PR #9, opened 2019-04-10
To avoid failure, warn customers to increase ES heap 

stevendanna:
This can be removed when the fix is in for 

https://chefio.atlassian.net/browse/MA-74
https://chefio.atlassian.net/browse/CLOUD-420
https://chefio.atlassian.net/browse/A2-288
...and so on...


stevendanna:
We now automatically configure Chef Automate to use 25% of available memory as ES heap.  While it is probably still valuable to document this (and all) configurables somewhere, I hope the new defaults mean that for most users we don't need such a prominent section in the install doc for it.

I'm closing this out, but we can reopen it if I'm mistaken.

PR #12, opened 2019-04-10
[A2-746] check if IAM beta flag set in middleware and pass along filtered projects

bcmdarroch:

This PR makes use of our newly updated policy version API, which now includes Major and Minor versions, in the gateway middleware. 

### :+1: Definition of Done

- [x] While on v1.0 or v2.0, projects in the project header have no effect on the returned data. Every API request returns all the things unfiltered
- [x] While on v2.1, projects in the project header get checked for authorization. Any authorized projects get sent along to the domain service in the auth context.
- [x] While on v2.1, the Projects API (or any other auth API with filtering) returns responses filtered by project 

1. rebuild services  and make some projects
```
rebuild components/authz-service && rebuild components/automate-gateway

# Top-level project is chef
# automate belongs to chef
# auth belongs to chef and automate

chef-automate dev psql chef_authz_service

insert into iam_projects (id, name, type, projects) values ('chef', 'Chef', 'custom', '{chef}');

insert into iam_projects (id, name, type, projects) values ('automate', 'Automate', 'custom', '{chef, automate}');

insert into iam_projects (id, name, type, projects) values ('auth', 'Auth', 'custom', '{chef, automate, auth}');

insert into iam_projects (id, name, type, projects) values ('some-project', 'Some other project', 'custom', '{some-project}');
```

2. try out project filters on v1
```
export TOK=`chef-automate admin-token`

curl -H "api-token: $TOK"  -H "projects: chef, automate, auth" https://localhost/apis/iam/v2beta/projects --insecure | jq

# all 4 projects returned (project headers ignored)
```

3. try out project filters on v2
```

curl -H "api-token: $TOK"  -H "projects: chef, automate, auth" https://localhost/apis/iam/v2beta/projects --insecure | jq

# all 4 projects returned (project headers ignored)
```

4. upgrade to 2.1 and make a test token
```
export TEST=`chef-automate iam token create test`
```

5. make a policy for that token
```
# save as policy.json
{
  "name": "beta flag test",
  "id": "beta-test",
  "members": [ "token:test" ],
  "statements": [
    {
      "effect": "ALLOW",
      "actions": [ "iam:projects:list" ],
      "projects": [ "chef", "automate", "auth" ]
    },
   {
      "effect": "ALLOW",
      "actions": [ "iam:projects:get" ],
      "projects": [ "automate", "auth" ]
    },
   {
      "effect": "ALLOW",
      "actions": [ "iam:projects:update" ],
      "projects": [ "auth" ]
    },
   {
      "effect": "ALLOW",
      "actions": [ "iam:projects:delete" ],
      "projects": [ "auth" ]
    }
 ]
}

cat policy.json | chef-automate dev grpcurl authz-service -- -d @ chef.automate.domain.authz.v2.Policies.CreatePolicy
```

4. curl for projects with different project headers
```
curl -H "api-token: $TOK"  https://localhost/apis/iam/v2beta/projects --insecure | jq
# admin can see all the projects

curl -H "api-token: $TEST"  -H "projects: chef, automate, auth" https://localhost/apis/iam/v2beta/projects --insecure | jq
# only chef, automate, auth projects should return

curl -H "api-token: $TEST"  -H "projects: auth" https://localhost/apis/iam/v2beta/projects/auth --insecure | jq
# only auth project should return

curl -H "api-token: $TEST"  -H "projects: automate" -X DELETE https://localhost/apis/iam/v2beta/projects/automate --insecure | jq
# request should fail

curl -H "api-token: $TEST"  -H "projects: auth" -X PUT -d '{"name": "New Auth Name"}' https://localhost/apis/iam/v2beta/projects/auth --insecure | jq
# request should succeed

curl -H "api-token: $TEST"  -H "projects: automate, auth, not-real" https://localhost/apis/iam/v2beta/projects --insecure | jq
# not-real project gets ignored

curl -sSkH "api-token: $TEST"  -H "projects: " https://localhost/apis/iam/v2beta/projects | jq -c '.projects | .[]'
# Empty projects header--should return all allowed ones

curl -sSkH "api-token: $TEST"  https://localhost/apis/iam/v2beta/projects | jq -c '.projects | .[]'
# no projects header--should return all allowed ones
```

5. revert back to v2
```

curl -H "api-token: $TEST"  -H "projects: chef, automate, auth" https://localhost/apis/iam/v2beta/projects --insecure | jq
# all projects returned
# v2.1 policies with projects become more permissive on v2 since projects get ignored
```

- [ ] Code actually executed?


components/automate-gateway/gateway/middleware/authv2/authv2.go
```diff
@@ -54,65 +54,31 @@ func (c *client) Handle(ctx context.Context, subjects []string, projectsToFilter
...
 	})
 	if err != nil {
 		if status.Convert(err).Code() == codes.FailedPrecondition {
 			return nil, err
 		}
 		log.WithError(err).Error("error authorizing request")
 		return nil, status.Errorf(codes.PermissionDenied,
-			"error authorizing action %q on resource %q for subjects %q: %s",
-			action, resource, subjects, err.Error())
+			"error authorizing action %q on resource %q for subjects %q filtered by project list %q: %s",
```
bcmdarroch: question: should we drop references to projects in these error messages so they don't leak into v2 responses? we could also maybe push down the error handling into ProjectsAuthorized so it returns different errors based on the version.

components/authz-service/server/v2/authz.go
```diff
@@ -60,9 +66,18 @@ func (s *authzServer) ProjectsAuthorized(
...
 		return nil, status.Error(codes.Internal, err.Error())
 	}
-	// Generally we return the engine's response verbatim
-	// but there are two cases that need to be intercepted and adjusted.
-	if stringutils.SliceContains(projectsAuthorized, constants.AllProjectsID) {
+
+	// we check the version set in the channel on policy server
+	// in order to determine whether or not to filter projects for the request
+	version := s.vSwitch.version
+	if version.Minor == api.Version_V0 && len(projectsAuthorized) > 0 {
```
bcmdarroch: i'm still mulling over the behavior here. i ran into this case while testing:

say i upgrade to v2.1 and create this policy:
```
{
  "name": "beta flag test",
  "id": "beta-test",
  "type": "CUSTOM",
  "members": [
    "token:test"
  ],
  "statements": [
    {
      "resources": [
        "*"
      ],
      "actions": [
        "iam:teams:list"
      ],
      "projects": [
        "chef",
        "automate",
        "auth"
      ]
    }
  ]
}
```

then this request always returns only `auth`, `automate`, `chef` teams
```
curl -H "api-token: $TEST" https://localhost/apis/iam/v2beta/teams --insecure | jq

{
  "teams": [
    {
      "name": "Chef",
      "id": "chef",
      "projects": [
        "chef"
      ]
    },
    {
      "name": "Automate",
      "id": "automate",
      "projects": [
        "chef",
        "automate"
      ]
    },
    {
      "name": "Auth Team",
      "id": "auth",
      "projects": [
        "chef",
        "automate",
        "auth"
      ]
    }
  ]
}
```

if i then downgrade back to v2, project filtering should no longer work. 

so my question: should this v2.1 policy apply at all? should we expand it to allow all teams? or should return no teams? 

here i've got it allowing access to ALL teams if any allowed project is returned, but now i'm thinking it might be safer to assume less permission.

susanev: wait, why does ur example 2.1 policy have `"resources": ["*"]`?

i dont really know the best approach here, maybe a good post standup convo for tomorrow i think there are pros/cons of both approaches.

components/authz-service/server/v2/authz.go
```diff
@@ -3,30 +3,36 @@ package v2
...
+	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
 	"github.com/chef/automate/lib/logger"
 	"github.com/chef/automate/lib/stringutils"
 
+	"github.com/chef/automate/api/interservice/authz/common"
 	api "github.com/chef/automate/api/interservice/authz/v2"
+	api_v2 "github.com/chef/automate/api/interservice/authz/v2"
```
msorens: This is a duplicate of the line above; condense to a single alias.

components/authz-service/server/v2/authz.go
```diff
@@ -59,8 +59,8 @@ func (s *authzServer) ProjectsAuthorized(
 	req *api.ProjectsAuthorizedReq) (*api.ProjectsAuthorizedResp, error) {
```
msorens: The first thing is this function is this:
```
projectsAuthorized, err := s.engine.V2ProjectsAuthorized(ctx,
		engine.Subjects(req.Subjects),
		engine.Action(req.Action),
		engine.Resource(req.Resource),
		engine.ProjectList(req.ProjectsFilter...))
```
But remember the above is being used for both v2.0 and v2.1. So you need to check the version before this call, so that you pass `AllProjectsID` instead of `req.ProjectsFilter` as the last param above. That way, if the system is on v2.0 and the user does include project headers, it will ignore them and ask for all projects in the above call.

components/authz-service/server/server.go
```diff
@@ -16,6 +14,8 @@ import (
 	"google.golang.org/grpc/reflection"
 	"google.golang.org/grpc/status"
 
+	automate_event "github.com/chef/automate/api/interservice/event"
```
msorens: test

components/authz-service/server/v2/authz.go
```diff
@@ -60,9 +66,18 @@ func (s *authzServer) ProjectsAuthorized(
 	if err != nil {
```
msorens: The very first thing in this function (not shown) is this:

projectsAuthorized, err := s.engine.V2ProjectsAuthorized(ctx,
		engine.Subjects(req.Subjects),
		engine.Action(req.Action),
		engine.Resource(req.Resource),
		engine.ProjectList(req.ProjectsFilter...))

But remember the above is being used for both v2.0 and v2.1. If it is v2.0, then you should pass `AllProjectsID` instead of `req.ProjectsFilter` as the last param above. That way, if the system is on v2.0 and the user does include project headers, it will ignore them and ask for all projects in the above call.

components/authz-service/server/v2/authz.go
msorens: Going from v2.1 to v2.0 is effectively saying "ignore projects" so IMO allowing access to all teams is the correct choice.

components/authz-service/server/v2/authz.go
```diff
@@ -148,3 +163,66 @@ func toEnginePairs(pairs []*api.Pair) []engine.Pair {
...
+	handler grpc.UnaryHandler) (interface{}, error) {
+	// For Authorization calls (IsAuthorized, FilterAuthorizedPairs), we decline
+	// cross-overs.
+	//
+	// Note: v2 policy related calls have their own service, so, for example, the
+	// endpoint for retrieving whether IAMv1 or v2 is used, GetPolicyVersion, is
+	// "/chef.automate.domain.authz.v2.Policies/GetPolicyVersion", and thus
+	// exempt from this version check.
+
+	// These methods skip the check, thought they are in the relevant service
```
msorens: s/thought/though/

components/authz-service/server/v2/authz_test.go
```diff
@@ -105,6 +105,68 @@ func TestProjectsAuthorized(t *testing.T) {
 	})
 }
 
+func TestV2ProjectsAuthorized(t *testing.T) {
```
msorens: Technically v2 is a beta, too, so instead of "TestV2ProjectsAuthorized" and "TestV2BetaProjectsAuthorized"
perhaps just "TestV2ProjectsAuthorized" and "TestV2_1ProjectsAuthorized" ?

components/authz-service/server/v2/policy.go
```diff
@@ -399,13 +408,6 @@ func (s *policyServer) CreateRole(
 func (s *policyServer) ListRoles(ctx context.Context,
 	_ *api.ListRolesReq) (*api.ListRolesResp, error) {
 
-	// TODO (TC): This is an example of how we can get the projects in the domain to filter on.
-	// projects, err := auth_context.ProjectsFromIncomingContext(ctx)
-	// if err != nil {
-	// 	return nil, status.Error(codes.Internal, err.Error())
-	// }
-	// s.log.Infof("PROJECTS: here are the authorized projects for this request: %s", projects)
-
```
msorens: Note to self: the above is handled closer to the source in final implementation (postgres.go::ListRoles)

components/authz-service/server/v2/policy.go
```diff
@@ -101,8 +100,18 @@ func NewPoliciesServer(
...
+	case storage.SuccessfulBeta1:
+		v = api.Version{Major: api.Version_V2, Minor: api.Version_V1}
+	case storage.Successful:
+		v = api.Version{Major: api.Version_V2, Minor: api.Version_V0}
+	default:
+		v = api.Version{Major: api.Version_V1, Minor: api.Version_V0}
+	}
+	srv.setVersion(v)
+
+	isV2 := v.Major == api.Version_V2
```
msorens: Suggest inlining the value of `isV2` and getting rid of the extra variable.

components/authz-service/server/v2/policy.go
```diff
@@ -101,8 +100,18 @@ func NewPoliciesServer(
...
+	var v api.Version
+	switch ms {
+	case storage.SuccessfulBeta1:
+		v = api.Version{Major: api.Version_V2, Minor: api.Version_V1}
+	case storage.Successful:
+		v = api.Version{Major: api.Version_V2, Minor: api.Version_V0}
+	default:
+		v = api.Version{Major: api.Version_V1, Minor: api.Version_V0}
+	}
+	srv.setVersion(v)
```
msorens: Well done on all the setVersion instrumentation in this file! 👍

components/automate-cli/cmd/chef-automate/iam.go
```diff
@@ -420,7 +420,7 @@ func runCreateTokenCmd(cmd *cobra.Command, args []string) error {
 		Projects: []string{},
 	})
 	if err != nil {
-		return status.Wrap(err, status.APIError, "Failed to generate new admin token.")
+		return status.Wrap(err, status.APIError, "Failed to generate new token.")
```
msorens: Good catch! However, if that changes I suspect the constant `adminTokenIAMPreconditionError` should also change.

components/automate-gateway/gateway/middleware/auth_interceptor.go
```diff
@@ -98,8 +99,9 @@ type AnnotatedAuthorizationResponse interface {
 }
 
 type authInterceptor struct {
-	authn authn.AuthenticationClient
-	authz GRPCAuthorizationHandler
+	authn  authn.AuthenticationClient
+	authz  GRPCAuthorizationHandler
+	policy v2.PoliciesClient
```
msorens: I think this change is no longer applicable...?

components/automate-gateway/gateway/middleware/authv2/authv2.go
msorens: The former at least (dropping references). The latter would be nice, though.

components/authz-service/server/v2/authz_test.go
bcmdarroch: this will be a little inconsistent with our migration status naming (`Successful` for v2, `SuccessfulBeta1` for v2.1, but i think i'll start using `p` in place of the decimal in camelcase names like google does: https://cloud.google.com/apis/design/versioning

components/automate-gateway/gateway/middleware/auth_interceptor.go
bcmdarroch: ha i deleted this like 5 times 😅

components/automate-gateway/gateway/middleware/authv2/authv2.go
bcmdarroch: i'll drop it for now 👍

components/automate-cli/cmd/chef-automate/iam.go
bcmdarroch: I checked, that error only triggers when you pass the `--admin` flag on v1. otherwise `chef-automate iam token create` works the same on v1 and v2

components/automate-gateway/gateway/middleware/authv2/authv2.go
```diff
@@ -82,37 +82,31 @@ func (c *client) Handle(ctx context.Context, subjects []string, projectsToFilter
...
+			log.WithError(err).Error("error authorizing request")
+			return nil, status.Errorf(codes.PermissionDenied,
+				"error authorizing action %q on resource %q for subjects %q filtered by project list %q: %s",
+				action, resource, subjects, projectsToFilter, err.Error())
+		}
+		if len(filteredResp.Projects) == 0 {
+			return nil, status.Errorf(codes.PermissionDenied,
+				"unauthorized: subjects %q has no project access for action %q on resource %q "+
+					"(filtered by project list %q)",
+				action, resource, subjects, projectsToFilter)
```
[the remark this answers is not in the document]
msorens: Looks like you overlooked this one--error message still says
```
unauthorized: subjects \"iam:projects:delete\" has no project access for action \"iam:projects:automate\" on resource [\"token:test\"]
```

components/authz-service/server/v2/authz.go
```diff
@@ -51,18 +56,35 @@ func (s *authzServer) IsAuthorized(
...
+		// if IAM version is set to v2.0
+		// as long as at least one project is authorized
+		// we override the filtered projects because no filter should be applied on v2
+		projectsAuthorized = []string{constants.AllProjectsExternalID}
+	} else if stringutils.SliceContains(projectsAuthorized, constants.AllProjectsID) {
+		// Generally we return the engine's response verbatim
+		// but there are two cases that need to be intercepted and adjusted.
 		if len(req.ProjectsFilter) == 0 {
 			// Engine allows all and we requested all, so signify it as all.
 			// This must be different than the requested notion of all,
```
msorens: Not shown, but I found a minor bug that I introduced:
L97 contains a call to `s.logProjectQuery`. That would have likely been quite helpful to you during development--if I had put it in the right place! It should not be nested in an if-then-else, so please move it down two lines, to immediately before the `return` in L100. With that in the right place, running your battery of "use cases" in the PR preamble shows this in the supervisor logs (truncated to fit here). Note the `projects` shows the _requested projects_ and `result` shows the _allowed projects_.
```
projects="[automate auth not-real]" resource="iam:projects" result="[automate auth]" subject="[token:test]"
projects="[]" resource="system:status" result="[*]" subject="[team:local:admins user:local:admin]"
projects="[auth]" resource="iam:projects:auth" result="[auth]" subject="[token:test]"
projects="[automate]" resource="iam:projects:automate" result="[]" subject="[token:test]"
```

components/authz-service/server/server.go
```diff
@@ -60,18 +60,18 @@ func NewGRPCServer(ctx context.Context,
 	dataMigrationsConfig datamigration.Config,
 	eventServiceAddress string) (*grpc.Server, error) {
 
-	// Note(sr): we're buffering one bool, as NewPostgresPolicyServer writes
+	// Note(sr): we're buffering one version struct, as NewPostgresPolicyServer writes
 	// to this before we've got readers
-	v2Chan := make(chan bool, 1)
-	switcher := NewSwitch(v2Chan)
+	vChan := make(chan api_v2.Version, 1)
+	switcher := v2.NewSwitch(vChan)
```
tylercloke: Did this still need to switch from a bool at the end of the day? It's no problem, just curious.

components/authz-service/server/v2/authz.go
```diff
@@ -51,18 +56,35 @@ func (s *authzServer) IsAuthorized(
 func (s *authzServer) ProjectsAuthorized(
 	ctx context.Context,
 	req *api.ProjectsAuthorizedReq) (*api.ProjectsAuthorizedResp, error) {
+	// we check the version set in the channel on policy server
+	// in order to determine whether or not to filter projects for the request
+	version := s.vSwitch.Version
+	var projects []string
+	if version.Minor == api.Version_V0 {
+		// if IAM version is set to v2.0
+		// we override the requested projects because no filter should be applied on v2
```
tylercloke: s/should be applied on v2/should be applied on v2.0

components/automate-gateway/gateway/middleware/authv2/authv2.go
```diff
@@ -54,65 +54,31 @@ func (c *client) Handle(ctx context.Context, subjects []string, projectsToFilter
...
-	// }
-	// if len(resp.Projects) == 0 {
-	// 	return nil, status.Errorf(codes.PermissionDenied,
-	// 		"unauthorized: subjects %q has no project access for action %q on resource %q "+
-	// 			"(filtered by project list %q)",
-	// 		action, resource, subjects, projectsToFilter)
-	// }
-	// projects = resp.Projects
-	// }
+	projects := filteredResp.Projects
```
tylercloke: I like how clean this code ended up being! Nice to keep as many auth implementation details as possible within the auth service!

components/authz-service/server/server.go
bcmdarroch: yeah we need to know the minor version so ProjectsAuthorized can decide if it's on v2 and should override projects in the request.

PR #13, opened 2019-04-10
Adds draft content

stevendanna:


stevendanna:
@kagarmoe @ryancragun I don't think this is a model we want users consuming anymore so it isn't clear to me we should be documenting this method.  We can reopen it if I'm wrong.

PR #14, opened 2019-04-10
[Workflow] Refresh session token for workflow UI

stevendanna:
To keep the session alive on A2 token get refreshed every minute. But for the workflow, there is no way to refresh to the session token.

### :+1: Definition of Done

 - Include jwt-decode npm library.
 - To make the session token consistent for both Automate 2 UI and workflow UI added refresh token logic to Workflow as well.
- Completed https://github.com/chef/a1/tree/rhass/fix-session-logout

* Install Automate 2.0
```
curl https://packages.chef.io/files/current/latest/chef-automate-cli/chef-automate_linux_amd64.zip | gunzip - > chef-automate && chmod +x chef-automate && \
./chef-automate deploy --channel dev --skip-preflight --accept-terms-and-mlsa --enable-chef-server --enable-workflow
```
* Login to Workflow.
* Stay on the dashboard page.
* After about two minutes the session will logout.

- [ ] Code actually executed?

### DIFF
--- a/components/automate-workflow-web/package.json
+++ b/components/automate-workflow-web/package.json
@@ -32,6 +32,7 @@
     "eventsource-polyfill": "^0.9.6",
     "font-awesome": "=4.7.0",
     "highlight.js": "=8.3.0",
+    "jwt-decode": "^2.2.0",
     "keymaster": "=1.6.2",
     "lodash": "=4.2.1",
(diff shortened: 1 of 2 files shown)


mattlqx:
Any idea on when this will be merged? Workflow on A2 is basically unusable without this.

PR #59, opened 2019-04-14
[workflow-ctl] Remove more unused commands

stevendanna:

This PR removes the following commands and their dependencies from workflow-ctl:

```
cleanse
data-summary
delete-node
doctor
graceful-kill
hup
int
kill
node-summary
once
reconfigure
restart
service-list
show-config
start
status
stop
tail
term
uninstall
usr1
usr2
```

As part of those removals it:

- Replaces `omnibus-ctl` with a small stub that has all of the functionality we were actually using from omnibus-ctl.
- Removes `lib/setup` which was used by the already removed `setup` command
- Renames most uses of `automate-ctl` to `workflow-ctl` inside the code

One of the riskier changes here is that the previous code was assuming omnibus-ctl
style argument ordering (where the first couple arguments are omnibus-ctl config) whereas now
we assume that we care about all arguments.  This is risky because many of the remaining
commands still do not work and thus it is hard to test that the modifications are correct.

### :+1: Definition of Done

Fewer commands that we don't support are available to workflow-ctl users.

- [ ] Code actually executed?


components/automate-workflow-ctl/Gemfile.lock
```diff
@@ -50,70 +50,103 @@ GEM
...
       multipart-post (>= 1.2, < 3)
-    ffi (1.9.18)
+    faraday-cookie_jar (0.0.6)
+      faraday (>= 0.7.4)
+      http-cookie (~> 1.0.0)
+    ffi (1.10.0)
     ffi-yajl (2.3.1)
       libyajl2 (~> 1.2)
     fuzzyurl (0.9.0)
+    google-api-client (0.23.9)
```
stevendanna: The new version of train unfortunately pulls in a lot of stuff

yzl:
I'm good with removing commands we don't support. Is there anything we might want to do to steer people to the automate equivalents of commands where they exist, or is that situation unlikely to come up?

stevendanna:
@yzl I've pushed a new commit that adds a message if you try to call one of the deprecated commands.  It looks like one of the following:

```
# workflow-ctl delete-node                                                                                                                                                                                                                
This command is not supported by workflow-ctl. workflow-ctl only
supports commands relevant to the Workflow componenent of Chef
Automate.

See

    chef-automate --help

for more details.
```

or, for commands with alternatives:

```
# workflow-ctl show-config
This command is not supported by workflow-ctl. workflow-ctl only
supports commands relevant to the Workflow component of Chef
Automate.

Did you mean: chef-automate config show
```

components/automate-workflow-ctl/Gemfile.lock
```diff
@@ -50,70 +50,103 @@ GEM
...
+      mime-types-data (~> 3.2015)
+    mime-types-data (3.2019.0331)
+    mixlib-archive (0.4.20)
       mixlib-log
     mixlib-authentication (1.4.2)
     mixlib-cli (1.7.0)
-    mixlib-config (2.2.4)
-    mixlib-install (3.3.1)
+    mixlib-config (2.2.18)
+      tomlrb
```
karendolan: FYI - upgrading maxlib-config from 2.2.4 to 2.2.18 stops it from being able to work with Ruby < v2.2 
https://github.com/chef/mixlib-config/commit/988c13aaecfae5afb7d4961a6181d8e6057a0a71

Issue #60, opened 2019-03-07
handle project rule updates for ingested resources


vjeffrey:
# User Story
when a user updates rules for a project or creates a new project, we need to re-evaluate all ingested nodes and actions to update project tagging.
in the spike, @lancewf demonstrated this could be done by running a painless script in elastic search that takes a list of project rules and evaluates/re-tags ingested nodes according to the new project rules.

that takes care of the actual re-tagging process.

the other thing that needs to be done is event handling around project rule updates.
@lancewf has a design for this:

<img width="679" alt="screen shot 2019-03-06 at 10 25 42 pm" src="https://user-images.githubusercontent.com/10341541/53934163-db307400-405e-11e9-80cf-0ff46522b2c6.png">

so, following the above design + some conversations we've had around message durability, the flow is:
- user creates one or more new projects or updates projects
- auth service triggers `rule update` event
- ingest service and compliance ingestion are listening for the `rule update` event
- upon receipt of the `rule update` event, ingest service and compliance ingestion make a call to elastic to start the painless script. this is an async call that returns a process id for later querying.
- ingest service and compliance ingestion trigger `rule update received` event
- auth is listening for the `rule update received` event. if that event is not received within a timely manner (tbd) after sending the `rule update` event, the auth service should retry
- when ingest service completes the update, it triggers `client runs update complete` event
- when compliance ingestion completes the update, it triggers `compliance update complete` event
- auth is listening for the `compliance update complete` event

_Notes: there will be many gotchas in this implementation. how to handle duplicate messages, what should the auth service do if it never receives a complete event from one of the services, etc. we are expecting these issues to crop up and expect to spend a couple weeks adjusting as needed_

# Definition of Done
a project rule update in auth triggers the event flow to start the painless script

# Demo Script / Repro Steps
update a project or create a new one
ensure the expected data is tagged with the correct projects 

# More Information
there have been concerns about data durability when using the event service. until we can build off the work currently being done in the event service for durability (NATS), we are pushing the data durability problem to the services sending events to the event service. this has been discussed at length with @gpeers and she is on board.

# Sub Tasks
- [x] Add project tag update for client runs nodes
- [x] Add project tag update for compliance reports
- [x] Add an authz-service manager to manage the update process with both compliance and client runs.

lancewf:
Future feature to update indexes individually from the most recent to the oldest. This will allow the system to resume from an elasticsearch restart. This will also allow the user to see changes to most recent data faster.

[...]

PR #86, opened 2019-04-16
Adding Compliance summary indexes to the project tag update

lancewf:

Adding compliance summary indexes to the project tag update process. Currently, the project tag updater only updates the report indexes. This change adds the summary indexes to the update process.  

1. `build components/compliance-service`
1. Add a project (https://github.com/chef/a2/blob/master/components/authz-service/server/v2/policy.go#L702) with rules using the `Environment` from above. Example below
```
	rules := []*rule{
		{
			ID:     "project9",
			Type:   "ChefEnvironment",
			Values: []string{"DevSec Prod Beta"},
		},
	}
```
1. `rebuild components/authz-service`
1. `start_all_services`
1. run `chef-automate dev grpcurl authz-service -- chef.automate.domain.authz.v2.Projects.ListProjectRules` to ensure the project rule was added. 
1. Remove the `job_uuid` field from the `` file components/compliance-service/ingest/examples/compliance-success-tiny-report.json (job_uuid is for scan jobs which are not tagged in ingestion). 
1. Send an InSpec report `send_inspec_example`
1. Ensure projects on the summary is tagged with 'project9' `curl $ELASTICSEARCH_URL/comp-3-s-*/_search | jq . | less` 
1. Update the project ID in authz like above to 'project3'
1. Create a project through the API `chef-automate dev grpcurl authz-service -- chef.automate.domain.authz.v2.Projects.CreateProject -d '{"name": "project3", "id": "project3"}'`
1. Update the project to trigger the compliance update process. `chef-automate dev grpcurl authz-service -- chef.automate.domain.authz.v2.Projects.UpdateProject -d '{"name": "project3", "id": "project3"}`
1. Ensure projects on the summary is tagged with 'project3' with this command `curl $ELASTICSEARCH_URL/comp-3-s-*/_search | jq . | less`. It should look like below. 
```
          "projects": [
            "project3"
          ],
```

https://github.com/chef/automate/issues/60

- [x] Code actually executed?


components/compliance-service/ingest/projectupdater/manager.go
```diff
@@ -177,12 +181,34 @@ func (manager *Manager) waitingForJobToComplete() {
...
+		PercentageComplete: 1.0,
+	}
+	for _, esJobID := range manager.esJobIDs {
+		jobStatus, err := manager.client.JobStatus(context.Background(), esJobID)
+		if err != nil {
+			return jobStatus, err
+		}
+
+		if !jobStatus.Completed {
+			combinedJobStatus.Completed = false
```
phiggins: This feels wrong to me. Doesn't this mean the first job in the array could be still going and the second one is completed, then the return value would be that both of them are completed?

components/compliance-service/ingest/projectupdater/manager.go
```diff
@@ -157,15 +164,12 @@ func (manager *Manager) waitingForJobToComplete() {
...
+		jobStatus, err = manager.combinedJobStatus()
 		if err != nil {
 			logrus.Errorf("Failed to check the running job: %v", err)
 			numberOfConsecutiveFails++
 			if numberOfConsecutiveFails > maxNumberOfConsecutiveFails {
-				logrus.Errorf("Failed to check Elasticsearch job %q %d times",
-					manager.esJobID, numberOfConsecutiveFails)
-				manager.sendFaildEvent(fmt.Sprintf("Failed to check Elasticsearch job %q %d times",
-					manager.esJobID, numberOfConsecutiveFails), manager.projectUpdateID)
+				manager.sendFaildEvent(err.Error(), manager.projectUpdateID)
```
vjeffrey: this is a total nitpick..but why "Faild"  and not "Failed"

components/compliance-service/ingest/projectupdater/manager.go
lancewf: I will add some tests to this function

lancewf: @phiggins here is a test for that case now https://github.com/chef/automate/pull/86/files#diff-5bef6c2a36a3b9fb536231672b4dc731R82

components/compliance-service/ingest/projectupdater/manager.go
```diff
@@ -144,47 +154,76 @@ func (manager *Manager) waitingForJobToComplete() {
...
 		numberOfConsecutiveFails++
 	} else {
-		manager.sendStatusEvent(jobStatus)
-		isJobComplete = jobStatus.Completed
+		mergedJobStatus := MergeJobStatus(jobStatuses)
+		isJobComplete = mergedJobStatus.Completed
+		manager.updateStatus(mergedJobStatus)
 	}
 
 	for !isJobComplete {
```
phiggins: nitpick: I think you could reduce a little bit of complexity here by not using `isJobComplete`, moving `mergedJobStatus` to the outer scope, and changing this condition to `!mergedJobStatus.Completed`.

components/compliance-service/ingest/projectupdater/manager.go
```diff
@@ -30,7 +30,7 @@ const (
 type Manager struct {
 	state                 string
 	projectUpdateID       string
-	esJobID               string
+	esJobIDs              []string
```
phiggins: Are we expecting more things to be added to this? Using a list for two things might be a bit of premature abstraction, IMO.

components/compliance-service/ingest/projectupdater/manager.go
lancewf: That is a good point! Thanks

components/compliance-service/ingest/projectupdater/manager.go
phiggins: I mean this in reference to other things introduced in this PR, not just this value specifically.

lancewf: Would you think it would be better to list out both types of job IDs like below? 
```
esReportJobID string
esSummaryJobID string
```

lancewf: I don't know of anything other jobs that are going to be added to this.

phiggins: That's what I was thinking, looping over a list that's only going to ever be two things seems a bit overkill.

components/compliance-service/ingest/projectupdater/manager.go
lancewf: I update it with this commit. https://github.com/chef/automate/pull/86/commits/6d58c39cc7d950f0538c62a141dda27096a3688b

components/compliance-service/ingest/projectupdater/manager.go
lancewf: I am going to leave it with the slice of IDs for now. We are still working on this code, so we can find a better solution later.

PR #90, opened 2019-04-16
[deployment] Represent waiting states in converger

stevendanna:
# Motivation

This PR started as a fix for the SupervisorUpgrade phase. That phase
was not restarting our systemd unit when required. Fixing that problem
revealed a deadlock in our systemd restart sequence. Namely,

- The Converge function would call `systemctl restart chef-automate`

- `systemctl restart chef-automate` would call the Stop() grpc call

- The Stop grpc call requires both the deployment lock (which the
  periodic converger likely held) and a responsive converger (which is
  unlikely as it is blocked waiting for `systemctl restart` to
  finish).

We solved the most immediate issue by passing the `--no-block` flag to
`systemctl restart` which allowed the converger/compiler to move on,
releasing the lock and responding to our Stop() request.

However, we were now faced with one of our common fears:

> How safe is it for us to be restarted in the middle of the compiler?

Previously we've solve this via three strategies:

- Making the compiler stop if it detected that a restart or
  reconfigure should be coming,

- Fixing bugs in `hab` that made killing commands unsafe, and

- Ad-hoc state mechanisms to prevent the compiler from moving forward.

While we could have added a new ad-hoc mechanism here, we decided we
wanted to model our waiting states in a small state machine.

# Converger State Machine

During upgrades and reconfiguration, we sometimes expect an external
process (hab-sup or systemd) to send our process a signal that will
result in our processes restarting. When this occurs, we want to wait
for that signal to arrive rather than starting new work that could be
interrupted.

To do that, the converger is now a small state machine that looks like
the following:

```
        +-----------+  convergeRequest           +-------------------+
        |           |  err = ErrRestartPending   |                   |
        |           +--------------------------->|                   |
        |    idle   |                            | waitingForRestart |
        |           |<---------------------------+                   |
        |           |       timeout              |                   |
        +-------+---+                            +-------------------+
            ^   |
            |   | convergeRequest
   timeout 2|   | err = ErrSelfReconfigurePending
            |   |
            |   v
         +--+--------------------+
         |                       | timeout 1
         |                       +---------+
         | waitingForReconfigure |         |
         |                       |<--------+
         |                       |
         +-----------------------+
```

Note that this state machine is expressing what I believe were the
previous semantics of the code.

# Converger version removal

Previously the Converge() request passed along a version. The idea of
this version was to make sure we weren't making changed based on
outdated deployment configuration.  I think this is a great idea but:

1. We weren't using it anywhere, and
2. For it to work we would have to centralize mutations to the
   deployment struct.

Thus, I've removed it for now.

# Debug Interface

The converger now has a debug channel where test code can get and set
the current state of the state machine. An alternative to this
approach would be to use a small mutex around the converger state.

# Follow-up Work

- Backup and Restore should probably be wired into this, however we
  want a week where we don't change the backup/restore code so we are
  deferring that work.

- Re-add the converge version more systematically.

- We really want to rename `converger` to `SOMETHING` and `compiler`
  to `converger` but I didn't do that here.


components/automate-deployment/pkg/converge/converger.go
```diff
@@ -127,23 +121,315 @@ func StartConverger(opts ...ConvergerOpt) Converger {
...
+}
+
+func (t *waitingForRestart) Timeout() time.Duration { return 600 * time.Second }
+func (t *waitingForRestart) String() string         { return "waiting for restart" }
+
+// waitingForReconfigure is a converer state that we enter whenever we
+// expect to have received a reconfiguration request from the Habitat
+// supervisor.
+type waitingForReconfigure struct {
+	sentSelfHup bool
```
stevendanna: Maybe rather than having this boolean, we can have two states: `waitingForReconfigure1` and `waitingForReconfigure2`.

components/automate-deployment/pkg/converge/compiler.go
```diff
@@ -704,14 +665,38 @@ func (phase *SupervisorUpgradePhase) Run(w *eventWriter) error {
...
+		// our process tree is:
+		//
+		// systemd +
+		//         |- hab-launcher +
+		//                         |- hab-sup
+		//                         |- deployment-service
+		//                         | ... other automate services
+		//
+		var pendingRestart bool
+		if phase.systemdRestartRequired(systemdReloaded) {
```
stevendanna: This is the fix for the initial bug.

components/automate-deployment/pkg/converge/converger.go
```diff
@@ -1,12 +1,16 @@
 package converge
```
stevendanna: All the state machine changes are in this file, you probably need to expand it in github.

components/automate-deployment/pkg/converge/converger.go
```diff
@@ -127,63 +136,460 @@ func StartConverger(opts ...ConvergerOpt) Converger {
 	return converger
 }
 
+func (converger *converger) run() {
+	for {
+		select {
+		case d := <-converger.debug:
+			d.Do(converger)
+		default:
+		}
```
stevendanna: The idea here was to not starve debug messages. While we are only using this in the test currently, I thought it might be good to eventually have grpcCalls that let us use it in production.

stevendanna:
This change looks a bit bigger than it is because I was a bit comment happy.  I also wrote some unit tests that I'm skeptical that we want to keep around.

components/automate-deployment/pkg/converge/converger.go
```diff
@@ -127,63 +136,460 @@ func StartConverger(opts ...ConvergerOpt) Converger {
...
+func (converger *converger) debugSetState(newState state) {
+	done := make(chan struct{})
+	converger.debug <- debugSetReq{
+		doneChan: done,
+		newState: newState,
+	}
+	<-done
+}
+
+func copyState(old state) state {
```
stevendanna: I'm unsure if we really need to do this.  The goal here was to make sure that we the converger and debug caller were never really sharing state.

components/automate-deployment/pkg/converge/converger.go
```diff
@@ -17,44 +21,38 @@ const (
 	elasticsearchServiceName = "automate-elasticsearch"
 )
 
-type Converger interface {
-	Converge(uint64, *Task, DesiredState, EventSink) error
-	Stop()
-}
+var (
+	waitingForReconfigureTimeout = 30 * time.Second
```
jaym: seems a bit short. Given that this is an anomoly, maybe we can be more lenient

components/automate-deployment/pkg/converge/converger.go
```diff
@@ -127,63 +136,460 @@ func StartConverger(opts ...ConvergerOpt) Converger {
...
+	respChan := make(chan debugResp, 1)
+	converger.debug <- debugGetReq{
+		responseChan: respChan,
+	}
+	resp := <-respChan
+	return resp.state
+}
+
+func (converger *converger) debugSetState(newState state) {
+	done := make(chan struct{})
```
jaym: I'm going to call this a nit because it's test only code atm, but we should probably give up if the channel is full. Or use 1 channel

components/automate-deployment/pkg/converge/converger.go
```diff
@@ -127,63 +136,460 @@ func StartConverger(opts ...ConvergerOpt) Converger {
...
+
+// idle is the default state of the converger. In this state we
+// are ready to accept any user-sendable requests. The idle state
+// never times out.
+type idle struct{}
+
+func (i *idle) ProcessMessage(c *converger, msg message) (state, error) {
+	switch msg.(type) {
+	case *convergeRequest:
+		c := c.compiler
```
jaym: [nit] this is a little confusing in that it shadows another c you already have

components/automate-deployment/pkg/converge/converger.go
```diff
@@ -127,63 +136,460 @@ func StartConverger(opts ...ConvergerOpt) Converger {
...
+		return i, nil
+	case *timeout:
+		logrus.Warn("received timeout when in idle state.")
+		return i, nil
+	default:
+		return i, UnknownMessageError(msg, i)
+	}
+}
+
+func (i *idle) handleError(err error) (state, error) {
```
jaym: this isn't your fault, but it annoys me we use errors this way. maybe one day compiler(we should also rename that) could just return the next state or a real error?

components/automate-deployment/pkg/converge/converger.go
```diff
@@ -127,63 +136,460 @@ func StartConverger(opts ...ConvergerOpt) Converger {
...
+			return
+		case req := <-converger.inbox:
+			logrus.Debugf("got message %s while %s", req, converger.currentState)
+			nextState, err := converger.currentState.ProcessMessage(converger, req)
+			converger.setState(nextState, false)
+			req.SendResponse(err)
+		case d := <-converger.debug:
+			d.Do(converger)
+		case <-converger.currentState.TimeoutChan():
+			nextState, err := converger.currentState.ProcessMessage(
```
jaym: should we log here like we do when we get a message

components/automate-deployment/pkg/converge/converger.go
stevendanna: I agree it feels wrong that these are "errors" since these are expected outcomes of certain operations.

components/automate-deployment/pkg/converge/converger.go
stevendanna: By "use 1 channel" do you mean that these debug messages should go to `converger.inbox`? If so, then I definitely agree as that has a number of benefits.  I didn't do that here just because this was a late addition and I didn't want to clutter up the production code path.

I'm not sure I follow why we would want to give up on the debug message though?

components/automate-deployment/pkg/converge/converger.go
stevendanna: This was the previous timeout (sort of), but I agree we could probably make it longer. I guess the trade-off is that the longer we stay in this state the longer we are prevented from doing anything else.  That said, considering that we've seen production nodes where hab starts taking 30 seconds per command, perhaps we can bump this to at least 1 minute?

components/automate-deployment/pkg/converge/converger.go
jaym: i think it's fine. I was just thinking of the case where we have multiple debug messages waiting just to get serviced and we're in a long converge loop, but i think it's fine since we service them first

components/automate-deployment/pkg/converge/converger.go
```diff
@@ -127,63 +136,460 @@ func StartConverger(opts ...ConvergerOpt) Converger {
...
+		logrus.Warn("received timeout when in idle state.")
+		return i, nil
+	default:
+		return i, UnknownMessageError(msg, i)
+	}
+}
+
+func (i *idle) handleError(err error) (state, error) {
+	switch err {
+	case nil:
```
jaym: [nit] i think this is the default case?

components/automate-deployment/pkg/converge/converger.go
```diff
@@ -444,12 +447,11 @@ func (t *timeoutable) TimeoutChan() <-chan time.Time {
...
-func (t *timeoutable) StopTimer(timedout bool) {
+func (t *timeoutable) StopTimer() {
 	if t.timer == nil {
 		return
 	}
-
-	if !t.timer.Stop() && !timedout {
+	if !t.timer.Stop() {
 		<-t.timer.C
 	}
```
stevendanna: We could consider setting `t.timer` to `nil` here to avoid possibly trying to Stop it again.

components/automate-deployment/pkg/events/mock_event_sender.go
```diff
@@ -1,5 +1,4 @@
 // Code generated by mockery v1.0.0. DO NOT EDIT.
-
```
ryancragun: super glad I expanded this diff :rofl:

components/automate-deployment/pkg/converge/converger.go
```diff
@@ -127,63 +137,482 @@ func StartConverger(opts ...ConvergerOpt) Converger {
...
+	// String is a convenience method used for logging. states should
+	// have uniq string descriptors so we can more easily
+	// conditionally log.
+	String() string
+}
+
+// message represents the messages that can be sent to our
+// converger. Messages move the converger from one state to another,
+// possibly triggering some system change.
+type message interface {
```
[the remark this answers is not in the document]
stevendanna: I can go either way, I don't think it simplifies ProcessMessage in any fundamental way since you'd still have to have a switch that covers each message type, unless I'm missing something?

components/automate-deployment/pkg/converge/converger.go
```diff
@@ -127,63 +137,482 @@ func StartConverger(opts ...ConvergerOpt) Converger {
...
+// converger to run the converge loop.
+type convergeRequest struct {
+	eventSink    EventSink
+	desiredState DesiredState
+	doneChan     chan error
+}
+
+func (c *convergeRequest) SendResponse(err error) {
+	c.doneChan <- err
+	close(c.doneChan)
```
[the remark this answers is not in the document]
stevendanna: Interesting.  I'm going to need to look a bit at how this done channel is used.  I feel like for "done" channels, it is a pretty common API pattern in the standard library that the library closes the channel since closing the channel is the only way for the "done channel" to always return.  

This might _feel_ different because the "task" is passed in by the user, but the fact that you even have to pass the task in rather than it being returned by the converge call seems like a quirk of the API. Perhaps if we changed around how the "Task" worked it would be more clear from an API perspective that "you only get to read from this chan"

components/automate-deployment/pkg/converge/converger.go
stevendanna: We should revert this change I think.  It turns out that we hit this case nearly all of the time because reconfiguration rarely leads to a template being rendered in the case of the deployment service.  I'll change this back to 30s and we can follow up with a change that removes some of our most confusing reconfiguration logic.

Issue #83, opened 2019-04-15
modify automate code to work with license accept changes from inspec


vjeffrey:
# User Story

InSpec will soon require users to accept the license by using one of the following methods:
```
* Interactively agree to license (not practical here, I think)
* run with the option `--chef-license accept`
* create the file `/etc/chef/accepted_licenses/inspec` or `~/.chef/accepted_licenses/inspec` . The file may be empty.
```

This will be needed for all the inspec commands, such as `json`, `check`, `detect`, and `exec`.

This change will also be modifying the stdout format. Please see https://github.com/chef/license-acceptance/issues/23 for more information.

These changes will affect: profile upload, profile json-ing, detect jobs, and scan jobs.

# Definition of Done
Able to run Automate as always with the new license acceptance functionality.

A gem of the new InSpec functionality is available here: http://artifactory.chef.co:8081/webapp/#/artifacts/browse/tree/General/omnibus-gems-local/gems/inspec-4.0.6.gem

We've requested a hab pkg for testing as well.

clintoncwolfe:
https://github.com/chef/license-acceptance/pull/26 has merged; use the option `--chef-license accept-silent` to accept the license and preserve clean JSON output.

alexpop:
Thank you @clintoncwolfe.
What do you think of [this](https://github.com/chef/automate/pull/95) approach in Automate?

vjeffrey:
we will have to do something for the remote (ssm) jobs as well: https://github.com/chef/automate/blob/master/components/compliance-service/inspec-agent/remote/remote.go

vjeffrey:
https://chefio.slack.com/archives/CGP1CVDAL/p1555522812041400
```
tball [12:51]
@vj Y’all are kicking off the remote `inspec` command, correct? If so could you run `inspec --chef-license accept-no-persist` or `CHEF_LICENSE=accept-no-persist inspec`? Then you don’t have to worry about manually creating that file, and if the format for the file changes it won’t break you


vj [12:51]
oooh interesting. ok ya we could do that — thank you!!

tball [12:52]
`accept-no-persist` also won’t output anything on stdout, so it shouldn’t affect the JSON response
```

jasonbellack:
The required step of accepting the license seems to be interfering with AWS State Manager. When attempting to run aws-RunInspecChecks on an instance, I am receiving this error: 
[...]

PR #95, opened 2019-04-16
Accept chef license for inspec 4 via ENV variable

alexpop:
Tested with `chef/inspec/4.1.3/20190417211329`
With this change, we use ENV variable `CHEF_LICENSE=accept-no-persist` to prevent inspec from prompting for chef license acceptance

### DIFF
--- a/components/compliance-service/inspec/cli.go
+++ b/components/compliance-service/inspec/cli.go
@@ -4,13 +4,13 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"os"
 	"os/exec"
 	"strings"
 	"time"
 
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
 
@@ -32,8 +32,9 @@ const logSensitiveData = false
 
 func inspecShimEnv() map[string]string {
 	return map[string]string{
-		"HOME":   TmpDir,
-		"TMPDIR": TmpDir,
+		"HOME":         TmpDir,
+		"TMPDIR":       TmpDir,
+		"CHEF_LICENSE": "accept-no-persist",
 	}
 }
 
@@ -43,6 +44,14 @@ func isTimeoutSane(timeout time.Duration, max time.Duration) error {
 	}
 	if timeout > max {
 		return errors.New(fmt.Sprintf("Timeouts of more than %s should not be set for the inspec cli.", max))
+	}
+	return nil
+}
+
+func AcceptInspecLicense() error {
+	err := os.Setenv("CHEF_LICENSE", "accept-no-persist")
+	if err != nil {
+		return errors.Wrap(err, "Unable to set CHEF_LICENSE env variable")
 	}
 	return nil
 }
--- a/components/compliance-service/compliance.go
+++ b/components/compliance-service/compliance.go
@@ -99,6 +99,11 @@ func runHungJobs(ctx context.Context, scheduledJobsIds []string, schedulerServer
 func initBits(ctx context.Context, conf *config.Compliance) (db *pgdb.DB, connFactory *secureconn.Factory, esr relaxting.ES2Backend, statusSrv *statusserver.Server, err error) {
 	statusSrv = statusserver.New()
 
+	err = inspec.AcceptInspecLicense()
+	if err != nil {
+		return db, connFactory, esr, statusSrv, err
+	}
+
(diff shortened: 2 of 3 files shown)


vjeffrey:
i know you mentioned that the license acceptance littering the stdout only happens the first time this is run.  what's the plan for handling that littered stdout the first time the license acceptance stuff actually runs?

vjeffrey:
will this also work for our inspec shim env? https://github.com/chef/automate/blob/master/components/compliance-service/inspec/cli.go#L33

alexpop:
> what's the plan for handling that littered stdout the first time the license acceptance stuff actually runs?
By creating the `~/.chef/accepted_licenses/inspec`, inspec assumes the license has already been accepted. No confirmation is added to STDOUT 

I'll test the `inspecShimEnv` now...

components/compliance-service/inspec/cli.go
```diff
@@ -30,11 +30,12 @@ const shimBinName = "inspec_runner"
 // variables via debug logs. Leave to false for release.
 const logSensitiveData = false
 
-func inspecShimEnv() map[string]string {
+func inspecShimEnv() (map[string]string, error) {
+	err := AcceptInspecLicense(TmpDir)
```
phiggins: I think it'd make more sense to leave this function as it was and call `AcceptInspecLicense` explicitly where its needed.

components/compliance-service/inspec/cli.go
```diff
@@ -47,6 +48,21 @@ func isTimeoutSane(timeout time.Duration, max time.Duration) error {
...
+		if err != nil {
+			return errors.Wrap(err, "InSpec license mkdir")
+		}
+		_, err = os.Create(licenseDir + "/inspec")
+		if err != nil {
+			return errors.Wrap(err, "InSpec license create")
+		}
+	}
+	return nil
+}
```
phiggins: We know what `TmpDir` is when the app starts up and don't allow changing it afterwards (right?), so we could do this check once while the app is starting. That feels simpler code-wise and also feels like a better user experience in the case where something goes wrong (ie, the service fails to start rather than every one of these function calls failing later).

jaym:
i'm not 100% convinced this is the right way to go about it. Your service shoudn't really be writing outside of /hab/svc. Maybe there is an environment variable that can be set?

vjeffrey:
@jaym the possible options for license acceptance are detailed here: https://github.com/chef/automate/issues/83#issue-433394461

phiggins:
An ENV variable is one of the options and the function being modified in this PR is for setting the ENV for our inspec calls. That seems to be a much simpler solution (for this specific case, anyways).

vjeffrey:
> An ENV variable is one of the options 
it is? where is that detailed?

phiggins:
> Adds a new value that can be passed as a value to --chef-license or as an ENV var: accept-silent. 

https://github.com/chef/license-acceptance/pull/26

vjeffrey:
that command is for silencing the output. it is not for actually accepting the license.
do we know that accepting the license can also be done via env var?

vjeffrey:
ohh, or did i misunderstand and that will accept the license _and_ silence the output?

phiggins:
> ohh, or did i misunderstand and that will accept the license and silence the output?

From the updates to README in that PR that's what it sounds like to me.

phiggins:
I could be misunderstanding this too though.

vjeffrey:
we should confirm, but ya the readme does seem to imply that will work.

vjeffrey:
nice catch @phiggins !

components/compliance-service/inspec/cli.go
alexpop: Talked with Victoria about that yesterday. Thing is `AcceptInspecLicense` is needed everywhere where `inspecShimEnv` is used. I'm going to test a new approach that @tyler-ball mentioned yesterday ( setting env `CHEF_LICENSE=accept-no-persist` )

alexpop:
`CHEF_LICENSE=accept-no-persist` worked like a charm, no files written on disk anymore, compatibility with inspec 3 & 4. I updated the code of the PR, title and description.

I tested with the latest unstable inspec 4.1.3 build the scanning, profile operation(inspec json, archive) and the scripts(bash & powershell) we run remotely.

components/compliance-service/inspec/cli.go
alexpop: Not needed anymore as AcceptInspecLicense is only called when the service starts now. Thank you!

PR #97, opened 2019-04-16
List reports cleanup

phiggins:
Some tangentially related changes I had made in #53 but didn't include for clarity.

* I ran `gofmt -s -w -l` on `compliance-service`. Our linter has simplification turned off for some reason (that's the `-s`), even though `make lint` tells you to run it:
    ```
    reporting/relaxting/reports.go:711: File is not `gofmt`-ed with `-s` (gofmt)
    ```
* I rearranged some code in the report filtering to make it more clear where some values are used.
* After our meeting this morning I realized I was embedding the special value `*` in some tests, so I replaced that with the constant from `authz_service`

### DIFF
--- a/components/compliance-service/ingest/projectupdater/manager.go
+++ b/components/compliance-service/ingest/projectupdater/manager.go
@@ -194,12 +194,12 @@ func (manager *Manager) sendFaildEvent(msg string, projectUpdateID string) {
 		},
 		Data: &_struct.Struct{
 			Fields: map[string]*_struct.Value{
-				project_update_tags.ProjectUpdateIDTag: &_struct.Value{
+				project_update_tags.ProjectUpdateIDTag: {
 					Kind: &_struct.Value_StringValue{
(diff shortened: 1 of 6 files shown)


components/compliance-service/ingest/projectupdater/manager.go
```diff
@@ -194,12 +194,12 @@ func (manager *Manager) sendFaildEvent(msg string, projectUpdateID string) {
...
 		Data: &_struct.Struct{
 			Fields: map[string]*_struct.Value{
-				project_update_tags.ProjectUpdateIDTag: &_struct.Value{
+				project_update_tags.ProjectUpdateIDTag: {
 					Kind: &_struct.Value_StringValue{
 						StringValue: projectUpdateID,
 					},
 				},
-				"message": &_struct.Value{
+				"message": {
```
lancewf: I did not know you could remove the type. Nice!

PR #101, opened 2019-04-17
Support for unassigned reports in reporting server GetReport endpoint

phiggins:
Followup to #23 and #53. ~~Branched from #97 so that code is in here as well.~~

This updates the project work the GetReport endpoing in reporting server to handle access to unassigned reports as detailed in #66.

### DIFF
--- a/components/compliance-service/reporting/relaxting/reports.go
+++ b/components/compliance-service/reporting/relaxting/reports.go
@@ -823,8 +823,24 @@ func (backend ES2Backend) getFiltersQueryForDeepReport(reportId string,
 	boolQuery = boolQuery.Must(idsQuery)
 
 	if len(filters["projects"]) > 0 {
-		termQuery := elastic.NewTermsQuery("projects", stringArrayToInterfaceArray(filters["projects"])...)
-		boolQuery = boolQuery.Must(termQuery)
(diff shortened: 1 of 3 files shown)


phiggins:
> could we extract that to a separate function to reduce duplicated code?

I have some ideas on better abstractions for this interface, but I want to see how it looks when repeated in more places before I go too crazy with the refactoring.

components/compliance-service/reporting/relaxting/reports.go
```diff
@@ -823,8 +823,24 @@ func (backend ES2Backend) getFiltersQueryForDeepReport(reportId string,
...
-		boolQuery = boolQuery.Must(termQuery)
+		projectsQuery := elastic.NewBoolQuery()
+
+		if stringutils.SliceContains(filters["projects"], authzConstants.UnassignedProjectID) {
+			emptyProjectQuery := elastic.NewBoolQuery()
+			emptyProjectQuery.MustNot(elastic.NewExistsQuery("projects"))
+			projectsQuery.Should(emptyProjectQuery)
+		}
+
+		assignedProjectIds := stringutils.SliceFilter(filters["projects"], func(projectId string) bool {
```
lancewf: I like the SliceFilter function! I did not know we had these.

PR #102, opened 2019-04-17
[A2-747] Set up e2e project tests

msorens:

Without supporting UI infrastructure, e2e tests for projects cannot yet be done with Cypress,
so this uses standard inspec integration tests. I randomly selected roles as the target scenario.
If there is sentiment to be exhaustive, we could extend this to policies, teams, etc.

This is the first time I have created a new "section" of integration tests (i.e. the .sh and .yml files) so do take a look at those with careful scrutiny.

**2019.04.28 Update**
Rounded out the set of integration tests for roles including List, Get, Update, and Delete.
The integration tests are shown failing here, but once these other PRs merge the integration tests will pass:
- PR #197 [A2-786] Correct project intersection query for roles 
- PR #200 [A2-787] authz-service and teams-service db cleanup
- PR #135 [A2-771] AuthorizedProjects handles deny

### :+1: Definition of Done

e2e tests show that one can filter results by supplying various values of requested projects via the CLI.

- Once the PRs mentioned above merge, then:

- Rebuild authz-service and automate-gateway.

```

inspec exec inspec/a2-iam-v2p1-only-integration --controls=iam-v2-projects-1
```
image

Further, if you revert to v2.0 most of the tests will fail, because project filtering is then off the table.

- [x] Code actually executed?


inspec/a2-iam-v2-1-only-integration/controls/role_projects.rb
```diff
@@ -0,0 +1,134 @@
...
+  CUSTOM_PROJECT_1 = {
+          id: CUSTOM_PROJECT_ID_1,
+          name: "Test Project 1"
+  }
+  CUSTOM_PROJECT_2 = {
+          id: CUSTOM_PROJECT_ID_2,
+          name: "Test Project 2"
+  }
+
+  Projects = [ CUSTOM_PROJECT_1, CUSTOM_PROJECT_2 ]
```
msorens: Have not seen the pattern in our integration tests before, but I am less fluent in ruby so I can be fearless in trying to cover new ground. The advantage of defining all the above is that the `before-all` and `after-all` below are quite clean and tidy.

integration/tests/iam_v2_1_only.sh
```diff
@@ -0,0 +1,45 @@
+#shellcheck disable=SC2034
```
msorens: This file was copied almost verbatim from existing code.

integration/tests/iam_v2_1_only.sh
```diff
@@ -0,0 +1,45 @@
...
+
+do_deploy() {
+    do_deploy_default
+    log_info "applying dev license"
+    chef-automate license apply "$A2_LICENSE"
+}
+
+do_test_deploy() {
+    log_info "run chef-automate iam upgrade-to-v2 --beta2.1"
+    chef-automate iam upgrade-to-v2 --beta2.1 || return 1
```
msorens: Here's my two lines of customization.

inspec/a2-iam-v2-1-only-integration/inspec.yml
```diff
@@ -0,0 +1,13 @@
+name: a2-iam-v2-1-only-integration
```
msorens: This file was copied almost verbatim from existing code.

inspec/a2-iam-v2-1-only-integration/controls/includes.rb
```diff
@@ -0,0 +1,5 @@
+# encoding: utf-8
```
msorens: This file was copied verbatim from existing code.

inspec/a2-iam-v2-1-only-integration/controls/role_projects.rb
tylercloke: Moving these to constants is perfect as long as we don't expect the tests to ever change them.

inspec/a2-iam-v2p1-only-integration/controls/role_projects.rb
```diff
@@ -0,0 +1,134 @@
...
+      # these are system default roles
+      resp = automate_api_request(
+        "/apis/iam/v2beta/roles",
+        request_headers: { 'projects': '(unassigned)' },
+        )
+      expect(resp.http_status).to eq 200
+      expect(resp.parsed_response_body[:roles]).not_to be_empty
+    end
+
+    it 'returns all roles when unfiltered' do
```
bcmdarroch: I would add a new control for deny behavior. say we have a policy like this:
```
deny policy:
         statement 1: deny all actions on proj1 roles
         statement 2: allow all actions on proj2, proj3 roles
```

what happens on these requests:
1. GET a role with projects `proj1`?
2. LIST roles with projects `proj1, proj2`?
3. LIST roles with no project headers?
4. DELETE a role with projects `proj1`?
5. UPDATE a role with projects `proj1`?
6. CREATE a role with projects `proj1`?

PR #103, opened 2019-04-17
[deployment] Don't run GC when stopping services

stevendanna:
Right now we use the converger to stop all services. Since the
converger is also where the package garbage collector runs, we were
deleting packages on shutdown, which seems unnecessary and just slows
down the restart.

### DIFF
--- a/components/automate-deployment/pkg/converge/stop_mitigation.go
+++ b/components/automate-deployment/pkg/converge/stop_mitigation.go
@@ -81,7 +81,7 @@ func (r *safeServiceShutdownRunner) RunIfRequired(t target.Target, pkg habpkg.Ve
 	if !required {
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


components/automate-deployment/pkg/converge/stop_mitigation.go
```diff
@@ -81,7 +81,7 @@ func (r *safeServiceShutdownRunner) RunIfRequired(t target.Target, pkg habpkg.Ve
 	if !required {
 		return nil
 	}
-	logrus.Info("Restart mitigation is required.")
+	logrus.Infof("Restart mitigation is required for service %s", pkg.Name())
```
stevendanna: These two logging changes were the result of me watching the log output of the shutdown sequence and being a bit confused about the ordering I saw. These two log changes helped clarify the situation.

PR #104, opened 2019-04-17
Remove fqdn from Summary doc struct since we don't have it in the data model

alexpop:
We don't have `fqdn` in the summary data model. By removing FQDN from the summary struct, we avoid adding a blank value in elasticsearch for every summary document.

I've also added the new fields to the tiny report we use for local dev & test.


alexpop:
Victoria and Pete, thank you for the reviews. Merged!

PR #106, opened 2019-04-17
Fix case where we could deadlock

jaym:
We've taken the lock here, but we return an error without giving it
back.


components/automate-deployment/pkg/server/backup.go
```diff
@@ -161,6 +161,7 @@ func (s *server) RestoreBackup(ctx context.Context, req *api.RestoreBackupReques
 	// file.
 	err = s.acquireLock(ctx)
 	if err != nil {
+		s.deployment.Unlock()
```
stevendanna: No thanks!

PR #110, opened 2019-04-17
Dan/upgrade golangci lint

danielsdeleo:

    Update golangci-lint to make it more usable in the hab studio
    
    golangci-lint 1.15 is affected by a bug that occurs when in a path with
    a symlink, which prevents it from properly applying `nolint` directives
    in comments. In the studio, we symlink the automate source from /src
    into the scaffolding gopath when you run `setup_go_workspace`; the
    combination of these factors makes golangci-lint not usable in the
    studio. 1.16 is supposed to contain a fix.
 
- [ ] Code actually executed?


stevendanna:
Thanks for digging into this.  We vendor golangci-lint to avoid constant problems downloading it in CI: https://github.com/chef/automate/tree/master/cache . We may want to update that vendored copy?

danielsdeleo:
@stevendanna done. I also brought back 1.15 for 64bit linux, since the best I could manage for compliance service was to pin. I put some of the failures I was seeing with compliance using 1.16 in the commit message for that.

PR #115, opened 2019-04-18
Enable additional static tests for apps svc and fix lint issues

danielsdeleo:

Changes applications-service to use `Makefile.common_go` stuff, which implies that we start using golangci-lint. That linter found a handful of issues, many of which are real bugs. Previously we were using the studio function `go_component_static_tests` which just runs `gofmt` so isn't very useful. We _could_ change `go_component_static_tests` to run golangci-lint, but the following projects use it:
*  config-mgmt-service
* ingest-service
* secrets-service

The other problem with the studio function is it doesn't have the same customization capabilities as the makefile approach. For example, with the makefile, it's possible to pin golangci-lint to a specific version if you need.

As far as I can tell, most devs are running the linter on their host and not the studio.  As of golangci-lint 1.15, it doesn't work very well in the studio due to a bug that breaks stuff when the CWD path has a symlink. That's fixed in 1.16, so I have a PR to upgrade us to that version: https://github.com/chef/automate/pull/110

### :+1: Definition of Done

- [x] Ci runs golangci-lint on applications service
- [x] Ci passes

* click the links to see the verify (public) pipeline run, find the applications service unit tests, and un-collapse the `Running /bin/bash -e -c 'scripts/install_golang` line. You'll see that it runs `make ci` which runs `../../cache/golangci-lint-1.16.0-linux-amd64/golangci-lint run -c ../../.golangci.toml ./...`
* In the studio: `go_component_make components/applications-service/ ci` or `go_component_make components/applications-service/ lint test build` (proper handling of multiple make target args)
* Run `make ci` from the `components/applications-service/` dir of your local clone of automate

- [ ] Code actually executed?

### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -69,9 +69,7 @@ steps:
 
   - label: "[unit] applications-service"
     command:
-      - scripts/install_golang
-      - cd components/applications-service
-      - make ci
+      - hab studio run "source .studiorc && go_component_make components/applications-service ci"
     timeout_in_minutes: 10
     retry:
       automatic:
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -69,7 +69,9 @@ steps:
 
   - label: "[unit] applications-service"
     command:
-      - hab studio run "source .studiorc && go_component_unit applications-service && go_component_static_tests applications-service"
+      - scripts/install_golang
+      - cd components/applications-service
+      - make ci
     timeout_in_minutes: 10
     retry:
       automatic:
--- a/components/applications-service/cmd/applications-publisher/main.go
+++ b/components/applications-service/cmd/applications-publisher/main.go
@@ -102,6 +102,10 @@ func main() {
 	flag.Usage = usage
 	flag.Parse()
 
+	if uniqID {
+		clientID = "applications-publisher-" + strconv.FormatInt(t.UnixNano(), 10)
+	}
+
 	if internalNats {
 		client = nats.New(
 			fmt.Sprintf("nats://0.0.0.0:%s", port),
@@ -127,10 +131,6 @@ func main() {
 	// Convert proto enums
 	event.HealthCheck = applications.HealthStatus(health)
 	event.Status = applications.ServiceStatus(status)
-
-	if uniqID {
-		clientID = "applications-publisher-" + strconv.FormatInt(t.UnixNano(), 10)
-	}
 
 	if infiniteLoop {
 		err := client.Connect()
--- a/components/applications-service/cmd/applications-service/commands/root.go
+++ b/components/applications-service/cmd/applications-service/commands/root.go
@@ -34,9 +34,20 @@ func init() {
 	RootCmd.PersistentFlags().String("key", "", "path to key")
 	RootCmd.PersistentFlags().String("cert", "", "path to cert")
 
-	viper.BindPFlag("cert", RootCmd.PersistentFlags().Lookup("cert"))
-	viper.BindPFlag("root-cert", RootCmd.PersistentFlags().Lookup("root-cert"))
-	viper.BindPFlag("key", RootCmd.PersistentFlags().Lookup("key"))
+	// https://github.com/spf13/viper/blob/7a605a50e69cd1ea431c4a1e6eebe9b287ef6de4/viper.go#L939
+	// err is non-nil when the flag doesn't exist. panic here should get caught quickly by Ci.
+	err := viper.BindPFlag("cert", RootCmd.PersistentFlags().Lookup("cert"))
+	if err != nil {
+		panic("viper configuration does not match command line flags")
+	}
+	err = viper.BindPFlag("root-cert", RootCmd.PersistentFlags().Lookup("root-cert"))
+	if err != nil {
+		panic("viper configuration does not match command line flags")
+	}
+	err = viper.BindPFlag("key", RootCmd.PersistentFlags().Lookup("key"))
+	if err != nil {
+		panic("viper configuration does not match command line flags")
+	}
 }
 
 // initConfig reads in config file and ENV variables if set.
--- a/components/applications-service/cmd/applications-service/commands/serve.go
+++ b/components/applications-service/cmd/applications-service/commands/serve.go
@@ -22,7 +22,7 @@ import (
 var serveCmd = &cobra.Command{
 	Use:   "serve",
 	Short: "Launches applications services",
-	Run: func(cmd *cobra.Command, args []string) {
+	RunE: func(cmd *cobra.Command, args []string) error {
 		conf, err := configFromViper()
 		if err != nil {
 			log.WithFields(log.Fields{
@@ -56,7 +56,10 @@ var serveCmd = &cobra.Command{
 			)
 
 			// Trigger NATS Subscription
-			natsClient.ConnectAndSubscribe()
+			err := natsClient.ConnectAndSubscribe()
+			if err != nil {
+				log.WithError(err).Fatal("could not connect to nats-streaming")
+			}
 
 			// Receive digested messages from the event channel and
 			// send them to the datastore for processing
@@ -96,7 +99,7 @@ var serveCmd = &cobra.Command{
 		}()
 
 		// GRPC Server
-		grpc.Spawn(conf, connFactory)
+		return grpc.Spawn(conf, connFactory)
 	},
 }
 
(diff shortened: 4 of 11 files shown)


components/applications-service/cmd/applications-publisher/main.go
```diff
@@ -102,6 +102,10 @@ func main() {
 	flag.Usage = usage
 	flag.Parse()
 
+	if uniqID {
```
danielsdeleo: We were assigning this variable too late for it to affect anything. We want to assign before it's used on line 112 (updated code)

components/applications-service/cmd/applications-service/commands/root.go
```diff
@@ -34,9 +34,20 @@ func init() {
 	RootCmd.PersistentFlags().String("key", "", "path to key")
 	RootCmd.PersistentFlags().String("cert", "", "path to cert")
 
-	viper.BindPFlag("cert", RootCmd.PersistentFlags().Lookup("cert"))
-	viper.BindPFlag("root-cert", RootCmd.PersistentFlags().Lookup("root-cert"))
-	viper.BindPFlag("key", RootCmd.PersistentFlags().Lookup("key"))
+	// https://github.com/spf13/viper/blob/7a605a50e69cd1ea431c4a1e6eebe9b287ef6de4/viper.go#L939
+	// err is non-nil when the flag doesn't exist. panic here should get caught quickly by Ci.
```
danielsdeleo: We could've done a `//nolint: errcheck` here, but I opted to make it panic as this should make it impossible for us to pass Ci if the flag names get out of sync.

components/applications-service/cmd/applications-service/commands/serve.go
```diff
@@ -56,7 +56,10 @@ var serveCmd = &cobra.Command{
 			)
 
 			// Trigger NATS Subscription
-			natsClient.ConnectAndSubscribe()
+			err := natsClient.ConnectAndSubscribe()
+			if err != nil {
+				log.WithError(err).Fatal("could not connect to nats-streaming")
```
danielsdeleo: This will make applications service crash if it can't connect to NATS, which means we can't silently fail to start ingesting messages any more.

components/applications-service/pkg/nats/nats.go
```diff
@@ -204,8 +204,8 @@ func (nc *NatsClient) ConnectAndPublish(msg *applications.HabService) error {
 // order to configure TLS, we are responsible for closing it.
 func (nc *NatsClient) Close() {
 	natsConn := nc.conn.NatsConn()
-	nc.conn.Close()
-	natsConn.Close()
+	nc.conn.Close()  // nolint: errcheck
+	natsConn.Close() // nolint: errcheck
```
danielsdeleo: This is usually used in a `defer` statement so best-effort is appropriate here. Plus the reason you probably fail is if the connection is closed already.

danielsdeleo:
> Another point is that, I am one of the developers (maybe also @lancewf ) that doesn't use/run anything outside the studio. So this will make us (not sure who else) install go and all the tools in our computers and deal with versioning and extra stuff (settings, etc)

I already fixed that and described how to run it in the studio in the demo script

.expeditor/verify.pipeline.yml
```diff
@@ -69,7 +69,9 @@ steps:
 
   - label: "[unit] applications-service"
     command:
-      - hab studio run "source .studiorc && go_component_unit applications-service && go_component_static_tests applications-service"
```
lancewf: I have the same question @afiune does. Are we moving away from using the hab studio for development or toward it? If we are moving toward using the studio more we should keep this test using the studio. 

From the poll that ssd did a few weeks ago, it looks like we most if not all people are using hab studio as their dev environment. 
Screen Shot 2019-04-19 at 4 35 46 PM

danielsdeleo: > Are we moving away from using the hab studio for development or toward it?

I don't understand this question in the context of this pull request. As I wrote in the description, I did a bunch of work already in https://github.com/chef/automate/pull/110 to ensure you can use this both inside and outside of the studio as you like. Does it have to have a wrapper function in .studio to count as the studio?

As for that poll, I'm not sure we're measuring the same things. I personally never run any automate services locally, but I very frequently run static analysis and unit tests on my Mac since it's faster; I would probably say I "use the studio" as "my dev environment." I am pretty sure that's how people working on the other projects using Makefile.common_go are working since it was broken in the studio until I fixed it yesterday. But in any case, it works in the studio now, so you don't have to change how you do development at all, except that the specific procedure for running static tests in the studio has changed for this one component.

afiune: @danielsdeleo I think the main problem is that you are switching the Buildkite test from running **_inside the studio_** to run **_outside with a Makefile_** that, as the poll says clearly, we are not running almost anything outside the studio. 

I think this PR would be perfectly fine if we switch the tests to run inside the studio as they were before. That is with `go_component_static_tests` and `go_componen_unit` methods.

stevendanna: FWIW I think that poll likely isn't too informative on this point since it only provided one option.  For instance, I use the studio but almost always run the linter outside of the studio.

Right now, every project that uses the linter does so using a Makefile.

I think Dan did a pretty good job of mentioning the various things we might want to fix to move everything to running the linter inside the studio using a helper function:

- Apply the linter to ingest-service
- Apply the linter to config-mgmt-service (https://github.com/chef/automate/pull/127)
- Apply the linter to secrets-service (https://github.com/chef/automate/pull/127)
- Provide some means for `go_component_static_tests` to continue to allow components to specify their own linter args or fix all of the issues that led to custom exceptions:

```
components/automate-cli/Makefile:LINTERARGS:=--exclude-use-default -D lll ./...
components/automate-deployment/Makefile:LINTERARGS=--exclude-use-default -D lll ./...
components/compliance-service/Makefile:LINTERARGS=-v -D lll -D unparam -D nakedret -D prealloc --deadline 1m ./...
components/es-sidecar-service/Makefile:LINTERARGS=-D lll ./... ../../api/config/es-sidecar-service/...
components/nodemanager-service/Makefile:LINTERARGS=-v -D lll -D unparam -D nakedret -D prealloc --deadline 1m ./...
```

- Provide some means to continue to run the lint tests outside of the studio since most Go tools are 6 to 7 times studio inside Docker for Mac.

Personally, I would like to see us converge on something where every project had a standard `Makefile` that worked inside the studio and (for some tasks) outside the studio easily.  But it is a long road from where we are to that.

I think y'all should do whatever you come to agreement on for the application-service, but perhaps we can open an issue to track the above?

afiune: I agree with @stevendanna — we should recognize all the amazing work that @danielsdeleo did, as usual, great work!

Having said that, I have never used a single make command since the creation of this repo, and even before that, since the repo was actually multiple repositories (before the monorepo) — I always run everything inside the studio and I would expect the CI system to do the same, that way we all have the same tools, compilers and versions. 

Now about the question that @lancewf asked, I am curious to know if this is a standard (to run the tests and linter outside the studio) and if it is, communicate it to everyone so that we all do the same. What I’m trying to avoid is maintaining multiple ways to develop and test. 

I hope this information clarified a bit more the situation.

lancewf: @danielsdeleo You are right. I don't see anything wrong with this PR. My question is more in general for Automate: which way are we moving towards? As @afiune said before, the old config-mgmt team always had everything run in the studio. Our reason for this was that it was easier to know that all the developers were using the same development environments (@afiune did a presentation on this a year back). This allowed us to assume if one person had a problem then we should see the same problem in our environment.  I know that not all the other teams have this same goal. Now we are the super team and a lot of the compliance tests and functions can only run outside of the studio. We are trying to move toward using the studio, but it is not a priority. 

It looks like you have answered my original question. We want to have a mix of development environments. Meaning people work outside of the studio and inside. Let me know if that is not correct.

danielsdeleo: @lancewf my preference would be that everything works inside the studio, and some stuff optionally also works outside. The docker Mac filesystem is pretty slow so I personally find it nice to run these quick kinds of checks outside of the studio. In any case I added a helper function to `cd` to the `GOPATH` and run `make` for you and made that the ci script for this service. This way we can support customization of the linter for each service/team's needs with the makefile and have a standardized way to run it which gets tested in Ci.

lancewf: @danielsdeleo Thank you for hearing me out. This is great to hear. I think this is a great addition. I am finishing looking over the code now.

components/applications-service/cmd/applications-service/commands/serve.go
afiune: @stevendanna this is the place where I think we are now bouncing constantly trying to connect to the NATS server, the server won't come up unless the user enables the feature flag so we will be in an infinite loop.

I can add a feature flag to avoid trying to connect unless the flag is enabled.

PR #118, opened 2019-04-18
Filter services by health status

afiune:
When users see a list of services in a group, they want to filter out the services by their health status (Critical, Warning, OK and Unknown), so that they can narrow down the target for further investigation.

We want to create a set of health status filters in the service group details panel to help users achieve the goal.

### :+1: Definition of Done

- In each filter, display one status label (Critical, Warning, OK or Unknown), status icon, **without** counts.
- A user should be able to click or tab through the filters to filter out the service cards by the selected health status.
- Can filter up to 1000 services. 25 items per page, the same as the run history panel.

This PR is modifying a number of components, so make sure to build them all before starting all the services:

- components/automate-gateway
- components/applications-service
- components/automate-ui (be careful, you have to actually build `automate-ui-devproxy`)
- components/chef-ui-library

Once you have built all of them inside your studio, you could just start all services:
```
start_all_services
```

This work looks like this in the UI:
services-filter-bar

https://chefio.atlassian.net/browse/A2-626

- [x] Code actually executed?


kmacgugan:
In my browser the health boxes end up being different sizes. Also when I click on them the URL doesn't show the filters being passed. What's expected behavior here? Also will these filters collide with the filters on the main applications tab component?

afiune:
@kmacgugan Thanks for your review!!! The boxes, I will fix them. :100:

The thing you mentioned about making everything persistent in the URL, we /could/ do that, but we really haven't done it for the clients or compliance, I wonder if we want to diverge and do it different or the same way? We should talk about it in our standup.

kmacgugan:
Looks good, the only thing is that the warning icon was black instead of orange.

afiune:
@kmacgugan the Warning color was yellow/orange but I changed it so that it is filled with color instead of empty (white). That was the request from @apriofrost - The page looks like this now:

image

**Same functionality as the gif in the description!**

components/chef-ui-library/src/molecules/chef-status-filter-group/chef-status-filter-group.scss
```diff
@@ -1,32 +1,76 @@
...
     &.filter {
-      font-size: 14px;
+      padding-bottom: 8px;
+      padding-right: 12px;
+      padding-left: 12px;
+      padding-top: 8px;
       height: auto;
-      display: flex;
+      width: auto;
+      margin-right: 16px;
```
afiune: @apriofrost here

components/chef-ui-library/src/molecules/chef-status-filter-group/chef-status-filter-group.scss
```diff
@@ -1,32 +1,76 @@
...
+      &.unknown {
+        chef-icon {
+          color: chef-hsl(var(--chef-dark-grey));
+        }
+      }
+    }
+
+    &.lean-filter {
+      padding: 8px;
+      margin-right: 8px;
```
afiune: @apriofrost and here

PR #123, opened 2019-04-19
[deployment] Do not display headers for zero-value config entries

yzl:

- [ ] Code actually executed?

### DIFF
--- a/api/config/deployment/init_config.go
+++ b/api/config/deployment/init_config.go
@@ -22,6 +22,7 @@ import (
 	"github.com/pkg/errors"
 
 	es "github.com/chef/automate/api/config/elasticsearch"
+	license_control "github.com/chef/automate/api/config/license_control"
 	global "github.com/chef/automate/api/config/shared"
 	w "github.com/chef/automate/api/config/shared/wrappers"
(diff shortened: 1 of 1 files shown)


api/config/deployment/init_config.go
```diff
@@ -158,38 +159,73 @@ func (c InitConfig) Render() (string, error) {
...
-	cfg.Deployment.V1.Svc.UpgradeStrategy = w.String(c.UpgradeStrategy)
-	cfg.Deployment.V1.Svc.DeploymentType = w.String(c.DeploymentType)
-	cfg.LicenseControl.V1.Svc.License = w.String(c.License)
-	cfg.Elasticsearch.V1.Sys.Runtime.Heapsize = w.String(c.ESHeapSize)
-	return cfg
+	// Fill out automate config
+	cfg := AutomateConfig{}
+	cfg.Global = g
+
+	cfg.SetChannel(c.Channel)                 // nolint: errcheck
```
yzl: I have a mental block on what action I should be taking with the errors I am currently ignoring. (Originally we were creating a new AutomateConfig and setting, instead of overriding, exactly the same values in the struct.)

PR #127, opened 2019-04-20
Apply golangci-lint to config-mgmt-service and secrets-service

stevendanna:
These are 2 of the 3 services that don't seem to be using the linter
currently. This applies the linter to those projects which will
hopefully make it a bit easier to transition to having
`go_component_static_tests` run the linter.

`ingest-service` is, I believe, likely the last service that needs to
have the linter applied to it. I have not done it here because there
are a large number of errcheck violations that need to be looked at
carefully.

### DIFF
--- a/components/config-mgmt-service/backend/elastic/nodes.go
+++ b/components/config-mgmt-service/backend/elastic/nodes.go
@@ -62,14 +62,10 @@ func (es Backend) GetNode(id string) (backend.Node, error) {
 // non-random access pagination and custom selection of how the docs are sorted.
 func (es Backend) GetInventoryNodes(ctx context.Context, start time.Time,
 	end time.Time, filters map[string][]string, cursorDate time.Time,
-	cursorID string, pageSize int, sortField string,
+	cursorID string, pageSize int, _ string,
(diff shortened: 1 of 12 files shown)


components/config-mgmt-service/backend/elastic/nodes.go
```diff
@@ -83,7 +83,7 @@ func (es Backend) GetInventoryNodes(ctx context.Context, start time.Time,
 		Query(mainQuery).
 		Index(IndexNodeState).
 		Size(pageSize).
-		Sort(CheckinTimestamp, ascending).
+		Sort(sortField, ascending).
```
stevendanna: ⚠️ This is potentially a behavior change since we were previously ignoring this variable.  The linter was complaining about code on lines 70-72 being an ineffectual assignment.  We could instead continue to ignore this argument and then remove those lines.

stevendanna: Whelp, a bunch of tests failed, so let me try ignoring this argument instead.

phiggins:
https://github.com/chef/a2/issues/4979

PR #128, opened 2019-04-20
Deployment History Log

jaym:
Adding the ability to log events in deployment service. Currently, just
prints them with logrus.

Example usage:
```golang
err := history.Record(context.Background(), EventInitializeDeploymentService, server.initialize)
```

or with tags:
```golang
err := history.WithContext(ctx).WithTags(
  history.Tag{"secret_group", secret.Group},
  history.Tag{"secret_name", secret.Name},
).Record(EventGenerateSecret, func(ctx context.Context) error {
```

You can also control it more manually:
```golang
ctx := history.RecordEvent(context.Background(), SomeEventClass)

err := DoSomething(ctx)
if err != nil {
  history.FailErr(ctx, err)
} else {
  history.Success(ctx)
}

```


components/automate-deployment/pkg/history/event.go
```diff
@@ -0,0 +1,34 @@
+package history
+
+import "time"
+
+type EventID string
```
stevendanna: Why a string and not a `uuid` type directly?

components/automate-deployment/pkg/history/logrus_interceptor.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func (logrusInterceptor) LogStart(ev *StartEvent) {
+	logrus.WithFields(logrus.Fields{
+		"id":          ev.ID,
+		"class":       ev.Name,
+		"description": ev.Description,
+		"cause":       ev.Cause,
+		"tags":        ev.Tags,
+	}).Info("Event Start")
```
stevendanna: [nit] In your example output, this ends up putting the information I care about pretty far into the log line.  I wonder if we want to make `ev.Description` the thing we pass to `Info`?  I see the value in driving this towards something more machine readable than our current logs but I worry it is gonna make debugging harder in the short term.

components/automate-deployment/pkg/history/history.go
```diff
@@ -0,0 +1,108 @@
...
+	Tags []Tag
+}
+
+func WithTags(tags ...Tag) *Logger {
+	return &Logger{
+		Tags: tags,
+	}
+}
+
+func LogEvent(ctx context.Context, name EventClass, description string) context.Context {
```
stevendanna: I wonder if `RecordEvent` might be better?

components/automate-deployment/pkg/history/history.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"github.com/gofrs/uuid"
+	"github.com/sirupsen/logrus"
+)
+
+var nowProvider = func() time.Time {
+	return time.Now()
+}
+
+type Logger struct {
```
stevendanna: I'm a little worried about this being confusing since we also have logrus which would be what most people expect from a thing called Logger.

components/automate-deployment/pkg/history/history.go
jaym: can we call `Do` -> `Record`?

components/automate-deployment/pkg/history/logrus_interceptor.go
jaym: we can.

components/automate-deployment/pkg/history/event.go
jaym: i hadn't decided i was going to use uuid when i wrote this line

jaym:
I think that warning needs to go. The first one will not have a value for the history ctx. Also, we should talk about what you were expecting to see before I get too attached to this code

components/automate-deployment/pkg/history/history.go
yzl: Recorder?

jaym:
we'll reopen this if we feel like fixing it up in the future

PR #132, opened 2019-04-22
Project update resume

lancewf:
### 📋 TODO
- [x] Fail on file io failures
- [x] Full Manual test
- [x] Fix test failures

The project update process can be a long running process. Therefore if the compliance-service, ingest-service, or authz-service is updated during this process it will need to resume where it left off.  To perform a resume the service needs to save its state during the process. Then during startup, the services need to check this saved state to resume the update process if needed. 

A restart of the elasticsearch service does not allow the project update process to resume. If this service restarts the compliance-service and ingest-service will detect that their jobs are not running any longer and send a failure message to the authz-service. The authz-service will mark the update process as a failure with a failure message. 

1. `build components/authz-service && build components/compliance-service && build components/ingest-service && build components/event-service`
1. `start_all_services`
1. `chef_load_nodes 30000`
1. `chef_load_compliance_scans -N10000 -D10 -T100000`
1. Watch the ingest-service status; `watch chef-automate dev grpcurl ingest-service -- chef.automate.domain.ingest.EventHandler.ProjectUpdateStatus`
1. Watch the compliance-service status; `watch chef-automate dev grpcurl compliance-service -- chef.automate.domain.compliance.ingest.ingest.ComplianceIngester.ProjectUpdateStatus`
1. Watch the authz-service status; `watch chef-automate dev grpcurl authz-service -- chef.automate.domain.authz.v2.Projects.ProjectUpdateStatus`
1. Create a project; `chef-automate dev grpcurl authz-service -- chef.automate.domain.authz.v2.Projects.CreateProject -d '{"name": "project9", "id": "project9"}'`
1. Update a project; `chef-automate dev grpcurl authz-service -- chef.automate.domain.authz.v2.Projects.UpdateProject -d '{"name": "project9", "id": "project9"}'`
1. Kill the ingest-service
1. Ensure the ingest-service resumes the update process where it left off. 
1. Update a project; `chef-automate dev grpcurl authz-service -- chef.automate.domain.authz.v2.Projects.UpdateProject -d '{"name": "project9", "id": "project9"}'`
1. Kill the compliance-service
1. Ensure the compliance-service resumes the update process where it left off.
1. Update a project; `chef-automate dev grpcurl authz-service -- chef.automate.domain.authz.v2.Projects.UpdateProject -d '{"name": "project9", "id": "project9"}'`
1. Kill the authz-service
1. Ensure the authz-service resumes the update process where it left off.
1. Update a project; `chef-automate dev grpcurl authz-service -- chef.automate.domain.authz.v2.Projects.UpdateProject -d '{"name": "project9", "id": "project9"}'`
1. Kill the automate-elasticsearch
1. Ensure the authz-service, compliance-service, and ingest-service jobs are stopped and the authz-service has failed status true and a failure message. 

Completes! 🎆  - https://github.com/chef/automate/issues/60

- [x] Code actually executed?


components/authz-service/config/config_mgmt.go
```diff
@@ -0,0 +1,103 @@
+package config
```
lancewf: Manage reading and writing a local config file.

components/authz-service/config/project_update_domain_service.go
```diff
@@ -0,0 +1,58 @@
+package config
```
lancewf: Pulled this code out into its own file.

components/authz-service/config/project_update_stage.go
```diff
@@ -0,0 +1,141 @@
...
+	"errors"
+	"fmt"
+	"time"
+
+	event_ids "github.com/chef/automate/lib/event"
+)
+
+// ProjectUpdateStage - the config for project updating
+// The stage is all the data that needs to be remembered if the service is restarted.
+type ProjectUpdateStage struct {
```
lancewf: The Stage contains all the changeable states of the project update process.

components/authz-service/habitat/config/config.yml
```diff
@@ -17,3 +17,5 @@ data-migrations-path: {{ pkg.svc_static_path }}/data-migrations
 event-address: {{event-service.sys.ip}}:{{event-service.cfg.port}}
   {{~/if}}
 {{~/eachAlive}}
+
+config: {{pkg.svc_data_path}}/.authz-service.toml
```
lancewf: Adding the config file to the services data directory. `/hab/svc/authz-service/data/.authz-service.toml`

components/authz-service/server/v2/project_update_manager.go
```diff
@@ -17,351 +17,424 @@ import (
...
+
+// Recovery
+// * Domain service was down during start
+// * Authz service restarts
+
 const (
-	RunningState    = "running"
-	NotRunningState = "not_running"
+	minutesWithoutCheckingInFailure = 5
+	sleepTimeBetweenStatusChecksSec = 5
```
lancewf: Only check the status of the domain services ever 5 seconds.

components/authz-service/server/v2/project_update_manager.go
```diff
@@ -17,351 +17,424 @@ import (
...
+// * Domain service sends failure message
+
+// Recovery
+// * Domain service was down during start
+// * Authz service restarts
+
 const (
-	RunningState    = "running"
-	NotRunningState = "not_running"
+	minutesWithoutCheckingInFailure = 5
```
lancewf: If the service has not checked in for 5 minutes the process is declared failed.

components/authz-service/server/v2/project_update_manager.go
```diff
@@ -17,351 +17,424 @@ import (
...
 
 // ProjectUpdateManager - project update manager
 type ProjectUpdateManager struct {
-	state              string
-	projectUpdateID    string
+	stage              config.ProjectUpdateStage
 	eventServiceClient automate_event.EventServiceClient
-	domainServices     []*domainService
+	configManager      *config.Manager
+	updateQueue        chan<- update
```
lancewf: This is a queue of updates to the ProjectUpdateStage. These change will only be processed in order by one thread.

lib/config/config_mgmt.go
```diff
@@ -0,0 +1,101 @@
+package config
```
lancewf: The base functionality for managing in-memory config data that is backed up to a file.

lib/config/config_mgmt.go
```diff
@@ -0,0 +1,101 @@
...
+			continue
+		}
+
+		manager.Config = c
+
+		update.errc <- manager.saveToFile(c)
+	}
+}
+
+func (manager *Manager) saveToFile(config interface{}) error {
```
lancewf: Only one thread should call this function.

components/ingest-service/projectupdater/manager.go
```diff
@@ -25,88 +26,126 @@ const (
 	maxNumberOfConsecutiveFails          = 10
 )
 
+type stage struct {
+	state           string
+	projectUpdateID string
+	esJobID         string
+}
```
lancewf: The stage is all the data to determine the state of the service.

components/ingest-service/projectupdater/manager.go
```diff
@@ -25,88 +26,126 @@ const (
...
-	esJobID               string
+	// stage should only be updated by using the 'updateStage' function
+	stage                 stage
 	percentageComplete    float32
 	estimatedEndTimeInSec int64
 	client                backend.Client
 	authzProjectsClient   iam_v2.ProjectsClient
 	eventServiceClient    automate_event.EventServiceClient
+	configManager         *config.Manager
+	updateQueue           chan<- func(stage) stage
```
lancewf: Allows only one thread to update the 'stage' data.

components/ingest-service/projectupdater/manager.go
```diff
@@ -25,88 +26,126 @@ const (
...
 		authzProjectsClient: authzProjectsClient,
 		eventServiceClient:  eventServiceClient,
+		configManager:       configManager,
+		updateQueue:         updateQueue,
 	}
+
+	// Single goroutine that updates the stage data
+	go manager.stageUpdater(updateQueue)
+
+	manager.resumePreviousState()
```
lancewf: After the `stage` data is read from the config file restore any running processes.

components/ingest-service/projectupdater/manager.go
```diff
@@ -25,88 +26,126 @@ const (
...
-		}
-		manager.esJobID = esJobID // Store the job ID and event ID
-		manager.projectUpdateID = projectUpdateID
-		manager.state = runningState
-		go manager.waitingForJobToComplete()
-	case runningState:
-		if manager.projectUpdateID == projectUpdateID {
-			//	Do nothing. The job has ready started
-		} else {
+	manager.updateStage(func(stage stage) stage {
```
lancewf: This adds this function to the `updateQueue` to be processed in order by a single thread.

components/ingest-service/projectupdater/manager.go
```diff
@@ -161,52 +200,101 @@ func (manager *Manager) waitingForJobToComplete() {
 		numberOfConsecutiveFails = 0
 	)
 
+	logrus.Info("waitingForJobToComplete")
```
lancewf: Removing this.

components/ingest-service/config/config_mgmt.go
```diff
@@ -89,6 +89,13 @@ type JobSchedulerConfig struct {
 	Running bool `toml:"running"`
 }
 
+// ProjectUpdateConfig - the config for project updating
+type ProjectUpdateConfig struct {
```
lancewf: Adding a new config section to store.

components/ingest-service/config/config_mgmt.go
```diff
@@ -226,85 +215,37 @@ func (manager *Manager) GetMissingNodesForDeletionSchedulerConfig() JobConfig {
...
 
-func (manager *Manager) send(updateFunc func(aggregateConfig) aggregateConfig) {
-	manager.updateQueue <- updateFunc
+func (manager *Manager) updateConfig(updateFunc func(aggregateConfig) (aggregateConfig, error)) error {
+	return manager.baseConfigManager.UpdateConfig(func(config interface{}) (interface{}, error) {
+		return updateFunc(config.(aggregateConfig))
+	})
 }
 
-func saveToFile(config aggregateConfig) error {
```
lancewf: Moved the core logic to lib/config/config_mgmt.go for authz-service and compliance-service use.

components/event-service/server/server.go
```diff
@@ -64,6 +64,7 @@ func (s *Server) initRegistry() map[string][]string {
 	registry[NodeTerminated] = []string{config.CFG_KEY}
 	registry[ProjectRulesUpdate] = []string{config.CFG_KEY, config.COMPLIANCE_INGEST_KEY}
 	registry[ProjectRulesUpdateStatus] = []string{config.AUTHZ}
+	registry[ProjectRulesUpdateFailed] = []string{config.AUTHZ}
```
lancewf: Authz listens for Failure events.

components/compliance-service/ingest/projectupdater/manager.go
```diff
@@ -12,6 +12,7 @@ import (
 
 	iam_v2 "github.com/chef/automate/api/interservice/authz/v2"
 	automate_event "github.com/chef/automate/api/interservice/event"
+	"github.com/chef/automate/components/compliance-service/config"
```
lancewf: These are the same changes as the ingest-service's project update manager. 
* The `stage` is only updated by a single thread. 
* The `stage` is back up to a file through the `config.ConfigManager`

components/compliance-service/habitat/hooks/run
```diff
@@ -124,6 +124,9 @@ CONFIG="$CONFIG --event-endpoint {{event.sys.ip}}:{{event.cfg.port}}"
  {{~/if}}
  {{~/eachAlive}}
 
+ # Create the compliance config file
+CONFIG="$CONFIG --config {{pkg.svc_data_path}}/.compliance-service.toml"
```
lancewf: Adding a config file for backing up the state of the project update. This file will be located at `/hab/svc/compliance-service/data/.compliance-service.toml`

components/compliance-service/config/config_mgmt.go
```diff
@@ -0,0 +1,90 @@
+package config
```
lancewf: Manages reading and writing a local config file.

components/authz-service/config/config_mgmt.go
```diff
@@ -0,0 +1,103 @@
...
+	})
+}
+
+func (manager *Manager) updateConfig(updateFunc func(aggregateConfig) (aggregateConfig, error)) error {
+	return manager.baseConfigManager.UpdateConfig(func(config interface{}) (interface{}, error) {
+		return updateFunc(config.(aggregateConfig))
+	})
+}
+
+func readinConfig(configFile string, defaultConfig aggregateConfig) interface{} {
```
phiggins: Nitpick: "readin" isn't a word. is `readConfig` ok? `readConfigFromFile` would be a better, more verbose name.

components/authz-service/config/config_mgmt_test.go
```diff
@@ -0,0 +1,155 @@
...
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/chef/automate/components/authz-service/config"
+	event_ids "github.com/chef/automate/lib/event"
+	"github.com/stretchr/testify/assert"
+)
+
+const cFile = "/tmp/.authz-service.toml"
```
phiggins: If the name of this file doesn't matter to the app code, something that mentioned this was a test file would be nice so it doesn't get mistaken for something important and left around.

components/authz-service/config/project_update_stage.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+	}
+	return true
+}
+
+func (stage ProjectUpdateStage) HasFailedDomainService() bool {
+	failure := false
+	for _, domainService := range stage.DomainServices {
+		if domainService.Failed {
+			failure = true
```
phiggins: This could be `return true`, right?

components/compliance-service/integration_test/reports_projects_update_test.go
```diff
@@ -1397,30 +1397,32 @@ func TestStartProjectUpdateWhenIDIsSent(t *testing.T) {
...
-		if lastEventSent.Type.Name != automate_event_type.ProjectRulesUpdateStatus {
-			continue
-		}
 
-		if lastEventSent.Data.Fields["Completed"].GetBoolValue() {
-			break
+		for _, event := range eventsSent {
+			assert.Equal(t, event.Type.Name, automate_event_type.ProjectRulesUpdateStatus)
+			if event.Data.Fields["Completed"].GetBoolValue() {
+				complete = true
```
phiggins: Instead of using a boolean, could this be `break`?

components/compliance-service/config/config_mgmt.go
```diff
@@ -0,0 +1,90 @@
...
+	baseConfigManager *base_config.Manager
+}
+
+// Config - stores the configuration for the service
+type aggregateConfig struct {
+	ProjectUpdateConfig ProjectUpdateConfig `toml:"project_update_config"`
+}
+
+// NewConfigManager - create a new config. There should only be one config for the service.
+func NewConfigManager(configFile string) *ConfigManager {
```
phiggins: The other `ConfigManager` packages had a `NewManager` function for this purpose. Any reason this one needs to be different?

phiggins: There are a lot of calls in this PR to `NewManager`/`NewConfigManager` that are passed an empty string, which returns the default config, I'm guessing. Instead I'd suggest making that an explicit operation like `NewManagerWithDefaults()`/`NewManager(filename)` or `NewManager()`/`NewManagerFromFile(filename)`, etc.

components/authz-service/cmd/authz-service/main.go
```diff
@@ -115,7 +116,7 @@ Please pass a config file as the only argument to this command.`))
 
 	// if server.GRPC() returns, it's with an error
 	fail(server.GRPC(ctx, cfg.GRPC, l, connFactory, engine, migrationConfig,
-		dataMigrationConfig, cfg.EventAddress))
+		dataMigrationConfig, cfg.EventAddress, cfg.ConfigFile))
```
blakestier: Is it worth being more specific about what kind of `ConfigFile` this is? The many config things in here is a bit confusing

components/authz-service/config/config_mgmt.go
```diff
@@ -0,0 +1,103 @@
...
+
+// Close - to close out the channel for this object.
+// This should only be called when the service is being shutdown
+func (manager *Manager) Close() {
+	manager.baseConfigManager.Close()
+}
+
+// GetProjectUpdateStage - get the project update stage
+func (manager *Manager) GetProjectUpdateStage() ProjectUpdateStage {
+	return manager.baseConfigManager.Config.(aggregateConfig).ProjectUpdateStage
```
blakestier: I'm not totally familiar with this syntax-- is this a type check? Could it fail? At a quick glance it seems like a lot of calls chained

components/authz-service/config/project_update_stage.go
blakestier: That would allow us to eliminate the var `failure`

components/authz-service/server/v2/project_update_manager_test.go
```diff
@@ -125,28 +151,35 @@ func TestProjectUpdateManagerSendCancelEvent(t *testing.T) {
...
+	waitForWithTimeout(t, func() bool {
+		return config.RunningState == manager.State()
+	}, time.Second*3, "State did not switch to Running")
 
-	eventData := lastestPublishedEvent.Data
+	waitFor(func() bool {
+		return len(eventsSent) > 0
+	})
 
+	eventData := eventsSent[0].Data
```
blakestier: This is just a test, so maybe it doesn't matter, but could `eventsSent` ever be `nil` or empty?

components/compliance-service/ingest/server/compliance.go
```diff
@@ -37,18 +38,18 @@ type ComplianceIngestServer struct {
 	mgrClient          manager.NodeManagerServiceClient
 	automateURL        string
 	notifierClient     notifier.Notifier
-	updateManager      projectupdater.Manager
+	updateManager      *projectupdater.Manager
```
blakestier: [huge nit] totally out of scope of this PR, but I was expecting `projectUpdater` (and project-updater for the file name)

components/ingest-service/config/config_mgmt.go
```diff
@@ -114,84 +121,66 @@ func defaultConfig() aggregateConfig {
 				Running:   true,
 			},
 		},
+		ProjectUpdateConfig: ProjectUpdateConfig{
+			State: "not_running",
```
blakestier: I think I saw this elsewhere. Worth pulling into a constant? 🤔

components/authz-service/cmd/authz-service/main.go
lancewf: I was thinking this would be a general config file for the service. For example, the ingest-service is using the config file for storing job scheduler information and for the project updater state. The config manager that uses this file is threadsafe to allow all the different parts of the service to use the same one.

components/authz-service/config/config_mgmt.go
```diff
@@ -0,0 +1,103 @@
+package config
+
+import (
+	"io/ioutil"
+	"os"
+	"time"
+
+	"github.com/chef/automate/components/automate-deployment/pkg/toml"
```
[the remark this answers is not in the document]
lancewf: Nice catch! It was supposed to be `toml "github.com/pelletier/go-toml"`

components/authz-service/config/config_mgmt.go
```diff
@@ -0,0 +1,103 @@
...
+func (manager *Manager) updateConfig(updateFunc func(aggregateConfig) (aggregateConfig, error)) error {
+	return manager.baseConfigManager.UpdateConfig(func(config interface{}) (interface{}, error) {
+		return updateFunc(config.(aggregateConfig))
+	})
+}
+
+func readinConfig(configFile string, defaultConfig aggregateConfig) interface{} {
+	config := defaultConfig
+
+	if _, err := os.Stat(configFile); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
lancewf: Thanks that is a nice tip. 
```
	tomlData, err := ioutil.ReadFile(configFile)
	if os.IsNotExist(err) {
		// config file does not exists
		return config
	} else if err != nil {
		log.WithFields(log.Fields{
			"config_file": configFile,
		}).WithError(err).Error("Unable to read config file")
	}
```

components/authz-service/config/config_mgmt.go
```diff
@@ -0,0 +1,103 @@
...
+		// config file does not exists
+		return config
+	}
+
+	tomlData, err := ioutil.ReadFile(configFile)
+	if err != nil {
+		log.WithFields(log.Fields{
+			"config_file": configFile,
+		}).WithError(err).Error("Unable to read config file")
```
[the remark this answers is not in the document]
lancewf: Yes, we should be returning here and at 99.

components/authz-service/server/v2/project_update_manager_test.go
lancewf: I don't think it could be `nil` because the 3 lines before are waiting until the `eventsSent` has at least one value in it.

components/compliance-service/config/config_mgmt.go
lancewf: I named this `NewConfigManager` because the struct created is `ConfigManager`. This is not `Manager` because it was already taken in the types.go file. 

That makes sense to have the NewConfigManagerWithDefaults() function. I will add it.

lancewf: After looking at creating a `NewConfigManagerWithDefaults` function I don't think that would be a good idea, because a config file path needs to be passed in. All the uses of `NewConfigManager("")` are for tests, because we do not care if the file is saved.

components/compliance-service/ingest/server/compliance.go
lancewf: Yeah, I agree with you, I don't like the mashing of 'project' and 'updater'. I remember this conversation a while ago when we merged all the projects to form the mono repo and we decided to follow the golang way from here https://blog.golang.org/package-names.

components/compliance-service/integration_test/reports_projects_update_test.go
lancewf: I tried that before. A `break` here will exit from the `for _, event := range eventsSent` for on line 1405. Here we want to break out of the `for` outside of that. 

But after looking at it again, I could just do a return 😄 .

components/compliance-service/config/config_mgmt.go
phiggins: My biggest hesitance here is that due to an error or misconfiguration, we expect a string but don't get one and instead of an error we get some confusing default behavior.

components/ingest-service/config/config_mgmt.go
lancewf: Thank you for pointing that out. I think I was being lazy. I updated them all in this commit. 
https://github.com/chef/automate/pull/132/commits/055ec162a0155d9212f6a737d328b13674e849f4

lib/config/config_mgmt.go
lancewf: I would love a different word for this because 'config' is already overloaded. I would like to do the new name after this PR merges.

lib/config/config_mgmt.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		log.WithFields(log.Fields{
+			"error": err,
+		}).Error("Error Marshaling Config struct")
+		return err
+	}
+
+	// create a file with the permissions of the running user can read/write other can only read.
+	var permissions os.FileMode = 0644
+	err = ioutil.WriteFile(manager.configFile, tomlData, permissions)
```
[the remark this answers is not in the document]
lancewf: 🤔 I have not heard of the `AtomicWriter`, but by the name alone it sounds correct. I will look into it.

components/authz-service/config/config_mgmt_test.go
lancewf: I renamed the file like this `/tmp/.authz-service-test-delete-me.toml`

lib/config/config_mgmt.go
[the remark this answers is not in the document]
stevendanna: It lives here: https://github.com/chef/automate/blob/master/lib/io/fileutils/atomic_write.go

I think you can probably do something like:

```golang
err := fileutils.AtomicWrite(manager.configFile, 
                             bytes.NewReader(tomlData), 
                             fileutils.WithAtomicWriteFileMode(0644))
```

lancewf:
@stevendanna It was intentional to not fail if the state was not persisted. My thinking was that if for some reason we could not write to the file or later Elasticsearch or Postgresql we would want to the update process to still run. I did not want the resume feature to prevent the update process from running. This was just my judgment call and I would like other opinions.

components/compliance-service/config/config_mgmt.go
lancewf: I agree with you now after talking about this in standup today. I made the changes in this commit https://github.com/chef/automate/pull/132/commits/0d8492f2ed178433a4031d3a631ce742c46f7dc0. Now if the file is not writable/readable an error is returned.

lancewf:
@stevendanna After thinking about it more I do think you are right we should fail when the state is not persisted. I have corrected this in this commit https://github.com/chef/automate/pull/132/commits/e2a82e80c8a6013528cd8b14271c2e400d257018.

PR #134, opened 2019-04-22
Added dynamic label to project multi-select form input

tylercloke:

* If no projects selected, it reads "(unassigned)"
Screen Shot 2019-04-22 at 4 29 18 PM
* If one project selected, it shows the name of the project
Screen Shot 2019-04-22 at 4 29 25 PM
* If multiple projects selected, it says "n projects"
Screen Shot 2019-04-22 at 4 29 32 PM

* Build UI and navigate to https://a2-dev.test/component_library/projects-dropdown

Try selecting the various permutations of projects.

- [ ] Code actually executed?

### DIFF
--- a/components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.html
+++ b/components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.html
@@ -9,9 +9,10 @@
     [disabled]="disabled"
     (keydown.arrowup)="moveFocus($event)"
     (keydown.arrowdown)="moveFocus($event)">
-    Select projects... <chef-icon aria-hidden>keyboard_arrow_down</chef-icon>
+    <div id="projects-selected">{{ (label.length &gt; 25)? (label | slice:0:25)+'...' : label }}</div>
+    <chef-icon aria-hidden>keyboard_arrow_down</chef-icon>
   </button>
-  <chef-click-outside omit="column-dropdown-toggle" (clickOutside)="closeColumnDropdown()">
+  <chef-click-outside (clickOutside)="closeDropdown()">
     <chef-dropdown [attr.visible]="active">
       <chef-checkbox
         *ngFor="let project of projects; index as i"
--- a/components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
+++ b/components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
@@ -2,7 +2,11 @@ import {
   Component, EventEmitter, OnInit, Input, Output
 } from '@angular/core';
 
-import { find, orderBy } from 'lodash/fp';
+import { orderBy } from 'lodash/fp';
+
+import { Project } from 'app/entities/projects/project.model';
+
+const UNASSIGNED_LABEL = '(unassigned)';
 
 @Component({
   selector: 'app-projects-dropdown',
@@ -11,12 +15,9 @@ import { find, orderBy } from 'lodash/fp';
 })
 export class ProjectsDropdownComponent implements OnInit {
 
-  active = false;
-  showError = false;
-
   // The array of projects that you pass into the component. This array gets updated when a user
   // makes changes and is emitted to the parent component.
-  @Input() projects = [];
+  @Input() projects:  Array<Project> = [];
 
   // Setting required to true means the dropdown will show an error if a user tries to close the
   // dropdown but hasn't selected any projects.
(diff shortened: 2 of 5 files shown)


components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
```diff
@@ -56,20 +64,44 @@ export class ProjectsDropdownComponent implements OnInit {
...
+    this.updateProjectsSelected(this.projects[index], checked);
   }
 
   checkboxKeyPress(event, checked, index) {
-    this.projects[index].enabled = !checked;
+    this.updateProjectsSelected(this.projects[index], !checked);
     event.target.setAttribute('checked', !checked);
   }
 
+  updateProjectsSelected(project: Project, enabled: boolean): void {
```
bcmdarroch: `updateSelectedProjects` reads a bit better. i might even simplify to `updateLabel`

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.html
```diff
@@ -9,7 +9,8 @@
     [disabled]="disabled"
     (keydown.arrowup)="moveFocus($event)"
     (keydown.arrowdown)="moveFocus($event)">
-    Select projects... <chef-icon aria-hidden>keyboard_arrow_down</chef-icon>
+    <div id="projects-selected">{{ label }}</div>
+    <chef-icon aria-hidden>keyboard_arrow_down</chef-icon>
   </button>
   <chef-click-outside omit="column-dropdown-toggle" (clickOutside)="closeColumnDropdown()">
```
msorens: Cut-and-paste residual: 
1. Suggest rename to just `closeDropdown` -- there is no "column" lurking around here.
2. Remove the omit property entirely; it is referencing a class that does not exist.
```suggestion
  <chef-click-outside (clickOutside)="closeDropdown()">
```

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
```diff
@@ -26,15 +27,22 @@ export class ProjectsDropdownComponent implements OnInit {
   @Input() disabled = false;
 
   // Emits the array of projects to the parent component
-  @Output() onSelection = new EventEmitter<any>();
+  @Output() onSelection = new EventEmitter<Array<Project>>();
+
+  active = false;
+  showError = false;
```
msorens: I do not think `showError` and its associated template code are necessary. (And I was not able to trigger it, anyway.) But the point of it, per the associated error message, is that it requires an item to be selected, and that is not true.

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.html
```diff
@@ -9,7 +9,8 @@
     [disabled]="disabled"
     (keydown.arrowup)="moveFocus($event)"
     (keydown.arrowdown)="moveFocus($event)">
-    Select projects... <chef-icon aria-hidden>keyboard_arrow_down</chef-icon>
+    <div id="projects-selected">{{ label }}</div>
```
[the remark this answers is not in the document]
susanev: Yah, that was the original intention along with a native tooltip on hover to reveal the entire name. @tylercloke would that be something that's quick for you to add now? if not, we can fix this up when @magwalk finishes up the a11y parts of it.

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
susanev: this component is going to be used in the future for required situations, in which we will need the error.

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.html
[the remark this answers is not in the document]
tylercloke: I'll commit what micheal has.

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
msorens: Makes sense. I withdraw my comment then. 😁

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.html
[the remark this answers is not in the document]
susanev: @tylercloke oh, then i think the user won't be able to access the full project name; i think if we are going to truncate we need to provide the full name in the tool tip, they kinda have to go together.

PR #136, opened 2019-04-23
[A2-626] Implement HealthCounts for services sidebar

afiune:
This is the continuation of a previous PR https://github.com/chef/automate/pull/118 where I left the __Health Counts out of scope__ to reduce the complexity and number of lines to review, this PR is finishing the A2-626 epic by adding that missing part.
>When users see a list of services in a group, they want to filter out the services by their health status (Critical, Warning, OK and Unknown), so that they can narrow down the target for further investigation.
>We want to create a set of health status filters in the service group details panel to help users achieve the goal.

This PR has backend and frontend code! 🥇 

tenor-197544338

### :+1: Definition of Done

- In each filter, display one status label (Critical, Warning, OK or Unknown), status icon, **WITH Health Counts.**
- A user should be able to click or tab through the filters to filter out the service cards by the selected health status.
- Can filter up to 1000 services. 25 items per page, the same as the run history panel.

This PR is modifying a number of components, so make sure to build them all before starting all the services:

- components/automate-gateway
- components/applications-service
- components/automate-ui (be careful, you have to actually build `automate-ui-devproxy`)
- components/chef-ui-library

Once you have built all of them inside your studio, you could just start all services:
```
start_all_services
```

This work looks like this in the UI:
image

https://chefio.atlassian.net/browse/A2-626

- [x] Code actually executed?


components/applications-service/integration_test/services_by_sg_test.go
```diff
@@ -92,12 +92,55 @@ func TestGetServicesBySGSingleService(t *testing.T) {
...
 
 		response, err := suite.ApplicationsServer.GetServicesBySG(ctx, request)
 		assert.Nil(t, err)
 		assert.Equal(t, expected.GetGroup(), response.GetGroup())
+		assert.Equal(t, expected.GetServicesHealthCounts(), response.GetServicesHealthCounts())
+		assertServicesEqual(t, expected.GetServices(), response.GetServices())
+	}
+}
+
+func TestGetServicesBySGWithPaginationParameter(t *testing.T) {
```
afiune: @kmacgugan here is a pagination test with the new additions. 💯

PR #152, opened 2019-04-23
Load Test Fixes

danielsdeleo:

This patch contains a handful of small-ish items I fixed/changed in the process of load testing the ingestion system:
* Set lower limits on NATS streaming message storage. We've had operational problems with unlimited queues in the past; NATS is not unlimited by default, but the limits are too large for our current performance characteristics. I opted to not do anything too fancy but left a lengthy code comment explaining.
* first step changing ingestion code to a pipeline. We use one goroutine to read messages from NATS and pass them via a go channel to a set of worker goroutines. Since go uses m:n threading we can just spawn a fairly large fixed number of workers.
* Add some rollup stats when describing a load generator configuration so you have a better idea what the generated load will be.
* Make the load generator initiate connections to the event gateway in a more sequential fashion; it takes a few minutes to setup thousands of connections.
* randomized exponential backoff when connecting to event gateway fails; this is faster for a single retry and slower but more likely to eventually succeed when lots of retries are needed.

### :+1: Definition of Done

This is a bit open-ended but the PR addresses all the issues I found during the performance test.

See the documentation of the results and process of the test for more about how to run your own load test.

- [ ] Code actually executed?

### DIFF
--- a/components/applications-service/cmd/applications-load-gen/commands/describe.go
+++ b/components/applications-service/cmd/applications-load-gen/commands/describe.go
@@ -41,6 +41,8 @@ func runDescribeCmd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	fmt.Print(supGroups.RollupStats())
+
 	for _, supGroup := range supGroups {
 		fmt.Print(supGroup.PrettyStr())
 	}
--- a/components/applications-service/pkg/generator/runner.go
+++ b/components/applications-service/pkg/generator/runner.go
@@ -122,15 +122,25 @@ func (s *SupervisorGroup) PrettyStr() string {
 func (s *SupervisorGroup) Run(cfg *RunnerConfig, stats *RunnerStatsKeeper) {
 	for n := int32(0); n < s.Count; n++ {
 		log.WithFields(log.Fields{"name": s.Name, "index": n}).Info("spawning supervisor")
-		go s.SpawnSup(n, cfg, stats) // nolint: errcheck
+		sup, err := s.NewSup(n, cfg, stats) // nolint: errcheck
+		if err != nil {
+			log.WithError(err).Error("failed to spawn supervisor")
+			continue
+		}
+		err = sup.Connect()
+		if err != nil {
+			log.WithError(err).Error("failed to spawn supervisor")
+			continue
+		}
+		go sup.Run()
 	}
 }
 
-func (s *SupervisorGroup) SpawnSup(n int32, cfg *RunnerConfig, stats *RunnerStatsKeeper) error {
+func (s *SupervisorGroup) NewSup(n int32, cfg *RunnerConfig, stats *RunnerStatsKeeper) (*SupSim, error) {
 	uuid, err := uuid.NewV4()
 	if err != nil {
 		log.WithError(err).Error("Failed to make a V4 UUID")
-		return err
+		return nil, err
 	}
 
 	sup := &SupSim{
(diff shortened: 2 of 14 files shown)


components/applications-service/cmd/applications-load-gen/commands/describe.go
```diff
@@ -41,6 +41,8 @@ func runDescribeCmd(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	fmt.Print(supGroups.RollupStats())
```
danielsdeleo: This prints something like this at the top of the description. I found it helpful as a sanity check of my load testing configurations.

```
Totals
======
Supervisor groups:  2
Total Supervisors:  6000
Total Services:     18000
HealthCheck rate/s: 600.00
```

components/applications-service/pkg/generator/runner.go
```diff
@@ -100,14 +150,15 @@ func (s *SupervisorGroup) SpawnSup(n int32, cfg *RunnerConfig, stats *RunnerStat
...
 		Stats:             stats,
 	}
+	return sup, nil
 
-	err = sup.Run()
-	if err != nil {
-		fmt.Printf("Supervisor failed to run: %s\n", err)
-		return err
-	}
+	// err = sup.Run()
```
danielsdeleo: Oops, I'll delete this

components/applications-service/pkg/nats/nats.go
```diff
@@ -50,7 +51,7 @@ func NewExternalClient(url, cluster, client, durable, subject string) *NatsClien
 		durableID:         durable,
 		subject:           subject,
 		HabServiceEventCh: make(chan *applications.HabService), // buffered channel?
-		retries:           10,
+		retries:           5,
```
danielsdeleo: With the randomized exponential backoff retry algorithm, 10 retries is A LOT.

components/applications-service/pkg/nats/nats.go
```diff
@@ -130,7 +131,10 @@ func (nc *NatsClient) Connect() error {
 			"max_retries": nc.retries,
 		}).Error("Unable to connect to server")
 
-		time.Sleep(5 * time.Second)
+		baseSleep := int64(uint32(1) << tries)
+		randomization := rand.Int63n(baseSleep)
+		totalSleep := baseSleep + randomization
```
danielsdeleo: Randomized exponential backoff. This breaks up simultaneous requests over time as it keeps retrying.

components/applications-service/pkg/server/ingest.go
```diff
@@ -0,0 +1,86 @@
...
+	return nil
+}
+
+func (i *Ingester) Run() {
+	for n := 0; n <= numWorkers; n++ {
+		go i.runWorkerLoop()
+	}
+	// Receive digested messages from the event channel and
+	// send them to the datastore for processing
+	// TODO @afiune Move this logic into an ingestion pipeline
```
danielsdeleo: I guess this is TODONE? It's not much of a pipeline but gives us something to iterate on.

components/applications-service/pkg/storage/postgres/db.go
```diff
@@ -11,19 +11,19 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-// postgres is a wrapping struct that will hold the database mapping object
+// Postgres is a wrapping struct that will hold the database mapping object
```
danielsdeleo: Had to make this public so I could embed it in the `Ingester`. But also it's kind of a pain to return private types from public functions. That said, we could make the members of the struct private if desired.

components/automate-prometheus/habitat/config/prometheus.yml
```diff
@@ -2,7 +2,7 @@
 #
 # my global config
 global:
-  scrape_interval:     1m # Set the scrape interval to every 15 seconds. Default is every 1 minute.
+  scrape_interval:     15s # Set the scrape interval to every 15 seconds. Default is every 1 minute.
```
danielsdeleo: prometheus causes very little load as far as I can tell. It's nicer to have the data refresh faster when you start testing something.

PR #155, opened 2019-04-23
Improve sending raw messages to NATS Server(s)

afiune:
We are improving the way we send raw messages to our Internal and
External NATS Servers.

You can now start only the applications service and send a raw message
through the internal NATS server:
```
[2][default:/src:0]# applications_publish_raw_message
INFO[0000] Connecting to NATS Server                     client_id=applications-publisher cluster_id=event-service mtls_enabled=true nats_url="nats://0.0.0.0:10140"
INFO[0000] Publishing message                            message=dummy-ping-message type=bytes
[3][default:/src:0]# sl
--> Tailing the Habitat Supervisor's output (use 'Ctrl+c' to stop)
[2019-04-23T19:36:43+00:00] applications-service.default(O): time="2019-04-23T19:36:43Z" level=error msg="Unknown message, dropping" data=dummy-ping-message subject=habitat
```

Or do it through the External NATS Server. (You will have to start all
services first since we need auth)

```
[1][default:/src:0]# start_all_services
[2][default:/src:0]# applications_publish_raw_message_via_gw "Hello world"
[3][default:/src:0]# sl
--> Tailing the Habitat Supervisor's output (use 'Ctrl+c' to stop)
[2019-04-23T19:36:43+00:00] applications-service.default(O): time="2019-04-23T19:36:43Z" level=error msg="Unknown message, dropping" data="Hello world" subject=habitat
```


.studio/applications-service
```diff
@@ -77,40 +77,39 @@ function applications_integration() {
...
+}
+
+document "applications_publish_raw_message_via_gw" <<DOC
+  Publish a raw message to the External NATS Server (event-gateway)
+
+  Example: Publish a hello world message
+  --------------------------------------
+  # applications_publish_raw_message_via_gw "Hello World!"
+DOC
+function applications_publish_raw_message_via_gw() {
```
danielsdeleo: Hmm, so in https://github.com/chef/automate/pull/168 I'm providing a config option that can disable TLS since TLS+NATS isn't yet working on the rust side. In order for the folks working on that stuff to make progress they'll have to disable TLS here. In the referenced PR, there is a flag to disable TLS when you run `applications-publisher`.... but I'm not sure how we best make that work with this helper. Ideas?

afiune: @danielsdeleo  This has nothing to do with TLS. This is a helper method that sends a *raw* message from both NATs Servers. Internal and External.

Do you see any problem with this PR and yours? Do you think they collide? 🤔

danielsdeleo: @afiune for one, there's git conflict since we both edited the applications-publisher. Not a big deal, just calling it out.

The bigger issue is that applications-publisher now needs to know whether it should use TLS when connecting to the server. The reason for this is that in dev, we use a self-signed certificate for TLS, so we have to disable certificate verification. When we pass the `&tls.Config{}` in to NATS with the `InsecureSkipVerify` option, this has the effect of making the NATS client _demand_ TLS--it will hang up on the server if the server isn't using TLS. So the way I "solved" for that was to just add a CLI flag to the applications publisher so you can turn off TLS. 

If folks on the habitat team are going to meaningfully use this helper in their testing, they will need it to work with TLS disabled (for now) since that's how they need to configure automate (because we're waiting for a proper implementation of TLS in the rust NATS library).

That means they can't use your helper as-is because there's no way to pass in the `--disable-tls` flag to applications publisher, so the command will always fail when testing against automate with TLS disabled at the event-gateway.

I'm not sure what feels most nice-to-use to you as the fix for that. One way to go could be to use `shift` to pop the first argument to the helper and then pass everything else verbatim to the `applications-helper`, similar to how I made the wrapper for `make` work https://github.com/chef/automate/blob/b3055c198189323f10b4da4c10185b1ef9374a1e/.studio/golang#L271-L275

components/applications-service/cmd/applications-publisher/main.go
```diff
@@ -22,11 +22,12 @@ NATS Flags:
 	--uniq-client-id     Generate a unique client-id to connect to server
 	--infinite-stream    Publish message every second intinitely
 	--internal-nats      Connect to the Automate Internal NATS Server
+	--disable-tls        Disables TLS when connecting to the server (only applies for event-gateway/external NATS)
```
afiune: @danielsdeleo I moved this parameter to the flags section.

PR #158, opened 2019-04-23
use test framework to spin up a2 cluster

jaym:
Currently, run's 2 A2 nodes and 2 HA backend nodes

```
EXTERNAL_ONLY=1 HAB_ORIGIN=jaym ./integration/run_test ./integration/tests/multi_node.sh
```

Still todo is a load balancer

### DIFF
--- a/integration/services/automate_cluster/init
+++ b/integration/services/automate_cluster/init
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+AUTOMATE_CLUSTER_DIR=$(dirname ${BASH_SOURCE[0]})
+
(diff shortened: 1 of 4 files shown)


integration/services/automate_cluster/init
```diff
@@ -0,0 +1,52 @@
...
+}
+
+automate_cluster_dump_logs() {
+    :
+}
+
+automate_cluster_teardown() {
+    docker stop "$automate_cluster_container1"
+    docker stop "$automate_cluster_container2"
+    docker stop "$automate_cluster_container_nginx"
```
yzl: [nit] it doesn't look like we're starting the nginx container?

Issue #149, opened 2019-04-23
Compliance Scan Results nodes sorted by end_time


alexpop:
In Compliance reporting, on a profile page, for example:

`https://automate.example.com/compliance/reporting/profiles/bf5ffac4d08aadd4120cf2316dc1633d90f0f2965b69c2974f925d8bed79c07f`

or on the reporting profiles page:

`https://automate.example.com/compliance/reporting/profiles`

if the `Scan Results` button is clicked, the sidebar lists the nodes using the default `latest_report.end_time` ASC sorting. Since only the node name is displayed, the list looks unsorted.

Screenshot from acceptance:
![Screen Shot 2019-04-23 at 4 27 24 PM](https://user-images.githubusercontent.com/107378/56594479-d3179e00-65e4-11e9-9084-1868e3f72670.png)

I propose to sort the list of node names alphabetically.

alexpop:
We brainstormed this during standup.

We considered a few things:
 * Node list on reporting page is sorted by default on `latest_report.end_time` (DESC)
 * Node list in the sidebar is limited to 1000 nodes without pagination, sorted by `latest_report.end_time` (ASC)

For consistency with the reporting page and due to the lack of pagination for the sidebar, we decided to maintain the current `latest_report.end_time` sorting (but DESC instead of ASC) for the sidebar nodes, and make the end_time sorting more obvious.

PR #164, opened 2019-04-24
 Add scan time on Scan Results node lists

alexpop:
Fixes https://github.com/chef/automate/issues/149

On top of adding the scan time under the node name, this PR also changes the `latest_report.end_time` sort order from  ASC(default) to DESC to put the recently scanned nodes at the top of the list.

Result:
Screen Shot 2019-04-24 at 11 09 50 AM


alexpop:
Thank you Victoria, polishing things as we go.

PR #166, opened 2019-04-24
Better UX for zero projects case

tylercloke:

If there are no projects, show better UX for the list view.

Screen Shot 2019-04-24 at 7 58 08 AM

- [ ] Code actually executed?

### DIFF
--- a/components/automate-ui/src/app/pages/project/list/project-list.component.html
+++ b/components/automate-ui/src/app/pages/project/list/project-list.component.html
@@ -14,37 +14,49 @@
     </section>
 
     <section *ngIf="(iamMinorVersion$ | async) === 'v1'" class="page-body">
-      <chef-toolbar>
-        <app-authorized [allOf]="['/iam/v2beta/projects', 'post']">
-          <chef-button id="create-button" primary (click)="openCreateModal()">Create Project</chef-button>
+      <ng-container *ngIf="(sortedProjects$ | async)?.length > 0">
+        <chef-toolbar>
+          <app-authorized [allOf]="['/iam/v2beta/projects', 'post']">
+            <chef-button id="create-button" primary (click)="openCreateModal()">Create Project</chef-button>
+          </app-authorized>
+        </chef-toolbar>
+        <app-authorized [allOf]="['/iam/v2beta/projects', 'get']">
+          <chef-table>
+            <chef-thead>
+              <chef-tr>
+                <chef-th>Name</chef-th>
+                <chef-th>Type</chef-th>
+                <chef-th class="controls"></chef-th>
+              </chef-tr>
+            </chef-thead>
+            <chef-tbody>
+              <chef-tr *ngFor="let project of sortedProjects$ | async">
+                <chef-td>
+                  <a [routerLink]="['/settings/projects', project.id]">{{ project.name }}</a>
+                </chef-td>
+                <chef-td>{{ project.type | iamType }}</chef-td>
+                <chef-td class="controls">
+                  <app-authorized [allOf]="['/iam/v2beta/projects/{id}', 'delete', project.id]">
+                    <chef-control-menu>
+                      <chef-option (click)="startProjectDelete(project)">Delete Project</chef-option>
+                    </chef-control-menu>
+                  </app-authorized>
+                </chef-td>
+              </chef-tr>
+            </chef-tbody>
+          </chef-table>
         </app-authorized>
-      </chef-toolbar>
-      <app-authorized [allOf]="['/iam/v2beta/projects', 'get']">
-        <chef-table>
-          <chef-thead>
-            <chef-tr>
-              <chef-th>Name</chef-th>
-              <chef-th>Type</chef-th>
-              <chef-th class="controls"></chef-th>
-            </chef-tr>
-          </chef-thead>
-          <chef-tbody>
-            <chef-tr *ngFor="let project of sortedProjects$ | async">
-              <chef-td>
-                <a [routerLink]="['/settings/projects', project.id]">{{ project.name }}</a>
-              </chef-td>
-              <chef-td>{{ project.type | iamType }}</chef-td>
-              <chef-td class="controls">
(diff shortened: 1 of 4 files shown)


components/automate-ui/src/app/pages/project/list/project-list.component.html
```diff
@@ -14,37 +14,49 @@
...
-        </chef-table>
-      </app-authorized>
+      </ng-container>
+      <ng-container *ngIf="(sortedProjects$ | async)?.length === 0">
+        <div class="empty-case-container">
+          <h1>Create the first project to get started!</h1>
+        </div>
+        <div class="empty-case-container">
+          <app-authorized [allOf]="['/iam/v2beta/projects', 'post']">
+            <chef-button id="create-button" primary (click)="openCreateModal()">Create Project</chef-button>
```
msorens: `id` attributes must be unique. Also, the only reason the other `chef-button` has the `id` is for a unit test. So I suspect you have a unit test failure due to this. I would suggest just remove it here unless you plan to add a unit test for this case.

components/automate-ui/src/app/pages/project/list/project-list.component.html
```diff
@@ -14,37 +14,49 @@
...
-          </chef-tbody>
-        </chef-table>
-      </app-authorized>
+      </ng-container>
+      <ng-container *ngIf="(sortedProjects$ | async)?.length === 0">
+        <div class="empty-case-container">
+          <h1>Create the first project to get started!</h1>
+        </div>
+        <div class="empty-case-container">
+          <app-authorized [allOf]="['/iam/v2beta/projects', 'post']">
```
msorens: This will look rather odd if the user is not authorized, having the friendly greeting (L52) but no button! Seems like the `app-authorized` for this case needs to move out to the `ng-container`... but then the screen will be blank.
Suggest a UX intervention -- @susanev ?

susanev: good point! hmm, i think we need a totally separate message in that case. i need some time to think about it. think we could come back to that later? or should we solve that now?

susanev: @tylercloke if its possible to do in this pr, here is a design for the case @msorens describes.
https://zpl.io/VkpeWKK

tylercloke: @msorens Do we have a way to say `not authorized` in the HTML? Not that I'm aware of 🤔. 

```
          <div class="empty-case-container">
            <h1>Create the first project to get started!</h1>
          </div>
          <div class="empty-case-container">
              <chef-button primary (click)="openCreateModal()">Create Project</chef-button>
          </div>
        </app-authorized>
        <ELSE?>
```

msorens: Copying my reply here as well, just for completeness:
Build an `authorizedChecker`, call its `setPermissions`, subscribe to its `isAuthorized$` observable, and plug that into an `ngIf`. :boom: See working example in landing.component.ts.

components/automate-ui/src/app/pages/project/list/project-list.component.html
```diff
@@ -14,37 +14,57 @@
...
+                    </chef-control-menu>
+                  </app-authorized>
+                </chef-td>
+              </chef-tr>
+            </chef-tbody>
+          </chef-table>
+        </app-authorized>
+      </ng-container>
+      <ng-container *ngIf="(sortedProjects$ | async)?.length === 0">
         <app-authorized [allOf]="['/iam/v2beta/projects', 'post']">
```
bcmdarroch: Let's add `'/iam/v2beta/projects', 'get'` to the permission check in L51

components/automate-ui/src/app/pages/project/list/project-list.component.html
```diff
@@ -14,37 +14,57 @@
...
-          <chef-button id="create-button" primary (click)="openCreateModal()">Create Project</chef-button>
+          <div class="empty-case-container">
+            <h1>Create the first project to get started!</h1>
+          </div>
+          <div class="empty-case-container">
+              <chef-button primary (click)="openCreateModal()">Create Project</chef-button>
+          </div>
+        </app-authorized>
+        <app-authorized [allOf]="['/iam/v2beta/projects', 'post']" [notAuthorized]="true">
```
bcmdarroch: it's a bit unclear how `notAuthorized` works with `allOf`. could we refactor introspection to include a `noneOf` option?

components/automate-ui/src/app/pages/project/list/project-list.component.html
```diff
@@ -14,37 +14,57 @@
...
+              </chef-tr>
+            </chef-tbody>
+          </chef-table>
+        </app-authorized>
+      </ng-container>
+      <ng-container *ngIf="(sortedProjects$ | async)?.length === 0">
         <app-authorized [allOf]="['/iam/v2beta/projects', 'post']">
-          <chef-button id="create-button" primary (click)="openCreateModal()">Create Project</chef-button>
+          <div class="empty-case-container">
+            <h1>Create the first project to get started!</h1>
```
susanev: sorry! didnt look closely enough, these should be paragraphs, not headings. its important for semantics and accessibility.

components/automate-ui/src/app/pages/project/list/project-list.component.html
```diff
@@ -14,37 +14,57 @@
...
+          </div>
+          <div class="empty-case-container">
+              <chef-button primary (click)="openCreateModal()">Create Project</chef-button>
+          </div>
+        </app-authorized>
+        <app-authorized not [allOf]="['/iam/v2beta/projects', 'post']">
+          <div class="empty-case-container">
+            <h1>It looks like no one has created any projects yet and you<br/>
```
susanev: same here, use a paragraph please

components/automate-ui/src/app/pages/project/list/project-list.component.html
tylercloke: @msorens simplified this a bit.

PR #172, opened 2019-04-25
[A2-248] Add global projects-filter components

scottopherson:

Adds some UI components for the global projects filter and wires them up to static placeholder data.

https://chefio.atlassian.net/browse/A2-248

- [x] Code actually executed?

### DIFF
--- a/components/automate-ui/src/app/page-components/navbar/navbar.component.html
+++ b/components/automate-ui/src/app/page-components/navbar/navbar.component.html
@@ -35,6 +35,7 @@
       </div>
     </div>
     <div class="right-nav" role="menubar">
+      <app-projects-filter></app-projects-filter>
       <app-profile></app-profile>
     </div>
   </nav>
--- a/components/automate-ui/src/app/page-components/navbar/navbar.component.html
+++ b/components/automate-ui/src/app/page-components/navbar/navbar.component.html
@@ -35,7 +35,9 @@
       </div>
     </div>
     <div class="right-nav" role="menubar">
-      <app-projects-filter></app-projects-filter>
+      <app-projects-filter
+        *ngIf="(iamMajorVersion$ | async) === 'v2' && (iamMinorVersion$ | async) === 'v1'">
+      </app-projects-filter>
       <app-profile></app-profile>
     </div>
   </nav>
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
@@ -28,7 +28,7 @@
 
   .selection-label {
     display: inline-block;
-    max-width: 150px;
+    max-width: 180px;
     white-space: nowrap;
     overflow: hidden;
     text-overflow: ellipsis;
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
@@ -1,0 +1,116 @@
+@import "~styles/variables";
+@import "~styles/mixins";
+
+:host {
+  display: block;
+  position: relative;
+  padding-left: 35px;
+}
+
+.dropdown-label {
+  display: flex;
+  align-items: center;
+  justify-content: flex-end;
+  position: relative;
+  z-index: 2;
+  padding: 10px 0;
+  margin-left: auto;
+  cursor: pointer;
+  border: none;
+  background: transparent;
+  font-weight: 600;
+
+  > chef-icon {
+    font-size: 16px;
+    color: $chef-primary-dark;
+    margin-left: 8px;
+  }
+
+  .selection-label {
+    display: inline-block;
+    max-width: 150px;
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+  }
+
+  .selection-count {
+    display: inline-block;
+    width: 20px;
+    height: 20px;
+    font-size: 14px;
+    line-height: 20px;
+    text-align: center;
+    border-radius: 50%;
+    margin-left: 10px;
+    background: $chef-grey;
+
+    &.active {
+      background: $chef-primary-bright;
+      color: $chef-white;
+    }
+  }
+}
+
+.dropdown {
+  display: block;
+  position: absolute;
+  right: -10px;
+  max-width: 250px;
+  border: 1px solid $chef-grey;
+  font-size: 14px;
+  z-index: 3;
+
+  &::before {
+    display: block;
+    position: absolute;
+    top: -6px;
+    right: 12px;
+    width: 10px;
+    height: 10px;
+    background: $chef-white;
+    transform: rotate(45deg);
+    border-left: 1px solid $chef-grey;
+    border-top: 1px solid $chef-grey;
+    content: '';
+  }
+
+  .dropdown-body {
+    display: block;
+
+    > span {
+      display: block;
+      color: $chef-dark-grey;
+      text-transform: uppercase;
+      font-size: 11px;
+      padding: 9px 18px 0;
+    }
+
+    chef-checkbox {
+      display: flex;
+      padding: 9px 18px;
+      align-items: start;
+      font-weight: 600;
+
+      &:first-child {
+        padding-top: 18px;
+      }
+
+      &:last-child {
+        padding-bottom: 18px;
+      }
+
+      ::ng-deep .label-wrap {
+        overflow: hidden;
+        text-overflow: ellipsis;
+        white-space: nowrap;
+      }
+    }
+  }
+
+  .dropdown-footer {
+    display: flex;
+    justify-content: center;
+    padding: 0 8px 8px 8px;
+  }
+}
--- a/components/automate-ui/src/app/services/projects-filter/projects-filter.actions.ts
+++ b/components/automate-ui/src/app/services/projects-filter/projects-filter.actions.ts
@@ -1,0 +1,35 @@
+import { Action } from '@ngrx/store';
+import { HttpErrorResponse } from '@angular/common/http';
+import { ProjectsFilterOption } from './projects-filter.reducer';
+
+export enum ProjectsFilterActionTypes {
+  LOAD_OPTIONS         = 'PROJECTS_FILTER::LOAD_OPTIONS',
+  LOAD_OPTIONS_SUCCESS = 'PROJECTS_FILTER::LOAD_OPTIONS::SUCCESS',
+  LOAD_OPTIONS_FAILURE = 'PROJECTS_FILTER::LOAD_OPTIONS::FAILURE',
+  SAVE_OPTIONS =         'PROJECTS_FILTER::SAVE_OPTIONS'
+}
+
+export class LoadOptions implements Action {
+  readonly type = ProjectsFilterActionTypes.LOAD_OPTIONS;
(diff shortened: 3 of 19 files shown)


components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
```diff
@@ -0,0 +1,116 @@
...
+    top: -6px;
+    right: 12px;
+    width: 10px;
+    height: 10px;
+    background: $chef-white;
+    transform: rotate(45deg);
+    border-left: 1px solid $chef-grey;
+    border-top: 1px solid $chef-grey;
+    content: '';
+  }
```
tylercloke: This is some heckin' crazy css. A comment might be a bit helpful to the less experienced :)

components/automate-ui/src/app/services/projects-filter/projects-filter.service.ts
```diff
@@ -0,0 +1,76 @@
...
+        label: 'Super Duper Project',
+        checked: false
+      }
+      // {
+      //   value: 'unassigned-resources',
+      //   label: 'Unassigned Resources',
+      //   checked: false
+      // }
+    ]);
+  }
```
tylercloke: This is a bit different from the pattern used in `components/automate-ui/src/app/entities/*.requests.ts`. Curious what the motivation is to use a separate service vs when it is correct to add a new entity.

scottopherson: It's just placeholder for async data; I'm not sure where projects available to the user will be loaded from or the shape they'll be received in. This can be replaced with a `*requests.ts` file or whatever else.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
scottopherson: This `::before` block is drawing and positioning the small white triangle in the top right corner of the dropdown overlay:

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  if (checked.length === options.length) {
+    return 'All projects';
+  }
+
+  if (checked.length > 1) {
+    return 'Multiple projects';
+  }
+
+  return 'All resources';
```
bcmdarroch: Question for @susanev: I'm not sure the difference between `All Projects` and `All Resources` is clear. Maybe `Unfiltered` or `No Projects Selected`?

components/automate-ui/src/app/page-components/navbar/navbar.component.html
```diff
@@ -35,6 +35,7 @@
       </div>
     </div>
     <div class="right-nav" role="menubar">
+      <app-projects-filter></app-projects-filter>
```
bcmdarroch: In order to hide this behind the beta 2.1 flag, you'll want to add an IAM minor version check.
https://github.com/chef/automate/blob/master/components/automate-ui/src/app/pages/project/list/project-list.component.ts#L53
https://github.com/chef/automate/blob/master/components/automate-ui/src/app/pages/project/list/project-list.component.html#L16

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.ts
susanev: So the difference between the two is only that _All Resources_ includes unassigned, since from the users perspective that is not a project. It does not have anything to do with whether a filter is applied or not. I haven't had a chance to review this yet though, so I will take a closer look later today.

components/automate-ui/src/app/page-components/navbar/navbar.component.html
scottopherson: 👍 Pushed a commit that adds a check for the 2.1 flag:  https://github.com/chef/automate/pull/172/commits/e0dbabdeabf6a4c8cf7ba83aa3bd755a62b16a21

It seems the iam version isn't fetched until you navigate to the Settings view though. We'll need to make that happen on app load instead.

components/automate-ui/src/app/services/projects-filter/projects-filter.effects.ts
```diff
@@ -0,0 +1,44 @@
...
+    private actions$: Actions,
+    private projectsFilter: ProjectsFilterService
+  ) { }
+
+  @Effect()
+  loadOptions$ = this.actions$.pipe(
+    ofType<LoadOptions>(ProjectsFilterActionTypes.LOAD_OPTIONS),
+    switchMap(() => {
+      return this.projectsFilter.fetchOptions().pipe(
+        map((fetched: ProjectsFilterOption[]) => {
```
msorens: What does this code do (L29-33) ?

components/automate-ui/src/app/services/projects-filter/projects-filter.service.ts
scottopherson: > Curious what the motivation is to use a separate service vs when it is correct to add a new entity.

It's actually still unclear to me what we should be considering an 'entity' and modeled as such (things that live in a DB somewhere and/or come from an API?).

But as for the service itself, its main purpose is to keep all the ngrx bits out of components (and other parts of the system that may need access to this data: interceptors, other services, etc): https://medium.com/@thomasburlesonIA/ngrx-facades-better-state-management-82a04b9a1e39

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.ts
```diff
@@ -0,0 +1,110 @@
...
+        set('selectionCountVisible', selectionCountVisible(action.payload)),
+        set('selectionCountActive', selectionCountActive(action.payload))
+      )(state) as ProjectsFilterState;
+    }
+  }
+
+  return state;
+}
+
+function selectionLabel(options: ProjectsFilterOption[]): string {
```
msorens: In this function and elsewhere it looks like you've implemented the line items under "Access to more than 1 project and not the unassigned resources" from A2-248 but not the "Access to at least 1 project and the unassigned resources".

(And uncommenting "Unassigned Resources" in the sample data seems to confirm that.)

components/automate-ui/src/app/services/projects-filter/projects-filter.actions.ts
```diff
@@ -0,0 +1,35 @@
+import { Action } from '@ngrx/store';
+import { HttpErrorResponse } from '@angular/common/http';
+import { ProjectsFilterOption } from './projects-filter.reducer';
+
+export enum ProjectsFilterActionTypes {
```
msorens: I was scratching my head at first trying to figure out what "options" this refers to throughout the code... It would help a lot, I think, to globally say "project" instead of "option".

components/automate-ui/src/app/services/projects-filter/projects-filter.service.ts
```diff
@@ -0,0 +1,76 @@
...
+        checked: false
+      },
+      {
+        value: 'super-project',
+        label: 'Super Duper Project',
+        checked: false
+      }
+      // {
+      //   value: 'unassigned-resources',
+      //   label: 'Unassigned Resources',
```
msorens: "Unassigned Resources" needs to fit without truncation when dropdown is closed.

msorens:
Generally looks great, Scott! Just a few questions/comments posted.

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.ts
scottopherson: @msorens that's still a todo. I wasn't sure how y'all were going to want to include 'unassigned resources' with all the other projects options available to a user. Will it have a special ID or some other flag to differentiate it from projects?

components/automate-ui/src/app/services/projects-filter/projects-filter.effects.ts
scottopherson: At the moment this PR has the projectOptions and their `checked` state being stored in localstorage to persist across page refreshes, sessions, etc.

This code is grabbing the previously saved projectOptions from localstorage (`restored`), and if any of them has the same `value` as one of the newly fetched projectOptions (`fetched`), its `checked` status is merged with it to create the final list of available options (`loaded`).

components/automate-ui/src/app/services/projects-filter/projects-filter.actions.ts
scottopherson: Dang, [`ProjectsFilterOption`](https://github.com/chef/automate/pull/172/files/e0dbabdeabf6a4c8cf7ba83aa3bd755a62b16a21#diff-b65e56f9fdc34d14ce2d28451cefeccaR8) is what I had intended to be in your head. I didn't go with just `Project` since that's a separate entity type not to be confused with. I originally had `LOAD_PROJECTS_FILTER_OPTIONS` but `ProjectsFilter.LoadProjectsFilterOptions` felt redundant compared to `ProjectsFilter.LoadOptions`.

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.ts
scottopherson: Eh, needed `if` statements will be the same regardless of what y'all decide so I'll go ahead and tweak this code.

components/automate-ui/src/app/services/projects-filter/projects-filter.service.ts
susanev: @scottopherson could you update the max-width on the top label to be 163px which i believe is the minimum to display _Unassigned resoureces_ up there fully

components/automate-ui/src/app/services/projects-filter/projects-filter.effects.ts
msorens: Nice! Thanks for the details!

components/automate-ui/src/app/services/projects-filter/projects-filter.actions.ts
msorens: Probably worth getting a consensus on this before changing it.

components/automate-ui/src/app/page-components/navbar/navbar.component.html
bcmdarroch: would that mean adding a call to ui.component.ts?

components/automate-ui/src/app/services/projects-filter/projects-filter.service.ts
scottopherson: [I updated it to 180px](https://github.com/chef/automate/pull/172/commits/e98d66c0310d47b53ad96a422f8d68a9ab863f42) before seeing @susanev's comment. For me it truncates around 170px in Chrome and FF so I did 180px to allow margin of error.

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.ts
scottopherson: https://github.com/chef/automate/pull/172/commits/8c9c813f05108db925930ff535cee3a45a48dba6

components/automate-ui/src/app/services/projects-filter/projects-filter.actions.ts
bcmdarroch: I'm alright with `options` as it differentiates them from project management calls to the Projects API. Adding `ProjectsFilterOption` where possible makes it more clear

PR #177, opened 2019-04-26
Project filtering for reporting server list suggestions

phiggins:
#66

### DIFF
--- a/components/compliance-service/api/reporting/server/server.go
+++ b/components/compliance-service/api/reporting/server/server.go
@@ -93,6 +93,10 @@ func (srv *Server) ListSuggestions(ctx context.Context, in *reporting.Suggestion
 		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Parameter 'type' not specified"))
 	}
 	formattedFilters := formatFilters(in.Filters)
+	formattedFilters, err := filterByProjects(ctx, formattedFilters)
+	if err != nil {
+		return nil, utils.FormatErrorMsg(err, "")
+	}
 	suggestionsList, err := srv.es.GetSuggestions(in.Type, formattedFilters, in.Text, in.Size)
 	if err != nil {
 		return nil, utils.FormatErrorMsg(err, "")
--- a/components/compliance-service/integration_test/reporting_server_list_suggestions_test.go
+++ b/components/compliance-service/integration_test/reporting_server_list_suggestions_test.go
@@ -1,0 +1,207 @@
+package integration_test
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/golang/protobuf/jsonpb"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	iam_v2 "github.com/chef/automate/api/interservice/authz/v2"
+	authzConstants "github.com/chef/automate/components/authz-service/constants/v2"
+	"github.com/chef/automate/components/compliance-service/api/reporting"
+	reportingServer "github.com/chef/automate/components/compliance-service/api/reporting/server"
+	"github.com/chef/automate/components/compliance-service/ingest/events/compliance"
+	"github.com/chef/automate/components/compliance-service/reporting/relaxting"
+)
+
+func ingestReport(fileName string, f func(*compliance.Report)) error {
+	fileData, err := os.Open(fileName)
+	if err != nil {
+		return err
+	}
+	defer fileData.Close()
+	var iReport compliance.Report
+	unmarshaler := &jsonpb.Unmarshaler{AllowUnknownFields: true}
+	if err = unmarshaler.Unmarshal(fileData, &iReport); err != nil {
+		return err
+	}
+
+	f(&iReport)
+
+	ctx := context.Background()
+	_, err = suite.ComplianceIngestServer.ProcessComplianceReport(ctx, &iReport)
+
+	return err
+}
+
+func waitFor(f func() bool) {
+	period := time.Millisecond * 10
+
+	for {
+		if f() {
+			break
+		}
+
+		time.Sleep(period)
+	}
+}
+
+func TestReportingListSuggestions(t *testing.T) {
+	reportFileName := "../ingest/examples/compliance-success-tiny-report.json"
+	everythingCtx := contextWithProjects([]string{authzConstants.AllProjectsExternalID})
+
+	server := reportingServer.New(&relaxting.ES2Backend{ESUrl: elasticsearchUrl})
+
+	n := 5
+
+	reportIds := make([]string, n)
+
+	for i := 0; i < n; i++ {
+		err := ingestReport(reportFileName, func(r *compliance.Report) {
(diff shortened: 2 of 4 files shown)


components/compliance-service/api/reporting/server/server.go
```diff
@@ -93,6 +93,10 @@ func (srv *Server) ListSuggestions(ctx context.Context, in *reporting.Suggestion
 		return nil, status.Error(codes.InvalidArgument, fmt.Sprintf("Parameter 'type' not specified"))
 	}
 	formattedFilters := formatFilters(in.Filters)
+	formattedFilters, err := filterByProjects(ctx, formattedFilters)
+	if err != nil {
+		return nil, utils.FormatErrorMsg(err, "")
+	}
```
phiggins: This is the only change to the application code, the rest is tests.

components/compliance-service/integration_test/reporting_server_list_suggestions_test.go
```diff
@@ -0,0 +1,206 @@
...
+		return err
+	}
+
+	f(&iReport)
+
+	ctx := context.Background()
+	_, err = suite.ComplianceIngestServer.ProcessComplianceReport(ctx, &iReport)
+
+	return err
+}
```
phiggins: This function takes the filename of one of our json profiles on disk, reads it, and sends it through the normal ingestion pipeline. This was done so that these tests could be somewhat unaware of where the data was coming from as far as summary index vs report index in elastic.

components/compliance-service/integration_test/reporting_server_list_suggestions_test.go
```diff
@@ -0,0 +1,206 @@
...
+		return err
+	}
+	defer fileData.Close()
+	var iReport compliance.Report
+	unmarshaler := &jsonpb.Unmarshaler{AllowUnknownFields: true}
+	if err = unmarshaler.Unmarshal(fileData, &iReport); err != nil {
+		return err
+	}
+
+	f(&iReport)
```
phiggins: This function takes a function that takes the report as an argument, so that it could be modified after reading from disk and json parsing, but before being sent through ingestion.

components/compliance-service/integration_test/reporting_server_list_suggestions_test.go
```diff
@@ -0,0 +1,206 @@
...
+			r.Platform.Name = id
+			r.Profiles[0].Controls = r.Profiles[0].Controls[:1]
+			r.Profiles[0].Controls[0].Id = id
+			r.Profiles[0].Controls[0].Title = id
+			r.Profiles = r.Profiles[:1]
+			r.Profiles[0].Sha256 = id
+			r.Profiles[0].Title = id
+			r.Recipes = []string{id}
+			r.ReportUuid = id
+			r.Roles = []string{id}
```
phiggins: This is probably the grossest part of this test. Because this test is testing suggestions, I need to do two things:
* Make each report have a unique value in the fields that we suggest on so that they don't get rolled up into one,
* Make it so that the suggestions returned can be tied back to the reports they came from.

components/compliance-service/integration_test/reporting_server_list_suggestions_test.go
```diff
@@ -0,0 +1,206 @@
...
+		})
+
+		require.NoError(t, err)
+	}
+
+	waitFor(func() bool {
+		response, _ := server.ListReports(everythingCtx, &reporting.Query{})
+
+		return response != nil && len(response.Reports) == n
+	})
```
phiggins: The ingestion pipeline is async (?), so wait until the report listing endpoint says it has as many reports as we tried to ingest.

components/compliance-service/integration_test/reporting_server_list_suggestions_test.go
```diff
@@ -0,0 +1,206 @@
...
+
+	reportsProjects := map[string][]string{
+		"project1": reportIds[1:3],
+		"project2": reportIds[2:5],
+		"project3": reportIds[3:],
+	}
```
phiggins: This is controlling which reports belong to which projects. This is the same logic as used in the list reports endpoint test, just inverted:

https://github.com/chef/automate/blob/945ce3f47b126336ce3dadabe2992145ef14f696/components/compliance-service/integration_test/reporting_server_list_reports_test.go#L17-L36

components/compliance-service/integration_test/reporting_server_list_suggestions_test.go
```diff
@@ -0,0 +1,206 @@
...
+			description:     "user has access to projects without reports and unassigned reports",
+			allowedProjects: []string{"project4", "project5", authzConstants.UnassignedProjectID},
+			expectedIds:     reportIds[:1],
+		},
+		{
+			description:     "user has access to unassigned reports",
+			allowedProjects: []string{authzConstants.UnassignedProjectID},
+			expectedIds:     reportIds[:1],
+		},
+	}
```
phiggins: These were copied pretty much verbatim from https://github.com/chef/automate/blob/945ce3f47b126336ce3dadabe2992145ef14f696/components/compliance-service/integration_test/reporting_server_list_reports_test.go. Only the descriptions have been changed.

components/compliance-service/integration_test/reporting_server_list_suggestions_test.go
```diff
@@ -0,0 +1,206 @@
...
+
+	suggestionTypes := []string{
+		"environment",
+		"node",
+		"platform",
+		"profile",
+		"recipe",
+		"role",
+		"control",
+	}
```
phiggins: This is all the types of suggestions currently supported.

components/compliance-service/integration_test/reporting_server_list_suggestions_test.go
```diff
@@ -0,0 +1,206 @@
...
+	iam_v2 "github.com/chef/automate/api/interservice/authz/v2"
+	authzConstants "github.com/chef/automate/components/authz-service/constants/v2"
+	"github.com/chef/automate/components/compliance-service/api/reporting"
+	reportingServer "github.com/chef/automate/components/compliance-service/api/reporting/server"
+	"github.com/chef/automate/components/compliance-service/ingest/events/compliance"
+	"github.com/chef/automate/components/compliance-service/reporting/relaxting"
+	rules_tags "github.com/chef/automate/lib/authz"
+)
+
+func ingestReport(fileName string, f func(*compliance.Report)) error {
```
lancewf: This is a nice tool for testing! I see moving this into the suit once other tests start using it.

rickmarry: At first, I was wondering why we want to pass the func instead of just doing the work of the func right in this function, then I realized how flexible this is in our ability to change the report to our liking from the calling side! 
Nice!

components/compliance-service/integration_test/reporting_server_list_suggestions_test.go
rickmarry: This looks like it is also a candidate for reuse.

PR #183, opened 2019-04-26
Optional EBS snapshot restore for ec2/vagrant setup

alexpop:
With this change, the ec2 Vagrant setup is not restoring EBS volumes by default for the elasticsearch data directory. And can restore any snapshot now.


alexpop:
⭐️ ⭐️

PR #188, opened 2019-04-26
notes for adding data in automate for compliance, scan jobs, node integrations

vjeffrey:

Just adding some notes on how to add data in Automate for compliance reporting, scan jobs, and node integrations.
I've also added a docker test target in the create-pg-data script. I'm not convinced this will actually work from the dev environment, since we're in a vagrant box or docker box and reaching out to a local docker machine isn't realllly possible, but maybe it works for the docker dev env as it is testing itself?anyways, I'm leaving it there for now as having an unreachable node is also good for dummy data.  

This also makes me wonder why we don't allow users to add docker containers for scanning in Automate? I'll look into this more..

### :+1: Definition of Done
Notes on how to add data in dev environment.

### DIFF
--- a/components/compliance-service/docs/adding_test_data.md
+++ b/components/compliance-service/docs/adding_test_data.md
@@ -6,10 +6,11 @@ from hab studio:
 
 # Adding nodes to populate the nodes and scan jobs pages
 
-_note: this should be run on your local system, not from within the studio_
-
 ensure you have `jq` installed
+the token is retrieved by running `get_admin_token` from within the studio
+if you've never run the get_secrets script, or haven't in a while, run `CHEF_USERNAME=username scripts/get_secrets` from the studio
 
+_note: this should be run on your local system, not from within the studio_
 `source dev/secrets-env.sh`
 `A2_URL='https://a2-dev.test' A2_TOKEN='token_val' components/compliance-service/scripts/create-pg-data.sh`
 
--- a/components/compliance-service/docs/adding_test_data.md
+++ b/components/compliance-service/docs/adding_test_data.md
@@ -1,0 +1,59 @@
+# Adding nodes to populate the compliance reporting page
+
+from hab studio: 
+
+`chef_load_compliance_scans -N20` 
+
+# Adding nodes to populate the nodes and scan jobs pages
+
+_note: this should be run on your local system, not from within the studio_
+
+ensure you have `jq` installed
+
+`source dev/secrets-env.sh`
+`A2_URL='https://a2-dev.test' A2_TOKEN='token_val' components/compliance-service/scripts/create-pg-data.sh`
+
+# Adding profiles
+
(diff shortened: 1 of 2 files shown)


components/compliance-service/docs/adding_test_data.md
```diff
@@ -0,0 +1,59 @@
...
+`chef_load_compliance_scans -N20` 
+
+# Adding nodes to populate the nodes and scan jobs pages
+
+_note: this should be run on your local system, not from within the studio_
+
+ensure you have `jq` installed
+
+`source dev/secrets-env.sh`
+`A2_URL='https://a2-dev.test' A2_TOKEN='token_val' components/compliance-service/scripts/create-pg-data.sh`
```
vjeffrey: maybe i should add a note here about how to retrieve the a2 token?

components/compliance-service/docs/adding_test_data.md
```diff
@@ -0,0 +1,59 @@
...
+from hab studio: 
+
+`chef_load_compliance_scans -N20` 
+
+# Adding nodes to populate the nodes and scan jobs pages
+
+_note: this should be run on your local system, not from within the studio_
+
+ensure you have `jq` installed
+
```
vjeffrey: should i add a quick note here along the lines of "if you haven't run it before or run it in a while, please run `CHEF_USERNAME=username scripts/get_secrets` first" ?

components/compliance-service/scripts/create-pg-data.sh
```diff
@@ -88,7 +107,7 @@ for i in $(seq 1 ${job_count}); do
       "name": "my job",
       "tags": [],
       "type": "exec",
-      "nodes": ['${vagrant_node_id}', '${ec2_node_id}'],
+      "nodes": ['${vagrant_node_id}', '${ec2_node_id}', '${docker_node_id}'],
```
vjeffrey: hmm. i might remove the docker node id from the scan job given that i assume it will always be unreachable

components/compliance-service/docs/adding_test_data.md
lancewf: I tried to run `CHEF_USERNAME=username scripts/get_secrets` and it needed vault. How do you install vault?

lancewf: `brew install vault` worked for me. Yeah, I think it would be nice to add these to the doc somewhere for people first time running this.

vjeffrey: https://github.com/chef/automate/blob/master/dev-docs/DEV_ENVIRONMENT.md#working-with-secrets 
ya i think running that part of it from the studio would also work?

vjeffrey: ill add some instructions

lancewf: Thank you. The steps worked for me.

PR #199, opened 2019-04-28
[multi-node] Replication for authz V2 OPA rules

jaym:
This PR introduces replicating authz policies into the local
OPA engine. Each OPA engine locally needs to stay in sync with
postgres. It is no longer a valid assumption that the updates
to OPA are only needed when authz's APIs are called on the node.
Requests can get routed to any node, so each node will also need
to check with postgres.

With this PR, each change that requires an OPA refresh modifies a
policy change ID in postgres. Whenever a policy refresh is requested,
the first thing that happens is we pull down the change id. If the ID
is not the same one as the last one we've seen, the OPA rules get
refreshed and the new ID is kept track of. On some interval (10
seconds), this check is automatically done.

As a follow up, we're going to use the postgres LISTEN/NOTIFY feature to
speed up replication. The OPA refresh will be kicked of every time a
message is published on the policychange channel. In this context, the
current timer will still be used to provide robustness against missing
an event.

I didn't do the V1 API because we can force people using multi-node to
use the V2 api. If we need it, we can implement it.


jaym:
The data fetched for rebuilding the OPA engine is not done transactionally, so i think there exist sequences that will lead to inconsistency if failovers ever were to happen.

components/authz-service/server/v2/policy_refresher.go
```diff
@@ -0,0 +1,276 @@
...
+		antiEntropyTimer.Reset(refresher.antiEntropyTimerDuration)
+	}
+	close(refresher.refreshRequests)
+}
+
+func (refresher *policyRefersher) refresh(ctx context.Context, lastPolicyID string) (string, error) {
+	curPolicyID, err := refresher.store.GetPolicyChangeID(ctx)
+	if err != nil {
+		refresher.log.WithError(err).Warn("Failed to get current policy change ID")
+		return lastPolicyID, err
```
srenatus: So the first returned value is _OK to use_ even if there's a non-nil error? That's a bit against conventions, I believe. 🤔

components/authz-service/server/v2/policy_refresher.go
```diff
@@ -0,0 +1,276 @@
...
+func (refresher *policyRefersher) Refresh(ctx context.Context) error {
+	m := policyRefresherMessageRefresh{
+		ctx:    ctx,
+		status: make(chan error, 1),
+	}
+	refresher.refreshRequests <- m
+	return m.Err()
+}
+
+func (refresher *policyRefersher) RefreshAsync() error {
```
srenatus: Is this used?

components/authz-service/server/v2/policy.go
```diff
@@ -78,12 +78,15 @@ func NewPoliciesServer(
 	pl storage_v1.PoliciesLister,
 	vChan chan api.Version) (PolicyServer, error) {
 
+	policyRefersher := NewPolicyRefresher(ctx, l, s, e)
```
srenatus: [nit] I guess if this is only used within this package, we could _not export_ it.

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -971,9 +971,11 @@ func TestDeletePolicy(t *testing.T) {
 	cases := map[string]func(*testing.T){
 		"empty database": func(t *testing.T) {
 			ctx := context.Background()
```
srenatus: [nit] This is fine as-is now, but I suppose it might have been more unit-testy to assert the ID (non-)change in one test case per op; not for _every test case_. But I might be wrong 😉

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -792,6 +874,11 @@ func (p *pg) UpdateRole(ctx context.Context, role *v2.Role) (*v2.Role, error) {
 		}
 	}
 
+	err = p.notifyPolicyChange(ctx, tx)
+	if err != nil {
+		return nil, p.processError(err)
+	}
+
```
srenatus: ⚠️ I didn't verify that every function that needs this now has this. But I guess your test change does cover this, after all.

components/authz-service/server/v2/policy_refresher.go
jaym: not sure i understand. The intent is that if there is an error, the change id doesn't change

components/authz-service/server/v2/policy_refresher.go
jaym: nope. I was going to use it for the pg notify stuff, but i'm not going to do that in this pr

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -347,6 +391,19 @@ func (p *pg) associatePolicyWithProjects(ctx context.Context,
...
 }
 
+func (p *pg) notifyPolicyChange(ctx context.Context, q Querier) error {
+	// We keep track of an id with each change. This lets us be smart about only updating
+	// the OPA rules when it might change.
+	_, err := q.ExecContext(ctx, "UPDATE policy_change_tracker SET policy_change_id = uuid_generate_v4();")
+	if err != nil {
+		return err
+	}
+	_, err = q.ExecContext(ctx,
```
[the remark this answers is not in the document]
jaym: we can remove it, i was testing to see if the golang pg notify library would work (it does)

components/authz-service/storage/v2/postgres/postgres.go
jaym: i did my best to line the up with what the interceptor does, but please do double check

components/authz-service/storage/v2/memstore/memstore.go
```diff
@@ -12,20 +13,27 @@ import (
...
 type State struct {
-	policies *cache.Cache
-	roles    *cache.Cache
-	projects *cache.Cache
-	ms       storage.MigrationStatus
+	policies       *cache.Cache
+	roles          *cache.Cache
+	projects       *cache.Cache
+	ms             storage.MigrationStatus
+	policyChangeID int64
```
stevendanna: I think if we are going to use AddInt64 we need to ensure this is 64byte aligned.  According to the sync/atomic docs, one way to do that is to make sure this is the first entry in the struct.  Perhaps we could move this to the top with a comment not to move it to protect against hard-to-find bugs in the future.

stevendanna: Note this is a supa-nit as I think this code is correct on x86 and likely everything we support currently

components/automate-cli/cmd/chef-automate/diagnostics.go
```diff
@@ -62,7 +69,10 @@ func runDiagnosticsRunCmd(cmd *cobra.Command, args []string) error {
 
 	dsClient, err := client.Connection(client.DefaultClientTimeout)
 	if err != nil {
-		return err
+		logrus.WithError(err).Warn("could not connect to deployment service")
+		if diagnosticsRunCmdOpts.adminToken == "" {
+			return err
```
stevendanna: Is the idea here that we can run a lot of the tests with a hand-created adminToken even if we can't connect to ds so we don't need this client later?  I'm cool with this for now, but as a follow up maybe we can move the adminToken retrieval code a bit closer to this code so that is clear.

jaym: that sounds right. i did things so i could run the cli locally

components/authz-service/server/v2/policy.go
```diff
@@ -26,11 +25,12 @@ import (
...
-	store  storage.Storage
-	engine engine.V2Writer
-	v1     storage_v1.PoliciesLister
-	vChan  chan api.Version
+	log             logger.Logger
+	store           storage.Storage
+	engine          engine.V2Writer
+	v1              storage_v1.PoliciesLister
+	vChan           chan api.Version
+	policyRefersher PolicyRefresher
```
blakestier: Looks like a typo/misspelling (s/policyRefersher/policyRefresher)

components/authz-service/server/v2/policy.go
blakestier: same spelling error-- but at least it's consistent!

components/authz-service/server/v2/policy.go
stevendanna: @blakestier wins code review today.

components/authz-service/storage/postgres/migration/sql/43_policy_changed.up.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+CREATE TABLE policy_change_tracker (
+  policy_change_id uuid DEFAULT uuid_generate_v4()
```
blakestier: It's worth adding a `NOT NULL` here

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -347,6 +391,17 @@ func (p *pg) associatePolicyWithProjects(ctx context.Context,
 	return nil
 }
 
+func (p *pg) notifyPolicyChange(ctx context.Context, q Querier) error {
+	// We keep track of an id with each change. This lets us be smart about only updating
+	// the OPA rules when it might change.
+	_, err := q.ExecContext(ctx, "UPDATE policy_change_tracker SET policy_change_id = uuid_generate_v4();")
```
blakestier: I think you might be able to say `UPDATE policy_change_tracker SET policy_change_id = DEFAULT;` but we would lose explicitness there so prob not worth it

PR #200, opened 2019-04-29
[A2-787] authz-service and teams-service db cleanup

msorens:

While performing an audit of our code to examine how project filtering is handled amongst all the auth entities, I noticed a few things that needed clean-up.

1. Swapping in `projects_match` in `query_roles` in lieu of the duplicate, inline code.

2. Removing deprecated functions that were still present due to signature changes due to two issues: (a) `create or replace` only replaces if the function signature matches; otherwise it is just `create`. 
(b) the `drop function` command also requires matching signatures; otherwise it fails silently.

3. DB functions in teams-service are a hodgepodge of v1 and v2 code. It is important to delineate which are v1 only so the casual reader does not get alarmed that project-handling is absent from those.

Examine the individual commits for the details.

### :+1: Definition of Done

Code is a bit cleaner than before.

- [x] Code actually executed?

### DIFF
--- a/components/authz-service/storage/postgres/migration/sql/41_query_roles_with_projects_match.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/41_query_roles_with_projects_match.up.sql
@@ -1,0 +1,30 @@
+BEGIN;
+
+CREATE OR REPLACE FUNCTION
+  query_roles(_projects_filter TEXT[])
+  RETURNS setof json AS $$
+
+  WITH t AS (
+    SELECT
+      r.id,
+      r.name,
+      r.type,
+      r.actions,
+      (SELECT array_agg(rp.project_id) FILTER (WHERE rp.project_id IS NOT NULL)) AS projects
+    FROM iam_roles AS r
+    LEFT OUTER JOIN iam_role_projects AS rp ON rp.role_id = r.db_id
+    GROUP BY r.id, r.name, r.type, r.actions
+  )
+  SELECT
+    json_build_object(
+      'id', t.id,
+      'name', t.name,
+      'type', t.type,
+      'actions', t.actions,
+      'projects', COALESCE(t.projects, '{}')
+    ) AS role FROM t
+  WHERE projects_match(t.projects::TEXT[],  _projects_filter);
+
+$$ LANGUAGE sql;
+
+COMMIT;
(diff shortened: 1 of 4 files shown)


components/authz-service/storage/postgres/migration/sql/41_query_roles_with_projects_match.up.sql
```diff
@@ -0,0 +1,30 @@
...
+  )
+  SELECT
+    json_build_object(
+      'id', t.id,
+      'name', t.name,
+      'type', t.type,
+      'actions', t.actions,
+      'projects', COALESCE(t.projects, '{}')
+    ) AS role FROM t
+  WHERE projects_match(t.projects::TEXT[],  _projects_filter);
```
msorens: The only change in this function is L26 with the call to `projects_match` replacing the inline code from components/authz-service/storage/postgres/migration/sql/38_query_roles_with_project_filter.up.sql.

components/authz-service/storage/postgres/migration/sql/42_remove_leftover_functions.up.sql
```diff
@@ -0,0 +1,13 @@
+BEGIN;
+
+-- These are leftover, deprecated functions, whose signatures have changed.
```
msorens: (That is, these should have been included in earlier migrations but were missed.)

components/teams-service/storage/postgres/postgres.go
```diff
@@ -76,10 +76,12 @@ func (p *postgres) UpgradeToV2(ctx context.Context) error {
 	return nil
 }
 
+// StoreTeam saves a team to the DB. This is used by IAM v1 ONLY!
```
msorens: Primarily adding this "v1 only" comment to appropriate code in this file.

components/authz-service/storage/postgres/migration/sql/42_remove_leftover_functions.up.sql
blakestier: oh THAT's why there were two of them when we were looking the other day-- makes sense.

msorens: Yep; decided to track that down.

PR #157, opened 2019-04-23
[A2-675] Create Header Bar with Status Health Counts

Shadae:
### :nut_and_bolt: Description
Screen Shot 2019-04-23 at 1 39 31 PM

### :+1: Definition of Done
No longer static!! 
### :white_check_mark: Checklist

- [x] Necessary tests added/updated?
- [x] Necessary docs added/updated?
- [x] Code actually executed?
- [ ] Vetting performed (unit tests, lint, etc.)?

PR #201, opened 2019-04-29
Revert commented health bar https://github.com/chef/automate/pull/157

afiune:
The health filter bar was commented out with the PR https://github.com/chef/automate/pull/157, this work is to uncomment it and have the bar visible again.

It also fixed the problem found in the above PR where the Redux state was not well initialized.

EXTRA: Fixed more little nitpicks from the codebase. Also used the selected status variable to make it persistent on reload.

### DIFF
--- a/components/automate-ui/src/app/entities/service-groups/service-groups.model.ts
+++ b/components/automate-ui/src/app/entities/service-groups/service-groups.model.ts
@@ -1,11 +1,11 @@
 import { RollupServiceStatus, SortDirection  } from '../../types/types';
 
 export interface HealthSummary {
-  total?: number;
-  ok?: number;
-  warning?: number;
-  critical?: number;
-  unknown?: number;
+  total: number;
+  ok: number;
+  warning: number;
+  critical: number;
+  unknown: number;
 }
 
 export interface Service {
--- a/components/automate-ui/src/app/entities/service-groups/service-groups.model.ts
+++ b/components/automate-ui/src/app/entities/service-groups/service-groups.model.ts
@@ -1,11 +1,11 @@
 import { RollupServiceStatus, SortDirection  } from '../../types/types';
 
 export interface HealthSummary {
-  total: number;
-  ok: number;
-  warning: number;
-  critical: number;
-  unknown: number;
+  total?: number;
+  ok?: number;
+  warning?: number;
+  critical?: number;
+  unknown?: number;
 }
 
 export interface Service {
--- a/components/automate-ui/src/app/entities/service-groups/service-groups.reducer.ts
+++ b/components/automate-ui/src/app/entities/service-groups/service-groups.reducer.ts
@@ -24,14 +24,26 @@ export interface ServiceGroupEntityState {
 
 export const ServiceGroupEntityInitialState: ServiceGroupEntityState = {
   serviceGroups: [],
-  serviceGroupHealthCounts: { },
+  serviceGroupHealthCounts: {
+    total: 0,
+    ok: 0,
+    warning: 0,
+    critical: 0,
+    unknown: 0
+  },
   status: EntityStatus.notLoaded,
   filters: { },
   servicesStatus: EntityStatus.notLoaded,
   errorResp: null,
   servicesFilters: { },
   servicesList: [],
-  servicesHealthSummary: { },
+  servicesHealthSummary: {
+    total: 0,
+    ok: 0,
+    warning: 0,
+    critical: 0,
+    unknown: 0
+  },
   selectedServiceGroupName: undefined
 };
 
--- a/components/automate-ui/src/app/entities/service-groups/service-groups.reducer.ts
+++ b/components/automate-ui/src/app/entities/service-groups/service-groups.reducer.ts
@@ -31,7 +31,7 @@ export const ServiceGroupEntityInitialState: ServiceGroupEntityState = {
   errorResp: null,
   servicesFilters: { },
   servicesList: [],
-  servicesHealthSummary: undefined,
+  servicesHealthSummary: { },
   selectedServiceGroupName: undefined
 };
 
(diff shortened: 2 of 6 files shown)


components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.ts
```diff
@@ -41,9 +41,9 @@ export class ServicesSidebarComponent implements OnInit {
       (state) => state.servicesHealthSummary));
 
     this.currentPage = 1;
-
-    // this.servicesHealthSummary$.subscribe(healthSummary =>
-    // this.totalServices = healthSummary.total);
+    this.servicesHealthSummary$.subscribe((healthSummary) => {
+      this.totalServices = healthSummary.total;
```
mpeck: If this field may or may not exist I would probably use `get()` from lodash/fp to get the value. It will always return a value even if one of the parent objects in undefined.

It should look something like `this.totalServices = get('total', healthSummary)`

This is the documentation from lodash, https://lodash.com/docs/4.17.11#get, but also take a look at the fp docs to see how to convert the syntax. https://github.com/lodash/lodash/wiki/FP-Guide

components/automate-ui/src/app/entities/service-groups/service-groups.model.ts
```diff
@@ -1,11 +1,11 @@
...
-  total: number;
-  ok: number;
-  warning: number;
-  critical: number;
-  unknown: number;
+  total?: number;
+  ok?: number;
+  warning?: number;
+  critical?: number;
+  unknown?: number;
```
mpeck: Generally I would stay away from optional fields in an interface. Instead you may want to just set the fields with some default values in the initial state.

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.ts
afiune: I can do that! Thanks for your feedback Matt.

components/automate-ui/src/app/entities/service-groups/service-groups.model.ts
afiune: I can do that! Thanks for your feedback Matt.

components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
```diff
@@ -26,40 +26,44 @@ import { find, includes } from 'lodash/fp';
...
+  private defaultFieldDirection: FieldDirection = {
     name: 'ASC',
     percent_ok: 'ASC'
   };
 
-  currentPage = 1;
-  // The number of service groups to display per page
-  pageSize = 25;
-  // TODO: Wire this up with real data
-  totalServiceGroups = 50;
```
afiune: This shouldn't be hardcoded to 50 anymore

components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
```diff
@@ -26,40 +26,44 @@ import { find, includes } from 'lodash/fp';
...
   public servicesSidebarVisible = false;
 
+  // The current page the user is visualizing
+  public currentPage = 1;
+
+  // The number of service groups to display per page
+  public pageSize = 25;
+
+  // Total number of service groups
+  public totalServiceGroups = 0;
```
afiune: I made the variables explicit to know if they are private or public.

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
```diff
@@ -8,8 +8,8 @@
       <chef-subheading>Services run by each Habitat Supervisor on a node.</chef-subheading>
     </chef-page-header>
     <div class="status-filter-bar">
-      <!-- <chef-status-filter-group lean>
-        <chef-option class="filter general" value="general" (click)="updateServicesFilters('total')">
+      <chef-status-filter-group lean>
+        <chef-option class="filter general" value="general" (click)="updateServicesFilters('total')" selected>
```
afiune: Marked the Total chef-option to be the default selected one.

components/automate-ui/src/app/pages/service-groups/service-groups.component.html
```diff
@@ -4,8 +4,8 @@
   <chef-subheading>Group of Habitat services running the same package and configuration.</chef-subheading>
 </chef-page-header>
 <div class="page-body">
-  <chef-status-filter-group>
-    <chef-option class="filter general" value="general" (click)="statusFilter('total')">
+  <chef-status-filter-group [value]="selectedStatus$ | async">
+    <chef-option class="filter general" value="general" (click)="statusFilter('total')" selected>
```
afiune: Marked the Total chef-option to be the default selected one.

components/automate-ui/src/app/pages/service-groups/service-groups.component.html
```diff
@@ -4,8 +4,8 @@
   <chef-subheading>Group of Habitat services running the same package and configuration.</chef-subheading>
 </chef-page-header>
 <div class="page-body">
-  <chef-status-filter-group>
-    <chef-option class="filter general" value="general" (click)="statusFilter('total')">
+  <chef-status-filter-group [value]="selectedStatus$ | async">
```
afiune: This line is using the `selectedStatus` Observable to make it persistent on reload. Before we were not using it at all.

components/automate-ui/src/app/entities/service-groups/service-groups.reducer.ts
```diff
@@ -11,7 +11,7 @@ import {
 
 export interface ServiceGroupEntityState {
   serviceGroups: ServiceGroup[];
-  serviceGroupHealthCounts: HealthSummary[];
+  serviceGroupHealthCounts: HealthSummary;
```
afiune: We didn't need an array of `HealthSummary` but instead a single instance of that interface. Changed that here! 💯

PR #204, opened 2019-04-29
Use upgrade_strategy none for airgap backup test

jaym:
There's currently a bug with this that should be fixed in #203.
Restoring does not use the correct manifest after restore and
deployment-service gets confused.


jaym:
this isn't right. we need to also deploy an old version

PR #207, opened 2019-04-30
[deployment] Set HAB_LICENSE=accept-no-persist in a lot of places

stevendanna:
This is a first pass at making sure we are calling hab with
HAB_LICENSE=no-accept-persist

Because `hab sup run` also requires a license check currently we also
need to set this in our systemd unit file. Since that is inherited by
all processes in our process tree, we are probably OK if we missed a
hab call or two, but I would rather not depend on that. Further, right
now we don't correctly restart our systemd unit file in all cases, so
this likely won't take effect unless chef/automate#90 is merged first.


stevendanna:
This is mostly untested since I was trying trouble getting hab to build locally.

PR #208, opened 2019-04-30
[secrets-service] Use platform secrets store for shared secret

stevendanna:
Currently, secrets-service stores the symmetric encryption key on
disk. Now, it uses the platform secrets code.

Right now the platform secrets store is also an insecure on-disk sore.
However, this makes it a bit easier for us to make sure we can export
all of the known shared secrets in the system.

To mitigate the very confusing naming, I've moved lib/secrets to
lib/platform/secrets.


components/secrets-service/cmd/secrets-service/commands/init.go
```diff
@@ -24,66 +23,71 @@ const (
...
+	// secrets service.
+	secretsServiceKeyPath = "/hab/svc/secrets-service/data/secrets_key"
 	// Number of bytes of random data to use to generate the key
 	defaultKeyLength = 16
-	// Permissions on key
-	defaultKeyMode = 0600
 )
 
 var initCmd = &cobra.Command{
 	Use:   "init KEYPATH",
```
phiggins: This doesn't take a path anymore, right?

stevendanna:
> I didn't know we had a pattern for this.

@phiggins That was a bit purposeful.  When we put this in it was really just to support some legacy bits in chef-server.  To be frank, I'm a bit nervous about spreading its use because as I said it just uses insecure on-disk storage; but, perhaps gives us some leverage to improve it more easily in the future.

stevendanna:
> There's other places in secrets-service and nodemanager-service where we store keys on disk that could probably use the same treatment. 

If you have a minute and want to point these out to me, I can include them in this PR or in a follow up.

phiggins:
> If you have a minute and want to point these out to me, I can include them in this PR or in a follow up.

The ones I was thinking of are where we fetch a node's credential information and hydrate any SSH keys it has by calling this function: https://github.com/chef/automate/blob/master/components/nodemanager-service/api/nodes/node.secrets.go

* https://github.com/chef/automate/blob/master/components/compliance-service/inspec-agent/resolver/resolver.go#L719 
* https://github.com/chef/automate/blob/master/components/nodemanager-service/api/nodes/server/server.go#L150

stevendanna:
We've gone a different way for now.

Issue #209, opened 2019-04-30
Include profile->group when storing profile


rickmarry:
# User Story
When storing a profile, we need to also include the group information of said profile.  Currently, this information is being received upon ingestion but is excluded from our store of it in ES.

vjeffrey:
@rickmarry  was this resolved with https://github.com/chef/automate/pull/211 ?

rickmarry:
@vjeffrey - yes it was.. thanks!

PR #211, opened 2019-04-30
take group info from profile when parsing

rickmarry:

When posting a profile to be ingested, we need to take the group information from the profile itself.

### DIFF
--- a/components/compliance-service/reporting/relaxting/profiles.go
+++ b/components/compliance-service/reporting/relaxting/profiles.go
@@ -88,10 +88,10 @@ func (esprofile *ESInspecProfile) parseInspecProfile(profile inspec.Profile) err
 	esprofile.Sha256 = profile.Sha256
 	// no need for Status and SkipMessage here as it's not static metadata of the profile
 	var groups []inspec.Group
-	for _, group := range groups {
+	for _, group := range profile.Groups {
 		eGroup := inspec.Group{
-			ID:       group.ID,
-			Title:    group.Title,
+			ID:       group.Id,
+			Title:    &group.Title,
 			Controls: group.Controls,
 		}
 		groups = append(groups, eGroup)


components/compliance-service/reporting/relaxting/profiles.go
```diff
@@ -88,10 +88,10 @@ func (esprofile *ESInspecProfile) parseInspecProfile(profile inspec.Profile) err
 	esprofile.Sha256 = profile.Sha256
 	// no need for Status and SkipMessage here as it's not static metadata of the profile
 	var groups []inspec.Group
-	for _, group := range groups {
+	for _, group := range profile.Groups {
 		eGroup := inspec.Group{
-			ID:       group.ID,
-			Title:    group.Title,
+			ID:       group.Id,
+			Title:    &group.Title,
```
stevendanna: I don't think taking a reference to the loop temporary is safe here.

Consider the example here https://play.golang.org/p/rb-sdxW-64v

Rather I think you'll need something like:

```golang
for _, group := range profile.Groups {
        title := group.Title
	eGroup := inspec.Group{
		ID:       group.Id,
		Title:    &title,
		Controls: group.Controls,
	}
}
```

rickmarry: profile.Groups is actually []*Group, so the the example would be more accurately represented as:
https://play.golang.org/p/BcU7Af4Tdut

stevendanna: I see! Cool, this is fine then

rickmarry: cool. thanks!

phiggins:
I think the danger with this, like @vjeffrey mentioned in standup, is that there's a lot of squirrelly data around profiles and we've messed things up "fixing" things in the past. See chef/a2#3158 for the example of when I caused a regression by trying to fix things in here.

rickmarry:
Fixes #209

Issue #62, opened 2019-03-06
rbac on actions


vjeffrey:
# User Story

Upon ingestion of an action, evaluate chef-server and org associated with action, and tag the action with the appropriate project.

- [x] ingestion service depends on authz service
- [x] ingestion service calls authz service for rules
- [x] action is tagged with project based on rules it matches
- [x] action is tagged with correct projects after project update event triggered

# Definition of Done
All incoming actions are evaluated and tagged on ingestion.
 - the only fields applicable for actions are chef-server and organization

End result: projects field on action

# Demo Script / Repro Steps
get some mocked ingest mapping rules
send some actions into a2
see action has been tagged with projects

lancewf:
## PRs
- [ ] Adding project rule types - https://github.com/chef/automate/pull/109

lancewf:
Use this PR to add the tagging and updating of project tags
https://github.com/chef/a2/pull/4736

PR #212, opened 2019-04-30
Elasticsearch actions project update

lancewf:

Adding a Chef Actions project tag updater function. This function takes the authz rule set and creates a painless script that is pushed to elasticsearch. This script walks through each Chef Action and updates their projects based on the rules. 

No demo until Chef actions can be ingested with projects. 

https://github.com/chef/automate/issues/62

This PR is the same work for updating nodes - https://github.com/chef/a2/pull/5301

- [x] Code actually executed?

### DIFF
--- a/components/ingest-service/integration_test/project_update_actions_test.go
+++ b/components/ingest-service/integration_test/project_update_actions_test.go
@@ -1,0 +1,342 @@
+package integration_test
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	iam_v2 "github.com/chef/automate/api/interservice/authz/v2"
+	iBackend "github.com/chef/automate/components/ingest-service/backend"
+	"github.com/chef/automate/components/ingest-service/backend/elastic/mappings"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestProjectUpdateActionsPainlessElasticsearchScript(t *testing.T) {
+	var (
+		ctx = context.Background()
+	)
+
+	cases := []struct {
+		description string
+		action      iBackend.InternalChefAction
+		projects    map[string]*iam_v2.ProjectRules
+		projectIDs  []string
+	}{
+		// Orgs
+		{
+			description: "Org: Single rule matching update",
+			action: iBackend.InternalChefAction{
+				OrganizationName: "org1",
+				Projects:         []string{"old_tag"},
+			},
+			projects: map[string]*iam_v2.ProjectRules{
+				"project9": &iam_v2.ProjectRules{
+					Rules: []*iam_v2.ProjectRule{
+						&iam_v2.ProjectRule{
+							Conditions: []*iam_v2.Condition{
+								&iam_v2.Condition{
+									Type:   iam_v2.ProjectRuleConditionTypes_CHEF_ORGS,
+									Values: []string{"org1"},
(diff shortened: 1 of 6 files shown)


components/ingest-service/integration_test/project_update_actions_test.go
```diff
@@ -0,0 +1,342 @@
...
+
+				jobStatus, err := suite.ingest.JobStatus(ctx, esJobID)
+				assert.Nil(t, err)
+				for !jobStatus.Completed {
+					time.Sleep(time.Millisecond * 5)
+					jobStatus, err = suite.ingest.JobStatus(ctx, esJobID)
+					assert.Nil(t, err)
+					if err != nil {
+						assert.FailNow(t, "testing job complete")
+					}
```
phiggins: I'm not entirely sure what this code is doing, but you might want to replace this if and the `assert.Nil` with `require.Nil` if you want the test to fail straight away.

components/ingest-service/integration_test/project_update_actions_test.go
```diff
@@ -0,0 +1,342 @@
...
+				suite.RefreshIndices(fmt.Sprintf("%s-%s", mappings.Actions.Index, "*"))
+
+				// assert the node's project IDs
+				actualActions, err := suite.GeActions(100)
+				assert.Nil(t, err)
+				assert.Equal(t, 1, len(actualActions), "wrong number of actions retrieved")
+
+				actualAction := actualActions[0]
+
+				assert.ElementsMatch(t, test.projectIDs, actualAction.Projects)
```
phiggins: I would definitely suggest using `require` here, because if an empty list is returned above you'll get a nil pointer exception trying to access the fields on this line.

components/ingest-service/integration_test/project_update_actions_test.go
```diff
@@ -0,0 +1,342 @@
...
+
+				suite.DeleteAllDocuments()
```
phiggins: I think we want to call `defer` before this (or other test cleanup functions). If this function returns before you expect it might never reach this line and tests could start polluting each other making it harder to debug the original failure.

components/ingest-service/integration_test/suite_test.go
```diff
@@ -153,6 +162,17 @@ func (s *Suite) IngestNodes(nodes []iBackend.Node) {
...
 
+// IngestActions ingests a number of actions then refreshe the all the action indexes
+func (s *Suite) IngestActions(actions []iBackend.InternalChefAction) {
+	// Insert actions
+	for _, action := range actions {
+		s.ingest.InsertAction(context.Background(), action)
+	}
+
+	// Refresh Indices
+	s.RefreshIndices(mappings.Actions.Index + "-*")
```
phiggins: In the PR I merged the other day, I moved some of these index strings into reusable variables and the refresh actions on them into helper methods, because I was copying and pasting them repeatedly. This one is referred to at least twice in this file so it might deserve the same treatment?

components/ingest-service/integration_test/project_update_actions_test.go
lancewf: Yeah, you are right. I moved the `DeleteAllDocuments` to right after the `suite.IngestActions` is called.

PR #135, opened 2019-04-23
[A2-771] AuthorizedProjects handles deny

bcmdarroch:
### :nut_and_bolt: Description
This PR refines the rego behind ProjectsAuthorized to ensure a deny overrides an allow on the same project, even with statements containing `All-Projects`

### :+1: Definition of Done

- [x] when request includes projects and statement denies all projects, ProjectsAuthorized returns no projects
- [x] when request includes no projects and statement denies all projects, ProjectsAuthorized returns no projects
- [x] when request includes no projects and one statement allows all projects and another statement denies one project, authz-service returns a list of all projects minus the denied projects

### :athletic_shoe: Demo Script / Repro Steps

1. rebuild services  and make some projects, teams, and roles
```
rebuild components/authz-service

export TOK=`chef-automate iam token create ADMIN --admin`
curl -kH "api-token: $TOK" -d '{"name":"project 1", "id": "p1"}' https://a2-dev.test/apis/iam/v2beta/projects | jq
curl -kH "api-token: $TOK" -d '{"name":"project 2", "id": "p2"}' https://a2-dev.test/apis/iam/v2beta/projects | jq

curl -kH "api-token: $TOK" -d '{"name":"role 1", "id": "r1", "actions": ["do:the:thing"], "projects": ["p1"]}' https://a2-dev.test/apis/iam/v2beta/roles | jq
curl -kH "api-token: $TOK" -d '{"name":"role 2", "id": "r2", "actions": ["do:the:thing"], "projects": ["p2"]}' https://a2-dev.test/apis/iam/v2beta/roles | jq

curl -kH "api-token: $TOK" -d '{"name":"team 2", "id": "t2", "projects": ["p2"]}' https://a2-dev.test/apis/iam/v2beta/teams | jq
```

2. upgrade to 2.1 and make a test token
```
chef-automate iam upgrade-to-v2 --beta2.1
export TEST=`chef-automate iam token create test`
```

3. make a policy for that token
```
# save as policy.json
{
  "name": "testing deny",
  "id": "test",
  "members": [ "token:test" ],
  "statements": [
    {
      "effect": "DENY",
      "actions": [ "iam:teams:list" ],
      "projects": [ "*" ]
    },
   {
      "effect": "ALLOW",
      "actions": [ "iam:teams:list" ],
      "projects": [ "p2" ]
    },
  {
      "effect": "ALLOW",
      "actions": [ "iam:roles:list" ],
      "projects": [ "*" ]
    },
   {
      "effect": "DENY",
[...]

PR #213, opened 2019-04-30
[A2-771] follow-up ProjectsAuthorized fix

bcmdarroch:
follow-up to https://github.com/chef/automate/pull/135 

### :+1: Definition of Done
- [x] when all projects are allowed and the request contains no project headers, authz-service passes on `*` to the domain services

```
[16][default:/src:0]# chef-automate iam upgrade-to-v2 --beta2.1

Enabling IAM v2.1
Migrating v1 policies...
Creating default teams Editors and Viewers...

Migrating existing teams...

Success: Enabled IAM v2.1

[22][default:/src:0]# curl -H "api-token: $TOK"  -d '{"id": "p1", "name": "project 1"}' https://localhost/apis/iam/v2beta/projects --insecure | jq
{
  "project": {
    "name": "project 1",
    "id": "p1",
    "type": "CUSTOM",
    "projects": [
      "p1"
    ]
  }
}

[18][default:/src:0]# curl -H "api-token: $TOK"  -d '{"id": "team1", "name": "team 1", "projects": ["p1"]}' https://localhost/apis/iam/v2beta/teams --insecure | jq
{
  "team": {
    "id": "team1",
    "name": "team 1",
    "projects": [
      "p1"
    ]
  }
}

[23][default:/src:0]# curl -H "api-token: $TOK"  https://localhost/apis/iam/v2beta/teams --insecure | jq
{
  "teams": [
    {
      "id": "admins",
      "name": "Members of the admins team, by default, have access to all parts of the API.",
      "projects": []
    },
    {
      "id": "editors",
      "name": "Editors",
      "projects": []
    },
    {
      "id": "viewers",
      "name": "Viewers",
      "projects": []
    },
    {
      "id": "team1",
      "name": "team 1",
      "projects": [
        "p1"
      ]
    }
  ]
}

[26][default:/src:0]# sl
--> Tailing the Habitat Supervisor's output (use 'Ctrl+c' to stop)
...
[2019-04-30T20:33:48+00:00] authz-service.default(O): time="2019-04-30T20:33:48Z" level=info msg="Projects Authorized Query" action="iam:teams:list" projects="[]" resource="iam:teams" result="[*]" subject="[token:e0a18396-7960-4a12-a185-3ef02f6df5f3]"
```
(punch line is `result=[*]` in the log)

- [ ] Code actually executed?

### DIFF
--- a/components/authz-service/engine/conformance/conformance_v2_test.go
+++ b/components/authz-service/engine/conformance/conformance_v2_test.go
@@ -83,7 +83,7 @@ func TestV2IsAuthorized(t *testing.T) {
 func TestV2ProjectsAuthorized(t *testing.T) {
 	ctx, engines := setup(t)
 	sub, act, res := "user:local:admin", "iam:users:create", "iam:users"
-	proj1, proj2, proj3, proj4, unassigned := "proj-1", "proj-2", "proj-3", "proj-4", "(unassigned)"
+	proj1, proj2, proj3, proj4, unassigned := "proj-1", "proj-2", "proj-3", "proj-4", constants.UnassignedProjectID
 	allProjects := []string{proj1, proj2, proj3, proj4, unassigned}
 
 	// We're always passing the same arguments to ProjectsAuthorized(). This allows for
@@ -359,7 +359,6 @@ func TestV2ProjectsAuthorized(t *testing.T) {
 			})
 
 			t.Run("policy that allows all projects returns all when all projects requested", func(t *testing.T) {
-				proj3, proj4, unassigned := "proj-3", "proj-4", "(unassigned)"
 				pol := map[string]interface{}{
 					"members": engine.Subject(sub),
 					"statements": map[string]interface{}{
--- a/components/authz-service/engine/conformance/conformance_v2_test.go
+++ b/components/authz-service/engine/conformance/conformance_v2_test.go
@@ -82,16 +82,14 @@ func TestV2IsAuthorized(t *testing.T) {
 
 func TestV2ProjectsAuthorized(t *testing.T) {
 	ctx, engines := setup(t)
-	sub, act, res, proj1, proj2 := "user:local:admin", "iam:users:create", "iam:users", "proj-1", "proj-2"
+	sub, act, res := "user:local:admin", "iam:users:create", "iam:users"
+	proj1, proj2, proj3, proj4, unassigned := "proj-1", "proj-2", "proj-3", "proj-4", "(unassigned)"
+	allProjects := []string{proj1, proj2, proj3, proj4, unassigned}
 
 	// We're always passing the same arguments to ProjectsAuthorized(). This allows for
 	// not having to wrap sub/act/res in _every call_; instead, we pass args().
 	args := func(projects []string) (context.Context, engine.Subjects, engine.Action, engine.Resource, engine.Projects) {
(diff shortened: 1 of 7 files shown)


components/authz-service/server/v2/authz.go
```diff
@@ -89,12 +89,16 @@ func (s *authzServer) ProjectsAuthorized(
 	}
 
 	requestedProjects := req.ProjectsFilter
+
+	allProjectsRequested := len(req.ProjectsFilter) == 0
```
tylercloke: Nice cleanup

components/authz-service/server/v2/system.go
```diff
@@ -171,6 +171,8 @@ func SystemPolicies() []*storage.Policy {
 					"iam:roles:" + constants.EditorRoleID,
 					"iam:roles:" + constants.ViewerRoleID,
 					"iam:roles:" + constants.IngestRoleID,
+					"iam:roles:" + constants.ProjectAdminRoleID,
+					"iam:roles:" + constants.IAMMembersViewerRoleID,
```
msorens: Good catch... but unrelated to this PR, is that correct...?

components/authz-service/server/v2/system.go
```diff
@@ -185,7 +185,6 @@ func SystemPolicies() []*storage.Policy {
 				},
 				Projects: []string{constants.AllProjectsID},
 			},
-			// TODO can't delete admin team from admin policy
```
msorens: By removing this comment you are claiming that it is now true that you cannot delete the admin team, but is that true now...?

components/authz-service/engine/conformance/conformance_v2_test.go
```diff
@@ -82,18 +82,16 @@ func TestV2IsAuthorized(t *testing.T) {
 
 func TestV2ProjectsAuthorized(t *testing.T) {
 	ctx, engines := setup(t)
-	sub, act, res, proj1, proj2 := "user:local:admin", "iam:users:create", "iam:users", "proj-1", "proj-2"
+	sub, act, res := "user:local:admin", "iam:users:create", "iam:users"
+	proj1, proj2, proj3, proj4, unassigned := "proj-1", "proj-2", "proj-3", "proj-4", "(unassigned)"
```
msorens: Please use constant so the definition is only given once per service:
s/"(unassigned)"/constants.UnassignedProjectID/

components/authz-service/engine/conformance/conformance_v2_test.go
```diff
@@ -362,7 +359,7 @@ func TestV2ProjectsAuthorized(t *testing.T) {
 			})
 
 			t.Run("policy that allows all projects returns all when all projects requested", func(t *testing.T) {
-				proj3, proj4 := "proj-3", "proj-4"
+				proj3, proj4, unassigned := "proj-3", "proj-4", "(unassigned)"
```
msorens: This line is a phantom; delete it please.

components/authz-service/server/v2/system.go
blakestier: We can't manage deletion of specific members of a policy with a policy (only general membership).  We'd need to modify the API. We could make a card for that if we think it's worth it.

components/authz-service/engine/conformance/conformance_v2_test.go
blakestier: nice catch!

PR #214, opened 2019-04-30
Action ingestion project tagging

lancewf:

Adds RBAC project tagging to the Chef Action ingestion.

1. `build components/ingest-service`
1. `start_all_services`
1. `send_chef_action_example`
1. `curl $ELASTICSEARCH_URL/actions-*/_search | jq . | less`
1. Ensure 'some project name' is in the project array

issue - https://github.com/chef/automate/issues/62
similar PR but for CCRs - https://github.com/chef/a2/pull/5236

- [x] Code actually executed?

### DIFF
--- a/components/ingest-service/pipeline/processor/action_project_tagger_test.go
+++ b/components/ingest-service/pipeline/processor/action_project_tagger_test.go
@@ -1,0 +1,416 @@
+package processor
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+
+	chef "github.com/chef/automate/api/external/ingest/request"
+	iam_v2 "github.com/chef/automate/api/interservice/authz/v2"
+	"github.com/chef/automate/components/ingest-service/pipeline/message"
+)
+
+// All the conditions must be true for a rule to be true (ANDed together).
+// Only one rule has to be true for the project to match (ORed together).
+// If there are no rules the project does not match
+// If the rule does not have any conditions it does not match any resources.
+func TestActionProjectRulesMatching(t *testing.T) {
+	cases := []struct {
+		description string
+		action      *chef.Action
+		rules       []*iam_v2.ProjectRule
+		matching    bool
+	}{
+		{
+			description: "A project with no rules does not match any resources",
+			matching:    false,
+			action:      &chef.Action{},
+			rules:       []*iam_v2.ProjectRule{},
+		},
+		{
+			description: "A project with one rule that does not have any conditions does not match any resources",
+			matching:    false,
+			action:      &chef.Action{},
+			rules: []*iam_v2.ProjectRule{
+				{
+					Type: iam_v2.ProjectRuleTypes_EVENT,
+				},
+			},
+		},
+		{
+			description: "Single rule single condition matching incorrect rule type",
+			matching:    false,
+			action: &chef.Action{
+				RemoteHostname:   "chef-server.org",
+				OrganizationName: "org1",
+			},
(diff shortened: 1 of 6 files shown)


components/ingest-service/pipeline/processor/action_project_tagger_test.go
```diff
@@ -0,0 +1,416 @@
...
+
+	<-out
+
+	assert.Equal(t, 1, listProjectRulesCount)
+}
+
+// When 5 messages are in the inbox the ListProjectRules function is only called once.
+func TestActionBundler5Messages(t *testing.T) {
+	inbox := make(chan message.ChefAction, 100)
+	listProjectRulesCount := 0
```
phiggins: I'm not familiar with gomock at all, but from using other mocking libraries I think this is what the `Times` function is for: https://godoc.org/github.com/golang/mock/gomock#Call.Times

components/ingest-service/pipeline/processor/action_project_tagger_test.go
```diff
@@ -0,0 +1,416 @@
...
+	inbox <- action2
+	close(inbox)
+
+	out := BuildActionProjectTagger(authzClient)(inbox)
+
+	processMsg1 := <-out
+	assert.Equal(t, []string{}, processMsg1.InternalChefAction.Projects)
+
+	processMsg2 := <-out
+	assert.Equal(t, []string{}, processMsg2.InternalChefAction.Projects)
```
phiggins: I'm finding these tests a bit hard to understand, and I wonder if the tests wouldn't need to do as much if the logic for how to treat the messages and rules was separated from the channel handling /message passing code.

components/ingest-service/pipeline/processor/action_project_tagger_test.go
lancewf: Thanks. I never thought of doing that. 
https://github.com/chef/automate/pull/214/commits/c1537a72e4990bfe1d952b044706ff0708d596c2

components/ingest-service/pipeline/processor/action_project_tagger_test.go
lancewf: 🤔 I will look into this.

lancewf: I think this last test (TestActionBundlerMatchProjectRuleNodeRuleType) is not needed, so I removed it. This is being tested here https://github.com/chef/automate/pull/214/files/ca2de00b00ff7b701f275efcf0cbcfea39652b8c#diff-7e0662f4628cc26ca7731861f082208fR63.

components/ingest-service/pipeline/processor/action_project_tagger_test.go
```diff
@@ -0,0 +1,410 @@
...
+	<-out
+	<-out
+	<-out
+	<-out
+	<-out
+}
+
+// A simple run through of the bundle project tagger processor.
+// Two messages are sent through with only one matching a project.
+func TestActionBundlerMatchProjectRule(t *testing.T) {
```
lancewf: This has mostly been tested with this test here https://github.com/chef/automate/pull/214/files#diff-7e0662f4628cc26ca7731861f082208fR20, but I wanted to do a whole run through of the entire processor.

components/ingest-service/pipeline/processor/action_project_tagger_test.go
lancewf: I went through and added more comments on what is going on in these tests. I do not know a way to break apart the channel handling from the bundler algorithm. This algorithm was designed to use the length of the channel to figure out how many messages to bundle together. Then the algorithm counts down each message as it goes through.

PR #215, opened 2019-04-30
[A2-690] Add site and channel data to the instance card

kmacgugan:
When site or channel is empty default to --
Added test helpers to default site and channel for existing tests
Added psql migrations to add site and channel columns

image

- start all services with ui pre-reqs completed
- rebuild components/applications-services
- source .studiorc to get data generator additions with site and channel
- run applications_populate_database
- open applications tab and click through service groups to see sidebar. You should see a mix of site and channel data filled out and when empty "--".

- [ ] Code actually executed?


.studio/applications-service
```diff
@@ -161,61 +161,61 @@ function applications_populate_database() {
   for i in $(seq 1 12); do
     applications_publish_supervisor_message --name nginx --group "kansas-prod"\
       --origin custom --version "1.0.1" --release "20190115184823" --health 2\
-      --sup-id "$(uuidgen)"
+      --sup-id "$(uuidgen)" --site "testsite" --channel "stable"
```
afiune: All these options are the same one for all our canned data inside this function. Can we make them a local function variable for easy editing later on?

Another idea is to not add these fields, site and channel, and instead put a default in the `applications-publisher` binary. (as I did for the `fqdn`)

api/external/applications/applications.proto
```diff
@@ -173,6 +175,9 @@ message HabService {
 
   // @afiune We need to figure out how the fqdn will be provided
   string fqdn = 9;
+  string channel = 11;
```
afiune: nit: Should this be the next available number? `10`?

api/external/applications/applications.proto
```diff
@@ -173,6 +175,9 @@ message HabService {
...
+  // site may not be provided by habitat so let's not use a reserved number for it.
```
afiune: Nice! Well thought. ⭐️

components/applications-service/cmd/applications-publisher/main.go
```diff
@@ -94,6 +94,8 @@ func main() {
 	flag.StringVar(&event.PkgIdent.Name, "name", "redis", "The name of a package")
 	flag.StringVar(&event.PkgIdent.Version, "version", "0.1.0", "The version of a package")
 	flag.StringVar(&event.PkgIdent.Release, "release", t.Format("20060102150405"), "The release of a package")
+	flag.StringVar(&event.Site, "site", "", "The site of the server where the service is running")
+	flag.StringVar(&event.Channel, "channel", "", "The channel that the supervisor is subscribed to")
```
afiune: maybe do the default channel to be `stable`?

components/applications-service/integration_test/global_test.go
```diff
@@ -24,6 +24,13 @@ var (
 // NewHabServiceMsg returns a new hab service message with the provided args
 func NewHabServiceMsg(supID, app, env, group,
 	origin, name, version, release, health string) *applications.HabService {
+	return NewHabServiceChannelMsg(supID, app, env, group,
+		origin, name, version, release, health, "stable", "test")
+}
+
+// NewHabServiceMsg returns a new hab service message with the provided args
```
afiune: Update the comment? 😅

components/applications-service/integration_test/global_test.go
```diff
@@ -24,6 +24,13 @@ var (
 // NewHabServiceMsg returns a new hab service message with the provided args
 func NewHabServiceMsg(supID, app, env, group,
 	origin, name, version, release, health string) *applications.HabService {
+	return NewHabServiceChannelMsg(supID, app, env, group,
```
afiune: I feel that this wrapper is going to make us follow a pattern we don't want to follow, that is creating multiple constructor functions `NewABC()`. IMO it would be worth changing this function to accept optional parameters and if any parameter is not provided, then take the defaults.

Something like:
```go
func NewHabServiceMsg(params map[string]string) *applications.HabService {
    // New message with defaults
    habServiceMsg := DefaultHabService()

    for key, value := range params {
        // Override here the provided parameters into 'habServiceMsg'
    }

    return habServiceMsg
}
```

components/applications-service/integration_test/services_by_sg_test.go
```diff
@@ -90,6 +90,7 @@ func TestGetServicesBySGSingleService(t *testing.T) {
 						Status:       applications.ServiceStatus_RUNNING,
 						HealthCheck:  applications.HealthStatus_OK,
 						Application:  a, Environment: e, Fqdn: "",
+						Channel: "stable", Site: "test",
```
afiune: Should we make all of these channel and site variables global? (since they are all the same) Just like `a` and `e` for application and environment. (Not really the best variable names though haha 😂 )

components/applications-service/pkg/storage/postgres/service.go
```diff
@@ -30,7 +30,9 @@ SELECT s.id
   , d.app_name AS application
   , d.environment AS environment
   , sup.member_id AS sup_member_id
-  , sup.fqdn AS fqdn
+	, sup.fqdn AS fqdn
```
afiune: fix indentation (tab for spacing)

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
```diff
@@ -41,9 +41,13 @@
         <div class="header">
           <div class ="service-line"><chef-icon>storage</chef-icon><b>{{ service.fqdn }}</b></div>
           <div class ="supervisor-line"><b>Supervisor:</b> {{ service.supervisor_id }}</div>
+          <div class ="supervisor-line" *ngIf="service.channel"><b>Site:</b> {{ service.site }}</div>
+          <div class ="supervisor-line" *ngIf="service.channel == ''"><b>Site:</b> --</div>
         </div>
         <div class="body">
           <div class ="service-line"><chef-icon>grain</chef-icon>{{ service.release }}</div>
+          <div class ="service-line" *ngIf="service.channel"><chef-pill>{{ service.channel }}</chef-pill></div>
+          <div class ="channel-line" *ngIf="service.channel == ''"><chef-pill> - - </chef-pill></div>
```
afiune: I can see we are always displaying the `Site` and the `Channel`, even when they have no values from the backend. I would recommend writing a function to do this logic in the `.ts` file and not in the HTML with `*ngIf`. That way you could even write a simple test to ensure they return the expected value. 😄 
(I say that because one returns `--`and the other one `- -`)

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.scss
```diff
@@ -50,8 +50,17 @@ app-page-picker {
     margin-bottom: .6em;
   }
 
+  .channel-line {
```
afiune: This new style is exactly as `service-line`, maybe renaming it to `data-line` and using it multiple times would be my advice.

danielsdeleo:
> I agree that the channel is data that is owned by the service but I feel that the site belongs to the supervisor table since it would be a tag added to supervisors and not services.

🤔 this is an interesting one. What you wrote is definitely how the system _should_ be used. Contrarily, in the initial implementation, habitat will only implement tagging at the supervisor level but we will need tagging at the service level to support hab-managed-chef style use-cases so we should expect that we'll have that capability eventually. The implementation in habitat that we've designed is that hab will just pass through arbitrary tags given by the user without knowing what they mean, so a user could potentially override site on a per-service basis. I don't know why they _would_, but it wouldn't be detected by habitat if they did it.

.studio/applications-service
kmacgugan: they actually do vary somewhat throughout the generation. Not by a lot though.  I wanted to have some be blank as well for ui testing. Which is why I didn't default.

components/applications-service/cmd/applications-publisher/main.go
kmacgugan: I want to be able to test a blank channel in the ui.

components/applications-service/integration_test/global_test.go
kmacgugan: I feel like this obfuscates the parameters and makes it harder to use the method though.

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
kmacgugan: I really only added the space because it looks better but maybe that is something we can fix with css?

apriofrost:
As discussed with @kmacgugan on Slack. There is CSS updates need to match the channel UI component with the latest mockup (sorry for the late mockup dates, the screenshot in the Jira has become outdated).

It currently uses the existing `chef-pill`, which is fine for now. We (the UX team) suggest adding a new UI component to display the information that is more suitable when put inside a table row or card when space is limited. It looks like this:
image

Maybe there will be help need from @Shadae or @scottopherson  for CSS. Here is the Inspect link to view the spec for the component: https://chef.invisionapp.com/d/main#/console/16786504/348323823/inspect

I haven't done a complete document on what the variations can be, but here is a peek into them (the numbers shows when semantic colors indicating success/warning/critical are applied to it) :
image

kmacgugan:
We made a new card for the css changes. A2-801.

PR #219, opened 2019-05-01
Update compliance-service to latest golangci-lint

phiggins:
Followup to #110 and #211.


components/compliance-service/config/types.go
```diff
@@ -71,7 +71,6 @@ type Manager struct {
 
 // InspecAgent service specific config options
 type InspecAgent struct {
-	Service
```
phiggins: 💭 I'm not sure what the correct solution is here. The problem is that the Compliance stuct embedded InspecAgent, and they both embedded Service which caused a conflict, generating a lint error.

components/compliance-service/reporting/relaxting/migrations.go
```diff
@@ -554,7 +554,7 @@ func migrateTimeSeriesDate(ctx context.Context, esClient *elastic.Client, dateTo
 				reportIds[i] = hit.Id
 			}
 
-			esInSpecReports, _ := getReportsA2v2(esClient, ctx, srcRepIndex, srcRepType, reportIds)
+			esInSpecReports, err := getReportsA2v2(esClient, ctx, srcRepIndex, srcRepType, reportIds)
```
phiggins: This one makes me nervous, if this function returns an error suddenly this entire function will error out when it wasn't before. Thoughts?

components/compliance-service/reporting/relaxting/profiles.go
```diff
@@ -146,12 +146,12 @@ func (esprofile *ESInspecProfile) parseInspecProfile(profile inspec.Profile) err
 		}
 		var refs string
 		byteRefs, _ := json.Marshal(control.Refs)
-		json.Unmarshal(byteRefs, refs) // nolint: errcheck
+		json.Unmarshal(byteRefs, &refs) // nolint: errcheck
```
stevendanna: All these nolint: errcheck's make me nervous but it is existing code so I think we should leave it.

components/compliance-service/api/profiles/conversion/conversion.go
```diff
@@ -47,7 +47,7 @@ func ConvertToPSProfile(profile inspec.Profile, namespace string) (profiles.Prof
 	if err != nil {
 		return convertedProfile, err
 	}
-	err = json.Unmarshal(jsonDepends, convertedDepends)
+	err = json.Unmarshal(jsonDepends, &convertedDepends)
```
stevendanna: I guess these were kinda, sorta working before because convertedDepends were slices

components/compliance-service/reporting/relaxting/migrations.go
stevendanna: The function returns `nil` for esInSpecReports for any non-nil error as far as I can tell.  So, provided we typically do have hits, then an error from here would likely already error out the function at line 595 if I'm reading this correctly, so I think we are probably OK here.

components/compliance-service/config/types.go
stevendanna: If it all compiles then we must not be referencing any fields provided by service directly on any InspecAgent structs so this seems reasonable to me.

components/compliance-service/reporting/relaxting/migrations.go
```diff
@@ -391,9 +391,9 @@ func (backend ES2Backend) reindex(src, dest, reindexScript, srcDocType string) (
 		Index(dest).
 		Type(mappings.DocType)
 	//source, err = reindexDestination.Source()
-	if err != nil {
-		return nil, indexToMigrateExists, errors.Wrap(err, fmt.Sprintf("%s unable to get reindexDest Source", myName))
-	}
+	//if err != nil {
+	//	return nil, indexToMigrateExists, errors.Wrap(err, fmt.Sprintf("%s unable to get reindexDest Source", myName))
+	//}
```
lancewf: Is there a reason we are not deleting this code?

components/compliance-service/reporting/relaxting/profiles.go
phiggins: I ran into trouble trying to sort this out before: https://github.com/chef/a2/pull/3158

components/compliance-service/reporting/relaxting/migrations.go
phiggins: There is other commented out code there so I was following that convention, but I think I'll just delete it instead. 👍

Issue #68, opened 2019-04-14
enable project filtering for all api calls in cfgmgmt.proto


vjeffrey:
# User Story
in order to return appropriately filtered data to users, each api call exposed in 
**`cfgmgmt.proto`**
must 
- make a function call on the incoming ctx to find out which projects they should filter on
- return the data that matches the project filters

An example of the function to be called on the ctx is:
```
func filterByProjects(ctx context.Context, filters map[string][]string) (map[string][]string, error) {
	projectsFilter, err := auth_context.ProjectsFromIncomingContext(ctx)
	if err != nil {
		return nil, err
	}
	if auth_context.AllProjectsRequested(projectsFilter) {
		return filters, nil
	}

	filters["projects"] = projectsFilter
	return filters, nil
}
```

An example of adding the project filter to the elasticsearch filters is:
```
if len(filters["projects"]) > 0 {
		termQuery := elastic.NewTermsQuery("projects", stringArrayToInterfaceArray(filters["projects"])...)
		query = query.Filter(termQuery)
	}
```

# Definition of Done
all apis in the defined proto make a function call on the incoming ctx to find out which projects they should filter by
all apis in the defined proto accept a []string projects filter and return the correct data based on those project filters
[ "project6", "project9" ] -> return all data tagged with project 6 or project 9 or both
[ "(unassigned)" ] -> return all data tagged with no projects tagged
[ "(unassigned)", "project8" ] -> return all data tagged with no projects tagged and return all data tagged with project8
["*" ] -> return all data

# Demo Script / Repro Steps
set up a project with ingest rules
send in a client run report, action, and compliance report that match the rules of the created project
make a query for resources in created project
expect to see the resources
make a query for resources in a different project
expect to not see the resources

lancewf:
- [ ] GetNodes
- [ ] GetRuns
- [ ] GetNodesCounts
- [ ] GetRunsCounts
- [ ] GetNodeRun
- [ ] GetSuggestions
- [ ] GetOrganizations
- [ ] GetSourceFqdns
- [ ] GetAttributes
- [ ] GetVersion
- [ ] GetPolicyCookbooks

PR #222, opened 2019-05-01
Project filtering for GetNodes

lancewf:

Add project filtering for the cfgmgmt.proto `GetNode` RPC. 

I am not sure I can do a demo for this 🤔, but there are a lot of tests ❓ ❗️ 😃 

This is mostly copying off the work that @phiggins is doing in PRs like these https://github.com/chef/automate/pull/249

- [x] Code actually executed?


components/config-mgmt-service/integration_test/nodes_test.go
```diff
@@ -139,6 +140,373 @@ func TestNodesWithSorting(t *testing.T) {
...
+					Exists:   true,
+				},
+			},
+			ctx: contextWithProjects([]string{authzConstants.UnassignedProjectID}),
+			request: request.Nodes{
+				Filter: []string{"status:success"},
+			},
+			expected: []string{"2"},
+		},
+	}
```
phiggins: There's some repetition in these test cases that might make sense to factor out, for example there's a contextWithProjects call in each one, and a lot of redundant info in the nodes; I'm pretty unfamiliar with this stuff and I had to read a few of the test cases to realize that the name and projects were the only significant fields.

lancewf: I will pull more of the data out of the cases and place it in the run section.

lancewf: @phiggins This commit cleaned up the tests https://github.com/chef/automate/pull/222/commits/103dee64149ee2432de04ae87327e9cd677fbf47

PR #223, opened 2019-05-01
[pipeline] Enable EAS Dashboard in dev & acceptance envs

afiune:
As an EAS Team Member,I would like to have proper development & acceptance environments that deploy automatically on every merge of a PR,So I can do proper "Acceptance Demos" and verify that the functionality we are shipping is the correct one.

We currently don't have a place where we as a Team can demo our work that is being shipped to our customers, that is our Acceptance Environment. As well as a place where we can do Demos to the entire organization and a place where anyone outside our team can go play with the work we are doing internally within the EAS team.

We also don't have a place where we are testing upgrades, so any modification to our database schema is not currently being tested.

This card is to *create such environment(s)* and start doing acceptance of all the work we are doing in the EAS Team.

### :+1: Definition of Done
Enable the Apps Beta Feature in one of our existing instances for our Development and Acceptance environments.

After this PR is merged we will have EAS Dashboard enabled in the following instances:
- a2-perf-test-single-local-inplace-upgrade-dev.cd.chef.co
- a2-perf-test-single-local-inplace-upgrade-acceptance.cd.chef.co

Merge and check the pipeline. (aka YOLO!) 😆 🚀 

https://chefio.atlassian.net/browse/A2-794

- [ ] Code actually executed?

### DIFF
--- a/terraform/test-environments/modules/chef_automate_install/main.tf
+++ b/terraform/test-environments/modules/chef_automate_install/main.tf
@@ -209,6 +209,17 @@ EOF
   [gateway.v1.sys]
     [gateway.v1.sys.service]
       trial_license_url = "https://licensing-${var.channel}.chef.io/create-trial"
+      enable_nats_feature = ${var.enable_eas_dashboard}
+
+[event_service.v1]
+  [event_service.v1.sys]
+    [event_service.v1.sys.service]
+      enable_nats_feature = ${var.enable_eas_dashboard}
+
+[applications.v1]
+  [applications.v1.sys]
+    [applications.v1.sys.service]
+      enable_nats_feature = ${var.enable_eas_dashboard}
 
 [load_balancer.v1]
   [load_balancer.v1.sys]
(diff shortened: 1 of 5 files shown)


jeremiahsnapp:
This PR looks great to me but does EAS require chef server and workflow? If it doesn't then what about just applying it to the performance instance where we have chef-load already applied?

https://github.com/chef/automate/blob/master/terraform/test-environments/scenarios/performance_test_single_local_inplace_upgrade.tf

afiune:
@jeremiahsnapp it does not need chef-server or workflow, I enable them into those instances because we talked over Slack and those are the once you suggested 😅 -- Should I use the other one that doesn't have workflow or chef server?

jeremiahsnapp:
Sorry about the confusion there @afiune. I misunderstood what your requirements were. It sounds like applying these changes to the performance testing instance is the better choice.

afiune:
Will do that Jeremiah! It’s all good. Thanks for your assistance! 💯

afiune:
@danielsdeleo I can load some data as we do in our studio but I would recommend waiting for the habitat integration to plug everything together. How does that sound?

danielsdeleo:
> I can load some data as we do in our studio but I would recommend waiting for the habitat integration to plug everything together. How does that sound?

Sounds good to me!

PR #225, opened 2019-05-01
nodes: refactor process node handling to be simpler

vjeffrey:
The logic in this file (node-managers-process-node) was hard to follow, so this is my attempt to simplify it.

I also modified the logic that reads the node from the db to use a pointer for target config so that we don't have to worry about always having a value there.

### :+1: Definition of Done
Tests still pass without modification and the code is easier to read.

### DIFF
--- a/components/nodemanager-service/pgdb/node-managers-process-node.go
+++ b/components/nodemanager-service/pgdb/node-managers-process-node.go
@@ -1,8 +1,6 @@
 package pgdb
 
 import (
-	"database/sql"
-	"encoding/json"
 	"time"
 
 	uuid "github.com/gofrs/uuid"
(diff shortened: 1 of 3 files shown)


components/nodemanager-service/pgdb/nodes.go
```diff
@@ -153,22 +153,22 @@ var nodesSortFields = map[string]string{
...
+	ManagerIDs      json.RawMessage  `db:"manager_ids"`
+	Name            string           `db:"name"`
+	Platform        string           `db:"platform"`
+	PlatformVersion string           `db:"platform_version"`
+	Projects        json.RawMessage  `db:"projects"`
+	ReportID        string           `db:"report_id"`
+	State           string           `db:"source_state"`
+	Status          string           `db:"status"`
+	Tags            json.RawMessage  `db:"tags"`
+	TargetConfig    *json.RawMessage `db:"target_config"`
```
phiggins: Could this change be avoided by checking for `if len(inNode.TargetConfig) > 0 {`?

I guess the deeper question is something like: do we need to differentiate between an empty TargetConfig vs nil?

vjeffrey: we do - because without doing the pointer there you end up with a nasty sql error about scanning a nil value. it's because we didnt give it a default value when we first set this all up.

components/nodemanager-service/pgdb/node-managers-process-node.go
```diff
@@ -55,98 +48,28 @@ func (db *DB) ProcessIncomingNode(node *manager.NodeMetadata) error {
...
-	return nil
-}
+	lastContact := ptypes.TimestampString(node.LastContact)
 
-func updateNode(nodeUuid string, node *manager.NodeMetadata, lastContact string, db *DB, nodeState string) (sql.Result, error) {
-	updated, err := db.Exec(sqlUpdateNodeLastContact, nodeState, lastContact, node.JobUuid, node.Name, node.PlatformName, node.PlatformRelease, nodeUuid)
-	if err != nil {
-		return nil, errors.Wrapf(err, "updateNode unable to update node %s", node.Uuid)
+	if len(node.GetUuid()) == 0 {
+		node.Uuid = uuid.Must(uuid.NewV4()).String()
```
lancewf: We might want to return an error if we can not create the UUID instead of exiting the service.

vjeffrey: ahhh ya i forgot about that darn uuid must functionality, good catch - ya!

PR #226, opened 2019-05-01
Trigger replication notification of authz policies

jaym:
This PR uses postgres' LISTEN/NOTIFY feature to notify all authz nodes
of changes to the policies. This makes the replication start almost
instantaneously, for sure faster that the 10 second anti entropy timer
we have right now.

### DIFF
--- a/components/authz-service/storage/v2/postgres/policy_change_notifier.go
+++ b/components/authz-service/storage/v2/postgres/policy_change_notifier.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"time"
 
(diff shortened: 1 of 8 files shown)


components/authz-service/storage/v2/postgres/policy_change_notifier.go
```diff
@@ -0,0 +1,77 @@
+package postgres
+
+import (
+	"context"
+	"time"
+
+	"github.com/chef/automate/components/authz-service/storage/v2"
```
srenatus: Nit import order

PR #227, opened 2019-05-01
[e2e] ensure we delete scan jobs after tests

sdelano:
in our long-lived test environments we're currently failing to delete
scan jobs for some reason. this change adds a test with the intent to
check that the scan job has been deleted before we proceed. this test
may end up being flaky, but we should fix the flakiness and ensure
that we're correctly deleting things in the test

Signed-off-by: Stephen Delano <stephen@chef.io>

- [x] Code actually executed?

### DIFF
--- a/e2e/cypress/integration/compliance/01_ssh_scan_job.spec.js
+++ b/e2e/cypress/integration/compliance/01_ssh_scan_job.spec.js
@@ -230,10 +230,20 @@ describe('create a manual node ssh scan job and cleanup after', () => {
       .click().then(() => {
         cy.get('chef-page-header').find('#uninstall-btn').click().then(() => {
           cy.get('chef-modal').find('chef-button').contains('Delete').click()
+        })
       })
+
+    // ensure that the profile is deleted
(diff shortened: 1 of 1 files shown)


sdelano:
@phiggins what's been happening is that, for some reason the scan job delete never finishes. I suspect that it might have something to do with it being the LAST thing we do in the test and the test just doesn't wait around before moving on to the next suite.

After ~100 tests, we start to fail because the scan job page becomes paginated and we can't reliably find the job we're testing. Ensuring we delete them every time will prevent that state, and will also test that delete works.

phiggins:
I get what is happening, what I was trying to say is that we might want to make it so that the deletion always happens regardless of whether that particular test gets run. This would guard against cases where that test failed before the deletion completed, if that test was ignored for some reason, etc.

sdelano:
Ah yes, of course. That's probably a better long-term solution. I just ran the tests again locally against the test environment without my changes and I was able to reproduce the case, so at least it seems like the fix does what is intended.

PR #228, opened 2019-05-01
[tools] Add jay's HTTP proxy

stevendanna:
Jay wrote this small HTTP proxy. We've been using it for very basic
testing as it is a bit nicer than starting up nginx in yet another
container.  Let's move it to the repo rather than having it in a gist.


tools/proxy/main.go
```diff
@@ -0,0 +1,61 @@
...
+		v := atomic.AddUint64(&counter, 1)
+		u := urls[int(v)%len(urls)]
+		req.URL.Scheme = u.Scheme
+		req.URL.Host = u.Host
+		logrus.Infof("%s %s upstream: %s", req.Method, req.URL.Path, req.URL.String())
+	}
+
+	return &httputil.ReverseProxy{
+		Director: director,
+		Transport: &http.Transport{
```
srenatus: No DefaultTransport copy here? Zero values are good enough?

tools/proxy/main.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		logrus.Fatalf("listen: %s", err.Error())
+	}
+}
+
+func newReverseProxy(urls []*url.URL) *httputil.ReverseProxy {
+	var counter uint64
+	director := func(req *http.Request) {
+		v := atomic.AddUint64(&counter, 1)
+		u := urls[int(v)%len(urls)]
```
srenatus: Round Robin? Heh.

PR #229, opened 2019-05-01
[deployment] Unpin automate-ha-backend packages.

yzl:
This at least gets us back to exercising the current automate-ha-backend packages.

For elasticsearch auth, we are no longer using nginx for elasticsearch auth, and we are generating certs for use with automate-backend-elasticsearch. Also, we are using a recent hab in the backend containers to get the nested config feature added in 0.75.

Signed-off-by: Yvonne Lam <yvonne@chef.io>

- [x] Code actually executed?

### DIFF
--- a/integration/services/ha_backend/init
+++ b/integration/services/ha_backend/init
@@ -2,14 +2,14 @@
 
 HA_BACKEND_DIR=$(dirname ${BASH_SOURCE[0]})
 HA_BACKEND_USER="admin"
-HA_BACKEND_PASSWORD="chefautomate"
+# changing to match the default in automate-ha-backend
+HA_BACKEND_PASSWORD="admin"
 
 ha_backend_container1=$(service_container_name "ha_backend_1")
 ha_backend_container2=$(service_container_name "ha_backend_2")
 
 ha_backend_private=$(service_config_path "ha_backend_private")
 ha_backend_config=$(service_config_path "ha_backend.toml")
-
 
 ha_backend_setup() {
     mkdir -p $ha_backend_private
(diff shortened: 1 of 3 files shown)


jaym:
you'll need to update habitat as well

integration/services/ha_backend/init
```diff
@@ -33,22 +34,16 @@ ha_backend_setup() {
         return 1
     fi
 
-    hab pkg exec core/openssl openssl req -x509 -nodes -days 3650 -newkey rsa:2048 \
```
jaym: Is it possible to generate the certificate and password and use that instead of relying on whatever the default is

integration/services/ha_backend/init
```diff
@@ -109,8 +126,14 @@ enable = true
 DOC
 
     sleep 45
-    curl -k "http://${ha_backend_container1_ip}:9200"
-    curl -k "http://${ha_backend_container2_ip}:9200"
+    curl -k "https://${ha_backend_container1_ip}:9200" \
+	    --cacert "${ha_backend_container1}/certificates/MyRootCA.pem" \
+	    --key "${ha_backend_container1}/certificates/odfe-admin.key" \
```
jaym: do we need `--key` and `--cert`? We expected this to change to use basic auth. Also, if you're passing in `-k`, you shouldn't need `--cacert`

integration/services/ha_backend/setup.sh
```diff
@@ -10,7 +10,7 @@ vm.dirty_ratio=20
 vm.dirty_background_ratio=30
 vm.dirty_expire_centisecs=30000
 net.ipv4.ip_local_port_range=1024 65024 net.ipv4.tcp_max_syn_backlog=60000
-net.ipv4.tcp_tw_reuse=1
+# net.ipv4.tcp_tw_reuse=1
```
jaym: why is this commented out? Should we remove it?

yzl: No, I will uncomment it for CI. I commented it out because docker on vagrant doesn't like it.

PR #230, opened 2019-05-01
Disable Create Project button if more than 6 projects

tylercloke:

Screen Shot 2019-05-01 at 4 12 10 PM

Screen Shot 2019-05-01 at 4 12 20 PM

- [ ] Code actually executed?

### DIFF
--- a/components/automate-ui/e2e/admin.e2e-spec.ts
+++ b/components/automate-ui/e2e/admin.e2e-spec.ts
@@ -1608,7 +1608,7 @@ describe('Admin pages', () => {
       ['project-3', true],
       ['project-4', true],
       ['project-5', true],
-      ['project-6', true],
+      ['project-6', true]
     ].forEach(([id, deletable]) => {
       const path = `/iam/v2beta/projects/${id}`;
       const endpoints = {
--- a/components/automate-ui/e2e/admin.e2e-spec.ts
+++ b/components/automate-ui/e2e/admin.e2e-spec.ts
@@ -1535,6 +1535,122 @@ describe('Admin pages', () => {
     });
   });
 
+  // TODO (tc): This whole block can go away once the 6 project limit is removed.
+  describe('projects list when the max of 6 projects is reached', () => {
+    beforeEach(() => {
+      fakeServer()
+      .post('/api/v0/auth/introspect_some', JSON.stringify(
+        {
+          paths: [
+            '/iam/v2beta/projects'
+          ]
+        }
+      ))
+      .many()
+      .reply(200, JSON.stringify(
+        {
+          endpoints: {
+            '/iam/v2beta/projects': {
+              get: true,
+              put: false,
+              post: true,
+              delete: false,
+              patch: false
+            }
+          }
+        }
+      ));
+
+    fakeServer()
+    .get('/apis/iam/v2beta/projects')
+    .many()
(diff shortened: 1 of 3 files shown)


components/automate-ui/src/app/pages/project/list/project-list.component.html
```diff
@@ -5,9 +5,12 @@
   <main>
     <chef-page-header>
       <chef-heading>Projects</chef-heading>
-      <chef-subheading>
+      <chef-subheading *ngIf="(sortedProjects$ | async)?.length < 6">
         Projects are named groups of resources used to define policies.
       </chef-subheading>
+      <chef-subheading *ngIf="(sortedProjects$ | async)?.length >= 6">
+        Projects group resources together. Max of 6 projects allowed.
```
msorens: Please, please, please put that magic number used throughout into a named constant. Even on this line.

components/automate-ui/e2e/admin.e2e-spec.ts
```diff
@@ -1533,6 +1533,43 @@ describe('Admin pages', () => {
         });
       });
     });
+
+    describe('when the max of 6 projects is reached', () => {
```
blakestier: 6 custom projects, or 6 total projects? If it's custom, can we include that in the code wherever possible?

blakestier: Or  . . . do we not display the chef-managed ones?

tylercloke: Great question! Lemme look.

tylercloke: Looks like they aren't included! Thanks for bringing that edge case up though :) 

https://github.com/chef/automate/blob/master/components/authz-service/server/v2/project.go#L202-L204

blakestier: oh, right. The chef-managed one is just internal

components/automate-ui/src/app/pages/project/list/project-list.component.html
```diff
@@ -17,7 +17,7 @@
       <ng-container *ngIf="(sortedProjects$ | async)?.length > 0">
         <chef-toolbar>
           <app-authorized [allOf]="['/iam/v2beta/projects', 'post']">
-            <chef-button id="create-button" primary (click)="openCreateModal()">Create Project</chef-button>
+            <chef-button [disabled]="(sortedProjects$ | async)?.length >= MAX_PROJECTS" id="create-button" primary (click)="openCreateModal()">Create Project</chef-button>
```
bcmdarroch: @susanev this is a UX nitpick, but I think it's hard to immediately tell why the create button is disabled. I know it's in the sub-heading but my eye tends to skip over that. Some kind of "no more projects allowed" message to the right of the button would be helpful. (though I know there's supposed to be an `Update` Button there soon so I get it you don't want to mess with that design)

susanev: I totally agree but unfortunately with ingest rules there are going to be other buttons to the right so we can't do that approach anymore. If we get user feedback on that from early customers we can find a better solution.

susanev: Also, this limit should not be in place for more than a few months... hopefully.

PR #231, opened 2019-05-02
[terraform] use correct config for nats settings

sdelano:
Signed-off-by: Stephen Delano <stephen@chef.io>

- [ ] Code actually executed?


sdelano:
@stevendanna haha I know :)

I typed it in like that because I hate the indentation, but then I decided against it because the rest of that file follows the unnecessary convention.

That indentation style we've co-opted in so many places is just an artifact of the way that the go libraries we use serialize things to TOML. Oh well....

PR #232, opened 2019-05-02
Adds missing feature flag to enable EAS Dashboard

afiune:
This is the missing feature flag: https://github.com/chef/automate/blob/master/api/config/gateway/config_request.proto#L38

I tested this time the change manually on the server to verify it works and after running the installation module from terraform it worked just fine, you can even visualize data in the Applications Tabs:

Screen Shot 2019-05-01 at 9 37 29 PM


afiune:
I will probably use my admin privileges to merge this change to fix the pipeline.

PR #233, opened 2019-05-02
Trigger replication notification of authz policies

jaym:
This PR uses postgres' LISTEN/NOTIFY feature to notify all authz nodes
to of changes to the policies. This makes the replication start almost
instantaneously, for sure faster that the 10 second anti entropy timer
we have right now.

### DIFF
--- a/components/authz-service/server/v2/policy_refresher.go
+++ b/components/authz-service/server/v2/policy_refresher.go
@@ -26,6 +26,7 @@ type policyRefresher struct {
 	engine                   engine.V2Writer
 	refreshRequests          chan policyRefresherMessageRefresh
 	antiEntropyTimerDuration time.Duration
+	changeNotifier           storage.PolicyChangeNotifier
 }
 
 type policyRefresherMessageRefresh struct {
@@ -48,16 +49,21 @@ func (m *policyRefresherMessageRefresh) Err() error {
 }
 
 func NewPolicyRefresher(ctx context.Context, log logger.Logger, store storage.Storage,
-	engine engine.V2Writer) PolicyRefresher {
+	engine engine.V2Writer) (PolicyRefresher, error) {
+	changeNotifier, err := store.GetPolicyChangeNotifier(ctx)
+	if err != nil {
+		return nil, err
+	}
 	refresher := &policyRefresher{
 		log:                      log,
 		store:                    store,
 		engine:                   engine,
 		refreshRequests:          make(chan policyRefresherMessageRefresh, 1),
 		antiEntropyTimerDuration: 10 * time.Second,
+		changeNotifier:           changeNotifier,
 	}
 	go refresher.run(ctx)
-	return refresher
+	return refresher, nil
 }
 
 func (refresher *policyRefresher) run(ctx context.Context) {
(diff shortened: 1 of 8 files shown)


components/authz-service/server/v2/policy_refresher.go
```diff
@@ -69,7 +75,18 @@ RUNLOOP:
...
 			break RUNLOOP
+		case <-refresher.changeNotifier.C():
+			refresher.log.Info("Received policy change notification")
+			var err error
+			lastPolicyID, err = refresher.refresh(context.Background(), lastPolicyID)
+			if err != nil {
+				refresher.log.WithError(err).Warn("Failed to refresh policies")
+			}
+			if !antiEntropyTimer.Stop() {
+				<-antiEntropyTimer.C
```
bcmdarroch: how is this timer used? it seems like the refresher is both listening for change notifications and also periodically refreshing based on the timer - is that right? are both needed?

jaym: the reason for both is you have to assume that notifications can be missed for various reasons. For example, you drop the connection and a notification comes in, you will miss it. Doing this is much easier than trying to guarantee delivery notifications and it's pretty cheap to do. The notifications just become an optimization that speeds up replication.

jaym:
in the HA world, your service will connect to a HA postgres (replicated). So while there will be multiple instances of postgres, they'll all have the same data. There will be multiple authz-service instances, all sharing that database. The OPA engine in each instance needs to stay consistent with whatever is in that database, which is what the 2 PRs i've submitted ensure.
What this changes about the policy update flow is that it is now eventually consistent. It's possible that a second node replicate faster or slower than the node that got the policy update request. It makes the update requests waiting for opa rules to refresh kind of meaningless, but I'm not sure that matters as I'm guessing those requests are generally going to time out for our largest customers anyway, and to them the changes will show up some time after that anyway.

components/authz-service/server/v2/policy_refresher.go
```diff
@@ -69,7 +75,18 @@ RUNLOOP:
 		case <-ctx.Done():
 			refresher.log.WithError(ctx.Err()).Info("Policy refresher exiting")
 			break RUNLOOP
+		case <-refresher.changeNotifier.C():
+			refresher.log.Info("Received policy change notification")
+			var err error
+			lastPolicyID, err = refresher.refresh(context.Background(), lastPolicyID)
+			if err != nil {
+				refresher.log.WithError(err).Warn("Failed to refresh policies")
```
blakestier: Do we have a convention around capitalizing logs? Lowercase errors, uppercase Info/Warn? Or are we all over the map?

components/authz-service/storage/v2/postgres/policy_change_notifier.go
```diff
@@ -0,0 +1,77 @@
...
+	maxReconnectInterval time.Duration
+	pingInterval         time.Duration
+	notificationChan     chan v2.PolicyChangeNotification
+	shutdown             func()
+}
+
+func newPolicyChangeNotifier(ctx context.Context, conninfo string) (v2.PolicyChangeNotifier, error) {
+	ctx, cancel := context.WithCancel(ctx)
+	p := &policyChangeNotifier{
+		conninfo:             conninfo,
```
blakestier: I think you camel case this elsewhere, fwiw

PR #239, opened 2019-05-02
Fix backup deadlock

jaym:
The backup runner was releasing the lock too early. It ended up
corrupting the backup status, as well as being able to deadlock backups
if they started one right after the other. The finishing backup with
close one of the channels the new backup created which caused the
deadlock

### DIFF
--- a/components/automate-deployment/pkg/backup/runner.go
+++ b/components/automate-deployment/pkg/backup/runner.go
@@ -479,13 +479,13 @@ func (r *Runner) startRestoreOperations(ctx context.Context) {
 		// Make sure that the event publisher always exits, we unlock the deployment,
 		// and that any listeners on the event stream will get the task complete
 		// signal.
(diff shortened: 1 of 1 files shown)


components/automate-deployment/pkg/backup/runner.go
```diff
@@ -479,13 +479,13 @@ func (r *Runner) startRestoreOperations(ctx context.Context) {
...
 		// and that any listeners on the event stream will get the task complete
 		// signal.
-		r.unlockDeployment()
 		if r.eventSender != nil {
 			r.eventSender.TaskComplete()
 		}
 		close(r.errChan)
 		close(r.eventTerm)
 		r.clearRunningTask()
+		r.unlockDeployment()
```
stevendanna: I'm worried about this being below the TaskComplete() because that can block

Issue #143, opened 2019-04-23
updates to user menu in top nav


vjeffrey:
original issue: https://chefio.atlassian.net/secure/RapidBoard.jspa?rapidView=261&modal=detail&selectedIssue=UI-158

# Overview

To unify the user menu with the incoming projects filter this task provides a number of visual changes and updates.

# Design Details

Zeplin (https://app.zeplin.io/project/5ba808703eb54372d860c737/dashboard?seid=5c4887fb20f12301729f6199)

All visual design and content are intentional if you have any questions or need more information please connect with susan evans.

## Dropdown button

- Remove name label from the top navigation bar
- Remove the circle behind the person icon
- Change the person icon to be 20x20 and #212234, it should be vertically centered in the top navigation bar
- Place an expand_more icon 8 pixels to the right of the person icon, its bottom most point should align with the text labels in the top navigation bar
- Together with the person icon and expand more icon should form a single button that opens a drop down

## Dropdown menu

- Add a caret to the top of the dropdown that aligns with the expand more icon
- Please reference the Zeplin designs for text style information, and padding/margins in the menu
- Signed in as and the user's name is not interactive
- Profile takes the user to their profile page
- Version number is not interactive
- About opens the about modal
- License

     should open 
     https://www.chef.io/online-master-agreement/
      in a new tab
            - Note: there is currently a modal and we want that to go away

- Release Notes opens the release notes in a new tab
- Sign Out signs the user out and sends them to the sign in page
- Hovering/focusing on any of the interactive items should result in a background color of #dee5fd

PR #240, opened 2019-05-02
Updates to user menu

magwalk:
Updates to the user menu in the nav. Also some minor tweaks to the projects filter dropdown for better matching with the user menu (so they both use the same markup conventions).

user-nav

https://github.com/chef/automate/issues/143

- [x] Code actually executed?


components/automate-ui/src/app/page-components/profile/profile.component.html
```diff
@@ -1,39 +1,46 @@
...
-    </li>
-    <li>
-      <button class="license-link" (click)="openLicenseModal()">License Information</button>
-    </li>
-    <li>
-      <a href="https://automate.chef.io/release-notes/?v={{ buildVersion }}" class="version-link" target="_blank">Chef Automate {{ buildVersion }}</a>
-    </li>
-    <li><chef-button primary (click)="logout()" data-cy="sign-out-button">Sign Out</chef-button></li>
-  </ul>
+      <app-authorized [allOf]="['/auth/users/{username}', 'get', userName]">
```
bcmdarroch: the updates here will need to reflected in our cypress tests (which are unfortunately not run in the PR `verify` pipeline yet so has to be tested locally)

Running the login spec locally, looks like this bit needs to be updated:
https://github.com/chef/automate/blob/master/e2e/cypress/integration/01_login.spec.js#L85-L92

magwalk: We should really add a link to this in the automate-ui readme in the testing section.

scottopherson:
Have [a commit here](https://github.com/chef/automate/commit/b7ddd8177f1dd5fd47297fa57fbbcc2ecfcfdaf1) that fixes issue mentioned by @bcmdarroch.



The `stopPropagation()` in the profile dropdown button handler prevents any of the `click-outside` components from doing their thing. Didn't see a reason to be preventing the event from propagating so just removed it; otherwise we'll need a different fix.

Everything else looks good 👍

components/automate-ui/src/app/page-components/profile/profile.component.scss
```diff
@@ -1,107 +1,96 @@
 @import "~styles/variables";
 @import "~styles/mixins";
 
-.profile,
+$dropdown-width: 196px;
+
 button {
   -webkit-appearance: none;
-  background: transparent;
```
scottopherson: Button seems to be gray in Firefox w/o this line. Not sure atm why its handling the background differently.

Screen Shot 2019-05-02 at 11 17 23 PM

msorens:
1. Once the dropdown is open, I can tab to move through the entries, but the up/down arrow keys should work, just as they do in the adjacent global project filter. (Same issue in Chrome and Firefox).
2. Tabbing through the dropdown works as expected in Chrome, but in Firefox it skips in a peculiar way: it does not land on Profile, License, or Release Notes.

components/automate-chef-io/content/docs/users.md
```diff
@@ -105,8 +105,8 @@ curl -X DELETE -H "api-token: $TOK" -H "Content-Type: application/json" https://
 ### User Self-Maintenance
 
 Local Automate users can manage their own name and password through the Chef Automate user interface.
-Select your full name in the top navigation bar,
-then select **Your Profile** from the drop-down.
+Click the user icon in the top navigation bar,
```
susanev: We've removed the word "Click" from the docs as it is centered on people who use pointing devices. Could we keep this as _Select_ here?

mjingle: Seconded.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
```diff
@@ -10,18 +10,20 @@
   </span>
   <chef-icon aria-hidden>keyboard_arrow_down</chef-icon>
 </button>
-<chef-dropdown class="dropdown" *ngIf="dropdownActive" (keydown.esc)="handleEscape()">
```
magwalk: Most of this is just changing order of the markup so the `chef-click-outside` element wraps around the `chef-dropdown` element (which is generally the convention we use for using the `chef-click-outside` element. 

Also removes an unnecessary argument on line 25 where `handleApplyClick()` had an unnecessary `$event` argument that was throwing errors in CI.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
```diff
@@ -53,8 +53,6 @@
 }
 
 .dropdown {
-  display: block;
```
magwalk: The `chef-dropdown` element ships with these styles when the `visible` attribute is being used properly.

components/automate-ui/src/app/page-components/profile/profile.component.ts
```diff
@@ -64,6 +67,38 @@ export class ProfileComponent implements OnInit, OnDestroy {
...
+  getCurrentIndex(currentTarget) {
+    let currentIndex;
+
+    this.focusElements.toArray().forEach((element, i) => {
+      if (element.nativeElement === currentTarget) {
+        currentIndex = i;
+      }
+    });
+
+    return currentIndex;
```
msorens: Consider a more functional approach to achieve the same result:
```
  getCurrentIndex(currentTarget: HTMLElement) {
    return this.focusElements.toArray()
      .map((e: ElementRef) => e.nativeElement)
      .indexOf(currentTarget);
  }
```

susanev:
Not sure if you have a follow-up commit for this, but it seems as though when selecting _About_ the menu does not close until after you close the modal. Could we close the menu right away, instead of after you close the modal?

Everything else looks great, thanks for your work on this!!

magwalk:
@msorens So for the firefox issue - I also had that same behavior when I tried it out on firefox but after doing some research I found out that I had the wrong settings on my Mac to enable tabbing through links on firefox. Are you able to tab to any links on firefox (i.e. the ones in the nav)?

magwalk:
@msorens Yeah it's super weird behavior. So! If you go to System Preferences > Keyboard > Shortcuts, then at the bottom of the window it says "Full Keyboard Access." Change the selection to "All Controls". That fixed it for me.

PR #241, opened 2019-05-02
[deployment] Remove AsyncOps from backup code

stevendanna:
The goal of this change is to reduce the surface area of the
backup/restore code we need to consider when making future changes to
this code.

* Why do we need to make future changes?

The backup code is *heavily* exercised by a number of our customers.

In the last few weeks we've fixed a number of 1-off deadlocks related
to the backup code. We would like to do some work to simplify the
concurrency control in his part of the code. Currently there are 2
locks and multiple channels in play. While the code removed in this PR
has not contributed to any of those bugs, removing it means we have
less to consider when making future changes.

* Why is this safe to remove?

We think this is safe to remove because no backup specs use the Async
version of operations.

* Won't this make it hard to speed up backups in the future?

We can always add it back.


stevendanna:
This is now out of date.  If someone is interested in moving this forward we can rebase it and continue.

PR #243, opened 2019-05-02
Project filtering for reporting server ReadNode

phiggins:
#66 

If you looked at this PR while it was WIP, the only change made since then was to remove some random code cleanups that I'll move to a new PR.

### DIFF
--- a/components/compliance-service/api/reporting/server/server.go
+++ b/components/compliance-service/api/reporting/server/server.go
@@ -285,7 +285,13 @@ func (srv *Server) ListNodes(ctx context.Context, in *reporting.Query) (*reporti
 
 // ReadNode returns a node based on id
 func (srv *Server) ReadNode(ctx context.Context, in *reporting.Id) (*reporting.Node, error) {
-	node, err := srv.es.GetNode(in.Id)
+	formattedFilters := formatFilters([]*reporting.ListFilter{})
+	formattedFilters, err := filterByProjects(ctx, formattedFilters)
+	if err != nil {
+		return nil, utils.FormatErrorMsg(err, in.Id)
+	}
+
+	node, err := srv.es.GetNode(in.Id, formattedFilters)
 	if err != nil {
 		return nil, utils.FormatErrorMsg(err, in.Id)
 	}
--- a/components/compliance-service/integration_test/reporting_server_read_node_test.go
+++ b/components/compliance-service/integration_test/reporting_server_read_node_test.go
@@ -12,13 +12,19 @@ import (
 	"github.com/chef/automate/components/compliance-service/reporting/relaxting"
 )
 
-func TestReadReport(t *testing.T) {
+func TestReadNode(t *testing.T) {
 	server := reportingServer.New(&relaxting.ES2Backend{ESUrl: elasticsearchUrl})
+
+	unassignedNodeId := newUUID()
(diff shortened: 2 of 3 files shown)


components/compliance-service/integration_test/reporting_server_read_node_test.go
```diff
@@ -0,0 +1,113 @@
...
+		t.Run(test.description, func(t *testing.T) {
+			ctx := contextWithProjects(test.allowedProjects)
+
+			response, err := server.ReadNode(ctx, &reporting.Id{Id: test.nodeId})
+
+			assert.Error(t, err)
+			assert.Nil(t, response)
+		})
+	}
+}
```
phiggins: This code is mostly the same as the test suite for ReadReport :https://github.com/chef/automate/blob/ac896d72efd26d007882ea031c4d9ce681671496/components/compliance-service/integration_test/reporting_server_read_report_test.go

components/compliance-service/reporting/relaxting/nodes.go
```diff
@@ -200,9 +199,12 @@ func (backend *ES2Backend) GetNode(nodeUuid string) (*reportingapi.Node, error)
 		"status",
 		"controls_sums")
 
+	filters["node_id"] = []string{nodeUuid}
+	query := backend.getFiltersQuery(filters, false)
```
phiggins: I *think* false is the value I want for the second argument, but I'm not really sure. That value becomes the `latestOnly` value used to choose a branch of this conditional: https://github.com/chef/automate/blob/master/components/compliance-service/reporting/relaxting/reports.go#L792-L799

My reasoning is that passing in true would add an additional filter that is not present currently.

components/compliance-service/integration_test/reporting_server_read_node_test.go
```diff
@@ -0,0 +1,113 @@
...
+			allowedProjects: []string{authzConstants.AllProjectsExternalID},
+			nodeId:          assignedNodeId,
+		},
+		{
+			description:     "Projects: user has access to all projects accessing an unassigned node",
+			allowedProjects: []string{authzConstants.AllProjectsExternalID},
+			nodeId:          unassignedNodeId,
+		},
+		{
+			description:     "Projects: user has access to all projects a node belongs to acessing an assigned node",
```
blakestier: lil typo (`acessing`)

blakestier:
🎉

components/compliance-service/api/reporting/server/server.go
```diff
@@ -285,7 +285,13 @@ func (srv *Server) ListNodes(ctx context.Context, in *reporting.Query) (*reporti
 
 // ReadNode returns a node based on id
 func (srv *Server) ReadNode(ctx context.Context, in *reporting.Id) (*reporting.Node, error) {
-	node, err := srv.es.GetNode(in.Id)
+	formattedFilters := formatFilters([]*reporting.ListFilter{})
```
lancewf: Why call `formatFilters` here and not create an empty `map[string][]string`?

phiggins: I considered that, but I thought it was worth having all of these go through the same code path in case we wanted to change how they all work at once (add some default filter, change the format, etc).

PR #245, opened 2019-05-02
[A2-779] Projects filter refinement

msorens:

This PR implements remaining bits of the global projects filter.
Essentially the first thing I did was back up the reducer with a comprehensive set of unit tests. 
But I did that using what I call "Extreme TDD": write all the test names first to convert  requirements to code: 
image

See the full list here:
https://github.com/chef/automate/blob/046fa3f5af0f1a28420aa482423c46f2687217f9/components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.spec.ts#L67

Once those requirements were confirmed, I then fleshed out all the tests.
This allowed me to easily determine what was left to do; the list below shows the failing tests, indicating that the component was not successfully doing what the test purports.
image

What made the creation of a comprehensive set of unit tests a very straightforward task was @scottopherson 's wonderfully clean code, wherein the visual components were utterly simple and  the complexity was kept in the reducer. That is the ideal design for testability because the reducer is, by definition, a pure function, and pure functions are the most one could desire for unit testing. I only needed to add one missing facet to the interface (`dropdownCaretVisible`) to be able to test the dropdown label content, the display of the dropdown caret, the count of projects in the badge adjacent to the label, the presence of the badge, and even the color of the badge.

Besides telling me what was left to be done, the set of unit tests made it fast to iterate over changes. I no longer had to manually try out multiple scenarios in the browser with every change. I just made some edits, and with the magic of the Wallaby continuous test runner, I had instant feedback as to whether the unit tests were getting better or worse.

This was amazingly useful in the reducer (projects-filter.reducer.ts):  if you take a look at the `selectionLabel` function it is essentially a codified Karnaugh map. I could thus swap things around and try different logic as I was worked toward a minimization of terms there.

BTW, see PR - #172 for video of how the visual component operates.

### :+1: Definition of Done

All unit tests pass--see chart below.
Also included in this PR:

1. Fixed the descenders being clipped in the dropdown label (thanks for pointing that out @susanev !).
2. Made the global project filter appear on all pages, not just settings pages. (Technically, it would appear on all pages before, but only _after_ you first visited the Settings page.)
3. Poll every 5 minutes to re-evaluate the IAM version so that, should the IAM version change, the global projects filter (and everything else depending on IAM version) will re-sync automatically without requiring any browser refresh.
4. Cleanup the use of the magic string `(unassigned)` in the ui project so that--like authz, teams, and authn before it--the string is only defined once in the project.

image

1. start_all_services
2. Build your UI component locally via the method of your choice. Ultimately the last step should be `make serve` so that if you make any edits as mentioned below, they will be automatically picked up.
3. Load the app in the browser.

The global projects filter will be present on every page.
It remembers its last value even after refreshing the browser (thanks, @scottopherson !)
The list of projects you find there are hard-coded, static data.
To test out the various scenarios in the requirements, just edit the list at the bottom of projects-filter.service.ts. Reload your browser once the ui rebuilds and then check the filter.

PR - #172 

- [x] Code actually executed?

### DIFF
--- a/components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
+++ b/components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
@@ -6,7 +6,7 @@ import { orderBy } from 'lodash/fp';
 
 import { Project, ProjectConstants } from 'app/entities/projects/project.model';
 
-const UNASSIGNED_PROJECT_ID = ProjectConstants.UNASSIGNED_PROJECT_ID;
+const { UNASSIGNED_PROJECT_ID } = ProjectConstants;
 
 @Component({
   selector: 'app-projects-dropdown',
--- a/components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
+++ b/components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
@@ -4,9 +4,9 @@ import {
 
 import { orderBy } from 'lodash/fp';
 
-import { Project } from 'app/entities/projects/project.model';
+import { Project, ProjectConstants } from 'app/entities/projects/project.model';
 
-const UNASSIGNED_LABEL = '(unassigned)';
+const UNASSIGNED_PROJECT_ID = ProjectConstants.UNASSIGNED_PROJECT_ID;
 
 @Component({
   selector: 'app-projects-dropdown',
@@ -31,7 +31,7 @@ export class ProjectsDropdownComponent implements OnInit {
 
   active = false;
   showError = false;
-  label = UNASSIGNED_LABEL;
+  label = UNASSIGNED_PROJECT_ID;
 
   // Map of projects currently selected
   selectedProjects = {};
@@ -109,7 +109,7 @@ export class ProjectsDropdownComponent implements OnInit {
         break;
       }
       case 0: {
-        this.label = UNASSIGNED_LABEL;
+        this.label = UNASSIGNED_PROJECT_ID;
         break;
       }
       default: {
--- a/components/automate-ui/src/app/entities/policies/policy.effects.ts
+++ b/components/automate-ui/src/app/entities/policies/policy.effects.ts
@@ -1,7 +1,7 @@
 import { Injectable } from '@angular/core';
 import { HttpErrorResponse } from '@angular/common/http';
 import { Actions, Effect, ofType } from '@ngrx/effects';
-import { of as observableOf } from 'rxjs';
+import { interval as observableInterval,  of as observableOf, Observable } from 'rxjs';
 import { catchError, mergeMap, map } from 'rxjs/operators';
 
 import { memberListToStringList, policyFromPayload } from './policy.model';
@@ -50,12 +50,17 @@ export class PolicyEffects {
   //
   // [1]: https://github.com/ngrx/platform/commit/8d56a6f7e9a0158d30c6a0f335f9805cc7d5a555
 
+  private POLLING_INTERVAL_IN_SECONDS = 300; // 5 minutes
+
   @Effect()
-  getiamMajorVersion$ = this.actions$.pipe(
+  newIamVersionStatus$ = observableInterval(1000 * this.POLLING_INTERVAL_IN_SECONDS)
+    .pipe(mergeMap(this.getIamVersionAction$()));
+
+
+  @Effect()
+  getIamVersion$ = this.actions$.pipe(
     ofType<GetIamVersion>(PolicyActionTypes.GET_IAM_VERSION),
-    mergeMap(() => this.requests.getIamVersion().pipe(
-      map(resp => new GetIamVersionSuccess(resp)),
-      catchError((error: HttpErrorResponse) => observableOf(new GetIamVersionFailure(error))))));
+    mergeMap(this.getIamVersionAction$()));
 
   @Effect()
   getIamVersionFailure$ = this.actions$.pipe(
@@ -174,5 +179,10 @@ export class PolicyEffects {
       type: Type.error,
       message: `Could not delete policy: ${error.error || error}`
     })));
-}
 
+  private getIamVersionAction$(): () => Observable<PolicyActions> {
+    return () => this.requests.getIamVersion().pipe(
+      map(resp => new GetIamVersionSuccess(resp)),
+      catchError((error: HttpErrorResponse) => observableOf(new GetIamVersionFailure(error))));
+  }
+}
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
@@ -8,7 +8,7 @@
     [ngClass]="{'active': selectionCountActive }">
     {{ selectionCount }}
   </span>
-  <chef-icon *ngIf="this.editableOptions.length > 1" aria-hidden>keyboard_arrow_down</chef-icon>
+  <chef-icon *ngIf="dropdownCaretVisible" aria-hidden>keyboard_arrow_down</chef-icon>
 </button>
 <chef-dropdown class="dropdown" *ngIf="dropdownActive" (keydown.esc)="handleEscape()">
   <chef-click-outside class="dropdown-body" omit="dropdown-label" (clickOutside)="handleEscape()">
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
@@ -8,7 +8,7 @@
     [ngClass]="{'active': selectionCountActive }">
     {{ selectionCount }}
   </span>
-  <chef-icon aria-hidden>keyboard_arrow_down</chef-icon>
+  <chef-icon *ngIf="this.editableOptions.length > 1" aria-hidden>keyboard_arrow_down</chef-icon>
 </button>
 <chef-dropdown class="dropdown" *ngIf="dropdownActive" (keydown.esc)="handleEscape()">
   <chef-click-outside class="dropdown-body" omit="dropdown-label" (clickOutside)="handleEscape()">
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
@@ -1,7 +1,7 @@
 <button class="dropdown-label"
   aria-label="Projects filter"
   [attr.title]="selectionLabel"
-  [attr.tabindex]="dropdownCaretVisible? 0 : -1"
+  [disabled]="!dropdownCaretVisible"
   (click)="handleLabelClick()">
   <span class="selection-label">{{ selectionLabel }}</span>
   <span class="selection-count"
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
@@ -1,6 +1,7 @@
 <button class="dropdown-label"
   aria-label="Projects filter"
   [attr.title]="selectionLabel"
+  [attr.tabindex]="dropdownCaretVisible? 0 : -1"
   (click)="handleLabelClick()">
   <span class="selection-label">{{ selectionLabel }}</span>
   <span class="selection-count"
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.spec.ts
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.spec.ts
@@ -31,17 +31,16 @@ describe('ProjectsFilterDropdownComponent', () => {
       component.selectionCountVisible = true;
       component.selectionCountActive = true;
       fixture.detectChanges();
-
-      label = fixture.nativeElement.querySelector('.dropdown-label');
-      count = fixture.nativeElement.querySelector('.selection-count');
     });
 
     it('displays label text', () => {
+      label = fixture.nativeElement.querySelector('.dropdown-label');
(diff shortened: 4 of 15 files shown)


components/automate-ui/src/app/entities/policies/policy.effects.ts
```diff
@@ -50,12 +50,17 @@ export class PolicyEffects {
...
+
+
   @Effect()
-  getiamMajorVersion$ = this.actions$.pipe(
+  getIamVersion$ = this.actions$.pipe(
     ofType<GetIamVersion>(PolicyActionTypes.GET_IAM_VERSION),
-    mergeMap(() => this.requests.getIamVersion().pipe(
-      map(resp => new GetIamVersionSuccess(resp)),
-      catchError((error: HttpErrorResponse) => observableOf(new GetIamVersionFailure(error))))));
+    mergeMap(this.getIamVersionAction$()));
```
msorens: Extracted the code to fetch the license version into a function, so that the very same function could be used in `newIamVersionStatus` just above for periodic polling after page load.

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
```diff
@@ -4,9 +4,9 @@ import {
 
 import { orderBy } from 'lodash/fp';
 
-import { Project } from 'app/entities/projects/project.model';
+import { Project, ProjectConstants } from 'app/entities/projects/project.model';
 
-const UNASSIGNED_LABEL = '(unassigned)';
+const UNASSIGNED_PROJECT_ID = ProjectConstants.UNASSIGNED_PROJECT_ID;
```
msorens: Drive-by: consolidate magic strings to one definition per service.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
```diff
@@ -8,7 +8,7 @@
     [ngClass]="{'active': selectionCountActive }">
     {{ selectionCount }}
   </span>
-  <chef-icon aria-hidden>keyboard_arrow_down</chef-icon>
+  <chef-icon *ngIf="dropdownCaretVisible" aria-hidden>keyboard_arrow_down</chef-icon>
```
msorens: Caret is now shown conditionally, when appropriate.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.spec.ts
```diff
@@ -31,28 +31,27 @@ describe('ProjectsFilterDropdownComponent', () => {
       component.selectionCountVisible = true;
       component.selectionCountActive = true;
       fixture.detectChanges();
-
-      label = fixture.nativeElement.querySelector('.dropdown-label');
-      count = fixture.nativeElement.querySelector('.selection-count');
     });
 
     it('displays label text', () => {
+      label = fixture.nativeElement.querySelector('.dropdown-label');
```
msorens: Drive-by: This and other changes in this file are mostly for clarity.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.spec.ts
```diff
@@ -180,12 +182,26 @@ describe('ProjectsFilterDropdownComponent', () => {
...
+          value: 'project-1',
+          label: 'Project 1',
+          checked: false
+        },
+        {
+          value: 'project-2',
+          label: 'Project 2',
+          checked: true
+        }
+      ];
```
msorens: The above converts this from a phantom test to a real test.

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.spec.ts
```diff
@@ -0,0 +1,549 @@
...
+  describe('LOAD_OPTIONS', () => {
+    const action = new LoadOptions();
+
+    it('sets status to loading', () => {
+      const { optionsLoadingStatus } = projectsFilterReducer(initialState, action);
+      expect(optionsLoadingStatus).toEqual(EntityStatus.loading);
+    });
+  });
+
+  describe('LOAD_OPTIONS_SUCCESS', () => {
```
msorens: This closely aligns to the requirements in the JIRA card. There are actually more cases here, though, filling in a few gaps in those requirements (derived from the video attached to the JIRA card).

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.spec.ts
```diff
@@ -0,0 +1,549 @@
...
+      ];
+      const action = new LoadOptionsSuccess(payload);
+
+      const { options } = projectsFilterReducer(initialState, action);
+
+      expect(map('value', options))
+        .toEqual(['a-proj', 'b-proj', 'd-proj', 'zz-proj', UNASSIGNED_PROJECT_ID]);
+    });
+
+    describe('with exactly one allowed project', () => {
```
msorens: Starting here each test section checks the dropdown label, the adjacent caret, and the adjacent badge (value and color) showing the count of selected or available items.

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.ts
```diff
@@ -13,20 +20,22 @@ export interface ProjectsFilterOption {
 
 export interface ProjectsFilterState {
   options: ProjectsFilterOption[];
-  optionsLoadingStatus: LoadingStatus;
+  optionsLoadingStatus: EntityStatus;
```
msorens: Converted from compliance's version to auth's version for consistency.

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.ts
```diff
@@ -36,22 +45,35 @@ export function projectsFilterReducer(
...
+        .sort((a, b) => {
+          const opts = { numeric: true, sensitivity: 'base' };
+          return a.label.localeCompare(b.label, undefined, opts)
+            || a.label.localeCompare(b.label, undefined, { numeric: true });
+        }
+        );
+      const unassignedProject = find(['value', UNASSIGNED_PROJECT_ID], action.payload);
+      if (unassignedProject) {
+        sortedOptions.push(unassignedProject);
+      }
```
msorens: The above forces unassigned to be at the bottom of the list.

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.ts
```diff
@@ -70,46 +93,43 @@ export function projectsFilterReducer(
...
   }
 
-  if (hasOnlyAllProjectsChecked) {
-    return 'All projects';
+  if (hasSomeProjectsChecked) {
+    return MULTIPLE_PROJECTS_LABEL;
   }
 
-  return 'All resources';
+  return ALL_PROJECTS_LABEL;
```
msorens: All the changes in the above `selectionLabel` function were to make the unit tests pass.

components/automate-ui/src/app/ui.component.ts
```diff
@@ -61,6 +62,7 @@ export class UIComponent implements OnInit {
     this.store.select(routeURL).subscribe((url: string) => {
       this.renderNavbar = url.split('/').pop() !== 'add-members';
     });
+    this.store.dispatch(new GetIamVersion());
```
msorens: This loads the IAM version on app load, as the global project filter only appear if it knows the system is on v2.1.

msorens:
@susanev 
1. Fixed tabbing issue with latest commit.
2. Very odd Safari behavior; not clear how to resolve that.

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.ts
scottopherson: We could keep this switch case simple by sorting the success payload instead: 
https://github.com/chef/automate/commit/4540debd3df47ef7a3121f91a4f6b45bda3873f0 <- working example but breaks the reducer tests

Moving these transformations into the effect also kinda helps clarify what's being described in [this previous comment](https://github.com/chef/automate/pull/172#discussion_r278750284). Or perhaps we could do the inverse by changing the success payload in a way that would allow all this [merging and sorting](https://github.com/chef/automate/commit/4540debd3df47ef7a3121f91a4f6b45bda3873f0#diff-4242aab319947ab25a57f84b34fb4240R47) to happen in the reducer file and keep the effect file super simple instead 🤔 

I'm +1 regardless though since this all works as advertised and is well-tested 👍👍

components/automate-ui/src/app/entities/policies/policy.effects.ts
msorens: Good question: It does not require an explicit triggering action. As soon as ngrx-effects initializes it starts up.
Oh, note, however, by "starts up" I do not mean it emits the first request--that's why there exists a separate call in ui.component.ts does. "Starts up" for an `observableInterval` means starts the timer; the first emission is at the end of the interval--you can see a visualization of that here: https://rxviz.com/examples/basic-interval

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.ts
```diff
@@ -1,10 +1,17 @@
...
+import { ProjectConstants } from 'app/entities/projects/project.model';
 import {
   ProjectsFilterActions,
   ProjectsFilterActionTypes
 } from './projects-filter.actions';
 
+const UNASSIGNED_PROJECT_ID = ProjectConstants.UNASSIGNED_PROJECT_ID;
+const ALL_RESOURCES_LABEL = ProjectConstants.ALL_RESOURCES_LABEL;
+const ALL_PROJECTS_LABEL = ProjectConstants.ALL_PROJECTS_LABEL;
+const MULTIPLE_PROJECTS_LABEL = ProjectConstants.MULTIPLE_PROJECTS_LABEL;
```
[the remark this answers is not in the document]
msorens: Ah, yes, I like your suggestion much better here, @scottopherson !

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.ts
msorens: Appreciate the suggestions--I will take a look at what you suggested.

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.ts
```diff
@@ -36,22 +45,35 @@ export function projectsFilterReducer(
...
 
     case ProjectsFilterActionTypes.LOAD_OPTIONS: {
-      return set('optionsLoadingStatus', LoadingStatus.loading, state);
+      return set('optionsLoadingStatus', EntityStatus.loading, state);
     }
 
     case ProjectsFilterActionTypes.LOAD_OPTIONS_SUCCESS: {
+      const sortedOptions = action.payload
+        .filter(o => o.value !== UNASSIGNED_PROJECT_ID)
+        .sort((a, b) => {
```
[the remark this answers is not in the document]
msorens: Good thought, Brenna, but each of those anonymous functions passed to `Array.sort()` is customized for a field of the given object. I know a way to do it but I may leave that for a separate PR as it is a little bit tricky.

srenatus:
IIRC the introduction of the periodic license check was when we got into the "don't wait for angular" situation in our e2e tests. This could be a decent moment to revisit that.

msorens:
@scottopherson After reviewing your suggests on co-locating the merge and fetch, I definitely like your approach. I chose to use your "inverse" idea though--putting it in the reducer--since that is readily testable. Ngrx/effects are testable, yes, but that requires getting up to speed on marble testing, since we have not yet introduced that into the code base (been on my "to do" list for sometime).

components/automate-ui/src/app/services/projects-filter/projects-filter.effects.ts
```diff
@@ -27,11 +27,10 @@ export class ProjectsFilterEffects {
       return this.projectsFilter.fetchOptions().pipe(
         map((fetched: ProjectsFilterOption[]) => {
           const restored = this.projectsFilter.restoreOptions() || [];
-          const loaded = fetched.map(fetchedOpt => {
-            const restoredOpt = restored.filter(opt => opt.value === fetchedOpt.value)[0];
-            return restoredOpt ? { ...fetchedOpt, checked: restoredOpt.checked } : fetchedOpt;
+          return new LoadOptionsSuccess(<ProjectsFilterOptionTuple>{
+            fetched: fetched,
+            restored: restored
```
msorens: Moved this logic into the reducer to (a) co-locate with the sorting of options and (b) to be able to readily add tests for it.

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.spec.ts
```diff
@@ -0,0 +1,692 @@
...
+        );
+
+        const { options } = projectsFilterReducer(initialState, action);
+
+        expect(map('value', options))
+          .toEqual(['a-proj', 'A-proj', 'b-proj']);
+      });
+    });
+
+    describe('merging fetched and stored options', () => {
```
msorens: New section of tests added to test merging the newly fetched options from the back-end with the browser-retained stored options.

msorens:
@susanev Adjusted the single project case to completely disable it (not tabbable and not selectable).
I will check with @magwalk about your second item, the "tabbing-off" action.

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.ts
```diff
@@ -36,22 +52,25 @@ export function projectsFilterReducer(
   switch (action.type) {
 
     case ProjectsFilterActionTypes.LOAD_OPTIONS: {
-      return set('optionsLoadingStatus', LoadingStatus.loading, state);
+      return set('optionsLoadingStatus', EntityStatus.loading, state);
     }
 
     case ProjectsFilterActionTypes.LOAD_OPTIONS_SUCCESS: {
+      const merged = mergeOptions(action.payload.fetched, action.payload.restored);
+      const sortedOptions = sortOptions(merged);
```
scottopherson: random nit: would go with `mergedOptions/sortedOptions` or `merged/sorted` for these var names.

Issue #192, opened 2019-04-27
one node view (backend work)


vjeffrey:
# User Story

In an effort to make it easier for users to get a quick view of their infrastructure as a whole, we would like to introduce something called the "one node view".

This "one node view" will be based on metadata information about all nodes, stored in postgres.

## how it works
at ingestion time, both client runs ingestion service and compliance ingestion service should call NodeManagerService.ProcessNode with metadata information about the node. This information should include:
- uuid
- node name
- platform name and release
- tags
- last scan or run:
      - id (report or run id)
      - end_time
      - status
- source information:
      - source_id
      - source_region
      - source_account_id

the end goal is to be able to populate a view that would contain (at least) the following information:
<img width="852" alt="Screen Shot 2019-04-27 at 11 50 27" src="https://user-images.githubusercontent.com/10341541/56853278-c5169580-68e2-11e9-9e73-cd9bdf4fa30a.png">

_Note: real designs are in progress. cc @jonong1972_ 

## filtering and endpoints
the endpoint used to populate this view will be the `api/v0/nodes/search` view.  this endpoint should support filtering by
- name
- platform
- platform version
- last check in range
- last scan time range
- tags 
- reachable/unreachable/unknown status
- terminated/stopped/running/unknown state
- failed client runs
- failed scans 
- nodes whose penultimate client run status was passed and latest is failed
- nodes whose penultimate scan job status was passed and latest is failed

and the combination of any of these filters.

the list of nodes produced should be downloadable (json/csv) 

the database storing this information should also keep the penultimate status and time for both runs and scans, allowing us to easily display a notification if a node that was previously passing a client run or scan begins to fail the client run or scan.

# Definition of Done
- [x] compliance ingestion calls NodeManagerService.ProcessNode with the necessary metadata https://github.com/chef/automate/pull/294)
- [x] client run ingestions calls NodeManagerService.ProcessNode with the necessary metadata (https://github.com/chef/automate/pull/345)
- [x] all above metadata is stored in the database (postgres) (https://github.com/chef/automate/pull/294)
- [x] able to filter by all above mentioned filters via api, combining any of them (https://github.com/chef/automate/pull/332)
- [x] ensure all necessary data is exposed at the gateway level (https://github.com/chef/automate/pull/347)
- [x] endpoints are documented (https://github.com/chef/automate/pull/347)

able to download the list of nodes via api, respecting all applied filters (separate issue: https://github.com/chef/automate/issues/366)

PR #253, opened 2019-05-06
include tags on nodemanager process node

vjeffrey:
this change is in support of the work towards the
one node view.  also included is a change to ensure we only
add a tag if it does not already exist in the db and only
add a node id-tag id association if it does not already exist

### :+1: Definition of Done
tags are passed through from scan job -> report
tags are added as part of the processNode codepath
tags are not added to the db if they already exist in the system
node tags associations are not added to the db if they already exist in the system


components/nodemanager-service/pgdb/node-managers-process-node.go
```diff
@@ -60,16 +60,31 @@ func (db *DB) ProcessIncomingNode(node *manager.NodeMetadata) error {
...
-		_, err = db.Exec(sqlUpsertByID, node.GetUuid(),
-			node.GetName(), node.GetPlatformName(), node.GetPlatformRelease(),
-			nodeState, lastContact, node.GetSourceRegion(), node.GetSourceAccountId(),
-			node.GetJobUuid())
-	} else {
-		_, err = db.Exec(sqlUpsertBySourceID, node.GetUuid(),
-			node.GetName(), node.GetPlatformName(), node.GetPlatformRelease(),
-			nodeState, lastContact, node.GetSourceId(), node.GetSourceRegion(),
-			node.GetSourceAccountId(), node.GetJobUuid())
+	err = Transact(db, func(tx *DBTrans) error {
```
lancewf: I like the Transact here.

components/compliance-service/inspec-agent/types/types.go
```diff
@@ -67,6 +68,7 @@ type InspecJob struct {
 	ProfilesOwner     string              `json:"profiles_owner,omitempty"`
 	InternalProfiles  []string            `json:"internal_profiles,omitempty"`
 	MachineIdentifier string              `json:"machine_identifier,omitempty"`
+	Tags              []*common.Kv        `json:"kv,omitempty"`
```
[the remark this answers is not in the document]
vjeffrey: oh ya it totally should!

components/nodemanager-service/pgdb/tags.go
```diff
@@ -29,15 +50,56 @@ func (trans *DBTrans) addTags(tags []*common.Kv) ([]string, error) {
...
+			link := NodeTag{
+				NodeID: nodeID,
+				TagID:  tagID,
+			}
+			links = append(links, &link)
 		}
-		links = append(links, &link)
 	}
 	return trans.Insert(links...)
 }
```
phiggins: I realize this is asking for quite a bit of rework, but I feel like this database syncing stuff would be better done in the database, by making the relevant columns unique and doing `INSERT ... ON CONFLICT DO NOTHING`.

phiggins:
```
CONFIG-008

One of the aliases you have provided in project.alias is already in use by another agent. Please update your list to remove the conflict.
```

The CI failure is probably not related to this PR and I'm guessing just needs a retry, but I've never seen this one before.

vjeffrey:
I think the CI failure is b/c i'm branched off a branch

components/nodemanager-service/pgdb/tags.go
vjeffrey: ah good point, that would be cleaner. i'll give that a try.

vjeffrey:
that's so annoying. i went to change the base branch and it closed my pr and then told me i can't change the base branch on a closed pr :/
i think it was closed because the base branch was deleted. bleh

vjeffrey:
well anyways, ill work on the improvements mentioned and the reopen with a new pull request since it won't allow me to re-open this one

PR #255, opened 2019-05-06
Accept CHEF_LICENSE for inspec check/exec

alexpop:
Also added an assert to ensure tests don't pass if `logSensitiveData` is not `false`

### DIFF
--- a/components/compliance-service/inspec-agent/runner/worker.go
+++ b/components/compliance-service/inspec-agent/runner/worker.go
@@ -196,16 +196,18 @@ func secretInfoExists(tc *inspec.TargetConfig) bool {
 }
 
 func cloudEnvVars(tc *inspec.TargetConfig) (map[string]string, error) {
+	envsMap := map[string]string{
+		"CHEF_LICENSE": "accept-no-persist",
+	}
 	switch tc.Backend {
 	case "aws":
 		if tc.AwsUser == "" || tc.AwsPassword == "" {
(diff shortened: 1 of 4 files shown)


components/compliance-service/inspec-agent/runner/worker.go
```diff
@@ -196,16 +196,18 @@ func secretInfoExists(tc *inspec.TargetConfig) bool {
...
 	switch tc.Backend {
 	case "aws":
 		if tc.AwsUser == "" || tc.AwsPassword == "" {
 			logrus.Debugf("no aws creds found in env vars, no aws creds found for node; attempting to use aws credential chain via inspec/train")
-			return map[string]string{}, nil
-		}
-		envsMap := map[string]string{
-			"AWS_ACCESS_KEY_ID":     tc.AwsUser,
-			"AWS_SECRET_ACCESS_KEY": tc.AwsPassword,
+			return envsMap, nil
```
phiggins: I'm not sure about this change (and the others like it). I don't think we should return any usable data in the cases where we're in a bad state.

alexpop: I know what you mean, I was thinking about it as well. I just wanted to ensure that `CHEF_LICENSE` ENV is never missed in any of the switch cases. And keep the code easy to read. More important I think is handling the error which the consumers of this function need to handle properly.

PR #258, opened 2019-05-06
Add v2 version of UpdateSelf

tylercloke:

Just want a v2 version of UpdateSelf so we have complete parity.

- [ ] Code actually executed?


srenatus:
Unless I'm missing something, this is only about a 1-to-1 correspondence of v1 methods vs v2 methods. Is that good enough a reason to introduce the code? I think the v1 method should have never been there in the first place, and took v2 as an opportunity to fix this oddity. Functionally, there's nothing missing in v2, and there's no need for this method. Let's not introduce something that is not needed? 😉

srenatus:
Internal previous PR with some context : https://github.com/chef/a2/pull/4791

tylercloke:
@srenatus The functionality between `UpdateUser` and `UpdateSelf` aren't the same. For `UpdateSelf`, you need to specify your correct, previous password to make password changes. `UpdateUser` doesn't have that requirement. There is no part of the V2 API that currently supports that functionality as far as I can tell, so we need this method.

components/automate-gateway/api/iam/v2beta/users.proto
```diff
@@ -51,4 +51,14 @@ service Users {
...
     option (chef.automate.api.iam.policy).action = "iam:users:update";
   };
+  rpc UpdateSelf (UpdateSelfReq) returns (UpdateSelfResp) {
+    option (google.api.http).put = "/iam/v2beta/self/{id}";
+    option (google.api.http).body = "*";
+    // The users namespace, as opposed for auth:users, is for actions users
+    // are allowed to perform on themselves by default without admin permissions.
+    option (chef.automate.api.policy).resource = "users:{id}";
+    option (chef.automate.api.policy).action = "update";
+    option (chef.automate.api.iam.policy).resource = "iam:usersSelf:{id}";
```
srenatus: But I still maintain that the extra resource and action is avoidable 😉

tylercloke: I'd argue we should remove the ability for users to hit UpdateUser on themselves by default.

tylercloke: Gonna merge for now so we can keep going, but feel free to refactor if you have a better approach :)

PR #260, opened 2019-05-06
[e2e] docs / readme cleanup

sdelano:
- set column width to 80 characters
- add notes about running cypress from `e2e` dir
- s/a2/automate/ (ish)

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -2,16 +2,20 @@
 
 Cypress is a tool for complete end-to-end testing.
 
-Cypress is ideal for testing any interactions with the A2 UI.
-This includes anything from logging in to filling out a form to create an A2 resource.
+Cypress is ideal for testing any interactions with the Chef Automate
+UI. This includes anything from logging in to filling out a form to
+create an Automate resource.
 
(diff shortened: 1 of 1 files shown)


e2e/README.md
```diff
@@ -26,59 +30,68 @@ don't already have `../dev/secrets-env.sh`:
...
-Your Cypress tests may be passing locally, but it's a good idea to check their stability
-when pointed at `dev` and `acceptance` environments.
+Your Cypress tests may be passing locally, but it's a good idea to
+check their stability when pointed at `dev` and `acceptance`
+environments.
 
 To do so, follow these steps.
 
-1. Log on to a docker container (using the same image buildkite uses), using your local A2 as the working directory.
+1. Log on to a docker container (using the same image buildkite uses),
```
msorens: While it is perfectly correct to number all list items as 1, and let the md post-processor generate the sequence, I have been suggesting the alternate approach, which is also valid: using the actual sequence numbers (1, 2, 3 in this case). That allows one to read the raw markdown file with more context. And the markdownlint tool vets it equally well--it will let you know if later you put things out of order.

PR #263, opened 2019-05-07
include tags on nodemanager process node 

vjeffrey:
original pr: https://github.com/chef/automate/pull/253

this change is in support of the work towards the
one node view.  also included is a change to ensure we only
add a tag if it does not already exist in the db and only
add a node id-tag id association if it does not already exist

### :+1: Definition of Done
tags are passed through from scan job -> report
tags are added as part of the processNode codepath
tags are not added to the db if they already exist in the system
node tags associations are not added to the db if they already exist in the system


components/nodemanager-service/pgdb/tags.go
```diff
@@ -5,41 +5,65 @@ import (
...
 	tagIDs := make([]string, 0, len(tags))
-	tagArr := make([]interface{}, 0, len(tags))
 
 	for _, keyValue := range tags {
-		tag := tag{
-			ID:    createUUID(),
-			Key:   keyValue.Key,
-			Value: keyValue.Value,
+		id := createUUID()
+		rows, err := trans.Query(sqlInsertTag, id, keyValue.Key, keyValue.Value)
```
phiggins: I *think* you can replace `Query` with `SelectStr` and it'll return the id string straight away without having to mess with the row.

vjeffrey: ooooh that would be nicer! ill give it a try/look it up

PR #264, opened 2019-05-07
do not run detect job if node has ssm backend

vjeffrey:
We do not currently support detect jobs on ssm nodes.  This is functionality we would like to one day support, but it will require a bit more work.  
In one case we were erroneously allowing detect ssm jobs to go through, marking them as "reachable" without ever having tried to reach them, leading to a confusing experience for the user.
In another case we were (again) erroneously allowing detect ssm jobs to go through, but not detecting that the node was ssm, and trying to run a detect on the job, resulting in a cryptic error of type `train ssm plugin missing`.
So this work ensures we skip nodes if they meet two criteria:
- job type is detect
- node backend is ssm

### :+1: Definition of Done
ssm nodes skip detect jobs

build compliance
add aws-ec2 integration (credentialless)
expect to _not_ see nodes be reachable. expect to _not_ see status update when rerun button is clicked.
run ssm job; expect it to work as usual

### DIFF
--- a/components/compliance-service/inspec-agent/resolver/resolver.go
+++ b/components/compliance-service/inspec-agent/resolver/resolver.go
@@ -466,6 +466,10 @@ func (r *Resolver) handleManagerNodes(ctx context.Context, m *manager.NodeManage
 							backend = inspec.BackendAZWindows
 						}
 					}
+					if ssmJob == true && job.Type == "detect" {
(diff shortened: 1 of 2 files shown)


vjeffrey:
we really need to improve the testing story for the inspec-agent. it has so few tests right now.. i might be able to write one up real quick...

vjeffrey:
i pulled the logic out into a separate function to make it more testable

PR #270, opened 2019-05-07
remove unique constraint on tags and nodes_tags

vjeffrey:
adding a unique constraint breaks on boxes with real data so here i am removing that and reverting back to the original logic.
https://github.com/chef/automate/pull/263/commits

### DIFF
--- a/components/nodemanager-service/pgdb/migration/sql/02_add_unique_to_tags.up.sql
+++ b/components/nodemanager-service/pgdb/migration/sql/02_add_unique_to_tags.up.sql
@@ -1,5 +1,0 @@
-ALTER TABLE IF EXISTS tags
-  ADD UNIQUE (key, value);
-
-ALTER TABLE IF EXISTS nodes_tags
-  ADD UNIQUE (node_id, tag_id);
--- a/components/nodemanager-service/pgdb/migration/sql/02_add_unique_to_tags.up.sql
+++ b/components/nodemanager-service/pgdb/migration/sql/02_add_unique_to_tags.up.sql
@@ -1,0 +1,10 @@
+-- ALTER TABLE IF EXISTS tags
+--   ADD UNIQUE (key, value);
+
+-- ALTER TABLE IF EXISTS nodes_tags
+--   ADD UNIQUE (node_id, tag_id);
+
+-- Setting a unique on these tables doesn't work for instances
+-- that already have a lot of data in them, because some duplicates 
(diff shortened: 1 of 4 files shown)


components/nodemanager-service/pgdb/migration/sql/02_add_unique_to_tags.up.sql
```diff
@@ -1,5 +0,0 @@
-ALTER TABLE IF EXISTS tags
-  ADD UNIQUE (key, value);
-
-ALTER TABLE IF EXISTS nodes_tags
-  ADD UNIQUE (node_id, tag_id);
```
sdelano: I don't think we can just delete a migration that's been applied to all the nodes in our infrastructure can we?

I think we might need to:

1. edit the migration here to either not add the uniqueness constraint or do nothing
1. add another migration that if the table and constraint exist, delete them

This change was deployed already and deleting the code that deployed it won't un-deploy it.

vjeffrey: ya, i was gonna manually do that for our envs but it's probably safer to do it in the migration. ill push a file up

vjeffrey: (in case anyone other than us is consuming dev)

sdelano: We've also got about 15 deployed environments now. It's probably easier here.

components/nodemanager-service/pgdb/migration/sql/02_add_unique_to_tags.up.sql
```diff
@@ -1,5 +0,0 @@
-ALTER TABLE IF EXISTS tags
```
sdelano: Probably not worth blocking over but it'd probably be nicer to have this just commented out with a comment explaining what happened.

phiggins:
I know this is blocking things so there's some urgency involved, but we could also migrate the data to remove the duplicate things preventing the uniqueness constraints from being applied.

vjeffrey:
> we could also migrate the data

ya, i started going down that path, but i'm unfamiliar with the process there so my attempts weren't going well. i opted for the quick fix.
are you familiar with deduping data in postgres? 
we could always try to do it in a next change

phiggins:
Yeah, I'd be fine doing that later in another PR. 👍

PR #273, opened 2019-05-08
[A2-808] OPA update for project rules

msorens:

Since project rules are stored in OPA alongside policies, we need to ensure that actions that update project rules will also trigger an OPA store update.
The mechanism for this is in place for policies and roles, and is (thanks to @jaym ) already wired for High Availability.
The few changes herein allow the as-yet-to-be-written API calls for project rule handling to tie in to the same mechanism.

### :+1: Definition of Done

Calls for project rules will be able to initiate OPA updates.

1. Add some code into a project handler function (say CreateProject) in authz-service/storage/v2/postgres/postgres.go:
```
       err = p.notifyPolicyChange(ctx, tx)
       if err != nil {
       }
```
You will see that code fragment used for policy and role actions already in place.

2. Adjust the gating regex in components/authz-service/server/v2/policy.go::L676 to allow "CreateProject"  for this demo. (So change the regex to this `^/chef.automate.domain.authz.v2.\w+/(Get|List|)`.

3. rebuild components/authz-service.

4. Invoke a curl command to create a project.

5. Observe the supervisor log--the last line is the proof of the pudding:
```
msg="Projects Authorized Query" action="iam:projects:create" projects="[]" resource="iam:projects" result="[*]" subject="[token:admin-1]"
msg="Accepted notification from postgres"
msg="Received policy change notification"
msg="Initiating store update for /chef.automate.domain.authz.v2.Projects/CreateProject"
```

- [x] Code actually executed?

### DIFF
--- a/components/authz-service/server/v2/policy.go
+++ b/components/authz-service/server/v2/policy.go
@@ -2,7 +2,6 @@ package v2
 
 import (
 	"context"
-	"regexp"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -670,12 +669,25 @@ func (s *policyServer) EngineUpdateInterceptor() grpc.UnaryServerInterceptor {
 			return resp, nil
 		}
 
-		// Note: the set of methods we _do not want to update_ our engine store
-		// is the smaller one. Also, updating too often is not as bad as
-		// updating too little.
-		if regexp.MustCompile(`^/chef.automate.domain.authz.v2.\w+/(Get|List|\w+Project$)`).
-			MatchString(info.FullMethod) {
-			return resp, nil
+		switch info.FullMethod {
+		// Important! Any new endpoint that requires refreshing the OPA cache must be added here.
+		case "/chef.automate.domain.authz.v2.Policies/ReplacePolicyMembers",
+			"/chef.automate.domain.authz.v2.Policies/CreatePolicy",
+			"/chef.automate.domain.authz.v2.Policies/DeletePolicy",
+			"/chef.automate.domain.authz.v2.Policies/UpdatePolicy",
+			"/chef.automate.domain.authz.v2.Policies/MigrateToV2",
+			"/chef.automate.domain.authz.v2.Policies/ResetToV1",
+			"/chef.automate.domain.authz.v2.Policies/CreateRole",
+			"/chef.automate.domain.authz.v2.Policies/DeleteRole",
+			"/chef.automate.domain.authz.v2.Policies/UpdateRole",
+			"/chef.automate.domain.authz.v2.Policies/RemovePolicyMembers",
+			"/chef.automate.domain.authz.v2.Policies/AddPolicyMembers",
+			"/chef.automate.domain.authz.v2.Policies/PurgeSubjectFromPolicies":
+			if err := s.updateEngineStore(ctx); err != nil {
+				return nil, status.Errorf(codes.Internal, "error updating engine store: %s", err.Error())
+			}
+		default:
+			// do nothing
 		}
 
 		s.log.Infof("Initiating store update for %s", info.FullMethod)
--- a/components/authz-service/server/v2/policy.go
+++ b/components/authz-service/server/v2/policy.go
@@ -690,7 +690,7 @@ func (s *policyServer) EngineUpdateInterceptor() grpc.UnaryServerInterceptor {
 			// do nothing
 		}
 
-		s.log.Infof("Initiating store update for %s", info.FullMethod)
+		s.log.Debugf("Initiating store update for %s", info.FullMethod)
 		if err := s.updateEngineStore(ctx); err != nil {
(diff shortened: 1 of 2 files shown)


components/authz-service/server/v2/policy.go
```diff
@@ -663,26 +664,23 @@ func (s *policyServer) EngineUpdateInterceptor() grpc.UnaryServerInterceptor {
...
-			"/chef.automate.domain.authz.v2.Policies/GetPolicy",
-			"/chef.automate.domain.authz.v2.Policies/GetRole",
-			"/chef.automate.domain.authz.v2.Policies/ListPolicyMembers",
-			"/chef.automate.domain.authz.v2.Policies/GetPolicyVersion":
-			// do nothing
-		default:
-			if err := s.updateEngineStore(ctx); err != nil {
-				return nil, status.Errorf(codes.Internal, "error updating engine store: %s", err.Error())
-			}
+		if regexp.MustCompile(`^/chef.automate.domain.authz.v2.\w+/(Get|List|\w+Project$)`).
```
srenatus: Two nits: RE are slower than the string predicates - how much and do we care? Secondly, the RE never changes, so it would commonly be put in to a package variable.

components/authz-service/server/v2/policy.go
```diff
@@ -663,26 +664,23 @@ func (s *policyServer) EngineUpdateInterceptor() grpc.UnaryServerInterceptor {
...
+			MatchString(info.FullMethod) {
+			return resp, nil
+		}
+
+		s.log.Infof("Initiating store update for %s", info.FullMethod)
```
srenatus: What would you think about making this debug level?

tylercloke: I'd argue that the logs in `updateEngineStore` should also be debug. Our auth logs are way too chatty on info imo.

msorens: Inside the subsequent call to `updateEngineStore` it writes a message at `info` level`. I added this at the same `info` level only so we could distinguish in the log which calls to `updateEngineStore` are synchronous and which are asynchronous.

components/authz-service/server/v2/policy.go
```diff
@@ -663,26 +664,23 @@ func (s *policyServer) EngineUpdateInterceptor() grpc.UnaryServerInterceptor {
...
-			"/chef.automate.domain.authz.v2.Policies/GetPolicyVersion":
-			// do nothing
-		default:
-			if err := s.updateEngineStore(ctx); err != nil {
-				return nil, status.Errorf(codes.Internal, "error updating engine store: %s", err.Error())
-			}
+		if regexp.MustCompile(`^/chef.automate.domain.authz.v2.\w+/(Get|List|\w+Project$)`).
+			MatchString(info.FullMethod) {
+			return resp, nil
+		}
```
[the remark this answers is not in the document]
blakestier: There are a few other endpoints that would fall in here . . .perhaps? `ProjectUpdateStatus` and `HandleEvent` at the least. At face value, those don't seem like they'd prompt a store update.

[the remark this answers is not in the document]
blakestier: From a cursory glance at the code, I don't think they should so we might need to modify this regex

blakestier:
I know this might be a larger refactor, but I am in favor of making these updates to OPA more explicit when possible. The way we're doing it currently is a bit indirect, particularly for readers unfamiliar with this part of the codebase.

components/authz-service/server/v2/policy.go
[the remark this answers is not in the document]
msorens: Opted to switch from the list-to-skip to the list-to-proceed because the former would have just gained a bunch more entries from the project.proto file.

components/authz-service/server/v2/policy.go
```diff
@@ -663,26 +663,36 @@ func (s *policyServer) EngineUpdateInterceptor() grpc.UnaryServerInterceptor {
...
+			"/chef.automate.domain.authz.v2.Policies/AddPolicyMembers",
+			"/chef.automate.domain.authz.v2.Policies/PurgeSubjectFromPolicies":
 			if err := s.updateEngineStore(ctx); err != nil {
 				return nil, status.Errorf(codes.Internal, "error updating engine store: %s", err.Error())
 			}
+		default:
+			// do nothing
+		}
+
+		s.log.Debugf("Initiating store update for %s", info.FullMethod)
```
srenatus: ❓ Why would we have done this? Above, we selectively run the engine update for specific methods, and after the "big switch", we _do it again regardless_? Am I missing something? 😕

msorens: 😮 No, don't think you are missing anything... Clearly needs fixing.

srenatus: 👉 https://github.com/chef/automate/pull/467/commits/99840f085dc0ff630ab4cb08a59699e3f5c38fb1 wrapped into #467

PR #244, opened 2019-05-02
[A2-782] fetch real data for global projects filter

bcmdarroch:
### :nut_and_bolt: Description
Swapped the fake data for real data 🎉 

### :+1: Definition of Done
- [x] options in global projects filter are populated by backend
- [x] data gets refreshed on login and periodically thereafter
- [x] global projects filter options are accurate for logged in user
- [x] tests
- [x] Follow-up: project names display in the global projects filter - API currently only returns project IDs, but we want the names in the filter labels. we could get the names from the UI store BUT we can't be certain that a user with partial access to a project (i.e. can get `proj3` Teams) also has access to `proj3` as a resource itself ➡️[chore card](https://chefio.atlassian.net/browse/A2-832)

### :athletic_shoe: Demo Script / Repro Steps
set up your policies and some resources to play with:
```
export TOK=`chef-automate admin-token`
chef-automate iam upgrade-to-v2 --beta2.1

# (you can also create these in the UI, just make sure the IDs are the same)
curl -H "api-token: $TOK" -XPOST -d '{"id":"proj3","name":"project numero 3"}' https://localhost/apis/iam/v2beta/projects --insecure | jq
curl -H "api-token: $TOK" -XPOST -d '{"id":"proj4","name":"project numero 4"}' https://localhost/apis/iam/v2beta/projects --insecure | jq
curl -H "api-token: $TOK" -XPOST -d '{"id":"proj9","name":"project numero 9"}' https://localhost/apis/iam/v2beta/projects --insecure | jq

curl -kH "api-token: $TOK" -d '{"name": "t3", "id": "t3", "projects": ["proj3"]}'  https://a2-dev.test/apis/iam/v2beta/tokens | jq
curl -kH "api-token: $TOK" -d '{"name": "t4", "id": "t4", "projects": ["proj4"]}'  https://a2-dev.test/apis/iam/v2beta/tokens | jq

curl -kH "api-token: $TOK" -d '{"name": "t3", "id": "t3", "projects": ["proj3", "proj4"]}'  https://a2-dev.test/apis/iam/v2beta/teams | jq
curl -kH "api-token: $TOK" -d '{"name": "t4", "id": "t4", "projects": ["proj4"]}'  https://a2-dev.test/apis/iam/v2beta/teams | jq

curl -kH "api-token: $TOK" -d '{"name": "coolbean", "id": "coolbean", "password": "chefautomate"}'  https://a2-dev.test/apis/iam/v2beta/users | jq

# save in policy.json
{
[...]

PR #277, opened 2019-05-08
[A2-782] global projects filter cypress test

bcmdarroch:
follow-up to: https://github.com/chef/automate/pull/244
MUST be merged after the changes above have hit acceptance

- [ ] Code actually executed?

### DIFF
--- a/e2e/cypress/integration/02_global_filter.spec.js
+++ b/e2e/cypress/integration/02_global_filter.spec.js
@@ -1,0 +1,78 @@
+describe('global projects filter', () => {
+  const nonAdminUsername = "nonadmin"
+  const proj1  = "cypress-project-1"
+  const proj2  = "cypress-project-2"
+  const pol_id = "cypress-policy"
+
+  before(() => {
+    cy.adminLogin('/').then(() => {
+      let admin = JSON.parse(localStorage.getItem('chef-automate-user'))
+
+      cy.createProject(admin.id_token, proj1)
+      cy.createProject(admin.id_token, proj2)
+      cy.createUser(admin.id_token, nonAdminUsername)
+      cy.createPolicy(admin.id_token, pol_id, nonAdminUsername, [proj1, proj2])
+      cy.logout()
+    })
+  })
+
+  after(() => {
+    cy.adminLogin('/').then(() => {
+      let admin = JSON.parse(localStorage.getItem('chef-automate-user'))
+      cy.cleanupProjects(admin.id_token)
+      cy.cleanupUsers(admin.id_token)
+      cy.request({
+        auth: { bearer: admin.id_token },
+        method: 'DELETE',
+        url: `/apis/iam/v2beta/policies/${pol_id}`,
+        failOnStatusCode: false
+      }).then((response) => {
+        expect([200, 404]).to.include(response.status)
+      })
+    })
+  })
+
+  it('shows all projects for admin', () => {
+    cy.adminLogin('/settings')
+    // hide modal unrelated to test flow
+    cy.get('app-welcome-modal').invoke('hide')
+
+    cy.get('chef-sidebar')
+      .should('have.attr', 'minor-version')
+      .then((version) => {
+        if (version === 'v1') {
+          cy.get('[data-cy=projects-filter-button]').click()
+          
+          const allowedProjects = [proj1, proj2, '(unassigned)'];
+          cy.get('[data-cy=projects-filter-dropdown] chef-checkbox')
+            .should(($elements) => { expect($elements).to.have.length(allowedProjects.length) })
+          allowedProjects.forEach(project => {
+            cy.get('[data-cy=projects-filter-dropdown]').contains(project)
+          })
+        }
+      })
+    cy.logout()
+  })
+
+  it('shows allowed projects for non-admin', () => {
+    cy.login('/settings', nonAdminUsername)
+    // hide modal unrelated to test flow
+    cy.get('app-welcome-modal').invoke('hide')
+
+    cy.get('chef-sidebar')
+      .should('have.attr', 'minor-version')
+      .then((version) => {
+        if (version === 'v1') {
+          cy.get('[data-cy=projects-filter-button]').click()
+
+          const allowedProjects = [proj1, proj2];
+          cy.get('[data-cy=projects-filter-dropdown] chef-checkbox')
+            .should(($elements) => { expect($elements).to.have.length(allowedProjects.length) })
+          allowedProjects.forEach(project => {
+            cy.get('[data-cy=projects-filter-dropdown]').contains(project)
+          })
+        }
+      })
+    cy.logout()
+  })
+})
--- a/e2e/cypress/integration/compliance/01_ssh_scan_job.spec.js
+++ b/e2e/cypress/integration/compliance/01_ssh_scan_job.spec.js
@@ -1,6 +1,6 @@
 describe('create a manual node ssh scan job and cleanup after', () => {
   before(() => {
-    cy.login('/')
+    cy.adminLogin('/')
   })
   beforeEach(() => {
     // cypress clears local storage between tests
--- a/e2e/cypress/support/commands.js
+++ b/e2e/cypress/support/commands.js
@@ -31,7 +31,12 @@ Cypress.Commands.add("restoreStorage", () => {
   })
 });
 
-Cypress.Commands.add("login", (url) => {
+Cypress.Commands.add("adminLogin", (url) => {
+  // CYPRESS_BASE_URL environment variable must be set
+  cy.login(url, 'admin')
+})
+
+Cypress.Commands.add("login", (url, username) => {
   // CYPRESS_BASE_URL environment variable must be set
   cy.visit(url)
   cy.get('button').contains('Log in with Username').click().then(() => {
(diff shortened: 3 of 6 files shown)


e2e/cypress/integration/02_global_filter.spec.js
```diff
@@ -0,0 +1,57 @@
...
+          auth: { bearer: admin.id_token },
+          method: 'POST',
+          url: `/apis/iam/v2beta/projects`,
+          failOnStatusCode: false,
+          body: {
+            id: "cypress-project-1",
+            name: "Cypress Project One"
+          }
+      }).then((response) => {
+        // if user already exists, move on
```
srenatus: [nit] it's "project", not "user", isn't it?

e2e/cypress/integration/02_global_filter.spec.js
```diff
@@ -0,0 +1,57 @@
+describe('global projects filter', () => {
+  before(() => {
+    cy.login('/').then(() => {
+      let admin = JSON.parse(localStorage.getItem('chef-automate-user'))
+      cy.request({
+          auth: { bearer: admin.id_token },
+          method: 'POST',
+          url: `/apis/iam/v2beta/projects`,
```
srenatus: [tiniest nit] no need for backticks, as there's no string interpolation happening here; let's stick with single or double quotes

e2e/cypress/integration/02_global_filter.spec.js
```diff
@@ -0,0 +1,57 @@
+describe('global projects filter', () => {
+  before(() => {
+    cy.login('/').then(() => {
+      let admin = JSON.parse(localStorage.getItem('chef-automate-user'))
```
srenatus: hrm it's OK if that's the pattern we're currently using; but I do wonder if there's better ways to pass around state in (cypress) tests than using the browser's local storage. After all, you _just happen to know_ what is in there; it's not a public API, or local state. It could change, and it would be fine if it did change, in a way such that the UI could still work, only these tests would break because they make use of something we have no guarantees about 🤔

e2e/cypress/support/commands.js
```diff
@@ -75,3 +75,23 @@ Cypress.Commands.add("cleanupUsers", (id_token) => {
     }
   })
 })
+
+Cypress.Commands.add("cleanupProjects", (id_token) => {
+  cy.request({
+    auth: { bearer: id_token },
+    method: 'GET',
+    url: `/apis/iam/v2beta/projects`,
+    failOnStatusCode: false
```
srenatus: [nit] Do we need this? Which non-success status codes are the ones we expect and want to ignore?

e2e/cypress/support/commands.js
```diff
@@ -75,3 +75,23 @@ Cypress.Commands.add("cleanupUsers", (id_token) => {
...
+  }).then((resp) => {
+    for (let project of resp.body.projects) {
+      if (project.id.startsWith('cypress')) {
+        cy.request({
+          auth: { bearer: id_token },
+          method: 'DELETE',
+          url: `/apis/iam/v2beta/projects/${project.id}`,
+          failOnStatusCode: false
```
srenatus: [same here]

e2e/cypress/integration/02_global_filter.spec.js
blakestier: I know-- seems like we should be able to use a token since these requests aren't even _using_ the UI. I wanted to get it "working" as a first pass, but I think it's worth investigating.

e2e/cypress/support/commands.js
blakestier: I think we're doing this bc we don't want to fail on conflict (for create) or 404 (for delete). I assume Brenna was encountering conflicts on reruns of the tests? We could maybe delete them . . .

blakestier: I'll ask when she gets back since I'm maybe missing context . . .that sound ok?

e2e/cypress/integration/02_global_filter.spec.js
```diff
@@ -0,0 +1,76 @@
...
+  const nonAdminUsername = "cypressTestUserNonAdmin"
+  const proj1  = "cypress-project-1"
+  const proj2  = "cypress-project-2"
+  const pol_id = "cypress-policy"
+
+  before(() => {
+    cy.adminLogin('/').then(() => {
+      let admin = JSON.parse(localStorage.getItem('chef-automate-user'))
+
+      cy.createProject(admin.id_token, proj1)
```
blakestier: I've pulled these into helpers because I think it reads better and I imagine they'll be helpful later. But, you can accuse me of YAGNI and I'll remove them and put the code inline if that seems better.

e2e/cypress/integration/compliance/01_ssh_scan_job.spec.js
```diff
@@ -1,6 +1,6 @@
 describe('create a manual node ssh scan job and cleanup after', () => {
   before(() => {
-    cy.login('/')
+    cy.adminLogin('/')
```
blakestier: It's not easy to make a constant available (`admin`) so I preferred the helper.

e2e/cypress/integration/02_global_filter.spec.js
```diff
@@ -0,0 +1,76 @@
...
+  it('shows all projects for admin', () => {
+    cy.adminLogin('/')
+    cy.visit('/settings')
+    // hide modal unrelated to test flow
+    cy.get('app-welcome-modal').invoke('hide')
+
+    cy.get('chef-sidebar')
+      .should('have.attr', 'minor-version')
+      .then((version) => {
+        if (version === 'v1') {
```
blakestier: Should we/could use this conditional at the top of the describe to not run the tests (or setup) at all if we're not in 2.1? 🤔

e2e/cypress/integration/02_global_filter.spec.js
```diff
@@ -0,0 +1,76 @@
...
+        failOnStatusCode: false
+      }).then((response) => {
+        expect([200, 404]).to.include(response.status)
+      })
+    })
+  })
+
+  it('shows all projects for admin', () => {
+    cy.adminLogin('/')
+    cy.visit('/settings')
```
msorens: Change the above TWO lines to just this one:
```suggestion
    cy.adminLogin('/settings')
```
(Similarly for non-admin test)

e2e/cypress/integration/02_global_filter.spec.js
```diff
@@ -0,0 +1,76 @@
...
+    // hide modal unrelated to test flow
+    cy.get('app-welcome-modal').invoke('hide')
+
+    cy.get('chef-sidebar')
+      .should('have.attr', 'minor-version')
+      .then((version) => {
+        if (version === 'v1') {
+          cy.get('[data-cy=projects-filter-button]').click()
+
+          const allowedProjects = [proj1, proj2];
```
msorens: Need to also check the length of projects on the list; otherwise this is actually a phantom test!
(Because the checks on L69-70 would pass for admin or non-admin.)

e2e/cypress/integration/02_global_filter.spec.js
```diff
@@ -0,0 +1,76 @@
...
+  const pol_id = "cypress-policy"
+
+  before(() => {
+    cy.adminLogin('/').then(() => {
+      let admin = JSON.parse(localStorage.getItem('chef-automate-user'))
+
+      cy.createProject(admin.id_token, proj1)
+      cy.createProject(admin.id_token, proj2)
+      cy.createUser(admin.id_token, nonAdminUsername)
+      cy.createPolicy(admin.id_token, "cypress-policy-id", nonAdminUsername, [proj1, proj2])
```
msorens: Oops.
```suggestion
      cy.createPolicy(admin.id_token, pol_id, nonAdminUsername, [proj1, proj2])
```

e2e/cypress/integration/02_global_filter.spec.js
msorens: Definitely. 👍

e2e/cypress/integration/02_global_filter.spec.js
```diff
@@ -0,0 +1,76 @@
+describe('global projects filter', () => {
+  const nonAdminUsername = "cypressTestUserNonAdmin"
```
msorens: Test failed on my box until I fixed this:
```suggestion
  const nonAdminUsername = "cypresstestusernonadmin"
```
Local storage apparently converted it all to lower case.

e2e/cypress/support/commands.js
```diff
@@ -54,6 +60,16 @@ Cypress.Commands.add("login", (url) => {
   })
 })
 
+Cypress.Commands.add("logout", () => {
+  cy.get('[data-cy=user-profile-button]').click().then(() => {
+    cy.get('[data-cy=sign-out-button]').click().then(() => {
```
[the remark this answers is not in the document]
blakestier: Oh, good to know. I copied this from our login spec. . .

e2e/cypress/integration/02_global_filter.spec.js
blakestier: odd, I thought I fixed that 🤔

e2e/cypress/integration/02_global_filter.spec.js
blakestier: yes, I thought i fixed that, too . . . I must have not pushed something . . .

blakestier: yep 😳

e2e/cypress/support/commands.js
```diff
@@ -75,3 +91,76 @@ Cypress.Commands.add("cleanupUsers", (id_token) => {
...
+        }
+      ]
+    }
+  }).then((response) => {
+    expect([200, 409]).to.include(response.status)
+  })
+})
+
+Cypress.Commands.add("createProject", (id_token, id) => {
+  const name = "cypress project" + Cypress.moment().format('MMDDYYhhmm')
```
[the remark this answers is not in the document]
blakestier: (it doesn't)

PR #279, opened 2019-05-08
Update Automate UI to use users V2 model and V2 APIs when beta enabled

tylercloke:

Replaces all of the UI's internal model of user with the V2 version and keeps all logic around the V1 vs V2 model encapsulated in `user.requests.ts`. Also updates NGRX store to query the V2 endpoints when the IAM V2 is enabled.

- [ ] Code actually executed?

### DIFF
--- a/components/automate-ui/src/app/entities/users/user.effects.ts
+++ b/components/automate-ui/src/app/entities/users/user.effects.ts
@@ -1,10 +1,14 @@
-import { catchError, mergeMap, map } from 'rxjs/operators';
+import { catchError, mergeMap, map, filter } from 'rxjs/operators';
 import { Injectable } from '@angular/core';
 import { HttpErrorResponse } from '@angular/common/http';
+import { Store } from '@ngrx/store';
 import { Actions, Effect, ofType } from '@ngrx/effects';
-import { of } from 'rxjs';
+import { of, combineLatest } from 'rxjs';
+import { identity } from 'lodash/fp';
 
+import { NgrxStateAtom } from 'app/ngrx.reducers';
 import { CreateNotification } from '../notifications/notification.actions';
+import { iamMajorVersion } from 'app/entities/policies/policy.selectors';
 import { Type } from '../notifications/notification.model';
 import {
   CreateUser,
@@ -35,17 +39,20 @@ import { SelfUser } from './userself.model';
 export class UserEffects {
   constructor(
     private actions$: Actions,
-    private requests: UserRequests
+    private requests: UserRequests,
+    private store$: Store<NgrxStateAtom>
   ) { }
 
   // 7/6/18: TODO in follow-up PR: fetch teams for all users - bd
   @Effect()
-  getUsers$ = this.actions$.pipe(
-    ofType(UserActionTypes.GET_ALL),
-    mergeMap((_action: GetUsers) =>
-      this.requests.getUsers().pipe(
-        map((resp: GetUsersSuccessPayload) => new GetUsersSuccess(resp)),
-        catchError((error: HttpErrorResponse) => of(new GetUsersFailure(error))))));
+  getUsers$ = combineLatest(
+    this.actions$.pipe(ofType<GetUsers>(UserActionTypes.GET_ALL)),
+    this.store$.select(iamMajorVersion).pipe(filter(identity)))
+    .pipe(
+      mergeMap(([_action, version]) =>
+        this.requests.getUsers(version).pipe(
+          map((resp: GetUsersSuccessPayload) => new GetUsersSuccess(resp)),
+          catchError((error: HttpErrorResponse) => of(new GetUsersFailure(error))))));
 
   @Effect()
   getUsersFailure$ = this.actions$.pipe(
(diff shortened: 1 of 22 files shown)


components/automate-ui/src/app/entities/users/user.effects.ts
```diff
@@ -35,17 +39,20 @@ import { SelfUser } from './userself.model';
 export class UserEffects {
   constructor(
     private actions$: Actions,
-    private requests: UserRequests
+    private requests: UserRequests,
+    private store$: Store<NgrxStateAtom>
   ) { }
 
   // 7/6/18: TODO in follow-up PR: fetch teams for all users - bd
```
bcmdarroch: i think we could drop this comment, don't think that's feature we need anymore

components/automate-ui/src/app/entities/users/user.effects.ts
```diff
@@ -58,18 +65,21 @@ export class UserEffects {
       });
     }));
 
+  // TODO rename
```
bcmdarroch: is this needed anymore?

tylercloke: Fixed both issues in follow up commits.

components/automate-ui/src/app/entities/users/user.model.ts
```diff
@@ -1,7 +1,8 @@
 export interface User {
   id: string;
   name: string;
-  username: string;
+  membership_id: string;
```
bcmdarroch: We need some further changes here so we can still handle `v1` users, otherwise the Users UI breaks when you reset to v1. 
Since v1 and v2 users use different unique identifiers (v1 uses a GUID, v2 uses string id), we have to convert all returned users into one standard user interface that includes both identifiers.

Here's how we did it with teams:
- two interfaces, one for v1, one for v2: https://github.com/chef/automate/blob/master/components/automate-ui/src/app/entities/teams/team.model.ts
- `versionize` function to convert v1 users to v2 interface
https://github.com/chef/automate/blob/master/components/automate-ui/src/app/entities/teams/team.effects.ts#L63
- versionize is called anywhere we get user objects back from the API
https://github.com/chef/automate/blob/master/components/automate-ui/src/app/entities/teams/team.requests.ts#L23

tylercloke: Hmmm I was trying to avoid V1 details anywhere besides requests. That's a ton of functionality to support a super specific situation. What if we just do something similar to what we are going to do with the project filters: cache the current version, and if it changes between v1 and v2 or vise versa, clear the store and reload? Seems like a much cleaner solution from a code perspective for something we only ever expect to happen once or twice in the lifetime of a customer's install.

tylercloke: Oh also I updated everywhere in the app to use v2 identifiers, even if it's on V1. That's only "breaking" in the sense that the URL changes, but that's a change that's going to have to happen eventually regardless and a change we've already made in a few spots if memory serves.

tylercloke: I updated the whole app to use the V2 user model. It's a "breaking" change in the sense that all the user UI routes will now use ID instead of the GUID, but to get the new code, they need to reload the app, so all the links within the app with be correct.

There are actually issues with resetting for teams. If you are on a teams details page and you reset to v1, you'll get this error:

Screen Shot 2019-05-08 at 1 21 44 PM

You also get that error if you are on the list view, reset, then click a team, because the IDs are different between v1 and v2.

This keeps all those details internal to `requests` and keeps the model consistent between v1 an v2. I'm not seeing any issues for users on this branch and the above exists on master.

bcmdarroch: ah 💡so we can just switch the user ID since we're getting the same information from the users API whether we're on v1 or v2. but teams come from different tables depending on the version so we have different data

components/automate-ui/src/app/page-components/user-team-membership-table/user-team-membership-table.component.ts
```diff
@@ -42,7 +42,7 @@ export class UserTeamMembershipTableComponent implements OnInit, OnDestroy {
           // sort by name then by username
```
msorens: s/username/id/

components/automate-ui/src/app/entities/users/user.effects.ts
```diff
@@ -35,17 +39,20 @@ import { SelfUser } from './userself.model';
...
   @Effect()
-  getUsers$ = this.actions$.pipe(
-    ofType(UserActionTypes.GET_ALL),
-    mergeMap((_action: GetUsers) =>
-      this.requests.getUsers().pipe(
-        map((resp: GetUsersSuccessPayload) => new GetUsersSuccess(resp)),
-        catchError((error: HttpErrorResponse) => of(new GetUsersFailure(error))))));
+  getUsers$ = combineLatest(
+    this.actions$.pipe(ofType<GetUsers>(UserActionTypes.GET_ALL)),
+    this.store$.select(iamMajorVersion).pipe(filter(identity)))
```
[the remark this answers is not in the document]
tylercloke: I'll try it without it

components/automate-ui/src/app/entities/users/user.requests.ts
```diff
@@ -13,35 +14,159 @@ export interface UsersResponse {
...
+  username: string;
+  password?: string; // only used for updating, never returned by the API
+}
+
+function convertUserV1ToUser(user: UserV1): User {
+  return {
+    id: user.username,
+    name: user.name,
+    membership_id: user.id,
+    projects: [] // TODO add real projects
```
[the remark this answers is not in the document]
tylercloke: ...right probably not. I'll remove it for now.

PR #289, opened 2019-05-08
Project filter GetRuns

lancewf:

Add project filtering for the cfgmgmt.proto GetRuns RPC.

This PR does change the behavior of the GetRuns RPC in that before if the node from the passed in NodeID an error with "Invalid ID" was returned. Now, no error is returned and an empty run list is returned. 

- [ ] Code actually executed?

### DIFF
--- a/components/config-mgmt-service/backend/client.go
+++ b/components/config-mgmt-service/backend/client.go
@@ -17,8 +17,8 @@ import (
 // use in the future. Think about it as a set of contracts that the API
 // requires to fulfill the requests
 type Client interface {
-	// @params (nodeid)
-	GetNode(string) (Node, error)
+	// @params (nodeid, filters)
+	NodeExists(nodeID string, projectFilters map[string][]string) (bool, error)
 	// @params (page, perPage, sortField, ascending, filters)
 	GetNodes(int, int, string, bool, map[string][]string) ([]Node, error)
 	// @params (filters)
--- a/components/config-mgmt-service/backend/elastic/nodes.go
+++ b/components/config-mgmt-service/backend/elastic/nodes.go
@@ -24,13 +24,10 @@ func init() {
 	}
 }
 
-// GetNode Returns a single Node
-func (es Backend) GetNode(id string) (backend.Node, error) {
-	var node backend.Node
-	filters := map[string][]string{
-		"exists":      []string{"true"},
-		"entity_uuid": []string{id},
-	}
+// NodeExists - Does a node with entity_uuid == 'nodeID' exists
+func (es Backend) NodeExists(nodeID string, filters map[string][]string) (bool, error) {
+	filters["exists"] = []string{"true"}
+	filters["entity_uuid"] = []string{nodeID}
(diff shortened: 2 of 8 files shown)


components/config-mgmt-service/backend/client.go
```diff
@@ -17,8 +17,8 @@ import (
 // use in the future. Think about it as a set of contracts that the API
 // requires to fulfill the requests
 type Client interface {
-	// @params (nodeid)
-	GetNode(string) (Node, error)
+	// @params (nodeid, filters)
+	NodeExists(nodeID string, projectFilters map[string][]string) (bool, error)
```
lancewf: The `GetNode` function was only being used by the `GetRuns` function to see if the node existed.

components/config-mgmt-service/grpcserver/nodes.go
```diff
@@ -48,6 +48,13 @@ func (s *CfgMgmtServer) GetRuns(
 		return runs, status.Errorf(codes.InvalidArgument, "Invalid start/end time. (format: YYYY-MM-DD)")
 	}
 
+	projectFilters, err := filterByProjects(ctx, map[string][]string{})
+	if err != nil {
+		return runs, status.Errorf(codes.Internal, err.Error())
+	}
+
+	nodeExistsChan := s.nodeExistsAsync(request.GetNodeId(), projectFilters)
```
lancewf: Running the node exists check concurrently with retrieving the runs.

components/config-mgmt-service/integration_test/runs_test.go
```diff
@@ -83,7 +83,15 @@ func TestRunsWithFilterReturnsListWithOneNode(t *testing.T) {
...
 		ctx    = context.Background()
+		node   = iBackend.Node{
+			Exists: true,
+			NodeInfo: iBackend.NodeInfo{
+				EntityUuid: nodeID,
+			},
+		}
 	)
+
+	suite.IngestNodes([]iBackend.Node{node})
```
lancewf: For some existing tests, an associated node had to be ingested.

components/config-mgmt-service/backend/elastic/nodes.go
```diff
@@ -24,38 +24,25 @@ func init() {
...
-			}).WithError(err).Debug("Unable to unmarshal the node object")
-		}
-	} else {
-		err = errors.New(errors.NodeNotFound, "Invalid ID")
+	if searchResult.Hits.TotalHits == 0 {
+		return false, nil
 	}
 
-	return node, err
+	return true, err
```
phiggins: `err` is checked a few lines up, so it will always be nil here, right? You could replace this last conditional and returns with one return statement like `return searchResult.Hits.TotalHits == 0, nil`

components/config-mgmt-service/backend/mock/mock_test.go
```diff
@@ -26,9 +26,9 @@ func TestNew(t *testing.T) {
 
 func TestGetNode(t *testing.T) {
```
phiggins: This function name is out of date now.

PR #290, opened 2019-05-08
[dev-docs] Fix typo

stevendanna:

### DIFF
--- a/dev-docs/DEV_ENVIRONMENT.md
+++ b/dev-docs/DEV_ENVIRONMENT.md
@@ -244,6 +244,10 @@ data. Chef employees have access to this data via an internal instance
 of Vault. To get started, the `scripts/get_secrets` script will
 download the mostly commonly needed secrets:
 
+1. Connect to the Chef VPN.
+
+2. Run the `get_secrets` script as below:
+
 ``` console
 CHEF_USERNAME=your-chef-ad-username scripts/get_secrets
 source dev/secrets-env.sh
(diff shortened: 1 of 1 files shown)


stevendanna:
> Should be obvious, but might want to mention one needs to be connected to the VPN. (I forget once in a while.)

Yup, let's put this in the docs.

> Should mention how to install vault in hab studio. (??)

When inside the studio, vault is installed for you:

https://github.com/chef/automate/blob/master/scripts/get_secrets#L32

If you don't see that happening, that is a bug that we should fix.

> Should mention how to install vault outside of hab studio (brew install vault).

I link to the official install instructions here:

https://github.com/chef/automate/blob/master/scripts/get_secrets#L28

> When I run vault kv list 'secrets/a2' outside the studio I get an error, so I suspect permissions needed for us non-super users...?

This is the isssue fixed in this PR.  The correct path is `secret/a2`

msorens:
Thanks, @stevendanna.  From your response, I realized what I need to clarify.
If I run `get_secrets` in the studio then, yes, vault is installed for me, so no bug. But that's not the scenario here. Here's the user story: _As a reader of dev_environment.md, I expect what it tells me to be true._

1. dev_environment.md says something about `get_secrets`. Fine. Noted. _But: not executed._ Next I read about running `vault` directly and I need to install it. Hence my comment about how to install vault in the studio. Reading the `get_secrets` script answers the question, but the reader of dev_environment.md should not be expected to know that. Probably does not make sense to duplicate a lot of what is in `get_secrets`.

2. I mentioned you should also tell how to install vault outside the studio and you referenced that was in `get_secrets`. See point 1. :grin:. How does the user know to look there?

3. For the actual error from running vault, I pasted the wrong output. I get this error when using the correct path `secret/a2` (outside the studio):
```
Error making API request.
URL: GET https://vault.chef.co:8200/v1/sys/internal/ui/mounts/secret/a2
Code: 500. Errors:
* missing client token
```
Perhaps that would succeed if I had run `get_secrets` first? Dunno.

Given all of the above, I do not think it makes sense to repeat a lot of details mentioned in `get_secrets`. But I would suggest something along the lines of one of the following:
"Please run get_secrets before running vault directly, so that it sets up your env correctly."
"Please read get_secrets for further details on setting up vault."

PR #292, opened 2019-05-08
Fix unit test bug in applications-service

msorens:

The unit test in question (TestBuildWhereConstraintsFromFiltersMatrix) was, as @phiggins divined, implicitly depending on the order of a hash which of course means it will fail... unless it happens to succeed. This removes that fun element of non-determinism.

### :+1: Definition of Done

Test always passes.

- [ ] Code actually executed?

### DIFF
--- a/components/applications-service/pkg/storage/postgres/service_internal_test.go
+++ b/components/applications-service/pkg/storage/postgres/service_internal_test.go
@@ -1,6 +1,7 @@
 package postgres
 
 import (
+	"regexp"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


afiune:
Curious to know if we call **"a test that fails intermittently"** a bug? 🤔

msorens:
Philosophically, I do lean towards calling it a bug, but I guess that's a personal preference. 😁

msorens:
Also, just for posterity, my response from Slack to Salim's comment above: 
msorens [4:51 PM]
I always lean towards the least amount of code to get the thing done. In this case, the array element that is an empty string is quite harmless, so I left it.

PR #294, opened 2019-05-09
process node: add run_data and scan_data to the db and process_node logic

vjeffrey:
This pull request adds the run_data and scan_data fields to the process node logic.
This will be used in support of https://github.com/chef/automate/issues/192
On ingestion of a compliance report (node report), we will send information about the report status, end_time, and report id to ProcessNode.  This will allow us to easily show users which nodes just started failing as well as give us the ability to link them directly to the report they care about.

### :+1: Definition of Done
Run data and scan data are part of the process node logic and stored in db.

send some nodes in
rebuild nodemanager and compliance
send more nodes in
check the db: note the scan_data on the newest nodes


components/nodemanager-service/pgdb/node-managers-process-node.go
```diff
@@ -86,9 +168,47 @@ func (db *DB) ProcessIncomingNode(node *manager.NodeMetadata) error {
 		}
 		return tx.tagNode(node.GetUuid(), tags)
 	})
+
+	return err
+}
+
+func (db *DB) getCurrentScanStatus(node *manager.NodeMetadata) (string, error) {
```
vjeffrey: i played around with the idea of doing away with this function and doing it all in postgres, but it didn't work well for me, as we overwrite the data later when we insert the run_data or scan_data. 
But if ppl know of a way to do this and not run into that problem, i'm all ears!

components/nodemanager-service/pgdb/node-managers-process-node.go
```diff
@@ -86,9 +168,47 @@ func (db *DB) ProcessIncomingNode(node *manager.NodeMetadata) error {
...
+	case err == sql.ErrNoRows:
+		return "", nil
+	case err != nil:
+		return "", errors.Wrap(err, "unable to query for current status")
+
+	}
+	if !status.Valid {
+		return "", nil
+	}
+	statusString, err := strconv.Unquote(status.String)
```
vjeffrey: is there something else i should be using here instead of Unquote?

components/compliance-service/api/tests/test_ingest_to_mgr_conn.rb
```diff
@@ -20,6 +20,12 @@
     nodes_list = MANAGER_GRPC nodes, :list, Nodes::Query.new()
     assert_equal(nodes_list.total, 9)
 
+    nodes_list['nodes'].each { |node|
+      assert_equal(node['scan_data'].id.length > 0, true)
+      assert_equal(node['scan_data'].status.length > 0, true)
+      assert_equal(node['scan_data'].end_time.length > 0, true)
+    }
```
phiggins: If this ever failed it'd give a pretty opaque error message. If you rewrote it a bit it could give a lot more information about what went wrong:

```
Petes-MBP:tmp pete$ cat test.rb
require "minitest/autorun"
require "ostruct"

def nodes_list
  {
    "nodes" => [
      { "scan_data" => OpenStruct.new(id: "abcd", status: "good", end_time: "a long time ago") },
      { "scan_data" => OpenStruct.new(id: "12345", status: "bad", end_time: "") },
    ]
  }
end

describe "nodes" do
  it "has an opaque error message" do
    nodes_list["nodes"].each { |node|
      assert_equal(node["scan_data"].id.length > 0, true)
      assert_equal(node["scan_data"].status.length > 0, true)
      assert_equal(node["scan_data"].end_time.length > 0, true)
    }
  end

  it "has a clearer error message" do
    assert nodes_list["nodes"].all? { |node|
      node["scan_data"].id.length > 0 &&
        node["scan_data"].status.length > 0 &&
        node["scan_data"].end_time.length > 0
    }, "Nodes did not have necessary scan_data: #{nodes_list}"
  end
end
```

Output:
```
Petes-MBP:tmp pete$ ruby test.rb
Run options: --seed 24770

# Running:

FF

Finished in 0.000998s, 2004.0087 runs/s, 7014.0303 assertions/s.

  1) Failure:
nodes#test_0001_has an opaque error message [test.rb:18]:
Expected: false
  Actual: true

  2) Failure:
nodes#test_0002_has a clearer error message [test.rb:23]:
Nodes did not have necessary scan_data: {"nodes"=>[{"scan_data"=>#<OpenStruct id="abcd", status="good", end_time="a long time ago">}, {"scan_data"=>#<OpenStruct id="12345", status="bad", end_time="">}]}

2 runs, 7 assertions, 2 failures, 0 errors, 0 skips
```

components/nodemanager-service/api/nodes/nodes.proto
```diff
@@ -110,6 +110,15 @@ message Node {
...
 	repeated string projects = 28;
+	LastContactData run_data = 29;
+	LastContactData scan_data = 30;
+}
+
+message LastContactData {
+	string id = 1; // run id or report id
+	string status = 2;
+	string penultimate_status = 3;
+	string end_time = 4;
```
phiggins: Should this be the timestamp type?

components/compliance-service/ingest/server/compliance.go
```diff
@@ -197,14 +203,39 @@ func (s *ComplianceIngestServer) handleNotifications(ctx context.Context, report
...
+		// setting the status to skipped if they match.
+		if profile.Status == "skipped" {
+			skippedCounter++
+		}
+	}
+	if skippedCounter == len(profiles) {
+		status = "skipped"
+	}
+	return
+}
```
phiggins: There's a linter that complains about the named return value (or whatever this is called). We've turned it off for compliance-service but we'll probably want to enable it eventually.

components/nodemanager-service/api/nodes/nodes.proto
```diff
@@ -110,6 +110,15 @@ message Node {
...
 	string name_prefix = 27;
 	repeated string projects = 28;
+	LastContactData run_data = 29;
+	LastContactData scan_data = 30;
+}
+
+message LastContactData {
+	string id = 1; // run id or report id
+	string status = 2;
+	string penultimate_status = 3;
```
phiggins: If these are always one of a set of values, should it be an enum or similar?

vjeffrey: yes, good point -- status and penultimate status should be an enum of passed, failed, skipped.  and probably empty string as the default

components/nodemanager-service/api/nodes/nodes.proto
vjeffrey: yes

components/compliance-service/api/tests/test_ingest_to_mgr_conn.rb
vjeffrey: oooh nice, thank you!

components/compliance-service/ingest/server/compliance.go
```diff
@@ -197,14 +203,39 @@ func (s *ComplianceIngestServer) handleNotifications(ctx context.Context, report
...
 	return nil
 }
+
+func getReportStatus(profiles []*ingest_inspec.Profile) (status string) {
+	// start with a status of passed as the default status
+	status = "passed"
+	skippedCounter := 0
+	for _, profile := range profiles {
+		// if any profile is failed, report is failed
+		if profile.Status == "failed" {
```
vjeffrey: i just realizeed this logic if faulty. profiles will have a status of skipped if skipped for a platform exception, but will not otherwise have the status. need to look at the controls to get that

components/nodemanager-service/pgdb/node-managers-process-node.go
phiggins: What does the data look like before and after Unquote?

components/nodemanager-service/pgdb/migration/sql/04_nodes_add_fields.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE IF EXISTS nodes
+  ADD COLUMN IF NOT EXISTS last_scan JSON not null default '{}'::json,
+  ADD COLUMN IF NOT EXISTS last_run JSON not null default '{}'::json;
```
phiggins: It looks from some places in the code that these two are exclusive, ie, a node should only have last_scan or last_run, but not both. Is that true? If so this could probably be simplified by not using json for this and just having the fields be regular columns on the nodes table.

components/nodemanager-service/pgdb/node-managers-process-node.go
phiggins: What did you try, and what happened? I'd have to play around with it a bit but I think you should be able to access the current value for the field during the upsert and set the appropriate field with it.

components/nodemanager-service/pgdb/node-managers-process-node.go
```diff
@@ -86,9 +168,47 @@ func (db *DB) ProcessIncomingNode(node *manager.NodeMetadata) error {
...
 	if err != nil {
-		return err
+		return "", errors.Wrap(err, "unable to read status")
 	}
+	return statusString, nil
+}
```
phiggins: Unless I misunderstand this function and the one below should be able to use `db.SelectStr` to get the status in one line.

components/nodemanager-service/pgdb/migration/sql/04_nodes_add_fields.up.sql
vjeffrey: a node can have one, the other, or both - but a single report will only have one or the other

components/nodemanager-service/pgdb/node-managers-process-node.go
vjeffrey: I was grabbing the value of status, setting penultimate_status to that value, and then going on with the rest of it, but it seemed like it was getting overwritten.
i'll play around again today with it. seems like it should be doable and it would simplify this logic alot

components/nodemanager-service/pgdb/node-managers-process-node.go
vjeffrey: `\"\passed\"\` and then the correct "passed".  Unquote works well there; I just wasn't sure if there was something else we should use instead

components/nodemanager-service/pgdb/node-managers-process-node.go
vjeffrey: ill give it a try!

components/nodemanager-service/pgdb/migration/sql/04_nodes_add_fields.up.sql
vjeffrey: ^ given this, doing individual columns would mean 8 columns...vs 2 with json

components/nodemanager-service/pgdb/node-managers-process-node.go
vjeffrey: hmmm. ok, i've tried revisiting this. first i was trying to just set the value of penultimate status to current value of status. that was not working as the intent of json in postgres is to store blurbs of data, not really made for updating individual fields within that data well.
so then i thought i could make an extra couple columns for just penult status for scan and run. the problem is updating any of these fields from within the queries is super awkward (i can't get it to work right) b/c they are upserts...

components/nodemanager-service/pgdb/node-managers-process-node.go
vjeffrey: `	ProcessIncomingNode unable to get current scan status: unable to read status: sql: Scan error on column index 0, name "status": unsupported Scan, storing driver.Value type <nil> into type *string`
looks like selectstr isn't great here..

phiggins: Can you use `COALESCE` to always return a string?

components/compliance-service/ingest/server/compliance.go
vjeffrey: this logic is fixed now

components/nodemanager-service/pgdb/node-managers-process-node.go
```diff
@@ -64,19 +110,69 @@ func (db *DB) ProcessIncomingNode(node *manager.NodeMetadata) error {
 	// 2) it is already registered in our db with diff uuid, same source_id: update the node by source_id
 	// 3) it is not in our db, we must add it
 
+	lastContactDataByte := []byte{}
+	if node.GetScanData() != nil {
+		scanData, err := translateToDBStruct(node.GetScanData())
+		if err != nil {
+			return errors.Wrap(err, "ProcessIncomingNode unable to translate struct to db struct")
+		}
```
vjeffrey: can this logic be simplified ?

vjeffrey: i moved some of this to another function to try to clear things up a bit

components/compliance-service/ingest/server/compliance.go
```diff
@@ -167,7 +169,11 @@ func (s *ComplianceIngestServer) ProcessComplianceReport(ctx context.Context, in
 		SourceRegion:    in.SourceRegion,
 		SourceAccountId: in.SourceAccountId,
 		Tags:            in.Tags,
-	})
+		ScanData: &nodes.LastContactData{
+			Id:      in.ReportUuid,
+			EndTime: endTimeTimestamp,
```
lancewf: This change will make it so you don't have to pass the `profiles` to the `sendNodeInfoToManager` function. 
```suggestion
			EndTime: endTimeTimestamp,
			Status: getReportStatus(in.Profiles),
```

components/compliance-service/ingest/server/compliance.go
```diff
@@ -197,14 +203,90 @@ func (s *ComplianceIngestServer) handleNotifications(ctx context.Context, report
...
 	}
 	return nil
 }
+
+func getReportStatus(profiles []*ingest_inspec.Profile) nodes.LastContactData_Status {
+	// start with a status of passed as the default status
+	status := nodes.LastContactData_PASSED
+	skippedCounter := 0
+	for _, profile := range profiles {
+		profile.Status = getProfileStatus(profile)
```
lancewf: This line is confusing me. This is updating the profiles passed in statuses because they are pointers. Do we need to update the status on these profiles?
If we do, line 228 is breaking out of the for loop and will not update statuses after a failed one. Is this desired?

vjeffrey: if any profile status is failed, the report (node) has a failed status
if all are skipped, report (node) is skipped
otherwise passed

components/compliance-service/ingest/server/compliance.go
vjeffrey: ah, good point!
i had left that off to be done in the function b/c i (incorrectly) assumed that it was being done async.  but now i just noticed that that we check the error. 
I wonder if that function should be executed via go routine instead, to avoid slowing down the ingestion process?

components/compliance-service/ingest/server/compliance.go
lancewf: Could we change the `profile.Status` to use a local variable like below? 
```suggestion
		profileStatus := getProfileStatus(profile)
```

vjeffrey: 👍

components/nodemanager-service/pgdb/node_managers_process_node_integration_test.go
```diff
@@ -18,11 +18,13 @@ func (suite *NodeManagersAndNodesDBSuite) TestProcessIncomingNodeWithNoUUID() {
 		Name:            "123.798.324.32",
 		PlatformName:    "ubuntu",
 		PlatformRelease: "16.04",
-		JobUuid:         "12345-389244-2433",
 		LastContact:     nowTime,
 		SourceId:        "i-078973",
 		SourceRegion:    "eu-west-1",
-		SourceAccountId: "8799247840",
+		SourceAccountId: "999999999999",
```
lancewf: That is a lot of tests!! 😱

components/compliance-service/ingest/server/compliance.go
lancewf: I would say not to run it in a separate go routine at this point. I would leave that for later when we look to increase the speed of ingestion.

PR #295, opened 2019-05-09
[A2-784] UI http interceptor

msorens:

This sends selected projects along with every API request from the UI.

### :+1: Definition of Done

Projects appear in the HTTP request headers

image

Select projects from the global projects filter then navigate somewhere in the UI (having self-refresh is in the browser is a separate card).

- [x] Code actually executed?

### DIFF
--- a/components/authz-service/server/v2/authz.go
+++ b/components/authz-service/server/v2/authz.go
@@ -9,6 +9,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
+	"github.com/chef/automate/lib/grpc/auth_context"
 	"github.com/chef/automate/lib/logger"
 	"github.com/chef/automate/lib/stringutils"
 
@@ -139,6 +140,10 @@ func (s *authzServer) FilterAuthorizedPairs(
 func (s *authzServer) FilterAuthorizedProjects(
 	ctx context.Context,
 	req *api.FilterAuthorizedPairsReq) (*api.FilterAuthorizedProjectsResp, error) {
+
+	// Introspection needs unfiltered access.
+	ctx = auth_context.ContextWithoutProjects(ctx)
+
 	resp, err := s.engine.V2FilterAuthorizedProjects(ctx,
 		engine.Subjects(req.Subjects),
 		toEnginePairs(req.Pairs))
@@ -172,6 +177,9 @@ func isBeta2p1(version api.Version) bool {
 }
 
 func (s *authzServer) getAllProjects(ctx context.Context) ([]string, error) {
+	// Need unfiltered access to the projects list.
+	ctx = auth_context.ContextWithoutProjects(ctx)
+
 	// we make this extra call to cover the case when the following are true:
 	// - no project filter has been provided
 	// - one statement allows All Projects for the given resource/action
--- a/components/authz-service/server/v2/policy_refresher.go
+++ b/components/authz-service/server/v2/policy_refresher.go
@@ -9,8 +9,8 @@ import (
 
 	"github.com/chef/automate/components/authz-service/engine"
 	storage "github.com/chef/automate/components/authz-service/storage/v2"
+	"github.com/chef/automate/lib/grpc/auth_context"
 	"github.com/chef/automate/lib/logger"
-	"google.golang.org/grpc/metadata"
 )
 
 var ErrMessageBoxFull = errors.New("Message box full")
@@ -152,14 +152,8 @@ func (refresher *policyRefresher) RefreshAsync() error {
 
 // updates OPA engine store with policy
 func (refresher *policyRefresher) updateEngineStore(ctx context.Context) error {
-	// We need to remove project filters from the request context
-	// otherwise they will be applied for store updates.
-	// This will fail on service start context, so only remove projects if ok.
-	md, ok := metadata.FromIncomingContext(ctx)
-	if ok {
-		delete(md, "projects")
-		ctx = metadata.NewOutgoingContext(ctx, md)
-	}
+	// Engine updates need unfiltered access to all data.
+	ctx = auth_context.ContextWithoutProjects(ctx)
 
 	policyMap, err := refresher.getPolicyMap(ctx)
 	if err != nil {
--- a/components/automate-gateway/gateway/middleware/auth_interceptor_internal_test.go
+++ b/components/automate-gateway/gateway/middleware/auth_interceptor_internal_test.go
@@ -79,6 +79,10 @@ func TestGetProjectsFromMetadata(t *testing.T) {
 		input    []string
 		expected []string
 	}{
+		"no projects": {
+			input:    []string{""},
+			expected: []string{""},
+		},
 		"a single project": {
 			input:    []string{"pikachu"},
 			expected: []string{"pikachu"},
--- a/components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
+++ b/components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
@@ -1,26 +1,46 @@
-import { throwError as observableThrowError,  Observable } from 'rxjs';
-
-import { catchError } from 'rxjs/operators';
-import { Injectable } from '@angular/core';
 import { HttpEvent, HttpHandler, HttpInterceptor, HttpRequest } from '@angular/common/http';
+import { Injectable } from '@angular/core';
+import { Store } from '@ngrx/store';
+import { throwError as observableThrowError, Observable } from 'rxjs';
+import { catchError } from 'rxjs/operators';
 import { get } from 'lodash/fp';
-import { ChefSessionService } from '../chef-session/chef-session.service';
+
+import { NgrxStateAtom } from 'app/ngrx.reducers';
+import { ChefSessionService } from 'app/services/chef-session/chef-session.service';
+import * as selectors from 'app/services/projects-filter/projects-filter.selectors';
+import { ProjectsFilterOption } from '../projects-filter/projects-filter.reducer';
 
 @Injectable()
 export class HttpClientAuthInterceptor implements HttpInterceptor {
 
-  constructor(private chefSession: ChefSessionService) {}
+  private projects: string;
+
+  constructor(
+    private chefSession: ChefSessionService,
+    private store: Store<NgrxStateAtom>
+  ) {
+    this.store.select(selectors.options)
+      .subscribe((options: ProjectsFilterOption[]) => {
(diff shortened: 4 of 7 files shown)


components/automate-gateway/gateway/middleware/auth_interceptor_internal_test.go
```diff
@@ -79,6 +79,10 @@ func TestGetProjectsFromMetadata(t *testing.T) {
 		input    []string
 		expected []string
 	}{
+		"no projects": {
```
msorens: Drive-by

components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
```diff
@@ -1,26 +1,44 @@
...
+  ) {
+    this.store.select(selectors.options)
+      .subscribe((options: ProjectsFilterOption[]) => {
+        this.projects = options.filter(p => p.checked).map(p => p.value).join(', ');
+      });
+  }
 
   intercept(request: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
+    let headers = request.headers.set('Authorization', `Bearer ${this.chefSession.id_token}`);
+    if (this.projects) {
```
msorens: Important to only send when there are some; otherwise we are making an invalid request and get a 403, because all projects must have non-zero length.

components/automate-ui/src/app/services/projects-filter/projects-filter.service.ts
```diff
@@ -9,17 +11,17 @@ const STORE_OPTIONS_KEY = 'projectsFilter.options';
 
 @Injectable()
 export class ProjectsFilterService {
```
msorens: All drive-by: add some types.

components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
```diff
@@ -1,26 +1,44 @@
...
-  constructor(private chefSession: ChefSessionService) {}
+  private projects: string;
+
+  constructor(
+    private chefSession: ChefSessionService,
+    private store: Store<NgrxStateAtom>
+  ) {
+    this.store.select(selectors.options)
+      .subscribe((options: ProjectsFilterOption[]) => {
+        this.projects = options.filter(p => p.checked).map(p => p.value).join(', ');
```
tylercloke: Will this work if the store returns undefined?

blakestier:
>Think we'll run into tricky business with the interceptor approach of injecting the project headers if we ever need to opt out of them, but we can cross that bridge when we get there :)

I think we're there-- the interceptor is adding the projects to the `introspect_projects` call (in lieu of empty `[]`)

We're still testing but we'll at least need some modifications

components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
```diff
@@ -1,26 +1,44 @@
...
+      });
+  }
 
   intercept(request: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
+    let headers = request.headers.set('Authorization', `Bearer ${this.chefSession.id_token}`);
+    if (this.projects) {
+      headers = headers.set('projects', this.projects);
```
bcmdarroch: one idea @scottopherson gave me: 
```
if (thisEndpointDoesntNeedOrSupportFilteringByProjects) {
      return next.handle(req.clone());
    }
```

we could write the helper `thisEndpointDoesntNeedOrSupportFilteringByProjects` and keep track of all the APIs that shouldn't be filtered inside it

msorens:
@blakestier Latest commit addresses your changes.
My preference is that any code that can, should use the new library call I added, `ContextWithoutProjects`. It makes it very clear that a given block of code wants to operate unfettered by project filtering, and that knowledge is all localized to the code needing it.
It seems a lot cleaner than having to hard-code a list of URLs that the UI knows about for stripping projects from the header, not to mention that how are our users going to know that they should not provide a projects header for a given API call?

components/authz-service/server/v2/policy_refresher.go
```diff
@@ -152,14 +152,8 @@ func (refresher *policyRefresher) RefreshAsync() error {
...
 func (refresher *policyRefresher) updateEngineStore(ctx context.Context) error {
-	// We need to remove project filters from the request context
-	// otherwise they will be applied for store updates.
-	// This will fail on service start context, so only remove projects if ok.
-	md, ok := metadata.FromIncomingContext(ctx)
-	if ok {
-		delete(md, "projects")
-		ctx = metadata.NewOutgoingContext(ctx, md)
-	}
+	// Engine updates need unfiltered access to all data.
```
srenatus: This comment is a bit confusing to me... We're inside Authz-service, does it even matter if this is set or not? Might be missing something 💭

lib/grpc/auth_context/auth_context.go
```diff
@@ -78,6 +78,18 @@ func NewOutgoingProjectsContext(ctx context.Context) context.Context {
 	return metadata.NewOutgoingContext(ctx, md)
 }
 
+// ContextWithoutProjects removes previously added projects from the GRPC metadata
+// for those system operations that must not be filtered by projects.
+func ContextWithoutProjects(ctx context.Context) context.Context {
```
srenatus: One relevant detail is missing from comment and name: the translation from incoming to outgoing metadata. Can we fix this? 😉

components/authz-service/server/v2/policy_refresher.go
blakestier: We aren't using the `ctx` to filter projects within the `V2SetPolicies` call so we can remove all of this (not sure why we had it before 🤷‍♂ )

components/authz-service/server/v2/authz.go
```diff
@@ -139,6 +140,10 @@ func (s *authzServer) FilterAuthorizedPairs(
 func (s *authzServer) FilterAuthorizedProjects(
 	ctx context.Context,
 	req *api.FilterAuthorizedPairsReq) (*api.FilterAuthorizedProjectsResp, error) {
+
+	// Introspection needs unfiltered access.
+	ctx = auth_context.ContextWithoutProjects(ctx)
```
blakestier: Consider moving this to just after 155 since that's where we use it (I got confused and thought `V2FilterAuthorizedProjects`

components/authz-service/server/v2/policy_refresher.go
```diff
@@ -152,14 +152,8 @@ func (refresher *policyRefresher) RefreshAsync() error {
...
+	ctx = auth_context.ContextWithoutProjects(ctx)
 
 	policyMap, err := refresher.getPolicyMap(ctx)
```
blakestier: Hmm we might be doing it here . . .

blakestier: We are! :bushes: :ignoreme:

components/authz-service/server/v2/policy_refresher.go
msorens: It is needed, because that method proceeds to get policies, roles, and rules from the store--and those would be filtered without this bit.

lib/grpc/auth_context/auth_context.go
msorens: Sure... but I am not sure on what is what; please suggest how to make it better!

components/authz-service/server/v2/policy_refresher.go
srenatus: ... but what's needed to avoid that is removing it from the incoming Metadata. Outgoing metadata is irrelevant here, since there's no grpc client calls involved in this code path, or is there?

lib/grpc/auth_context/auth_context.go
srenatus: See above, I wonder if this helper is a good idea as is. The only point where we need this incoming/outgoing translation is the gateway.

srenatus: Well, it's got a use for sure. Just not inside authz-service. 😉

components/authz-service/server/v2/authz.go
blakestier: 🤔 Do we need this in `FilterAuthorizedProjects` at all?  I wonder if we really want to push it into `getAllowedProjects` (which calls `FilterAuthorizedProjects`) and is what `IntrospectAllProjects` calls.

components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
scottopherson: Stores have an initial value so they should never be undefined unless a reducer sets a value to be so. `selectors.options` starts off as an empty array.

components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
scottopherson: Didn't see this PR until this morning. I'd also recommend having separate `Auth` and `ProjectsFilter`  interceptors in the request chain, similar to the branch I was playing with last night:

https://github.com/chef/automate/compare/sc/projects-filter-data#diff-673f4abc0b622315151c256ba3238a8b
https://github.com/chef/automate/compare/sc/projects-filter-data#diff-b99ce0ca235bf0716d691eab97a28bf5R311

components/authz-service/server/v2/authz.go
blakestier: (or `IntrospectAllProjects` itself)

blakestier: I guess it's all the same flow so 🤷‍♂ nvmd

components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
scottopherson:
```
@Injectable()
export class ProjectsFilterInterceptor implements HttpInterceptor {
  constructor(private projectsFilter: ProjectsFilterService) {}

  intercept(request: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {

    if (thisEndpointDoesntNeedOrSupportFilteringByProjects) {
      // pass the request through
      return next.handle(req.clone());
    }

    return this.projectsFilter.selectionValue$.pipe(
      first(),
      flatMap(selectionValue => {
        const headers = request.headers.set('projects', selectionValue);

        if (selectionValue === ProjectsConstants.ALL_RESOURCES_VALUE) {
          headers.delete('projects');
        }

        return next.handle(request.clone({ headers }));
      })
    );
  }
}
```

PR #304, opened 2019-05-09
[A2-655] Add user V2 API integration testing

tylercloke:

Fills in some missing base level integration testing for IAM V2 users endpoints.

- [ ] Code actually executed?

### DIFF
--- a/inspec/a2-api-integration/controls/iam_v2.rb
+++ b/inspec/a2-api-integration/controls/iam_v2.rb
@@ -18,6 +18,7 @@ control 'iam-v2-1' do
   CUSTOM_TOKEN_ID = 'inspec-token'
   CUSTOM_TEAM_ID = 'inspec-team'
   CUSTOM_USER_ID = 'inspec-user'
+  ADMIN_USER_ID = 'admin'
 
   describe 'v2beta policy API' do
     before(:all) do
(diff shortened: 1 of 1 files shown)


inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -427,63 +428,207 @@
...
+      end
+
+      describe "GET /iam/v2beta/users/" do
+        it "returns the list of users" do
+          resp = automate_api_request("/apis/iam/v2beta/users")
+          expect(resp.http_status).to eq 200
+          expect(resp.parsed_response_body[:users].length).to eq 3
+          expect(resp.parsed_response_body[:users][0][:id]).to match(/#{TEST_USER_ID_2}|#{CUSTOM_USER_ID}|#{ADMIN_USER_ID}/)
+          expect(resp.parsed_response_body[:users][1][:id]).to match(/#{TEST_USER_ID_2}|#{CUSTOM_USER_ID}|#{ADMIN_USER_ID}/)
+          expect(resp.parsed_response_body[:users][2][:id]).to match(/#{TEST_USER_ID_2}|#{CUSTOM_USER_ID}|#{ADMIN_USER_ID}/)
```
msorens: That is not as strong as it could be; it would still pass if all three users in the response were the same user.
Here is one way to ensure that the response really contains all 3 users:
```
idList = resp.parsed_response_body[:users].map { |user| user.id }
expect((idList & [TEST_USER_ID_2, CUSTOM_USER_ID, ADMIN_USER]).size).to eq idList.size
```
Or, without introducing a temp variable:
```
resp.parsed_response_body[:users].map { |u| u.id }.all? 
    { |user| [TEST_USER_ID_2, CUSTOM_USER_ID, ADMIN_USER].include? user }
```

inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -427,63 +428,207 @@
...
-        http_method: 'PUT',
-        request_body: {
-          name: updatedName,
-          password: "newpassword",
-          previous_password: "wrongagain"
-        }.to_json
-      )
-      expect(resp.http_status).to eq 400
+    TEST_USER = {
+        id: CUSTOM_USER_ID,
```
msorens: Consider making this TEST_USER_ID to be more consistent.

inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -427,63 +428,207 @@
...
+    end
+
+    describe "when a user exists" do
+      before(:each) do
+        resp = automate_api_request("/apis/iam/v2beta/users",
+         http_method: 'POST',
+         request_body: TEST_USER.to_json
+       )
+       expect(resp.http_status).to eq 200
+     end
```
msorens: I should think it is more realistic to start with multiple users (i.e. your list of 3 users) rather than just a single user in the DB. That makes the test stronger because rather than getting the right data because it is the _only_ one, instead it is because it is the _correct_ one.

inspec/a2-api-integration/controls/iam_v2.rb
blakestier: I also think `match_array` might work, here. 

`expect(resp.parsed_response_body[:users].map(&:id)).to match_array([TEST_USER_ID_2, CUSTOM_USER_ID, ADMIN_USER])` 

sometimes that `&:id` helper is weird around json though

inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -427,63 +428,207 @@
...
+          )
+          expect(resp.http_status).to eq 200
+          expect(resp.parsed_response_body[:user][:name]).to eq createdName
+          expect(resp.parsed_response_body[:user][:id]).to eq CREATED_ID
+        end
+      end
+    end
+
+    describe "when multiple users exists" do
+      TEST_USER_ID_2 = 'inspec-user-2'
```
blakestier: I know using the constant is an established pattern, but I'm not clear on why we need it. I think a local variable would work fine.

inspec/a2-api-integration/controls/iam_v2.rb
blakestier: (the one @msorens suggested, iterating over users, is better if we expect that there will be other users lurking around that our tests haven't created, which, sadly, has been true for some other things . . . tokens, I think)

inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -427,63 +428,207 @@
...
+            request_body: {
+              name: updatedName,
+              password: "newpassword",
+              previous_password: "wrongagain"
+            }.to_json
+          )
+          expect(resp.http_status).to eq 400
+        end
+
+        # TODO (tc): right now, it returns 400 for a non-existent user because
```
blakestier: what is the `TODO`? To return a `404`?

inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -427,63 +428,207 @@
...
+          )
+          expect(resp.http_status).to eq 200
+          expect(resp.parsed_response_body[:user][:name]).to eq updatedName
+        end
+      end
+
+      describe "DELETE /iam/v2beta/users/:id" do
+        it "deletes the user if it exists" do
+          resp = automate_api_request("/apis/iam/v2beta/users/#{CUSTOM_USER_ID}", http_method: 'DELETE')
+          expect(resp.http_status).to eq 200
```
blakestier: For testing `DELETE`, first verifying that the user exists, deleting, then checking that it's gone would be more robust, here. 

Recently we were getting `200s` even when the object wasn't being deleted.

inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -427,63 +428,207 @@
...
+        end
+
+        it "user gets a 404 when the user does not exist" do
+          resp = automate_api_request("/apis/iam/v2beta/users/some_wrong_id", http_method: 'DELETE')
+          expect(resp.http_status).to eq 404
+        end
+      end
+
+      describe "PUT /iam/v2beta/users/:id" do
```
blakestier: Since these test cases are no longer about the user updating itself, could we 

1) Put them in a describe block that clarifies that 
2) Change the `updatedName` text (which makes the reader think we're still in the `self` block)?

inspec/a2-api-integration/controls/iam_v2.rb
msorens: As long as we keep the length test in L497 then @blakestier 's is better--use the test helpers when available.

inspec/a2-api-integration/controls/iam_v2.rb
tylercloke: Yeah, the API is wrong.

inspec/a2-api-integration/controls/iam_v2.rb
tylercloke: I'm asserting that it's a 200 on create in the before block but I can add that.

inspec/a2-api-integration/controls/iam_v2.rb
blakestier: Can we update the comment?

inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -461,12 +461,14 @@
     end
 
     describe "when multiple users exists" do
-      TEST_USER_ID_2 = 'inspec-user-2'
-      TEST_USER_2 = {
-        id: TEST_USER_ID_2,
-        name: "display name 2",
-        password: "chefautomate"
-      }
+      let (:custom_user_id_2) { 'inspec-user-2' }
```
blakestier: :D :D :D

inspec/a2-api-integration/controls/iam_v2.rb
blakestier: Yeah it's that we were getting a `200` for a DELETE when the delete wasn't actually deleting-- that's why the `GET` check afterward helps also, since we're not using the db to insert. (I don't see the `GET` check yet)

PR #305, opened 2019-05-10
improve error handling for remote execution nodes

vjeffrey:
the commit msgs probably describe this best.
this includes a couple diff fixes, but the primary intent was to improve error handling for ssm nodes and ensure that we are adding the name information to nodes if available when polling for nodes.

### :+1: Definition of Done
better error handling for ssm nodes

Screen Shot 2019-05-09 at 16 00 00
Screen Shot 2019-05-09 at 17 49 29

rebuild ui, compliance, nodemanager
add an azure-vm nodemanager
patch the config so it has this:
```
[nodemanager.v1.sys.nodemanager]
azurevm_polling_interval = 2
```
to ensure you're polling for nodes from azure every 2 minutes
delete all the nodes associated with the azure-vm mgr (you can do this from mgr details page)
wait for them to be rediscovered: should see the same info as previous
if some are unreachable, check error. it should not be blank.
run a scan job.
check errors. they should not be blank.


components/compliance-service/inspec-agent/resolver/resolver.go
```diff
@@ -452,7 +452,7 @@ func (r *Resolver) handleManagerNodes(ctx context.Context, m *manager.NodeManage
 				// then let's prioritize that -- otherwise try ssm
 				if len(credsArr) == 0 {
 					var skip bool
-					skip, ssmJob = handleSSMNodes(node, job, &backend)
+					ssmJob, skip = handleSSMNodes(node, job, &backend)
```
alexpop: good catch!

components/nodemanager-service/managers/azure/azure.go
```diff
@@ -759,11 +761,10 @@ func (creds *Creds) SendRunCommandJob(ctx context.Context, job *types.InspecJob,
 	client.Authorizer = getAuthorizer(creds.Token)
 	future, err := sendCommand(ctx, job, script, scriptType, client)
 	if err != nil {
-		logrus.Errorf("unable to run scan: %s %s %s", job.NodeName, job.MachineIdentifier, err.Error())
 		*job.NodeStatus = types.StatusFailed
-		return err
+		return errors.Wrap(err, fmt.Sprintf("unable to run scan: %s %s", job.NodeName, job.MachineIdentifier))
 	}
-	err = future.WaitForCompletion(ctx, client.Client)
+	err = future.WaitForCompletionRef(ctx, client.Client)
```
alexpop: Makes sense 👍 
```
// Deprecated: Please use WaitForCompletionRef() instead
```

components/nodemanager-service/pgdb/node_managers_cloud_integration_test.go
```diff
@@ -122,6 +122,32 @@ func (suite *NodeManagersAndNodesDBSuite) TestUpdateOrInsertInstanceSourceStateI
...
+		Key:    "manager_id",
+		Values: []string{mgrID},
+	}
+	mgrNodes, _, err := suite.Database.GetNodes("", nodes.Query_ASC, 1, 100, []*common.Filter{filter})
+	if err != nil {
+		suite.FailNow(err.Error())
+	}
+	suite.Equal(1, len(mgrNodes))
+
+	node := mgrNodes[0]
```
phiggins: If the assertion above fails (ie, if `len(mgrNodes) == 0`) this line will panic with a null pointer exception. If line 144 is changed to use `suite.Require().Equal(...)` the test will stop on that line instead.

components/nodemanager-service/pgdb/node_managers_cloud_integration_test.go
```diff
@@ -122,6 +122,32 @@ func (suite *NodeManagersAndNodesDBSuite) TestUpdateOrInsertInstanceSourceStateI
...
+	suite.Equal("my-new-node", node.Name)
+	_, err = suite.Database.DeleteNode(node.Id)
```
phiggins: This error isn't being checked.

PR #313, opened 2019-05-10
Adding project filters for GetNodeRun

lancewf:

Adding project filtering for the cfgmgmt.proto GetNodeRun RPC.

- [ ] Code actually executed?

### DIFF
--- a/components/config-mgmt-service/integration_test/node_run_test.go
+++ b/components/config-mgmt-service/integration_test/node_run_test.go
@@ -141,7 +141,7 @@ func TestNodeRunProjectFilter(t *testing.T) {
 			expectedError: false,
 		},
 		{
-			description:   "Node has a projects; request unassigned projects allowed",
+			description:   "Node has a project; request only unassigned projects",
 			ctx:           contextWithProjects([]string{authzConstants.UnassignedProjectID}),
 			nodeProjects:  []string{"project9"},
 			expectedError: true,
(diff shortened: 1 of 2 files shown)


components/config-mgmt-service/integration_test/node_run_test.go
```diff
@@ -61,13 +70,128 @@ func TestNodeRunWithRuns(t *testing.T) {
...
+			expectedError: false,
+		},
+		{
+			description:   "Node has no projects; request unassigned projects allowed",
+			ctx:           contextWithProjects([]string{authzConstants.UnassignedProjectID}),
+			nodeProjects:  []string{},
+			expectedError: false,
+		},
+		{
+			description:   "Node has a projects; request unassigned projects allowed",
```
msorens: 1. s/projects/project/ (here and on L150).
2. If I take your meaning correctly, it should also say s/allowed/not allowed/

components/config-mgmt-service/integration_test/node_run_test.go
```diff
@@ -61,13 +70,128 @@ func TestNodeRunWithRuns(t *testing.T) {
...
+			nodeProjects:  []string{"project9"},
+			expectedError: true,
+		},
+		{
+			description:   "Node has a projects; request unassigned and matching project allowed",
+			ctx:           contextWithProjects([]string{authzConstants.UnassignedProjectID, "project9"}),
+			nodeProjects:  []string{"project9"},
+			expectedError: false,
+		},
+	}
```
msorens: Consider these tests as well:
1 Node has no projects; request a project not allowed
2 Node with one project matching one of several requested projects allowed
3 Node with one project matching not matching any of several requested projects not allowed
4 Node with several projects where one matches a single requested project allowed
5 Node with several projects where one matches one of several requested project allowed
6 Node with several projects where none matches several requested project not allowed

lancewf: Thank you for all these. I have added them and one other.

components/config-mgmt-service/integration_test/node_run_test.go
```diff
@@ -61,13 +70,170 @@ func TestNodeRunWithRuns(t *testing.T) {
...
+			},
+		}
+	)
+
+	suite.IngestNodes([]iBackend.Node{node})
+	suite.IngestRuns([]iBackend.Run{run})
+	defer suite.DeleteAllDocuments()
+
+	_, err := cfgmgmt.GetNodeRun(ctx, &req)
+	assert.Error(t, err)
```
phiggins: Would it be worth making this test that the first return value is nil?

components/config-mgmt-service/integration_test/node_run_test.go
```diff
@@ -61,13 +70,170 @@ func TestNodeRunWithRuns(t *testing.T) {
...
+			}
+			suite.IngestNodes([]iBackend.Node{node})
+			suite.IngestRuns([]iBackend.Run{run})
+			defer suite.DeleteAllDocuments()
+
+			res, err := cfgmgmt.GetNodeRun(test.ctx, &req)
+			if test.expectedError {
+				assert.Error(t, err)
+				return
+			}
```
phiggins: Would it be worth adding an assertion on the value of `res`?

components/config-mgmt-service/integration_test/node_run_test.go
lancewf: I was under the impression that the golang way was that if an error was returned the other return value could be anything. If this is the case then I don't think we should test the second value.

phiggins: Hmm, it seems like that would allow undefined behavior, where some half-baked data could leak through because we're not explicit about our error cases.

lancewf: Does it matter if we have half-baked data when we return an error? I thought when an error was returned the other value should not be used.

phiggins: It shouldn't, but we can always test to make sure it isn't. ;)

PR #316, opened 2019-05-10
[A2-783] Wire up  the projects filter "Apply Changes" button

msorens:

Before this change, the project filter would only take effect when you navigate to a different page within Automate. Now, when you change the filter, the current page will refresh itself to reflect the new set of projects, too.

Here I have project p2 selected, and nodes in that project display:
image

When I switch to project p4, which has no nodes, the list becomes empty:
image

### :+1: Definition of Done

Filters work throughout Automate.

Rebuild automate-ui.
Create some projects and assign them to some entities.
Exercise the global project filter on the appropriate page by selecting new projects and observe the list filtering change (almost) immediately when "Apply Changes" is selected.

- [x] Code actually executed?

### DIFF
--- a/components/automate-ui/src/app/app-routing.module.ts
+++ b/components/automate-ui/src/app/app-routing.module.ts
@@ -297,6 +297,10 @@ const routes: Routes = [
     pathMatch: 'prefix',
     redirectTo: 'settings/node-credentials'
   },
+  { // ued by projects-filter.service.ts
+    path: 'reload',
+    children: []
+  },
   // END Deprecated routes.
   { // everything unknown goes to client runs
     path: '**',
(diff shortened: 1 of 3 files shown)


components/automate-ui/src/app/app-routing.module.ts
```diff
@@ -297,6 +297,10 @@ const routes: Routes = [
     pathMatch: 'prefix',
     redirectTo: 'settings/node-credentials'
   },
+  { // ued by projects-filter.service.ts
```
tylercloke: `used`

msorens: Ah, thanks.

blakestier:
This is good-to-go but I think we need a follow-up with some tests 🎉

PR #322, opened 2019-05-13
project filtering for stats summary

rickmarry:
### :nut_and_bolt: Description

Adding project filtering for the stats.proto ReadSummary RPC.

### :chains: Related Resources

https://github.com/chef/automate/issues/67

### :white_check_mark: Checklist

- [x] Necessary tests added/updated?
- [ ] Necessary docs added/updated?
- [ ] Code actually executed?
- [x] Vetting performed (unit tests, lint, etc.)?

PR #324, opened 2019-05-13
reporting stats trends filter by project

rickmarry:

Adding project filtering for the stats.proto ReadTrend RPC.

https://github.com/chef/automate/issues/67

- [ ] Code actually executed?

### DIFF
--- a/components/compliance-service/integration_test/stats_server_read_summary_test.go
+++ b/components/compliance-service/integration_test/stats_server_read_summary_test.go
@@ -91,15 +91,9 @@ func TestReadSummary(t *testing.T) {
 
 	suite.RefreshComplianceReportIndex()
 
-	octoberTwentyFifthQuery := stats.Query{
-		Filters: []*stats.ListFilter{
-			{Type: "end_time", Values: []string{"2018-10-25T23:59:59Z"}},
-		},
-	}
 	successCases := []struct {
 		description     string
 		allowedProjects []string
-		query           stats.Query
 
 		//report summary
 		expectedEnvironmentCnt int32
@@ -122,7 +116,6 @@ func TestReadSummary(t *testing.T) {
 		{
 			description:     "Projects: user has access to all projects",
 			allowedProjects: []string{authzConstants.AllProjectsExternalID},
-			query:           octoberTwentyFifthQuery,
 
 			//report summary
 			expectedEnvironmentCnt: 5,
@@ -140,7 +133,6 @@ func TestReadSummary(t *testing.T) {
 		{
 			description:     "Projects: user has access to one project with reports",
 			allowedProjects: []string{"project1"},
-			query:           octoberTwentyFifthQuery,
 
 			//report summary
 			expectedEnvironmentCnt: 2,
@@ -158,7 +150,6 @@ func TestReadSummary(t *testing.T) {
 		{
 			description:     "Projects: user has access to some projects with reports",
 			allowedProjects: []string{"project1", "project2"},
-			query:           octoberTwentyFifthQuery,
 
 			//report summary
 			expectedEnvironmentCnt: 4,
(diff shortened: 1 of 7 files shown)


components/compliance-service/integration_test/stats_server_read_trend_test.go
```diff
@@ -0,0 +1,187 @@
...
+		"project1": reportIds[1:3],
+		"project2": reportIds[2:5],
+		"project3": reportIds[3:],
+	}
+
+	projectRules := map[string]*iam_v2.ProjectRules{}
+	for k, v := range reportsProjects {
+		projectRules[k] = &iam_v2.ProjectRules{
+			Rules: []*iam_v2.ProjectRule{
+				{
```
lancewf: You should add the below line even though your test is bypassing the check for it. 
```suggestion
				{
					Type: iam_v2.ProjectRuleTypes_EVENT,
```

components/compliance-service/integration_test/stats_server_read_trend_test.go
```diff
@@ -0,0 +1,187 @@
...
+
+			expectedPassedCnt: 1,
+		},
+	}
+
+	for _, test := range successCases {
+		t.Run(test.description, func(t *testing.T) {
+			ctx := contextWithProjects(test.allowedProjects)
+
+			test.query.Type = "nodes"
```
lancewf: Is there a reason this is not placed in the creation of the query at line 94? The same with the line 177 `test.query.Type = "controls"`

rickmarry: Yes.  The reason these are set late like this is because they don't coexist for each call.
I set Type to "nodes" first, then make the call to the backend "ReadTrend" api. This returns node oriented trend data.  I then make appropriate assertions and then move on.  The next test is with type "control".. again, making backend call with "control" set as Type.  

I could have, instead, had a separate block dedicated to each respective type, but there would have been a lot of duplicate code.. so I combined.

components/compliance-service/integration_test/stats_server_read_summary_test.go
```diff
@@ -0,0 +1,314 @@
...
+
+			assert.NoError(t, err)
+			require.NotNil(t, response)
+
+			controlSummary := response.ControlsSummary
+
+			//control summary
+			assert.Equal(t, test.expectedPassedCnt, controlSummary.Passed, "Passed count")
+		})
+	}
```
phiggins: I'd make the same comment as on the other PR: https://github.com/chef/automate/pull/322#discussion_r283510492

components/compliance-service/integration_test/stats_server_read_trend_test.go
```diff
@@ -0,0 +1,179 @@
...
+			octoberTwentyFifthQuery.Type = "controls"
+			response, err = statsServer.ReadTrend(ctx, octoberTwentyFifthQuery)
+
+			assert.NoError(t, err)
+			require.NotNil(t, response)
+
+			//controls trends
+			assert.Equal(t, test.expectedPassedCnt, response.Trends[0].Passed, "Controls Passed count")
+		})
+	}
```
phiggins: I'd make the same comment here as on the other PR: https://github.com/chef/automate/pull/322#discussion_r283510492

components/compliance-service/reporting/relaxting/util.go
```diff
@@ -7,6 +7,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/chef/automate/lib/grpc/auth_context"
+	"golang.org/x/net/context"
```
phiggins: This import should just be `"context"`.

components/compliance-service/integration_test/stats_server_read_summary_test.go
rickmarry: making that change now.. thanks

components/compliance-service/integration_test/stats_server_read_trend_test.go
lancewf: That makes sense. Thanks

PR #330, opened 2019-05-13
break the summary tests into separate tests for nodes, reports and controls

rickmarry:

In order to improve test readability and, in case of test failures, to more easily see failure origin, the stats summary tests are being separated into three distinct sections:
1 - reports
2 - nodes
3 - controls

- [ ] Code actually executed?

### DIFF
--- a/components/compliance-service/integration_test/stats_server_read_summary_test.go
+++ b/components/compliance-service/integration_test/stats_server_read_summary_test.go
@@ -18,78 +18,9 @@ import (
 	"github.com/chef/automate/components/compliance-service/reporting/relaxting"
 )
 
-func TestReadSummary(t *testing.T) {
-	reportFileName := "../ingest/examples/compliance-success-tiny-report.json"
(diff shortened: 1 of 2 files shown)


phiggins:
Sorry if I wasn't clear in my other comment, but I think it'd be ok to keep the test setup shared but have a different test for each of the code paths. For the suggestions tests, the test setup and results were the same for each suggestion type but they take different ways of getting the data, so I split those into multiple tests for each test case: https://github.com/chef/automate/pull/322#discussion_r283510492

rickmarry:
> Sorry if I wasn't clear in my other comment, but I think it'd be ok to keep the test setup shared but have a different test for each of the code paths. For the suggestions tests, the test setup and results were the same for each suggestion type but they take different ways of getting the data, so I split those into multiple tests for each test case: [#322 (comment)](https://github.com/chef/automate/pull/322#discussion_r283510492)

Thanks @phiggins.  I've refactored a bit. I moved some of the boilerplate stuff into a setup func and call it from each test now.  This cuts down on the the duplicate code and makes things more readable.

PR #332, opened 2019-05-14
add filtering for last scan and last run data on nodes endpoint

vjeffrey:
This work adds support for filtering by:
 - last run and last scan end time ranges
 - last run and last scan status
 - last run and last scan penultimate status

### :+1: Definition of Done
able to filter by above defined criteria on the `api/v0/nodes` endpoint

### DIFF
--- a/components/nodemanager-service/pgdb/filtering.go
+++ b/components/nodemanager-service/pgdb/filtering.go
@@ -120,6 +120,10 @@ func acceptedJSONFilterFields(field string) bool {
 		return true
 	case "last_scan ->> 'Status'":
 		return true
+	case "last_run ->> 'PenultimateStatus'":
+		return true
+	case "last_scan ->> 'PenultimateStatus'":
+		return true
 	default:
 		return false
 	}
--- a/components/nodemanager-service/pgdb/filtering.go
+++ b/components/nodemanager-service/pgdb/filtering.go
@@ -91,8 +91,10 @@ func buildWhereFilter(mergeableFilters []*common.Filter, tableAbbrev string, fil
 
 // Builds an IN where condition like: j.parent_id IN ('e57605ed-bb8a-49b8-606c-af0e2b31b139')
 func whereFieldIn(field string, arr []string, tableAbbrev string) (condition string, err error) {
-	if !utils.IsSqlSafe(field) {
-		return "", &utils.InvalidError{Msg: fmt.Sprintf("Unsupported character found in field: %s", field)}
+	if !acceptedJSONFilterFields(field) {
+		if !utils.IsSqlSafe(field) {
+			return "", &utils.InvalidError{Msg: fmt.Sprintf("Unsupported character found in field: %s", field)}
+		}
 	}
 	condition += fmt.Sprintf("%s.%s IN (", tableAbbrev, field)
 	if len(arr) > 0 {
@@ -113,6 +115,10 @@ func acceptedJSONFilterFields(field string) bool {
 	case "last_run ->> 'EndTime'":
 		return true
 	case "last_scan ->> 'EndTime'":
+		return true
+	case "last_run ->> 'Status'":
+		return true
+	case "last_scan ->> 'Status'":
 		return true
 	default:
 		return false
--- a/components/nodemanager-service/pgdb/filtering.go
+++ b/components/nodemanager-service/pgdb/filtering.go
@@ -53,7 +53,6 @@ func buildWhereFilter(mergeableFilters []*common.Filter, tableAbbrev string, fil
 	for _, filter := range filters {
 		var newCondition string
 		var err error
-
 		if strings.HasPrefix(filter.Key, "tags:") {
 			tagKeyFilter := strings.TrimPrefix(filter.Key, "tags:")
 			newCondition, err = wherePatternMatchTags(tagKeyFilter, filter.Values, "t")
@@ -63,7 +62,7 @@ func buildWhereFilter(mergeableFilters []*common.Filter, tableAbbrev string, fil
 				return "", &utils.InvalidError{Msg: fmt.Sprintf("Unsupported filter field: %s", filter.Key)}
 			case "source_region", "name":
 				newCondition, err = wherePatternMatch(filterField[filter.Key], filter.Values, tableAbbrev)
-			case "statechange_timestamp", "last_contact":
+			case "statechange_timestamp", "last_contact", "last_run ->> 'EndTime'", "last_scan ->> 'EndTime'":
 				newCondition, err = whereFieldBetween(filterField[filter.Key], filter.Values, tableAbbrev)
 			case "manager_id":
 				newCondition, err = whereNodeManagerNodeExists(filterField[filter.Key], filter.Values, tableAbbrev)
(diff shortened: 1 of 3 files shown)


components/nodemanager-service/pgdb/nodes_integration_test.go
```diff
@@ -366,3 +368,291 @@ func (suite *NodesIntegrationSuite) TestProjectsAreRoundtrippedThroughNodeLifecy
...
+	}
+
+	listNodes, _, err := suite.Database.GetNodes("", nodes.Query_ASC, 1, 100, []*common.Filter{
+		{Key: "last_run_penultimate_status", Values: []string{"FAILED"}},
+	})
+	if err != nil {
+		suite.FailNow(err.Error())
+	}
+	suite.Equal(1, len(listNodes))
+	suite.Equal(nodes.LastContactData_FAILED, listNodes[0].GetRunData().GetPenultimateStatus())
```
lancewf: Nice. I did not know we stored the second to the latest status. I had to look up 'Penultimate' 😄

components/nodemanager-service/pgdb/filtering.go
```diff
@@ -92,8 +91,10 @@ func buildWhereFilter(mergeableFilters []*common.Filter, tableAbbrev string, fil
 
 // Builds an IN where condition like: j.parent_id IN ('e57605ed-bb8a-49b8-606c-af0e2b31b139')
 func whereFieldIn(field string, arr []string, tableAbbrev string) (condition string, err error) {
-	if !utils.IsSqlSafe(field) {
-		return "", &utils.InvalidError{Msg: fmt.Sprintf("Unsupported character found in field: %s", field)}
+	if !acceptedJSONFilterFields(field) {
+		if !utils.IsSqlSafe(field) {
+			return "", &utils.InvalidError{Msg: fmt.Sprintf("Unsupported character found in field: %s", field)}
+		}
```
phiggins: I know this is not all new code from this PR, but it's not clear what this is guarding against. All of the possible values this function could receive are hardcoded in the `nodesFilterField` map in `pgdb/nodes.go`. If we still aren't comfortable interpolating plain strings into our SQL (which I think is totally reasonable!), `pq` has a function that I think should guard against any shenanigans here: https://godoc.org/github.com/lib/pq#QuoteIdentifier

components/nodemanager-service/pgdb/nodes_integration_test.go
```diff
@@ -366,3 +368,291 @@ func (suite *NodesIntegrationSuite) TestProjectsAreRoundtrippedThroughNodeLifecy
...
+		ScanData: &nodes.LastContactData{
+			Id:      "1003-9254-2004-1322",
+			EndTime: nowTime,
+			Status:  nodes.LastContactData_PASSED,
+		},
+	}
+	err := suite.Database.ProcessIncomingNode(node)
+	if err != nil {
+		suite.FailNow(err.Error())
+	}
```
phiggins: This could be replaced with `suite.Require().NoError(err)`.

components/nodemanager-service/pgdb/nodes_integration_test.go
```diff
@@ -366,3 +368,291 @@ func (suite *NodesIntegrationSuite) TestProjectsAreRoundtrippedThroughNodeLifecy
...
+
+	timeRangeMax := time.Now().Add(time.Hour * 24).Format(time.RFC3339)
+	listNodes, _, err := suite.Database.GetNodes("", nodes.Query_ASC, 1, 100, []*common.Filter{
+		{Key: "last_scan_timerange", Values: []string{"2019-03-05T00:00:00Z", timeRangeMax}},
+	})
+	if err != nil {
+		suite.FailNow(err.Error())
+	}
```
phiggins: This could be replaced with suite.Require().NoError(err).

components/nodemanager-service/pgdb/nodes_integration_test.go
```diff
@@ -366,3 +368,291 @@ func (suite *NodesIntegrationSuite) TestProjectsAreRoundtrippedThroughNodeLifecy
...
+			Status:  nodes.LastContactData_PASSED,
+		},
+	}
+	err := suite.Database.ProcessIncomingNode(node)
+	if err != nil {
+		suite.FailNow(err.Error())
+	}
+
+	timeRangeMax := time.Now().Add(time.Hour * 24).Format(time.RFC3339)
+	listNodes, _, err := suite.Database.GetNodes("", nodes.Query_ASC, 1, 100, []*common.Filter{
```
phiggins: Since the second return value is generated with a different sql query, is it worth adding a test that it is correct also?

components/nodemanager-service/pgdb/nodes_integration_test.go
```diff
@@ -366,3 +368,291 @@ func (suite *NodesIntegrationSuite) TestProjectsAreRoundtrippedThroughNodeLifecy
...
+		RunData: &nodes.LastContactData{
+			Id:      "1003-9254-2004-1322",
+			EndTime: nowTime,
+			Status:  nodes.LastContactData_FAILED,
+		},
+	}
+	err := suite.Database.ProcessIncomingNode(node)
+	if err != nil {
+		suite.FailNow(err.Error())
+	}
```
phiggins: This could be replaced with `suite.Require().NoError(err)`.

components/nodemanager-service/pgdb/nodes_integration_test.go
```diff
@@ -366,3 +368,291 @@ func (suite *NodesIntegrationSuite) TestProjectsAreRoundtrippedThroughNodeLifecy
...
+		SourceId:        "",
+		SourceRegion:    "",
+		SourceAccountId: "",
+		JobUuid:         "12343-232324-1231242",
+		RunData: &nodes.LastContactData{
+			Id:      "1003-9254-2004-1322",
+			EndTime: nowTime,
+			Status:  nodes.LastContactData_PASSED,
+		},
+	}
```
phiggins: It might end up less neat than in my head, but with this test structure mostly repeated in every test, it might be nicer to extract the base metadata to a constant that gets copied and updated with whatever relevant values in each test. It'd make it clearer in each test what the significant values are.

components/nodemanager-service/pgdb/filtering.go
vjeffrey: > All of the possible values this function could receive are hardcoded 

Are they? tags aren't hardcoded, iirc. those just get passed through

components/nodemanager-service/pgdb/nodes_integration_test.go
vjeffrey: I don't know why i always forget about that one :/

components/nodemanager-service/pgdb/nodes_integration_test.go
vjeffrey: ya, that makes sense. there's a lot of stuff that's repeated

components/nodemanager-service/pgdb/filtering.go
vjeffrey: https://github.com/chef/automate/blob/master/components/nodemanager-service/pgdb/nodes.go#L582
https://github.com/chef/automate/blob/master/components/nodemanager-service/pgdb/filtering.go#L57

vjeffrey: though i guess that means we need the issqlsafe bit there but elsewhere? hmmm. im not sure.

phiggins: Ahh yeah, I forgot about the tags.

PR #341, opened 2019-05-14
User real Habitat HealthCheck event message

afiune:
Update Automate so that all the features we have now work with real data from Habitat and not just canned data from the generators. Though, generators should also be updated.

**IMPORTANT:** All this work is based of @christophermaier PR and mine:
- https://github.com/habitat-sh/habitat/pull/6344
- https://github.com/habitat-sh/habitat/pull/6520

### :+1: Definition of Done
When we merge this change, we will work the same exact way we were working before, that is, ingesting a single message to shape our system (inside the applications-service) but we will be using the same protobuf message that Habitat will be sending (`HealthCheckEvent`) after the PR https://github.com/habitat-sh/habitat/pull/6344 gets merged.

Additionally, we should be able to work with the generator as we were doing it before this PR.

To test that the generators work as expected you should still be able to run the following commands and navigate to the Apps Tab in the UI:
```
$ hab studio enter
[2][default:/src:0]# applications_populate_database
[3][default:/src:0]# start_all_services
```
To test that Automate can accept messages from habitat you will need to setup your Habitat dev environment, build the habitat supervisor component from the PR https://github.com/habitat-sh/habitat/pull/6344 and the start the supervisor with the following parameters:

```
sudo HAB_FEAT_EVENT_STREAM=1 HAB_SUP_BINARY=/home/afiune/github/habitat/target/debug/hab-sup /home/afiune/github/habitat/target/debug/hab sup run --event-stream-application=MY_APP --event-stream-environment=MY_ENV --event-stream-url=192.168.11.134:4222 --event-stream-token="ptNAt4G7lm7V8ofBo0XidlEhdi0=" &
```  
Then you can start any service and you will see it immediately in the Applications Tab:
```
afiune@afiune-ubuntu-vb:~/github/habitat/components/sup$ sudo HAB_SUP_BINARY=/home/afiune/github/habitat/target/debug/hab-sup /home/afiune/github/habitat/target/debug/hab svc load core/redis

afiune@afiune-ubuntu-vb:~/github/habitat/components/sup$ sudo HAB_SUP_BINARY=/home/afiune/github/habitat/target/debug/hab-sup /home/afiune/github/habitat/target/debug/hab sup status
package                           type        desired  state  elapsed (s)  pid    group
core/redis/4.0.14/20190319155852  standalone  up       up     79           31789  redis.default
```

Screen Shot 2019-05-15 at 5 10 43 PM

https://chefio.atlassian.net/browse/A2-770

- [x] Code actually executed?


api/external/applications/applications.proto
```diff
@@ -133,53 +133,6 @@ message HealthCounts {
   int32 unknown = 5;
 }
 
-// A HabService message is the representation of one single service in habitat
-message HabService {
```
afiune: Long live to our "**HabService**" message!
tenor-204203012

components/applications-service/pkg/storage/postgres/hab_event.go
```diff
@@ -76,64 +99,113 @@ func (db *Postgres) IngestHabEventWithoutMetrics(event *applications.HabService)
...
 			if err := tx.Insert(sup); err != nil {
 				return errors.Wrap(err, "Unable to insert supervisor")
 			}
 			sid = sup.ID
 		}
 
 		// 4) Service
-		svc := serviceFromHabEvent(event, did, sid, gid)
+		svc := newService(pkgIdent, health, did, sid, gid)
+		if svcMetadata.GetUpdateConfig() != nil {
```
kmacgugan: I think we can move this to newService, for better encapsulation. Also we should add update strategy, as channel is irrelevant if update strategy is nil we should update the business logic when retrieving this as well.

kmacgugan:
I think we need to add update strategy and also update the logic that presents the channel information to the UI. Do we also want to add helpers to regenerate the event proto messages? We should merge this PR after https://github.com/habitat-sh/habitat/pull/6344

components/applications-service/integration_test/global_test.go
```diff
@@ -129,24 +193,70 @@ func assertServiceGroupsEqual(t *testing.T, expected, actual *applications.Servi
...
 // TODO @afiune add more apps & envs
-func habServicesMatrix() []*applications.HabService {
-	return []*applications.HabService{
+func habServicesMatrix() []*habitat.HealthCheckEvent {
+	return []*habitat.HealthCheckEvent{
 		// service_group 1 <-> With a Health Status = 'OK'
-		NewHabServiceMsg("sup1", a, e, "default", "core", "redis", "0.1.0", "20190101121212", "OK", "stable", "test"),
-		NewHabServiceMsg("sup2", a, e, "default", "core", "redis", "0.1.0", "20190101121212", "OK", "stable", "test"),
-		NewHabServiceMsg("sup3", a, e, "default", "core", "redis", "0.1.0", "20190101121212", "OK", "stable", "test"),
+		NewHabitatEvent([]MessageOverrides{
```
danielsdeleo: can't we pass the arguments instead of making a slice and "splatting" it?

afiune: How so? Like it was before? I feel this is a better way but I would like to hear why and how would you like this to be updated! Looove any feedback! 💯

afiune: I could change the function argument from:
```go
func NewHabitatEvent(overrides ...MessageOverrides) *habitat.HealthCheckEvent {
```
To:
```go
func NewHabitatEvent(overrides []MessageOverrides) *habitat.HealthCheckEvent {
```

Would that work @danielcbright ?

components/applications-service/integration_test/service_groups_test.go
```diff
@@ -626,24 +676,32 @@ func habServicesABCD() []*habitat.HealthCheckEvent {
 			withServiceGroup("a.default"),
 			withPackageIdent("core/a/0.1.0/20190101121212"),
 			withHealth("UNKNOWN"),
+			withApplication("a_app"),
+			withEnvironment("a_env"),
```
afiune: This is the cool part about the new test function to create Hab Event Messages. Just add a single func to update it instead of all the tests and functions and options! cc/ @kmacgugan @danielsdeleo 

tenor-200387754

PR #343, opened 2019-05-14
Fixing flaky UI e2e test

lancewf:
active

Fixing a flaky UI e2e test. When selecting the event feed dropdown filter the 'active' attribute should be added. The problem was that the test was checking before that 'active' attribute was added. So, the fix here is to wait until after the click and then check for the 'active' attribute for about a second. The other option was to add a sleep statement and sleeps cause other problems. 

I found a solution thanks to this website https://www.captechconsulting.com/blogs/keep-your-promises-making-the-jump-from-selenium-to-protractor

### DIFF
--- a/components/automate-ui/e2e/event-feed-page.e2e-spec.ts
+++ b/components/automate-ui/e2e/event-feed-page.e2e-spec.ts
@@ -9,9 +9,18 @@ describe('Event Feed: Guitar String Graph Interaction', () => {
 
   it('should have a working dropdown filter', () => {
     const select = element(by.css('app-event-feed-select chef-select'));
-    select.click();
 
-    expect(select.getAttribute('class')).toContain('active');
+    // Click the event feed dropdown filter and wait until the filter turns 'active'
+    select.click().then(() => {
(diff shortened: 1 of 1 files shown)


components/automate-ui/e2e/event-feed-page.e2e-spec.ts
```diff
@@ -9,9 +9,18 @@ describe('Event Feed: Guitar String Graph Interaction', () => {
...
     const select = element(by.css('app-event-feed-select chef-select'));
-    select.click();
 
-    expect(select.getAttribute('class')).toContain('active');
+    // Click the event feed dropdown filter and wait until the filter turns 'active'
+    select.click().then(() => {
+      return new Promise(function(resolve) {
+        browser.wait(() => {
+          return select.getAttribute('class').then((values: string) => {
+            return values.indexOf('active') > 0;
```
msorens: Probably OK, but note that this would be happy with a class name of `inactive` also. Ideally should separate the class by words, e.g., `values.split(' ').includes('active')`

PR #345, opened 2019-05-14
 ingest-service calls nodemanager-service with node metadata upon ingestion

vjeffrey:
continuing work to support https://github.com/chef/automate/issues/192

On ingestion, the ingest-service calls nodemanager ProcessNode with some metadata about the node.
If anything in the call fails, we just move along and log the error. This is the same way it is implemented in compliance ingestion. We can revisit this in the future if needed.

thanks to @lancewf for helping me figure out where this fits in the pipeline and how to test it! those mocks are a great tool to have!

### :+1: Definition of Done
ingest service calls nodemanager service on ccr ingestion

rebuild deployment and ingest service
send some nodes through (chef_load_nodes)
check the nodemanager db - see nodes have run data

https://github.com/chef/automate/pull/332


components/ingest-service/pipeline/publisher/node_manager_publisher_test.go
```diff
@@ -0,0 +1,81 @@
...
+		})
+	errc := make(chan error)
+
+	inbox <- message.NewChefRun(context.Background(), &chef.Run{}, errc)
+	close(inbox)
+	out := nodeManagerPublisher(inbox, nodeMgrClient)
+
+	<-out
+
+	assert.Equal(t, 1, processNodeCount)
```
phiggins: I'm not too familiar with gomock, but from using other mocking libraries I think gomock's `Times` function is used for this kind of test: https://godoc.org/github.com/golang/mock/gomock#Call.Times

.studio/common
```diff
@@ -83,6 +83,7 @@ function compile_client_mocks() {
     mockgen -source authn/authenticate.pb.go -destination authn/authenticate.pb.client_mock.go -package authn
     mockgen -source cfgmgmt/service/cfgmgmt.pb.go -destination cfgmgmt/service/cfgmgmt.pb.client_mock.go -package service
     mockgen -source ingest/chef.pb.go -destination ingest/chef.pb.client_mock.go -package ingest
+    mockgen -source ../../components/nodemanager-service/api/manager/manager.pb.go -destination ../../components/nodemanager-service/api/manager/manager.pb.client_mock.go -package manager
```
lancewf: Nice! Thank you for adding this.

components/ingest-service/pipeline/publisher/node_manager_publisher_test.go
lancewf: I would say to leave it how it is. I have a lot of problems with using this gomock libraries `Times` function. 

I would think that the below code with the `Times` function would work, but I have commented out the `ProcessNode` call in the `nodeManagerPublisher` function and the test does not fail. With the code how it is now it does fail with the `ProcessNode` call commented out. 

```
	inbox := make(chan message.ChefRun, 100)
	nodeMgrClient := manager.NewMockNodeManagerServiceClient(gomock.NewController(t))
	nodeMgrClient.EXPECT().ProcessNode(gomock.Any(), gomock.Any()).Times(1).Return(&manager.ProcessNodeResponse{}, nil)
	errc := make(chan error)

	inbox <- message.NewChefRun(context.Background(), &chef.Run{}, errc)
	close(inbox)
	out := nodeManagerPublisher(inbox, nodeMgrClient)

	<-out
```

lancewf: Ok, I figure out the problem. The below code will work with the `Times` function. I did not know I had to use the `Controller`. 
```
	inbox := make(chan message.ChefRun, 100)
	ctr := gomock.NewController(t)
	nodeMgrClient := manager.NewMockNodeManagerServiceClient(ctr)
	nodeMgrClient.EXPECT().ProcessNode(gomock.Any(), gomock.Any()).Times(1).Return(&manager.ProcessNodeResponse{}, nil)
	errc := make(chan error)

	inbox <- message.NewChefRun(context.Background(), &chef.Run{}, errc)
	close(inbox)
	out := nodeManagerPublisher(inbox, nodeMgrClient)

	<-out

	ctr.Finish()
```

phiggins: https://github.com/chef/automate/pull/352#issuecomment-492801186

Quoting myself:

> I know I was the one suggesting using Times in the first place, but I'm starting to think just using a counter variable is better if its less likely to turn into a silent failure.

PR #346, opened 2019-05-14
PROTOTYPE/EXPERIMENT - Wrapping Messages in protobuf.Any

danielsdeleo:

**THIS IS JUST A DEMONSTRATION OF protobuf.any not intended for any real use**

This is to show one way we can handle multiple message types from hab using a single undifferentiated stream of messages (i.e., no _implicit_ type information). I'll make some notes about it in code comments.

- [ ] Code actually executed?


api/external/applications/applications.proto
```diff
@@ -212,3 +213,18 @@ enum HealthStatus {
   CRITICAL = 2;
   UNKNOWN = 3;
 }
+
+message Wrapper {
+  string type = 1;
```
danielsdeleo: One case should should think about how we handle, is what happens when Automate is "old" and hab is "new," specifically if a new message type is added and hab starts to send these messages but automate isn't upgraded so it doesn't know what they are. In this demonstration, we would switch the code path based on the content `Wrapper.Type` and drop any messages with unknown values, e.g.,

```go
	switch wrapperMessage.Type {
	case "ExampleOne":
		var fullyUnwrappedMessage applications.ExampleOne
		err = ptypes.UnmarshalAny(wrap1rt.Message, &fullyUnwrappedMessage)
		// errcheck
		handleExampleOneMessage(message)
	case "ExampleTwo":
		// etc.
	default: // drop it
	}
```

components/applications-service/pkg/wrap/wrap_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	packedEx1, err := ptypes.MarshalAny(&ex1)
+	require.NoError(t, err)
+
+	wrap1 := applications.Wrapper{
+		Type:    "ExampleOne",
+		Message: packedEx1,
+	}
+
+	fmt.Printf("w: %+v\n", wrap1)
```
danielsdeleo: If you look at the output of this, it adds some fixed amount of overhead we have to deal with. The output is like:

```
{Type:ExampleOne Message:type_url:"type.googleapis.com/chef.automate.api.applications.ExampleOne" value:"\010\027\020*"  XXX_NoUnkeyedLiteral:{} XXX_unrecognized:[] XXX_sizecache:0}
```

So we have stored our own definition of what the type is, and then `ptypes.MarshalAny()` has also converted our message to a thing with its own definition of what the type is. It might be possible to remove our type "tag" and use the one added by protobuf.Any, see https://github.com/golang/protobuf/blob/e91709a02e0e8ff8b86b7aa913fdc9ae9498e825/ptypes/any/any.pb.go#L104-L129

components/applications-service/pkg/wrap/wrap_test.go
```diff
@@ -0,0 +1,51 @@
...
+
+	err = proto.Unmarshal(data, &wrap1rt)
+
+	require.NoError(t, err)
+
+	fmt.Printf("w (rt): %+v\n", wrap1rt)
+
+	var extractedEx1 applications.ExampleOne
+
+	err = ptypes.UnmarshalAny(wrap1rt.Message, &extractedEx1)
```
danielsdeleo: Another possibility (I think) is to unmarshal the any into an `interface{}` and do some coercion? But that seems not good.

components/applications-service/pkg/wrap/wrap_test.go
danielsdeleo: So one alternative would be to make the wrapper like so (proto):

```protobuf
message Wrapper {
  string type = 1;
  bytes message = 2;
}
```

This is what `protobuf.Any` does internally anyway: https://github.com/golang/protobuf/blob/e91709a02e0e8ff8b86b7aa913fdc9ae9498e825/ptypes/any/any.proto#L122-L154

The advantages of rolling our own would be:
* shorter type annotations in the message for less message size overhead
* wire compatibility is not broken if we change the protobuf package name(space)

danielsdeleo:
the "roll your own any" approach in the other PR seems better.

PR #348, opened 2019-05-14
DEMONSTRATION OF A "Roll your own protobuf.Any" Approach

danielsdeleo:

**PROTOTYPE/DEMO ONLY**

This PR is for comparison with https://github.com/chef/automate/pull/346

- [ ] Code actually executed?


components/applications-service/pkg/wrap/wrap_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/chef/automate/api/external/applications"
+)
+
+func TestWrap(t *testing.T) {
+	ex1 := applications.ExampleOne{
+		IntOne: 23,
+		IntTwo: 42,
+	}
+
+	packedEx1, err := proto.Marshal(&ex1)
```
danielsdeleo: Even though we have to "manually" marshal and unmarshal the nested protobufs, it's a similar amount of code because when we used `protobuf.Any` we needed to call `ptypes.MarshalAny()` and `ptypes.UnmarshalAny()`

components/applications-service/pkg/wrap/wrap_test.go
```diff
@@ -0,0 +1,50 @@
...
+	require.NoError(t, err)
+
+	wrap1 := applications.Wrapper{
+		Type:    "ExampleOne",
+		Message: packedEx1,
+	}
+
+	fmt.Printf("w: %+v\n", wrap1)
```
danielsdeleo: This has a lot smaller content compared to protobuf.Any:

```
w: {Type:ExampleOne Message:[8 23 16 42] XXX_NoUnkeyedLiteral:{} XXX_unrecognized:[] XXX_sizecache:0}
```

api/external/applications/applications.proto
```diff
@@ -212,3 +212,30 @@ enum HealthStatus {
...
+	// siblings of their type, not children of it.
+	//
+	// So we need to nest this one extra level
+	enum WrappedMessageType {
+		UNKNOWN = 0;
+		EXAMPLE_ONE = 1;
+		EXAMPLE_TWO = 2;
+	}
+
+  WrappedMessageType type = 1;
```
danielsdeleo: Making this an enum I think gives us the lowest possible overhead in terms of additional bytes on the wire. If you run the test code, it prints:
```
inner msg: [8 23 16 42]
inner msg len: 4
data: [8 1 18 4 8 23 16 42]
data len: 8
```

So we only added 4 bytes

components/applications-service/pkg/wrap/wrap_test.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	data2, err := proto.Marshal(&wrap2)
+	require.NoError(t, err)
+
+	dispatchWrappedMessage(data1, t)
+	dispatchWrappedMessage(data2, t)
+}
+
+func dispatchWrappedMessage(m []byte, t *testing.T) {
```
danielsdeleo: I'm thinking if we implement the wrapper approach we'll end up with one function like this where we contain all the boiler-plate-y stuff. IMO it's not too awful and it's probably plenty efficient.

api/external/applications/applications.proto
danielsdeleo: For further discussion, in the last load test, we had about 600 messages per second, so this would add 4 * 600 == 2400 bps, which should be pretty much insignificant.

danielsdeleo:
The concept is demonstrated and we decided on a design here so no need for this any more.

PR #349, opened 2019-05-14
workflow library

stevendanna:
*This code is in flux and not ready for code-level review. We've opened
this PR to facilitate a discussion about the high-level design.*

This is a spike implementation of a "workflow" library. The goal of
this library is to provide applications with a means of describing
application-specific processes such that they can be run across
multiple machines and resumed after crashes or reboots.

This library is inspired by the following systems:

https://cadenceworkflow.io/
https://docs.aws.amazon.com/swf/index.html

It is centered around workflows (nothing to do with delivery) and
tasks. Workflows coordinate the execution of Tasks. Tasks are discrete
bits of application logic.

A Workflow can

- start Tasks,
- make decisions based on the completion of Task, and
- make decisions based on a cancellation signal.

The logic of a workflow is defined in a WorkflowExecutor: Go code that
application authors write and register with the library. The executor
is defined as a set of callbacks. The workflow executor is started
with user-defined parameters that allow the same workflow executor to
be reused. The workflow executor can store state in a payload that is
persistent between steps, allowing it to implement more complex logic
based on past events.

A simple WorkflowExecutor might look like this (API not final):

```golang
type SimpleWorkflowExecutor struct{}
type SimpleWorkflowPayload struct {
	State string
}

// The OnStart callback is the entry point for a workflow’s execution
func (*SimpleWorkflowExecutor) OnStart(w workflow.WorkflowInstanceHandler,
	ev workflow.StartEvent) workflow.Decision {

	// Request "simple_task" be run
	w.EnqueueTask("simple_task", SimpleTaskParams{EchoMe: "Weeeee"})
	return w.Continue(SimpleWorkflowPayload{State: "REMEMBER ME"})
}

// OnTaskComplete is called whenever a task related to this workflow completes
func (p *SimpleWorkflowExecutor) OnTaskComplete(w workflow.WorkflowInstanceHandler,
	ev workflow.TaskCompleteEvent) workflow.Decision {
	switch ev.TaskName {
	case "simple_task":
		if err := ev.Result.Err(); err != nil {
			logrus.WithError(err).Error("simple_task failed")
		} else {
			var result string
			ev.Result.Get(&result)
			logrus.WithField("result", result).Info("simple_task done")
		}
	default:
		logrus.Error("Unknown task")
	}
	return w.Complete()
}
// OnCancel is called if any user-code sends a cancel single to this workflow
func (p *SimpleWorkflowExecutor) OnCancel(w workflow.WorkflowInstanceHandler, ev workflow.CancelEvent) workflow.Decision {
	return w.Complete()
}
```

The logic of a task is defined in a Task Executor: Go code that
application authors write and register with the library. For example,
a simple TaskExecutor might look like this:

```golang
type SimpleTaskParams struct {
	EchoMe string
}

type SimpleTaskExecutor struct{}

func (t *SimpleTaskExecutor) Run(ctx context.Context, task workflow.TaskQuerier) (interface{}, error) {
	params := SimpleTaskParams{}
	if err := task.GetParameters(&params); err != nil {
		return nil, err
	}
	fmt.Printf("%s\n", params.EchoMe)
	return "echoed", nil
}
```

While structuring application logic in this way is not always
straight-forward, it comes with a benefit. As long as
WorkflowExecutors and TaskExecutors use the provided payload and
parameter functions to receive and store state, the execution of a
workflow can occur across multiple machines in a Chef Automate
installation and resume after an application crash or restart.

We’ve implemented this system as a library that can be included in any
Go-based service.  It is built on top of PostgreSQL.

TaskExecutors work by starting a transaction, selecting an unprocessed
task from the tasks table, running the user-level code related to that
task, and then committing a transaction that removes the unprocessed
task from the task list and adds the result to the task_result
table. Row-level locking via SELECT FOR UPDATE is used to ensure only
one executor runs a given task at a time. If a TaskExecutor
disappears, the postgresql session will end, aborting the transaction
and leaving the task for another executor to restart.  Note that this
implies that Tasks should generally strive to be re-runnable.

WorkflowExecutors operate similarly to TaskExecutors.

Scheduled workflows are built on a polling system that checks all
scheduled workflows and starts any workflows that are due to start.
Uniqueness constraints on the workflow_instances table help us
identify if the task is still running from the last invocation.

TODO:

- [ ] Sort out use of `TIMESTAMP` vs `TIMEZONE WITH TIME STAMP` in the PG backend
- [ ] Tests
- [ ] Public-facing APIs for workflow cancellation
- [ ] Public-facing APIs for listing scheduled jobs
- [ ] Tests
- [ ] Tests
- [ ] Refactor type names. ssd has a plan
- [ ] Actually use pg_notify for lower latency workflow dequeueing 

Co-authored-by: Jay Mundrawala <jmundrawala@chef.io>
Co-authored-by: Yvonne Lam <yvonne@chef.io>


lib/workflow/backend/backend.go
```diff
@@ -0,0 +1,137 @@
...
+
+type TaskStatusType string
+
+const (
+	TaskStatusSuccess TaskStatusType = "success"
+	TaskStatusFailed  TaskStatusType = "failed"
+)
+
+type WorkflowEvent struct {
+	InstanceID         int64
```
stevendanna: It might be cleaner for this to live on the implementation of WorkflowCompleter since that is the only thing that should depend on the internal state of the backend?

lib/workflow/backend/backend.go
```diff
@@ -0,0 +1,137 @@
...
+	Type               WorkflowEventType
+	EnqueuedTaskCount  int
+	CompletedTaskCount int
+
+	TaskResult *TaskResult
+}
+
+type Task struct {
+	Name               string
+	WorkflowInstanceID int64
```
stevendanna: [minor] Similarly, this *also* seems like it could be part of the internal implementation of TaskCompleter.

lib/workflow/backend/backend.go
```diff
@@ -0,0 +1,137 @@
+package backend
```
stevendanna: We think we want to comment these types a bit as internal documentation.  We'll likely want to do the same for the public API.

lib/workflow/backend/backend.go
```diff
@@ -0,0 +1,137 @@
...
+type TaskEnqueueOpts struct {
+	TryRemaining int
+	StartAfter   time.Time
+}
+
+type WorkflowCompleter interface {
+	EnqueueTask(task *Task, opts TaskEnqueueOpts) error
+
+	Continue(payload []byte) error
+	Abandon() error
```
jaym: We might not need this anymore

lib/workflow/backend/backend.go
```diff
@@ -0,0 +1,137 @@
...
+
+type RecurringWorkflowCompleter interface {
+	EnqueueRecurringWorkflow(s *Schedule) error
+	Close()
+}
+
+type Schedule struct {
+	// NOTE(ssd) 2019-05-13: Since name and workflow-name are
+	// user-controlled in the case of many scheduled workflows, we
+	// need the ID to create unique workflow names.
```
stevendanna: This comment isn't true anymore.

lib/workflow/backend/backend.go
```diff
@@ -0,0 +1,137 @@
...
+	ID             int64
+	Enabled        bool
+	Name           string
```
stevendanna: s/Name/InstanceName/

lib/workflow/integration/error_task_test.go
```diff
@@ -0,0 +1,56 @@
...
+					return w.Complete()
+				},
+			},
+		),
+	)
+	defer m.Stop()
+	err := m.EnqueueWorkflow(context.Background(), workflowName, instanceName, nil)
+	suite.Require().NoError(err, "Failed to enqueue workflow")
+	wgTask.Wait()
+	time.Sleep(20 * time.Millisecond)
```
stevendanna: 👀

lib/workflow/postgres/migration_array_driver.go
```diff
@@ -0,0 +1,90 @@
...
+		// TODO(ssd) 2019-05-14: If we wanted this to be
+		// re-usable, we would need to take the passed string
+		// and lookup this data. For now, we just assume it is
+		// defined in this package.
+		migrations: sqlMigrations,
+	}, nil
+}
+
+func (a *ArrayMigrationDriver) Close() error { return nil }
+func (a *ArrayMigrationDriver) First() (uint, error) {
```
stevendanna: This seems super unit-testable, maybe some tests?

lib/workflow/postgres/postgres.go
```diff
@@ -0,0 +1,726 @@
...
+	"github.com/sirupsen/logrus"
+
+	"github.com/chef/automate/lib/workflow"
+	"github.com/chef/automate/lib/workflow/backend"
+)
+
+const (
+	enqueueTaskQuery   = `SELECT enqueue_task($1, $2, $3, $4, $5)`
+	dequeueTaskQuery   = `SELECT * FROM dequeue_task($1)`
+	completeTaskQuery  = `SELECT complete_task($1::bigint, $2::task_status, $3::text, $4)`
```
stevendanna: Do we need these type casts?

lib/workflow/postgres/sql_schema.go
```diff
@@ -0,0 +1,309 @@
...
+-- workflow that create and then waits for the completion of a number
+-- of inspec scan tasks.
+--
+-- The schedule for recurring workflows are stored in SQL and (TODO)
+-- go-level code will push on new workflow_instances and the appointed
+-- time.
+CREATE TABLE recurring_workflow_schedules (
+    id BIGSERIAL PRIMARY KEY,
+
+    name TEXT NOT NULL,
```
stevendanna: s/name/instance_name/

lib/workflow/postgres/sql_schema.go
```diff
@@ -0,0 +1,309 @@
...
+-- Workflows
+--
+-- Workflows coordinate a set of tasks. For example, a scan job is a
+-- workflow that create and then waits for the completion of a number
+-- of inspec scan tasks.
+--
+-- The schedule for recurring workflows are stored in SQL and (TODO)
+-- go-level code will push on new workflow_instances and the appointed
+-- time.
+CREATE TABLE recurring_workflow_schedules (
```
stevendanna: We should probably prefix these with something like `__workflow_`

lib/workflow/postgres/sql_schema.go
```diff
@@ -0,0 +1,309 @@
...
+		upSQL: `
+BEGIN;
+
+-- Workflows
+--
+-- Workflows coordinate a set of tasks. For example, a scan job is a
+-- workflow that create and then waits for the completion of a number
+-- of inspec scan tasks.
+--
+-- The schedule for recurring workflows are stored in SQL and (TODO)
```
stevendanna: Comment is out of date.

lib/workflow/postgres/sql_schema.go
```diff
@@ -0,0 +1,309 @@
...
+        SELECT * FROM recurring_workflow_schedules WHERE id = _id FOR UPDATE
+    )
+    UPDATE recurring_workflow_schedules SET enabled = _enabled WHERE id = _id;
+$$ LANGUAGE SQL;
+
+CREATE TYPE workflow_instance_status AS ENUM('running', 'abandoned');
+
+CREATE TABLE workflow_instances (
+    id BIGSERIAL PRIMARY KEY,
+    name TEXT NOT NULL,
```
jaym: s/name/instance_name

lib/workflow/postgres/sql_schema.go
```diff
@@ -0,0 +1,309 @@
...
+    enqueued_tasks INTEGER NOT NULL DEFAULT 0,
+    completed_tasks INTEGER NOT NULL DEFAULT 0,
+    status workflow_instance_status NOT NULL DEFAULT 'running',
+
+    CONSTRAINT say_my_name1 UNIQUE(name, workflow_name)
+);
+
+CREATE TABLE workflow_results (
+    id BIGSERIAL PRIMARY KEY,
+    name TEXT NOT NULL,
```
jaym: s/name/instance_name

lib/workflow/postgres/sql_schema.go
```diff
@@ -0,0 +1,309 @@
...
+    workflow_name TEXT NOT NULL,
+    parameters BYTEA,
+    payload BYTEA,
+    start_at TIMESTAMP NOT NULL DEFAULT NOW(),
+    updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    enqueued_tasks INTEGER NOT NULL DEFAULT 0,
+    completed_tasks INTEGER NOT NULL DEFAULT 0,
+    status workflow_instance_status NOT NULL DEFAULT 'running',
+
+    CONSTRAINT say_my_name1 UNIQUE(name, workflow_name)
```
jaym: we're not doing `say_my_name`

lib/workflow/postgres/sql_schema.go
```diff
@@ -0,0 +1,309 @@
...
+    workflow_name TEXT NOT NULL,
+    parameters BYTEA,
+    start_at TIMESTAMPTZ NOT NULL,
+    end_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    result BYTEA
+);
+
+-- Tasks
+--
+-- Tasks are units of descrete work that need to be done.
```
stevendanna: descrete -> discrete

lib/workflow/postgres/sql_schema.go
```diff
@@ -0,0 +1,309 @@
...
+    workflow_instance_id BIGINT NOT NULL,
+    try_remaining INT NOT NULL DEFAULT 1,
+    enqueued_at   TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    updated_at    TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    start_after   TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    task_name     TEXT NOT NULL,
+    parameters    BYTEA
+);
+
+CREATE TABLE tasks_results (
```
stevendanna: s/tasks_results/task_results/

lib/workflow/postgres/sql_schema.go
```diff
@@ -0,0 +1,309 @@
...
+    completed_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    status       task_status,
+    error        TEXT,
+    result       BYTEA
+);
+
+CREATE TYPE workflow_event_type AS ENUM('start', 'task_complete', 'cancel', 'tasks_abandoned');
+
+-- NOTE(ssd) 2019-05-09: Workflow events are defined here because they
+-- may reference task_restuls
```
stevendanna: comment is out of date

lib/workflow/postgres/sql_schema.go
```diff
@@ -0,0 +1,309 @@
...
+    id BIGSERIAL PRIMARY KEY,
+    event_type workflow_event_type NOT NULL,
+    workflow_instance_id BIGINT,
+    enqueued_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    -- task_complete members
+    task_result_id BIGINT
+);
+
+-- Workflow Functions
+CREATE OR REPLACE FUNCTION enqueue_workflow(
```
stevendanna: Consistent ordering of name and workflow name

lib/workflow/postgres/sql_schema.go
```diff
@@ -0,0 +1,309 @@
...
+        completed_tasks = _completed_tasks + (select count(*) from unclaimed_tasks);
+
+    DELETE FROM workflow_events WHERE id = eid;
+
+    INSERT INTO workflow_events(event_type, workflow_instance_id)
+        VALUES('tasks_abandoned', _workflow_instance_id);
+$$ LANGUAGE SQL;
+
+
+CREATE OR REPLACE FUNCTION cancel_workflow(workflow_instance_id BIGINT)
```
stevendanna: This is a redefinition of the function at line 197 but is wrong, let's remove it!

lib/workflow/postgres/sql_schema.go
```diff
@@ -0,0 +1,309 @@
...
+RETURNS VOID
+LANGUAGE SQL
+AS $$
+    UPDATE workflow_instances SET updated_at = NOW(), payload = _payload,
+        enqueued_tasks = _enqueued_tasks, completed_tasks = _completed_tasks WHERE id = wid;
+    -- We've decided there is more to do but are done processing this event.
+    DELETE FROM workflow_events WHERE id = eid
+$$;
+
+-- (jaym): I think we can just have 1 notification channel for workflow events
```
jaym: delete me

lib/workflow/postgres/postgres.go
```diff
@@ -0,0 +1,726 @@
...
+	getTaskResultQuery = `SELECT task_name, parameters, status, error, result from tasks_results WHERE id = $1`
```
stevendanna: s/from/FROM/

lib/workflow/postgres/postgres.go
```diff
@@ -0,0 +1,726 @@
...
+	// wid is the WorkflowInstance id in our postgresql database.  We need
+	// this to complete the correct workflow
+	wid int64
+	// eid is the WorkflowEvent id
+	eid int64
+
+	// tx is the transaction that is holding our dequeued workflow.
+	tx *sql.Tx
+	// ctx is the context that the transaction was started with.
+	ctx context.Context
```
stevendanna: Currently we store the ctx, the cancellation func, and the tx.  Unfortunately, because of bugs in the pq library, we don't think it is safe to refer on simply cancelling this context and thus we should probably be handling rollback ourselves.

lib/workflow/postgres/postgres.go
```diff
@@ -0,0 +1,726 @@
...
+		return err
+	}
+
+	m, err := migrate.NewWithInstance("built-in schema", sourceDrv, dbName, dbInstance)
+	if err != nil {
+		return err
+	}
+	// defer m.Close() I don't think we want to call close here because it'll close our db instance
+
+	err = m.Up()
```
stevendanna: We probably need to do something like this:

```golang
	version, dirty, err := m.Version()
	if err != nil && err != migrate.ErrNilVersion {
		return errors.Wrap(err, "init migrator - error getting migration version")
	}

	if dirty {
		// force to prior version to reattempt migration
		err := m.Force(int(version) - 1)
		if err != nil {
			return errors.Wrap(err, "force to working schema version")
		}
		logrus.Infof("Forced to previous version: %v to reattempt migration", int(version)-1)
	} else {
		logrus.Infof("Current schema version: %v", version)
	}
```

lib/workflow/postgres/postgres.go
```diff
@@ -0,0 +1,726 @@
...
+
+	err := row.Scan(&data)
+	if err != nil {
+		return nil, err
+	}
+
+	return data, nil
+}
+
+func (pg *PostgresBackend) GetScheduledWorkflowRecurrence(ctx context.Context, scheduleName string, workflowName string) (string, error) {
```
stevendanna: We can collapse this and the previous function into `GetScheduleByName`

lib/workflow/postgres/postgres.go
```diff
@@ -0,0 +1,726 @@
...
+			&scheduledWorkflow.Recurrence,
+			&scheduledWorkflow.NextDueAt,
+			&scheduledWorkflow.LastStart,
+			&scheduledWorkflow.LastEnd,
+		)
+		if err != nil {
+			return nil, errors.Wrap(err, "could not read workflow schedules")
+		}
+		schedules = append(schedules, &scheduledWorkflow)
+	}
```
stevendanna: Here we need to check rows.Err() since Next() returns false on the end of the results AND on an error.

lib/workflow/postgres/postgres.go
```diff
@@ -0,0 +1,726 @@
...
+		return err
+	}
+
+	// Lock row to update
+	r := tx.QueryRow("SELECT id FROM recurring_workflow_schedules WHERE name = $1 AND workflow_name = $2 FOR UPDATE",
+		scheduleName, workflowName)
+	var id int64
+	if err := r.Scan(&id); err != nil {
+		if err != sql.ErrNoRows {
+			// TODO(jaym) return not found
```
jaym: doit

lib/workflow/postgres/postgres.go
```diff
@@ -0,0 +1,726 @@
...
+		_, err := tx.Exec("SELECT update_recurring_workflow_recurrence($1, $2, $3)",
+			id, o.Recurrence, o.NextRunAt)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// TODO(ssd) 2019-05-13: We need to decide on what "update" will look like
```
stevendanna: TODONE, see above.

lib/workflow/postgres/postgres.go
```diff
@@ -0,0 +1,726 @@
...
+		scheduleName, workflowName, parameters, recurrence, enabled, nextRunAt)
+	if err != nil {
+		return wrapErr(err, "could not update workflow schedule")
+	}
+
+	return wrapErr(tx.Commit(), "failed to commit workflow schedule update")
+}
+
+func (pg *PostgresBackend) EnqueueWorkflow(ctx context.Context, w *backend.WorkflowInstance) error {
+	wrapErr := func(err error, msg string) error {
```
jaym: unwrapErr

lib/workflow/postgres/postgres.go
```diff
@@ -0,0 +1,726 @@
...
+		cancel()
+		return nil, nil, errors.Wrap(err, "failed to dequeue workflow")
+	}
+
+	stmt, err := tx.Prepare(enqueueTaskQuery)
+	if err != nil {
+		cancel()
+		return nil, nil, err
+	}
+	// TODO: allow multiple workflow names
```
stevendanna: TODONE

lib/workflow/postgres/postgres.go
```diff
@@ -0,0 +1,726 @@
...
+	if event.Type == backend.TaskComplete {
+		event.CompletedTaskCount++
+		row := tx.QueryRowContext(ctx, getTaskResultQuery, taskResultID)
+		tr := backend.TaskResult{}
+		err = row.Scan(
+			&tr.TaskName, &tr.Parameters, &tr.Status, &tr.ErrorText, &tr.Result)
+		if err != nil {
+			cancel()
+			// FIXME FIXME FIXME
+			// TODO is this an infinite loop (this is panic situation)
```
stevendanna: We've convinced ourselves this is actually the right thing to do.  But we should wrap this error and probably log a warning.

lib/workflow/postgres/postgres.go
```diff
@@ -0,0 +1,726 @@
...
+
+	tid, task, err := pg.dequeueTask(tx, taskName)
+	if err != nil {
+		cancel()
+		return nil, nil, err
+	}
+	taskc.tid = tid
+
+	// pinger
+	go func() {
```
stevendanna: Let's move this into a named function so it is easier to find when we inevitably are reading through dlv output.

PR #352, opened 2019-05-15
Fixing the incorrect use of the gomock library

lancewf:
When using the goMock library the `Controller. Finish()` needs to be called when using `Times`, `MaxTimes`, or `MinTimes` functions.


phiggins:
I know I was the one suggesting using `Times` in the first place, but I'm starting to think just using a counter variable is better if its less likely to turn into a silent failure.

lancewf:
Yeah, I can go either way.

PR #353, opened 2019-05-15
Rule Mapping Schema and Database Functionality

tylercloke:

Wire in all database calls and schema we'll need for the Rule Mapping API.

Run `postgres_test.go`

- [ ] Code actually executed?


components/authz-service/storage/v2/memstore/memstore.go
```diff
@@ -240,6 +240,31 @@ func (s *State) GetPolicyChangeNotifier(ctx context.Context) (v2.PolicyChangeNot
 	return notifier, nil
 }
 
+func (s *State) CreateRule(_ context.Context, rule *storage.Rule) (*storage.Rule, error) {
+	// TODO implement
```
tylercloke: These will get implemented in the API stories.

components/authz-service/storage/errors.go
```diff
@@ -28,6 +28,10 @@ var (
 	// ErrMaxProjectsExceeded indicates that a new project cannot be created
 	// since the max allowed are already created.
 	ErrMaxProjectsExceeded = errors.New("max projects allowed")
+
+	// ErrChangeProjectForRule indicates that an updated attempted to change
```
blakestier: I'm not sure what this comment is supposed to mean 🤔 .

components/authz-service/storage/postgres/migration/sql/44_policy_rules.up.sql
```diff
@@ -0,0 +1,71 @@
...
+  id TEXT NOT NULL UNIQUE,
+  project_id TEXT REFERENCES iam_projects ON DELETE CASCADE,
+  name TEXT NOT NULL,
+  type iam_rule_type NOT NULL
+);
+
+CREATE TABLE iam_rule_conditions (
+  db_id SERIAL PRIMARY KEY,
+  rule_id INTEGER REFERENCES iam_project_rules ON DELETE CASCADE,
+  value TEXT NOT NULL,
```
blakestier: I thought value could sometimes be an array, depending on the operator. We should confirm.

components/authz-service/storage/v2/conditionattribute.go
```diff
@@ -0,0 +1,76 @@
+package v2
```
blakestier: The name of this file seems odd. Should it be train-cased?

components/authz-service/storage/v2/conditionattribute.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"fmt"
+	"strings"
+)
+
+// ConditionAttribute is an enum of attributes a project rule condition can be.
+type ConditionAttribute int
+
+const (
+	Organization ConditionAttribute = iota
```
blakestier: Could we alphabetize these? 😬 🤓

components/authz-service/storage/v2/conditionattribute.go
```diff
@@ -0,0 +1,76 @@
...
+
+func (c ConditionAttribute) String() string {
+	if c < Organization || c > PolicyGroup {
+		panic(fmt.Sprintf("unknown value from iota ConditionAttribute on String() conversion: %d", c))
+	}
+
+	return conditionAttributeStringValues()[c]
+}
+
+// NewConditionAttribute converts a string to an ConditionAttribute or returns an error.
```
blakestier: [nit] typo in the comment (s/an/a)

components/authz-service/storage/v2/conditionoperator.go
```diff
@@ -0,0 +1,56 @@
...
+	case "member-of":
+		return MemberOf, nil
+	default:
+		return Equals,
+			fmt.Errorf("condition operator type must be one of %q, you passed %s",
+				conditionOperatorStringValues(), in)
+	}
+}
+
+// UnmarshalJSON implements json unmarshalling for an RuleType reference
```
blakestier: same here

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -950,6 +950,211 @@ func (p *pg) insertRoleWithQuerier(ctx context.Context, role *v2.Role, q Querier
...
+			WHERE id = $3 AND projects_match_for_rule(project_id, $4) RETURNING db_id, project_id`,
+		rule.Name, rule.Type.String(), rule.ID, pq.Array(projectsFilter))
+	var ruleDbID string
+	var projectID string
+	if err := row.Scan(&ruleDbID, &projectID); err != nil {
+		return nil, p.processError(err)
+	}
+
+	// If they tried to change the project_id, abort the transaction with an error.
+	if projectID != rule.ProjectID {
```
blakestier: The order of operations feels awkward, here, but I see how you've added this flow to simplify fetching/validating the project_id. Can't think of a better way, though. 🤷‍♂

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3307,6 +3307,638 @@ func TestUpdatePolicy(t *testing.T) {
...
+			resp, err := store.CreateRule(ctx, &rule)
+			assert.Nil(t, resp)
+			assert.Error(t, storage_errors.ErrForeignKey, err)
+		},
+		"cannot use improper condition attributes for events": func(t *testing.T) {
+			_, err := storage.NewCondition(storage.Event, "chef-server-1", storage.ChefTag, storage.Equals)
+			assert.Error(t, err)
+		},
+		"creating a rule with no conditions returns an error": func(t *testing.T) {
+			_, err := storage.NewRule("new-id-1", "project-1", "name", storage.Node, []storage.Condition{})
```
blakestier: Based on the first test, it looks like this test could also fail because the project doesn't exist, correct?

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3307,6 +3307,638 @@ func TestUpdatePolicy(t *testing.T) {
...
+			assert.Error(t, err)
+		},
+		"creating a rule with inconsistent child condition type returns an error": func(t *testing.T) {
+			ruleType := storage.Node
+			differentRuleType := storage.Event
+			condition1, err := storage.NewCondition(ruleType,
+				"chef-server-1", storage.ChefServer, storage.Equals)
+			require.NoError(t, err)
+			condition2, err := storage.NewCondition(differentRuleType,
+				"org1,org2,org3", storage.Organization, storage.MemberOf)
```
blakestier: it's a string? not an array of strings? 🤔

blakestier: do we want to do validation on the existence of the organizations?

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3307,6 +3307,638 @@ func TestUpdatePolicy(t *testing.T) {
...
+				"role1", storage.ChefRole, storage.Equals)
+			require.NoError(t, err)
+
+			_, err = storage.NewRule("new-id-1", "project-1", "name", ruleType,
+				[]storage.Condition{condition1, condition2, condition3})
+			require.Error(t, err)
+		},
+		"create node rule with multiple conditions": func(t *testing.T) {
+			projID := "project-1"
+			insertTestProject(t, db, projID, "let's go jigglypuff - topsecret", storage.Custom)
```
blakestier: jigglypuuuuuuuff

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3307,6 +3307,638 @@ func TestUpdatePolicy(t *testing.T) {
...
+
+	cases := map[string]func(*testing.T){
+		"when no rules exist, returns an empty list": func(t *testing.T) {
+			ctx := context.Background()
+			resp, err := store.ListRules(ctx)
+			assert.NoError(t, err)
+			assert.Nil(t, resp)
+			assert.Equal(t, 0, len(resp))
+		},
+		"when mutiple rules exist with no project fitler, returns the full list": func(t *testing.T) {
```
blakestier: typo

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3307,6 +3307,638 @@ func TestUpdatePolicy(t *testing.T) {
...
+			ctx := context.Background()
+
+			projID := "project-1"
+			insertTestProject(t, db, projID, "let's go jigglypuff - topsecret", storage.Custom)
+
+			ruleType := storage.Node
```
blakestier: normally I'm all for redundancy in tests but this setup is getting dizzying. Worth pulling into a helper IMO

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3307,6 +3307,638 @@ func TestUpdatePolicy(t *testing.T) {
...
+			condition4, err := storage.NewCondition(ruleType,
+				"chef-server-2", storage.ChefServer, storage.Equals)
+			rule2, err := storage.NewRule("new-id-2", projID, "name2", ruleType,
+				[]storage.Condition{condition4})
+			require.NoError(t, err)
+			_, err = store.CreateRule(ctx, &rule2)
+			require.NoError(t, err)
+
+			resp, err := store.ListRules(ctx)
+			fmt.Println("wtf")
```
blakestier: 😂

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3307,6 +3307,638 @@ func TestUpdatePolicy(t *testing.T) {
...
+			condition4, err := storage.NewCondition(ruleType,
+				"chef-server-2", storage.ChefServer, storage.Equals)
+			rule2, err := storage.NewRule("new-id-2", projID2, "name2", ruleType,
+				[]storage.Condition{condition4})
+			require.NoError(t, err)
+			_, err = store.CreateRule(ctx, &rule2)
+			require.NoError(t, err)
+
+			resp, err := store.ListRules(ctx)
+			fmt.Println("wtf")
```
blakestier: 😂 😂

components/authz-service/storage/v2/ruletype.go
```diff
@@ -0,0 +1,56 @@
...
+type RuleType int
+
+const (
+	Node RuleType = iota
+	Event
+)
+
+func (r RuleType) String() string {
+	if r < Node || r > Event {
+		panic(fmt.Sprintf("unknown value from iota RuleType on String() conversion: %d", r))
```
blakestier: Do we have to panic here bc `String()` doesn't return an error?

components/authz-service/storage/v2/conditionattribute.go
```diff
@@ -0,0 +1,76 @@
...
+func NewConditionAttribute(in string) (ConditionAttribute, error) {
+	switch in {
+	case "organization":
```
blakestier: I wonder if it's worth pulling these into a hash (instead of `conditionAttributeStringValues`) and then checking for the keys/returning the node, erring if that doesn't succeed. It would get us closer to having some constants around the strings. 💭

components/authz-service/storage/postgres/migration/sql/44_policy_rules.up.sql
```diff
@@ -0,0 +1,71 @@
...
+  db_id SERIAL PRIMARY KEY,
+  id TEXT NOT NULL UNIQUE,
+  project_id TEXT REFERENCES iam_projects ON DELETE CASCADE,
+  name TEXT NOT NULL,
+  type iam_rule_type NOT NULL
+);
+
+CREATE TABLE iam_rule_conditions (
+  db_id SERIAL PRIMARY KEY,
+  rule_id INTEGER REFERENCES iam_project_rules ON DELETE CASCADE,
```
blakestier: is this the rule `db_id`? Can we name it `rule_db_id` in that case?

components/authz-service/storage/postgres/migration/sql/44_policy_rules.up.sql
blakestier: I see later that we're using a string-- I'm curious about the choice if we have time to chat about that!

components/authz-service/storage/postgres/migration/sql/44_policy_rules.up.sql
```diff
@@ -0,0 +1,71 @@
+BEGIN;
+
+CREATE TYPE iam_rule_type AS ENUM ('node', 'event');
+CREATE TYPE iam_attributes AS ENUM ('organization', 'chef-server',
+  'environment', 'role', 'chef-tag', 'policy-name', 'policy-group');
+CREATE TYPE iam_condition_operator AS ENUM ('equals', 'member-of');
+
+CREATE TABLE iam_project_rules (
+  db_id SERIAL PRIMARY KEY,
+  id TEXT NOT NULL UNIQUE,
```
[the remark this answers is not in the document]
blakestier: Ok @susanev updated me a bit on the background behind these decisions (name/description/id) changes. Totally 👍 on id and db_id

components/authz-service/storage/v2/condition.go
```diff
@@ -0,0 +1,63 @@
...
+
+func validateConditionInputs(value string, attribute ConditionAttribute, ruleType RuleType) error {
+
+	if value == "" {
+		return errors.New("a condition must have a value")
+	}
+
+	if ruleType == Event {
+		if attribute != Organization && attribute != ChefServer {
+			return errors.New("rules of type Event only accepts Conditions with attributes Organization or ChefServer")
```
msorens: s/accepts/accept/ for grammar

components/authz-service/storage/v2/conditionattribute.go
```diff
@@ -0,0 +1,76 @@
...
+	case "policy-group":
+		return PolicyGroup, nil
+	default:
+		return Organization,
+			fmt.Errorf("condition attribute type must be one of %q, you passed %s",
+				conditionAttributeStringValues(), in)
+	}
+}
+
+// UnmarshalJSON implements json unmarshalling for an RuleType reference
```
msorens: s/an RuleType/a ConditionAttribute/

components/authz-service/storage/v2/conditionoperator.go
```diff
@@ -0,0 +1,56 @@
+package v2
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ConditionOperator is an enum of operator a project rule condition can have.
```
msorens: s/operator/operators/

components/authz-service/storage/v2/conditionoperator.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (c ConditionOperator) String() string {
+	if c < Equals || c > MemberOf {
+		panic(fmt.Sprintf("unknown value from iota ConditionOperator on String() conversion: %d", c))
+	}
+
+	return conditionOperatorStringValues()[c]
+}
+
+// NewConditionOperator converts a string to an ConditionOperator or returns an error.
```
msorens: s/an/a/

components/authz-service/storage/v2/conditionoperator.go
```diff
@@ -0,0 +1,56 @@
...
+			fmt.Errorf("condition operator type must be one of %q, you passed %s",
+				conditionOperatorStringValues(), in)
+	}
+}
+
+// UnmarshalJSON implements json unmarshalling for an RuleType reference
+// so we can pull them out of the database directly as the correct type.
+func (c *ConditionOperator) UnmarshalJSON(b []byte) error {
+	// After byte conversion, things coming out of db as
+	// '"node"' and '"event"'.
```
msorens: copy/paste error => equals, member-of

components/authz-service/storage/postgres/migration/sql/44_policy_rules.up.sql
```diff
@@ -0,0 +1,71 @@
...
+CREATE TABLE iam_rule_conditions (
+  db_id SERIAL PRIMARY KEY,
+  rule_id INTEGER REFERENCES iam_project_rules ON DELETE CASCADE,
+  value TEXT NOT NULL,
+  attribute iam_attributes NOT NULL,
+  operator iam_condition_operator NOT NULL
+);
+
+CREATE OR REPLACE FUNCTION
+  projects_match_for_rule(_project_id TEXT, _projects_filter TEXT[])
```
msorens: Function already exists as `projects_match`, so delete this function and just use `projects_match` in the query functions; see my next comment.

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -950,6 +950,211 @@ func (p *pg) insertRoleWithQuerier(ctx context.Context, role *v2.Role, q Querier
...
+
+	tx, err := p.db.BeginTx(ctx, nil)
+	if err != nil {
+		return nil, p.processError(err)
+	}
+
+	row := tx.QueryRowContext(ctx,
+		`UPDATE iam_project_rules SET (name, type) = ($1, $2)
+			WHERE id = $3 AND projects_match_for_rule(project_id, $4) RETURNING db_id, project_id`,
+		rule.Name, rule.Type.String(), rule.ID, pq.Array(projectsFilter))
```
msorens: Please see https://chefio.atlassian.net/browse/A2-789. I think you were around when we discussed that card. The piece missing here is to add a `RowsAffected` check after the operation to have an explicit check for returning a not found for project filtering.
(Same thing on get operation--you've already got it on delete.)

components/authz-service/storage/v2/ruletype.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	return ruleTypeStringValues()[r]
+}
+
+// NewRuleType converts a string to an RuleType or returns an error.
```
msorens: s/an/a/

components/authz-service/storage/v2/conditionoperator.go
msorens: Also s/RuleType/ConditionOperator/

components/authz-service/storage/postgres/migration/sql/44_policy_rules.up.sql
```diff
@@ -0,0 +1,71 @@
...
+CREATE TYPE iam_attributes AS ENUM ('organization', 'chef-server',
+  'environment', 'role', 'chef-tag', 'policy-name', 'policy-group');
+CREATE TYPE iam_condition_operator AS ENUM ('equals', 'member-of');
+
+CREATE TABLE iam_project_rules (
+  db_id SERIAL PRIMARY KEY,
+  id TEXT NOT NULL UNIQUE,
+  project_id TEXT REFERENCES iam_projects ON DELETE CASCADE,
+  name TEXT NOT NULL,
+  type iam_rule_type NOT NULL
```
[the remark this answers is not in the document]
blakestier: oh yes, thanks @msorens :ignore_me:

srenatus:
> Run `postgres_test.go`

🤔 that's a little sparse, isn't it? 😉

components/authz-service/storage/v2/conditionattribute.go
```diff
@@ -0,0 +1,76 @@
...
+	Environment
+	ChefRole
+	ChefTag
+	PolicyName
+	PolicyGroup
+)
+
+func (c ConditionAttribute) String() string {
+	if c < Organization || c > PolicyGroup {
+		panic(fmt.Sprintf("unknown value from iota ConditionAttribute on String() conversion: %d", c))
```
srenatus: This can never happen. Let's remove L22-24.

*Update*: [you can make it happen](https://play.golang.org/p/_lF3iyHow3B). And that fact makes me sad.

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3307,6 +3307,638 @@ func TestUpdatePolicy(t *testing.T) {
...
+			condition3, err := storage.NewCondition(ruleType,
+				"role1", storage.ChefRole, storage.Equals)
+			require.NoError(t, err)
+
+			rule, err := storage.NewRule("new-id-1", "project-1", "name", ruleType,
+				[]storage.Condition{condition1, condition2, condition3})
+			require.NoError(t, err)
+			resp, err := store.CreateRule(ctx, &rule)
+			require.NoError(t, err)
+			require.Equal(t, &rule, resp)
```
srenatus: This assertion is not particularly interesting. I'd propose we query the database, and see what has been stored.

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3307,6 +3307,638 @@ func TestUpdatePolicy(t *testing.T) {
...
+			condition3, err := storage.NewCondition(ruleType,
+				"chef-server-2,chef-server-3", storage.ChefServer, storage.MemberOf)
+			require.NoError(t, err)
+
+			rule, err := storage.NewRule("new-id-1", "project-1", "name", ruleType,
+				[]storage.Condition{condition1, condition2, condition3})
+			require.NoError(t, err)
+			resp, err := store.CreateRule(ctx, &rule)
+			require.NoError(t, err)
+			require.Equal(t, &rule, resp)
```
srenatus: [same here]

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3307,6 +3307,638 @@ func TestUpdatePolicy(t *testing.T) {
...
+			condition2, err := storage.NewCondition(ruleType,
+				"org1,org2,org3", storage.Organization, storage.MemberOf)
+			require.NoError(t, err)
+			condition3, err := storage.NewCondition(ruleType,
+				"chef-server-2,chef-server-3", storage.ChefServer, storage.MemberOf)
+			require.NoError(t, err)
+			rule1, err := storage.NewRule("new-id-1", projID, "name", ruleType,
+				[]storage.Condition{condition1, condition2, condition3})
+			require.NoError(t, err)
+			_, err = store.CreateRule(ctx, &rule1)
```
srenatus: ⚠️ The test for `ListRules` now depends on `CreateRule`. Let's try to avoid that and do a straightforward DB query to set the stage database-wise for these tests.

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3307,6 +3307,638 @@ func TestUpdatePolicy(t *testing.T) {
...
```
srenatus: [same here]

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3307,6 +3307,638 @@ func TestUpdatePolicy(t *testing.T) {
...
+			condition2, err := storage.NewCondition(ruleType,
+				"org1,org2,org3", storage.Organization, storage.MemberOf)
+			require.NoError(t, err)
+			condition3, err := storage.NewCondition(ruleType,
+				"role1", storage.ChefRole, storage.Equals)
+			require.NoError(t, err)
+			rule, err := storage.NewRule("new-id-1", projID, "name", ruleType,
+				[]storage.Condition{condition1, condition2, condition3})
+			require.NoError(t, err)
+			_, err = store.CreateRule(ctx, &rule)
```
srenatus: Let's not use `CreateRule` in the tests for `UpdateRule`, please. 😉

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3307,6 +3307,638 @@ func TestUpdatePolicy(t *testing.T) {
...
+			condition2, err := storage.NewCondition(ruleType,
+				"org1,org2,org3", storage.Organization, storage.MemberOf)
+			require.NoError(t, err)
+			condition3, err := storage.NewCondition(ruleType,
+				"role1", storage.ChefRole, storage.Equals)
+			require.NoError(t, err)
+			rule, err := storage.NewRule("new-id-1", "project-1", "name", ruleType,
+				[]storage.Condition{condition1, condition2, condition3})
+			require.NoError(t, err)
+			_, err = store.CreateRule(ctx, &rule)
```
srenatus: [same here]

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3307,6 +3307,638 @@ func TestUpdatePolicy(t *testing.T) {
...
+			require.NoError(t, err)
+			assertCount(t, 3, db.QueryRow(`SELECT count(*) FROM iam_rule_conditions`))
```
srenatus: 👏 that's (in spirit) what I'd add where I've criticized the assertions above.

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3307,6 +3307,638 @@ func TestUpdatePolicy(t *testing.T) {
...
+
+			condition4, err := storage.NewCondition(ruleType,
+				"new-chef-server", storage.ChefServer, storage.Equals)
+			ruleUpdated, err := storage.NewRule("new-id-1", projID, "name", ruleType,
+				[]storage.Condition{condition1, condition2, condition3, condition4})
+			require.NoError(t, err)
+			resp, err := store.UpdateRule(ctx, &ruleUpdated)
+			assert.Nil(t, err)
```
srenatus:
```suggestion
			assert.NoError(t, err)
```

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3307,6 +3307,638 @@ func TestUpdatePolicy(t *testing.T) {
...
+			projID := "project-1"
+			insertTestProject(t, db, projID, "let's go jigglypuff - topsecret", storage.Custom)
+
+			ruleType := storage.Node
+			condition1, err := storage.NewCondition(ruleType,
+				"chef-server-1", storage.ChefServer, storage.Equals)
+			require.NoError(t, err)
+			rule, err := storage.NewRule("new-id-1", projID, "name", ruleType, []storage.Condition{condition1})
+			require.NoError(t, err)
+			_, err = store.CreateRule(ctx, &rule)
```
srenatus: same old same old, I'll not pick these everywhere, ok? 😉

components/authz-service/storage/v2/ruletype.go
srenatus: yup

components/authz-service/storage/postgres/migration/sql/44_policy_rules.up.sql
```diff
@@ -0,0 +1,71 @@
...
+);
+
+CREATE TABLE iam_rule_conditions (
```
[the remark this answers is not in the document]
tylercloke: I named it `iam_rule_conditions` to make it clear that the table is a one to many relationship between rules and conditions and that conditions are a concept that don't make sense outside of the context of rules.

components/authz-service/storage/postgres/migration/sql/44_policy_rules.up.sql
tylercloke: Not clear what the proper value is yet, but if we did wanna go the array route for `member-of` type conditions, we'll have to rethink the schema a bit, since that'd be different than what we wanted for `equals.

components/authz-service/storage/postgres/migration/sql/44_policy_rules.up.sql
tylercloke: They are a bit different:

```
CREATE OR REPLACE FUNCTION
  projects_match(_projects TEXT[], _projects_filter TEXT[])
  RETURNS BOOLEAN AS $$
    BEGIN
      RETURN (
        -- no projects filter requested (length 0) will be the case for v2.0 or v2.1 ["*"]
        array_length(_projects_filter, 1) IS NULL
        -- projects filter intersects with projects for row
        OR _projects && _projects_filter
        -- projects for row is an empty array, check if (unassigned) in project filter
        OR (array_length(_projects, 1) IS NULL AND '{(unassigned)}' && _projects_filter)
      );
    END
$$ LANGUAGE PLPGSQL;
```

```
CREATE OR REPLACE FUNCTION
  projects_match_for_rule(_project_id TEXT, _projects_filter TEXT[])
  RETURNS BOOLEAN AS $$
    BEGIN
      RETURN (
        array_length(_projects_filter, 1) IS NULL
        OR _project_id = ANY (_projects_filter)
      );
    END
$$ LANGUAGE PLPGSQL;
```

Rules can only have a single project which will not be in an array already like other objects. It will also never be unassigned so we don't wanna go down that logic path. Rules will always have a one and only one project (which in fact cannot change).

components/authz-service/storage/v2/conditionattribute.go
tylercloke: Ah I thought that wasn't idiomatic in go but looks like it's ok on further googling. Can do.

components/authz-service/storage/v2/conditionattribute.go
tylercloke: We want conditionAttributeStringValues to be a slice because that's how we convert between the enum and string (the enum is composed of integers that corresponds to slice indices).

components/authz-service/storage/v2/conditionattribute.go
tylercloke: I updated them to be alphabetic, but in general you don't wanna change the order of enums because the order matters, so if we ever add another one, it'll likely be un-alphabetized anyway.

components/authz-service/storage/v2/postgres/postgres.go
tylercloke: It actually behaves correctly (process error returns ErrNotFound), but I can add the check to make it more explicit.

tylercloke: We need it in delete because I'm doing an `ExecContext`. In get and update, we are using a `QueryRowContext` which will be converted to `storage_errors.ErrNotFound` correctly by `processError`. I have test coverage that indicates the correct thing is happening.

components/authz-service/storage/v2/postgres/postgres.go
tylercloke: It's just saving us a fetch query to get it back out of the delete, should be faster this way.

components/authz-service/storage/v2/postgres/postgres_test.go
tylercloke: a) It's not currently. We should talk about how we want that to behave for sure. I wasn't sure so I stuck with the simplest thing. Let's chat about that IRL.

b) Not for now. We aren't wired to talk to the chef servers API from our automate services which is where that source of truth lives.

components/authz-service/storage/v2/postgres/postgres_test.go
tylercloke: tenor-227982126

components/authz-service/storage/v2/postgres/postgres_test.go
tylercloke: I did that later in the tests but forgot to come back to these. Will add everywhere I missed.

components/authz-service/storage/v2/postgres/postgres_test.go
tylercloke: That adds so much complexity imo. I'm really not a big fan of all of our tests that do that. So much to refactor if you change anything.

components/authz-service/storage/v2/postgres/postgres_test.go
tylercloke: Open to suggestions on how that helper would look. There's some variance. Not a super clear pattern to me but I'm probably pretty in the weeds at this point.

components/authz-service/storage/v2/postgres/postgres_test.go
tylercloke: Mostly I'm not following the logic here. We have tests for CreateRule and UpdateRule. We are confident they work. If they don't those tests will break. If we aren't confident those work, we should expand those tests?

srenatus: If we were changing, say, the database layout in some of way, the only approach would be all-at-once. The different CRUD methods could not be worked on in increments or in parallel.

It's always a tradeoff, but this is a learning from previous difficult situations we had gotten  ourselves into. 😉

components/authz-service/storage/postgres/migration/sql/44_policy_rules.up.sql
```diff
@@ -0,0 +1,71 @@
+BEGIN;
+
+CREATE TYPE iam_rule_type AS ENUM ('node', 'event');
+CREATE TYPE iam_attributes AS ENUM ('organization', 'chef-server',
+  'environment', 'role', 'chef-tag', 'policy-name', 'policy-group');
+CREATE TYPE iam_condition_operator AS ENUM ('equals', 'member-of');
```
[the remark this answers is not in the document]
tylercloke: I'm fine with them for now, personally.

components/authz-service/storage/v2/condition.go
```diff
@@ -0,0 +1,63 @@
...
+	"encoding/json"
+
+	"github.com/pkg/errors"
+
+	storage_errors "github.com/chef/automate/components/authz-service/storage"
+)
+
+// Condition defines a condition for an ingest rule for a project.
+type Condition struct {
+	Value     string             `json:"value"`
```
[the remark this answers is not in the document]
tylercloke: We are still discussing if it should be a comma-separated string or  if we should change the schema. Will follow up here.

components/authz-service/storage/v2/postgres/postgres_test.go
tylercloke: After chatting with the team, I added precondition assertions so we felt confident that the calls to `CreateRule` weren't giving us false positives elsewhere. If you still feel super strongly, would it be ok if we merged as is and refactored so we aren't blocking the rest of the story?

components/authz-service/storage/postgres/migration/sql/44_policy_rules.up.sql
```diff
@@ -0,0 +1,71 @@
...
+      -- A rule can't exist without conditions so we don't need to worry
+      -- about null case here.
+      (SELECT json_agg(rc) AS conditions)
+      FROM iam_project_rules AS r
+      LEFT OUTER JOIN iam_rule_conditions
+      AS rc ON rc.rule_id=r.db_id WHERE id=_rule_id
+      AND projects_match_for_rule(project_id, _project_filter)
+      GROUP BY r.id, r.project_id, r.name, r.type)
+  SELECT row_to_json(t) AS rule FROM t;
+
```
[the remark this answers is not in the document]
msorens: Leaving out the change about projects_match, the line breaks here still need to get in, if you don't mind. 😁 

```
    FROM iam_project_rules AS r
    LEFT OUTER JOIN iam_rule_conditions AS rc
    ON rc.rule_id = r.db_id
    WHERE id = _rule_id AND AND projects_match_for_rule(r.project_id, _project_filter)
```

components/authz-service/storage/postgres/migration/sql/44_policy_rules.up.sql
```diff
@@ -0,0 +1,71 @@
...
+      -- A rule can't exist without conditions so we don't need to worry
+      -- about null case here.
+      (SELECT json_agg(rc) AS conditions)
+      FROM iam_project_rules AS r
+      LEFT OUTER JOIN iam_rule_conditions
+      AS rc ON rc.rule_id=r.db_id
+      WHERE projects_match_for_rule(project_id, _project_filter)
+      GROUP BY r.id, r.project_id, r.name, r.type)
+  SELECT row_to_json(t) AS rule FROM t;
+
```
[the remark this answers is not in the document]
msorens: Similar to above, need these line breaks, please:
```
  FROM iam_project_rules AS r
  LEFT OUTER JOIN iam_rule_conditions AS rc
  ON rc.rule_id = r.db_id
  WHERE projects_match_for_rule(r.project_id, _project_filter)
```

components/authz-service/storage/v2/postgres/postgres.go
msorens: Again, per A2-789, we agreed as a team that we want to be explicit on all-the-things (get, update, delete). If you want to re-open the discussion, that's cool, though.

PR #354, opened 2019-05-15
adjust the e2e scan job tests to fit with new navigation changes

vjeffrey:
with recent changes to the navigation, we need to update e2e tests


vjeffrey:
merging this so we can stop failing the deploy/dev jobs

PR #356, opened 2019-05-15
cloud integrations: if no name is defined for node when adding, use id

vjeffrey:
I noticed some nodes were getting added in with no name. 
When we query the Azure node state we get the name of the instance, but when we query aws node state we do not.  So this logic just ensures that if we are adding a node b/c it is new to us, and it does not have a name defined, we use the id as the name, so there's some way to identify the node

### :+1: Definition of Done
nodes discovered via aws instance check have the id as their name

### DIFF
--- a/components/nodemanager-service/pgdb/node-managers-cloud.go
+++ b/components/nodemanager-service/pgdb/node-managers-cloud.go
@@ -324,25 +324,29 @@ func (db *DB) UpdateOrInsertInstanceSourceStateInDb(instance InstanceState, mgrI
 	}
 	var id string
 	connectionErr := fmt.Sprintf("node state is %s", instanceState)
+	name := instance.Name
+	if len(name) == 0 {
+		name = instance.ID
+	}
 	switch instance.State {
 	case "terminated":
 		// if the instance state we're getting is terminated, we only want to update the node if we already
 		// have it in the system. it's silly to *add* nodes that are already terminated here.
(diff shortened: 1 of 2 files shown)


components/nodemanager-service/pgdb/node-managers-cloud.go
```diff
@@ -324,25 +313,27 @@ func (db *DB) UpdateOrInsertInstanceSourceStateInDb(instance InstanceState, mgrI
...
-	switch instance.State {
-	case "terminated":
-		// if the instance state we're getting is terminated, we only want to update the node if we already
-		// have it in the system. it's silly to *add* nodes that are already terminated here.
+	name := instance.Name
+	if len(name) == 0 {
+		name = instance.ID
+	}
+	switch instanceState {
+	case "TERMINATED", "STOPPED":
```
phiggins: I'm confused about why these changed from lowercase to uppercase...

vjeffrey: i was using instance.State before, but i actually translate instance.State into instanceState via this function earlier up in the function, so I figured I may as well use that: https://github.com/chef/automate/blob/master/components/nodemanager-service/pgdb/node-managers-cloud.go#L350

phiggins: Ahhh, I totally missed that. Thanks for the explanation.

phiggins: Would it make sense to add a default case to the switch that logs/fails with an unrecognized value?

PR #360, opened 2019-05-15
Adding project filter for GetEventTaskCounts

lancewf:

Adding project filtering for the event_feed.proto GetEventTaskCounts RPC.

https://github.com/chef/automate/issues/69

- [ ] Code actually executed?

### DIFF
--- a/components/config-mgmt-service/integration_test/event_task_counts_test.go
+++ b/components/config-mgmt-service/integration_test/event_task_counts_test.go
@@ -699,13 +699,7 @@ func TestEventTaskCountsProjectFilter(t *testing.T) {
 
 			// test response
 			assert.Equal(t, test.expected.Total, res.Total)
-
-			for _, expectedCount := range test.expected.Counts {
-				matchingResCount, err := findEventCountByName(expectedCount.Name, res.Counts)
-				assert.NoError(t, err)
-
-				assert.Equal(t, expectedCount.Count, matchingResCount.Count)
-			}
+			assert.ElementsMatch(t, test.expected.Counts, res.Counts)
 		})
 	}
 }
--- a/components/config-mgmt-service/integration_test/event_task_counts_test.go
+++ b/components/config-mgmt-service/integration_test/event_task_counts_test.go
@@ -6,11 +6,13 @@ import (
 	"testing"
 	"time"
 
+	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


components/config-mgmt-service/integration_test/event_task_counts_test.go
```diff
@@ -164,6 +166,561 @@ func TestEventTaskCountsCountOnlyFilteredChefServers(t *testing.T) {
...
+			// test response
+			assert.Equal(t, test.expected.Total, res.Total)
+
+			for _, expectedCount := range test.expected.Counts {
+				matchingResCount, err := findEventCountByName(expectedCount.Name, res.Counts)
+				assert.NoError(t, err)
+
+				assert.Equal(t, expectedCount.Count, matchingResCount.Count)
+			}
+		})
```
phiggins: I'm a little confused why the test cases build up a `response.EventCounts` struct and then only refer to its properties and not the struct itself in the test. That seems to couple the test implementation to the code's implementation unnecessarily.

phiggins: Also, because of how the test is structured you could have a passing test where the implementation included more information than was included in the expected data in the test case. Was that intentional?

lancewf: For the second comment. I only see that is the case if there are two counts for the same type returned. I think that should be tested somewhere else.

lancewf: For the second comment. I used the `response.EventCounts` because it matches the response closer. I would have like to have `assert.Equal(t, test.expected.Counts, res)` but I wanted to account for the `Counts` being in a different order.

phiggins: Is the order significant? Does `ElementsMatch` work for these?

lancewf: The order is not significant. I just assumed that `ElementsMatch` would not work because the `Count` is an array of structs, but now I want to test it.

lancewf: @phiggins It does work with `ElementsMatch`! Updating now. Thanks.

phiggins: Thank you for humoring my pedantry 😅

PR #361, opened 2019-05-15
Get project info from context in streaming endpoint.

phiggins:
For #66, followup to #318 

I thought it was strange that streaming endpoints wouldn't get access to the context object, and it turns out my instinct was correct; the `ServerStream` struct passed in has a `Context()` function we can use to get access to the projects. https://godoc.org/google.golang.org/grpc#ServerStream

### DIFF
--- a/components/compliance-service/api/reporting/server/server.go
+++ b/components/compliance-service/api/reporting/server/server.go
@@ -129,7 +129,11 @@ type exportHandler func(*reporting.Report) error
 
 // Export streams a json or csv export
 func (srv *Server) Export(in *reporting.Query, stream reporting.ReportingService_ExportServer) error {
-	formattedFilters := filterByProjectsFilter(formatFilters(in.Filters))
+	formattedFilters := formatFilters(in.Filters)
+	formattedFilters, err := filterByProjects(stream.Context(), formattedFilters)
+	if err != nil {
+		return err
+	}
 
 	if len(formattedFilters["profile_name"]) > 0 && len(formattedFilters["profile_id"]) > 0 {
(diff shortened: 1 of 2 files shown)


phiggins:
For reference, this was the ticket that gave me the clues I was missing https://github.com/grpc/grpc-go/issues/156

components/compliance-service/integration_test/reporting_server_export_test.go
```diff
@@ -136,10 +136,8 @@ func TestReportingServerExport(t *testing.T) {
 
 	for _, test := range successCases {
 		t.Run(test.description, func(t *testing.T) {
-			response, err := client.Export(ctx, &reporting.Query{
-				Filters: []*reporting.ListFilter{{Type: "projects", Values: test.allowedProjects}},
-				Type:    "json",
-			})
+			ctx := auth_context.NewOutgoingContext(contextWithProjects(test.allowedProjects))
```
phiggins: This was the other necessary change, because this test more closely simulates an actual request it needs to prepare the context like an actual network request would.

components/compliance-service/api/reporting/server/server.go
```diff
@@ -129,7 +129,11 @@ type exportHandler func(*reporting.Report) error
 
 // Export streams a json or csv export
 func (srv *Server) Export(in *reporting.Query, stream reporting.ReportingService_ExportServer) error {
-	formattedFilters := filterByProjectsFilter(formatFilters(in.Filters))
+	formattedFilters := formatFilters(in.Filters)
+	formattedFilters, err := filterByProjects(stream.Context(), formattedFilters)
```
vjeffrey: oh wow, nice find!

Issue #185, opened 2019-04-26
[EAS] Reconfigure page to make sidebar static


Shadae:
Jira link: https://chefio.atlassian.net/secure/RapidBoard.jspa?rapidView=314&projectKey=A2&modal=detail&selectedIssue=A2-768

Acceptance criteria:

The right hand side bar is persistent. it is visible when you load the apps tab and cannot be hidden

when I first load the applications tab, the first row on the left side table is selected and the right sidebar is populated with the right data for that row

persist the state of which row on the left is selected in the URL

persist the state of which filters are active on the right sidebar in the URL

persist the state of the pages on the right sidebar in the URL

when I select a new row on the left table, it resets the state of the sidebar things

susanev:
fixed in #367

PR #367, opened 2019-05-16
Reconfigure EAS dashboard to make sidebar static

scottopherson:

Reconfigures the *Applications* view to keep the sidebar permanently visible and to serialize its state to the URL.

### :+1: Definition of Done

- The right hand side bar is persistent. it is visible when you load the apps tab and cannot be hidden
- when I first load the applications tab, the first row on the left side table is selected and the right sidebar is populated with the right data for that row
- persist the state of which row on the left is selected in the URL
- persist the state of which filters are active on the right sidebar in the URL
- persist the state of the pages on the right sidebar in the URL
- when I select a new row on the left table, it resets the state of the sidebar things

- Nav to *Applications* view
- See sidebar remain permanently visible
- See sidebar filter/paging state serialized to URL query params

https://github.com/chef/automate/issues/185

- [x] Code actually executed?

### DIFF
--- a/components/automate-ui/src/app/entities/service-groups/service-groups.model.ts
+++ b/components/automate-ui/src/app/entities/service-groups/service-groups.model.ts
@@ -39,7 +39,7 @@ export interface ServiceGroupFilters {
 }
 
 export interface ServicesFilters {
-  service_group_id?: number;
+  service_group_id?: string;
   health: string;
   page: number;
   pageSize?: number;
--- a/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.ts
+++ b/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnDestroy, OnInit, Input, Output, EventEmitter } from '@angular/core';
+import { Component, OnDestroy, OnInit, Input } from '@angular/core';
 import { Store } from '@ngrx/store';
 import { NgrxStateAtom } from 'app/ngrx.reducers';
 import { Observable, Subject } from 'rxjs';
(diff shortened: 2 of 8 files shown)


components/automate-ui/src/app/entities/service-groups/service-groups.model.ts
```diff
@@ -39,7 +39,7 @@ export interface ServiceGroupFilters {
 }
 
 export interface ServicesFilters {
-  service_group_id?: number;
+  service_group_id?: string;
```
scottopherson: Existing interface uses string type: https://github.com/chef/automate/blob/master/components/automate-ui/src/app/entities/service-groups/service-groups.model.ts#L25 which is what the API returns ID as.

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.ts
```diff
@@ -45,22 +51,23 @@ export class ServicesSidebarComponent implements OnInit {
 
     this.svcHealthSummary$ = this.store.select(createSelector(serviceGroupState,
       (state) => state.servicesHealthSummary));
-    this.svcHealthSummary$.subscribe((servicesHealthSummary) => {
+    this.svcHealthSummary$.pipe(takeUntil(this.isDestroyed)).subscribe((servicesHealthSummary) => {
```
scottopherson: This component and sidebar component had a few subscriptions that never got closed when navigating away from *Applications* view.

https://github.com/chef/automate/pull/367/commits/54d183cbc920047c9e23c64f84e153fa5f3f310b
https://github.com/chef/automate/pull/367/commits/ecb68c7da1add60df64768ef9a8b960687449baf

PR #369, opened 2019-05-16
Removing dup function

lancewf:
Removing duplicate function

### DIFF
--- a/components/config-mgmt-service/integration_test/event_counts_test.go
+++ b/components/config-mgmt-service/integration_test/event_counts_test.go
@@ -3,7 +3,6 @@ package integration_test
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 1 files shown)


components/config-mgmt-service/integration_test/event_counts_test.go
```diff
@@ -659,17 +659,6 @@ func TestEventTypeCountsProjectFilter(t *testing.T) {
 	}
 }
 
-func findEventCountByName(eventName string,
```
lancewf: This function is here also https://github.com/chef/automate/blob/3ebc63c6506f4d570626a057594ef1a87e53d847/components/config-mgmt-service/integration_test/event_task_counts_test.go#L707

PR #376, opened 2019-05-16
[A2-839] Remove duplicates from IAM roles table in DB

msorens:

In a recent migration we used the convenient add-to-array syntax of PSQL:
```
actions = actions || '{ secrets:*:get, secrets:*:list }'
```
Unfortunately, that resulted in duplicates of those entries.
image

This PR remove the duplicates by resetting the list.

### :+1: Definition of Done

Above actions only appear once in the DB.

In the authz database:
```
SELECT * FROM iam_roles WHERE id = 'viewer';
```
From the command-line:
```
a2 -2 -o roles id viewer
```
or, in long-hand:
```
curl -sSkH "api-token: $TOK" "https://a2-dev.test/apis/iam/v2beta/roles" | \
jq -c '.roles[] | select (.id == "viewer")'
```

- [x] Code actually executed?

### DIFF
--- a/components/authz-service/storage/postgres/datamigration/sql/06_v2_roles.up.sql
+++ b/components/authz-service/storage/postgres/datamigration/sql/06_v2_roles.up.sql
@@ -1,0 +1,24 @@
+BEGIN;
+
+UPDATE iam_roles
+    SET
+        actions = '{
+            secrets:*:get,
+            secrets:*:list,
+            infra:*:get,
+            infra:*:list,
+            compliance:*:get,
+            compliance:*:list,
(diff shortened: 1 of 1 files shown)


blakestier:
This looks good though I'm not 100% clear on why the above syntax failed us . . .

msorens:
@blakestier The reason it failed is that it is incremental, rather than setting the entire list as this new migration does. It works fine for someone already on v2 (no duplicates), but if you do the upgrade-to-v2, those entries were put in by the upgrade code first, then again by this data migration that runs after the upgrade.

Issue #370, opened 2019-05-16
unable to filter by multiple tags when creating a scan job


vjeffrey:
# User Story

When creating a scan job, and filtering by tag values for the Automate node manager, we are currently unable to filter by multiple tags. 
This is confusing for the user, because we present them with the option to filter by multiple tag kvs but it doesn't work.
I was able to reproduce the issue in acceptance:
One node had tags:
`test-tag`, `is this`
`my-tag`, `val-1`
And another had tags:
`my-tag`, `val-1`

Attempting to filter on more than one tag resulted in an empty list

# Definition of Done
Able to filter by multiple tags

PR #377, opened 2019-05-16
nodes: filtering by multiple tags

vjeffrey:
fixes https://github.com/chef/automate/issues/370

When creating a scan job, and filtering by tag values for the Automate node manager, we are currently unable to filter by multiple tags.

### :+1: Definition of Done
Able to filter nodes by multiple tags (Automate nodemanager)

rebuild nodemanager
create some nodes with different tags
create a scan job filtering by those tags

### DIFF
--- a/components/nodemanager-service/pgdb/filtering.go
+++ b/components/nodemanager-service/pgdb/filtering.go
@@ -38,6 +38,18 @@ func mergeFilters(mergeableFilters []*common.Filter) ([]common.Filter, error) {
 	return filters, nil
 }
 
+func handleTagFilters(filter common.Filter) (string, error) {
+	tagKeyFilter := strings.TrimPrefix(filter.Key, "tags:")
+	newTagCondition, err := wherePatternMatchTags(tagKeyFilter, filter.Values, "t")
+	if err != nil {
+		return "", errors.Wrap(err, "buildWhereFilter error")
+	}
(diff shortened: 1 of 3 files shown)


components/nodemanager-service/pgdb/filtering.go
```diff
@@ -84,7 +101,15 @@ func buildWhereFilter(mergeableFilters []*common.Filter, tableAbbrev string, fil
 		conditions = append(conditions, newCondition)
 	}
 
-	whereFilter = fmt.Sprintf("WHERE (%s)", strings.Join(conditions, " AND "))
+	if len(oRConditionsForTags) == 0 {
+		whereFilter = fmt.Sprintf("WHERE (%s)", strings.Join(conditions, " AND "))
+	} else {
+		if len(conditions) > 0 {
```
vjeffrey: is this too messy? is there a cleaner way?

phiggins: The logic is getting a bit convoluted with special cases. I think you could simplify somewhat by separating all the tag handling from the other filters. If you moved all the tag filter handling first, that stuff could be `OR`ed together and added to the `conditions` as just another element, since those all get `AND`ed together at the end.

Hopefully that made sense, I was trying to avoid just writing a bunch of pseudocode.

PR #378, opened 2019-05-16
fix some variable package collisions

rickmarry:

- [ ] Code actually executed?

### DIFF
--- a/components/compliance-service/integration_test/stats_server_read_failures_test.go
+++ b/components/compliance-service/integration_test/stats_server_read_failures_test.go
@@ -3,7 +3,7 @@ package integration_test
 import (
 	"testing"
 
-	iam_v2 "github.com/chef/automate/api/interservice/authz/v2"
+	iamV2 "github.com/chef/automate/api/interservice/authz/v2"
 	"github.com/chef/automate/components/compliance-service/ingest/events/compliance"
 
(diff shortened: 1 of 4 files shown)


components/compliance-service/integration_test/stats_server_read_failures_test.go
```diff
@@ -3,7 +3,7 @@ package integration_test
 import (
 	"testing"
 
-	iam_v2 "github.com/chef/automate/api/interservice/authz/v2"
+	iamV2 "github.com/chef/automate/api/interservice/authz/v2"
```
vjeffrey: why? 
https://blog.golang.org/package-names

rickmarry: I'm a little bit torn here. I know that both snake and camel case are discouraged in package names.  
snake_case is more generally discouraged in go.
in a way, I think that just naming it iam is the way to go here.  what do you think?

rickmarry: I find myself needing to use multi-word aliases for packages from time to time.  snake_case and camelCase being antipatterns make this difficult.

vjeffrey: ya, in the go examples they do "syscall" for system call. but it can get messy looking. our package naming (and aliasing) is really all over the place in this codebase tbh

Issue #311, opened 2019-05-10
Don't show 0 count while loading


alexpop:
# User Story

As an Automate User, I would like the following UI pages not to show a value of 0 for counts while data from the backend in pulled in order to render them. It creates a situation where the page is visited and for a few seconds, the user will see 0 counts and then as data is rendered, the counts change.

**1. Asset Store > `Profiles` and `Available` counts**

![Screen Shot 2019-05-10 at 1 34 53 PM](https://user-images.githubusercontent.com/107378/57540472-b8fbf080-7344-11e9-9662-e80f0214e080.png)

------------

**2. Scan Jobs > `Scan jobs` and `Nodes added` counts**
<img width="908" alt="Screen Shot 2019-05-10 at 5 05 08 PM" src="https://user-images.githubusercontent.com/107378/57541028-03ca3800-7346-11e9-862b-6bfba167bd93.png">

-----------

This issue is not showing on the Compliance page for the `Nodes` and `Profiles` counts.

susanev:
@msorens created a pattern for page loading with an overlay.. i wonder if that might be useful here.
https://github.com/chef/a2/pull/4165

PR #379, opened 2019-05-17
Adding loading pattern to profiles page + hide 0 until after page loads

vinay033:

As an Automate User, I would like the following UI pages not to show a value of 0 for counts while data from the backend in pulled in order to render them. It creates a situation where the page is visited and for a few seconds, the user will see 0 counts and then as data is rendered, the counts change.
### :+1: Definition of Done
new changes video 
https://drive.google.com/file/d/131ST9sOB8xBg4LHQy1yMjg-g5pMrc2jg/view?usp=sharing
fixes https://github.com/chef/automate/issues/311


susanev:
@vinay033 could you make it so the overlay is just for the profiles page content and not the left-nav? otherwise, it looks good!

susanev:
@vinay033 ignore my previous comment, we can come back and fix that in a separate task.

components/automate-ui/src/app/pages/+compliance/+scanner/containers/scanner/scanner.component.ts
```diff
@@ -13,12 +13,22 @@ export class ScannerComponent implements OnInit {
   jobsCount$: Observable<number>;
   nodesCount$: Observable<number>;
 
+  // show spinner befor loading data
```
msorens: s/befor/before/

components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.ts
```diff
@@ -60,6 +60,11 @@ export class ProfileOverviewComponent implements OnInit, OnDestroy {
   // notification data
   downloadErrorVisible = false;
 
+  // show spinner befor loading data
```
msorens: s/befor/before/

vinay033: changes added :)

PR #380, opened 2019-05-17
[A2-835] make custom endpoints projects aware for iam v2.1

srenatus:

TODO

### :+1: Definition of Done

TBD

TODO

- [A2-835](https://chefio.atlassian.net/browse/A2-835)

- [ ] Code actually executed?


components/automate-gateway/gateway/middleware/authv2/authv2.go
```diff
@@ -83,13 +83,44 @@ func (c *client) Handle(ctx context.Context, subjects []string, projectsToFilter
...
-		Subjects: subjects,
-		Resource: resource,
-		Action:   action,
+	projectsToFilter []string) (middleware.AuthorizationResponse, error) {
+	log := ctxlogrus.Extract(ctx)
+	filteredResp, err := c.client.ProjectsAuthorized(ctx, &authz.ProjectsAuthorizedReq{
+		Subjects:       subjects,
+		Resource:       resource,
+		Action:         action,
+		ProjectsFilter: projectsToFilter,
```
tylercloke: This is looking similar to the `Handle` function. When do we hit `Handle` vs `IsAuthorized`?

srenatus:
ℹ️ That's the authorizer in a nutshell:

- It wraps an authorizer for v1 and one for v2; and exposes roughly the same methods that these do:
1. `Handle` is for use in the GRPC interceptor: it reads the method from the context, and retrieves the policy mappings (v1 or v2 respectively, in the separate authv{1,2} packages); and uses that to constrict an authz query to the respective service, interpreting the result accordingly. v2 will take care of injecting projects into the context
2. `IsAuthorized` is a (bolted-on) method for using the authorizer construct in our custom handlers: it needs to be passed the resource/action directly.
- The authorizer package wrapping the v1 and v2 authorizers takes care of **switching** between v1  and v2: it has the version in its state, and if the authorizer for v1 fails with FailedPrecondition, switches to v2, and vice-versa. This is how we've achieved that the gateway config doesn't have to be changed for IAM v2 upgrade, but authz-service is the single source of truth.
- To expose versioned error messages, the authorizer interface needs both v1 and v2 info; whereas the versioned authv{1,2} packages only take one. This is because there's no way for the `IsAuthorized` methods to retrieve the info themselves. 
- [The relevant types are defined here](https://github.com/chef/automate/blob/a2a442f16459f887d517bc43447a154d4b729b3b/components/automate-gateway/gateway/middleware/auth_interceptor.go#L40-L45)

components/automate-gateway/gateway/middleware/authv2/authv2.go
srenatus: Is it? [`Handle`](https://github.com/chef/automate/blob/a2a442f16459f887d517bc43447a154d4b729b3b/components/automate-gateway/gateway/middleware/authv2/authv2.go#L26-L84) is to be used in a GRPC interceptor, and depends on 
a) the grpc method invoked, and
b) the policy info being mapped in our proto files
i.e. [this half of the method](https://github.com/chef/automate/blob/a2a442f16459f887d517bc43447a154d4b729b3b/components/automate-gateway/gateway/middleware/authv2/authv2.go#L26-L55), neither of which exist for custom handlers.

I'll outline how the authorizer is supposed to work in another comment, give me a minute 😉

srenatus:
Everything would be nice and tidy if there was only GRPC (and if we rearranged how introspection works, maybe?). 😉 

Also: please note that I'm not particularly attached to how this currently works (I entirely blame my lack of imagination and not my tools) -- if anyone comes up with some better means to reach the same end, I'm all ears.

components/automate-gateway/gateway/middleware/auth_interceptor.go
```diff
@@ -42,7 +41,7 @@ type SwitchingAuthorizationHandler interface {
 	GRPCAuthorizationHandler
 	SwitchingFilterHandler
 	IsAuthorized(ctx context.Context, subjects []string,
-		resourceV1, actionV1, resourceV2, actionV2 string) (AnnotatedAuthorizationResponse, error)
+		resourceV1, actionV1, resourceV2, actionV2 string, projects []string) (AnnotatedAuthorizationResponse, error)
```
blakestier: I suppose this is the "last param" but oof we're passing around a lot of args at this point!

components/automate-gateway/gateway/middleware/auth_interceptor.go
```diff
@@ -182,7 +182,7 @@ func getProjectsFromMetadata(projectHeaderEntries []string) []string {
 	for _, entry := range projectHeaderEntries {
 		for _, project := range strings.Split(entry, ",") {
 			newProject := strings.TrimSpace(project)
-			if _, value := keys[newProject]; !value {
+			if !keys[newProject] {
```
blakestier: I'm sort of surprised this worked-- I would expect that if the value of `keys[newproject]` was nil, this would fail.

Don't we need to both check the value and the success of extracting it (`if val, ok := keys[newProject]; ok {`)?

Maybe I'm missing something 🤔

components/automate-gateway/gateway/middleware/authv2/authv2.go
```diff
@@ -83,13 +83,44 @@ func (c *client) Handle(ctx context.Context, subjects []string, projectsToFilter
...
 	})
+	if err != nil {
+		if status.Convert(err).Code() == codes.FailedPrecondition {
+			return nil, err
+		}
+		log.WithError(err).Error("error authorizing request")
+		return nil, status.Errorf(codes.PermissionDenied,
```
blakestier: Should this message/error also include projects?

components/automate-gateway/gateway/middleware/authv2/authv2.go
```diff
@@ -83,13 +83,44 @@ func (c *client) Handle(ctx context.Context, subjects []string, projectsToFilter
...
+		Subjects:       subjects,
+		Resource:       resource,
+		Action:         action,
+		ProjectsFilter: projectsToFilter,
 	})
+	if err != nil {
+		if status.Convert(err).Code() == codes.FailedPrecondition {
+			return nil, err
+		}
+		log.WithError(err).Error("error authorizing request")
```
blakestier: I'm curious about the choice to log here-- can we not guarantee that the error itself will surface in the logs? 

If that's the case, I imagine it would be helpful to have the subj/resource/etc in the log message as well as the error.

components/automate-gateway/gateway/middleware/auth_interceptor.go
msorens: `keys[newProject]`, if not present, will return the zero-value for the type, in this case `false`. So this statement is still asserting "if there is no newProject entry in keys then..."

components/automate-gateway/gateway/services.go
```diff
@@ -662,10 +665,21 @@ func (s *Server) authRequest(r *http.Request,
 	return authzResp.Err()
 }
 
-func logError(err error, resource, action string) {
-	log.WithFields(log.Fields{
```
blakestier: Did we move away from `WithFields`?

components/automate-gateway/gateway/services.go
```diff
@@ -662,10 +665,21 @@ func (s *Server) authRequest(r *http.Request,
...
-		"error":    err,
-		"resource": resource,
-		"action":   action,
-	}).Error("error authorizing request")
+// XXX needs headers, doesn't it?
+func getProjectsFromMetadata(projectHeaderEntries []string) []string {
+	if projectHeaderEntries == nil {
```
blakestier: 👍 but also 😢

components/automate-gateway/gateway/middleware/auth_interceptor.go
blakestier: @msorens I know that it reads that way, but I was under the impression that golang didn't necessarily behave how we'd expect here. I've tried this in a playground and I keep getting the error I'd expect (invalid operation)-- would you mind linking me to an example plaground?

components/automate-gateway/gateway/middleware/authv2/authv2.go
msorens: As this is the gatekeeper for both v2 and v2.1, and it does not know which for any given call, we cannot appropriately include projects.

components/automate-gateway/gateway/services.go
srenatus: Sorry that one should have been deleted -- I've done the refactor here, after all. 😄

components/automate-gateway/gateway/services.go
srenatus: Nope, just realised that the function hadn't been used 🤷‍♂

components/automate-gateway/gateway/middleware/auth_interceptor.go
srenatus: yeah -- the root cause is that this is used for authorizing requests in the custom endpoints' helper function -- and there's no way they can retrieve the v1 or v2 mappings (no grpc/proto files involved). So, it'll have to be presented up front.

components/automate-gateway/gateway/middleware/auth_interceptor.go
srenatus: https://play.golang.org/p/y1GXa4D4vdv does that do the trick? Go really isn't dynamic at all, but non-existing values in maps will return their zero value, as @msorens suggested. [Here's the spec section](https://golang.org/ref/spec#Index_expressions):

>  For a of map type M:
>
> - x's type must be assignable to the key type of M
> - if the map contains an entry with key x, a[x] is the map element with key x and the type of a[x] is the element type of M
> - if the map is nil or does not contain such an entry, **a[x] is the zero value for the element type of M**

components/automate-gateway/gateway/middleware/authv2/authv2.go
srenatus: 😄 a) I don't think I've spent much time deliberating the choice of log here.
b) *ANY* GRPC response that has a status !=`OK` will be logged by the gateway. So, we should end up having [most of the stuff in the logs because of the status message below](https://github.com/chef/automate/pull/380/files#diff-cecefbe2195758a5155eeb1d419ed6cfR115).

Good enough? 😉

components/automate-gateway/gateway/middleware/auth_interceptor.go
blakestier: Ugh thank you for the example-- I'd read the docs but was having trouble manually testing . . . I had my map constructed incorrectly (the value as a string, not a bool) 🙄

PR #382, opened 2019-05-17
Adding project filters for NodeExport

lancewf:

Adding project filtering for the cfgmgmt.proto NodeExport RPCs.

- [ ] Code actually executed?

### DIFF
--- a/components/config-mgmt-service/integration_test/node_export_test.go
+++ b/components/config-mgmt-service/integration_test/node_export_test.go
@@ -1,0 +1,498 @@
+package integration_test
+
+import (
+	// "bytes"
+	"context"
+	// "encoding/json"
+	"fmt"
+	"github.com/buger/jsonparser"
+	"github.com/chef/automate/api/interservice/cfgmgmt/request"
+	api "github.com/chef/automate/api/interservice/cfgmgmt/service"
+	authzConstants "github.com/chef/automate/components/authz-service/constants/v2"
+	"github.com/chef/automate/components/config-mgmt-service/backend/elastic"
+	"github.com/chef/automate/components/config-mgmt-service/config"
+	"github.com/chef/automate/components/config-mgmt-service/grpcserver"
+	iBackend "github.com/chef/automate/components/ingest-service/backend"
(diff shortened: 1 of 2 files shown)


components/config-mgmt-service/integration_test/node_export_test.go
```diff
@@ -0,0 +1,498 @@
...
+
+			actualNodeNames := make([]string, 0)
+			jsonparser.ArrayEach(data, func(node []byte, _ jsonparser.ValueType, _ int, err error) {
+				if err == nil {
+					nodeName, err := jsonparser.GetString(node, "name")
+					if err == nil {
+						actualNodeNames = append(actualNodeNames, nodeName)
+					}
+				}
+			})
```
phiggins: I'm not sure I understand this part. Do we not have a type for the response returned here?

components/config-mgmt-service/integration_test/node_export_test.go
```diff
@@ -0,0 +1,498 @@
...
+					break
+				}
+
+				require.NoError(t, err)
+				data = append(data, tdata.GetContent()...)
+			}
+
+			actualNodeNames := make([]string, 0)
+			jsonparser.ArrayEach(data, func(node []byte, _ jsonparser.ValueType, _ int, err error) {
+				if err == nil {
```
phiggins: I know this test is not trying to verify basic correctness of the endpoint, but it feels bad to ignore errors here...

components/config-mgmt-service/integration_test/node_export_test.go
lancewf: We do not have an exposed type. The type is here https://github.com/chef/automate/blob/master/components/config-mgmt-service/grpcserver/node_export.go#L31. 

I found it easier to just parse the JSON.

components/config-mgmt-service/integration_test/node_export_test.go
lancewf: That sounds right. I will update this test to fail on a returned error.

PR #383, opened 2019-05-17
Clean up data before test

blakestier:

This test was hitting our precondition that only 6 projects exist-- when we attempted to create data for this test, it collided with 4 existing projects on the `inplace` box.

I don't see those projects being created programmatically in our code, so I'm assuming someone did it manually? 

JIC, we're now blowing away existing projects before running the test. This should be less troublesome when we remove the precondition.


srenatus:
ℹ️ rebase, then merge.

srenatus:
Nah, won't rebase. Tested this locally against acceptance, and it works beautifully:

image

PR #385, opened 2019-05-17
[A2-763] Hide "Access Management" on sidebar

blakestier:

Previously, if logged in as an "Editor" on 2.0, we were seeing the "Access Management" title in the sidebar without any subcategories.

With this change, an Editor does not see the title on 2.0. However, on 2.1, the editor can see both the title and "Projects" underneath. 

On 2.0:
Screen Shot 2019-05-17 at 1 58 55 PM

On 2.1:
Screen Shot 2019-05-17 at 1 58 29 PM

### DIFF
--- a/components/automate-ui/src/app/components/admin-sidebar/admin-sidebar.component.html
+++ b/components/automate-ui/src/app/components/admin-sidebar/admin-sidebar.component.html
@@ -26,7 +26,7 @@
       <chef-sidebar-entry route="/settings/tokens" icon="vpn_key">API Tokens</chef-sidebar-entry>
     </app-authorized>
 
-    <div *ngIf="(iamMajorVersion$ | async) === 'v2'">
+    <div *ngIf="(iamMajorVersion$ | async) !== 'v1'">
       <div class="group" *ngIf="policies.visible || roles.visible || projects.visible && (iamMinorVersion$ | async) === 'v1'">Access Management</div>
       <app-authorized #policies [allOf]="['/iam/v2beta/policies', 'get']">
         <chef-sidebar-entry route="/settings/policies" icon="security">Policies</chef-sidebar-entry>
--- a/components/automate-ui/src/app/components/admin-sidebar/admin-sidebar.component.html
+++ b/components/automate-ui/src/app/components/admin-sidebar/admin-sidebar.component.html
@@ -27,7 +27,7 @@
     </app-authorized>
 
     <div *ngIf="(iamMajorVersion$ | async) !== 'v1'">
-      <div class="group" *ngIf="policies.visible || roles.visible || projects.visible && (iamMinorVersion$ | async) === 'v1'">Access Management</div>
+      <div class="group" *ngIf="policies.visible || roles.visible || (projects.visible && (iamMinorVersion$ | async) === 'v1')">Access Management</div>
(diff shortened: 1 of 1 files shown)


components/automate-ui/src/app/components/admin-sidebar/admin-sidebar.component.html
```diff
@@ -26,8 +26,8 @@
       <chef-sidebar-entry route="/settings/tokens" icon="vpn_key">API Tokens</chef-sidebar-entry>
     </app-authorized>
 
-    <div *ngIf="(iamMajorVersion$ | async) !== 'v1'">
-      <div class="group" *ngIf="policies.visible || roles.visible || projects.visible">Access Management</div>
+    <div *ngIf="(iamMajorVersion$ | async) === 'v2'">
```
blakestier: I could see the case for keeping this as a `!== v1` (if we bump to a 3?) . . . I just found the positive `===` so much easier to read 🤷‍♂

msorens: Better to keep the not-equal-v1 because if we go to v3, etc. we likely want it still to work.

components/automate-ui/src/app/components/admin-sidebar/admin-sidebar.component.html
```diff
@@ -26,8 +26,8 @@
...
+      <div class="group" *ngIf="policies.visible || roles.visible || projects.visible && (iamMinorVersion$ | async) === 'v1'">Access Management</div>
```
msorens: Seemed like this was being slightly redundant putting it here--why not just move the one on the chef-sidebar up to its parent app-authorized?--but as we discovered, doing that, and changing this to `projects?.visible` failed! Seems like the Angular reference variable (`projects`) is not quite like a real variable after all, so this is a good solution!

components/automate-ui/src/app/components/admin-sidebar/admin-sidebar.component.html
msorens: LOL -- just refreshed and saw your comment on the very same thing!

msorens: So I will add in that not-equal-v1 is also consistent with all the other uses in the front-end.

components/automate-ui/src/app/components/admin-sidebar/admin-sidebar.component.html
```diff
@@ -27,7 +27,7 @@
     </app-authorized>
 
     <div *ngIf="(iamMajorVersion$ | async) !== 'v1'">
-      <div class="group" *ngIf="policies.visible || roles.visible || projects.visible">Access Management</div>
+      <div class="group" *ngIf="policies.visible || roles.visible || projects.visible && (iamMinorVersion$ | async) === 'v1'">Access Management</div>
```
msorens: Sorry, got distracted and forgot one comment.
I would urge adding parentheses around the new expression `(projects.visible && (iamMinorVersion$ | async) === 'v1'")`. Should not rely on assumptions about operator precedence.

srenatus:
ℹ️ I'll rebase out of politeness and merge this when green.

Issue #326, opened 2019-05-13
narrowing suggestions for client runs searchbar


vjeffrey:
# User Story

In an effort to drive parity across the search-bars available in Automate, we would like to implement narrowing suggestions for the client runs searchbar. This functionality has already been implemented in the compliance reporting searchbar and has been well received by customers. 
"Narrowing suggestions" basically means including the already selected filters in the api request when calling the suggestions endpoint, so that the suggestions the user is presented with are valid ones.  For example, if a user selects Environment:Prod and there are no ubuntu nodes that match that environment value, they should not be presented with ubuntu as a possible platform filter choice.

# Definition of Done
Suggestions on client runs searchbar respects the already-selected filters.

PR #391, opened 2019-05-20
Client Runs Narrowing Suggestions

lancewf:

Add Client runs search bar suggestion narrowing. The suggestions that appear in the search bar should be filtered to only show values from the nodes currently visible on the page. This filters the suggestions with the filters already selected in the search bar. 

One exception is that the suggestions are not narrowed by the type they are for. For example, if a user is trying to add a new node name filter in the search bar the priorly added node name filters will not be used to narrow the values returned. This is because filters of the same type are or'ed together. Therefore the filters of the same type should not limit new filters.  

1. `build components/config-mgmt-service && build components/automate-gateway && build components/automate-ui`
1. `start_all_services`
1. Update 'components/ingest-service/examples/converge-success-report.json' node.chef_environment to 'A'
1. `send_chef_run_example`
1. Update 'components/ingest-service/examples/converge-success-report.json' node.chef_environment to 'B' and node_name to 'insights.B'
1. `send_chef_run_example`
1. Open `https://a2-dev.test/client-runs` and add a filter for 'environment' 'A'
1. In the search bar select type 'Node Name' and type 'insights'. 
1. Only the suggestion of 'insights.chef.co' should be seen. 

https://github.com/chef/automate/issues/326

- [x] Code actually executed?


components/config-mgmt-service/grpcserver/cfg_mgmt.go
```diff
@@ -315,6 +312,30 @@ func (s *CfgMgmtServer) GetSourceFqdns(ctx context.Context,
 	return sourceFqdns, err
 }
 
+func removeSuggestionTypeFromFilter(filter []string, suggestionType string) []string {
+	adjustedFilter := make([]string, 0)
+	for _, filter := range filter {
+		keyValuePair := strings.Split(filter, ":")
+
+		if len(keyValuePair) == 2 && keyValuePair[0] != suggestionType {
```
alexpop: What do we do with filters that have `:` in the value? `len(keyValuePair)` will be > 2 in that case, right?

lancewf: Good point. 🤔  I will create an issue to look into problem with search terms with `:` in them.

lancewf: @alexpop This has not been a problem yet because the value from the UI is URL encoded. So, "chef:v2" would be "chef%253Av2*". This would be a problem for an API user if they do not encode the value.

alexpop: Ok, thank you

PR #399, opened 2019-05-21
try to fix the pipeline

vjeffrey:
i haven't tested this but i think it should fix things.
sorry!


srenatus:
If noone else does, I'll get this merged tomorrow and 👀 the pipeline

PR #400, opened 2019-05-21
Fix a few issues with the policy membership UI

tylercloke:

* Fixed policy member add UI issue where it was incorrectly using local team names instead of IDs to add members to policy. This was resulting in being able to add "Viewers" to the viewers policy already containing the local team "viewers".
* Fixed policy membership table in policy details view to have the correct table column header of ID instead of Name.
* Fixed issue where you could get a console error if policy isn't loaded on time.
* Fixed TODO where we were populating links for local users and teams in policy details membership table.

Working links to local teams and users:

Screen Shot 2019-05-21 at 1 24 32 PM

Capital V "Viewers" no longer showing up and properly showing local team IDs instead of Names:

Screen Shot 2019-05-21 at 1 24 57 PM

- [ ] Code actually executed?

### DIFF
--- a/components/automate-ui/src/app/pages/policy/details/policy-details.component.html
+++ b/components/automate-ui/src/app/pages/policy/details/policy-details.component.html
@@ -40,23 +40,25 @@
         <div *ngIf="(members$ | async)?.length === 0" id="getting-started">Add the first members to get started!</div>
         <chef-toolbar>
           <chef-button primary
-            [routerLink]="['/settings', 'policies', policy.id, 'add-members']">Add Members</chef-button>
+            [routerLink]="['/settings', 'policies', policy?.id, 'add-members']">Add Members</chef-button>
         </chef-toolbar>
         <chef-table *ngIf="(members$ | async)?.length > 0">
           <chef-thead>
             <chef-tr>
-              <chef-th>Name</chef-th>
+              <chef-th>ID</chef-th>
               <chef-th>Type</chef-th>
               <chef-th></chef-th>
             </chef-tr>
           </chef-thead>
           <chef-tbody>
             <chef-tr *ngFor="let member of members$ | async">
-              <!-- TODO (tc) we should link to local users and teams once we switch those APIs to named IDs.
-              can't now since we don't have the GUIDs for them unless we queried them by name for each
-              row somehow. -->
               <chef-td>
-                {{ member.displayName }}
+                <a *ngIf="memberURLs.hasOwnProperty(member.name)" [routerLink]="memberURLs[member.name]" target="_blank">
+                  {{ member.displayName }}
+                </a>
+                <ng-container *ngIf="!memberURLs.hasOwnProperty(member.name)">
+                  {{ member.displayName }}
+                </ng-container>
               </chef-td>
               <chef-td>
                 {{ member.displayType }}
--- a/components/automate-ui/src/app/pages/policy/details/policy-details.component.html
+++ b/components/automate-ui/src/app/pages/policy/details/policy-details.component.html
@@ -53,7 +53,7 @@
           <chef-tbody>
             <chef-tr *ngFor="let member of members$ | async">
               <chef-td>
-                <a *ngIf="memberURLs.hasOwnProperty(member.name)" [routerLink]="memberURLs[member.name]" target="_blank">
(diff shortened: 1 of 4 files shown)


components/automate-ui/src/app/pages/policy/details/policy-details.component.ts
```diff
@@ -63,7 +68,14 @@ export class PolicyDetailsComponent implements OnInit, OnDestroy {
         this.policyJSON = this.policyToString(this.policy);
         const members = [];
```
msorens:
```suggestion
        const members = <Member[]>[];
```

components/automate-ui/src/app/pages/policy/details/policy-details.component.ts
```diff
@@ -63,7 +68,14 @@ export class PolicyDetailsComponent implements OnInit, OnDestroy {
...
         this.policy.members.forEach(element => {
-          members.push(stringToMember(element));
+          const member = stringToMember(element);
+          members.push(member);
+          if (member.type === Type.LocalUser) {
+            this.memberURLs[member.name] = ['/settings', 'users', member.displayName];
```
msorens: Should member.name be changed to member.id ?

components/automate-ui/src/app/pages/policy/details/policy-details.component.ts
```diff
@@ -63,7 +68,14 @@ export class PolicyDetailsComponent implements OnInit, OnDestroy {
...
+          }
+          if (member.type === Type.LocalTeam) {
```
msorens: s/if/else if/

components/automate-ui/src/app/pages/policy/details/policy-details.component.html
```diff
@@ -40,23 +40,25 @@
...
             </chef-tr>
           </chef-thead>
           <chef-tbody>
             <chef-tr *ngFor="let member of members$ | async">
-              <!-- TODO (tc) we should link to local users and teams once we switch those APIs to named IDs.
-              can't now since we don't have the GUIDs for them unless we queried them by name for each
-              row somehow. -->
               <chef-td>
-                {{ member.displayName }}
+                <a *ngIf="memberURLs.hasOwnProperty(member.name)" [routerLink]="memberURLs[member.name]" target="_blank">
```
bcmdarroch: i noticed when i opened the first new tab the welcome modal popped up again in the new tab. didn't happen on the second time though.

if we want the default behavior to be `open a new tab` on team/policy member links, let's do a quick inventory and make that consistent for all our pages. i noticed clicking a user from the team membership table doesn't open a new tab.

components/automate-ui/src/app/pages/policy/details/policy-details.component.ts
```diff
@@ -27,6 +29,9 @@ export class PolicyDetailsComponent implements OnInit, OnDestroy {
   public members$: Observable<Member[]>;
   public tabValue = 'definition';
   public url: string;
+  // Map of local user and team member IDs to URLs.
+  // Will not contain LDAP, SAML, or * members.
+  private memberURLs: { [id: string]: string[] } = {};
```
bcmdarroch: can we include `token` members too?

components/automate-ui/src/app/pages/policy/details/policy-details.component.ts
tylercloke: Not for now, because of how the Member type is defined. I think we should refactor Member's `name` field to be `id` everywhere, but let's do that in a refactor PR and not mix that in with the bug fix imo :)

components/automate-ui/src/app/pages/policy/details/policy-details.component.html
tylercloke: Should definitely make that consistent. Let's take an inventory. I'll add a tech debt card to track down the places the behavior varies and we'll make it consistent!

susanev: if they are in a user flow already then it should open in a new tab, otherwise, it should not. when we update the add users flow for teams then those links will be fixed to match the flow when adding members to a policy.

i dont think anything else is a user flow like that so no need to open in a new tab... unless im overlooking something.

tylercloke: I'll make it not open a new tab in policy-details but open a new tab in the add members modal in that case.

PR #402, opened 2019-05-22
automate-gateway: cleanup GRPC/HTTP multiplexing cruft

srenatus:

This was carried over from some ancient grpc-gateway example code[1]; and
we're not exposing the GRPC API that way right now.

Using (*grpc.Server).ServeHTTP has some issues:

1. it's known to be slow, very slow[2]
2. it's seems to be not working with go-grpc >= 1.19 [3]

Furthermore, we're exposing the proper GRPC server (the native go-grpc
one) on a different port. If decide to (finally!) expose our GRPC API,
that's the thing we should expose.

So, since we don't need this multiplexing business, we shouldn't keep
the code around. This is the cleanup.

[1]: https://github.com/philips/grpc-gateway-example/
[2]: https://github.com/grpc/grpc-go/issues/586
[3]: https://github.com/soheilhy/cmux/issues/64#issuecomment-491999193

### DIFF
--- a/components/automate-gateway/gateway/server.go
+++ b/components/automate-gateway/gateway/server.go
@@ -7,7 +7,6 @@ import (
 	"net"
 	"net/http"
 	"net/url"
-	"strings"
 
 	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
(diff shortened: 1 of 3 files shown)


srenatus:
✔️ trivial PR. merging.

Issue #397, opened 2019-05-21
Remove _Node State_ link from _Client Runs_


susanev:
## Overview
During usability studies at ChefConf 2019 participants repeatedly clicked on the _Node State_ link on the _Client Runs_ page expecting something to happen. This task removes that link if we add more items to the left-nav of that page in the future we can add it back.

## What it looks like when its done
* The _Node State_ links is removed

### Before
<img width="1054" alt="client-runs-node-state" src="https://user-images.githubusercontent.com/5489125/58109292-396eec80-7ba2-11e9-9590-fc680f4ac6d5.png">

### After
<img width="1040" alt="client-runs-no-node-state" src="https://user-images.githubusercontent.com/5489125/58109460-805ce200-7ba2-11e9-9aee-d3a403a64ab0.png">

PR #404, opened 2019-05-22
remove node state link from client run page

vinay033:

During usability studies at ChefConf 2019 participants repeatedly clicked on the Node State link on the Client Runs page expecting something to happen. This task removes that link if we add more items to the left-nav of that page in the future we can add it back.

After talking with @jonong1972 this task is only removing the link from the _Client Runs_ landing page, it will still be visible on node detail pages, we can revisit that in the future if necessary.

### :+1: Definition of Done
client_runs_page

https://github.com/chef/automate/issues/397

### DIFF
--- a/components/automate-ui/src/app/pages/client-runs/client-runs.component.html
+++ b/components/automate-ui/src/app/pages/client-runs/client-runs.component.html
@@ -4,7 +4,6 @@
 </div>
 
 <app-server-org-filter-sidebar>
-  <!-- <chef-sidebar-entry route="/client-runs" icon="storage">Node State</chef-sidebar-entry> -->
   <a *ngIf="isWorkflowEnabled$ | async" href="/workflow">
     <chef-icon class="sidebar-icon" aria-hidden="true">local_shipping</chef-icon>
     Workflow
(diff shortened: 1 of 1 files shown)


components/automate-ui/src/app/pages/client-runs/client-runs.component.html
```diff
@@ -4,7 +4,7 @@
 </div>
 
 <app-server-org-filter-sidebar>
-  <chef-sidebar-entry route="/client-runs" icon="storage">Node State</chef-sidebar-entry>
+  <!-- <chef-sidebar-entry route="/client-runs" icon="storage">Node State</chef-sidebar-entry> -->
```
srenatus: [nit] I'd rather just delete the line. git doesn't forget 😉

susanev: i think you might be able to pull out the whole nav too, and not just that one entry.

vinay033: @srenatus I have removed the commented line. :)

srenatus: I'd merge, but I'm not sure I fully follow you there, @susanev: which _whole nav_ would you like to see pulled?

susanev: yah, sorry i was not being clear. right now it renders a totally empty nav element when there are no items in the left-nav. i realize now, that workflow complicates things. the empty nav element can be annoying for people using screen readers. but its not important enough to fix in this pr i dont think. so ignore my comment, heh.

PR #406, opened 2019-05-22
Add proper IAMv2p1 machines to deployment farm

msorens:

Previously I had repurposed the IAM v2 "fresh install" machine in the deployment farm to 2.1 (leaving the "inplace upgrade" at v2). But that "conservation of resources" led to more confusion than savings, so this PR reverts the change on the IAM v2 "fresh install" machine _and_ creates first-class siblings for v2.1, both fresh install and inplace upgrade.

### :+1: Definition of Done

New v2p1 machines are provisioned on each of the dev and acceptance dashboards.

- [ ] Code actually executed?

### DIFF
--- a/terraform/test-environments/scenarios/single_local_iamv2p1_inplace_upgrade.tf
+++ b/terraform/test-environments/scenarios/single_local_iamv2p1_inplace_upgrade.tf
@@ -18,7 +18,7 @@ module "single_local_iamv2p1_inplace_upgrade" {
   vpc            = "${var.environment}"
   platform       = "ubuntu-16.04"
   key_name       = "cd-infrastructure"
-  instance_type  = "m4.large"
+  instance_type  = "m5.large"
   root_volume_gb = "200"
 
   # Required AWS Tags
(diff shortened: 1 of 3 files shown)


terraform/test-environments/scenarios/single_local_iamv2p1_inplace_upgrade.tf
```diff
@@ -0,0 +1,57 @@
...
+  # Metadata
+  meta_title       = "Single Local (Inplace Upgrade) with IAMv2p1"
+  meta_description = "A2 stack with IAMv2p1 (using SAML) deployed locally as Habitat packages on a single host using the chef-automate CLI."
+  meta_type        = "habitat"
+
+  # AWS Instance Configuration
+  vpc            = "${var.environment}"
+  platform       = "ubuntu-16.04"
+  key_name       = "cd-infrastructure"
+  instance_type  = "m4.large"
```
tduffield: Is there a reason this is a m4.large while the other is a m5.large?

msorens: Yes: because somebody started it that way. 😁

If you look at all the other "fresh-install / implace-upgrade" pairs, the fresh install uses **m5** and the inplace-upgrade uses **m4**. 

(I personally have no context what those values actually do or mean, though.)

tduffield: We should move to m5 large. We don't need to backfill now since that would recreate the instance, but going forward we should use m5 class instances.

Issue #398, opened 2019-05-21
filter by status on client runs searchbar


vjeffrey:
# User Story

We already support filtering by status on the client runs nodes list by doing some button clicks above the table. 
We've had a few customers mention they would like to have this search functionality in the searchbar as well, so that the values show up in the search url they send to people.
This should be a pretty easy change to make.

# Definition of Done
- client runs searchbar displays an option to filter by `status`, with possible options `failed`, `successful`, and `missing`
- the status search selection shows up in the url

lancewf:
@vjeffrey The status selection already shows up in the URL. So, the user can share this selection with other users. 

The only difficulty here is allowing the user to select multiple statuses from the search bar. Currently, we only allow one status to be selected in the UI and the backend. 

I will look into adding this tomorrow since I have been updating the search bar in the last two PRs.

lancewf:
This task is on hold because of some questions of the UI changes.

vjeffrey:
i'm going to remove this card from the board until we figure this out

PR #407, opened 2019-05-22
WIP: Client Runs Filter Status

lancewf:
Screen Shot 2019-05-22 at 4 05 08 PM

Adding status filters to Client Runs the search bar. Now multiple statuses can be selected at the same time. Multiple statuses can only be selected by using the search bar. Only one of the status can be selected using the status buttons above the node list.  

1. `build components/ingest-service && build components/config-mgmt-service && build components/automate-ui`
1. `start_all_services`
1. `chef_load_nodes 10`
1. Open https://a2-dev.test/client-runs?status=success&status=failure

https://github.com/chef/automate/issues/398

- [x] Code actually executed?


components/automate-ui/src/app/pages/client-runs/client-runs.component.ts
```diff
@@ -245,29 +249,40 @@ export class ClientRunsComponent implements OnInit, OnDestroy {
...
-    this.isFailureStatusSelected$ = this.selectedStatus$.pipe(map(status => status === 'failure'));
-    this.isTotalStatusSelected$ = this.selectedStatus$.pipe(map(status => status === undefined));
+    this.isMissingStatusSelected$ = this.selectedStatuses$.pipe(map(statuses =>
+      statuses.findIndex(status => status === 'missing') >= 0));
+    this.isSuccessStatusSelected$ = this.selectedStatuses$.pipe(map(statuses =>
+      statuses.findIndex(status => status === 'success') >= 0));
+    this.isFailureStatusSelected$ = this.selectedStatuses$.pipe(map(statuses =>
+      statuses.findIndex(status => status === 'failure') >= 0));
+    this.isTotalStatusSelected$ = this.selectedStatuses$.pipe(map(statuses =>
+      statuses.length === 0));
```
lancewf: These above changes are to account that multiple statuses can be selected. Unlike before only one status could have been selected.

components/ingest-service/backend/elastic/mappings/node_state.go
```diff
@@ -65,7 +65,13 @@ var nodeProps = `
...
 	},
 	"status": {
-		"type": "keyword"
+		"type": "keyword",
+		"fields": {
+			"engram" : {
+				"type": "text",
+				"analyzer": "autocomplete"
+			}
+		}
```
lancewf: Allowing the status to have autocomplete. Without this, the suggestions do not work.

components/automate-ui/src/app/pages/client-runs/client-runs.component.spec.ts
```diff
@@ -554,10 +546,9 @@ describe('ClientRunsComponent', () => {
         new UpdateNodeFilters({filters: {
           page: 1,
           pageSize: 100,
-          searchBar: [],
+          searchBar: [{type: 'status', text: 'wrong'}],
```
lancewf: I think this should be allowed. This would support future new types of node statuses.

vjeffrey:
thank you for doing this work! 
i haven't reviewed yet, will shortly, but it sounds like we should chat with ux first to make sure this is in line with what they're thinking.
for background, this came from a customer request, and we will also need status filtering in compliance reporting.
let's chat with ux sometime over the next couple days, i'll see if i can get a mtg on the calendar

lancewf:
I jumped the gun on this PR. The UI has not been decided yet.

PR #410, opened 2019-05-23
[A2-802] CreateRule projects API

srenatus:

Adds the CreateRule GRPC method to `authz-service`, wrapping its storage's `CreateRule`.

#### Items on Note

- This includes some property-based tests in [`rules_property_test.go`](https://github.com/chef/automate/pull/410/files#diff-66769720a1732f29e7206a24759eea52), where we encode what we want the system to do, and what "random inputs" look like. Then, the gopter library takes care of generating 100 inputs, and asserting the property. If it fails to hold, the library will attempt to _shrink_ the input into a manageable test case (as anything random can be unwieldy).

### :+1: Definition of Done

- [X] you can create a rule from `chef-automate dev grpcurl authz-service`
- [x] the `authz-service` method is exposed via `automate-gateway`
- [ ] expose operator (it's currently only "member-of" but it should be in the API) -- _❓ need input there, I'm missing context on the decision_

- `start_all_services`
- `rebuild components/authz-service`, `rebuild components/automate-gateway`
- create a rule: (`hab pkg install -b core/jq-static` for `jq`)
```
[121][default:/src:130]# curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/rules -d "$(jq -n '{ id: "foo-rule", name: "my foo rule", type: "NODE", project_id: "foo-project", conditions: [{ type: "CHEF_SERVERS", values: ["prod", "staging"]}]}')"
{"error":"error creating rule with ID \"foo-rule\": foreign key violation","message":"error creating rule with ID \"foo-rule\": foreign key violation","code":13,"details":[]}[122][default:/src:0]#
```
- note that this naive attempt failed -- the project doesn't exist -- and the error messaging **could be better** ⚠️ 
- try again, creating the project first:
```
[122][default:/src:0]# curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects -d "$(jq -n '{ id: "foo-project", name: "my foo project" }')"
{"project":{"name":"my foo project","id":"foo-project","type":"CUSTOM","projects":["foo-project"]}}
[123][default:/src:0]# curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/rules -d "$(jq -n '{ id: "foo-rule", name: "my foo rule", type: "NODE", project_id: "foo-project", conditions: [{ type: "CHEF_SERVERS", values: ["prod", "staging"]}]}')" | jq .
{
  "rule": {
    "id": "foo3-rule",
    "project_id": "foo-project",
    "name": "my foo rule",
    "type": "NODE",
    "conditions": [
      {
        "type": "CHEF_SERVERS",
        "values": [
          "prod",
          "staging"
        ]
      }
    ]
  }
}
```
- check the database that it was properly created:
```
[129][default:/src:113]# chef-automate dev psql chef_authz_service
psql (9.6.11)
SSL connection (protocol: TLSv1.2, cipher: ECDHE-RSA-AES256-GCM-SHA384, bits: 256, compression: off)
Type "help" for help.

chef_authz_service=# select * from iam_project_rules where id='foo-rule';
 db_id |    id    | project_id  |    name     | type
-------+----------+-------------+-------------+------
     3 | foo-rule | foo-project | my foo rule | node
(1 row)
chef_authz_service=# select * from iam_rule_conditions where rule_db_id=3;
 db_id | rule_db_id |     value      |  attribute  | operator
-------+------------+----------------+-------------+-----------
     1 |          3 | {prod,staging} | chef-server | member-of
(1 row)
```

- [A2-650](https://chefio.atlassian.net/browse/A2-650)
  - [A2-802](https://chefio.atlassian.net/browse/A2-802) 👈 

- [X] Code actually executed?


components/authz-service/server/v2/project.go
```diff
@@ -398,17 +397,49 @@ func fromStorageConditions(cs []storage.Condition) ([]*api.Condition, error) {
...
+	storage.ChefServer:   api.ProjectRuleConditionTypes_CHEF_SERVERS,
+	storage.ChefTag:      api.ProjectRuleConditionTypes_CHEF_TAGS,
+	storage.Environment:  api.ProjectRuleConditionTypes_CHEF_ENVIRONMENTS,
+	storage.Organization: api.ProjectRuleConditionTypes_CHEF_ORGS,
+	storage.PolicyGroup:  api.ProjectRuleConditionTypes_POLICY_GROUP,
+	storage.PolicyName:   api.ProjectRuleConditionTypes_POLICY_NAME,
+}
+
+var apiToStorageConditionAttributes map[api.ProjectRuleConditionTypes]storage.ConditionAttribute
+var onceReverseConditionAttributesMapping sync.Once
```
bcmdarroch: what does `sync.Once` do?

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -0,0 +1,60 @@
...
+		"Name":      gen.AnyString(),
+		"ProjectId": gen.Identifier(),
+		"Type": gen.OneConstOf(
+			api.ProjectRuleTypes_NODE,
+			api.ProjectRuleTypes_EVENT,
+		),
+		"Conditions": gen.SliceOf(conditionsGen),
+	})
+
+	params := gopter.DefaultTestParametersWithSeed(seed)
```
bcmdarroch: what does `gopter` do for us?

components/automate-gateway/api/iam/v2beta/common/rules.proto
```diff
@@ -0,0 +1,34 @@
+syntax = "proto3";
+
+package chef.automate.api.iam.v2beta;
+option go_package = "github.com/chef/automate/components/automate-gateway/api/iam/v2beta/common";
+
+enum RuleType {
```
bcmdarroch: thanks for splitting this out! we probably should do the same for roles one of these days

components/authz-service/server/v2/rules_property_test.go
tylercloke: The [readme](https://github.com/leanovate/gopter#synopsis) isn't the best, but it's a property based tester. What it looks like we are using it for here is having it generate lots of different combinations of all the conditions possibilities (since we have a lot of different condition types) so we can test `CreateRule` against many possible different condition inputs without manually managing how they are generated.

components/authz-service/server/v2/project_test.go
```diff
@@ -371,9 +371,14 @@ func addProjectToStore(t *testing.T, store *cache.Cache, id, name string, projTy
 }
 
 func setupProjects(t *testing.T) (api.ProjectsClient, *cache.Cache, *mockEventServiceClient) {
+	cl, ca, _, mc, _ := setupProjectsAndRules(t)
+	return cl, ca, mc
+}
```
msorens: Add blank line after this

components/authz-service/server/v2/project_test.go
```diff
@@ -371,9 +371,14 @@ func addProjectToStore(t *testing.T, store *cache.Cache, id, name string, projTy
...
+func setupProjectsAndRules(t *testing.T) (api.ProjectsClient, *cache.Cache, *cache.Cache, *mockEventServiceClient,
```
msorens: Odd place to break the line; consider:
```
func setupProjectsAndRules(t *testing.T) (
api.ProjectsClient, *cache.Cache, *cache.Cache, *mockEventServiceClient, int64) {
```

components/authz-service/server/v2/project_test.go
```diff
@@ -371,9 +371,14 @@ func addProjectToStore(t *testing.T, store *cache.Cache, id, name string, projTy
...
+	int64) {
 	t.Helper()
 	ctx := context.Background()
-	prng.Seed(t)
+	seed := prng.GenSeed(t)
```
msorens: Why the seeding change?

components/authz-service/server/v2/project_test.go
```diff
@@ -371,9 +371,14 @@ func addProjectToStore(t *testing.T, store *cache.Cache, id, name string, projTy
 }
 
 func setupProjects(t *testing.T) (api.ProjectsClient, *cache.Cache, *mockEventServiceClient) {
+	cl, ca, _, mc, _ := setupProjectsAndRules(t)
```
msorens: Why does `setupProjectsAndRules` return arguments that are never used?

components/authz-service/server/v2/rules_test.go
```diff
@@ -0,0 +1,163 @@
...
+					ProjectId: "foo",
+					Type:      api.ProjectRuleTypes_NODE,
+					Conditions: []*api.Condition{
+						{
+							Type:   api.ProjectRuleConditionTypes_CHEF_ORGS,
+							Values: []string{"chef"},
+						},
+					},
+				},
+			}, resp) // FIXME
```
msorens: What needs fixing?

components/authz-service/server/v2/rules_test.go
```diff
@@ -0,0 +1,163 @@
...
+				Name:       "my other foo",
+				ProjectId:  "bar",
+				Conditions: apiConditions,
+			})
+			grpctest.AssertCode(t, codes.AlreadyExists, err)
+			assert.Nil(t, resp)
+		}},
+		// happy path
+		{"with valid rule data, returns no error and creates the rule in storage", func(t *testing.T) {
+			resp, err := cl.CreateRule(ctx, &api.CreateRuleReq{
```
msorens: Was curious whether this reduced form would work, but it fails. Any thoughts on why?
```
			ruleReq := &api.CreateRuleReq{
				Id:        "any-name",
				Name:      "any name",
				ProjectId: "foo",
				Type:      api.ProjectRuleTypes_NODE,
				Conditions: []*api.Condition{
					{
						Type:   api.ProjectRuleConditionTypes_CHEF_ORGS,
						Values: []string{"chef"},
					},
				},
			}
			resp, err := cl.CreateRule(ctx, ruleReq)
			assert.NoError(t, err)
			rule := api.ProjectRule(*ruleReq)
			assert.Equal(t, &api.CreateRuleResp{
				Rule: &rule,
			}, resp)
```

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -0,0 +1,124 @@
+package v2_test
+
+import (
+	"context"
+	"reflect"
+	"testing"
+
+	"github.com/leanovate/gopter"
```
msorens: We have used one data generator in our code base already, https://github.com/jaswdr/faker. Is gopter more useful?

components/authz-service/server/v2/project_test.go
srenatus: because it's controlling randomness in gopter

components/authz-service/server/v2/rules_property_test.go
srenatus: > Is gopter more useful?

100 times _yes_ -- since it's not _just_ a data generator. Back when we've introduced faker, you might remember our conversations (also with @phiggins, https://github.com/chef/a2/pull/4294 and previous gopter usage introduced earlier: https://github.com/chef/a2/pull/4369): fake data is nice for tests, but the icing on the cake is _property based tests_. This is what gopter gives us: a way to state **properties**, and assert them using a large battery of random inputs.

I.e. instead of
> generate one random policy, see that it can be created

we formulate a property
> when CreateRule is called with some input, it will lead to a rule in storage

and have the framework try that a hundred times with different inputs.

**Also**, any decent PBT framework gives you _shrinking_: when it finds an input such that the property _not holds_, it attempts to find the smallest input that does the same thing. It's vital for dealing with random inputs. 

Now, this doesn't mean what we don't need unit tests. And it also doesn't mean that the way we're using it here is the one and only correct way.

components/authz-service/server/v2/project.go
srenatus: Exactly! It's the lazy way for not having to maintain both mappings manually. Only A -> B is hardcoded, B -> A is initialized _once_ in the way @tylercloke described.

components/authz-service/server/v2/project_test.go
srenatus: Not used for `setupProjects`. They _are_ used for `setupRules`. 🤷‍♂ We could clean this up, I suppose. But I don't find it that terrible.

components/authz-service/server/v2/rules_test.go
srenatus: Didn't dive in deeper, but with this:
```diff
diff --git a/components/authz-service/server/v2/rules_test.go b/components/authz-service/server/v2/rules_test.go
index 08a82b6c..c1be18cf 100644
--- a/components/authz-service/server/v2/rules_test.go
+++ b/components/authz-service/server/v2/rules_test.go
@@ -5,6 +5,7 @@ import (
 	"math/rand"
 	"testing"
 
+	"github.com/kylelemons/godebug/pretty"
 	cache "github.com/patrickmn/go-cache"
 	"github.com/stretchr/testify/assert"
 	"google.golang.org/grpc/codes"
@@ -103,7 +104,8 @@ func TestCreateRule(t *testing.T) {
 		}},
 		// happy path
 		{"with valid rule data, returns no error and creates the rule in storage", func(t *testing.T) {
-			resp, err := cl.CreateRule(ctx, &api.CreateRuleReq{
+
+			ruleReq := &api.CreateRuleReq{
 				Id:        "any-name",
 				Name:      "any name",
 				ProjectId: "foo",
@@ -114,22 +116,16 @@ func TestCreateRule(t *testing.T) {
 						Values: []string{"chef"},
 					},
 				},
-			})
+			}
+			actual, err := cl.CreateRule(ctx, ruleReq)
 			assert.NoError(t, err)
-			assert.Equal(t, &api.CreateRuleResp{
-				Rule: &api.ProjectRule{
-					Id:        "any-name",
-					Name:      "any name",
-					ProjectId: "foo",
-					Type:      api.ProjectRuleTypes_NODE,
-					Conditions: []*api.Condition{
-						{
-							Type:   api.ProjectRuleConditionTypes_CHEF_ORGS,
-							Values: []string{"chef"},
-						},
-					},
-				},
-			}, resp)
+			rule := api.ProjectRule(*ruleReq)
+			expected := &api.CreateRuleResp{
+				Rule: &rule,
+			}
+			if !assert.Equal(t, expected, actual) {
+				t.Log(pretty.Compare(expected, actual))
+			}
 		}},
 	}
```
...we get:
```
    --- FAIL: TestCreateRule/with_valid_rule_data,_returns_no_error_and_creates_the_rule_in_storage (0.00s)
        Error Trace:    rules_test.go:126
        Error:          Not equal:
                        expected: &v2.CreateRuleResp{Rule:(*v2.ProjectRule)(0xc000192690), XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0}
                        actual  : &v2.CreateRuleResp{Rule:(*v2.ProjectRule)(0xc000192230), XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0}

                        Diff:
        Test:           TestCreateRule/with_valid_rule_data,_returns_no_error_and_creates_the_rule_in_storage
        rules_test.go:127:  {
              Rule: {
               Id: "any-name",
               ProjectId: "foo",
               Name: "any name",
               Type: 0,
               Conditions: [
                {
                 Type: 1,
                 Values: [
                  "chef",
                 ],
                 XXX_NoUnkeyedLiteral: {
                 },
                 XXX_unrecognized: [
                 ],
            -    XXX_sizecache: 8,
            +    XXX_sizecache: 0,
                },
               ],
               XXX_NoUnkeyedLiteral: {
               },
               XXX_unrecognized: [
               ],
            -  XXX_sizecache: 35,
            +  XXX_sizecache: 0,
              },
              XXX_NoUnkeyedLiteral: {
              },
              XXX_unrecognized: [
              ],
              XXX_sizecache: 0,
             }
```
So, the trivial answer is: the `XXX_sizecache` fields don't match. 😉

components/authz-service/server/v2/project_test.go
msorens: Ah, still used to thinking about private to namespace means private to file, but not true in Go. 👍

components/authz-service/server/v2/rules_test.go
msorens: Hmm... I did not see the expanded diff when I ran it, just this. Any idea why my output is different?
```
expected: &v2.CreateRuleResp{Rule:(*v2.ProjectRule)(0xc0001930a0), XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0}
            
actual  : &v2.CreateRuleResp{Rule:(*v2.ProjectRule)(0xc000192bd0), XXX_NoUnkeyedLiteral:struct {}{}, XXX_unrecognized:[]uint8(nil), XXX_sizecache:0}
                        
Diff:
```

srenatus: For more information please reread -- my comment includes the diff of what I had to do to see that, too ☝️

msorens: 🤔 Except it does not include the actual command; "diff --git..." is output... Am I missing something...?

srenatus: Sorry that was the output of `git diff`.

PR #411, opened 2019-05-23
[automate-cli] Generate unique serial for self-signed certs

stevendanna:
Automate CLI generates a self-signed SSL certificate for the front-end
load balancer if the user does not provide one.

Previously, this certificate had a hard-coded Serial of 1. This PR
changes it to a random serial to avoid errors in Firefox and
potentially other browsers.

### DIFF
--- a/api/config/deployment/init_config.go
+++ b/api/config/deployment/init_config.go
@@ -356,9 +356,31 @@ func generatePrivateKey() (*rsa.PrivateKey, error) {
 	return rsa.GenerateKey(rand.Reader, keyLength)
 }
 
+func generateSerial() (*big.Int, error) {
+	// According to
+	// https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.6.4.pdf:
+	//
+	// Effective September 30, 2016, CAs SHALL generate
+	// non-sequential Certificate serial numbers greater than zero
+	// (0) containing at least 64 bits of output from a CSPRNG.
+	//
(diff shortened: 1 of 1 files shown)


api/config/deployment/init_config.go
```diff
@@ -356,9 +356,31 @@ func generatePrivateKey() (*rsa.PrivateKey, error) {
...
+	// According to
+	// https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.6.4.pdf:
+	//
+	// Effective September 30, 2016, CAs SHALL generate
+	// non-sequential Certificate serial numbers greater than zero
+	// (0) containing at least 64 bits of output from a CSPRNG.
+	//
+	// Here, we set the limit to double this requirement.
+	limit := new(big.Int).Lsh(big.NewInt(1), 128)
+	ret, err := rand.Int(rand.Reader, limit)
```
msorens: Just wondering how this is seeded... would it generate the same sequence the next time the service is started? And is that necessarily a bad thing?

stevendanna: @msorens Great question.  The `crypto/rand` package uses an operating system entropy source.

On the systems we support, this means we'll likely be calling this system call:

http://man7.org/linux/man-pages/man2/getrandom.2.html

This means that we generally should be getting high-quality random data that will not be the same between runs.  On systems that do not have that system call, it may fall back to another source that could produce predictable results if called during very early boot on some systems, but I don't think that is likely for our use case.

Issue #336, opened 2019-05-14
Compliance remote jobs to pin version of inspec to install


alexpop:
# User Story

As an Automate User I want a compatible version of InSpec to be installed on the remote systems when running remote jobs, like SSM ones.

Right now, latest version is installed. This might cause an older version of Automate that does not accept the InSpec license automatically to install InSpec 4 which requires license acceptance. I suspect this to be the issue here: https://github.com/chef/automate/issues/296

With this change, InSpec should not be installed on the remote system if the version is already installed from a previous run.

PR #414, opened 2019-05-23
Pin inspec version that is run remotely

alexpop:
Fixes https://github.com/chef/automate/issues/336 by pinning the version of inspec to the same version installed with Automate. The same version of inspec is only installed once on a remote target due to the 'once' install strategy.


components/compliance-service/habitat/plan.sh
```diff
@@ -38,7 +38,9 @@ pkg_deps=(
   core/jq-static            # Used in habitat/hooks/health_check
   ${local_platform_tools_origin:-chef}/automate-platform-tools
   # WARNING: Update with care. The chef/inspec is managed with Expeditor.
+
   # See .expeditor/update-inspec-version.sh for details
+  # When updating this, also update 'remote_inspec_version' in ./default.toml
```
stevendanna: Can we modify the script to update the version pin in the new locations as well so that our existing automation still does the right thing?

components/compliance-service/habitat/default.toml
```diff
@@ -22,3 +22,4 @@ manual_polling_interval = 1440
 [agent]
 workers = 10
 buffer_size = 1000
+remote_inspec_version = "3.9.0"
```
stevendanna: Typically we also specify the default in the Go code here: https://github.com/chef/automate/blob/master/api/config/compliance/config_request.go#L25

It's annoying, likely isn't strictly required anymore, but might still be good to do for consistency.

alexpop: Thank you, auto-updating the value with the most recent commit.

components/compliance-service/habitat/plan.sh
alexpop: Yes, please check the most recent commit ( https://github.com/chef/automate/pull/414/commits/e17e1e204e806bf68d0b26ad55ee312818a631f3 ). Is that in line with your thinking?

alexpop: Good catch Steven! Take a look at this ( https://github.com/chef/automate/pull/414/commits/ab45ccb6237861ede6ae040d5c4a669ead00ed46 ) commit I pushed in order for expeditor to continue to match the inspec package.

PR #416, opened 2019-05-24
Cleanup FilterAuthorizedProjects

msorens:

When originally implemented, `FilterAuthorizedProjects` was modeled after `FilterAuthorizedPairs` for expediency... but it really was not very clean, artificially introducing exposing it to IAM v1 when it had no business being there. This PR strips out all the artificial v1 artifacts.

### :+1: Definition of Done

Global project filter dropdown still gets populated as before.

```
rebuild components/authz-service/
rebuild components/automate-gateway/
```
Add a couple projects in the UI or command-line. Wait 2 minutes (or refresh browser page) to see the global projects filter dropdown populated.

PR - #5396 (https://github.com/chef/a2/pull/5396)

- [x] Code actually executed?


components/automate-gateway/handler/authz.go
```diff
@@ -345,17 +345,23 @@ func (a *AuthzServer) getAllowedProjects(
...
+		Pairs:    pairsV2,
+	})
 	if err != nil {
 		log.WithError(err).Debug("Error on client.FilterAuthorizedProjects")
 		return nil, err
 	}
-	return resp.Projects, nil
+	respProjects := make([]string, len(resp.Projects))
+	copy(respProjects, resp.Projects)
+	return respProjects, nil
```
srenatus: Why do we do this `copy` business? In any real-world usage, I don't think this matters, does it?

components/automate-gateway/handler/authz.go
```diff
@@ -345,17 +345,23 @@ func (a *AuthzServer) getAllowedProjects(
 	// Fetches the id of the current user PLUS the team ids for that user
```
srenatus: 🤔 Shouldn't the entire IAM v2[.1] business live in [automate-gateway/handler/iam/v2beta](https://github.com/chef/automate/tree/master/components/automate-gateway/handler/iam/v2beta)? What do you think?

components/automate-gateway/handler/authz.go
msorens: You are probably right; don't recall why that was here...

components/automate-gateway/handler/authz.go
msorens: I agree; I may save that for a separate PR, though.

components/automate-gateway/handler/authz.go
stevendanna: I think this might have come from being consistent with some other code that was the result an overly-verbose comment I left on another PR:

https://github.com/chef/a2/pull/5396#discussion_r269438700

As I mentioned there though, I think we can return resp.Projects directly in this case.

components/automate-gateway/handler/authz.go
srenatus: 😄 Would fit quite nicely under this PR's description though 😉 But yeah, it would involve changing the UI as well...

components/automate-gateway/handler/authz_test.go
```diff
@@ -181,11 +181,11 @@ func TestIntrospectAllProjectsV2(t *testing.T) {
 	}{
 		"empty response": {
 			&authz_v2.FilterAuthorizedProjectsResp{},
-			[]string{},
+			[]string(nil),
```
srenatus: ℹ️ By writing `[]string(nil)`, you're "typing `nil`" -- and we don't need that, its type is clear from [the struct definition](https://github.com/chef/automate/pull/416/files/4d8ce51c08272cc3d1b0202a81ccf420cf4773c2..7c509bccf2eddae87ba6ef588037a58c42b47b92#diff-e23e2853bf09746e5ad1182dbabfa42bR180).

[See this playground for some illustration.](https://play.golang.org/p/0ZGC0DiTSN5)

msorens: This change was introduced to make the test pass. Are you suggesting there is a better way?

srenatus: Yeah just try `nil` there now

msorens: Changed it there, and in a few other places as well.

PR #418, opened 2019-05-24
HOLD: Generate self-signed CA and separate certificate

stevendanna:
This is a potential fix for reports of Chrome reporting
ERR_CERT_INVALID. Our suspicion is that Chrome may be rejecting the
use of a certificate with both `CA: TRUE` and the serverAuth extended
key usage.

Here we generated a self-signed CA and then generate a cert signed by
that CA.  Both certificates are concatenated for use in nginx.

### DIFF
--- a/api/config/deployment/init_config.go
+++ b/api/config/deployment/init_config.go
@@ -4,8 +4,10 @@ import (
 	"bytes"
 	"crypto/rand"
 	"crypto/rsa"
+	"crypto/sha1"
 	"crypto/x509"
 	"crypto/x509/pkix"
+	"encoding/asn1"
 	"encoding/hex"
 	"encoding/pem"
 	"fmt"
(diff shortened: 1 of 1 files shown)


stevendanna:
Unfortunately, this is even worse on Windows.  On Windows 10 and Chrome, it appears there is no easy path to allow the chained certificate like you would allow the current self-signed certificate.

I'm thinking we ship the already-merged Serial fix and see if that helps the users who reported this.

There are a couple of things identified here that we might want to look into:

- [ ] Having `CA: true` and the `serverAuth` extended key usage on the same cert seems to cause a violation according to x509lint.  Perhaps we can just remove one or both of these for our self-signed cert?

- [ ] Having `CA: true` without a Subject Key ID is also flagged by x509lint.  Perhaps we can just add that to our self-signed cert?

Now that I can test a bit of this on at least 2 OSes I'll play with this sometime this week.

stevendanna:
I'll close this if I don't make progress on it this week.

PR #425, opened 2019-05-28
Rename auth get endpoints to list

msorens:

Per my spreadsheet of auth endpoints (https://docs.google.com/spreadsheets/d/190c3VvCQGpsJ3eWq8DvqgTsIdZw3Mmk1os_9itjADgY/edit#gid=0) an anomaly stood out for GetTeams and GetUsers, which should be more appropriately ListTeams and ListUsers in v2.

This is done with this PR; also changed the interservice API for teams, as it seemed useful to do that, but we seem to only have a single interservice proto file for users. So not sure if it was worth adding a new v2 proto file just for that. Opinions?

### :+1: Definition of Done

Exposed endpoints are now ListTeams and ListUsers.

- [x] Code actually executed?


components/automate-gateway/handler/iam/v2beta/users/users.go
```diff
@@ -43,8 +43,8 @@ func (p *Server) CreateUser(
 	return &pb_resp.CreateUserResp{User: convert(resp)}, nil
 }
 
-func (p *Server) GetUsers(
-	ctx context.Context, _ *pb_req.GetUsersReq) (*pb_resp.GetUsersResp, error) {
+func (p *Server) ListUsers(
+	ctx context.Context, _ *pb_req.ListUsersReq) (*pb_resp.ListUsersResp, error) {
 	resp, err := p.users.GetUsers(ctx, &local_user.GetUsersReq{})
```
bcmdarroch: If consistency is the main goal here, I think we should update the method names on the domain level as well (which you already did with teams)

bcmdarroch: i spose you've kept GetUsers here since it's used by the v1 API. that works for me, don't want to blow this up into a huge rename

tylercloke:
@srenatus Yeah, that's using V1 still. Should be safe.

srenatus:
rebased with the intention to merge when green

PR #430, opened 2019-05-28
[A2-806] GetRule + ListRules projects API

bcmdarroch:
We can create rules - now we can get them too! 🌟 

### :+1: Definition of Done
- [x] GetRule API works
- [x] ListRules API works

- `start_all_services`
- `rebuild components/authz-service && rebuild components/automate-gateway`
-  `hab pkg install -b core/jq-static` for `jq`
- create a project then create a few rules associated with that project
```
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects -d "$(jq -n '{ id: "foo-project", name: "my foo project" }')"

curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/rules -d "$(jq -n '{ id: "foo-rule", name: "my foo rule", type: "NODE", project_id: "foo-project", conditions: [{ type: "CHEF_SERVERS", values: ["prod", "staging"]}]}')" | jq .

curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/rules -d "$(jq -n '{ id: "coo-foo-rule", name: "my coo foo rule", type: "EVENT", project_id: "foo-project", conditions: [{ type: "CHEF_ORGS", values: ["dev"]}]}')" | jq .

```
- now get one of those rules!
```
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/rules/foo-rule | jq
{
  "rule": {
    "id": "foo-rule",
    "project_id": "foo-project",
    "name": "my foo rule",
    "type": "NODE",
    "conditions": [
      {
        "type": "CHEF_SERVERS",
        "values": [
          "prod",
          "staging"
        ]
      }
    ]
  }
}

curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/rules/coo-foo-rule | jq
{
  "rule": {
    "id": "coo-foo-rule",
    "project_id": "foo-project",
    "name": "my coo foo rule",
    "type": "EVENT",
    "conditions": [
      {
        "type": "CHEF_ORGS",
        "values": [
          "dev"
        ]
      }
    ]
  }
}

curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/rules | jq .
{
  "rules": [
   ... <those 2 rules>
  ]
}
```

- [ ] Code actually executed?


components/authz-service/server/v2/rules_test.go
```diff
@@ -143,6 +144,171 @@ func TestCreateRule(t *testing.T) {
...
+			resp, err := cl.GetRule(ctx, &api.GetRuleReq{Id: ""})
+			grpctest.AssertCode(t, codes.InvalidArgument, err)
+			assert.Nil(t, resp)
+		}},
+		{"if the rule id is invalid, returns 'invalid argument'", func(t *testing.T) {
+			resp, err := cl.GetRule(ctx, &api.GetRuleReq{Id: "no_underscore_allowed"})
+			grpctest.AssertCode(t, codes.InvalidArgument, err)
+			assert.Nil(t, resp)
+		}},
+		{"if the rule does not exist, returns code 'not found'", func(t *testing.T) {
```
msorens: Delete "code" for consistency with above tests.

components/authz-service/server/v2/rules_test.go
```diff
@@ -143,6 +144,171 @@ func TestCreateRule(t *testing.T) {
...
+	cases := []struct {
+		desc string
+		f    func(*testing.T)
+	}{
+		{"if no rules exist, returns empty list", func(t *testing.T) {
+			resp, err := cl.ListRules(ctx, &api.ListRulesReq{})
+			require.NoError(t, err)
+			assert.Equal(t, &api.ListRulesResp{}, resp)
+		}},
+		{"if many rules exist, returns all rules", func(t *testing.T) {
```
msorens: How about "multiple" instead of "many"? Two is hardly "many". 😁

components/authz-service/server/v2/project.go
```diff
@@ -333,6 +333,43 @@ func (s *state) CreateRule(ctx context.Context, req *api.CreateRuleReq) (*api.Cr
...
+	return &api.GetRuleResp{Rule: apiRule}, nil
+}
+
+func (s *state) ListRules(ctx context.Context, req *api.ListRulesReq) (*api.ListRulesResp, error) {
+	resp, err := s.store.ListRules(ctx)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "error retrieving rules: %s", err.Error())
+	}
+
+	rules := []*api.ProjectRule{}
```
msorens: When you have a known quantity, should use the more efficient pattern:
```
	rules := make([]*api.ProjectRule, len(resp))
	for i, rule := range resp {
	. . .
```

api/interservice/authz/v2/project.proto
```diff
@@ -72,13 +71,15 @@ service Projects {
 
     // "proper" methods go here (and the ones above will be phased out eventually)
     rpc CreateRule(CreateRuleReq) returns (CreateRuleResp) {};
+    rpc GetRule(GetRuleReq) returns (GetRuleResp) {};
+    rpc ListRules(ListRulesReq) returns (ListRulesResp) {};
```
msorens: What about `ListProjectRules` and `GetProjectRules` above? Are these replacing those?

components/authz-service/server/v2/rules_test.go
```diff
@@ -143,6 +144,171 @@ func TestCreateRule(t *testing.T) {
...
+			resp, err := cl.GetRule(ctx, &api.GetRuleReq{Id: "no_underscore_allowed"})
+			grpctest.AssertCode(t, codes.InvalidArgument, err)
+			assert.Nil(t, resp)
+		}},
+		{"if the rule does not exist, returns code 'not found'", func(t *testing.T) {
+			resp, err := cl.GetRule(ctx, &api.GetRuleReq{Id: "foo"})
+			grpctest.AssertCode(t, codes.NotFound, err)
+			assert.Nil(t, resp)
+		}},
+		{"if a rule with the requested id exists, returns the rule", func(t *testing.T) {
```
msorens: From the name, it is not clear how this test differs from the next; suggest rename to "if there is only one rule and it matches..." 
However, you might just want to delete this test; I do not think it adds any value beyond that of the next test.

api/interservice/authz/v2/project.proto
bcmdarroch: No those two endpoints are [slightly different](https://github.com/chef/automate/blob/master/components/authz-service/projects_spike.md#endpoints) and will be updated when we get to [this card](https://chefio.atlassian.net/browse/A2-809).

components/automate-gateway/api/iam/v2beta/rules.proto
```diff
@@ -22,4 +22,20 @@ service Rules {
...
+    option (chef.automate.api.policy).resource = "auth:rules:{id}";
+    option (chef.automate.api.policy).action = "get";
+    option (chef.automate.api.iam.policy).resource = "iam:rules:{id}";
+    option (chef.automate.api.iam.policy).action = "iam:rules:get";
+  };
+
+  rpc ListRules (ListRulesReq) returns (ListRulesResp) {
+    option (google.api.http).get= "/iam/v2beta/rules";
+    option (chef.automate.api.policy).resource = "auth:rules";
+    option (chef.automate.api.policy).action = "get";
```
tylercloke: We are inconsistent with how we use `read` vs `get` in V1, but I don't see any `get` for list endpoints. Should probably be `read` but it also probably doesn't matter since no one will be using this in V1.

PR #439, opened 2019-05-29
Added equals operator that can only accept a single value

tylercloke:

Most of the rest of the code coming in follow up PR where we add the equals operator to the CreateRule API.

- [ ] Code actually executed?

### DIFF
--- a/components/authz-service/storage/v2/condition.go
+++ b/components/authz-service/storage/v2/condition.go
@@ -34,7 +34,7 @@ func (p *Condition) Scan(src interface{}) error {
 func NewCondition(ruleType RuleType, value []string,
 	attribute ConditionAttribute, operator ConditionOperator) (Condition, error) {
 
-	err := validateConditionInputs(value, attribute, ruleType)
+	err := validateConditionInputs(value, attribute, ruleType, operator)
 	if err != nil {
 		return Condition{}, err
 	}
@@ -47,10 +47,17 @@ func NewCondition(ruleType RuleType, value []string,
 	}, nil
 }
 
-func validateConditionInputs(value []string, attribute ConditionAttribute, ruleType RuleType) error {
+func validateConditionInputs(value []string,
+	attribute ConditionAttribute, ruleType RuleType, operator ConditionOperator) error {
 
 	if len(value) == 0 {
-		return errors.New("a condition must have a value list that is not empty")
+		return errors.New("a condition must specify one or more values")
+	}
+
+	if operator == Equals {
+		if len(value) != 1 {
+			return errors.New("a condition must specify exactly one value for 'equals' operator")
+		}
 	}
 
 	if ruleType == Event {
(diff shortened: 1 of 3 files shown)


components/authz-service/storage/v2/condition.go
```diff
@@ -47,12 +47,19 @@ func NewCondition(ruleType RuleType, value []string,
...
+func validateConditionInputs(value []string,
+	attribute ConditionAttribute, ruleType RuleType, operator ConditionOperator) error {
 
 	if len(value) == 0 {
 		return errors.New("a condition must have a value list that is not empty")
 	}
 
+	if operator == Equals {
+		if len(value) != 1 {
+			return errors.New("if the operator is equals, you must have a single entry in the value list")
```
msorens: Suggest just: "must specify exactly one value for 'equals' operator"

components/authz-service/storage/v2/condition.go
```diff
@@ -47,12 +47,19 @@ func NewCondition(ruleType RuleType, value []string,
 	}, nil
 }
 
-func validateConditionInputs(value []string, attribute ConditionAttribute, ruleType RuleType) error {
+func validateConditionInputs(value []string,
+	attribute ConditionAttribute, ruleType RuleType, operator ConditionOperator) error {
 
 	if len(value) == 0 {
 		return errors.New("a condition must have a value list that is not empty")
```
msorens: Suggest: "must specify one or more values"

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3333,6 +3333,17 @@ func TestCreateRule(t *testing.T) {
...
+			condition1, err := storage.NewCondition(storage.Event, []string{}, storage.ChefServer, storage.Equals)
+			assert.Equal(t, storage.Condition{}, condition1)
+			assert.Error(t, err)
+		},
+		"creating an equals condition with multiple entries returns an error": func(t *testing.T) {
+			condition1, err := storage.NewCondition(storage.Event,
+				[]string{"chef-server-1", "chef-server-2"}, storage.ChefServer, storage.Equals)
+			assert.Equal(t, storage.Condition{}, condition1)
+			assert.Error(t, err)
+		},
```
msorens: Suggest a couple more tests:
"multiple entries for a 'member-of' condition are allowed"
"single entry for a 'member-of' condition is allowed"

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3333,6 +3333,17 @@ func TestCreateRule(t *testing.T) {
 			_, err := storage.NewRule("new-id-1", projID, "name", storage.Node, []storage.Condition{})
 			assert.Error(t, err)
 		},
+		"creating a condition with zero entries returns an error": func(t *testing.T) {
```
msorens: Suggest explicitly say "for 'equals' operator" in this test,
and add another test for the member-of operator.

components/authz-service/storage/v2/postgres/postgres_test.go
tylercloke: We have a million tests that cover this case but I can add them explicitly.

Issue #443, opened 2019-05-29
Add Status Filters to Compliance Reporting  Profile Details View


jonong1972:
# User Story

As a Automate User, I want to filter my controls on the Compliance Reporting Profile detall view so that I may see which controls are failing.

# Definition of Done
The Profile Details page within a compliance report will have our standard filter buttons above the table.

For now we will just limit to:
All Controls | Failed Controls | Skipped Controls

# Demo Script / Repro Steps
1. User navigates to Compliance > Profiles Tab
2. Click on any profile
3. User can then click on a status filter above the Controls data table
4. System filters controls within table

vjeffrey:
related: #328

vjeffrey:
this is old; we are doing something different here; closing

PR #445, opened 2019-05-29
add a test for filtering by profile status

vjeffrey:
I wasn't clear on whether or not status filtering on compliance reporting profiles is supported or not, so i added a test. And it is!! 🎉 
Figured we should commit the test, so we know for sure that it's supported

### :+1: Definition of Done
test for filtering by status on compliance reporting profiles

### Related Resources
https://github.com/chef/automate/issues/443

### DIFF
--- a/components/compliance-service/api/tests/08_search_profiles_spec.rb
+++ b/components/compliance-service/api/tests/08_search_profiles_spec.rb
@@ -180,6 +180,30 @@ describe File.basename(__FILE__) do
     }.to_json
     assert_equal(expected_data, actual_data.to_json)
 
+    # Filter by status
+    actual_data = GRPC reporting, :list_profiles, Reporting::Query.new(
+        filters: [
+            Reporting::ListFilter.new(type: "end_time", values: ["2018-03-04T#{END_OF_DAY}"]),
+            Reporting::ListFilter.new(type: 'status', values: ['passed'])
+        ],
+        page: 1, per_page: 2)
+    expected_data = {
+        "profiles" => [
+            {
+                "name" => "nginx-baseline",
+                "title" => "DevSec Nginx Baseline",
+                "id" => "09adcbb3b9b3233d5de63cd98a5ba3e155b3aaeb66b5abed379f5fb1ff143988",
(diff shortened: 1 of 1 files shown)


vjeffrey:
in conversations with @rickmarry, he stated this may be reflecting the overall report status and not the profile status. i'm going to look into this locally.

vjeffrey:
i've confirmed it is doing the profile status.
basically i did a scan job with 3 profiles. 2 failed, one skipped. overall status is obviously failed, so i tried to pull the compliance/profiles endpoint with a filter of status: skipped and it only returned the skipped profile

components/compliance-service/api/tests/08_search_profiles_spec.rb
```diff
@@ -180,6 +180,30 @@ def reporting;
...
+                "version" => "2.1.0",
+                "status" => "passed"
+            }
+        ],
+        "counts" => {
+            "total" => 1,
+            "passed" => 1
+        }
+    }.to_json
+    assert_equal(expected_data, actual_data.to_json)
```
phiggins: We have a helper function for this kind of test: https://github.com/chef/automate/blob/53665d66aaeb9c4cebb1d54ce41fb16375665ae0/components/compliance-service/api/tests/run.rb#L105-L107

phiggins:
> basically i did a scan job with 3 profiles. 2 failed, one skipped. overall status is obviously failed, so i tried to pull the compliance/profiles endpoint with a filter of status: skipped and it only returned the skipped profile

Is there a way to test this scenario in code? Probably not worth redoing the ruby integration test data for this but it seems worth testing if we can.

components/compliance-service/api/tests/08_search_profiles_spec.rb
vjeffrey: ohh nice!

vjeffrey:
> Is there a way to test this scenario in code? 

i'm gonna parse through our data we send in and see if i can find a good example. if we run a skipped profile on a failed node and then try to filter for the skipped profile that should give us what we want

PR #448, opened 2019-05-29
[A2-833] Generalize project introspection

msorens:

IntrospectAllProjects was evaluating a subset of endpoints in determining the set of allowed projects--not taking into account _parameterized_ endpoints. It turns out that for determining projects we can safely ignore action/resource pairs, which makes the parameterized question moot to boot.

### :+1: Definition of Done

The global projects filter will populate with all allowed projects, irrespective of whether they are on parameterized endpoints or not.

image

```
export TOK=`chef-automate admin-token`

chef-automate iam upgrade-to-v2 --beta2.1 --skip-policy-migration

# Create a project
jo id='proj3' name='Project 3' | \
curl -sSkH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects --data @-  | jq -c

# Create a non-admin user
jo name=coolbean id=coolbean password=chefautomate | \
curl -sSkH "api-token: $TOK" https://localhost/apis/iam/v2beta/users --data @-  | jq -c

# save below as policy.json
{
  "id": "test",
  "name": "testing global project filter",
  "members": ["user:local:coolbean"],
  "projects": ["proj3"],
  "statements": [
    {
      "effect": "ALLOW",
      "actions": ["iam:teams:get"],
      "projects": ["proj3"]
    }
  ]
}
curl -sSkH "api-token: $TOK" -d @policy.json https://localhost/apis/iam/v2beta/policies | jq
```

(Above leverages sample repro steps on the jira card--thanks @bcmdarroch!)

With all that in place: 
- login as user `coolbean` and you will see `project-p1` as the one and only project in the global project filter.
- login as user `admin` and you will see that plus `(unassigned)` plus any others you happen to have in your system.

- [x] Code actually executed?


components/authz-service/engine/conformance/conformance_v2_test.go
```diff
@@ -499,19 +499,16 @@ func TestV2FilterAuthorizedProjects(t *testing.T) {
...
+		return ctx, engine.Subject(sub)
 	}
 
 	for desc, e := range engines {
 		t.Run(desc, func(t *testing.T) {
 
 			t.Run("no matching policies", func(t *testing.T) {
 				policy0 := map[string]interface{}{
-					"members": engine.Subject(sub),
+					"members": engine.Subject("team:local:foo"),
```
msorens: Test now needs to fail to find matches on the subject only.

components/authz-service/engine/opa/policy/introspection_v2_test.rego
```diff
@@ -98,57 +92,58 @@ test_authorized_project_ignores_system_policies {
...
 	actual_projects == {"proj1", "proj2", "proj3"}
 }
 
-test_authorized_project_include_all_allowed_policies_and_omits_denied_policies {
+# As long as some statement allows a project, it is possible for the user to see something there
+test_authorized_project_includes_project_both_allowed_and_denied {
 	actual_projects = authorized_project with data.policies.polid as {
 		"members": ["bob"],
 		"statements": {
-			# Allowed with single project
```
msorens: This complicated block is now more digestibly covered in the couple new tests.

components/authz-service/engine/opa/policy/introspection_v2.rego
```diff
@@ -46,7 +46,8 @@ authorized_pair[pair] {
 
 allowed_project[project] {
 	project := policies[pol_id].statements[statement_id].projects[_]
-	match_pair[["allow", _, pol_id, statement_id]]
+	"allow" == policies[pol_id].statements[statement_id].effect
+	authz.has_member[pol_id]
 	not policies[pol_id].type == const_system_type
```
srenatus: Is this still necessary? If it's a system policy allowing something, a why exclude it? (but I hadn't been around when this was introduced)

msorens: Indeed it is: system policies have wildcard projects, which means a normal user will get literally all projects as their allowed list. This line prevents that from happening.

srenatus: Hmm we could exclude policies with wildcard projects then, rather, couldn't we? 😃

msorens: No, we do not want to do that if it is not a system policy. The above effect _should_ result in all  projects if the user has that access.

srenatus: Ah I see. Thanks!

Issue #160, opened 2019-04-24
add search support for new fields supported by audit cookbook 7.5.0


vjeffrey:
# User Story

As part of the latest changes to the Audit cookbook, we added some fields to the reports. This work was done in support of IAMv2 work, but we can leverage this work to give users more search options in compliance reporting.
The newly added fields are:
- chef-server
- organization
- chef tags
- policy group
- policy name

This will require work to be completed in the reporting backend for:
- suggestions support for each of the fields
- filtering support for each of the fields

and work to be completed in the ui to add the new fields to the search-bar menu as well as ensuring the correct fields are being sent to the api.

# Definition of Done
Able to filter in compliance reporting ui by
 - chef-server
 - organization
 - chef tags
 - policy group
 - policy name

PR #450, opened 2019-05-29
Adding compliance search bar types

lancewf:
Screen Shot 2019-05-29 at 4 05 47 PM

Adding newly added fields to the compliance report to the search bar. 

The new types are: 
* chef-server
* organization
* chef tags
* policy group
* policy name

1. `build components/compliance-service && build components/automate-ui && start_all_services`
1. `send_inspec_example`
1. `chef_load_compliance_scans -N100 -D3 -M100 -T100`
1. Open https://a2-dev.test/compliance/reports/overview
1. Add a filter for the organization type with a value of 'tiny'
1. Ensure the suggestion is shown and that only the "My Compliant Node" node listed

https://github.com/chef/automate/issues/160

- [x] Code actually executed?


components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.ts
```diff
@@ -33,10 +33,28 @@ export class ReportingComponent implements OnInit, OnDestroy {
   // Query search bar
   availableFilterTypes = [
     {
-      'name': 'profile',
-      'title': 'Profile',
-      'description': 'Add the name or ID to filter this report against a profile',
-      'placeholder': 'Name or ID'
+      'name': 'chef_server',
```
lancewf: Added the new types to the UI and sorted them.

components/compliance-service/integration_test/reporting_server_list_nodes_test.go
```diff
@@ -6,13 +6,297 @@ import (
...
+	"golang.org/x/net/context"
+
 	authzConstants "github.com/chef/automate/components/authz-service/constants/v2"
 	"github.com/chef/automate/components/compliance-service/api/reporting"
 	reportingServer "github.com/chef/automate/components/compliance-service/api/reporting/server"
 	"github.com/chef/automate/components/compliance-service/reporting/relaxting"
 )
 
-func TestListNodes(t *testing.T) {
+func TestListNodesFiltering(t *testing.T) {
```
lancewf: Test for filtering with the new types.

components/compliance-service/integration_test/reporting_server_list_suggestions_test.go
```diff
@@ -15,6 +17,376 @@ import (
 	"github.com/chef/automate/components/compliance-service/reporting/relaxting"
 )
 
+func TestReportingListSuggestionsFiltering(t *testing.T) {
```
lancewf: Test for getting suggestions for the new types.

components/compliance-service/reporting/relaxting/reports.go
```diff
@@ -614,6 +614,31 @@ func (backend ES2Backend) getFiltersQuery(filters map[string][]string, latestOnl
 		boolQuery = boolQuery.Must(termQuery)
 	}
 
+	if len(filters["organization"]) > 0 {
```
lancewf: Adding the new types to the filters.

components/compliance-service/reporting/relaxting/suggestions.go
```diff
@@ -26,13 +26,18 @@ func (backend ES2Backend) GetSuggestions(typeParam string, filters map[string][]
...
 
 	var SUGGESTIONS_TYPES = map[string]string{
-		"environment": "environment",
-		"platform":    "platform.name",
-		"node":        "node_name",
-		"role":        "roles",
-		"recipe":      "recipes",
-		"profile":     "profiles.title",
-		"control":     "profiles.controls.title",
+		"environment":  "environment",
```
lancewf: Adding the new types to the suggestions.

components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.spec.ts
```diff
@@ -61,51 +61,26 @@ describe('ReportingComponent', () => {
...
+        'policy_group', 'policy_name', 'profile', 'recipe', 'role'];
+
+      expect(availableFilterTypesNames.length).toBe(expected.length);
+
+      // Each expected type is in the availableFilterTypesNames
+      expected.forEach(expectedType =>
+          expect(availableFilterTypesNames.indexOf(expectedType)).toBeGreaterThanOrEqual(0));
+
+      // Ensure availableFilterTypes.Titles are sorted
+      const availableFilterTypesTitles = component.availableFilterTypes.map( type => type.title);
```
lancewf: A test to ensure the types in the UI are sorted.

components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.spec.ts
```diff
@@ -61,51 +61,26 @@ describe('ReportingComponent', () => {
...
+
+      const expected = [
+        'chef_server', 'chef_tags', 'control', 'environment', 'node', 'organization', 'platform',
+        'policy_group', 'policy_name', 'profile', 'recipe', 'role'];
+
+      expect(availableFilterTypesNames.length).toBe(expected.length);
+
+      // Each expected type is in the availableFilterTypesNames
+      expected.forEach(expectedType =>
+          expect(availableFilterTypesNames.indexOf(expectedType)).toBeGreaterThanOrEqual(0));
```
phiggins: What does this do that testing for equality doesn't? This would allow `availableFilterTypesNames` to have elements not in `expected`, is that necessary behavior?

components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.spec.ts
```diff
@@ -61,51 +61,26 @@ describe('ReportingComponent', () => {
...
+
+      // Ensure availableFilterTypes.Titles are sorted
+      const availableFilterTypesTitles = component.availableFilterTypes.map( type => type.title);
+      const expectedAvailableFilterTypesTitle =
+        component.availableFilterTypes.map( type => type.title).sort(
+        (type1, type2) => (type1 < type2 ? -1 : 1));
```
phiggins: Does `sort` do something we don't want by default?

components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.spec.ts
phiggins: The length check above would preclude my concern above, but I'm still not sure what this is doing that an equality check wouldn't.

lancewf: I did not want to use equals because that would dictate the order. The order could be different than the `expected` array because the titles are what is being used to sort. 
I think it is correct that `availableFilterTypesNames` can have values not in `expected`. I would not want to fail when someone adds a new element to the search bar.

phiggins: If there isn't an assertion for "same elements in any order" you could do something like `expect(list1.sort()).toEqual(list2.sort())`.

components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.spec.ts
lancewf: This is sorting the types' titles then checking if the original titles were in that order.

components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.spec.ts
lancewf: I could not find an assertion for "same elements in any order" or ElementMatch in jasmine. 

I will try out your suggested code.

components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.spec.ts
phiggins: Right, but is `.sort((type1, type2) => (type1 < type2 ? -1 : 1))` different from the default `sort()` behavior?

lancewf: Oh, I thought I had to add a function to sort. Thanks for checking that. 
It should be the default. I will test it out.

components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.spec.ts
lancewf: That works `expect(expected.sort()).toEqual(availableFilterTypesNames.sort());`. 
Thanks. I am pushing it up now.

components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.spec.ts
lancewf: The default was the same. This is updated now.

PR #462, opened 2019-05-31
[a2-694] Add policy for applications service IAM v1 and v2

kmacgugan:
- For v1 add resource service_groups
- For v2 add resource applications:serviceGroups

- [ ] Code actually executed?


susanev:
do we want any docs changes related to this?
v1: https://automate.chef.io/docs/default-policies/
v2: https://automate.chef.io/docs/iam-v2-api-reference/

should we be updating the viewer and editor roles as well?

api/external/applications/applications.proto
```diff
@@ -21,41 +21,41 @@ service ApplicationsService {
     // TODO (dan, 2/2019): need to replace this once we have resources and such
```
srenatus: So can this comment now go away?

components/authz-service/storage/v1/storage.go
```diff
@@ -300,6 +300,14 @@ func DefaultPolicies() (map[string]*Policy, error) {
 			Effect:   "allow",
 			Version:  1,
 		},
+		constants.ApplicationsServiceGroupsPolicyID: {
```
bcmdarroch: You've added this policy to our system policies, which are hidden policies that govern permissions that Automate needs to work correctly (i.e. every user always needs to be able to get the license status so they won't be always blocked by the license modal). These policies never change.

My guess is that we should be able to change permissions on this API. In that case, instead of adding this policy, we'll want to [add the API's v2 action to the role(s)](https://github.com/chef/automate/blob/master/components/authz-service/storage/v2/storage.go#L218) that should have access. Maybe Viewer and Editor?

api/external/applications/applications.proto
```diff
@@ -21,41 +21,41 @@ service ApplicationsService {
...
     // TODO (afiune, 3/2019): need to replace this once we have resources and such
     // created in the auth system
     option (google.api.http).get = "/beta/applications/service_groups_health_counts";
-    option (chef.automate.api.policy).resource = "nodes";
+    option (chef.automate.api.policy).resource = "service_groups";
     option (chef.automate.api.policy).action = "list";
-    option (chef.automate.api.iam.policy).resource = "infra:nodes";
-    option (chef.automate.api.iam.policy).action = "infra:nodes:list";
+    option (chef.automate.api.iam.policy).resource = "applications:serviceGroups";
+    option (chef.automate.api.iam.policy).action = "applications:serviceGroups:list";
```
bcmdarroch: If you can GetServices, can you _always_ get their healthcounts and get them by SG?

components/authz-service/storage/postgres/migration/sql/45_add_application_policies.up.sql
```diff
@@ -0,0 +1,13 @@
+BEGIN;
+
+INSERT INTO policies
+    VALUES ('aee14d59-da0b-4974-ba6d-1a018b024874',
+            '{"action": "*", "effect": "allow", "resource": "applications:serviceGroups", "subjects": ["user:*"]}',
```
bcmdarroch:
```suggestion
            '{"action": "*", "effect": "allow", "resource": "service_group", "subjects": ["user:*"]}',
```

components/authz-service/storage/v1/storage.go
bcmdarroch: actually just kidding about that system policy comment! i misread the code. this is correct for adding a v1 default policy 👍 

for v2 permissions:
1. add `applications:*` to the [Editor role](https://github.com/chef/automate/blob/master/components/authz-service/storage/v2/storage.go#L229)
2. add `applications:serviceGroups:list` to the [Viewer role](https://github.com/chef/automate/blob/master/components/authz-service/storage/v2/storage.go#L244)
3. add a datamigration updating those two roles with their new actions (similar to [this migration](https://github.com/chef/automate/blob/master/components/authz-service/storage/postgres/datamigration/sql/06_v2_roles.up.sql))

susanev: If we make changes to these roles please also update the iam v2 docs, thx!

api/external/applications/applications.proto
kmacgugan: Yes, at least you should be able to.

bcmdarroch: gotcha, makes sense that all the resources/actions should be the same then. 👍

components/authz-service/storage/v2/storage.go
```diff
@@ -254,6 +255,7 @@ func DefaultRoles() []Role {
 			"event:*:list",
 			"ingest:*:get",
 			"ingest:*:list",
+			"applications:*:list",
```
bcmdarroch: is there a possibility there will be an individual `getService` in the near future? if so we may want to add `applications:*:get` as well while we're here

components/authz-service/storage/postgres/migration/sql/45_add_application_policies.up.sql
```diff
@@ -0,0 +1,24 @@
...
+    VALUES ('aee14d59-da0b-4974-ba6d-1a018b024874',
+            '{"action": "*", "effect": "allow", "resource": "service_groups", "subjects": ["user:*"]}',
+            CURRENT_TIMESTAMP,
+            1,
+            TRUE)
+    ON CONFLICT (id) DO UPDATE
+    SET policy_data='{"action": "list", "effect": "allow", "resource": "service_groups", "subjects": ["user:*"]}',
+        deletable=TRUE;
+
+UPDATE iam_roles
```
bcmdarroch: ah i forgot to mention one thing 😅 
hopefully last thing to fix: move L13-23 into a new migration file under the `postgres/datamigrations/sql` [directory](https://github.com/chef/automate/tree/master/components/authz-service/storage/postgres/datamigration/sql)

components/authz-service/storage/postgres/migration/sql/45_add_application_policies.up.sql
```diff
@@ -0,0 +1,24 @@
+BEGIN;
+
+INSERT INTO policies
+    VALUES ('aee14d59-da0b-4974-ba6d-1a018b024874',
+            '{"action": "*", "effect": "allow", "resource": "service_groups", "subjects": ["user:*"]}',
+            CURRENT_TIMESTAMP,
+            1,
+            TRUE)
+    ON CONFLICT (id) DO UPDATE
+    SET policy_data='{"action": "list", "effect": "allow", "resource": "service_groups", "subjects": ["user:*"]}',
```
msorens: This action must be the same as the action on L5.

components/authz-service/storage/v2/storage.go
```diff
@@ -234,6 +234,7 @@ func DefaultRoles() []Role {
 			"ingest:*",
 			"secrets:*",
 			"telemetry:*",
+			"applications:*",
```
msorens: This is incompatible with 45_add_application_policies.up.sql and will result in the entry being added to the database twice. Just delete this line and all is well.

Same for L258 below.

components/authz-service/storage/postgres/datamigration/sql/07_update_v2_roles.up.sql
```diff
@@ -0,0 +1,16 @@
...
+        actions = actions || '{applications:*:list, applications:*:get}'
+    WHERE
+        id = 'viewer';
+
+UPDATE iam_roles
+    SET
+        actions = actions || '{applications:*}'
+    WHERE
+        id = 'editor';
+END;
```
blakestier: What is this `END` doing that's distinct from `COMMIT`?

kmacgugan: probably i just wanted commit and end snuck in there. Commit will commit a transaction, end just delineates a block of code, like if it were in a loop. I think maybe I copied this structure from a loop somewhere?

kmacgugan:
@susanev I added docs for default policies.

api/external/applications/applications.proto
```diff
@@ -18,44 +18,42 @@ import "api/external/common/query/parameters.proto";
...
   };
   rpc GetServices(ServicesReq) returns (ServicesRes) {
     // TODO (afiune, 4/2019): need to replace this once we have resources and such
     // created in the auth system
     option (google.api.http).get = "/beta/applications/services";
-    option (chef.automate.api.policy).resource = "nodes";
+    option (chef.automate.api.policy).resource = "service_groups";
```
afiune: I don't think this endpoint has anything to do with service-groups here. This is only listing all services so we would probably need a different resource?

danielsdeleo: For any future readers of this, we decided to stick with `service_groups` because this matches the labeling in the UI, so that is what users will expect to see.

PR #464, opened 2019-05-31
Project filtering for reporting server ListProfiles endpoint.

phiggins:
#66 

In the process of debugging some test failures on this branch I tried cleaning up a few things and while this didn't end up being necessary, I kept the cleanups because I thought they were worth doing.


components/compliance-service/integration_test/reporting_server_list_suggestions_test.go
```diff
@@ -370,11 +370,12 @@ func TestReportingListSuggestionsFiltering(t *testing.T) {
 
 	for _, test := range cases {
 		t.Run(test.description, func(t *testing.T) {
-			suite.InsertInspecSummaries(test.summaries)
+			_, err := suite.InsertInspecSummaries(test.summaries)
+			require.NoError(t, err)
```
lancewf: Why remove the test for an error here?

PR #465, opened 2019-05-31
Re-sync introspection logging with code operation

msorens:

Introspection changed some time back,
to return results for every endpoint,
even if all methods were false.

### :+1: Definition of Done

Supervisor log accurately reflects introspection results.

1. rebuild components/automate-gateway
2. Enable debug logging in gateway:
```
 chef-automate debug set-log-level automate-gateway debug
```
3. Run `sl | grep Allowed` in hab studio to monitor the log.
4. Navigate around in the browser and observe log lines like these:
```
level=debug msg="Allowed paths: /iam/v2beta/policies[NONE]" ...
level=debug msg="Allowed paths: /retention/nodes/status[Get]" ...
level=debug msg="Allowed paths: /auth/users[Get,Post]" ...
```

- [x] Code actually executed?

### DIFF
--- a/components/automate-gateway/handler/authz.go
+++ b/components/automate-gateway/handler/authz.go
@@ -2,6 +2,7 @@ package handler
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	"github.com/grpc-ecosystem/go-grpc-middleware/logging/logrus/ctxlogrus"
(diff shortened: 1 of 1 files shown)


components/automate-gateway/handler/authz.go
```diff
@@ -250,14 +251,30 @@ func (a *AuthzServer) IntrospectAllProjects(
...
-		log.Debug("Allowed paths: NONE")
-		return
-	}
 	paths := make([]string, len(endpointMap))
 	i := 0
-	for k := range endpointMap {
-		paths[i] = k
+	for k, v := range endpointMap {
+		methods := []string{}
+		if v.Get == true {
```
srenatus: uhm... we could simplify this:
```suggestion
		if v.Get {
```
(same with the others below)

PR #467, opened 2019-05-31
[A2-798] split partial result update calls for v2 and v2.1

srenatus:

`authz-service` uses OPA's partial eval feature to save time when authorizing requests by "pre-calculating" everything it can figure out without knowing the (variable) input data. However, that process itself takes time, any change to the relevant data (policies, roles, projects) requires a rebuild of that entire state.

When introducing IAM v2.1, we've taken a bit of a shortcut, and have rebuilt both the v2 and the v2.1 queries' partial eval state. This PR is about fixing that.

The need to fix it arises from the rebuild time, which is dependent on the data it's built with, increasing when the data increases (linearly? haven't checked) -- and it's become noticable, and annoying.

#### ℹ️ Notes

- This PR also expands the current upgrade-to-v2-and-reset-to-v1 tests by going some extra loops through v2.1 as well, see [`integration/tests/upgrade_reset_iam_v2.sh`](https://github.com/chef/automate/pull/467/files#diff-c2011348cfa003a68d46851e7d3216c7)
- If you look at the commits, you'll find that I had to revert some of the cleanups I thought necessary. The need here was `ingest-service`, it seems to query the v2.1-related projects-rules stuff on ingest, regardless of whether A2 is using IAM v1, v2 or v2.1.

### :+1: Definition of Done

When using IAM v2.1, only the v2.1 partial eval result is rebuilt on updated data, and when using v2, only the v2 state is.

- `rebuild components/authz-service`
- upgrade to IAM v2.1 (`chef-automate iam upgrade-to-v2 --beta2.1`)
- create two projects and 30 policies, measuring the time it takes:
  - projects foo and bar: `curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects -d "$(jq -n '{ id: "foo", name: "my foo project" }')"` and `curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects -d "$(jq -n '{ id: "bar", name: "my bar project" }')"`
```
[118][default:/src:0]# time for i in $(seq 1 30); do curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/policies -d "$(sed "s/ID/$i/" pol.json)"; done
real    0m5.211s
user    0m0.252s
sys     0m0.060s
```
this is with `pol.json` containing
```
{
  "name": "testpolicy",
  "id": "testpolicy-id-ID",
  "members": [
    "team:local:viewers"
  ],
  "statements": [
    {
      "effect": "ALLOW",
      "actions": [],
      "role": "viewer",
      "projects": [
        "*"
      ]
    }
  ],
  "projects": [ "foo", "bar" ]
}
```
- delete those policies, 
```
[119][default:/src:2]# chef-automate dev psql chef_authz_service -- -c "delete from iam_policies where id ilike 'testpolicy-id-%'"
DELETE 30
```
- downgrade to v2, redo timing the 30 inserts:
```
[121][default:/src:130]# chef-automate iam upgrade-to-v2

Upgrading to IAM v2
Migrating v1 policies...
Creating default teams Editors and Viewers...
Skipped: Editors team already exists
Skipped: Viewers team already exists

Migrating existing teams...

Success: Enabled IAM v2
[122][default:/src:0]# time for i in $(seq 1 30); do curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/policies -d "$(sed "s/ID/$i/" _dump/pol.json)"; done
[output omitted]
real    8m41.688s
user    0m0.300s
sys     0m0.112s
```

⚠️ This is odd, isn't it? So, this doesn't solve the issue for IAM v2 customers; but it does promise us a bright and fast future in IAM v2.1. **Now I don't really understand (yet) why v2.1 is so much faster to rebuild**, but at least we don't drag v2 around 😅. Also, not updating v2.1 when updating v2 takes _some_ of the work away; combined with not updating the engine store twice, this hopefully helps a little.

- [A2-798](https://chefio.atlassian.net/browse/A2-798) 👈 this is about the perceivable delay 🐛 this hopes to improve.

- [x] Code actually executed?

### DIFF
--- a/components/authz-service/engine/opa/opa_internal_test.go
+++ b/components/authz-service/engine/opa/opa_internal_test.go
@@ -316,11 +316,11 @@ func BenchmarkInitPartialResultV2(b *testing.B) {
 
 	for d := 0; d < 5; d++ {
 		b.Run(fmt.Sprintf("default policies, run %d times", d), func(b *testing.B) {
-			for n := 0; n < b.N; n++ {
-				s, err := New(ctx, l)
-				require.NoError(b, err, "init state")
-				s.v2Store = store
+			s, err := New(ctx, l)
+			require.NoError(b, err, "init state")
+			s.v2Store = store
 
+			for n := 0; n < b.N; n++ {
 				for e := 0; e <= d; e++ {
 					r = s.initPartialResultV2(ctx)
 					if r != nil {
--- a/components/authz-service/server/v2/policy.go
+++ b/components/authz-service/server/v2/policy.go
@@ -116,15 +116,15 @@ func NewPoliciesServer(
 	}
 	srv.setVersion(v)
 
-	if err := srv.updateEngineStore(ctx); err != nil {
-		return nil, errors.Wrap(err, "initialize engine storage")
-	}
-
 	if v.Major == api.Version_V2 {
 		err = srv.store.ApplyV2DataMigrations(ctx)
 		if err != nil {
 			return nil, errors.Wrap(err, "error migrating v2 data")
 		}
+	}
+
+	if err := srv.updateEngineStore(ctx); err != nil {
+		return nil, errors.Wrap(err, "initialize engine storage")
 	}
 
 	return srv, nil
--- a/components/authz-service/server/v2/policy.go
+++ b/components/authz-service/server/v2/policy.go
@@ -117,14 +117,13 @@ func NewPoliciesServer(
 	srv.setVersion(v)
 
 	if v.Major == api.Version_V2 {
-		err = srv.store.ApplyV2DataMigrations(ctx)
-		if err != nil {
+		if err := srv.store.ApplyV2DataMigrations(ctx); err != nil {
 			return nil, errors.Wrap(err, "error migrating v2 data")
 		}
 	}
 
 	if err := srv.updateEngineStore(ctx); err != nil {
-		return nil, errors.Wrap(err, "initialize engine storage")
+		return nil, errors.Wrapf(err, "initialize engine storage (%v)", v)
 	}
 
 	return srv, nil
--- a/components/authz-service/server/v2/policy.go
+++ b/components/authz-service/server/v2/policy.go
@@ -122,6 +122,7 @@ func NewPoliciesServer(
 		}
 	}
 
+	// now that the data is all set, attempt to feed it into OPA:
 	if err := srv.updateEngineStore(ctx); err != nil {
 		return nil, errors.Wrapf(err, "initialize engine storage (%v)", v)
 	}
--- a/components/authz-service/server/v2/policy.go
+++ b/components/authz-service/server/v2/policy.go
@@ -28,9 +28,10 @@ import (
 type policyServer struct {
 	log             logger.Logger
 	store           storage.Storage
-	engine          engine.V2Writer
+	engine          engine.V2pXWriter
 	v1              storage_v1.PoliciesLister
 	vChan           chan api.Version
+	version         api.Version
 	policyRefresher PolicyRefresher
 }
 
@@ -46,7 +47,7 @@ type PolicyServer interface {
 func NewMemstorePolicyServer(
 	ctx context.Context,
 	l logger.Logger,
-	e engine.V2Writer,
+	e engine.V2pXWriter,
 	pl storage_v1.PoliciesLister,
 	vChan chan api.Version) (PolicyServer, error) {
 
@@ -57,7 +58,7 @@ func NewMemstorePolicyServer(
 func NewPostgresPolicyServer(
 	ctx context.Context,
 	l logger.Logger,
-	e engine.V2Writer,
+	e engine.V2pXWriter,
 	migrationsConfig migration.Config,
 	dataMigrationsConfig datamigration.Config,
 	pl storage_v1.PoliciesLister,
@@ -75,7 +76,7 @@ func NewPoliciesServer(
 	ctx context.Context,
 	l logger.Logger,
 	s storage.Storage,
-	e engine.V2Writer,
+	e engine.V2pXWriter,
 	pl storage_v1.PoliciesLister,
 	vChan chan api.Version) (PolicyServer, error) {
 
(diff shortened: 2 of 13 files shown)


components/authz-service/server/v2/policy.go
```diff
@@ -704,16 +705,12 @@ func (s *policyServer) EngineUpdateInterceptor() grpc.UnaryServerInterceptor {
 			// do nothing
 		}
 
-		s.log.Debugf("Initiating store update for %s", info.FullMethod)
```
srenatus: 🐛 drive-by bug fix

components/authz-service/server/v2/policy.go
```diff
@@ -119,6 +116,10 @@ func NewPoliciesServer(
 	}
 	srv.setVersion(v)
 
+	if err := srv.updateEngineStore(ctx); err != nil {
```
srenatus: ℹ️ this needs to happen _after_ `setVersion`, so the proper store will be updated in the process (v2 vs v2.1). (v1 is done per-call in its corresponding handlers)

components/authz-service/server/v2/policy_refresher.go
```diff
@@ -107,35 +114,39 @@ RUNLOOP:
...
-		status: make(chan error, 1),
+		ctx:     ctx,
+		status:  make(chan error, 1),
+		version: vsn,
 	}
 	refresher.refreshRequests <- m
 	return m.Err()
 }
 
+// TODO: handle version: v2 vs v2.1
```
srenatus: Let's fix that when we get there? 😉

components/authz-service/server/v2/policy_refresher.go
```diff
@@ -185,10 +185,10 @@ func (refresher *policyRefresher) updateEngineStore(ctx context.Context, vsn api
 	}
 
 	switch {
-	case vsn.Minor == api.Version_V1: // v2
```
srenatus: `E_IMANIDIOT`

components/authz-service/server/v2/policy_refresher.go
```diff
@@ -168,7 +179,14 @@ func (refresher *policyRefresher) updateEngineStore(ctx context.Context) error {
...
 
-	return refresher.engine.V2SetPolicies(ctx, policyMap, roleMap, ruleMap)
+	switch {
+	case vsn.Minor == api.Version_V1: // v2.1
+		return refresher.engine.V2p1SetPolicies(ctx, policyMap, roleMap, ruleMap)
+	default: // v2.0 OR v1.0
+		return refresher.engine.V2SetPolicies(ctx, policyMap, roleMap, ruleMap)
+	}
+	// Note 2019/06/04 (sr): v1?! Yes, IAM v1. Our POC code depends on this query to be
+	// answered regardless of whether IAM is v1, v2 or v2.1.
```
tylercloke: Helpful comment!

components/authz-service/server/v2/policy_refresher.go
```diff
@@ -295,3 +313,7 @@ func (refresher *policyRefresher) getRuleMap(_ context.Context) (map[string][]in
 	}
 	return data, nil
 }
+
+func pretty(vsn api.Version) string {
```
tylercloke: tenor-62368220

components/authz-service/server/v2/policy_test.go
```diff
@@ -3209,7 +3227,13 @@ type testEngine struct {
 }
 
 func (te *testEngine) V2SetPolicies(
-	_ context.Context, policies map[string]interface{},
+	ctx context.Context, policies map[string]interface{},
+	roles map[string]interface{}, rules map[string][]interface{}) error {
+	return te.V2p1SetPolicies(ctx, policies, roles, rules)
```
msorens: Perhaps add a comment explaining why it is OK to use the same test code for v2 and v2p1 here.

integration/tests/upgrade_reset_iam_v2.sh
```diff
@@ -33,41 +33,28 @@ hab_curl() {
 do_test_deploy() {
```
msorens: Nice!

srenatus:
So this most likely has issues with switching between v2 and v2.1 in a multi-node setting. I'm not sure how this is currently tested, so _thinking hard_ is the only method we have right now for figuring this out. I'm torn between adding a card outlining the problematic scenario; and adding some commits on top of this PR. I'd rather not do the latter, but since I have no idea who depends on this functionality when, I'm probably opting for that anyways.

components/authz-service/server/v2/policy_test.go
```diff
@@ -2933,15 +2936,26 @@ type testSetup struct {
 }
 
 func setupV2WithWriter(t *testing.T,
-	writer engine.V2Writer) testSetup {
-	return setupV2(t, nil, writer, nil, nil)
+	writer engine.V2pXWriter) testSetup {
+	return setupV2WithMigrationState(t, nil, writer, nil, make(chan api_v2.Version, 1),
+		func(s storage.MigrationStatusProvider) error { return s.Success(context.Background()) }) // IAM v2
```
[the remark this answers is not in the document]
srenatus: I think it does. I'll elaborate.

components/authz-service/engine/opa/opa_internal_test.go
```diff
@@ -316,11 +316,11 @@ func BenchmarkInitPartialResultV2(b *testing.B) {
 
 	for d := 0; d < 5; d++ {
 		b.Run(fmt.Sprintf("default policies, run %d times", d), func(b *testing.B) {
-			for n := 0; n < b.N; n++ {
-				s, err := New(ctx, l)
-				require.NoError(b, err, "init state")
-				s.v2Store = store
+			s, err := New(ctx, l)
+			require.NoError(b, err, "init state")
+			s.v2Store = store
```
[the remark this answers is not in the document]
srenatus: The intention was to only benchmark the relevant bits, not what happens in `s.New`. Setting the `s.v2Store` is not relevant, either, but also likely not a thing that influences measurements too much.

components/authz-service/server/v2/policy_refresher.go
```diff
@@ -107,35 +114,39 @@ RUNLOOP:
...
 	curPolicyID, err := refresher.store.GetPolicyChangeID(ctx)
 	if err != nil {
 		refresher.log.WithError(err).Warn("Failed to get current policy change ID")
 		return lastPolicyID, err
 	}
-	if curPolicyID != lastPolicyID {
+	if curPolicyID != lastPolicyID || forceUpdate {
 		refresher.log.WithFields(logrus.Fields{
 			"lastPolicyID": lastPolicyID,
 			"curPolicyID":  curPolicyID,
```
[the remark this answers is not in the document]
srenatus: I'll touch this code soon to fix the multi-node issue mentioned in the comments. I'll probably do this, then.

Issue #438, opened 2019-05-29
Client runs: search bar wildcard case insensitive filters


lancewf:
# User Story
As an Automate user, I want to filter with the Client Runs's search bar and wildcards where the case is insensitive. Currently filtering with the search bar all filters are case insensitive except when wildcards are added. 

# Definition of Done
When using the Error, Node name, or any other search bar type the filter value is case insensitive.

# Demo Script / Repro Steps

1. Create two nodes with names 'Chef-a2' and 'chef-a3'
1. Add a filter of 'chef-*' from the Client Runs page of type 'Node Name'
1. Ensure both nodes appear in the node list.

lancewf:
We want to look into this task now to see if there are any migrations needed. If there is a migration needed we want to merge the migrations with this PR. 
https://github.com/chef/automate/pull/476

PR #476, opened 2019-06-01
Migration node-state index to allow projects to be indexed

lancewf:
This change performs a migration on the node-state index to allow the projects field to be indexed. 

When the projects field was added it had two properties ('index=false' and 'doc_values=false') unintentionally added that prevented it from being index or queriable. 
```
"projects": {
    "type": "keyword",
    "index": false,
    "doc_values": false
},
```

Later an update was made to remove those fields from the node-state mapping to make the 'projects' field look like below. 
```
"projects": {
        "type": "keyword"
},
```
When the ingest-service was updated with this change the below error was logged and the mapping was not update. The ingest-service started with no problems. 
```
Error creating index node-state-5 with error: elastic: Error 400 (Bad Request): Mapper for [projects] conflicts with existing mapping in other types:\n[mapper [projects] has different [index] values, mapper [projects] has different [doc_values] values] [type=illegal_argument_exception]"
```

A second change is to exit the ingest-service when an error like above happens. 

#### Merged in "Client Runs search bar with case insensitive wildcard filters" PR
PR https://github.com/chef/automate/pull/516 is merged into this PR because they both need migration of the node-state index. 
 
#### Migration
1. To simulate the state of clients node-state mappings, change 'node-state-6' to 'node-state-5' here https://github.com/chef/automate/blob/master/components/ingest-service/backend/elastic/mappings/node_state.go#L326. 
1. Replace this line https://github.com/chef/automate/blob/master/components/ingest-service/backend/elastic/mappings/node_state.go#L65 with the below
```
    "type": "keyword",
    "index": false,
    "doc_values": false
```
1. `build components/ingest-service`
1. `start_all_services`
1. Ensure that the 'projects' field has the 'index' and 'doc_values' properties with this command  `curl $ELASTICSEARCH_URL/node-state/_mapping | jq . | less`
1. Watch the ingest-service logs ` sl | grep "ingest-service.default" &`
1. Revert all the changes to this file https://github.com/chef/automate/blob/master/components/ingest-service/backend/elastic/mappings/node_state.go
1. To simulate the update project and kick off the migration run `rebuild components/ingest-service`
1. Ensure no errors are in the logs then run `curl $ELASTICSEARCH_URL/node-state/_mapping | jq . | less` again to ensure that the 'projects' field look like below. 
```
"projects": {
        "type": "keyword"
},
```

#### Mapping update failure
1. Delete 'index' and 'doc_values' properties from the 'run_list' field from here https://github.com/chef/automate/blob/master/components/ingest-service/backend/elastic/mappings/node_state.go#L61
1. Watch the ingest-service logs with `sl | grep "ingest-service.default"&`
1. Apply the changes to the running service with `rebuild components/ingest-service/`
1. Ensure the ingest-service keeps failing to start with an error like the above one ("Error 400 (Bad Request)").
1. To fix the ingest-service remove the changes to this file https://github.com/chef/automate/blob/master/components/ingest-service/backend/elastic/mappings/node_state.go#L61 and `rebuild components/ingest-service/`

#### wildcard case insensitive filters
1. Update the node name https://github.com/chef/automate/blob/master/components/ingest-service/examples/converge-success-report.json#L2827 to "A2-prod"
1. Ingest the run `send_chef_run_example`
1. Update the node name again https://github.com/chef/automate/blob/master/components/ingest-service/examples/converge-success-report.json#L2827 to "a2-prod"
1. Ingest the run `send_chef_run_example`
1. Open the Client Runs page https://a2-dev.test/client-runs
1. Add a filter for `Node Name` value `a2-*`
1. Ensure both "A2-prod" and "a2-prod" are in the list
1. Remove all filters and add a filter for `Node Name` value `A2-*`
1. Again ensure both "A2-prod" and "a2-prod" are in the list

https://github.com/chef/automate/issues/438 of the merged in PR here https://github.com/chef/automate/pull/516 
Copied off this PR: https://github.com/chef/a2/pull/4587

- [x] Code actually executed?

### DIFF
--- a/components/config-mgmt-service/integration_test/nodes_test.go
+++ b/components/config-mgmt-service/integration_test/nodes_test.go
@@ -235,10 +235,1606 @@ func TestGetNodesRegexWithExactSameField(t *testing.T) {
 			},
 			expected: []string{"a2-dev", "a2-prod", "a1-prod", "a1-dev"},
 		},
+		{
+			description: "Case sensitive regex filters 1",
+			nodes: []iBackend.Node{
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "a2-dev",
+					},
+				},
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "A2-Dev",
+					},
+				},
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "A2-prod",
+					},
+				},
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "a2-Prod",
+					},
+				},
+			},
+			request: request.Nodes{
+				Filter: []string{"name:a2-*"},
+			},
+			expected: []string{"a2-dev", "A2-Dev", "A2-prod", "a2-Prod"},
+		},
+		{
+			description: "Case sensitive regex filters 2",
+			nodes: []iBackend.Node{
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "a2-dev",
+					},
+				},
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "A2-Dev",
+					},
+				},
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "A2-prod",
+					},
+				},
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "a2-Prod",
+					},
+				},
+			},
+			request: request.Nodes{
+				Filter: []string{"name:A2-*"},
+			},
+			expected: []string{"a2-dev", "A2-Dev", "A2-prod", "a2-Prod"},
+		},
 	}
 
 	for _, test := range cases {
 		t.Run(fmt.Sprintf("Regex with exact filter: %s", test.description), func(t *testing.T) {
+
+			// Adding required node data
+			for index := range test.nodes {
+				test.nodes[index].Exists = true
+				test.nodes[index].NodeInfo.EntityUuid = newUUID()
+			}
+
+			// Add node with project
+			suite.IngestNodes(test.nodes)
+			defer suite.DeleteAllDocuments()
+
+			// call GetNodes
+			res, err := cfgmgmt.GetNodes(context.Background(), &test.request)
+			assert.NoError(t, err)
+
+			names := getFieldValues(res, "name")
+
+			// Test what nodes are returned.
+			assert.ElementsMatch(t, test.expected, names)
+		})
+	}
+}
+
+func TestGetNodesWildcardCaseInsensitive(t *testing.T) {
+
+	cases := []struct {
+		description string
+		nodes       []iBackend.Node
+		request     request.Nodes
+		expected    []string
+	}{
+		// Node name
+		{
+			description: "Node name: term lowercase with values with varying cases",
+			nodes: []iBackend.Node{
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "a2-dev",
+					},
+				},
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "A2-Dev",
+					},
+				},
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "A2-prod",
+					},
+				},
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "a2-Prod",
+					},
+				},
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "a3-Prod",
+					},
+				},
+			},
+			request: request.Nodes{
+				Filter: []string{"name:a2-*"},
+			},
+			expected: []string{"a2-dev", "A2-Dev", "A2-prod", "a2-Prod"},
+		},
+		{
+			description: "Node name: term uppercase with values with varying cases",
+			nodes: []iBackend.Node{
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "a2-dev",
+					},
+				},
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "A2-Dev",
+					},
+				},
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "A2-prod",
+					},
+				},
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "a2-Prod",
+					},
+				},
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "a3-Prod",
+					},
+				},
+			},
+			request: request.Nodes{
+				Filter: []string{"name:A2-*"},
+			},
+			expected: []string{"a2-dev", "A2-Dev", "A2-prod", "a2-Prod"},
+		},
+
+		// organization
+		{
+			description: "Org: term lowercase with values with varying cases",
+			nodes: []iBackend.Node{
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName:         "a",
+						OrganizationName: "a2-dev",
+					},
+				},
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName:         "b",
+						OrganizationName: "A2-Dev",
+					},
+				},
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName:         "c",
+						OrganizationName: "A2-prod",
+					},
+				},
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName:         "d",
+						OrganizationName: "a2-Prod",
+					},
+				},
+				{
+					NodeInfo: iBackend.NodeInfo{
+						NodeName:         "e",
+						OrganizationName: "a3-Prod",
+					},
+				},
+			},
+			request: request.Nodes{
+				Filter: []string{"organization:a2-*"},
+			},
+			expected: []string{"a", "b", "c", "d"},
+		},
+		{
+			description: "Org: term uppercase with values with varying cases",
+			nodes: []iBackend.Node{
+				{
+					NodeInfo: iBackend.NodeInfo{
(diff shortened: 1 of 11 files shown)


components/ingest-service/backend/elastic/elastic.go
```diff
@@ -241,53 +241,70 @@ func (es *Backend) createBulkUpdateRequestToIndexWithID(
...
+				err = es.createStoreAliasIfNotExists(ctx, esMap.Alias, esMap.Index)
+				if err != nil {
+					return err
+				}
 			}
 		}
 	}
 	es.initialized = true
+
+	return nil
```
lancewf: We are now returning an error for problems initializing elasticsearch.

components/ingest-service/backend/elastic/elastic.go
```diff
@@ -241,53 +241,70 @@ func (es *Backend) createBulkUpdateRequestToIndexWithID(
...
 			if esMap.Timeseries {
-				es.createTemplate(ctx, esMap.Index, esMap.Mapping)
+				err := es.createTemplate(ctx, esMap.Index, esMap.Mapping)
+				if err != nil {
+					return err
+				}
 			} else {
-				es.createOrUpdateStore(ctx, esMap)
-				es.createStoreAliasIfNotExists(ctx, esMap.Alias, esMap.Index)
+				err := es.createOrUpdateStore(ctx, esMap)
```
lancewf: An error will be returned here for node-state mapping updates. This would have prevented the 'projects' field problem.

components/ingest-service/grpc/grpc.go
```diff
@@ -45,7 +48,7 @@ func Spawn(opts *serveropts.Opts) error {
 
 	var (
 		grpcServer  = opts.ConnFactory.NewServer()
-		A1migration = migration.New(client)
+		migrationer = migration.New(client)
```
lancewf: Updated the variable name because it is not only for A1 migrations.

components/ingest-service/migration/migration.go
```diff
@@ -14,11 +14,14 @@ import (
 var (
 	// This variable will allow us to add a test so that if the mappings.NodeState.Index
 	// changes we might need to add some migration bits over here
-	a2CurrentNodeStateIndex = "node-state-5"
+	a2CurrentNodeStateIndex = "node-state-6"
```
lancewf: Adding a new migration from node-state-5 to node-state-6.

components/ingest-service/grpc/grpc.go
```diff
@@ -32,6 +33,8 @@ import (
 //
 // Maybe even spawn multiple servers
 func Spawn(opts *serveropts.Opts) error {
+	var client backend.Client
```
afiune: What is this for? 🤔

afiune: curiour: What is this for? 🤔

components/ingest-service/grpc/grpc.go
```diff
@@ -57,24 +60,33 @@ func Spawn(opts *serveropts.Opts) error {
...
 
 	// Initialize elasticsearch indices or trigger a migration
 	//
 	// This initialization will happen inside a goroutine so that we can then
 	// report the health of the system. We might have to do the same for the
 	// migration tasks.
-	migrationNeeded, err := A1migration.MigrationNeeded()
+	migrationNeeded, err := migrationer.MigrationNeeded()
 	if err != nil {
+		log.WithFields(log.Fields{"error": err}).Fatal("Failed checking for migration")
```
afiune: nit: Sine all these logs are just adding an error as a field, I would suggest:
```suggestion
		log.WithError(err).Fatal("Failed checking for migration")
```

components/ingest-service/grpc/grpc.go
```diff
@@ -57,24 +60,33 @@ func Spawn(opts *serveropts.Opts) error {
...
 		return err
 	}
 
 	if migrationNeeded {
-		A1migration.Start()
+		err = migrationer.Start()
+		if err != nil {
+			log.WithFields(log.Fields{"error": err}).Fatal("Failed starting a migration")
```
afiune:
```suggestion
			log.WithError(err).Fatal("Failed starting a migration")
```

components/ingest-service/grpc/grpc.go
```diff
@@ -57,24 +60,33 @@ func Spawn(opts *serveropts.Opts) error {
...
+			return err
+		}
 	} else {
-		A1migration.MarkUnneeded()
-		client.InitializeStore(context.Background())
+		migrationer.MarkUnneeded()
+		err = client.InitializeStore(context.Background())
+		if err != nil {
+			log.WithFields(log.Fields{"error": err}).Fatal("Failed initializing elasticsearch")
```
afiune:
```suggestion
			log.WithError(err).Fatal("Failed initializing elasticsearch")
```

components/ingest-service/grpc/grpc.go
lancewf: I put this here to have the client be a `backend.Client` and not a `elastic.Backend`.

afiune:
@lancewf Is this last commit a new modification that you're trying to include in this PR? or it is related to the same project indexing problem? 🤔

lancewf:
@afiune The last commit is from this PR https://github.com/chef/automate/pull/516. It was merged in with this PR because they both need to migrate the node-state index. So, instead of doing two migrations (one for each PR) this will allow us to do only one migration.

components/ingest-service/backend/elastic/mappings/node_state.go
```diff
@@ -121,6 +129,7 @@ var nodeProps = `
 	},
 	"resource_names": {
 		"type": "keyword",
+		"normalizer": "case_insensitive",
```
lancewf: The `case_insensitive` normalizer is needed to allow wildcard filters to be case insensitive.

components/ingest-service/backend/elastic/mappings/node_state.go
```diff
@@ -337,6 +365,13 @@ var NodeState = Mapping{
 				"refresh_interval": "5s"
 			},
 			 "analysis": {
+				"normalizer": {
+					"case_insensitive": {
+						"type": "custom",
+						"char_filter": [],
+						"filter": ["lowercase", "asciifolding"]
+					}
+				},
```
lancewf: Adding the `case_insensitive` normalizer. This change requires migration of the node-state index.

components/ingest-service/backend/elastic/mappings/node_state.go
```diff
@@ -262,10 +280,18 @@ var nodeProps = `
...
 	"source_fqdn": {
-		"type": "keyword"
+		"type": "keyword",
+		"normalizer": "case_insensitive",
+		"fields": {
+			"engram" : {
+				"type": "text",
+				"analyzer": "autocomplete"
+			}
+		}
```
lancewf: Adding the `source_fqdn` and `organization_name` field to allow wildcard filters.

components/config-mgmt-service/integration_test/suggestions_test.go
```diff
@@ -734,13 +734,13 @@ func TestSuggestionsWithTableDriven(t *testing.T) {
 		// Suggestions for Policy Group
 		{"should return all policy_group suggestions",
 			request.Suggestion{Type: "policy_group"},
-			[]string{"sports", "tv_series", "RPGs", "heros", "movies", "nintendo"}},
+			[]string{"sports", "tv_series", "rpgs", "heros", "movies", "nintendo"}},
```
lancewf: Suggestions will all come back now in lowercase.

components/config-mgmt-service/integration_test/nodes_test.go
```diff
@@ -235,6 +235,64 @@ func TestGetNodesRegexWithExactSameField(t *testing.T) {
...
+					NodeInfo: iBackend.NodeInfo{
+						NodeName: "a2-Prod",
+					},
+				},
+			},
+			request: request.Nodes{
+				Filter: []string{"name:A2-*"},
+			},
+			expected: []string{"a2-dev", "A2-Dev", "A2-prod", "a2-Prod"},
+		},
```
phiggins: Are these tests doing something the ones below aren't?

components/config-mgmt-service/integration_test/suggestions_test.go
phiggins: Could this cause a problem with suggestions not being specific enough? If I have things named `a2-Dev` and `A2-dev`, the suggestion would be `a2-dev`?

components/ingest-service/grpc/grpc.go
phiggins: This might fail the eventual spellcheck run. `migrator` is used in a few places though.

components/config-mgmt-service/integration_test/suggestions_test.go
lancewf: That is correct. Only the `a2-dev` suggestion would be returned. If you selected that suggestion both nodes would be shown. With this change, `A2-dev` and `a2-dev` are treated as the same term.

components/ingest-service/grpc/grpc.go
lancewf: I thought `migrator` was a word https://en.oxforddictionaries.com/definition/migrator

components/config-mgmt-service/integration_test/nodes_test.go
lancewf: Good catch! Those are not supposed to be there. I had them there initially when I was working on a solution. 

Removing now.

vjeffrey:
pulling this change down now to test it out

components/ingest-service/grpc/grpc.go
phiggins: That's what I meant, I think `migrator` is a word but I'm not sure `migrationer` is.

lancewf: Thanks, I did not see that. I am updating now.

PR #477, opened 2019-06-01
Reporting cleanup

phiggins:
Extracting code from #464 while I debug test failures on that branch.

### DIFF
--- a/components/compliance-service/reporting/relaxting/profiles.go
+++ b/components/compliance-service/reporting/relaxting/profiles.go
@@ -641,8 +641,6 @@ func (backend ES2Backend) getProfileMinsFromNodes(
 	filters map[string][]string) (map[string]reporting.ProfileMin, *reportingapi.ProfileCounts, error) {
 	myName := "getProfileMinsFromNodes"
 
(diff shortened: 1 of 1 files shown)


components/compliance-service/reporting/relaxting/profiles.go
```diff
@@ -605,17 +602,13 @@ func (backend *ES2Backend) GetAllProfilesFromNodes(from int32, size int32, filte
...
 
 	LogQueryPartMin(esIndex, searchResult, fmt.Sprintf("%s - search result", myName))
 
 	profiles := make([]*reportingapi.ProfileMin, 0)
-	if err != nil {
-		// should we be returning the error here...instead of logging it and then returning nil for the error?
-		logrus.Errorf("%s: Error while trying to get data from ES for index: %s error: %s",
-			myName, esIndex, err.Error())
-		return profiles, nil, nil
-	}
```
phiggins: I was getting a null pointer exception in a test before I changed this. Ignoring the error and continuing with potentially half-baked data seems like a bad idea.

PR #478, opened 2019-06-01
Integration test error checking

phiggins:
Extracting code from #464 while I debug test failures on that branch.

### DIFF
--- a/components/compliance-service/integration_test/reports_projects_update_test.go
+++ b/components/compliance-service/integration_test/reports_projects_update_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/golang/protobuf/ptypes"
 	_struct "github.com/golang/protobuf/ptypes/struct"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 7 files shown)


components/compliance-service/integration_test/reports_projects_update_test.go
```diff
@@ -1293,35 +1294,37 @@ func TestProjectUpdate(t *testing.T) {
 	for _, test := range cases {
 		t.Run(fmt.Sprintf("report match: %s", test.description),
 			func(t *testing.T) {
-				suite.InsertInspecReports([]*relaxting.ESInSpecReport{test.report})
-				suite.InsertInspecSummaries([]*relaxting.ESInSpecSummary{test.summary})
+				_, err := suite.InsertInspecReports([]*relaxting.ESInSpecReport{test.report})
+				require.NoError(t, err)
+				_, err = suite.InsertInspecSummaries([]*relaxting.ESInSpecSummary{test.summary})
+				require.NoError(t, err)
```
phiggins: In #464 I'm seeing failures from this test when the changes in that branch are seemingly unrelated. Previously this would fail silently.

PR #480, opened 2019-06-02
[A2-832] Populate global project filter with names instead of ids

msorens:

This updates the global project filter to display project names instead of project ids.

It turns out that this rather benign-sounding task involved more intricacies than one might expect.
First, one must deal with the fact that every UI call to the back-end now includes _requested projects_ (i.e. the current selection from the global project filter). But if that is done for the call that populates the global project filter it would restrict its universe of discourse to just the currently selected projects, when it should always be looking at all allowed projects.

This could be solved in a number of ways. The back-end could, for example, just be aware of the situation and choose to ignore the project headers that were sent. Another approach is to provide a mechanism for front-end calls to opt-out of sending the requested projects at all. I chose to implement the latter for two reasons. First, we now have an easily-available mechanism for future work that might require it. Second, and more importantly, I ended up re-using some existing back-end code and did not want to clutter it with this special case. 

As to that "re-using some existing back-end code"...  The next problem was how to marry project names with introspected project IDs. This also had a number of solutions. The front-end could do two calls, one for IntrospectAllProjects and one for ListProjects, then marry the data. But then there's that finickly special case where the latter never returns "(unassigned)" while that would possibly show up in the former. So more special casing there it would seem.

Another solution is to do essentially the same in the back-end: get the introspection result, get the projects list, and marry them. But same issue with "(unassigned)" there, too. 

And then I had an epiphany that we actually did not need IntrospectAllProjects on the back-end at all! 
Didn't ListProjects already return everything we need--the project names with the project IDs and, as long as project filtering was disabled, it would be a complete list--oh, except for "(unassigned)". 

So let's talk about "(unassigned)". A couple solutions there as well: add a parameter to the ListProjects API method to allow including it in the list. Ugly. A cleaner way is to in fact keep the IntrospectAllProjects endpoint but rather than invoke the introspection mechanism as it used to, simply call ListProjects on authz with a now internal parameter that lets it know to include "(unassigned)". This was an almost trivial code change to solve the problem.

**UPDATE:**
So, the above is not all true. In order to make that work to this point,
I had to adjust the action on IntrospectAllProjects in the proto file to *not* be under iam:introspect:*.
Without that change IntrospectAllProjects returns *all* projects rather than all *allowed* projects (because the system policy for iam:introspect:* includes projects: *).
BUT... and that's a big but I realized last night--that means that everyone would need
a special policy reference for the new action. Ugh. (So while things worked for good ol' user bob here, it did not work for mary who was just made a 'viewer'.)

OK, so re-jiggering the new pieces, components/automate-gateway/handler/iam/v2beta/policy/policy.go::IntrospectAllProjects now gets the correct result for bob and mary: It does require the original introspection function to fetch the correct set of project IDs (getAllowedProjects). It then, blends that with the project names retrieved from ListProjects--which still returns all projects, not just all allowed projects. But now that is OK, because only the ones from `getAllowedProjects` are included in the constructed result. 

One bit of work remaining: delete the now unused IntrospectAllProjects v1-only endpoint.

### :+1: Definition of Done

image

Components to build:
- authz-service
- automate-gateway

Create project-p1, project-p2, and project-p3 (UI or CLI).

Create local user "bob".

Create this policy for bob. 
```
{
  "id": "filtering-test",
  "name": "testing filtering of a list by project",
  "projects": [],
  "statements": [
    {
      "effect": "ALLOW",
      "projects": ["project-p3","project-p2","(unassigned)"]
    }
  ]
}
```
This policy includes a subset of the 3 projects above, not all of them. That allows one to see that the global projects filter gets just _allowed_ projects rather than all projects.

This policy also gives bob access to "(unassigned)" so one can see that does show up in the list.

After logging in as bob you should find that bob's global project filter looks like the screenshot above,
containing the three projects in the policy above.

Go to the projects list view and you can see entries appear when either project-p2 or project-p3 is selected.
Go to the teams list view and you can see entries appear when "(unassigned)" is selected.

- [x] Code actually executed?


susanev:
@msorens apologies i had a comment in that card and added it to the body once you picked it up...

would like for _(unassigned)_ to stay as is.. it was well received in all usability studies at chef conf.

api/interservice/authz/v2/project.proto
```diff
@@ -32,7 +32,9 @@ message GetProjectResp {
     Project project = 1;
 }
 
-message ListProjectsReq {}
+message ListProjectsReq {
+    bool globalFilterView = 1;
```
msorens: This name describes the usage, not what it actually does. That is, if you want to use authz/ListProjects for the global projects filter, you set this flag. Alternately, could name it for what it does "alsoIncludeUnassignedWhereApplicable" if there is strong sentiment for this.

components/authz-service/server/v2/project.go
```diff
@@ -19,7 +18,7 @@ import (
 	api "github.com/chef/automate/api/interservice/authz/v2"
 	automate_event "github.com/chef/automate/api/interservice/event"
 	"github.com/chef/automate/components/authz-service/config"
-	v2_constants "github.com/chef/automate/components/authz-service/constants/v2"
+	constants_v2 "github.com/chef/automate/components/authz-service/constants/v2"
```
msorens: Drive-by: the uses of "constants_v2" far outweighed "v2_constants" in the codebase so I changed those few of the latter to the former.

components/authz-service/server/v2/project.go
```diff
@@ -205,8 +203,10 @@ func (s *state) ListProjects(ctx context.Context,
 			return nil, status.Errorf(codes.Internal,
 				"error converting project with ID %q: %s", p.ID, err.Error())
 		}
-		// exclude "all projects" meta-project from the API
-		if !stringutils.SliceContains(systemProjects, apiProject.Id) {
+		// Exclude "all projects" meta-project from the API.
+		// Exclude "unassigned" meta-project from the API--except for introspection!
+		if (apiProject.Id != constants_v2.AllProjectsID) &&
+			(req.GlobalFilterView || (apiProject.Id != constants_v2.UnassignedProjectID)) {
```
msorens: This is where GlobalFilterView does its work.

components/automate-gateway/api/iam/v2beta/policy.proto
```diff
@@ -177,4 +177,12 @@ service Policies {
     option (chef.automate.api.iam.policy).resource = "system:iam:resetToV1";
     option (chef.automate.api.iam.policy).action = "system:iam:reset";
   };
+  rpc IntrospectAllProjects (ListProjectsReq) returns (ListProjectsResp) {
+    option (google.api.http).get = "/iam/v2beta/introspect_projects";
+    option (chef.automate.api.policy).resource = "auth_introspection:introspect_all_projects";
+    option (chef.automate.api.policy).action = "read";
+    option (chef.automate.api.iam.policy).resource = "iam:introspect";
+    option (chef.automate.api.iam.policy).action = "iam:projects:getAllProjects";
```
msorens: Re-purposing what IntrospectAllProjects does--and at the same time putting it on the /iam/v2beta endpoint path, where it belongs. There is one anomaly here: Had to adjust the action to *not* be under `iam:introspect:*`.
Without that change IntrospectAllProjects returns *all* projects rather than all *allowed* projects (because the system policy for iam:introspect:* includes projects: `*`).

components/automate-gateway/handler/iam/v2beta/policy/policy.go
```diff
@@ -424,6 +424,25 @@ func (p *Server) ListProjects(
...
+	for i, proj := range resp.Projects {
+		apiProj, err := domainProjectToAPIProject(proj)
+		if err != nil {
+			return nil, err
+		}
+		projects[i] = apiProj
+	}
+	return &pb_resp.ListProjectsResp{Projects: projects}, nil
+}
+
```
msorens: TODO: extract common code from ListProjects and IntrospectAllProjects in this file into a helper function to be DRYer.

components/automate-ui/src/app/pages/user-details/user-details.component.ts
```diff
@@ -1,4 +1,4 @@
-import { combineLatest as combineLatest,  Observable ,  Subscription } from 'rxjs';
+import { combineLatest,  Observable ,  Subscription } from 'rxjs';
```
msorens: Drive-by: remove redundant bit.

components/automate-ui/src/app/services/http/http-client-auth.interceptor.spec.ts
```diff
@@ -123,6 +123,45 @@ describe('HttpClientAuthInterceptor', () => {
...
+
+      using([
+        ['does not include project header', true],
+        ['includes project header', false]
+      ], function (description: string, setting: boolean) {
+
+        it(description + 'when unfiltered flag set to ' + setting, done => {
+          const options = {
+            params: new HttpParams().set('unfiltered', String(setting))
+          };
```
msorens: Any call that needs to say "do not send project header" can now just pass this special `unfilitered` parameter set to true.

components/automate-ui/src/app/services/projects-filter/projects-filter.effects.ts
```diff
@@ -1,11 +1,14 @@
 import { Injectable } from '@angular/core';
```
msorens: This file contains what is really all the front-end tweaks needed to use names instead of ids.

msorens:
@susanev: Changing "Unassigned Project" to "(unassigned)" is actually quite separate from the work in this PR so far (it just needs a data migration), but I can certainly add that in.

blakestier:
This is still WIP so I haven't reviewed yet, but something feels off about modifying the back end for the sake of a single front-end call. Is it crazy to add `unassigned` as part of the front-end work?

msorens:
@blakestier  You raise two separate issues that are not actually related.
1. The motivation for modifying the back-end call is because we need it to return different data (ids and names instead of just ids). Nothing to do with "unassigned".
2. Issues with "unassigned" just made that work a bit more complicated than expected.

components/automate-gateway/handler/iam/v2beta/policy/policy.go
msorens: Roger. Point is sort of moot now anyway, with the latest commit.

api/interservice/authz/v2/project.proto
msorens: So @srenatus has a comment below about unifying the terms "introspection" and "globalFilterView", since this is really need just for introspection. So as a third choice, we could rename this to just "introspection". Please vote for your choice here!

components/authz-service/server/v2/project.go
```diff
@@ -205,8 +203,10 @@ func (s *state) ListProjects(ctx context.Context,
 			return nil, status.Errorf(codes.Internal,
 				"error converting project with ID %q: %s", p.ID, err.Error())
 		}
-		// exclude "all projects" meta-project from the API
-		if !stringutils.SliceContains(systemProjects, apiProject.Id) {
+		// Exclude "all projects" meta-project from the API.
+		// Exclude "unassigned" meta-project from the API--except for introspection!
```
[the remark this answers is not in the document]
msorens: "Merged" this with an earlier comment I had started--see above.

components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
```diff
@@ -27,12 +27,15 @@ export class HttpClientAuthInterceptor implements HttpInterceptor {
 
   intercept(request: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
```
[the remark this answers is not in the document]
msorens: Yes! Thanks for that. We should definitely prefer `unknown` over `any`. Interestingly, this will be the first use in our code base!

api/interservice/authz/v2/project.proto
srenatus: ⚠️ let's investigate why `protoc-gen-lint` doesn't yell at you there -- it should be `global_filter_view`, and it should be raised [like this one](https://buildkite.com/chef-oss/chef-automate-master-verify/builds/1677#fd0dc244-d0c7-4d0c-b0ad-750a34bcec7e/326-708)

components/authz-service/server/v2/project.go
srenatus: Fine. It is a bit distracting, though 😉 But it's a separate commit, so there's nothing to be desired 😄

components/automate-gateway/gateway/services.go
```diff
@@ -73,7 +73,8 @@ func (s *Server) RegisterGRPCServices(grpcServer *grpc.Server) error {
 	if err != nil {
 		return errors.Wrap(err, "create client for deployment service")
 	}
-	pb_deployment.RegisterDeploymentServer(grpcServer, handler.NewDeploymentServer(deploymentClient))
```
srenatus: now this is a bit hidden in your "Remove deprecated v1 IntrospectAllProjects" commit, isn't it? https://github.com/chef/automate/pull/480/commits/f4d4c74dff09c6d518c11e5b9028327afcadb255

I know it'll eventually get squashed, but still 😉

srenatus: Also I'd opt for either _changing the linter settings to flag this_ or _not caring_ 🙈 -- fixing stuff like this that isn't detected by the linter is an uphill battle, I suppose 😉

components/automate-gateway/handler/iam/v2beta/policy/policy.go
```diff
@@ -424,6 +430,62 @@ func (p *Server) ListProjects(
...
+func mapProjectNamesToIds(
+	projectIDs []string, refProjects []*authz.Project) (*pb_resp.ListProjectsResp, error) {
+	// refProjects is *all* projects, not just all *allowed* projects.
+	// This uses the correctly filtered id list to generate the desired return list.
+	projMap := toMap(refProjects)
+	projects := make([]*pb_common.Project, len(projectIDs))
+	for i, id := range projectIDs {
+		apiProj, err := domainProjectToAPIProject(projMap[id])
```
srenatus: This could fail, I believe. `projMap[id]` will be `nil` if no project with that id is in `refProjects`. Now, I know that this should be highly unlikely, but when facing concurrent user/API requests, this could perhaps still happen.
```suggestion
		domainProj, ok := projMap[id]
		if !ok {
			// deal with it
		}
		apiProj, err := domainProjectToAPIProject(domainProj)
```

components/automate-ui/src/app/pages/user-details/user-details.component.ts
srenatus: Spaces look fishy, though 🐟  Since you've touched the line, let's do
```suggestion
import { combineLatest, Observable, Subscription } from 'rxjs';
```

components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
```diff
@@ -27,12 +27,15 @@ export class HttpClientAuthInterceptor implements HttpInterceptor {
...
     let headers = request.headers.set('Authorization', `Bearer ${this.chefSession.id_token}`);
-    if (this.projects) {
+    const filtered = request.params.get('unfiltered') !== 'true';
+    const params = request.params.delete('unfiltered');
```
srenatus: ❓ Is one endpoint used for both filtered and unfiltered requests? If so, fine; if not, we could also have the endpoint ignore an incoming list of projects, I suppose.

components/automate-gateway/handler/iam/v2beta/policy/policy.go
```diff
@@ -424,6 +430,62 @@ func (p *Server) ListProjects(
...
+func (p *Server) IntrospectAllProjects(
+	ctx context.Context, req *pb_req.ListProjectsReq) (*pb_resp.ListProjectsResp, error) {
+
+	projectIDs, err := p.getAllowedProjects(ctx)
+	if err != nil {
+		return nil, err
+	}
+	// To correlate names with project IDs, use the standard ListProjects
+	// with the special flag to include "(unassigned)" too.
+	resp, err := p.projects.ListProjects(ctx, &authz.ListProjectsReq{GlobalFilterView: true})
```
bcmdarroch: I don't feel that ListProjects needs to change here. I can't think of another scenario when ListProjects would need to return the `unassigned` project, so it seems like this is leaking concerns between two APIs.
Instead of tying ListProjects to IntrospectAllProjects, could we handle this all within the IntrospectAllProjects API? I'm thinking of a helper that checks projectIDs for `unassigned` and if found adds an `unassigned` Project to the list of projects.

msorens: I agree that sounds nicer, @bcmdarroch , except it means that IntrospectAllProjects needs a further call to `GetProject("(unassigned)")` to find out its name. But, on balance, that sounds reasonable.

components/automate-gateway/gateway/services.go
msorens: It is an uphill battle, yes, but I really find it cumbersome to work with long lines. 🤷‍♂

components/automate-gateway/handler/iam/v2beta/policy/policy.go
bcmdarroch: Isn't the `unassigned` name a constant somewhere? If not can we make it one?

components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
msorens: The intention is that this is a general purpose mechanism; not designed for any specific endpoint.

components/automate-gateway/handler/iam/v2beta/policy/policy.go
msorens: Thanks for that; good point.

components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
srenatus: Yeah nevermind. I don't know what I've been thinking -- this is the client req interceptor, after all. 😄

msorens:
Closing this PR -- replaced with PR #546

PR #489, opened 2019-06-03
authz-service/policy: only update engine store when needed and only do it once

srenatus:
🐛 be gone -- this removes extra work being done that isn't making any difference.


components/authz-service/server/v2/policy.go
```diff
@@ -704,10 +704,6 @@ func (s *policyServer) EngineUpdateInterceptor() grpc.UnaryServerInterceptor {
 			// do nothing
```
srenatus: ℹ️ the work happens selectively on like 700 above

PR #491, opened 2019-06-04
Default sorting by percent_ok for service-groups table

afiune:
Set the default sorting field to `percent_ok` for the service-group table.

### :+1: Definition of Done
1) By default, sort by Group Health OK percentage number: from smallest to largest. (0-100) Within the same percentage number sort by level of criticality. When the sorting order is toggled, within the same percentage number, still sort by level of criticality. The level of criticality must always be Critical, Unknown and Warning in that order.

2) Switch the primary sorting criteria with the secondary. That is by default, sort by Level of criticality: The order is Critical, Unknown, Warning and Ok. Within the same level of criticality sort by “group health Ok percentage number”. The group health OK percentage number must always be ordered from 0-100.

- Start all services and build the UI. 
- Populate the applications database. 
- Then navigate to the Applications Tab.

https://chefio.atlassian.net/browse/A2-840

- [x] Code actually executed?

### DIFF
--- a/components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
+++ b/components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
@@ -52,13 +52,15 @@ export class ServiceGroupsComponent implements OnInit, OnDestroy {
   // The collection of allowable sort directions
   private allowedSortDirections = ['asc', 'desc', 'ASC', 'DESC'];
 
+  // Sorft field by default
+  readonly defaultSortField = 'percent_ok';
+
   private selectedFieldDirection$: Observable<SortDirection>;
   private selectedSortField$: Observable<string>;
   private healthSummary$: Observable<HealthSummary>;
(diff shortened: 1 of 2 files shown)


components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
```diff
@@ -58,6 +58,7 @@ export class ServiceGroupsComponent implements OnInit, OnDestroy {
   private currentPage$: Observable<number>;
   private currentFieldDirection: SortDirection;
   private currentSortField: string;
+  private defaultSortField = 'percent_ok';
```
srenatus: 💭 We could throw in a `readonly` here, I suppose? ([See this post for details](https://mariusschulz.com/blog/typescript-2-0-read-only-properties))

afiune: Me gusta. 💯

PR #490, opened 2019-06-03
Applications service-group deep sorting

afiune:
### :nut_and_bolt: Description
We are enabling two custom ways of sorting the service-groups:

1) **Sort by Group Health OK percentage number:** from smallest to largest.
(0-100) Within the same percentage number sort by level of criticality.
The level of criticality must always be Critical, Unknown, Warning, OK
in that order.

2) **Sort by Level of Criticality:** The order is Critical, Unknown, Warning
and Ok. Within the same level of criticality sort by “group health Ok
percentage number”. The percentage number must always be ordered
from 0-100

We are also changing the default sort field to be `percent_ok`.
Finally, a quick fix of a flaky test that was failing randomly.

### :+1: Definition of Done
1) By default, sort by Group Health OK percentage number: from smallest to largest. (0-100) Within the same percentage number sort by level of criticality. When the sorting order is toggled, within the same percentage number, still sort by level of criticality. The level of criticality must always be Critical, Unknown and Warning in that order.

2) Switch the primary sorting criteria with the secondary. That is by default, sort by Level of criticality: The order is Critical, Unknown, Warning and Ok. Within the same level of criticality sort by “group health Ok percentage number”. The group health OK percentage number must always be ordered from 0-100.

### :athletic_shoe: Demo Script / Repro Steps
Enter the studio then start, build and populate applications-service.
Verify the new default service-group order.
```
$ hab studio enter
[1][default:/src:0]# start_applications_service
[2][default:/src:0]# applications_populate_database
[3][default:/src:0]# applications_get_service_groups | grep healthPercentage
      "healthPercentage": 20,
      "healthPercentage": 20,
      "healthPercentage": 60,
      "healthPercentage": 80,
      "healthPercentage": 100,
      "healthPercentage": 100,
```
You can just play a bit more with the GRPC functions. 😄

### :chart_with_upwards_trend: Code Benchmark (MASTER Vs PR)
We are introducing some benchmarks for the applications
service at this PR https://github.com/chef/automate/pull/495 and made it official with https://github.com/chef/automate/pull/510

Here is a comparison between what is running currently in
**master versus this PR.**
```
[18][default:/src:0]# benchstat /src/benchmarks/applications-service/txt/PR_510.txt /src/benchmarks/applications-service/txt/PR_490.txt
name                                                   old time/op    new time/op    delta
IngestionNewServices-4                                   9.05ms ±14%    8.60ms ±11%     ~     (p=0.690 n=5+5)
[...]

PR #495, opened 2019-06-04
Benchmarks for Applications Service

afiune:
Currently, in the EAS Team, we are struggling a little bit to make decisions
in the backend, that is inside the `applications-service`.

We don't have a way to know the implications of adding or modifying code
and therefore, as engineers, we make lots of theories and deductions of
what "it could happen" based on assumptions.

An example is the following PR that is adding some extra complexity in
the query side of the micro-service. https://github.com/chef/automate/pull/490

This work is the first iteration to introduce benchmarks in our integration
tests to measure two main GRPC functions that we expose to consumers
(endpoints and the UI). These measurements will be recorded in our pipeline
so that we can understand how the system progress or degrades on code
changes.

### :+1: Definition of Done
After this change is merged, we will start recording times based on
Go Benchmarks for the GRPC functions:

- `GetServiceGroups` - Lists all service-groups in the system.
- `GetServices` - Lists all services in the system.

- Start the studio
- Start the applications service: (you don't have to build the applications-service) 
```
[1][default:/src:0]# start_deployment_service;
[2][default:/src:0]# chef-automate dev deploy-some --with-deps "chef/applications-service"
```
- Run the integration tests. `applications_integration`

Here is the [(link)](https://buildkite.com/chef/chef-automate-master-verify-private/builds/1725#04637108-dcfb-4052-817e-1f8953d9f206) of the BuildKite Job that ran them

Output Gist: https://gist.github.com/afiune/f298f71788294d698e134dd82ad4b39c#file-pr-495-md

N/A

- [x] Code actually executed?

### DIFF
--- a/components/applications-service/integration_test/ingestion_benchmark_test.go
+++ b/components/applications-service/integration_test/ingestion_benchmark_test.go
@@ -1,0 +1,76 @@
+//
+//  Author:: Salim Afiune <afiune@chef.io>
+//  Copyright:: Copyright 2019, Chef Software Inc.
+//
+
+package integration_test
+
+import (
+	"fmt"
+	"math/rand"
+	"testing"
+
+	uuid "github.com/chef/automate/lib/uuid4"
+)
+
+const letterBytes = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+
+func RandString(n int) string {
+	b := make([]byte, n)
+	for i := range b {
+		b[i] = letterBytes[rand.Intn(len(letterBytes))]
+	}
+	return string(b)
+}
+
+func BenchmarkIngestionNewServices(b *testing.B) {
+	var criticality = []string{"ok", "critical", "warning", "unknown"}
+
+	defer suite.DeleteDataFromStorage()
+
+	for n := 0; n < b.N; n++ {
+		b.StopTimer()
+		var (
+			uuid, _ = uuid.NewV4()
+			svcName = RandString(10)
+			health  = criticality[rand.Intn(4)]
+		)
+
+		event := NewHabitatEvent(
+			withSupervisorId(uuid.String()),
+			withPackageIdent(fmt.Sprintf("core/%s/0.1.0/20190101121212", svcName)),
+			withServiceGroup(fmt.Sprintf("%s.default", svcName)),
+			withStrategyAtOnce("stable"),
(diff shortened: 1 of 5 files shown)


components/applications-service/integration_test/ingestion_benchmark_test.go
```diff
@@ -0,0 +1,44 @@
...
+		withStrategyAtOnce("stable"),
+		withApplication("cool-app"),
+		withEnvironment("development"),
+		withFqdn("app.example.com"),
+		withHealth("CRITICAL"),
+		withSite("us"),
+	)
+	suite.IngestService(event)
+
+	b.StartTimer()
```
srenatus: ❓ Does it matter whether `StartTimer` is inside the `b.N` loop or not?

afiune: @srenatus I don't think so, but in reality, I don't know 😅 . I will change it to be inside the loop and see if it changes the result.

components/applications-service/integration_test/service_groups_benchmark_test.go
```diff
@@ -0,0 +1,84 @@
...
+	var (
+		ctx            = context.Background()
+		request        = new(applications.ServiceGroupsReq)
+		iterations     = 10
+		incremental    = 100
+		maxGroups      = incremental * iterations
+		nServiceGroups = 0
+	)
+
+	for i := 0; i <= iterations; i++ {
```
srenatus: AFAIK the golang benchmark suite will do as many iterations as necessary to get a meaningful result -- so what are we measuring here? Since the `b.N` loop is the innermost loop, it seems we're executing _10 runs of golang's benchmark logic_. Why are we doing that?

The alternative I'd see here is benchmarking _10 GetServiceGroups requests_, but then I'd expect the loop with `iterations` to be inside the `b.N` loop 🤔

afiune: I am just doing multiple benchmarks with different loads. Instead of writing 10 benchmark functions, I am just doing one that tests the RPC function with loads of groups from 0 to 50,000. 

I could separate them if that is less confusing.

srenatus: Ah! [this](https://github.com/chef/automate/pull/495/files#diff-8383c9c14d839f5204cba7037be032d8R43), right? I didn't see it 😎 sorry for the noise

PR #497, opened 2019-06-04
[WIP] Use workflow for compliance ES migrations

jaym:
Use the workflow library to run compliance migrations


jaym:
this is not happening right now

PR #501, opened 2019-06-05
[A2-798] fix v2.1 upgrade issue for multi-node settings

srenatus:

Since there's two different stores for these now, it matters in a
multi-node setting which version the server is operating under. Since
the version is derived from the migration status, which in turn is
stored in the database, we make a change to that a "policy change
event": any instance that is connected to the database, but hasn't been
the one processing the migration GRPC call, will be informed that it
needs to update its store.

Figuring out which store to update is done by querying the database for
its migration status; it no longer is retrieved from local state of the
server instance.

Note: this does not notify other nodes in unhappy paths, since a failure
to migrate v1 policies should prohibit a switch to v2 (or v2.1).

### :+1: Definition of Done

Stuff still works single-node. -- This is a _best effort_ fix for getting switches between v2 and v2.1 in multi-node settings work, I don't believe we have any proper testing in place for this (_I don't even know how to spin it up!_)... 😅 

NA

- #467 for splitting the store updates in v2 vs v2.1

- [x] Code actually executed?

### DIFF
--- a/components/authz-service/server/v2/policy_refresher.go
+++ b/components/authz-service/server/v2/policy_refresher.go
@@ -69,10 +69,7 @@ func NewPolicyRefresher(ctx context.Context, log logger.Logger, store storage.St
 }
 
 func (refresher *policyRefresher) run(ctx context.Context) {
-	var (
-		lastVersion  api.Version
-		lastPolicyID string
-	)
+	var lastPolicyChangeID string
 	antiEntropyTimer := time.NewTimer(refresher.antiEntropyTimerDuration)
 RUNLOOP:
 	for {
@@ -83,7 +80,7 @@ RUNLOOP:
 		case <-refresher.changeNotifier.C():
 			refresher.log.Info("Received policy change notification")
 			var err error
-			lastPolicyID, err = refresher.refresh(context.Background(), lastPolicyID)
+			lastPolicyChangeID, err = refresher.refresh(context.Background(), lastPolicyChangeID)
 			if err != nil {
 				refresher.log.WithError(err).Warn("Failed to refresh policies")
 			}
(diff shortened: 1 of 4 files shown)


components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1338,12 +1338,29 @@ func (p *pg) Pristine(ctx context.Context) error {
 	return p.recordMigrationStatus(ctx, enumPristine)
 }
 
+func (p *pg) recordMigrationStatusAndNotifyPG(ctx context.Context, ms string) error {
+	tx, err := p.db.BeginTx(ctx, nil /* use driver default */)
```
msorens: Probably a naive question: don't you need a "defer _some-kind-of-rollback_" on this transaction? I don't see it anywhere else this construct is used, so I'm guessing the answer is no, but why is that?

components/authz-service/server/v2/policy_refresher.go
```diff
@@ -162,10 +153,16 @@ func (refresher *policyRefresher) RefreshAsync() error {
 }
 
 // updates OPA engine store with policy
-func (refresher *policyRefresher) updateEngineStore(ctx context.Context, vsn api.Version) error {
+func (refresher *policyRefresher) updateEngineStore(ctx context.Context) error {
 	// Engine updates need unfiltered access to all data.
 	ctx = auth_context.ContextWithoutProjects(ctx)
 
+	vsn, err := refresher.getIAMVersion(ctx)
```
msorens: Not following all the intricacies of this PR, but I think this line is the crux: it retrieves the version from the DB instead of the local server state so that in a multi-node situation it will be updating the correct store. Is that right? If so, please add an in-code comment to that effect; thx!

components/authz-service/storage/v2/postgres/postgres.go
srenatus: It's all in the `ctx`, see this note: https://github.com/chef/automate/blob/02fcff1a6660ba1d3f1fe847a5d9785c744afa79/components/authz-service/storage/v2/postgres/postgres.go#L58-L63

Now, whether not cancelling this myself is a problem or not (I haven't put a `ctx, cancel := context.WithCancel(ctx); defer cancel()` in this function), I'll figure out now.

srenatus: Update: We don't need to cancel ourselves, grpc takes care of this -- see this toy experiment https://gist.github.com/srenatus/ce12b31ea517f16c024e4f8736fa5f2b

srenatus: https://github.com/chef/automate/pull/501/commits/329e8ef6fbef7337c3d5161ebd3b9a00cca23dbd Added a cleanup accordingly.

srenatus:
will merge when green

srenatus:
### ⚠️ Unit tests for `authz-service` keep timing out because they don't finish within 10 minutes. This could be a symptom of some issue. TODO

srenatus:
Hrm maybe my experiment re: `context.WithCancel` was misleading 🤔

srenatus:
will merge when green

components/authz-service/storage/v2/postgres/postgres.go
srenatus: And reverted the cleanup. Experiment must have been flawed, we _need_ the `context.WithCancel`.

Issue #432, opened 2019-05-29
Display version and build in UI, CLI, Release Notes, and Downloads page


vjeffrey:
## User Story

This is a requirement for some customers. We publish the build number but in order to meet the upgrade requirements set by certain companies, we need to also call out the major version of Automate.

### UI
* Add major version number in the user menu
* Label the build number in the user menu

<img width="214" alt="add-version-label-build" src="https://user-images.githubusercontent.com/5489125/58572717-0efbdf80-81f1-11e9-9208-6e51b69ec6e4.png">

### CLI
* When checking the version with `chef-automate version` display the major version, CLI build number, and Server build number 

```
Version: 2
CLI Build: 20190528210257
Server Build: 20190528210257
```

### Release Notes Page
* Change left-nav to say _Build_ instead of _Version_
* Change the h1's on the pages to say `Version: 2 Build: <build number>`

<img width="1029" alt="docs-changes" src="https://user-images.githubusercontent.com/5489125/58573454-9564f100-81f2-11e9-8109-e517aabab0c3.png">

### Downloads Page
* Update the paragraph text describing automate to list the major version number

<img width="483" alt="downloads-page" src="https://user-images.githubusercontent.com/5489125/58573675-17551a00-81f3-11e9-9a83-98b51c9357a5.png">

## Related Resources
https://chef-internal.ideas.aha.io/ideas/A2-I-44

btm:
> Downloads Page
> * Update the paragraph text describing automate to list the major version number

I can make this change. However, it used to say Automate 2, but when we made the last round of changes for the licensing changes we tried to make it more "the one and only automate you should use, except if... " instead of there being v1 and v2. So we should discuss if there's a better way to do that and not have the [marketing] product name be "Chef Automate 2".

kagarmoe:
Automate 1 EOLs in December, doesn't it? We may have a problem with "Version: 2" because it isn't how we handle versioning in the docs. (As @btm mentioned, there is only one Automate. It is a marketing & metaphysical position) Even if we settle on versioning, we may not be able to get the build ID on the same line as the version number. OK to stack them?

kagarmoe:
Also, you use "provide" 2x in your paragraph.

susanev:
@kagarmoe that's the current copy.. all I added was the _Version 2_

btm:
@kagarmoe yes, we're targeting an EOL of A1 for 12/31.

[...]

PR #502, opened 2019-06-05
modify automate cli and ui user menu to include major automate version

vjeffrey:
As noted in https://github.com/chef/automate/issues/432, we have made the decision to include the "major" Automate version (2) where the build number is listed for the Automate product.

This change makes those modifications for the user menu in the Automate ui and the automate cli.
automate-cli version
Screen Shot 2019-06-05 at 07 40 54

### :+1: Definition of Done
Version: 2 is displayed above build number for the user menu and cli

rebuild ui & cli
chef-automate version
check user menu in ui

### DIFF
--- a/components/automate-cli/cmd/chef-automate/version.go
+++ b/components/automate-cli/cmd/chef-automate/version.go
@@ -34,6 +34,7 @@ type versionResult struct {
 var verbose bool
 
 func runVersionCmd(cmd *cobra.Command, args []string) error {
+	writer.Printf("Version: %s\n", "2")
 	printClientVersion()
 	return printServerVersion()
 }
@@ -43,11 +44,10 @@ func printClientVersion() {
 	if verbose {
 		writer.Println("CLI")
 		writer.Println("===")
-		writer.Printf("Version: %s\n", version.BuildTime)
-		writer.Printf("Build time: %s\n", version.BuildTime)
+		writer.Printf("CLI Build: %s\n", version.BuildTime)
 		writer.Printf("Git SHA: %s\n", version.GitSHA)
 	} else {
-		writer.Printf("CLI version: %s\n", version.BuildTime)
+		writer.Printf("CLI Build: %s\n", version.BuildTime)
 	}
 }
 
@@ -81,8 +81,7 @@ func printServerVersion() error {
 		writer.Println("======")
 		// TODO: (yzl) When we know what we want to display for Version, update it to be something other
 		// than the timestamp.
-		writer.Printf("Version: %s\n", response.BuildTimestamp)
-		writer.Printf("Build time: %s\n", response.BuildTimestamp)
+		writer.Printf("Server Build: %s\n", response.BuildTimestamp)
 		writer.Printf("Git SHA: %s\n", response.BuildSha)
 		err := printUpgradeStatus()
 		if err != nil {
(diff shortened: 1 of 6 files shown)


components/automate-ui/src/app/page-components/profile/profile.component.html
```diff
@@ -27,8 +27,9 @@
         </li>
       </app-authorized>
       <li class="dropdown-list-item">
-        <span class="version">Version: {{ buildVersion }}</span>
-        <ul class="dropdown-sub-list">
+        <span class="version">Version: 2</span>
+        <span class="build">Build: {{ buildVersion }}</span>
+		<ul class="dropdown-sub-list">
```
srenatus: Indentation looks a bit wonky, doesn't it?

vjeffrey: it does. my editor does not get along with ui code. or, well, i guess i should say i haven't taken the time to set up my editor so that it does the right thing when it auto-detects the language in the file

components/automate-ui/e2e/app.e2e-spec.ts
```diff
@@ -142,7 +143,9 @@ describe('User Dropdown', () => {
     element(by.css('app-profile')).click();
 
     const versionLink = $('.version');
-    expect(versionLink.getText()).toBe(`Version: ${expectedVersion}`);
+    expect(versionLink.getText()).toBe(`Version: 2`);
+    const buildLink = $('.build');
+    expect(versionLink.getText()).toBe(`Build: ${expectedVersion}`);
```
phiggins: Should this be testing `expectedBuild` instead?

components/automate-cli/cmd/chef-automate/version.go
```diff
@@ -43,11 +44,10 @@ func printClientVersion() {
 	if verbose {
 		writer.Println("CLI")
 		writer.Println("===")
-		writer.Printf("Version: %s\n", version.BuildTime)
-		writer.Printf("Build time: %s\n", version.BuildTime)
+		writer.Printf("CLI Build: %s\n", version.BuildTime)
 		writer.Printf("Git SHA: %s\n", version.GitSHA)
 	} else {
-		writer.Printf("CLI version: %s\n", version.BuildTime)
+		writer.Printf("CLI Build: %s\n", version.BuildTime)
```
phiggins: Super nitpicky, but `Println` can be used when you want the thing to appear at the end of the line, like `writer.Println("CLI Build:", version.BuildTime)`. 

https://play.golang.org/p/XxFm4WS-mNg

stevendanna: Unfortunately, it looks like the api of that writer object doesn't exactly matching the api of fmt and it only accepts a single string.  We should probably make them match to avoid confusion.

components/automate-ui/src/app/page-components/profile/profile.component.scss
```diff
@@ -98,3 +98,8 @@ a, button:not(.dropdown-toggle) {
   right: 16px;
   top: 8px;
 }
+
+.version:after {
+  content: '\A';
+  white-space: pre;
+}
```
phiggins: I'm no CSS expert but it seems kinda non-obvious what's going on here. I had to look up what `content: '\A'` does to see why that's here. :)

This might be less prone to breaking in the future with a separately styled `<ul>` instead.

Issue #456, opened 2019-05-30
`EOF` bug on compliance/profiles


susanev:
## Overview
in acceptance (not current)
navigating directly to `compliance/profiles`
or refreshing that page once you're there results in a `EOF`

## What it looks like when its done
- change route to `compliance/compliance-profiles`
- this will need to be updated throughout the docs as well
    - https://automate.chef.io/docs/asset-store/#using-the-asset-store
    - https://automate.chef.io/docs/asset-store/#about-the-profile-identifier
    - https://automate.chef.io/docs/asset-store/#api-calls

srenatus:
We've got a clash here:

a) automate-load-balancer config:
https://github.com/chef/automate/blob/master/components/automate-load-balancer/habitat/config/nginx.conf#L199-L204
b) new compliance ui route: https://github.com/chef/automate/pull/291/files#diff-fc147971f13617ccf3d96c5f26ef3f2aR227

If you navigate, no request is really sent (UI routes are a bit of a scam), **but** if you refresh, or browse directly, the browser sends a request to that path - and the load balancer will forward the request to the API endpoint. The request sent isn't good, so you'll get a bad response, which in this case is EOF.

I _think_ we cannot change (a) as too many clients (inspec, audit cookbook) depend on this exact API route (which is why the redirect exists in the first place, i.e., it's not a route below `/api/v0/...` etc). However, we could change (b) easily, I suppose

srenatus:
Looking at the logs while accessing `https://a2-dev.test/compliance/profiles` _directly_, I only find
```
+ echo '[2019-05-31T12:33:52+00:00]' 'automate-load-balancer.default(O): - [31/May/2019:12:33:52 +0000]  "GET /compliance/profiles/ HTTP/1.1" 400 "0.001" 4 "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/74.0.3729.169 Safari/537.36" "10.0.2.15:2000" "400" "0.000" 738'
```
which is a bit odd -- I'd have expected _some_ domain-service response there.

[...]

PR #505, opened 2019-06-05
UI-456 EOF compliance/profiles bug Signed-Off-By: Tara Black <tblack@…

tarablack01:
…chef.io>

in acceptance (not current)
navigating directly to compliance/profiles
or refreshing that page once you're there results in a EOF

### :+1: Definition of Done
change route to compliance/compliance-profiles
this will need to be updated throughout the docs as well
https://automate.chef.io/docs/asset-store/#using-the-asset-store
https://automate.chef.io/docs/asset-store/#about-the-profile-identifier
https://automate.chef.io/docs/asset-store/#api-calls

https://github.com/chef/automate/issues/456

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/+profile/profile.routing.ts
+++ b/components/automate-ui/src/app/pages/+compliance/+profile/profile.routing.ts
@@ -6,16 +6,11 @@ import { ProfileOverviewComponent } from './+profile-overview/profile-overview.c
 const routes: Routes = [
   {
     path: '',
-    children: [
-        {
-        path: '',
-        component: ProfileOverviewComponent
-        },
-        {
-        path: 'profile-details',
-        component: ProfileDetailsComponent
-        }
-    ]
+    component: ProfileOverviewComponent
+  },
+  {
+    path: 'profile-details',
+    component: ProfileDetailsComponent
   }
 ];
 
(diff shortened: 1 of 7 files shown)


components/automate-ui/src/app/pages/+compliance/+profile/profile.routing.ts
```diff
@@ -6,16 +6,11 @@ import { ProfileOverviewComponent } from './+profile-overview/profile-overview.c
...
-        {
-        path: 'profile-details',
-        component: ProfileDetailsComponent
-        }
-    ]
+    component: ProfileOverviewComponent
+  },
+  {
+    path: 'profile-details',
+    component: ProfileDetailsComponent
```
srenatus: This is just a simplification, I suppose? Or has it any consequence?

tarablack01:
The UI route: compliance/profiles was changed to compliance/compliance-profiles because we have a conflicting location route compliance/profiles defined in automate-load-balancer/habitat.

Docs didn't need updating, /profiles is still used to redirect to /compliance/compliance-profiles.

components/automate-ui/src/app/pages/+compliance/+profile/profile.routing.ts
tarablack01: this url is compliance/compliance-profiles/profile-details...
The /profile-details is the same and hasn't been changed and the /compliance-profiles has been updated from /profiles.

Issue #266, opened 2019-05-07
Projects list page updates


susanev:
[Jira subtask](https://chefio.atlassian.net/browse/A2-824)
[Jira parent task](https://chefio.atlassian.net/browse/A2-318) with video and more context of where we are headed.

PR #511, opened 2019-06-05
Projects list updates - update projects

scottopherson:

Adds updates to the settings/projects-list view to support updating project rules.

https://github.com/chef/automate/issues/266

### DIFF
--- a/components/automate-ui/src/app/entities/projects/project.actions.ts
+++ b/components/automate-ui/src/app/entities/projects/project.actions.ts
@@ -34,6 +34,14 @@ export interface ProjectSuccessPayload {
   project: Project;
 }
 
+export interface GetApplyRulesStatusSuccessPayload {
+  state: string;
+  estimated_time_complete: string;
+  percentage_complete: number;
+  failed: boolean;
+  failure_message: string;
+}
+
 export class GetProjects implements Action {
   readonly type = ProjectActionTypes.GET_ALL;
 }
@@ -157,7 +165,7 @@ export class GetApplyRulesStatus implements Action {
 
 export class GetApplyRulesStatusSuccess implements Action {
   readonly type = ProjectActionTypes.GET_APPLY_RULES_STATUS_SUCCESS;
-  constructor(public payload: any) { }
+  constructor(public payload: GetApplyRulesStatusSuccessPayload) { }
 }
 
 export class GetApplyRulesStatusFailure implements Action {
--- a/components/automate-ui/src/app/entities/projects/project.actions.ts
+++ b/components/automate-ui/src/app/entities/projects/project.actions.ts
@@ -4,21 +4,30 @@ import { Action } from '@ngrx/store';
 import { Project } from './project.model';
 
 export enum ProjectActionTypes {
-  GET_ALL         = 'PROJECT::GET_ALL',
-  GET_ALL_SUCCESS = 'PROJECT::GET_ALL::SUCCESS',
-  GET_ALL_FAILURE = 'PROJECT::GET_ALL::FAILURE',
-  GET             = 'PROJECT::GET',
-  GET_SUCCESS     = 'PROJECT::GET::SUCCESS',
-  GET_FAILURE     = 'PROJECT::GET::FAILURE',
-  CREATE          = 'PROJECT::CREATE',
-  CREATE_SUCCESS  = 'PROJECT::CREATE::SUCCESS',
-  CREATE_FAILURE  = 'PROJECT::CREATE::FAILURE',
-  DELETE          = 'PROJECT::DELETE',
-  DELETE_SUCCESS  = 'PROJECT::DELETE::SUCCESS',
-  DELETE_FAILURE  = 'PROJECT::DELETE::FAILURE',
-  UPDATE          = 'PROJECT::UPDATE',
-  UPDATE_SUCCESS  = 'PROJECT::UPDATE::SUCCESS',
-  UPDATE_FAILURE  = 'PROJECT::UPDATE::FAILURE'
+  GET_ALL                        = 'PROJECT::GET_ALL',
+  GET_ALL_SUCCESS                = 'PROJECT::GET_ALL::SUCCESS',
+  GET_ALL_FAILURE                = 'PROJECT::GET_ALL::FAILURE',
+  GET                            = 'PROJECT::GET',
+  GET_SUCCESS                    = 'PROJECT::GET::SUCCESS',
+  GET_FAILURE                    = 'PROJECT::GET::FAILURE',
+  CREATE                         = 'PROJECT::CREATE',
+  CREATE_SUCCESS                 = 'PROJECT::CREATE::SUCCESS',
+  CREATE_FAILURE                 = 'PROJECT::CREATE::FAILURE',
+  DELETE                         = 'PROJECT::DELETE',
+  DELETE_SUCCESS                 = 'PROJECT::DELETE::SUCCESS',
+  DELETE_FAILURE                 = 'PROJECT::DELETE::FAILURE',
+  UPDATE                         = 'PROJECT::UPDATE',
+  UPDATE_SUCCESS                 = 'PROJECT::UPDATE::SUCCESS',
+  UPDATE_FAILURE                 = 'PROJECT::UPDATE::FAILURE',
+  APPLY_RULES_START              = 'PROJECT::APPLY_RULES_START',
+  APPLY_RULES_START_SUCCESS      = 'PROJECT::APPLY_RULES_START::SUCCESS',
+  APPLY_RULES_START_FAILURE      = 'PROJECT::APPLY_RULES_START::FAILURE',
+  APPLY_RULES_STOP               = 'PROJECT::APPLY_RULES_STOP',
+  APPLY_RULES_STOP_SUCCESS       = 'PROJECT::APPLY_RULES_STOP::SUCCESS',
+  APPLY_RULES_STOP_FAILURE       = 'PROJECT::APPLY_RULES_STOP::FAILURE',
+  GET_APPLY_RULES_STATUS         = 'PROJECT::GET_APPLY_RULES_STATUS',
+  GET_APPLY_RULES_STATUS_SUCCESS = 'PROJECT::GET_APPLY_RULES_STATUS::SUCCESS',
+  GET_APPLY_RULES_STATUS_FAILURE = 'PROJECT::GET_APPLY_RULES_STATUS::FAILURE'
 }
 
 export interface ProjectSuccessPayload {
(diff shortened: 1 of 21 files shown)


susanev:
thanks for working on this @scottopherson!!
looks great, just noticed one thing...
- space between the buttons should be 16px, currently, it looks to be 22px

susanev:
@scottopherson looks great!
i did notice a few things, that it would be great if we could fix in this pr.
- the control-menus on the project list page are loading a few moments after everything else, can we fix that so it works like the other pages in settings?
- focus is not being placed on the modals when they open, can we start the focus on the close-x button?

components/automate-ui/src/app/entities/projects/project.reducer.ts
```diff
@@ -21,16 +32,26 @@ const CREATE_STATUS = 'createStatus';
...
 const DELETE_STATUS = 'deleteStatus';
 const UPDATE_STATUS = 'updateStatus';
+const APPLY_RULES_STATUS = 'applyRulesStatus';
 
 export const projectEntityAdapter: EntityAdapter<Project> = createEntityAdapter<Project>();
 
+const initialApplyRulesStatus: ApplyRulesStatus = {
+  state: 'not_running',
+  estimatedTimeComplete: '0001-01-01T00:00:00Z',
+  percentageComplete: 61,
```
bcmdarroch: why 61?

components/automate-ui/src/app/entities/projects/project.effects.ts
```diff
@@ -151,5 +162,41 @@ export class ProjectEffects {
...
         });
       }));
+
+  @Effect()
+  applyRulesStart$ = this.actions$.pipe(
+    ofType<ApplyRulesStart>(ProjectActionTypes.APPLY_RULES_START),
+    mergeMap(() =>
+      this.requests.applyRulesStart().pipe(
+        switchMap(() => [
+          new ApplyRulesStartSuccess(),
```
msorens: This is not currently hooked up to do anything. If that was deliberate, please at least add a comment here to the effect "needs to be wired up". Same goes for `ApplyRulesStartSuccess` below.

components/automate-ui/src/app/entities/projects/project.effects.ts
```diff
@@ -151,5 +162,41 @@ export class ProjectEffects {
           message: `Could not update project: ${msg || payload.error}`
         });
       }));
+
+  @Effect()
+  applyRulesStart$ = this.actions$.pipe(
+    ofType<ApplyRulesStart>(ProjectActionTypes.APPLY_RULES_START),
+    mergeMap(() =>
+      this.requests.applyRulesStart().pipe(
+        switchMap(() => [
```
msorens: Help me understand what `switchMap` is doing here. (I am familiar with the operator; talking about this specific implementation. 😁 ) I am guessing that since its anonymous function is returning a list of two Observables it emits on the first (ApplyRulesStartSuccess) until the second (GetApplyRulesStatus) comes back...? If so, why is that useful to do here.

components/automate-ui/src/app/pages/project/confirm-apply-stop-modal/confirm-apply-stop-modal.component.ts
```diff
@@ -0,0 +1,38 @@
...
+  progressSuffixText: string;
+
+  ngOnChanges() {
+    const { percentageComplete, estimatedTimeComplete } = this.applyRulesStatus;
+    const now = moment();
+    const etc = moment(estimatedTimeComplete);
+    const dur = moment.duration(etc.diff(now));
+    const durHours = Math.floor(dur.get('hours')).toString().padStart(2, '0');
+    const durMins = dur.get('minutes').toString().padStart(2, '0');
+    const durSecs = dur.get('seconds').toString().padStart(2, '0');
```
msorens: For the above three `get` calls, suggest methods instead of string arguments, i.e. `dur.hours()`, `dur.minutes()`, and `dur.seconds()`. Is there a reason not to do that?

components/automate-ui/src/app/pages/project/confirm-apply-stop-modal/confirm-apply-stop-modal.component.ts
```diff
@@ -0,0 +1,38 @@
...
+  progressPrefixText: string;
+
+  progressSuffixText: string;
+
+  ngOnChanges() {
+    const { percentageComplete, estimatedTimeComplete } = this.applyRulesStatus;
+    const now = moment();
+    const etc = moment(estimatedTimeComplete);
+    const dur = moment.duration(etc.diff(now));
+    const durHours = Math.floor(dur.get('hours')).toString().padStart(2, '0');
```
msorens: Don't see why you need `Math.floor` here...? (And why for hours but not for minutes or seconds.)

components/automate-ui/src/app/pages/project/confirm-apply-stop-modal/confirm-apply-stop-modal.component.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  @Output() cancel: EventEmitter<void> = new EventEmitter();
+
+  progressValue: number;
+
+  progressPrefixText: string;
+
+  progressSuffixText: string;
+
+  ngOnChanges() {
```
msorens: Seems like there is enough going on here to merit unit testing these calculations, if it is not too cumbersome to set up.

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -53,6 +59,17 @@ export class ProjectListComponent implements OnInit {
     this.iamMajorVersion$ = store.select(iamMajorVersion);
     this.iamMinorVersion$ = store.select(iamMinorVersion);
 
+    this.applyRulesButtonText$ = this.projects.applyRulesStatus$.pipe(
+      map(({ state, percentageComplete }) => {
+        switch (state) {
+          case 'not_running':
+            return 'Update Projects';
+          case 'running':
```
msorens: Could you grab the named constants for `not_running` and `running` from lib/authz/project_update_manager.go ?
Similar note for project.reducer.ts.

components/automate-ui/src/app/entities/projects/project.reducer.ts
```diff
@@ -96,6 +117,9 @@ export function projectEntityReducer(
     case ProjectActionTypes.UPDATE_FAILURE:
       return set(UPDATE_STATUS, EntityStatus.loadingFailure, state);
 
+    case ProjectActionTypes.GET_APPLY_RULES_STATUS_SUCCESS:
+      return set(APPLY_RULES_STATUS, mapKeys(key => camelCase(key), action.payload), state);
```
msorens: What are you camel-casing here? Seems like the only field would be the state, and that uses snake case, no?

components/automate-ui/src/app/entities/projects/project.actions.ts
```diff
@@ -114,6 +123,48 @@ export class UpdateProjectFailure implements Action {
...
+  constructor(public payload: HttpErrorResponse) { }
+}
+
+export class GetApplyRulesStatus implements Action {
+  readonly type = ProjectActionTypes.GET_APPLY_RULES_STATUS;
+}
+
+export class GetApplyRulesStatusSuccess implements Action {
+  readonly type = ProjectActionTypes.GET_APPLY_RULES_STATUS_SUCCESS;
+  constructor(public payload: any) { }
```
msorens: Could you make this a specific type instead of `any` ?

components/automate-ui/src/app/entities/projects/project.requests.ts
```diff
@@ -41,4 +41,16 @@ export class ProjectRequests {
...
+    return this.http.post<any>(`${env.auth_v2_url}/apply-rules`, '{}');
+  }
+
+  public applyRulesStop(): Observable<any> {
+    return this.http.delete<any>(`${env.auth_v2_url}/apply-rules`);
+  }
+
+  public getApplyRulesStatus(): Observable<any> {
+    return this.http.get<any>(`${env.auth_v2_url}/apply-rules`);
+  }
```
msorens: Could these use specific types instead of `any` ?

components/automate-ui/src/app/entities/projects/project.reducer.ts
scottopherson: Ha oops, just had that there while I was testing some stuff out w/ the progress bar. I'll change it back to 0.

components/automate-ui/src/app/entities/projects/project.reducer.ts
scottopherson: The GET response payload (and all other API responses) returns as snakecase:
```
{
  "state": "not_running",
  "estimated_time_complete": "0001-01-01T00:00:00Z",
  "percentage_complete": 1,
  "failed": false,
  "failure_message": ""
}
```

However I have [the interface](https://github.com/chef/automate/blob/0c43a66cc0f46bd36c40fab549a08e7e5d352f21/components/automate-ui/src/app/entities/projects/project.reducer.ts#L11-L17) set up as camelcase (since its JS and thats what we use everywhere):
```
interface ApplyRulesStatus {
  state: ApplyRulesStatusState;
  estimatedTimeComplete: string;
  percentageComplete: number;
  failed: boolean;
  failureMessage: string;
}
```

As for camel-casing here in the reducer specifically, I could've also done it in the `project.requests` method that gets the response or in the effect that handles the response. I just went w/ doing it in the reducer; I'd be fine doing it elsewhere (or not at all and instead change the interface to snakecase).

I wish we had chosen camelcase for our JSON APIs 😐

components/automate-ui/src/app/pages/project/confirm-apply-stop-modal/confirm-apply-stop-modal.component.ts
scottopherson: @msorens ah oops, I actually forgot to change this back to using `dur.as('hours')` for hours. `.get` doesn't return what I need for hours.

So the countdown string is setup as `hours:mins:secs` remaining. For mins and secs remaining, we use `.get` method:
```js
dur.get('seconds')  // <= returns num between 0-59
dur.get('minutes')  // <= returns num between 0-59
dur.get('hours')    // <= returns num between 0-23

// 16:51:17 remaining
// 16 hours, 51 mins, 17 secs remaining
```
^ using `.get` method for hours is fine as well if duration is less than 24 hours but if we want the hours to display beyond 24 hours then we want to use `.as` method for hours and round off the remainder (to be displayed by mins and secs):

```js
dur.get('seconds')  // <= returns num between 0-59
dur.get('minutes')  // <= returns num between 0-59
dur.as('hours')     // <= returns length of duration in hours as float

dur.as('hours') // => 36.854893...

// 36:51:17 remaining
// 36 hours, 51 mins, 17 secs remaining
```
Screen Shot 2019-07-08 at 3 49 09 PM

Perhaps this is all moot since time remaining for updating projects supposedly should never go above 12 hours 🤔 

Either way, I should add unit tests for this as noted in your other comment. Will add some shortly 👍 

As for the methods used, I have no preference for
`.get('hours') / .as('hours')` vs `.hours() / .asHours()`.
I'll use what y'all prefer 👍

components/automate-ui/src/app/entities/projects/project.effects.ts
scottopherson: @msorens I'm using the switchMap to dispatch `ApplyRulesStartSuccess` and `GetApplyRulesStatus` actions simultaneously.

`GetApplyRulesStatus` is just so that we immediately fetch the latest update status after successfully starting an update. This isn't necessary, we could instead just wait for the ongoing 5 sec interval to eventually update the status. 

`ApplyRulesStartSuccess` currently isn't used by anything, I just included the corresponding `Success` and `Error` actions to be consistent with other async actions.

scottopherson:
@susanev ya I'll add focus for the modal. But on other modals that we focus, we focus on the modal container itself first, not the close button (or other content) within it. The `create project` modal on this page focuses on the container for example. Do we want to differ here? The initial focus target is not currently configurable in `chef-modal` iirc, we'd need to modify the component to allow that.

susanev:
@scottopherson oh sorry, the container itself is just fine, whatever the other modals use is good with me!

components/automate-ui/src/app/pages/project/confirm-apply-stop-modal/confirm-apply-stop-modal.component.ts
msorens: Thanks for the detailed note!

components/automate-ui/src/app/entities/projects/project.reducer.ts
msorens: Ah, I was thinking about the data, not the field names. 👍

components/automate-ui/src/app/pages/project/confirm-apply-stop-modal/confirm-apply-stop-modal.component.ts
```diff
@@ -0,0 +1,45 @@
...
+    const etc = moment(estimatedTimeComplete);
+    const dur = moment.duration(etc.diff(now));
+    const pad = num => num.toString().padStart(2, '0');
+    const durHours = pad(Math.floor(dur.asHours()));
+    const durMins = pad(dur.minutes());
+    const durSecs = pad(dur.seconds());
+    const durCountdown = etc.diff(now) < 0 ? '00:00:00' : `${durHours}:${durMins}:${durSecs}`;
+
+    this.progressValue = percentageComplete;
+    this.progressPrefixText = `${Math.floor(percentageComplete * 100)}% complete`;
```
msorens: Hmm... Curious why you changed from round to floor here...? Seems like round would be accurate.

susanev: @msorens currently 99.5 will display as 99 instead of 100 with this method, so let's leave it as is, for now, we can revisit in the future if needed.

scottopherson: I don't feel strongly either way; I'd happily follow either choice. It just seemed more appropriate to me to display values like `99.6` as `99% complete` as opposed to `100% complete`.

PR #514, opened 2019-06-05
Explain some tricky bits of nats custom auth

danielsdeleo:

Our NATS authentication and authorization behavior is based on feeding callbacks to NATS and this is something you have to read the NATS code to understand (docs mention the capability exists but don't explain how to use it). I've added some comments to explain a bit about how the interactions with NATS are expected to work and why they are written the way they are.

- [ ] Code actually executed?

### DIFF
--- a/components/event-gateway/pkg/nats/authenticator.go
+++ b/components/event-gateway/pkg/nats/authenticator.go
@@ -104,6 +104,18 @@ func newAutomateAuthenticator(c *config.EventGatewayConfig) (*automateAuthentica
 	}, nil
 }
 
+// Check is a callback function that NATS will call to authenticate connection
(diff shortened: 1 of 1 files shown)


components/event-gateway/pkg/nats/authenticator.go
```diff
@@ -116,7 +128,7 @@ func (a *automateAuthenticator) Check(client natsd.ClientAuthentication) bool {
 	}
 
 	err := a.checkToken(token)
-	if err == nil {
+	if err == nil { // success
```
afiune: I would change the logic in this `if` instead of adding a comment. 😅 (just my humble opinion)

PR #517, opened 2019-06-06
[A2-809] back up ListProjectRules API with real data

bcmdarroch:
The ListProjectRules API is one of our POC APIs we built with the superteam during our initial spike to tag ingested resources with projects. This API returns a map of rules keyed by the project they belong to, which is then used by the ingest/compliance service to tag ingested resources with the appropriate project based on whatever rules apply to that project.

Now, this API is backed by a simple cache in OPA. Before, the intention was to store the rules in a proper OPA store and fetch them with a special `rule_mappings` rego function. However, since we weren't using any of rego's capabilities to manipulate the returned rules and the cost of re-converting the stored rules and their nested conditions into their correct storage types was high, I decided to switch to storing them in a basic go cache (still a part of OPA's state, so we still get all opa refresh logic needed for multi-node)).

We will need to rerun benchmarks to ensure this change doesn't impact performance.

### :+1: Definition of Done
- [x] rules are loaded into a cache on authz server init
- [x] the rule cache gets updated anytime a rule gets created, deleted, or updated
- [x] load rules into test setup, get tests passing
- [x] delete GetProjectRules POC APIs, project mappings rego, old engine rule structs
- [x] rename API so it's not so similar to ListRulesForProject

```
1. create projects
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects -d "$(jq -n '{ id: "foo-project", name: "my foo project" }')"
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects -d "$(jq -n '{ id: "foo-project-2", name: "my foo project" }')"

2. create rules
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/rules -d "$(jq -n '{ id: "foo-rule", name: "my foo rule", type: "NODE", project_id: "foo-project", conditions: [{ type: "CHEF_SERVERS", operator: "MEMBER_OF", values: ["prod", "staging"]}]}')" | jq .

curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/rules -d "$(jq -n '{ id: "foo-rule-b", name: "my foo rule", type: "EVENT", project_id: "foo-project-2", conditions: [{ type: "CHEF_ORGS", operator: "EQUALS", values: ["staging"]}]}')" | jq .

curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/rules -d "$(jq -n '{ id: "foo-rule-z", name: "my foo rule", type: "EVENT", project_id: "foo-project-2", conditions: [{ type: "CHEF_ORGS", operator: "EQUALS", values: ["staging"]}]}')" | jq .

chef-automate dev grpcurl authz-service chef.automate.domain.authz.v2.Projects.ListRulesForAllProjects
{
  "projectRules": {
    "foo-project": {
      "rules": [
        {
          "id": "foo-rule",
          "projectId": "foo-project",
          "name": "my foo rule",
          "conditions": [
            {
              "values": [
                "prod",
                "staging"
              ]
            }
          ]
        }
      ]
    },
    "foo-project-2": {
      "rules": [
        {
          "id": "foo-rule-z",
          "projectId": "foo-project-2",
          "name": "my foo rule",
          "type": "EVENT",
          "conditions": [
            {
              "type": "CHEF_ORGS",
              "values": [
                "staging"
              ],
              "operator": "EQUALS"
            }
          ]
        },
        {
          "id": "foo-rule-b",
          "projectId": "foo-project-2",
          "name": "my foo rule",
          "type": "EVENT",
          "conditions": [
            {
              "type": "CHEF_ORGS",
              "values": [
                "staging"
              ],
              "operator": "EQUALS"
            }
          ]
        }
      ]
    }
  }
}
```

Note: in the grpcurl response, some of the rule types, condition attributes, or condition operators are suppressed because they're the `0` value of those types. We could fix this, but I've confirmed the values are correct in logging, and since this API is for internal use only, it should be fine leaving it as is.

- [ ] Code actually executed?


components/authz-service/server/v2/policy_refresher.go
```diff
@@ -183,7 +183,7 @@ func (refresher *policyRefresher) updateEngineStore(ctx context.Context, vsn api
 	case vsn.Minor == api.Version_V1: // v2.1
 		return refresher.engine.V2p1SetPolicies(ctx, policyMap, roleMap, ruleMap)
 	default: // v2.0 OR v1.0
-		return refresher.engine.V2SetPolicies(ctx, policyMap, roleMap, ruleMap)
+		return refresher.engine.V2SetPolicies(ctx, policyMap, roleMap)
```
srenatus: 👏

components/authz-service/engine/opa/opa.go
```diff
@@ -425,12 +429,24 @@ func (s *State) RulesForProject(
...
+func (s *State) ListProjectMappings(ctx context.Context) (map[string][]v2.Rule, error) {
+	items := s.ruleStore.Items()
+	s.log.Infof("HEY! 2. got this %#v in items", items)
+	// nothing stored while on v1 or v2.0
+	if len(items) == 0 {
+		return map[string][]v2.Rule{}, nil
+	}
+	projectRules := make(map[string][]v2.Rule, len(items))
+	for project, item := range items {
+		rules, ok := item.Object.([]v2.Rule)
```
srenatus: you can `fmt.Printf("object type is %T\n", item)` to check the type -- it might not be `[]v2.Rule`?

bcmdarroch: ya i've tested with logging and it's coming back as the correct type.

api/interservice/authz/v2/project.proto
```diff
@@ -76,6 +74,7 @@ service Projects {
     rpc ListRules(ListRulesReq) returns (ListRulesResp) {};
     rpc ListRulesForProject(ListRulesForProjectReq) returns (ListRulesForProjectResp) {};
     rpc DeleteRule(DeleteRuleReq) returns (DeleteRuleResp) {};
+    rpc ListProjectRules(ListProjectRulesReq) returns (ProjectCollectionRulesResp) {};
```
blakestier: Not your change (I see it's from the old proto) but would be good to have the req/resp in alignment

bcmdarroch: yeah totally i'm gonna rename the whole API 👍

components/authz-service/server/v2/system_integration_test.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	"github.com/chef/automate/lib/logger"
 
+	api "github.com/chef/automate/api/interservice/authz/v2"
```
blakestier: this is identical to the import below it

components/authz-service/server/v2/system_integration_test.go
```diff
@@ -160,7 +160,7 @@ func TestListProjectRules(t *testing.T) {
 
 		expectedRules1 := []*api.ProjectRule{createResp1.Rule, createResp2.Rule}
 		expectedRules2 := []*api.ProjectRule{createResp3.Rule}
-
+
```
blakestier: is this spaces? looks funny

api/interservice/authz/v2/project.proto
blakestier: any movement on the value of "ListRulesForProject" and whether it could just move into the "GetProject" call? (If it's indeed only used on the detail project view of the UI).

components/authz-service/server/v2/system_integration_test.go
srenatus: It's superfluous at any rate ;) `gofmt -w` should remove these. 😃

components/authz-service/engine/opa/policy.bindata.go
```diff
@@ -1,18 +1,17 @@
-// Code generated by go-bindata.
+// Code generated by go-bindata. DO NOT EDIT.
```
srenatus: Hmm this could mean we're using different versions of `go-bindata`. That's kind of asking for trouble...

I'm on
```
$ go-bindata  -version
go-bindata 3.1.0 (Go runtime go1.11.3).
Copyright (c) 2010-2013, Jim Teeuwen.
```
and I've installed it using `go get`, I suppose. What version do you use? Let's standardize. 😉

bcmdarroch: i'm on ```go-bindata 3.13.0 (Go runtime go1.11.4).
Copyright (c) 2010-2015, Jim Teeuwen.
Copyright (c) 2017-2018, Kevin Burke.```

srenatus: Yeah. Where did you get that? 😉

srenatus: Ah, https://github.com/kevinburke/go-bindata

srenatus:
```
$ go get -u github.com/kevinburke/go-bindata/go-bindata
$ go-bindata  -version
go-bindata 3.13.0 (Go runtime go1.12.5).
Copyright (c) 2010-2015, Jim Teeuwen.
Copyright (c) 2017-2018, Kevin Burke.
$
```
:woohoo:

api/interservice/authz/v2/project.proto
bcmdarroch: not yet, maybe we can chat at standup?

api/interservice/authz/v2/project.proto
```diff
@@ -92,19 +91,11 @@ message ProjectUpdateStatusResp {
...
-    string project_id = 1;
-}
-
-message GetProjectRulesResp {
-    ProjectRules rules_for_project = 2;
-}
-
 message ProjectRules {
     repeated ProjectRule rules = 2;
 }
```
msorens: Later in this file, it defines these:
```
enum ProjectRuleTypes {
    NODE = 0;
    EVENT = 1;
}
```
But in components/automate-gateway/api/iam/v2beta/common/rules.proto there are these:
```
enum RuleType {
    RULE_TYPE_UNSET = 0;
    NODE = 1;
    EVENT = 2;
}
```
Two concerns:
1. NODE type is a zero-value in the former, so it is suppressed in the output (as you noted in your preamble). Having the explicit unset/zero value, as the latter does, would improve the output, always showing "node" and "event".
2. Perhaps more significantly, these two definitions are inconsistent for no particular reason. Would be much better if they were consistent.

components/authz-service/engine/opa/opa.go
```diff
@@ -636,16 +533,27 @@ func (s *State) V2SetPolicies(
...
 	s.v2p1Store = inmem.NewFromObject(map[string]interface{}{
 		"policies": policyMap,
 		"roles":    roleMap,
-		"rules":    ruleMap,
 	})
 
 	return s.initPartialResultV2p1(ctx)
 }
 
+// SetRules replaces OPA's rule cache with an updated rule map
```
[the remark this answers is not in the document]
bcmdarroch: The rule cache is [a part of OPA's state](https://github.com/chef/automate/pull/517/files/60eccbda6bd6a847c3894b7ab6787d23f444ab34#diff-2d317bb33ff7babca7e1290b0fb1dcf2R31). The rule store is not a fully-fledged OPA store, just a go cache. But by keeping the rule cache as part of OPA's state, we can take advantage of the policy_refresher to keep everything in sync for HA.

components/authz-service/projects_spike.md
```diff
@@ -13,12 +13,12 @@ rebuild components/authz-service
 
 ## Endpoints
```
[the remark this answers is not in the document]
bcmdarroch: I'd rather delete the file since the spike details no longer apply. ListRulesForProject doesn't even depend on OPA anymore

components/authz-service/storage/v2/memstore/memstore.go
```diff
@@ -246,6 +246,7 @@ func (s *State) CreateRule(_ context.Context, rule *storage.Rule) (*storage.Rule
 	if err := s.rules.Add(rule.ID, rule, cache.NoExpiration); err != nil {
 		return nil, storage_errors.ErrConflict
 	}
+	s.bumpPolicyVersion()
```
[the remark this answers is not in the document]
bcmdarroch: I'd rather save that rename for a follow-up PR. Changing that function name would require changing the `policyChangeID` to a more appropriate name as well, so I'll need to take care of all those ripples

components/authz-service/server/v2/policy_refresher.go
```diff
@@ -184,9 +184,12 @@ func (refresher *policyRefresher) updateEngineStore(ctx context.Context) error {
 
 	switch {
 	case vsn.Minor == api.Version_V1: // v2.1
-		return refresher.engine.V2p1SetPolicies(ctx, policyMap, roleMap, ruleMap)
+		if err := refresher.engine.SetRules(ctx, ruleMap); err != nil {
```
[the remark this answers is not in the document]
bcmdarroch: Updating the caches together is a good first step. But I do agree some performance analysis would be worthwhile to see if separating their updates is worth the cost of implementing another way to notify authz-server instances of changes

api/interservice/authz/v2/project.proto
bcmdarroch: tbh I'm not sure why they were defined that way. Looks like we're using the UNSET value [here](https://github.com/chef/automate/blob/master/components/automate-gateway/handler/iam/v2beta/rules/rules.go#L167) to do some validation at the gateway level.

In any case, I'd like to discuss these APIs and come to an agreement on what they should look like as a team.

msorens: I'll just add that I particularly don't like that if the rule is "node" type then it does not include the type in the output changing the enum values to 1 and 2 would fix that.

components/authz-service/server/v2/policy_refresher.go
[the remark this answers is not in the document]
srenatus: I agree -- given that we've initially planned to put all the data _into one OPA store_, where we'd have had to update the partial eval state for any rule change, too, this is a good first step! Having the cache separate allows for further optimizations. Let's keep an eye on v2.1 partial eval update times, but I think they're pretty fast right now.

api/interservice/authz/v2/project.proto
srenatus: It's a good practice for GRPC APIs, I believe: due to the way that an unset enum _is its first value_. When using `grpc-gateway`, as we do, it means the user _has to provide_ the rule type -- if they don't, it'll fail as unset; without the first value, it would default to the first value. At least in my eyes, that would be a bit surprising.

Random googled up blog post re: protobuf enums: http://androiddevblog.com/protocol-buffers-pitfall-adding-enum-values/

srenatus:
ℹ️ This still leaves `ingest-service` querying an empty rule cache when the system is not using v2.1 -- but it's better than before (and #561), where these queries would fail ("empty eval" OPA error), in that now, they'll just return nothing. Merging.

PR #518, opened 2019-06-06
Rename SQL migration to have unique, monotonic sequence number

msorens:

Problem reported in https://chefio.slack.com/archives/C6JED4R8W/p1559784169026800:
```
Jun 06 01:22:24 a2-local-inplace-upgrade-dev.cd.chef.co hab[6080]: authz-service.default(O): could not initialize v1 server: init postgres for IAM v1: apply database migrations from /hab/svc/authz-service/static/migrations: init migrate: unable to parse file 45_query_rules_for_project.up.sql
```

The cause of this: looks like merge-contention. @kmacgugan  merged a PR with an authz SQL migration 3 hours after @tylercloke  merged another PR with another authz SQL migration... but both used the same migration sequence number (45), which is a Bad Thing :tm: . This PR corrects the numbering, but given the hysteresis of migrations, I'm not completely sure if it will fix the problem.

### :+1: Definition of Done

Error goes away.

- [ ] Code actually executed?


srenatus:
So, as mentioned in slack, "failed to parse file" here is caused by the double use of number 45. (see https://github.com/mattes/migrate/blob/4768a648fbd9e04389a73a21139d14a4ccb1a61a/source/file/file.go#L71 and https://github.com/mattes/migrate/blob/4768a648fbd9e04389a73a21139d14a4ccb1a61a/source/migration.go#L57-L60).

Since loading the migrations fails, *nothing is executed* against the database. No schema entries should have been marked as "dirty". We thus **don't have to do this dance again**: https://github.com/chef/automate/blame/a298421c1eb69411ab9f55fe5e567579aa4c0f7f/components/authz-service/storage/postgres/migration/migration.go#L40

I'd like to verify this is the same in our terraformed dev envs, will do that in a oment.

srenatus:
```
Jun 06 07:34:00 a2-iamv2-local-inplace-upgrade-dev.cd.chef.co hab[23363]: authz-service.default(O): could not initialize v1 server: init postgres for IAM v1: apply database migrations from /hab/svc/authz-service/static/migrations: init migrate: unable to parse file 45_query_rules_for_project.up.sql
```
exactly the same thing we see in local dev envs happens in the other envs, and our in-place-upgrade envs still have a "non-dirty" schema entry:
```
root@a2-iamv2-local-inplace-upgrade-dev:~# chef-automate dev psql chef_authz_service -- -c 'select * from schema_migrations'
 version | dirty
---------+-------
      45 | f
(1 row)
```

So, **let's force-merge this**.

PR #527, opened 2019-06-06
Add ability to complete workflow instance with a result

jaym:
The nomenclature here needs some discussion. Perhaps payload/result is the wrong thing and we could just call it state. Then you wouldn't need to check if you're running or not to determine if you should grab the payload or the result.

### DIFF
--- a/lib/workflow/postgres/postgres.go
+++ b/lib/workflow/postgres/postgres.go
@@ -437,7 +437,7 @@ func (pg *PostgresBackend) GetWorkflowInstanceByName(ctx context.Context, instan
 	if err != nil {
 		if err == sql.ErrNoRows {
 			row := tx.QueryRowContext(ctx,
-				"SELECT parameters, result FROM workflow_results WHERE workflow_name = $1 AND instance_name = $2 ORDER BY end_at DESC",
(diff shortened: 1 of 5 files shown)


lib/workflow/postgres/postgres.go
```diff
@@ -412,6 +412,56 @@ VALUES ($1, $2, $3, $4, $5, $6)`,
...
+	}
+	err = row.Scan(
+		&workflowInstance.Status,
+		&workflowInstance.Parameters,
+		&workflowInstance.Payload,
+	)
+	if err != nil {
+		if err == sql.ErrNoRows {
+			row := tx.QueryRowContext(ctx,
+				"SELECT parameters, result FROM workflow_results WHERE workflow_name = $1 AND instance_name = $2 ORDER BY end_at DESC",
```
stevendanna: I'm a little hesitant to use the `end_at` time here.  I wonder, since for a given named workflow, we know that we can't have overlapping executions of it, might we instead order by the ID serial?  We'll have a potential rollover issue then. 🤷‍♂ Perhaps depend on the time isn't too bad.

jaym: i like that better. let me give it a go

PR #530, opened 2019-06-06
Update the 30_jobs_docker tests to run reliably in local dev env

alexpop:
Because of the two `before` statements in `30_jobs_docker_spec.rb`, first run of the tests failed locally.

I was also getting this error when asserting the grpc outputs:
```
No visible difference in the Array#inspect output.
You should look at the implementation of #== on Array or its members.
```

So, I switched to json based assertions like we have in other test files.


components/compliance-service/api/tests/Gemfile
```diff
@@ -6,4 +6,5 @@ gem 'rb-readline', '~> 0.5'
 gem 'minitest-hooks', '~> 1.5.0'
 gem 'grpc', '= 1.18.0'
 gem 'grpc-tools', '= 1.18.0'
-gem 'google-protobuf', '= 3.6.1'
+gem 'google-protobuf', '= 3.8.0'
```
phiggins: There are a lot of changes in the tests but no app updates, is updating this gem the reason we had to make so many changes? Is there any particular reason we need to update this gem?

components/compliance-service/api/tests/run.rb
```diff
@@ -105,6 +106,10 @@ def assert_grpc_error(details, code)
   def assert_equal_json_content(expected, actual)
     assert_equal JSON.parse(expected.to_json), JSON.parse(actual.to_json)
   end
+
+  def assert_equal_json_sorted(expected, actual)
+    assert_equal JSON.parse(expected).deep_sort, JSON.parse(actual).deep_sort
+  end
```
phiggins: If there are tests that rely on ordering, should we order the test data explicitly?

components/compliance-service/api/tests/Gemfile
vjeffrey: aren't most of the changes a result of updating these to be json?

alexpop: Latest ruby version from chefdk didn't work with google-protobuf 3.6.1.
Some of the tests updated are fixes to tests that were ignored due to a testing framework bug Rick is working on.

components/compliance-service/api/tests/run.rb
alexpop:
```
    expected_job1 = {
      "id": job_id1,
      "name": "my Detect Job For two nodes",
      "type": "detect",
      "timeout": 600,
      "tags": [
        {
          "key": "reason",
          "value": "checking the connectivity"
        },
        {
          "key": "team",
          "value": "Spain"
        }
      ],
```
In the jobs output, tags(see example above) came with a random order, hence the deep sort of the json.

components/compliance-service/api/tests/Gemfile
rickmarry: the 'exit' bug was taken care of in this pr. 
7aa54818ac797c60855b913a6cf23859b39b7d8f

PR #542, opened 2019-06-07
take exit out of test that caused premature termination

rickmarry:

- [ ] Code actually executed?


components/compliance-service/api/tests/08_deep_search_profiles_spec.rb
```diff
@@ -124,7 +124,7 @@ def reporting;
                         },
                         "status" => "failed",
                         "controls" => {
-                            "total" => 1,
+                            "total" => 2,
```
rickmarry: putting this 2 in here just to test that the pipeline will properly fail.  this should cause a test failure.  once this works, i'll change it back to 1

PR #544, opened 2019-06-07
authz-service/{datamigrations,migrations}: add Makefile for timestamp versions

srenatus:
This is @phiggins idea -- it's _very_ unlikely that two PRs ever touch
the same migration again.

To create a new one, use `make new`. You can also pass a name:

    $ cd components/authz-service/storage/postgres/migration/sql/
    $ make new
    => 20190607130445_TODO.up.sql
    $ make new name=do_foo_things
    => 20190607130452_do_foo_things.up.sql

Of course, it's possible that the migrations of two different PRs do not
commute (i.e., A;B is different from B;A), but that situation has always
been a problem.

The library we use ends up [calling `sort.Sort` on the versions](https://github.com/mattes/migrate/blob/4768a648fbd9e04389a73a21139d14a4ccb1a61a/source/migration.go#L73), casted to integers. So, this is fine.

### DIFF
--- a/components/authz-service/storage/postgres/datamigration/sql/Makefile
+++ b/components/authz-service/storage/postgres/datamigration/sql/Makefile
@@ -1,0 +1,1 @@
+../../migration/sql/Makefile
--- a/components/authz-service/storage/postgres/migration/sql/Makefile
+++ b/components/authz-service/storage/postgres/migration/sql/Makefile
@@ -1,0 +1,5 @@
+name:=TODO
+new: filename=$(shell echo "$$(date -u '+%Y%m%d%H%M%S')_$(name).up.sql")
+new:
+	@touch $(filename)
+	@echo "=> $(filename)"


sdelano:
I believe there are specific reasons that we haven't used timestamp-based numbering for migrations in the past. @danielsdeleo do you happen to remember? I seem to recall there was a good explanation on the philosophy behind it somewhere on the Internet at some point in time :)

danielsdeleo:
Yeah, I'm starting to remember the downsides of this now. The problem is that developer Alice can craft a `migration1` and some time later in the day, Bob independently crafts `migration2`. It's possible that Bob's pull request gets merged first, and now any system exposed to that code will have a schema `oldmaster+migration2`. Later we can still merge Alice's pull request since git will not complain. At that point the behavior for systems that have deployed the previous package depends on the implementation of the migrator library, but either Alice's migration gets skipped, or we will have some systems out in the world that ran `oldmaster->migration2->migration1` and some that run `oldmaster->migration1->migration2`. Most of the time that probably turns out fine but since schema migrations are basically mutation of shared state, there can be nasty surprises. And debugging it will be unpleasant since you'll see a different behavior on a fresh install than the systems that deployed the intermediate package first.

PR #546, opened 2019-06-07
 [A2-832] Populate global project filter with names instead of ids (round 2)

msorens:

This updates the global project filter to display project names instead of project ids.

It turns out that this rather benign-sounding task involved more intricacies than one might expect.
First, one must deal with the fact that every UI call to the back-end now includes _requested projects_ (i.e. the current selection from the global project filter). But if that is done for the call that populates the global project filter it would restrict its universe of discourse to just the currently selected projects, when it should always be looking at all allowed projects.

This could be solved in a number of ways. The back-end could, for example, just be aware of the situation and choose to ignore the project headers that were sent. Another approach is to provide a mechanism for front-end calls to opt-out of sending the requested projects at all. I started with the latter approach, as it was straightforward. But after doing some coding, and getting some feedback, I opted to go back to the former, making the back-end aware of the situation.

The next problem was how to marry project names with introspected project IDs. This also had a number of solutions. The front-end could do two calls, one for IntrospectAllProjects and one for ListProjects, then marry the data. But then there's that finicky special case where the latter never returns "(unassigned)" while that would possibly show up in the former. So more special casing there it would seem.

Another solution is to do essentially the same in the back-end: get the introspection result, get the projects list, and marry them. But rather than use ListProjects use a new ListProjectsForIntrospection, so that it would be able to return  "(unassigned)"  if appropriate. This is what you will find here.

### :+1: Definition of Done

image

Components to build:
- authz-service
- automate-gateway

Create project-p1, project-p2, and project-p3 (UI or CLI).

Create local user "bob".

Create a policy for bob:
```
curl -sSkH "api-token: $TOK" $TARGET_HOST/apis/iam/v2beta/policies --data @policy.json | jq
```
With this text in policy.json:
```
{
  "id": "filtering-test",
  "name": "testing filtering of a list by project",
  "projects": [],
  "statements": [
    {
      "effect": "ALLOW",
      "projects": ["project-p3","project-p2","(unassigned)"]
    }
  ]
}
```
This policy includes a subset of the 3 projects above, not all of them. That allows one to see that the global projects filter gets just _allowed_ projects rather than all projects.

This policy also gives bob access to "(unassigned)" so one can see that does show up in the list.

After logging in as bob you should find that bob's global project filter looks like the screenshot above,
containing the three projects in the policy above.

Go to the projects list view and you can see entries appear when either project-p2 or project-p3 is selected.
Go to the teams list view and you can see entries appear when "(unassigned)" is selected.

- [x] Code actually executed?


components/authz-service/server/v2/project_test.go
```diff
@@ -336,6 +336,54 @@ func TestListProjects(t *testing.T) {
...
+			apiProject1 := addProjectToStore(t, store, id1, "my bar", storage.Custom)
+
+			resp, err := cl.ListProjectsForIntrospection(ctx, &api.ListProjectsReq{})
+			require.NoError(t, err)
+			assert.ElementsMatch(t, []*api.Project{&apiProject0, &apiProject1}, resp.Projects)
+		}},
+		{"suppresses hidden system projects except for unassigned", func(t *testing.T) {
+			require.Zero(t, store.ItemCount())
+			addProjectToStore(t, store, constants.AllProjectsID, "All Projects", storage.ChefManaged)
+			unassignedProject := addProjectToStore(t, store, constants.UnassignedProjectID, "Unassigned", storage.ChefManaged)
```
blakestier: Seems like these projects should *always* exist in the store (rather than be added for this test) as that is more comparable to how our system is in production/

components/automate-ui/src/app/services/http/http-client-auth.interceptor.spec.ts
```diff
@@ -123,6 +123,56 @@ describe('HttpClientAuthInterceptor', () => {
...
+        ];
+        configure(projectList);
+        httpClient = TestBed.get(HttpClient);
+        httpMock = TestBed.get(HttpTestingController);
+        chefSession = TestBed.get(ChefSessionService);
+        spyOnProperty(chefSession, 'id_token', 'get').and.returnValue('token');
+      });
+
+      using([
+        ['does not include project header', true],
```
blakestier: fwiw, I found this abstraction made it a little trickier to read the tests and might prefer the redundancy. No biggie though.

components/automate-ui/src/app/services/projects-filter/projects-filter.requests.ts
```diff
@@ -5,16 +5,16 @@ import { Observable } from 'rxjs';
...
+}
 
 @Injectable()
 export class ProjectsFilterRequests {
 
   constructor(private http: HttpClient) { }
 
   fetchOptions(): Observable<AuthorizedProjectsResponse[]> {
-    return this.http.get<AuthorizedProjectsResponse[]>(`${env.auth_url}/introspect_projects`);
+    return this.http.get<AuthorizedProjectsResponse[]>(`${env.auth_v2_url}/introspect_projects`);
```
blakestier: 👍

components/automate-gateway/handler/iam/v2beta/policy/policy.go
```diff
@@ -406,7 +412,7 @@ func (p *Server) GetProject(
 	return &pb_resp.GetProjectResp{Project: proj}, nil
 }
 
-// ListProjects returns the list of all projects.
+// ListProjects returns the list of all projects for display on the List Projects view.
```
blakestier: I'm curious about this comment-- `ListProjects` isn't used elsewhere? It's exposed, right?

blakestier:
oops just saw the WIP 😬

components/authz-service/server/v2/project.go
```diff
@@ -19,7 +20,7 @@ import (
 	api "github.com/chef/automate/api/interservice/authz/v2"
 	automate_event "github.com/chef/automate/api/interservice/event"
 	"github.com/chef/automate/components/authz-service/config"
-	v2_constants "github.com/chef/automate/components/authz-service/constants/v2"
+	constants_v2 "github.com/chef/automate/components/authz-service/constants/v2"
```
msorens: Drive-by: the uses of "constants_v2" far outweighed "v2_constants" in the codebase so I changed those few of the latter to the former.

components/automate-ui/src/app/pages/user-details/user-details.component.ts
```diff
@@ -1,10 +1,10 @@
-import { combineLatest as combineLatest,  Observable ,  Subscription } from 'rxjs';
+import { combineLatest, Observable, Subscription } from 'rxjs';
```
msorens: Drive-by: remove redundant bit.

components/automate-ui/src/app/services/http/http-client-auth.interceptor.spec.ts
```diff
@@ -123,6 +123,56 @@ describe('HttpClientAuthInterceptor', () => {
...
+        ['includes project header', false]
+      ], function (description: string, setting: boolean) {
+
+        it(description + 'project header when unfiltered flag set to ' + setting, done => {
+          const options = { params: { unfiltered: String(setting) } };
```
msorens: Any call that needs to say "do not send project header" can now just pass this special unfilitered parameter set to true.

components/automate-ui/src/app/services/projects-filter/projects-filter.effects.ts
```diff
@@ -42,7 +45,7 @@ export class ProjectsFilterEffects {
 
   private loadOptionsAction$(): () => Observable<ProjectsFilterActions> {
     return () => this.requests.fetchOptions().pipe(
-      map((fetched: AuthorizedProjectsResponse) => {
+      map((fetched: GetProjectsSuccessPayload) => {
```
msorens: This file contains what is really all the front-end tweaks needed to use names instead of ids.

components/authz-service/server/v2/project.go
```diff
@@ -214,6 +217,38 @@ func (s *state) ListProjects(ctx context.Context,
 	return &resp, nil
 }
 
+func (s *state) ListProjectsForIntrospection(
```
msorens: Two differences from ListProjects:
1. Needs to skip project filtering.
2. Needs to include "(unassigned)" when appropriate.

components/automate-gateway/api/authz/authz.proto
```diff
@@ -66,11 +66,4 @@ service Authorization {
     option (chef.automate.api.iam.policy).resource = "iam:introspect";
     option (chef.automate.api.iam.policy).action = "iam:introspect:get";
     };
-  rpc IntrospectAllProjects (authz.request.IntrospectAllProjectsReq) returns (authz.response.IntrospectProjectsResp) {
```
msorens: Removing from the v1 proto file since it is a v2-only function.

components/automate-gateway/gateway/services.go
```diff
@@ -139,7 +144,7 @@ func (s *Server) RegisterGRPCServices(grpcServer *grpc.Server) error {
 		return errors.Wrap(err, "create client for authzV2 service")
 	}
 	pb_authz.RegisterAuthorizationServer(grpcServer,
-		handler.NewAuthzServer(authzClient, authzV2Client, s.authorizer))
+		handler.NewAuthzServer(authzClient, s.authorizer))
```
msorens: This line is the only substantive change in this file, removing a no-longer-needed argument.
All the remaining changes in this file are making lines shorter.

components/authz-service/storage/v2/storage.go
```diff
@@ -198,7 +198,7 @@ func DefaultProjects() []Project {
 
 	unassignedProject := Project{
 		ID:       constants_v2.UnassignedProjectID,
-		Name:     "Unassigned Project",
+		Name:     constants_v2.UnassignedProjectID,
```
msorens: Originally I did not include this--having the change only in the migration--but the integration test failed without it.

components/authz-service/storage/postgres/datamigration/sql/08_update_v2_projects.up.sql
```diff
@@ -0,0 +1,9 @@
+BEGIN;
+
+UPDATE iam_projects
+    SET
+        name = '(unassigned)'
```
msorens: New UX change: We want this instead of "Unassigned Project"

api/interservice/authz/v2/project.proto
```diff
@@ -59,6 +59,7 @@ service Projects {
     rpc GetProject(GetProjectReq) returns (GetProjectResp) {};
     rpc DeleteProject(DeleteProjectReq) returns (DeleteProjectResp) {};
     rpc ListProjects(ListProjectsReq) returns (ListProjectsResp) {};
+    rpc ListProjectsForIntrospection(ListProjectsReq) returns (ListProjectsResp) {};
```
bcmdarroch: I like the separate API 👍

PR #548, opened 2019-06-07
[A2-845] Add integration testing for rules API

tylercloke:
Signed-off-by: Tyler Cloke <tylercloke@gmail.com>

- [ ] Code actually executed?

### DIFF
--- a/inspec/a2-api-integration/controls/iam_v2.rb
+++ b/inspec/a2-api-integration/controls/iam_v2.rb
@@ -750,4 +750,285 @@ control 'iam-v2-1' do
       expect(resp.parsed_response_body[:user_ids]).to eq([user])
     end
   end
+
+  describe "v2beta project rules API" do
+    before(:all) do
+       resp = automate_api_request("/apis/iam/v2beta/projects",
+        http_method: 'POST',
+        request_body: {
+          id: CUSTOM_PROJECT_ID,
+          name: "display name !#$#"
+        }.to_json
+      )
+      expect(resp.http_status).to eq 200
+    end
+
+    after(:all) do
+      resp = automate_api_request("/apis/iam/v2beta/projects/#{CUSTOM_PROJECT_ID}", http_method: 'DELETE')
+      expect(resp.http_status.to_s).to match(/200|404/)
+    end
+
+    context "when the project does not exist" do
+      before(:each) do
+        resp = automate_api_request("/apis/iam/v2beta/projects/#{CUSTOM_PROJECT_ID}", http_method: 'DELETE')
+        expect(resp.http_status.to_s).to match(/200|404/)
+      end
+
+      after(:all) do
+        resp = automate_api_request("/apis/iam/v2beta/projects",
+          http_method: 'POST',
+          request_body: {
+            id: CUSTOM_PROJECT_ID,
+            name: "display name !#$#"
+          }.to_json
+        )
+        expect(resp.http_status).to eq 200
+      end
+
+      describe "GET /iam/v2beta/projects/:id/rules" do
+        it "returns an empty list" do
+          resp = automate_api_request("/apis/iam/v2beta/projects/project-not-found/rules")
+          expect(resp.http_status).to eq 200
+          expect(resp.parsed_response_body[:rules]).to eq([])
+        end
+      end
+
+      describe "GET /iam/v2beta/rules" do
+        it "returns an empty list" do
+          resp = automate_api_request("/apis/iam/v2beta/rules")
+          expect(resp.http_status).to eq 200
+          expect(resp.parsed_response_body[:rules]).to eq([])
+        end
+      end
+    end
+
+    context "when there are no rules but the project exists" do
+      describe "GET /iam/v2beta/projects/:id/rules" do
+        it "returns an empty list" do
+          resp = automate_api_request("/apis/iam/v2beta/projects/#{CUSTOM_PROJECT_ID}/rules")
+          expect(resp.http_status).to eq 200
+          expect(resp.parsed_response_body[:rules]).to eq([])
+        end
+      end
+
+      describe "GET /iam/v2beta/rules" do
+        it "returns an empty list" do
+          resp = automate_api_request("/apis/iam/v2beta/rules")
+          expect(resp.http_status).to eq 200
+          expect(resp.parsed_response_body[:rules]).to eq([])
+        end
+      end
+
+      describe "GET /iam/v2beta/rules/:id" do
+        it "returns a 404" do
+          resp = automate_api_request("/apis/iam/v2beta/rules/not-found")
+          expect(resp.http_status).to eq 404
+          expect(resp.parsed_response_body[:rule]).to eq(nil)
+        end
+      end
+
+      describe "POST /iam/v2beta/rules" do
+        CUSTOM_RULE = {
+          id: "custom-rule",
+          name: "display name !#$#",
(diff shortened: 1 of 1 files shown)


inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -750,4 +750,285 @@
...
+      )
+      expect(resp.http_status).to eq 200
+    end
+
+    after(:all) do
+      resp = automate_api_request("/apis/iam/v2beta/projects/#{CUSTOM_PROJECT_ID}", http_method: 'DELETE')
+      expect(resp.http_status.to_s).to match(/200|404/)
+    end
+
+    context "when the project does not exist" do
```
bcmdarroch: could we rearrange these tests so the project only gets created in a before block for the test cases that require it?

blakestier: I agree that this would be worthwhile.

inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -750,4 +750,285 @@
...
+          request_body: {
+            id: CUSTOM_PROJECT_ID,
+            name: "display name !#$#"
+          }.to_json
+        )
+        expect(resp.http_status).to eq 200
+      end
+
+      describe "GET /iam/v2beta/projects/:id/rules" do
+        it "returns an empty list" do
```
blakestier: Can you make this test description a bit more explanatory? I was confused why three tests had the same name and had to read the setup detail (`project-not-found`) to sort it out.

inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -750,4 +750,285 @@
...
+      describe "GET /iam/v2beta/rules" do
+        it "returns an empty list" do
+          resp = automate_api_request("/apis/iam/v2beta/rules")
+          expect(resp.http_status).to eq 200
+          expect(resp.parsed_response_body[:rules]).to eq([])
+        end
+      end
+    end
+
+    context "when there are no rules but the project exists" do
```
blakestier: The way these tests are organized (with a context block) is different than above-- let's stick with one if possible.

I'm also not convinced we need this many describe blocks but I'd have to look for a bit to see if there's a better option.

inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -750,4 +750,285 @@
...
+            }
+          ]
+        }
+
+        after(:each) do
+          resp = automate_api_request("/apis/iam/v2beta/rules/#{CUSTOM_RULE[:id]}", http_method: 'DELETE')
+          expect(resp.http_status.to_s).to match(/200|404/)
+        end
+
+        it "creates a new rule" do
```
blakestier: These, for example, are organized in a way that makes a little more sense to me. 

Perhaps we have a describe block for the general api ("rules") and then each test is like "creates the rule with correct params" or "updates the rule" or "update fails with 404 when rule is not found"

inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -803,28 +771,50 @@
...
+       )
+       expect(resp.http_status).to eq 200
+     end
+
+     after(:all) do
+       resp = automate_api_request("/apis/iam/v2beta/projects/#{CUSTOM_PROJECT_ID}", http_method: 'DELETE')
+       expect(resp.http_status.to_s).to match(/200|404/)
+     end
+
+      it "GET /iam/v2beta/projects/:id/rules returns an empty list" do
```
blakestier: when there are no rules?

inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -803,28 +771,50 @@
...
+        expect(resp.parsed_response_body[:rules]).to eq([])
       end
 
-      describe "GET /iam/v2beta/rules/:id" do
-        it "returns a 404" do
-          resp = automate_api_request("/apis/iam/v2beta/rules/not-found")
-          expect(resp.http_status).to eq 404
-          expect(resp.parsed_response_body[:rule]).to eq(nil)
-        end
+      it "GET /iam/v2beta/rules/:id returns a 404" do
```
blakestier: this is when the rule doesn't exist? or if we aren't authorized to view? (i'm guessing the former, but worth clarifying in the test name)

inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -965,43 +951,36 @@
         expect(resp.http_status.to_s).to match(/200|404/)
       end
 
-      describe "GET /iam/v2beta/projects/:id/rules" do
-        it "returns the rules for the project" do
-          resp = automate_api_request("/apis/iam/v2beta/projects/#{CUSTOM_PROJECT_ID}/rules")
-          expect(resp.http_status).to eq 200
-          expect(resp.parsed_response_body[:rules]).to match_array([CUSTOM_RULE_1, CUSTOM_RULE_2])
-        end
+      it "GET /iam/v2beta/projects/:id/rules returns the rules for the project" do
```
blakestier: 👍

inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -965,43 +951,36 @@
...
+        expect(resp.parsed_response_body[:rules]).to match_array([CUSTOM_RULE_1, CUSTOM_RULE_2])
       end
 
-      describe "GET /iam/v2beta/rules" do
-        it "returns all the rules" do
-          resp = automate_api_request("/apis/iam/v2beta/rules")
-          expect(resp.http_status).to eq 200
-          expect(resp.parsed_response_body[:rules]).to match_array([CUSTOM_RULE_1, CUSTOM_RULE_2, CUSTOM_RULE_3])
-        end
+      it "GET /iam/v2beta/rules returns all the rules" do
```
blakestier: 👍

inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -965,43 +951,36 @@
...
-          expect(resp.http_status).to eq 404
-        end
+        resp = automate_api_request("/apis/iam/v2beta/rules/#{CUSTOM_RULE_1[:id]}")
+        expect(resp.http_status).to eq 404
       end
 
-      describe "PUT /iam/v2beta/rules/:id" do
-        UPDATED_RULE = {
+
+      it "PUT /iam/v2beta/rules/:id returns updates the rule" do
```
blakestier: typo in here

inspec/a2-api-integration/controls/iam_v2.rb
tylercloke: Yes, that's the context block it's in in line 773:

```
context "when there are no rules but the project exists" do
```

Setting up preconditions via context is a core part of Rspec / Inspec. Otherwise, we'd need to copy-pasta a lot of setup / teardown for every test. I've simplified the tests so there are almost no nested context / describe blocks though to make it as human-readable as possible. Same for the comment below. The actual output is very clear:

```
     ✔  v2beta project rules API when the project does not exist GET /iam/v2beta/projects/:id/rules returns an empty list
```

blakestier: Oh right-- that context was not part of the commit so I glazed over it.

I understand the value of the context block and was not commenting on that previously. Feel free to just point out the line I didn't see 👍

PR #549, opened 2019-06-07
Switching ingest migrations to work async

lancewf:
The migration of the node-state index is timing out on the a2-local-inplace-upgrade-dev.cd.chef.co 
 server. This could be timing out because the migration is leaving the connection open while the reindex happens. This PR is performing the reindex async. 

Below is the error seen on the a2-local-inplace-upgrade-dev.cd.chef.co  server:
`Jun 07 19:25:12 a2-local-inplace-upgrade-dev.cd.chef.co hab[6080]: ingest-service.default(O): time="2019-06-07T19:25:12Z" level=fatal msg="Unable to reindex node-state to latest" error="elastic: Error 504 (Gateway Timeout)" type=migration`

This code has been tested locally, but I can not get the migration to take as long as it does on the `a2-local-inplace-upgrade-dev.cd.chef.co` server.

### DIFF
--- a/components/ingest-service/backend/client.go
+++ b/components/ingest-service/backend/client.go
@@ -83,5 +83,7 @@ type Client interface {
 	ReindexInsightstoActions(context.Context, string, string) error
 	RefreshIndex(context.Context, string) error
 	GetNodeCount(context.Context, string) (int64, error)
-	ReindexNodeStateToLatest(context.Context, string) error
+	// @parm (context, previousIndex)
+	// @return (taskID, error)
+	ReindexNodeStateToLatest(context.Context, string) (string, error)
 }
--- a/components/ingest-service/backend/client.go
+++ b/components/ingest-service/backend/client.go
@@ -83,7 +83,7 @@ type Client interface {
 	ReindexInsightstoActions(context.Context, string, string) error
 	RefreshIndex(context.Context, string) error
 	GetNodeCount(context.Context, string) (int64, error)
-	// @parm (context, previousIndex)
+	// @param (context, previousIndex)
 	// @return (taskID, error)
 	ReindexNodeStateToLatest(context.Context, string) (string, error)
 }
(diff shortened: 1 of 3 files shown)


components/ingest-service/migration/migration.go
```diff
@@ -256,11 +257,28 @@ func (ms *Status) migrateNodeStateToCurrent(previousIndex string) error {
...
 	ms.update(fmt.Sprintf("Reindexing %s index to current", previousIndex))
-	err := ms.client.ReindexNodeStateToLatest(ms.ctx, previousIndex)
+	reindexTaskID, err := ms.client.ReindexNodeStateToLatest(ms.ctx, previousIndex)
 	if err != nil {
 		ms.updateErr(err.Error(), "Unable to reindex node-state to latest")
 		return err
 	}
+
+	// Wait for Reindex task to complete
+	for {
```
lancewf: Should we time out of this loop? We don't really know how long this could take. If elasticsearch restarts then this loop would fail at line 273.

components/ingest-service/backend/client.go
```diff
@@ -83,5 +83,7 @@ type Client interface {
 	ReindexInsightstoActions(context.Context, string, string) error
 	RefreshIndex(context.Context, string) error
 	GetNodeCount(context.Context, string) (int64, error)
-	ReindexNodeStateToLatest(context.Context, string) error
+	// @parm (context, previousIndex)
```
vjeffrey: should this say `param` ?

components/ingest-service/migration/migration.go
vjeffrey: what would the benefit of timing out give us?  is there any chance this would just go on forever and ever?

components/ingest-service/backend/client.go
lancewf: Yes it should I will update it.

components/ingest-service/migration/migration.go
lancewf: I don't think there would any benefit. This would not go on forever because the job has to end in elasticsearch at some point.

vjeffrey:
also fyi https://github.com/chef/automate/pull/552/files

PR #550, opened 2019-06-07
[A2-866] Condition Type -> Attribute rename

bcmdarroch:
Currently, Conditions are exposed with a `Type` field. This corresponds to the `Attribute` field on the internal storage definition of Condition. These two fields should match for consistency and clarity. Plus, the UI mocks use `Attribute`, so we might as well use Attribute everywhere we can.

### :+1: Definition of Done

- [x] everything still works 

- [ ] Code actually executed?


blakestier:
Code looks good to me-- I noticed in demos that someone mentioned how the word `attribute` was already pretty loaded and confusing. I wonder if this will help or not with that? I know, for instance, that we have something called a `node attribute` . . .is that similar? So a rule has a type that could be `node` and then a condition will have an `attribute` that might be . . . `chef-server`? 

I think that my feeling is that we should name all the things more specifically-- `ResourceType` and `ResourceAttribute` would be helpful here and would maybe relieve some confusion. Also, I agree that most docs talk about attribute over type so at a minimum I'm 👍 on this change if the other is too far-reaching for now.

components/authz-service/server/v2/project.go
```diff
@@ -595,13 +595,13 @@ func fromAPIType(t api.ProjectRuleTypes) (storage.RuleType, error) {
 func rulesToProjectRules(rules []engine.Rule) ([]*api.ProjectRule, error) {
 	conditions := make([]*api.Condition, len(rules))
 	for j, rule := range rules {
-		conditionType, exists := api.ProjectRuleConditionTypes_value[rule.Type]
+		conditionAttribute, exists := api.ProjectRuleConditionAttributes_value[rule.Type]
```
msorens: The thing used to index into `api.ProjectRuleConditionAttributes_value` is an attribute, not a type, now, so not sure why this should work as written.

components/authz-service/server/v2/project.go
```diff
@@ -595,13 +595,13 @@ func fromAPIType(t api.ProjectRuleTypes) (storage.RuleType, error) {
...
 		if !exists {
 			return []*api.ProjectRule{}, errors.New(fmt.Sprintf("Condition type %s is not supported", rule.Type))
```
msorens: The previous comment reflects here as well: this text should say "condition attribute" not "condition type".

components/automate-gateway/handler/iam/v2beta/rules/rules.go
```diff
@@ -281,19 +281,19 @@ func fromInternalCondition(c *authz.Condition) (*pb_common.Condition, error) {
...
+	onceReverseConditionAttributesMapping.Do(func() {
+		for k, v := range externalToAPIConditionAttributes {
+			apiToExternalConditionAttributes[v] = k
 		}
 	})
-	if s, ok := apiToExternalConditionTypes[t]; ok {
+	if s, ok := apiToExternalConditionAttributes[t]; ok {
 		return s, nil
 	}
 	return 0, fmt.Errorf("invalid condition type %s", t.String())
```
msorens: s/type/attribute/
Also consider changing `t` variable to `a`.

components/authz-service/server/v2/project.go
bcmdarroch: This is getting deleted in https://github.com/chef/automate/pull/517 so I didn't go beyond fixing the compiler error

components/authz-service/server/v2/project.go
bcmdarroch: This is getting deleted in https://github.com/chef/automate/pull/517

PR #551, opened 2019-06-07
[A2-861] Expose skip policy migration option to end user

blakestier:

After exposing this flag to the end user, we also needed to update docs.

I have done this work in accord with the ticket, but I wonder if it would make more sense (at least, eventually) to make skipping policy migration the default behavior (as that is what we recommend) and provide a flag to automatically migrate. 🤷‍♀

### DIFF
--- a/components/automate-chef-io/content/docs/iam-v2-guide.md
+++ b/components/automate-chef-io/content/docs/iam-v2-guide.md
@@ -37,7 +37,8 @@ without porting existing policies, use the `--skip-policy-migration` flag: `chef
 
 After you've logged in to Chef Automate, select the **Settings** tab in the top navigation bar, then select and locate the `Policies` section on the left hand panel.
 
-This lists all of your v2 policies:
+Here you can view all of your v2 policies. If you have upgraded without using the `--skip-policy-migration` flag, you will
+also see v1 policies. This includes the following:
 
 * New default (Chef-managed) policies: Administrator, Ingest, Editors, and Viewers.
 * Imported v1 default policies--now called *legacy policies*--in the new v2 policy format and marked with the `[Legacy]` prefix.
@@ -49,6 +50,8 @@ This lists all of your v2 policies:
 
 Now that you are up-and-running with v2 policies, as the first step we recommend that you reconstitute your v1 policies as v2 policies.
 Once that is done, then delete the old legacy v1 policies and you will have a clean, up-to-date system.
+
+Alternately, you can run `--skip-policy-migration` on upgrade to start without any policies at all. You will still need to create new v2 policies to preserve any IAM behavior from v1.
 
 The next few sections explain how to use Chef-managed policies and how to create custom policies.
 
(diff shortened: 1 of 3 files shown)


susanev:
could we take out the bits about recommending not migrating v1 policies, we can address that stuff in the future. these docs changes were intended to focus more on letting users know the flag exists if they decide they don't want to bring any v1 policies to v2.

we can work through the recommended flow and make changes before ga, as that will also require changes to our current iam v2 beta deck that we've been presenting to customers.

components/automate-cli/cmd/chef-automate/iam.go
```diff
@@ -106,9 +105,14 @@ func newIAMUpgradeToV2Cmd() *cobra.Command {
 		false,
 		"Upgrade to version 2.1 with beta project authorization.")
 
-	// all flags are hidden right now
-	cmd.PersistentFlags().VisitAll(func(f *pflag.Flag) { f.Hidden = !isDevMode() })
-
+	var err error
+	if !isDevMode() {
+		err = cmd.PersistentFlags().MarkHidden("beta2.1")
+		if err != nil {
```
srenatus: [nit] How about
```go
if !isDevMode() {
	if err := cmd.PersistentFlags().MarkHidden("beta2.1"); err != nil {
		fmt.Printf("failed marking --beta2.1 hidden: %s\n", err.Error())
	}
}
```
So, this can never happen. No need to panic. In other places, we just ignore the error return,
```go
_ = cmd.PersistentFlags().MarkHidden("beta2.1")
```
I believe.

(ℹ️ the only way this could fail is if the string you're providing didn't match the flag.)

blakestier: I see it both ways in our codebase and was being a 🤓 about always checking errors. But happy to make the change!

components/automate-chef-io/content/docs/iam-v2-guide.md
```diff
@@ -226,6 +231,8 @@ A warning appears if members are still attached to the policy, because deleting
 However, you can still delete the policy.
 Repeat for each legacy policy.
 
+Alternately, you can run `--skip-policy-migration` on upgrade to start without any policies at all.
```
mjingle: This line should be removed as at this stage of the document, you're past the point of upgrading. You've also done a very thorough job of including this information elsewhere in the document.

components/automate-chef-io/content/docs/iam-v2-guide.md
```diff
@@ -30,11 +30,14 @@ Creating default teams Editors and Viewers...
...
 ```
 
+Note: To upgrade without porting existing policies, use the `--skip-policy-migration` flag: `chef-automate iam upgrade-to-v2 --skip-policy-migration`.
+
 ## View Policies
 
 After you've logged in to Chef Automate, select the **Settings** tab in the top navigation bar, then select and locate the `Policies` section on the left hand panel.
 
-This lists all of your v2 policies:
+Here you can view all of your v2 policies. If you have upgraded without using the `--skip-policy-migration` flag, you will
```
mjingle:
```suggestion
In this section, you can view all of your v2 policies. If you have upgraded without using the `--skip-policy-migration` flag, you will
```

mjingle: Also, ideally, it would be great if lines 39 and 40 were put together rather than suddenly split mid-sentence.

components/automate-chef-io/content/docs/iam-v2-guide.md
blakestier: Erring on the side of redundancy 😂

Issue #469, opened 2019-05-31
Service groups table columns update


apriofrost:
[Jira link
](https://chefio.atlassian.net/browse/A2-851)

PR #553, opened 2019-06-10
[apps-service] Update service-groups table

afiune:
Update the service-groups table inside the Applications Tab. Match the mockups.

**Table Column in Mockup**

image-20190531-213458

**Current state**

Screen Shot 2019-06-10 at 2 45 20 PM

### :+1: Definition of Done

- Change column header "Services" to "Service Group"
- Merge the value of "Name" column to "Service Group" column, putting it on top of services counts
- Remove the "Name" column, apply the same sorting logic for "Name" to "Service Group" column.
- Make the services counts "xx of xx" font style bold
- Split the value of "Release" column to "Package" and "Release": "Package" value includes the "origin/package_name", and "Release" value includes the "version/build_stamp".
- Change the alignment of the row content (including the donut chart and the health status icon) to the top, instead of central vertically.

Build and start the UI. Populate the application's database and navigate to the Applications Tab.

#### Normal View
Screen Shot 2019-06-10 at 1 46 59 PM

#### Large/Expanded View
Screen Shot 2019-06-10 at 1 47 12 PM

#### Small/Contracted View
Screen Shot 2019-06-10 at 1 47 40 PM

Jira: https://chefio.atlassian.net/browse/A2-851
Closes: https://github.com/chef/automate/issues/469
 
- [x] Code actually executed?

### DIFF
--- a/components/automate-ui/src/app/pages/service-groups/service-groups.component.html
+++ b/components/automate-ui/src/app/pages/service-groups/service-groups.component.html
@@ -37,11 +37,11 @@
           <chef-th class="health sort" (click)="onToggleSort('percent_ok')">
             Health <i class="sort-icon {{sortIcon('percent_ok')}}"></i>
           </chef-th>
-          <chef-th class="services">
-            Services
+          <chef-th class="service-group">
+            Service Group
           </chef-th>
-          <chef-th class="name sort" (click)="onToggleSort('name')">
-            Name <i class="sort-icon {{sortIcon('name')}}"></i>
+          <chef-th class="package-name sort" (click)="onToggleSort('name')">
+            Package <i class="sort-icon {{sortIcon('name')}}"></i>
           </chef-th>
           <chef-th class="release">
             Release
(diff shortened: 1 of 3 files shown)


danielsdeleo:
So I noticed in the mockup, the wrapping that we have when the package overflows the box is like:

```
the-origin/
long-package-name
```

But the behavior shown in the pictures is like:

```
the-origin/long-
package-name
```

I did a little research and I think there's no way to do what we want in CSS. There are some options to customize hyphenation, but the character at which the break occurs is determined by a language setting (i.e., English or whatever) on the content. I suppose it could be fixed in javascript but I don't really know. Do we care enough to look into this?

danielsdeleo:
Summarizing some discussion from Chef internal slack, we think breaking on `/` makes it more readable but we are okay with how it is now. There is a zero-width space unicode character, and if you put it after the `/`, then the package name does break where we want it to. BUT, that isn't an acceptable solution because the zero-width space gets included in copy/paste operations, which would be annoying if users want to copy information from the table for use in `hab pkg promote` or anything like that.

PR #561, opened 2019-06-10
Add neglected version-adjustment for OPA call

msorens:

Amongst the ad-hoc OPA queries that were all assigned to specific stores in opa.go,
there was one that was assigned to v2 when it is actually used both by v2 and v2.1.
Added the mechanism to provide the version so it can appropriately version-adjust on the OPA call.

### :+1: Definition of Done

We no longer see the **absence** of the header bar:
image

This is the set of steps that manifested the issue:
1. Start fresh hab studio and start the world.
2. Upgrade to 2.1: `chef-automate iam upgrade-to-v2 --beta2.1`
3. Login as admin and observe that all tabs are present across the top nav bar.
4. Restart authz-service: `pkill authz-service`
5. Refresh browser. 

Before the fix, all introspection calls now return "deny" so the top nav bar was essentially absent.
Now it is correctly populated.

- [x] Code actually executed?

### DIFF
--- a/components/authz-service/engine/opa/opa.go
+++ b/components/authz-service/engine/opa/opa.go
@@ -127,6 +127,8 @@ func (s *State) initModules() error {
 		}
 		s.modules = mods
 	}
+
+	// this compiler is for the ad-hoc queries (those *not* having partial results prepped)
 	compiler, err := s.newCompiler()
 	if err != nil {
 		return errors.Wrap(err, "init compiler")
@@ -352,6 +354,7 @@ func (s *State) FilterAuthorizedPairs(
 		"pairs":    pairs,
 	}
 
+	// NB: V1 only, so s.store used here
 	rs, err := s.evalQuery(ctx, s.queries[filteredPairsQuery], opaInput, s.store)
 	if err != nil {
 		return nil, &ErrEvaluation{e: err}
@@ -365,14 +368,21 @@ func (s *State) FilterAuthorizedPairs(
 func (s *State) V2FilterAuthorizedPairs(
 	ctx context.Context,
 	subjects engine.Subjects,
-	pairs []engine.Pair) ([]engine.Pair, error) {
+	pairs []engine.Pair,
+	isBeta2p1 bool,
+) ([]engine.Pair, error) {
 
 	opaInput := map[string]interface{}{
 		"subjects": subjects,
 		"pairs":    pairs,
 	}
 
-	rs, err := s.evalQuery(ctx, s.queries[filteredPairsV2Query], opaInput, s.v2Store)
+	// NB: V2 --or-- V2.1 only, so need to version-adjust the store here
+	store := s.v2Store
+	if isBeta2p1 {
+		store = s.v2p1Store
+	}
+	rs, err := s.evalQuery(ctx, s.queries[filteredPairsV2Query], opaInput, store)
 	if err != nil {
 		return nil, &ErrEvaluation{e: err}
 	}
@@ -390,6 +400,7 @@ func (s *State) V2FilterAuthorizedProjects(
 		"subjects": subjects,
 	}
 
+	// NB: V2.1 only, so s.v2p1Store used here
 	rs, err := s.evalQuery(ctx, s.queries[filteredProjectsV2Query], opaInput, s.v2p1Store)
 	if err != nil {
 		return nil, &ErrEvaluation{e: err}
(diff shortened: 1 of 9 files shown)


components/authz-service/engine/opa/opa.go
```diff
@@ -135,6 +135,8 @@ func (s *State) initModules() error {
 		}
 		s.modules = mods
 	}
+
+	// this compiler is for the ad-hoc queries (those *not* having partial results prepped)
```
msorens: Drive-by: clarification comment.

components/authz-service/engine/opa/opa.go
```diff
@@ -360,6 +362,7 @@ func (s *State) FilterAuthorizedPairs(
 		"pairs":    pairs,
 	}
 
+	// NB: V1 only, so s.store used here
```
msorens: Added an explicit line like this on each `evalQuery` to show that it has been consciously considered.

components/authz-service/engine/opa/opa.go
```diff
@@ -373,14 +376,21 @@ func (s *State) FilterAuthorizedPairs(
...
 		"pairs":    pairs,
 	}
 
-	rs, err := s.evalQuery(ctx, s.queries[filteredPairsV2Query], opaInput, s.v2Store)
+	// NB: V2 --or-- V2.1 only, so need to version-adjust the store here
+	store := s.v2Store
+	if isBeta2p1 {
+		store = s.v2p1Store
+	}
+	rs, err := s.evalQuery(ctx, s.queries[filteredPairsV2Query], opaInput, store)
```
msorens: This is the crux of the fix ^^

components/authz-service/server/v2/authz.go
```diff
@@ -127,7 +126,8 @@ func (s *authzServer) FilterAuthorizedPairs(
 	req *api.FilterAuthorizedPairsReq) (*api.FilterAuthorizedPairsResp, error) {
 	resp, err := s.engine.V2FilterAuthorizedPairs(ctx,
 		engine.Subjects(req.Subjects),
-		toEnginePairs(req.Pairs))
+		toEnginePairs(req.Pairs),
+		s.isBeta2p1())
```
msorens: Pass in the awareness of the IAM version to opa here.

components/authz-service/engine/opa/opa.go
```diff
@@ -398,6 +408,7 @@ func (s *State) V2FilterAuthorizedProjects(
 		"subjects": subjects,
 	}
 
+	// NB: V2.1 only, so s.v2p1Store used here
```
srenatus: Shall we add a v2p1Eval instead of dozens of comments, maybe? 😉

components/authz-service/engine/opa/opa.go
```diff
@@ -426,7 +438,8 @@ func (s *State) RulesForProject(
 
 // ListProjectMappings returns a map of all the rules for each projectID.
 func (s *State) ListProjectMappings(ctx context.Context) (map[string][]engine.Rule, error) {
-	rs, err := s.evalQuery(ctx, s.queries[listProjectMapQuery], map[string]interface{}{}, s.v2Store)
+	// NB: V2.1 only, so s.v2p1Store used here
+	rs, err := s.evalQuery(ctx, s.queries[listProjectMapQuery], map[string]interface{}{}, s.v2p1Store)
```
srenatus: :x: this is why tests fail: you've switched `v2p1Store` for `v2Store`. `ingest-service` **still doesn't know whether A2 is running in IAM v2, v1, or v2.1 mode** -- and queries for this *using this API* regardless. So, it'll only keep working as-is if this uses the v2Store.

How does this ever work? Well, the `policy_refresher` doesn't know about IAM v1. So, on server startup, it'll default to updating the `v2Store` if we haven't upgrade to v2.1.

There is likely another bug here: if the server starts in v2.1 mode, ingestion should fail, as the v2Store has never been initialized.

components/authz-service/server/v2/authz.go
```diff
@@ -170,7 +170,8 @@ func (s *authzServer) FilterAuthorizedProjects(
 	}, nil
 }
 
-func isBeta2p1(version api.Version) bool {
+func (s *authzServer) isBeta2p1() bool {
+	version := s.vSwitch.Version
```
srenatus: It's a bit of a hack, but fine, I expected eventually regretting that this field is public 😅 After all, we've already got a "re-architect `authz-service`" card in the backlog.

srenatus:
@msorens I've added a few commits to make this issue apparent on buildkite. See https://github.com/chef/automate/pull/563 for details, and how it *failed* when put ontop of master. It should now pass on this branch.

components/authz-service/engine/opa/opa.go
msorens: Ah, forgot I changed that one for ListProjectMappings, too. Aarggh. Thanks, @srenatus .

PR #563, opened 2019-06-11
inspec/a2-iam-v2p1-only-integration: add introspection controls

srenatus:
🚧 This is to assure that the test works. It'll likely be wrapped into #561.


srenatus:
Assumption 1: This will pass, since the envs we're resting this with (pre-merge integration) haven't restarted `authz-service`.

srenatus:
Assumption 2: with the next commit, some integration tests should fail.

srenatus:
✔️ This has served its purpose.

PR #570, opened 2019-06-12
[studio] build_all should actually skip non-OSS software

stevendanna:
Kudos to @phiggins for spotting this when reading code.


.studio/common
```diff
@@ -231,6 +231,7 @@ function build_all() {
     [[ -f "$d/habitat/plan.sh" ]] || continue # skip what can't be built
     if [[ "$d" = "/src/components/automate-chef-io" ]] && [[ -n "$AUTOMATE_OSS_BUILD" ]]; then
         log_line "Skipping build of $d since AUTOMATE_OSS_BUILD is set."
+        continue
```
afiune: Shouldn’t this be a `return` instead?

stevendanna: I don't think so because this is inside a for loop where we are trying to build all of the packages and the point of the variable is "build all packages, except those that aren't open source"

afiune: You are right. Disregard

PR #571, opened 2019-06-12
ci: don't wrap emoji jobnames in backticks

srenatus:
image

We just cannot let this happen. Also freeing two `cat`s in the process.


srenatus:
This is rather trivial. Merging.

PR #575, opened 2019-06-12
[A2-282] Wire property based AuthZ tests into the database

tylercloke:

Will be greatly expanding on these tests soon but need to shift focus for a bit and don't want this to get stale.

- [ ] Code actually executed?

### DIFF
--- a/components/authz-service/server/v2/rules_property_test.go
+++ b/components/authz-service/server/v2/rules_property_test.go
@@ -11,7 +11,7 @@ import (
 	"google.golang.org/grpc/codes"
 
 	api "github.com/chef/automate/api/interservice/authz/v2"
-	storage "github.com/chef/automate/components/authz-service/storage/v2"
+	"github.com/chef/automate/components/authz-service/testhelpers"
 	"github.com/chef/automate/lib/grpc/grpctest"
 )
 
@@ -21,7 +21,8 @@ const (
 
 func TestCreateRuleProperties(t *testing.T) {
 	ctx := context.Background()
-	cl, store, _, seed := setupRules(t)
+	// cl, store, _, seed := setupRules(t)
+	cl, testDB, store, _, seed := testhelpers.SetupProjectsAndRulesWithDB(t)
 
 	conditionsGenNode := gen.StructPtr(reflect.TypeOf(&api.Condition{}), map[string]gopter.Gen{
 		"Type": gen.OneConstOf(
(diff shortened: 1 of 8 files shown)


components/authz-service/testhelpers/testhelpers.go
```diff
@@ -0,0 +1,213 @@
+package testhelpers
```
tylercloke: I moved some functions we needed between testing package to a common package since go doesn't let you export functions defined in `_test.go` files outside of the module it resides in.

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -33,25 +37,25 @@ func TestCreateRuleProperties(t *testing.T) {
 			api.ProjectRuleConditionAttributes_POLICY_GROUP,
 			api.ProjectRuleConditionAttributes_POLICY_NAME,
 		),
-		"Values": gen.SliceOf(gen.AnyString()),
+		"Values": gen.SliceOf(gen.UnicodeString(graphicRange)),
 	})
 	conditionsGenEvent := gen.StructPtr(reflect.TypeOf(&api.Condition{}), map[string]gopter.Gen{
 		"Type": gen.OneConstOf(
```
bcmdarroch: These `Type` keys should be `Attribute` now

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -33,25 +37,25 @@ func TestCreateRuleProperties(t *testing.T) {
 			api.ProjectRuleConditionAttributes_POLICY_GROUP,
 			api.ProjectRuleConditionAttributes_POLICY_NAME,
 		),
-		"Values": gen.SliceOf(gen.AnyString()),
+		"Values": gen.SliceOf(gen.UnicodeString(graphicRange)),
```
bcmdarroch: Should we be generating an `Operator` as well?

components/authz-service/testhelpers/testhelpers.go
```diff
@@ -0,0 +1,213 @@
...
+	return &automate_event.StartResponse{}, nil
+}
+
+func (t *MockEventServiceClient) Stop(ctx context.Context,
+	in *automate_event.StopRequest,
+	opts ...grpc.CallOption) (*automate_event.StopResponse, error) {
+	return &automate_event.StopResponse{}, nil
+}
+
+// TODO More testing
```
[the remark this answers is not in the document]
tylercloke: Heh me neither. Was in the code I moved it from originally. Can just delete it.

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -21,7 +23,9 @@ const (
 
 func TestCreateRuleProperties(t *testing.T) {
 	ctx := context.Background()
-	cl, store, _, seed := setupRules(t)
+	cl, testDB, _, _, seed := testhelpers.SetupProjectsAndRulesWithDB(t)
+
+	graphicRange := rangetable.Merge(unicode.GraphicRanges...)
```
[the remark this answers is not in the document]
tylercloke: Yeah, with the previous method, we were getting `NULL` and trying to insert that into postgres, resulting in :sadpanda:

components/authz-service/server/v2/rules_property_test.go
tylercloke: ...Surprised this worked then because it's actually doing database inserts.

components/authz-service/server/v2/rules_property_test.go
tylercloke: Yep, I'm gonna do a follow up PR where I add a bunch of test coverage. Just getting this in with the DB changes and will pick up after we wrap up everything we need to see projects + ingest work e2e.

PR #579, opened 2019-06-12
[backup] Respect context deadlines when restoring the deployment-service

ryancragun:

* When possible, respect the request deadline when restoring the
  deployment-service from backup
* Only resolve the manifest once during restore and pass the manifest to the Restore() gRPC routine

- [x] Code actually executed?


components/automate-deployment/pkg/client/restore.go
```diff
@@ -196,65 +198,69 @@ func (r *DeploymentRestore) stopDeploymentService() error {
...
 
 	// Wait for Habitat to "unload" the process
-	for {
-		deployed, err := r.target.DeployedServices(ctx)
-		if err == nil {
-			alive := false
-			for _, svc := range deployed {
-				if svc.Pkg.Name() == deploymentServiceName {
-					alive = true
+	go func() {
```
stevendanna: [nit] I've been thinking we should prefer named functions for any `go func` statements that extend for more than a handful of lines.  It makes the goroutine a bit easier to spot in dlv and allows us to give a name to the thing we think this is doing.

components/automate-deployment/pkg/client/restore.go
```diff
@@ -196,65 +198,69 @@ func (r *DeploymentRestore) stopDeploymentService() error {
...
+		for {
+			deployed, err := r.target.DeployedServices(ctx)
+			if err == nil {
```
stevendanna: This is existing code, but I wonder if it might be a bit easier to read if we did

```golang
if err != nil {
    // Maybe log?
    // Maybe sleep  a sec?
    continue
}
```

and then outdent the rest of the code.

components/automate-deployment/pkg/client/restore.go
```diff
@@ -196,65 +198,69 @@ func (r *DeploymentRestore) stopDeploymentService() error {
...
-				// There's a small delay between hab "unloading" the service and terminating
-				// it's PID. To ensure that it's dead enough we'll ping the service until
-				// the connection fails.
-				con, err := Connection(500 * time.Millisecond)
-				// If we can't connect to the service we can assume it's down
-				if err != nil {
-					break
+				if !alive {
+					// There's a small delay between hab "unloading" the service and terminating
+					// it's PID. To ensure that it's dead enough we'll ping the service until
```
stevendanna: s/it's/its/

components/automate-deployment/pkg/client/restore.go
```diff
@@ -471,6 +478,7 @@ func (r *DeploymentRestore) Restore() error {
 	// Make the spec actually execute commands
 	spec := *r.metadata.Spec
 	backup.SetCommandExecutor(spec, command.NewExecExecutor())
+	ctx, cancel := context.WithCancel(ctx)
```
stevendanna: Ultimately, I doubt it matters because the parent context of this is going to get canceled at the end of this request, but I think in the general case, not calling cancel on this context is a leak of sorts since the executor only cancels contexts on error unless I've missed something.

components/automate-deployment/pkg/server/server.go
```diff
@@ -1761,6 +1761,45 @@ func (s *server) ManifestVersion(ctx context.Context, d *api.ManifestVersionRequ
...
+		desiredManifest *manifest.A2
+		jsonManifest    []byte
+	)
+
+	jsonManifest = req.GetManifest().GetJson()
+	if jsonManifest == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "failed to parse the manifest json")
+	}
+	manifestClient = client.NewInMemoryClient(jsonManifest)
+	desiredManifest, err = manifestClient.GetCurrentManifest(ctx, "")
```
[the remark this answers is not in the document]
ryancragun: Yeah, that would certainly work but I sorta feel like using the clients instead of relying on the parser directly might be better. I'd rather tie us to a client interface than what the underlying parser is doing, especially as the parser and manifest evolve after time.

components/automate-deployment/pkg/client/restore.go
```diff
@@ -295,38 +298,41 @@ func (r *DeploymentRestore) startDeploymentService(ctx context.Context) error {
...
-		// deployment-service is responding and returning the
-		// correct version than it is up. We don't currently
-		// depend on the Habitat health status here because
-		// that status can be delayed for up to 30 seconds if
-		// the first health check happens to fail.
-		currentRel, err := r.serviceRelease()
-		if err != nil {
-			currentRel = "unknown" // make the error message pretty
-		}
+	go func() {
```
jaym: I'm not sure why we need a goroutine here. The call to will bail and you can check if the context is expired then?

components/automate-deployment/pkg/client/restore.go
jaym: I agree. I think we made a mistake passing the cancel function to the executor. If it wants to be able to cancel stuff, it should create its own context with a cancel func. I'm guessing this does leak though given that we were canceling before

jaym: also, this isn't the server code, so whatever

components/automate-deployment/pkg/client/restore.go
```diff
@@ -521,6 +526,56 @@ func (r *DeploymentRestore) Restore(ctx context.Context) error {
...
+	if err != nil {
+		return status.Wrap(err, status.MarshalError, "Failed to marshal package manifest to JSON")
+	}
+
+	deadline, ok := ctx.Deadline()
+	if !ok {
+		deadline = time.Now().Add(5 * time.Second)
+	}
+
+	con, err := Connection(time.Until(deadline))
```
jaym: it might be worth creating a function `ConnectionCtx(context.Context)` so you don't have to keep doing this deadline calculation

components/automate-deployment/pkg/server/server.go
```diff
@@ -1761,6 +1761,45 @@ func (s *server) ManifestVersion(ctx context.Context, d *api.ManifestVersionRequ
...
+
+	if err == manifest.ErrCannotParse {
+		return nil, status.Errorf(codes.InvalidArgument, "failed to parse the manifest")
+	} else if err != nil {
+		return nil, err
+	}
+
+	if err = s.acquireLock(ctx); err != nil {
```
jaym: I put a comment on acquireLock that is worth a read to see if you really want to use it.

components/automate-deployment/pkg/server/server.go
```diff
@@ -1761,6 +1761,45 @@ func (s *server) ManifestVersion(ctx context.Context, d *api.ManifestVersionRequ
 	}, nil
 }
 
+// SetManifest sets the current package manifest
+func (s *server) SetManifest(ctx context.Context, req *api.SetManifestRequest) (*api.SetManifestResponse, error) {
```
jaym: I think this can get people into a bad state. Perhaps we should limit it to the case where we're doing a restore. I think the case where autoupgrades are off and the deployment-service in the manifest is not older than the current running one.

components/automate-deployment/pkg/server/server.go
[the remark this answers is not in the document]
stevendanna: I'm happy to go with y'all on this. To me, the format of the manifest is a first class citizen and we don't really gain much by intermediating our parsing via the client.

components/automate-deployment/pkg/manifest/parser/parser.go
```diff
@@ -70,7 +68,7 @@ func parseA2Manifest(body []byte) (*manifest.A2, error) {
 	a2 := &manifest.A2{}
 	err := json.Unmarshal(body, a2)
 	if err != nil {
-		return nil, errors.Wrap(err, "Failed to A2 manifest")
+		return nil, manifest.ErrCannotParse
```
stevendanna: I generally like error constants like this, but here we are losing most of the helpful error information. It is a _lot_more verbose, but if you want be able to check for a specific error and not lose detail information, we might consider making a custom error. For instance, over in the certauthority package we do:

https://github.com/chef/automate/blob/master/components/automate-deployment/pkg/certauthority/certauthority.go#L141-L152

which we can then check specifically using a type check:

https://github.com/chef/automate/blob/master/components/automate-deployment/pkg/certauthority/certauthority_test.go#L231-L232

components/automate-deployment/pkg/client/restore.go
ryancragun: There's a TODO to change Connection() to take a context, I was thinking I'd just do that there on another change e v e n t u a l l y

components/automate-deployment/pkg/server/server.go
ryancragun: I think it's fine, it's same thing as Lock(), it'll just check my context first.

components/automate-deployment/pkg/server/server.go
ryancragun: :thinking: Yeah, I suppose you could break yourself in some scenarios, but this was intended as a "cleaner" way to get the deployment-service into the state we want before we start a restore. The `chef-automate` CLI doesn't expose and way to do this, so they'd have to gRPC curl it. I think the chances of a person doing that are super low, but then again, who knows?

jaym: i was more thinking for ourselves. We might forget in the future when it is safe to do something like that and end up making things sad. Also, another solution might be to not have SetManifest at all and pass it into the restore call?

ryancragun: I thought about adding it to the restore call but that just pollutes it even more. I suppose we could tho.

components/automate-deployment/pkg/manifest/parser/parser.go
ryancragun: I'll give you your parse errors :dango:

components/automate-deployment/pkg/server/server.go
ryancragun: Went ahead and changed it, we'll see how it goes

PR #580, opened 2019-06-12
Improved service group query to use a view

kmacgugan:
Added column for full package identity so the query doesn't have to re-concat it

- [ ] Code actually executed?


components/applications-service/pkg/storage/postgres/db.go
```diff
@@ -96,6 +96,7 @@ type service struct {
 	DeploymentID int32  `db:"deployment_id"`
 	SupID        int32  `db:"sup_id"`
 	Channel      string `db:"channel"`
+	FullPkgIdent string `db:"full_pkg_id"`
```
afiune: Can we match the Habitat terminology here? I would suggest changing this to `package_ident`.

From the [habitat docs](https://www.habitat.sh/docs/reference/#package_identifier):
> package_identifier
A Habitat package identifier, split apart into its constituent components

components/applications-service/pkg/storage/postgres/hab_event.go
```diff
@@ -201,7 +201,7 @@ func (db *Postgres) insertNewServiceFromHealthCheckEvent(
 		}
 
 		// 4) Service
-		svc := newService(pkgIdent, health, did, sid, gid)
+		svc := newService(pkgIdent, health, did, sid, gid, svcMetadata.GetPackageIdent())
```
afiune: You already have the package ident here, it is the first argument in this function.

components/applications-service/pkg/storage/postgres/service_groups.go
```diff
@@ -24,47 +24,8 @@ const (
 	// 1) Counts health of services (How many ok, critical, warning and unknown) and its total
 	// 2) Calculates the percentage of services with an ok health.
```
afiune: we might need to update all the comments so that developers don't get confused with them.

components/applications-service/pkg/storage/postgres/hab_event.go
kmacgugan: That is a struct of it broken apart though. Maybe I can refactor further.

kmacgugan: Ok, I moved the full string into the package ident struct so that we don't have a new param here. Let me know what you think.

kmacgugan:
@afiune Ok, I changed the name of the PR

components/applications-service/pkg/storage/postgres/db.go
afiune: @kmacgugan I saw your 👍which led me to believe you agree on the suggestion but I didn't see any change? 🤔

components/applications-service/pkg/storage/postgres/schema/sql/03_service_group_view.up.sql
```diff
@@ -0,0 +1,30 @@
+ALTER TABLE service ADD COLUMN full_pkg_ident TEXT NOT NULL DEFAULT '';
+
+CREATE OR REPLACE VIEW service_groups AS
```
afiune: The name of this file doesn't match with the view name: 

- File singular: `03_service_group_view.up.sql`
- View plural: `VIEW service_groups`

More in the **naming** topic, _I would suggest calling this view differently_ since I have a feeling that it will be confusing in the near future, currently we have a table called `service_group` and this view is created as `service_groups`, if I am new and look at them both, I wouldn't be able to differentiate them. 

- To me, the `service_group` table is the actual definition of a service-group.
- Now, I think the new view is **an aggregation of the service-groups health**, such information is combined from all its services, deployment id, application, and environment name, etc. 

I would suggest this view to be called: `service_group_health_aggregation`. (or something similar)

components/applications-service/pkg/storage/postgres/service_groups.go
```diff
@@ -21,115 +21,74 @@ import (
...
-        AND health_unknown  = 0
-    ) AS warning
-  , COUNT(*) FILTER (
-      WHERE health_ok > 0
-        AND health_critical = 0
-        AND health_warning  = 0
-        AND health_unknown  = 0
-    ) AS ok
-FROM (` + selectServiceGroupHealth + `) AS service_groups_health_counts
+	SELECT COUNT(*) AS total
```
afiune: nit: I think this file has spaces and not tabs. Maybe this was done by your editor? 🤔

components/applications-service/pkg/storage/postgres/schema/sql/03_service_group_view.up.sql
```diff
@@ -0,0 +1,30 @@
+ALTER TABLE service ADD COLUMN full_pkg_ident TEXT NOT NULL DEFAULT '';
```
afiune: What happens with existing records? 

Example: How would this impact our dev environment?  https://a2-perf-test-single-local-inplace-upgrade-dev.cd.chef.co/applications

components/applications-service/pkg/storage/postgres/schema/sql/03_service_group_view.up.sql
lancewf: I agree with @afiune the naming is confusing. Just by looking at the names of the 'service_groups' view and 'service_group' table I would be confused. It looks like the view is the one that needs the rename.

components/applications-service/pkg/storage/postgres/service_groups.go
kmacgugan: I thought we were going with spaces?

afiune: Yes, they were spaces, but here you are changing them for tabs.

components/applications-service/pkg/storage/postgres/db.go
```diff
@@ -96,6 +96,7 @@ type service struct {
 	DeploymentID int32  `db:"deployment_id"`
 	SupID        int32  `db:"sup_id"`
 	Channel      string `db:"channel"`
+	FullPkgIdent string `db:"full_pkg_ident"`
```
afiune: I think we agreed this field should be called `package_ident` or the long name `package_identifier`.

components/applications-service/pkg/storage/postgres/schema/sql/03_service_group_view.up.sql
```diff
@@ -0,0 +1,33 @@
...
+
+UPDATE service SET full_pkg_ident =
+  CONCAT (service.origin, '/', service.name, '/', service.version, '/', service.release);
+
+CREATE OR REPLACE VIEW service_group_health AS
```
afiune: Thank you so much for changing the name of the view! 💯 -- Could you please also rename this file to `03_service_group_health_view.up.sql` to match the new name? 🥇

components/applications-service/pkg/storage/postgres/service_groups.go
```diff
@@ -21,115 +21,74 @@ import (
...
+						 WHERE health_warning  > 0
+							 AND health_critical = 0
+							 AND health_unknown  = 0
+			 ) AS warning
+	,COUNT(*) FILTER (
+						 WHERE health_ok > 0
+							 AND health_critical = 0
+							 AND health_warning  = 0
+							 AND health_unknown  = 0
+			 ) AS ok
```
afiune: all these modifications above are introducing tabs in our queries, we want spaces.

components/applications-service/pkg/storage/postgres/service_groups.go
```diff
@@ -21,115 +21,74 @@ import (
...
 `
 	selectServiceGroupHealthFilterOK = `
-SELECT * FROM (` + selectServiceGroupHealth + `) AS service_groups_health
-WHERE health_ok > 0
-  AND health_critical = 0
-  AND health_warning  = 0
-  AND health_unknown  = 0
-ORDER BY %s
-LIMIT $1
+SELECT * FROM service_groups AS service_groups_health
```
afiune: You forgot to update all the queries with the new view name. (`s/service_groups/service_group_health/g`)
```suggestion
SELECT * FROM service_group_health AS service_group_health
```

Also, we don't need the `AS` statement here anymore and it is not matching the new view name since it is plural, so perhaps you could remove it or update it as above.

components/applications-service/pkg/storage/postgres/db.go
kmacgugan: oh I thought you meant I should change it from id to ident, thus it became full_pkg_ident. I can change it again though I don't have an attachment to any name.

components/applications-service/pkg/storage/postgres/service_groups.go
```diff
@@ -21,115 +21,74 @@ import (
...
+               AND health_critical = 0
+               AND health_unknown  = 0
+       ) AS warning
+  ,COUNT(*) FILTER (
+             WHERE health_ok > 0
+               AND health_critical = 0
+               AND health_warning  = 0
+               AND health_unknown  = 0
+       ) AS ok
+FROM service_groups_health AS service_groups_health_counts
```
afiune: name still wrong. it should be singular, not plural.
```suggestion
FROM service_group_health AS service_group_health_counts
```

PR #583, opened 2019-06-13
ci: don't build the merge ref, try to rebase ourselves

srenatus:
This has proven to be buggy, and misleading. When force-pushing a PR
branch, the merge head doesn't get updated reliably, and the CI run thus
builds the wrong thing, old code.

Additionally, the BK UI will claim it was building the correct commit,
which confuses even more.

~This adds an info message with the commit being built, for extra
visibility.~ Now, we're attempting to non-interactively rebase onto what's `origin/master` at _build time_; and warn if it fails. If it fails, we'll run the build job with the PR's HEAD.

If everything's fine rebasing, it'll show
image

If not, you'll see
image

*however*, your build will still be executed ([see this for a test run where the rebase fails](https://buildkite.com/chef/chef-automate-master-verify-private/builds/2054))

### DIFF
--- a/.buildkite/hooks/pre-command
+++ b/.buildkite/hooks/pre-command
@@ -12,14 +12,8 @@ git fetch origin master
 # Fetch the git tags to see if that addresses the weird smart build behavior for Habitat
 git fetch --tags --force
 
-# By default, Buildkite pulls down HEAD. If we're on a pull-request, pull down
-# the merged head: https://github.com/buildkite/agent/blob/master/bootstrap/bootstrap.go#L698
-if [[ "$BUILDKITE_PULL_REQUEST" != "false" ]]; then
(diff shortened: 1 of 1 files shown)


srenatus:
@vjeffrey this is "our issue" 😉

stevendanna:
The original goal of this code was to allow people to avoid rebasing while still making sure we were testing something close to the state of master post-merge.  Because of the passage of time, it is never really the case that we are testing what will get merged, but now we will be even further from that goal.

This isn't an objection, but it might imply that if we merge this we also want to make sure we are rebasing more complex changes.

srenatus:
@stevendanna I'm open to changing this to do the merge instead; it's just the `+refs/pull/<ID>/merge` thing I want to avoid.

PR #588, opened 2019-06-13
[A2-811] Table to stage rule changes

blakestier:

Before kicking off a long-running re-indexing of tagged resources, users need to be able to stage changes.  This table is the starting point (along with the function `CreateRule` and `GetStagedOrAppliedRule` for POC). 

NOTE:

Technically, we don't want to trigger a reindex if only the name of a rule/project has changed. We've included it anyway to reduce complexity on this first go, but we will eventually want to handle those differently.

- [ ] Code actually executed?

### DIFF
--- a/components/authz-service/server/v2/project.go
+++ b/components/authz-service/server/v2/project.go
@@ -367,7 +367,7 @@ func (s *state) UpdateRule(ctx context.Context, req *api.UpdateRuleReq) (*api.Up
 }
 
 func (s *state) GetRule(ctx context.Context, req *api.GetRuleReq) (*api.GetRuleResp, error) {
-	resp, err := s.store.GetStagedOrCurrentRule(ctx, req.Id)
+	resp, err := s.store.GetStagedOrAppliedRule(ctx, req.Id)
 	if err != nil {
 		if err == storage_errors.ErrNotFound {
 			return nil, status.Errorf(codes.NotFound, "could not find rule with ID %q", req.Id)
--- a/components/authz-service/server/v2/project.go
+++ b/components/authz-service/server/v2/project.go
@@ -379,7 +379,6 @@ func (s *state) GetRule(ctx context.Context, req *api.GetRuleReq) (*api.GetRuleR
 		return nil, status.Errorf(codes.NotFound, "could not find rule with ID %q", req.Id)
 	}
 
-
 	apiRule, err := fromStorageRule(resp)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal,
--- a/components/authz-service/server/v2/project.go
+++ b/components/authz-service/server/v2/project.go
@@ -367,7 +367,7 @@ func (s *state) UpdateRule(ctx context.Context, req *api.UpdateRuleReq) (*api.Up
 }
 
 func (s *state) GetRule(ctx context.Context, req *api.GetRuleReq) (*api.GetRuleResp, error) {
-	resp, err := s.store.GetRule(ctx, req.Id)
+	resp, err := s.store.GetStagedOrCurrentRule(ctx, req.Id)
 	if err != nil {
 		if err == storage_errors.ErrNotFound {
 			return nil, status.Errorf(codes.NotFound, "could not find rule with ID %q", req.Id)
@@ -375,6 +375,10 @@ func (s *state) GetRule(ctx context.Context, req *api.GetRuleReq) (*api.GetRuleR
 		return nil, status.Errorf(codes.Internal,
 			"error retrieving rule with ID %q: %s", req.Id, err.Error())
 	}
+	if resp.Deleted {
+		return nil, status.Errorf(codes.NotFound, "could not find rule with ID %q", req.Id)
+	}
+
 
 	apiRule, err := fromStorageRule(resp)
 	if err != nil {
--- a/components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
@@ -4,8 +4,9 @@ CREATE TABLE iam_staged_projects (
   db_id SERIAL,
   id TEXT PRIMARY KEY,
   name TEXT NOT NULL,
-  deleted BOOLEAN,
   type iam_policy_type NOT NULL DEFAULT 'custom',
+  projects TEXT[],
+  deleted BOOLEAN,
   updated_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
   created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
 );
@@ -76,6 +77,17 @@ CREATE OR REPLACE FUNCTION
     UNION
     SELECT 'staged' as TableName from iam_staged_project_rules a where a.id=_rule_db_id
     );
+
+$$ LANGUAGE sql;
+
+CREATE OR REPLACE FUNCTION
+  query_staged_project(_project_id TEXT, _projects_filter TEXT[])
+  RETURNS json AS $$
+
+  WITH t AS
+    (SELECT p.id, p.name, p.type, p.projects p.deleted FROM iam_staged_projects p
+      WHERE p.id = _project_id AND (array_length(_projects_filter, 1) IS NULL OR p.projects && _projects_filter))
+  SELECT row_to_json(t) AS role FROM t;
 
 $$ LANGUAGE sql;
 
--- a/components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
@@ -1,0 +1,32 @@
+BEGIN;
+
+CREATE TYPE staged_rule_state
+  AS ENUM (
+    'new',
+    'deleted',
+    'updated'
+    );
+
+CREATE TABLE iam_staged_project_rules (
+  db_id SERIAL PRIMARY KEY,
+  id TEXT NOT NULL UNIQUE,
+  project_id TEXT REFERENCES iam_projects,
+  name TEXT NOT NULL,
+  type TEXT NOT NULL,
+  updated_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  state (pending)
+);
+
+CREATE TABLE iam_staged_rule_conditions (
+  db_id SERIAL PRIMARY KEY,
+  rule_db_id INTEGER REFERENCES iam_staged_project_rules ON DELETE CASCADE,
+  value TEXT[] NOT NULL,
+  attribute TEXT NOT NULL,
+  operator TEXT NOT NULL,
+  updated_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  state staged_rule_state NOT NULL
+);
+
+COMMIT;
--- a/components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
@@ -19,10 +19,11 @@ CREATE TABLE iam_staged_rule_conditions (
 );
 
 CREATE OR REPLACE FUNCTION
-  query_staged_rule(_rule_db_id TEXT, _project_filter TEXT[])
+  query_staged_rule(_rule_id TEXT, _project_filter TEXT[])
   RETURNS json AS $$
 
   WITH t AS (
+    -- check for staged rule
     SELECT
       r.id,
       r.project_id,
@@ -35,8 +36,26 @@ CREATE OR REPLACE FUNCTION
     FROM iam_staged_project_rules AS r
     LEFT OUTER JOIN iam_staged_rule_conditions
     AS rc ON rc.rule_db_id=r.db_id
-    WHERE id=_rule_db_id AND projects_match_for_rule(project_id, _project_filter)
+    WHERE id=_rule_id AND projects_match_for_rule(project_id, _project_filter)
     GROUP BY r.id, r.project_id, r.name, r.type, r.deleted
+
+    UNION ALL
+
+    -- check for applied rule
+    SELECT 
+      r.id,
+      r.project_id,
+      r.name,
+      r.type,
+      false as deleted,
+      json_agg(rc) AS conditions
+      FROM iam_project_rules AS r
+      LEFT OUTER JOIN iam_rule_conditions
+      AS rc ON rc.rule_db_id=r.db_id
+      WHERE id=_rule_id AND projects_match_for_rule(project_id, _project_filter)
+      AND
+        NOT EXISTS (SELECT 1 FROM iam_staged_project_rules WHERE id=_rule_id)
+      GROUP BY r.id, r.project_id, r.name, r.type, deleted
   )
   SELECT row_to_json(t) AS rule FROM t;
 
--- a/components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
@@ -43,13 +43,13 @@ CREATE OR REPLACE FUNCTION
 $$ LANGUAGE sql;
 
 CREATE OR REPLACE FUNCTION
-  query_staged_rule_table(_rule_db_id TEXT)
+  query_staged_rule_table(_id TEXT)
   RETURNS TEXT[] AS $$
 
   SELECT ARRAY(
-    SELECT 'current' as TableName from iam_project_rules c where c.id=_rule_db_id
+    SELECT 'current' as TableName from iam_project_rules c where c.id=_id
     UNION
-    SELECT 'staged' as TableName from iam_staged_project_rules a where a.id=_rule_db_id
+    SELECT 'staged' as TableName from iam_staged_project_rules a where a.id=_id
     );
 
 $$ LANGUAGE sql;
--- a/components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
@@ -17,41 +17,56 @@ CREATE TABLE iam_staged_rule_conditions (
   value TEXT[] NOT NULL,
   attribute TEXT NOT NULL,
   operator TEXT NOT NULL,
+  deleted BOOLEAN,
   updated_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
   created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
 );
 
-CREATE FUNCTION valid_staged_rule_state_change(
-  old_state staged_rule_state, new_state staged_rule_state
-) RETURNS boolean
-LANGUAGE sql AS
-$$
-  SELECT EXISTS(
-    SELECT 1
-    FROM (VALUES
-      ('updated', 'deleted'),
-      ('updated', 'updated')
-    ) AS valid(old, new)
-    WHERE valid.old::staged_rule_state=old_state
-    AND valid.new::staged_rule_state=new_state
+CREATE OR REPLACE FUNCTION
+  projects_match_for_staged_rule(_project_id TEXT, _projects_filter TEXT[])
+  RETURNS BOOLEAN AS $$
+    BEGIN
+      RETURN (
+        array_length(_projects_filter, 1) IS NULL
+        OR _project_id = ANY (_projects_filter)
+      );
+    END
+$$ LANGUAGE PLPGSQL;
+
+CREATE OR REPLACE FUNCTION
+  query_staged_rule(_rule_db_id TEXT, _project_filter TEXT[])
+  RETURNS json AS $$
+
+  WITH t AS (
+    SELECT
+      r.id,
+      r.project_id,
+      r.name,
+      r.type,
+      r.deleted,
+      r.updated_at,
+      r.created_at
+      -- A rule can't exist without conditions so we don't need to worry
+      -- about null case here.
+      json_agg(rc) AS conditions
+    FROM iam_staged_project_rules AS r
+    LEFT OUTER JOIN iam_staged_rule_conditions
+    AS rc ON rc.rule_db_id=r.db_id
+    WHERE id=_rule_db_id AND projects_match_for_staged_rule(project_id, _project_filter)
+    GROUP BY r.id, r.project_id, r.name, r.type, r.deleted, r.updated_at, r.created_at
   )
-$$;
+  SELECT row_to_json(t) AS rule FROM t;
 
-CREATE OR REPLACE FUNCTION staged_rule_state_trigger_func() RETURNS trigger
-LANGUAGE plpgsql AS $$
-DECLARE
-BEGIN
-  IF OLD IS NULL THEN
-    RETURN NEW;
-  END IF;
-  IF valid_staged_rule_state_change(OLD.state, NEW.state) THEN
-    RETURN NEW;
-  END IF;
-  RAISE EXCEPTION 'invalid state change: % -> %', OLD.state, NEW.state;
-END
-$$;
+CREATE OR REPLACE FUNCTION
+  query_staged_rule_table(_rule_db_id TEXT)
+  RETURNS TEXT[] AS $$
 
-CREATE TRIGGER staged_rule_state_trigger_func BEFORE UPDATE ON iam_staged_project_rules
-  FOR EACH ROW EXECUTE PROCEDURE staged_rule_state_trigger_func();
+  SELECT ARRAY(
+    SELECT 'current' as TableName from iam_project_rules c where c.id=_rule_db_id
+    UNION
+    SELECT 'staged' as TableName from iam_staged_project_rules a where a.id=_rule_db_id
+    );
+
+$$ LANGUAGE sql;
 
 COMMIT;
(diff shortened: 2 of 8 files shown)


components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
```diff
@@ -0,0 +1,94 @@
...
+CREATE TABLE iam_staged_projects (
+  db_id SERIAL,
+  id TEXT PRIMARY KEY,
+  name TEXT NOT NULL,
+  type iam_policy_type NOT NULL DEFAULT 'custom',
+  projects TEXT[],
+  deleted BOOLEAN,
+  updated_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+);
```
tylercloke: We are staging projects? This is news to me but I guess it makes sense. Must have missed that.

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
```diff
@@ -0,0 +1,94 @@
...
+
+CREATE TABLE iam_staged_project_rules (
+  db_id SERIAL PRIMARY KEY,
+  id TEXT NOT NULL UNIQUE,
+  project_id TEXT REFERENCES iam_staged_projects ON DELETE CASCADE,
+  name TEXT NOT NULL,
+  type TEXT NOT NULL,
+  deleted BOOLEAN,
+  updated_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
tylercloke: I remember us deciding not to have these timestamps when we were kicking off the V2 work (we had them in V1 inconsistently) but I can't really remember the reason or conversation 😓 I guess my feeling is that if we add them here we should add them everywhere for IAM.

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
```diff
@@ -0,0 +1,94 @@
...
+CREATE OR REPLACE FUNCTION
+  projects_match_for_staged_rule(_project_id TEXT, _projects_filter TEXT[])
+  RETURNS BOOLEAN AS $$
+    BEGIN
+      RETURN (
+        array_length(_projects_filter, 1) IS NULL
+        OR _project_id = ANY (_projects_filter)
+      );
+    END
+$$ LANGUAGE PLPGSQL;
```
tylercloke: Not seeing how these functions are different besides the name:

https://github.com/chef/automate/blob/master/components/authz-service/storage/postgres/migration/sql/44_policy_rules.up.sql#L19-L28

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
```diff
@@ -0,0 +1,94 @@
...
+  query_staged_rule_table(_rule_db_id TEXT)
+  RETURNS TEXT[] AS $$
+
+  SELECT ARRAY(
+    SELECT 'current' as TableName from iam_project_rules c where c.id=_rule_db_id
+    UNION
+    SELECT 'staged' as TableName from iam_staged_project_rules a where a.id=_rule_db_id
+    );
+
+$$ LANGUAGE sql;
```
tylercloke: Not quite clear on what this function is for based on the name.

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
blakestier: I just added it bc if a user deletes a project we'd want to trigger an update. . . right? or is that unrelated to this work?

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
blakestier: yeah i agree on that point. still mulling on what might be better.

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
```diff
@@ -17,41 +17,56 @@ CREATE TABLE iam_staged_rule_conditions (
...
-  SELECT EXISTS(
-    SELECT 1
-    FROM (VALUES
-      ('updated', 'deleted'),
-      ('updated', 'updated')
-    ) AS valid(old, new)
-    WHERE valid.old::staged_rule_state=old_state
-    AND valid.new::staged_rule_state=new_state
+CREATE OR REPLACE FUNCTION
+  projects_match_for_staged_rule(_project_id TEXT, _projects_filter TEXT[])
```
bcmdarroch: i don't think we need this function, we could re-use the old `projects_match_for_rule`. unless i'm missing how this one is different 🤔

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
blakestier: Oh yep -- missing the corresponding `query_staged_rule`. I suppose no need for the rename but maybe rename both for clarity 🤔 . Not sure yet how much of the db work around "current" rules will need to stay since we'll mostly be using/displaying/manipulating staged data.

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
bcmdarroch: It's related to this work but we might want to tackle it after getting the rule staging right.

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
blakestier: we can re-use the old one-- see my note to Tyler, though, for other consideration 🤷‍♀

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
blakestier: well we'll put a pin it in but def worth thinking about as a separate task. I haven't seen it anywhere so we should at min make a card.

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
blakestier: (renamed it to `query_rule_table_associations`)

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -6363,6 +6356,152 @@ func insertStatementProject(t *testing.T, db *testhelpers.TestDB, statementID uu
...
+	for _, c := range rule.Conditions {
+		_, err = db.Exec(`
+			INSERT INTO iam_rule_conditions (rule_db_id, value, attribute, operator) VALUES ($1, $2, $3, $4);`,
+			dbID, pq.Array(c.Value), c.Attribute.String(), c.Operator.String())
+		require.NoError(t, err)
+	}
+	return &rule
+}
+
+func setup(t *testing.T) (storage.Storage, *testDB, *prng.Prng) {
```
blakestier: this and code below isn't new, just moved

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
```diff
@@ -0,0 +1,57 @@
...
+    AS rc ON rc.rule_db_id=r.db_id
+    WHERE id=_rule_db_id AND projects_match_for_rule(project_id, _project_filter)
+    GROUP BY r.id, r.project_id, r.name, r.type, r.deleted
+  )
+  SELECT row_to_json(t) AS rule FROM t;
+
+$$ LANGUAGE sql;
+
+CREATE OR REPLACE FUNCTION
+  query_rule_table_associations(_id TEXT)
```
blakestier: behavior changes contingent on whether the rule exists in current/staging/both, hence the need for this helper function

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -962,19 +962,33 @@ func (p *pg) CreateRule(ctx context.Context, rule *v2.Rule) (*v2.Rule, error) {
...
+		`SELECT query_rule_table_associations($1);`, rule.ID)
+	if err != nil {
+		return nil, p.processError(err)
+	}
+	var associations []string
+	if err := row.Scan(pq.Array(&associations)); err != nil {
+		return nil, p.processError(err)
+	}
+	// If any associations return, then the rule already exists in current, staged, or both tables
+	if len(associations) > 0 {
```
blakestier: We could maybe try to do this check in the database with a trigger? Thoughts?

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3309,6 +3309,32 @@ func TestCreateRule(t *testing.T) {
 			assert.Nil(t, resp)
 			assert.Error(t, storage_errors.ErrForeignKey, err)
 		},
+		"when rule exists in the current rules table, return Err": func(t *testing.T) {
```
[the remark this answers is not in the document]
blakestier: lol i dunno. just copied from existing tests. will change.

components/authz-service/server/v2/project.go
```diff
@@ -384,6 +384,24 @@ func (s *state) GetRule(ctx context.Context, req *api.GetRuleReq) (*api.GetRuleR
...
+	resp, err := s.store.GetStagedRule(ctx, req.Id)
+	if err != nil {
+		if err == storage_errors.ErrNotFound {
+			return nil, status.Errorf(codes.NotFound, "could not find rule with ID %q", req.Id)
+		}
+		return nil, status.Errorf(codes.Internal,
+			"error retrieving rule with ID %q: %s", req.Id, err.Error())
+	}
+
+	apiRule, err := fromStorageRule(resp)
```
blakestier: Will this work? I would imagine the struct is a bit different bc of the additional field 🤔

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1137,6 +1137,40 @@ func (p *pg) GetRule(ctx context.Context, id string) (*v2.Rule, error) {
...
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	tx, err := p.db.BeginTx(ctx, nil)
+	if err != nil {
+		return nil, p.processError(err)
+	}
+
+	var rule v2.Rule
+	row := tx.QueryRowContext(ctx, `SELECT query_staged_rule from query_staged_rule($1, $2);`,
```
blakestier: i think we should only need `SELECT query_staged_rule($1, $2);`

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3309,6 +3309,32 @@ func TestCreateRule(t *testing.T) {
...
+			insertTestProject(t, db, projID, "let's go jigglypuff - topsecret", storage.Custom)
+
+			condition1, err := storage.NewCondition(storage.Node, []string{"chef-server-1"}, storage.ChefServer, storage.MemberOf)
+			require.NoError(t, err)
+			rule, err := storage.NewRule("new-id-1", projID, "name", storage.Node, []storage.Condition{condition1})
+			insertStagedRule(t, db, rule)
+			require.NoError(t, err)
+			resp, err := store.CreateRule(ctx, &rule)
+			assert.Nil(t, resp)
+			assert.Error(t, storage_errors.ErrConflict, err)
```
srenatus: ⚠️ I don't think that's how it works: [`assert.Error`](https://godoc.org/github.com/stretchr/testify/assert#Error) has type `func Error(t TestingT, err error, msgAndArgs ...interface{}) bool`, so, what this line does is:

**it's checking that `storage_errors.ErrConflict` is indeed an error, and it's using `err` as extra information to display for when it isn't**

I think we want either
```go
assert.Error(t, err)
```
or
```go
assert.Equal(t, storage_errors.ErrConflict, err)
```
here.

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -6363,6 +6427,68 @@ func insertStatementProject(t *testing.T, db *testhelpers.TestDB, statementID uu
...
+}
+
+func insertStagedRule(t *testing.T, db *testhelpers.TestDB, rule storage.Rule) {
+	t.Helper()
+	row := db.QueryRow(`
+		INSERT INTO iam_staged_project_rules (id, project_id, name, type, deleted) VALUES ($1, $2, $3, $4, $5) RETURNING db_id;`,
+		rule.ID, rule.ProjectID, rule.Name, rule.Type.String(), false)
+	var dbID string
+	err := row.Scan(&dbID)
+	require.NoError(t, err)
```
srenatus: [nit] This is fine, but we could also do this in one line: `require.NoError(t, row.Scan(&dbID))`

components/authz-service/storage/v2/postgres/postgres.go
srenatus: Some bits strike me as a bit odd:
1. we explicitly convert whatever is returned into an _array_ in the SQL call, so that only one row is returned -- the array. This is a bit odd, I'd expect this to just return the `TableName`s as multiple rows, if there's multiple of them.
2. if we only care about _any of them existing_ (`len(associations) > 0`), we might as well build the query for just that, instead of shuffling around a `[]string` we're not interested in. 

(2.) with the change of (1.) could become something using `EXISTS`, I suppose. 🤔 Alternatively, we could perhaps use `ExecContext` (instead of `QueryRowContext`) and check `(Result).RowsAffected`? (Not sure if that second proposal works, though!)

srenatus:
💁‍♂ just a reminder: `QueryRowContext` does not error out if there's more than one row returned -- it'll just give you the first one.

components/authz-service/storage/v2/postgres/postgres.go
srenatus: :+1: I think a trigger is a good idea here. Forbid adding to `iam_staged_rule_conditions` if a rule with that ID exists in either one of the two tables 🤔

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
```diff
@@ -0,0 +1,76 @@
+BEGIN;
+
+CREATE TABLE iam_staged_project_rules (
+  db_id SERIAL PRIMARY KEY,
+  id TEXT NOT NULL UNIQUE,
+  project_id TEXT REFERENCES iam_projects ON DELETE CASCADE,
+  name TEXT NOT NULL,
+  type TEXT NOT NULL,
+  deleted BOOLEAN NOT NULL DEFAULT FALSE
+);
```
srenatus: I know we've discussed this to a large extent, and I'm a jerk for bringing this up now -- but I need the DDL to _think RDBMS_ 😅 

1. These two tables are basically copies of the other ones, right? With an extra field.

   I think that means [we could use something like this](https://dba.stackexchange.com/a/207103):
   ```sql
   CREATE TABLE iam_staged_project_rules ( 
     deleted BOOLEAN NOT NULL DEFAULT FALSE, 
     LIKE iam_project_rules INCLUDING ALL
   );
   ```
2. We didn't want to add a column with some enum -- like `deleted, staged, active` -- to the other ones... would anyone mind iterating the reason for that for me once more? 😉

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -962,19 +962,33 @@ func (p *pg) CreateRule(ctx context.Context, rule *v2.Rule) (*v2.Rule, error) {
...
 	var ruleDbID string
 	if err := row.Scan(&ruleDbID); err != nil {
 		return nil, p.processError(err)
 	}
 
 	for _, condition := range rule.Conditions {
 		_, err := tx.ExecContext(ctx,
-			`INSERT INTO iam_rule_conditions (rule_db_id, value, attribute, operator) VALUES ($1, $2, $3, $4);`,
-			ruleDbID, pq.Array(condition.Value), condition.Attribute.String(), condition.Operator.String(),
+			`INSERT INTO iam_staged_rule_conditions (rule_db_id, value, attribute, operator, deleted) VALUES ($1, $2, $3, $4, $5);`,
```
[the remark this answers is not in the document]
blakestier: We don't need it, but I kind of liked the extra transparency. Would you consider it safer/cleaner to omit it?

components/authz-service/storage/v2/postgres/postgres.go
blakestier: For context, at this juncture, we only care about whether it already exists (for `Create`). We will, however, have to look at the returned values for calls such as `Update` (if it exists in the `applied` table only, we'll create a new row in `staged` . . . if it exists in `staged`, we'll update that row.) So we'll eventually be interested in the `[]string`. We don't need it here, really, though.

Totally open to better ideas on how to do ⬆️ though. (An array seemed like a better data structure for this purpose but perhaps we should just be examining the rows returned?)

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
blakestier: Oh, yes, #1 is nicer. I think we were underwater considering whether that would copy data, too. 👍 

I'll draw up a document outlining pros/cons of our choices and what seems to have led us here. I don't quite know what you mean by " to the other ones," though. Could you explain?

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
```diff
@@ -0,0 +1,76 @@
...
+      r.type,
+      false as deleted,
+      json_agg(rc) AS conditions
+      FROM iam_project_rules AS r
+      LEFT OUTER JOIN iam_rule_conditions
+      AS rc ON rc.rule_db_id=r.db_id
+      WHERE id=_rule_id AND projects_match_for_rule(project_id, _project_filter)
+      AND
+        NOT EXISTS (SELECT 1 FROM iam_staged_project_rules WHERE id=_rule_id)
+      GROUP BY r.id, r.project_id, r.name, r.type, deleted
```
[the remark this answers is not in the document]
blakestier: Much of this SQL is ported from the presently existing table, I confess.

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
tylercloke: Ooooh didn't know about the `LIKE` keyword. That's awesome.

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
```diff
@@ -0,0 +1,76 @@
...
+  deleted BOOLEAN NOT NULL DEFAULT FALSE
+);
+
+CREATE TABLE iam_staged_rule_conditions (
+  db_id SERIAL PRIMARY KEY,
+  rule_db_id INTEGER REFERENCES iam_staged_project_rules ON DELETE CASCADE,
+  value TEXT[] NOT NULL,
+  attribute TEXT NOT NULL,
+  operator TEXT NOT NULL,
+  deleted BOOLEAN NOT NULL DEFAULT FALSE
```
tylercloke: I imagine we could use `LIKE` here as well!

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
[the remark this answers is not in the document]
tylercloke: Any fields not in an [aggregate function](https://www.postgresql.org/docs/9.5/tutorial-agg.html) but be included in the `GROUP BY` clause.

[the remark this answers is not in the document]
msorens: s/but/must/ in @tylercloke 's comment. ^^

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
blakestier: Ok, turns out that while the `LIKE` is prettier, the tables have a few distinctions which make maintaining the foreign key constraints a little less straightforward. 

`iam_staged_project_rules` references `iam_projects`, while `iam_staged_rule_conditions` references `iam_staged_project_rules`. We don't yet need a staging table for projects (though we might), so I think it's more straightforward to leave it like this.

components/authz-service/storage/v2/postgres/postgres.go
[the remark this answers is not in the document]
srenatus: It's fine as-is 😄

components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql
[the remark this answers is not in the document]
srenatus: Really? :wink:

The error message, if I drop the `GROUP BY`, is slightly different:

```
ERROR:  column "r.project_id" must appear in the GROUP BY clause or be used in an aggregate function
```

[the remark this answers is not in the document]
srenatus: Is it just me? Proliferating these kinds of queries in our code base makes me slightly unhappy.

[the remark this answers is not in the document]
srenatus: Spend some quality time with `psql` and the PG docs. 🎉 Let's do something like this in the future:
```sql
SELECT json_agg(json_build_object(
  'id', r.id,
  'project_id', r.project_id,
  'name', r.name,
  'type', r.type,
  'deleted', r.deleted,
  'conditions', json_build_array(
    json_build_object(
      'value', rc.value,
      'operator', rc.operator,
      'attribute', rc.attribute
    )
  )
)) AS rules
FROM iam_staged_project_rules AS r
INNER JOIN iam_staged_rule_conditions AS rc ON rc.rule_db_id = r.db_id
```
The output, with some example data, is

```
                                                                                                                                                                                                                                                                                                                              rules                                                                                                                                                                                                                                                                                                                               
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 [{"id" : "foo-rule", "project_id" : "foo-project", "name" : "my foo rule", "type" : "node", "deleted" : false, "conditions" : [{"value" : ["prod","staging"], "operator" : "member-of", "attribute" : "chef-server"}]}, {"id" : "bar-rule", "project_id" : "foo-project", "name" : "my foo rule", "type" : "node", "deleted" : false, "conditions" : [{"value" : ["prod","staging"], "operator" : "member-of", "attribute" : "chef-server"}]}, {"id" : "bar-rule", "project_id" : "foo-project", "name" : "my foo rule", "type" : "node", "deleted" : false, "conditions" : [{"value" : ["foo","bar"], "operator" : "member-of", "attribute" : "chef-server"}]}]
(1 row)

```

<details><summary>pretty-printed</summary>

```
[
  {
    "id": "foo-rule",
    "project_id": "foo-project",
    "name": "my foo rule",
    "type": "node",
    "deleted": false,
    "conditions": [
      {
        "value": [
          "prod",
          "staging"
        ],
        "operator": "member-of",
        "attribute": "chef-server"
      }
    ]
  },
  {
    "id": "bar-rule",
    "project_id": "foo-project",
    "name": "my foo rule",
    "type": "node",
    "deleted": false,
    "conditions": [
      {
        "value": [
          "prod",
          "staging"
        ],
        "operator": "member-of",
        "attribute": "chef-server"
      }
    ]
  },
  {
    "id": "bar-rule",
    "project_id": "foo-project",
    "name": "my foo rule",
    "type": "node",
    "deleted": false,
    "conditions": [
      {
        "value": [
          "foo",
          "bar"
        ],
        "operator": "member-of",
        "attribute": "chef-server"
      }
    ]
  }
]
```
</details>

Now, I'd claim this to be easier to understand. 😅

components/authz-service/storage/v2/postgres/postgres.go
srenatus: Ah, I see. Thanks for the context. It's fine I think. 👍

components/authz-service/server/v2/project.go
```diff
@@ -367,14 +367,17 @@ func (s *state) UpdateRule(ctx context.Context, req *api.UpdateRuleReq) (*api.Up
...
 	if err != nil {
 		if err == storage_errors.ErrNotFound {
 			return nil, status.Errorf(codes.NotFound, "could not find rule with ID %q", req.Id)
 		}
 		return nil, status.Errorf(codes.Internal,
 			"error retrieving rule with ID %q: %s", req.Id, err.Error())
 	}
+	if resp.Deleted {
+		return nil, status.Errorf(codes.NotFound, "could not find rule with ID %q", req.Id)
+	}
```
tylercloke: Interesting. So we just give them a 404 if the rule is staged to be deleted? Is this correct? What is list going to do with a rule that is staged to be deleted?

bcmdarroch: This behavior comes from our interpretation of [one of the Staging Changes requirements](https://chefio.atlassian.net/browse/A2-591):

> If multiple users are staging changes, each user should be able to see the changes of the other user

users should always see the most recent change to any given rule. when fetching a rule, we return the staged version if it exists in the staged table, even though the staged change hasn't been applied yet. so if the staged version of the rule is marked deleted, we treat it as if it's already been deleted.

msorens: Presuming that `resp.Deleted` will only be true if the item is marked deleted in the staging table, I would suggest make the message more transparent: "rule with ID %q is staged for deletion".

blakestier: @msorens At first I was onboard with that suggestion, but I think it might create some additional confusion for users at this junction given that while _we_ know it's staged for deletion, they don't need to since 1) It's effectively deleted-- there's nothing they can do about it and 2) there will be no way for them to distinguish staged vs 'applied' in any other context (updated/new, etc). 

In the future, if we do begin to distinguish, I think your error msg makes lots of sense.

PR #590, opened 2019-06-13
[A2-671] Default token's active status to true if not specificed on create or update for V2 API

tylercloke:

If `active` field not passed to token create or update for v2 API, default to true.

Defaults to true if `active` not set:

```
# curl -kH "api-token: $TOK" -XPOST -d '{"id": "test2", "name": "test2"}' https://a2-dev.test/apis/iam/v2beta/tokens
{"token":{"id":"test2","name":"test2","value":"token","active":true,"created_at":"2019-06-13T20:29:59Z","updated_at":"2019-06-13T20:29:59Z","projects":[]}}
```

Can set to false or true explicitly:

```
# curl -kH "api-token: $TOK" -XPOST -d '{"id": "test3", "name": "test3", "active":false}' https://a2-dev.test/apis/iam/v2beta/tokens
{"token":{"id":"test3","name":"test3","value":"P8pPfmlcZI_NwOvqtRozboH2RRI=","active":false,"created_at":"2019-06-13T20:32:11Z","updated_at":"2019-06-13T20:32:11Z","projects":[]}}
```

```
curl -kH "api-token: $TOK" -XPOST -d '{"id": "test4", "name": "test4", "active":true}' https://a2-dev.test/apis/iam/v2beta/tokens
{"token":{"id":"test4","name":"test4","value":"qgbi6ofhrxsVO41K53Vf5OAm6ZE=","active":true,"created_at":"2019-06-13T20:32:37Z","updated_at":"2019-06-13T20:32:37Z","projects":[]}
```

- [ ] Code actually executed?


components/automate-gateway/api/iam/v2beta/request/tokens.proto
```diff
@@ -3,10 +3,12 @@ syntax = "proto3";
 package chef.automate.api.iam.v2beta;
 option go_package = "github.com/chef/automate/components/automate-gateway/api/iam/v2beta/request";
 
+import "google/protobuf/wrappers.proto";
+
 message CreateTokenReq {
     string id = 1;
     string name = 2;
-    bool active = 3;
+    google.protobuf.BoolValue active = 3;
```
tylercloke: Reviewer note: in general, it's bad to change existing protobuf values. In this case, I think we should because 1) it doesn't actually change the type for the external HTTP API because BoolValue accepts JSON true and false and 2) we are in beta still for these APIs so changing them is ok for now.

components/automate-gateway/handler/iam/v2beta/tokens/tokens.go
```diff
@@ -67,10 +74,18 @@ func (s *Server) GetToken(
 // UpdateToken uses a token ID to update the Name/Active fields only
 func (s *Server) UpdateToken(
 	ctx context.Context, req *pb_req.UpdateTokenReq) (*pb_resp.UpdateTokenResp, error) {
+	// If active value not specified, default to True.
```
msorens: Or should we leave the existing value alone if not specified? Not sure which is preferable.

components/automate-gateway/handler/iam/v2beta/tokens/tokens.go
```diff
@@ -23,13 +23,20 @@ func NewServer(client authn.TokensMgmtClient) *Server {
 func (s *Server) CreateToken(
 	ctx context.Context, in *pb_req.CreateTokenReq) (*pb_resp.CreateTokenResp, error) {
 
+	// If active value not specified, default to True.
+	var active bool
+	if in.Active == nil {
+		active = true
+	} else {
+		active = in.Active.GetValue()
+	}
```
msorens: Consider more concise code:
```
active := true
if in.Active != nil {
    active = in.Active.GetValue()
}
```

components/automate-gateway/handler/iam/v2beta/tokens/tokens.go
tylercloke: The previous behavior was that it was defaulting to `false` if the request did not contain the `active` field, which we almost certainly don't want. We should either default it to `true` like create, allow partial updates (which none of our other APIs do and would require changes to the domain), or return a 400 if they didn't pass the active field on update, but that's much more breaking change. Given all of that, seemed like defaulting to true was the best option!

components/automate-gateway/handler/iam/v2beta/tokens/tokens.go
srenatus: Consider more concise code:
```go
active := in.Active == nil || in.Active.GetValue()
```
just kidding. what we have is fine 👍

components/automate-cli/cmd/chef-automate/iam.go
```diff
@@ -420,7 +421,7 @@ func runCreateTokenCmd(cmd *cobra.Command, args []string) error {
 	tokenResp, err := apiClient.TokensV2Client().CreateToken(ctx, &policies_req.CreateTokenReq{
 		Id:     id,
 		Name:   name,
-		Active: true,
+		Active: &wrappers.BoolValue{Value: true},
```
srenatus: [nit] it defaults to _active_, so, we might as well go without this line
```suggestion
```
but being more explicit is fine. LGTM.

tylercloke: Smart.

components/automate-gateway/handler/iam/v2beta/tokens/tokens.go
msorens: Actually, I kind of like the single-line. 😀

PR #595, opened 2019-06-14
[compliance] Use the "cereal" library for scan jobs

stevendanna:

To facilitate multi-node operations, this moves us to using the cereal library to run compliance scan jobs.  With this change, scan job tasks will 

1) Run across multiple compliance-service instances talking to the same database, and
2) Safe to resume after restart

To do this, the scan job process, has to be re-formulated into a set of tasks.  In doing this, we've attempted to change as little of the core scan jobs logics as possible.

Currently, we are not yet using the built-in scheduler in the cereal library.  We've done this to avoid a data migration for now.  Rather we rely on the fact that the library prevents two instances of the same workflow from running at once.  Thus, even if two nodes try to independently schedule a job only one should win.  Since the scheduled_time of the workflow is modified as part of the workflow, we think that this is also safe against cases that would otherwise result in duplicate scans.  There is still a window where we could end up with a duplicate job:

1. Thread A gets Job 1
2. Thread B gets Job 1
3. Thread A pushes and finishes Job 1
4. Thread B pushes and finished Job 1

But the existing code already has a number of bugs where a job can get run twice so we are defering working on this for the moment.

### :+1: Definition of Done

We should not merge this until we are comfortable that the workflow library is ready for production use.

When done, the compliance service should be able to:

- Run multiple instances against the same database without a scheduled scan being run more than expected.
- Run multiple instances against the same database and have the inspec scan work spread out across all instances.

- [ ] Code actually executed?


components/compliance-service/api/jobs/server/server.go
```diff
@@ -89,19 +78,8 @@ func (srv *Server) getComplianceAndSecretsConnection(connectionFactory *secureco
...
-	}
-	// get secrets client with secrets conn
-	secretsClient := secrets.NewSecretsServiceClient(secretsConn)
-	if secretsClient == nil {
-		logrus.Errorf("getComplianceAndSecretsConnection, could not obtain secrets service client: %s", err)
-		return
-	}
-	srv.schedulerServer = scheduler.New(mgrClient, nodesClient, db, ingestClient, secretsClient, remoteInspecVer)
+	scanner := scanner.New(mgrClient, nodesClient, db)
+	srv.schedulerServer = scheduler.New(scanner, workflowManager)
```
stevendanna: If we move some code around in startup, I think we can further simplify this by passing in the scheduler directly.

components/compliance-service/compliance.go
```diff
@@ -444,22 +439,18 @@ func setup(ctx context.Context, connFactory *secureconn.Factory, conf config.Com
...
 	// this function will sleep for one minute, then query the db for all jobs
 	// with recurrence and check if it's time to run the job
 	go schedulerServer.PollForJobs(ctx)
 
-	// start the InspecAgent workers.  the workers represent the amount of goroutines
-	// available to execute a scan.  the buffer size represents the maximum amount of jobs
-	// that may get backed up in the agent queue
-	logrus.Infof("compliance service initializing %d job workers with %d job buffer size",
-		conf.InspecAgent.JobWorkers, conf.InspecAgent.JobBufferSize)
-	runnerServer.SetWorkers(conf.InspecAgent.JobWorkers, conf.InspecAgent.JobBufferSize)
```
stevendanna: fwiw, We think this was probably a no-op as it was setting  the worker count on the wrong runnerServer.

components/compliance-service/compliance.go
```diff
@@ -517,7 +491,13 @@ func Serve(conf config.Compliance, grpcBinding string) error {
 		return err
 	}
 	SERVICE_STATE = serviceStateStarting
```
stevendanna: I _think_ we can probably get rid of SERVICE_STATE as well, but the startup is complicated enough that I need to put some more thought into it.

components/compliance-service/inspec-agent/resolver/resolver.go
```diff
@@ -582,6 +582,12 @@ func (r *Resolver) resolveDynamicJob(ctx context.Context, job *jobs.Job) ([]*typ
 		}
 		jobs = append(jobs, agentJobs...)
 	}
+
+	// Keeping the child - parent job reference alive in the resolved jobs array
+	for _, j := range jobs {
+		j.ParentJobID = job.Id
+	}
```
stevendanna: We can technically remove this again now that resolve-job happens as part of the workflow, but we figured that it is nice to keep the parent ID around for logging and debugging.

components/compliance-service/inspec-agent/runner/runner.go
```diff
@@ -0,0 +1,661 @@
...
+	})
+}
+
+func (p *ScanJobWorkflow) OnTaskComplete(w workflow.WorkflowInstance,
+	ev workflow.TaskCompleteEvent) workflow.Decision {
+
+	var payload ScanJobWorkflowPayload
+
+	if err := w.GetPayload(&payload); err != nil {
+		logrus.WithError(err).Fatal("Could not decode payload")
```
stevendanna: ⚠️ We don't want to use fatal here.  

I think we want to extend the workflow library to be able to do `return w.Fail(err)`

components/compliance-service/inspec-agent/runner/runner.go
```diff
@@ -0,0 +1,661 @@
...
+		for _, job := range jobs {
+			logrus.Infof("Enqueueing individual scan job %q for %q (child of %q)", job.JobID, payload.ChildJobID, payload.ParentJobID)
+			w.EnqueueTask("scan-job", job)
+		}
+
+		payload.OutstandingJobs = len(jobs)
+		return w.Continue(&payload)
+	case "scan-job":
+		payload.OutstandingJobs--
+
```
stevendanna: We should probably be checking ev.Result.Err() here as well

components/compliance-service/inspec-agent/runner/worker.go
```diff
@@ -1,348 +0,0 @@
-package runner
```
stevendanna: most of this code is still around in a new form over in runner.go

components/compliance-service/inspec-agent/runner/runner.go
```diff
@@ -0,0 +1,712 @@
...
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	job.NodeStatus = types.StatusCompleted
+	return osInfo, nil
+}
+
+func doExec(job *types.InspecJob) (jsonBytes []byte, err *inspec.Error) {
```
jaym: this is a question for the compliance folks: do we want to wire through cancellation? it looks like the run function in inspec/cli.go already uses a context, so it would require threading our context to that function. Doing so would mean that when cancelling a workflow, all the inspec processes get killed.

jaym:
I think we'll need a sane way for people to cancel running scan jobs as restarting services won't do it anymore

components/compliance-service/inspec-agent/runner/runner.go
vjeffrey: "do we want to wire through cancellation" -- can you define what you mean by cancellation? like, as a user, how does one "cancel"?

jaym: I think there are 2 parts.
1.) how does a user cancel
2.) Once a user cancels, what do we do. Assuming a user had some way to cancel, currently running scan scans (inspec is running) would wait to finish. If we thread the context to the doExec function, we can kill the inspec process when the user cancels as the context will automatically get canceled in that case

vjeffrey:
a few things:

in today's world, if a user starts a scan job that is scheduled to run agains 100 nodes, and then chooses they do not want to run the scan job anymore about halfway through execution, and delete the scan job, we would allow all current inspec processes to complete but prevent any more inspec processes from being triggered. in other words, if 40 nodes have already executed inspec, 10 more nodes are in the process of being executing inspec, and 50 more have not started to execute inspec, we would end up with 50 node results from that scan job.

if we introduce a cancel functionality that stops currently running inspec processes, that should probably be a new "button"/action?  cc @jonong1972

jonong1972:
Yup, looking at the new designs, i have a place for this.  But for existing designs, it would be just another button on the list row.

vjeffrey:
my next question for everyone is, if we support a cancel functionality/api/button, should that be a new auth "action"? https://github.com/chef/automate/blob/master/components/automate-gateway/api/compliance/scanner/jobs/jobs.proto#L65 
as in, does this fall under the delete action or should it be a new "cancel" action.

jaym:
i would expect that if i can start a scan job, i should be able to cancel. Could it be part of that permission?

jaym:
> a few things:
> 
> in today's world, if a user starts a scan job that is scheduled to run agains 100 nodes, and then chooses they do not want to run the scan job anymore about halfway through execution, and delete the scan job, we would allow all current inspec processes to complete but prevent any more inspec processes from being triggered. in other words, if 40 nodes have already executed inspec, 10 more nodes are in the process of being executing inspec, and 50 more have not started to execute inspec, we would end up with 50 node results from that scan job.
> 
> if we introduce a cancel functionality that stops currently running inspec processes, that should probably be a new "button"/action? cc @jonong1972

Also, i don't think we need to block on a UI redesign if there is a way to stop things that are running. Once we get the new UI going, it should all be fairly easy to wire up

vjeffrey:
"i would expect that if i can start a scan job, i should be able to cancel. Could it be part of that permission?"  you mean the create action? we'll have to coordinate with auth/ @susanev on this

susanev:
hi folks, @jonong1972 @vjeffrey @jaym could someone schedule a meeting for next week for us all to get on the same page? id also like some clear descriptions of delete and cancel from a user's perspective. that should help us identify if there may be users with cancel permission but not delete permission or vice versa 🤷‍♀.

also, are create and start the same thing? or are those separate actions?

vjeffrey:
pulled this down locally to test it out.
i rebuilt compliance and nodemanager service
started a bunch of jobs and watched them complete
started a new scan job and the restarted services immediately.
a bit after restart, i saw `[2019-08-12T18:48:57+00:00] compliance-service.default(O): time="2019-08-12T18:48:57Z" level=error msg="scan-job failed with abnormal error" error="task lost before reporting its status"` in the logs
the job was set to failed, but no error was reported in the ui:
Screen Shot 2019-08-12 at 12 51 57

so it looks that error msg is not being propagated all the way through..

vjeffrey:
it also might be useful if that "scan-job failed with abnormal error" error msg included some reference to the scan job id/other identifying information

vjeffrey:
tested a second time to ensure it wasn't a fluke and same deal - failed scan job with no error msgs

components/compliance-service/inspec-agent/runner/runner.go
```diff
@@ -476,6 +499,23 @@ func (t *InspecJobSummaryTask) Run(ctx context.Context, task cereal.Task) (inter
...
 
+	for _, jobID := range jobsPayload.LostJobIDs {
+		endTime := time.Now()
+		startTime := jobsPayload.StartTime
+		if startTime.IsZero() {
+			startTime = endTime
+		}
+		job := &types.InspecJob{
+			InspecBaseJob: types.InspecBaseJob{
+				JobID: jobID,
```
vjeffrey: Can the node id be included as well?

not including it results in all rows with this error opening at once; see screenshot:
Screen Shot 2019-08-12 at 21 19 40

stevendanna:
Are we sure we want to fail the job on process restart or should we handle the lost worker case in the workflow and re-enqueue the job?

vjeffrey:
> Are we sure we want to fail the job on process restart or should we handle the lost worker case in the workflow and re-enqueue the job?

if it's possible, sure!

jaym:
i think we should hold off on handling lost workers. It'll add more complexity as we'd have to deal with backoff and keeping track of retries to make sure we dont run forever

PR #598, opened 2019-06-14
 Allow workflows to indicate failure 

stevendanna:

This allows you do to:

```
return w.Fail(err)
```

inside a workflow executor callback, which is useful because 
of all the error checking that has to happen around the serialization
and deserialization of the payload and task results.

While here, I cleaned up some SQL that was using various in_val tables
that I don't think are necessary.

### DIFF
--- a/lib/workflow/postgres/postgres.go
+++ b/lib/workflow/postgres/postgres.go
@@ -26,6 +26,7 @@ const (
 	enqueueWorkflowQuery  = `SELECT enqueue_workflow($1, $2, $3)`
 	dequeueWorkflowQuery  = `SELECT * FROM dequeue_workflow(VARIADIC $1)`
 	completeWorkflowQuery = `SELECT complete_workflow($1, $2)`
+	failWorkflowQuery     = `SELECT fail_workflow($1, $2)`
 	continueWorkflowQuery = `SELECT continue_workflow($1, $2, $3, $4, $5)`
 	abandonWorkflowQuery  = `SELECT abandon_workflow($1, $2, $3)`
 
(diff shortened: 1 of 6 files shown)


lib/workflow/postgres/postgres.go
```diff
@@ -684,6 +706,18 @@ func (workc *PostgresWorkflowCompleter) Done(result []byte) error {
 	return errors.Wrapf(workc.tx.Commit(), "failed to mark workflow %d as complete", workc.wid)
 }
 
+func (workc *PostgresWorkflowCompleter) Fail(error error) error {
```
stevendanna:
```suggestion
func (workc *PostgresWorkflowCompleter) Fail(err error) error {
```

Want to avoid using `error` as a variable name I think.

lib/workflow/postgres/postgres.go
```diff
@@ -153,6 +154,22 @@ func (pg *PostgresBackend) Init() error {
 	}
 	// defer m.Close() I don't think we want to call close here because it'll close our db instance
 
+	version, dirty, err := m.Version()
+	if err != nil && err != migrate.ErrNilVersion {
+		return errors.Wrap(err, "init migrator - error getting migration version")
+	}
+
+	if dirty {
```
srenatus: That's #597, isn't it?

stevendanna: Yup, I'll rebase this once that merges.  I have it on this branch because it is not fun to iterate on schema changes without it.

lib/workflow/workflow.go
```diff
@@ -735,32 +755,43 @@ func (m *WorkflowManager) processWorkflow(ctx context.Context, workflowNames []s
...
-
-	if decision.complete {
+	logctx := logrus.WithFields(logrus.Fields{
+		"enqueued_tasks":  wevt.EnqueuedTaskCount,
+		"completed_tasks": wevt.CompletedTaskCount,
+	})
+	if decision.failed {
+		s.Begin("failed")
+		if wevt.CompletedTaskCount != wevt.EnqueuedTaskCount {
+			logctx.Info("Abandoning workflow because it has pending tasks")
```
jaym: we might still want to log `decision.err` here

jaym:
There's some weirdness around failing when we end up abandoning the workflow. However, I kind of think we won't need abandoning for much longer, so it's probably fine

jaym:
Oh, if you could add a quick integration test to make sure we can get back the error for a failed workflow, that'd be cool

PR #604, opened 2019-06-14
Proceed on update version conflict for compliance reports

alexpop:
I created two recurrent scan jobs with the same schedule. They targetted the same nodes. This caused ingestion on a number of scan jobs to conflict in ElasticSearch when doing the update in the `setDailyLatestToFalse` function.

```
Report processing error: rpc error: code = Internal desc = elastic: Error 409 (Conflict) elastic: Error 409 (Conflict)
```

Screen Shot 2019-06-14 at 8 16 59 PM

The report would be ingested, but because of the update failure, the scan job is failed and no link to see the results from the scan jobs page.

We have two options here:
1. Use ProceedOnVersionConflict to abort the update but not fail the request.
2. Implement a retry logic in go for the NewUpdateByQueryService call.

The PR implements `1.` and testing shows the scan jobs completing.

### DIFF
--- a/components/compliance-service/ingest/ingestic/ingestic.go
+++ b/components/compliance-service/ingest/ingestic/ingestic.go
@@ -1,6 +1,7 @@
 package ingestic
 
 import (
+	"math/rand"
 	"time"
 
 	"fmt"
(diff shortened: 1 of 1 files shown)


phiggins:
I tried to look up what `ProceedOnVersionConflict` and I'm more confused now.

```
 ProceedOnVersionConflict aborts the request on version conflicts. It is an alias to setting Conflicts("proceed"). 
```

... but then:

```
Conflicts indicates what to do when the process detects version conflicts. Possible values are "proceed" and "abort". 
```

Maybe this is an error in the docs, but the options are to abort the request, or to proceed which aborts the request?

alexpop:
Tested this change with two scan jobs running at the same time for the same nodes. There were no more failures of the scan jobs, but the `latest_report` field was not being updated all the time and in some cases it would be set to false for all documents. Talked with Rick and decided to close this for now since it's a corner case that no users have reported. Will revisit and look for a solution if needed in the future.

phiggins:
I wonder if this is related to #613. 🤔

PR #605, opened 2019-06-14
Cover more ui cases with studio helper

danielsdeleo:

Updates the `applications_populate_database` studio function to trigger a few more UI behaviors we have added since it was first written:
* Create enough service groups so that pagination happens
* Create service groups with no "application," this shows up with a special badge
* Create service groups with no "environment," this shows up with a special badge
* Create service groups with no "channel," this shows up with a special badge (on the right side detail panel)

1. `start_all_services` in the studio
2. `source .studiorc` to load the new studio helper code
3. `applications_populate_database`
4. Visit the UI at a2-dev.test, you should have more than 25 service groups and service groups that are missing application, environment, and channel (different service group for each case)

- [x] Code actually executed?

### DIFF
--- a/.studio/applications-service
+++ b/.studio/applications-service
@@ -316,6 +316,32 @@ function applications_populate_database() {
       --origin custom --version "1.9.2" --release "20190115181111" --health 0\
       --sup-id "$(uuidgen)"
   done
+
+  # We have a custom visual treatment when application, environment, or channel
+  # are blank. Create one of each
+  applications_publish_supervisor_message --name "blank-app" --group "default"\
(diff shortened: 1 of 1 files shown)


apriofrost:
Thanks for adding more use case examples! 
I wonder if we can add another use where there are more than two types of health status for a service group. For example, have at least one group that has services in critical, warning, and OK status. 

Asking this because I'd like to test if users would expect to see the service group be included when filtered by "warning" status in the primary filter above the table - to see if our definition of "group status" resonates well with users.

danielsdeleo:
@apriofrost k, I added one service group with all four kinds of status and one that has ok, unknown, and warning.

PR #606, opened 2019-06-14
WIP: Case Insensitive wildcard searching

lancewf:

- [ ] Code actually executed?


components/compliance-service/reporting/relaxting/nodes.go
```diff
@@ -70,8 +70,8 @@ func (backend *ES2Backend) GetNodes(from int32, size int32, filters map[string][
 		Size(int(size))
 
 	// Adding a second sort field when we don't sort by node name
-	if sortField != "node_name.lower" {
-		searchSource.Sort("node_name.lower", true)
```
lancewf: This makes the request fail. I am assuming the problem is that with the changes in the mapping there is no `.lower` because all values are lower. I would have assumed this would not be a problem.

PR #609, opened 2019-06-17
Minor UI package maintenance

msorens:

Starting to think about an Angular upgrade (Angular 8 just came out and we are still on Angular 6).
Ran into various issues when attempting to upgrade that need more investigation, but this PR is at least a start to tidy up a couple minor things.

Upon upgrading `tslint` a few new linting errors appeared, and this PR resolves them.

1. Error: _Tried to lint *file*.js but found no valid, enabled rules for this file type and file path in the resolved configuration._ The fix for that is to add a `jsRules` section to the tslint.json file, as described here: https://github.com/palantir/tslint/issues/3735

2. The recommended content of the `jsRule` section from that suggested fix--to turn on the "no empty blocks" rule caused, a few instances of, yep (`block is empty`). It turns out that the content of that `jsRules` for the first fix is completely arbitrary; _any_ rule will do, so I changed that to something innocuous (`no-debugger`), and that resolved this second issue.

3. Error: _no-use-before-declare is deprecated._ This refers to a linting rule in tslint.json that is both expensive to compute and not at all needed in a ts application (would be in a plain js application). So just deleting that line from the config in tslint.json was the fix here. Reference: https://palantir.github.io/tslint/rules/no-use-before-declare/

### :+1: Definition of Done

1. Sass builds complete without error.
2. linting in automate-ui and in chef-ui-library returns no errors.

- [x] Code actually executed?

### DIFF
--- a/components/automate-ui/package-lock.json
+++ b/components/automate-ui/package-lock.json
@@ -1487,6 +1487,34 @@
         "tslib": "1.9.3"
       }
     },
+    "@babel/code-frame": {
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
+      "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
+      "dev": true,
+      "requires": {
+        "@babel/highlight": "7.0.0"
+      }
+    },
+    "@babel/highlight": {
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/highlight/-/highlight-7.0.0.tgz",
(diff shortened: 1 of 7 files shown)


components/release-mgmt-dashboard/tslint.json
```diff
@@ -69,7 +69,6 @@
     "no-trailing-whitespace": true,
     "no-unnecessary-initializer": true,
     "no-unused-expression": true,
-    "no-use-before-declare": true,
```
srenatus: Hrm, I'd have expected either one of:
1. leaving this file alone (i.e. ignoring this component)
2. updating this component's tslint in the same manner as the other ones, and updating this file.

components/automate-ui/package-lock.json
```diff
@@ -10383,79 +10357,12 @@
             "path-is-absolute": "1.0.1"
           }
         },
-        "har-schema": {
```
srenatus: Interesting. I wonder what had dragged this in, and why it's gone now. 🤷‍♂

PR #610, opened 2019-06-17
Add new bash helper for a2 auth resources

msorens:

1. Rename `a2` bash helper to `authQuery`.
2. Add new `authLoad` helper.

The latter makes it very convenient to copy resources from one machine to another for diagnostic purposes.
You can, for example, now do this on your development machine to copy all tokens from an acceptance machine to your own:

```
# Generate a token for the target machine
$ export TARGET_HOST=https://a2-local-inplace-upgrade-acceptance.cd.chef.co
$ export TOK="r7sDLTujkijLLhwe5L4xyRPkCXY="

$ authQuery --one-line tokens > captures/tokens.out

# Reset to local machine
$ export TARGET_HOST=https://a2-dev.test
$ export TOK="r7sDLTujkijLLhwe5L4xyRPkCXY="

$ authLoad tokens captures/tokens.out
```

This will work for IAM v1 tokens, team, users, and policies.
Will likely work for most IAM v2 resources, but have not needed to try those yet.

### :+1: Definition of Done

See above.

- [x] Code actually executed?

### DIFF
--- a/scripts/a2_collections
+++ b/scripts/a2_collections
@@ -41,7 +41,8 @@
 
 authQuery () {
   # prerequisites
-  ([ -z "$TOK" ] || [ -z "$TARGET_HOST" ]) && { echo "Need TOK and TARGET_HOST variables"; return; }
+  local token=${TOK:?Need this env variable to contain ChefAutomate admin token}
+  local host=${TARGET_HOST:?Need this env variable to point to your ChefAutomate host}
   if ! type jq &> /dev/null
   then
     echo "Need jq installed"; return;
@@ -63,19 +64,19 @@ authQuery () {
   done
 
   # generate jq script
-  if [ "$3" != "" ]; then
+  if [[ "$3" != "" ]]; then
     jq_script=".$1[] | select (.$2 == \"$3\")"
-  elif [ "$jq_option" == "-c" ]; then 
+  elif [[ "$jq_option" == "-c" ]]; then 
     jq_script=".$1[]"
   else
     jq_script="."
   fi
 
   # execute
-  if [ "$show_cmd" == 1 ]; then 
-    echo "curl -sSkH \"api-token: \$TOK\" -H \"$projects\" \"$TARGET_HOST/$auth_path/$1\" | jq $jq_option '$jq_script'"
+  if [[ "$show_cmd" == 1 ]]; then 
+    echo "curl -sSkH \"api-token: \$token\" -H \"$projects\" \"$host/$auth_path/$1\" | jq $jq_option '$jq_script'"
   fi
-  curl -sSkH "api-token: $TOK" -H "$projects" "$TARGET_HOST/$auth_path/$1" | jq $jq_option "$jq_script"
+  curl -sSkH "api-token: $token" -H "$projects" "$host/$auth_path/$1" | jq $jq_option "$jq_script"
 }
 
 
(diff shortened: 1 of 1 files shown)


scripts/a2_collections
```diff
@@ -34,7 +39,7 @@
 #     To run on a different environment you can adjust the env vars for a single invocation. Example:
 #     TARGET_HOST=https://a2-local-fresh-install-dev.cd.chef.co TOK=stV8b5NqP6qepD9I23OFn_U2OME= a2 -s policies
 
-a2 () {
+authQuery () {
   # prerequisites
   ([ -z "$TOK" ] || [ -z "$TARGET_HOST" ]) && { echo "Need TOK and TARGET_HOST variables"; return; }
```
srenatus: [nit]
Could go with
```bash
local token=${TOK:?Need TOK variable}
local host=${TARGET_HOST:?Need TARGET_HOST variable}
```
[via posix](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02)

scripts/a2_collections
```diff
@@ -73,3 +78,65 @@ a2 () {
...
+    echo "Need jq installed"; return;
+  fi
+  RESOURCE=$1
+  FILE=$2
+
+  # file-lines-to-array
+  old_IFS=$IFS
+  IFS=$'\n'
+  lines=($(cat $FILE))
+  IFS=$old_IFS
```
srenatus: Interesting construct! I'd have expected something like `while read line; do ...; done <"$FILE"`. But this is fine, too.

scripts/a2_collections
```diff
@@ -1,8 +1,21 @@
 #!/bin/bash
 
+# Prerequisites:
+#    jq preprocessor
```
srenatus: "preprocessor" lol. It's like calling your laptop a "desk calculator". 😉

scripts/a2_collections
```diff
@@ -73,3 +78,65 @@ a2 () {
...
+  then
+    echo "Need jq installed"; return;
+  fi
+  RESOURCE=$1
+  FILE=$2
+
+  # file-lines-to-array
+  old_IFS=$IFS
+  IFS=$'\n'
+  lines=($(cat $FILE))
```
stevendanna: [nit] Double quote to avoid wordsplitting for $FILE:
```suggestion
  lines=($(cat "$FILE"))
```

scripts/a2_collections
```diff
@@ -73,3 +78,65 @@ a2 () {
...
+  IFS=$old_IFS
+
+  for line in "${lines[@]}"
+  do
+    if [ ! -z "$line" ]; then
```
stevendanna: Since we are in bash, might we use `[[` here for consistency?

[info] `! -z` can be replaced by `-n` if you fancy.  I personally always have to look up which is which so I'm not sure that is actually a useful change though :D

scripts/a2_collections
stevendanna: [info] If we wanted to enforce bash v4+, you could do: `readarray -t lines <"$FILE"`

scripts/a2_collections
```diff
@@ -1,8 +1,21 @@
...
+# These works either inside or outside Habitat Studio.
+
+
+#############
+# authQuery
+#############
 # Purpose:
-#     Provide a very concise way to fetch collections of a2 auth entities at the API level.
-#     This works either inside or outside Habitat Studio.
+#     Provide a very concise way to fetch collections of a2 auth resources at the API level.
```
blakestier: Are "auth" resources the same as "IAM" resources? If so, we should be consistent with using one. Given that this is likely to only be used internally, it probably doesn't matter too much, but it gave me a moment's pause.

msorens: I would think they mean the same; what would be an example of an auth resource that is not an IAM resource?

PR #614, opened 2019-06-17
automate-gateway: add ?pretty parameter to every grpc-gateway method

srenatus:
Now, if you query, for example, either one of

    https://a2-dev.test/apis/iam/v2beta/policies?pretty
    https://a2-dev.test/apis/iam/v2beta/policies?pretty=1
    https://a2-dev.test/apis/iam/v2beta/policies?pretty=true
    https://a2-dev.test/apis/iam/v2beta/policies?pretty=please

you'll get a pretty-printed JSON response. 
Any handcrafted handlers are _not_ getting this, only the handlers we get from grpc-gateway.

### TODOs
- [x] update docs where we pipe our output into `jq .`
- [x] add integration test

### DIFF
--- a/components/automate-gateway/gateway/services.go
+++ b/components/automate-gateway/gateway/services.go
@@ -321,6 +321,7 @@ func versionedRESTMux(grpcURI string, dopts []grpc.DialOption, toggles gwRouteFe
 func muxFromRegisterMap(grpcURI string, dopts []grpc.DialOption, localEndpoints map[string]registerFunc) (*runtime.ServeMux, error) {
 	opts := []runtime.ServeMuxOption{
 		runtime.WithIncomingHeaderMatcher(headerMatcher),
+		runtime.WithMarshalerOption("application/json+pretty", &runtime.JSONPb{OrigName: true, EmitDefaults: true, Indent: "  "}),
 		runtime.WithMarshalerOption(runtime.MIMEWildcard, &runtime.JSONPb{OrigName: true, EmitDefaults: true}),
 		runtime.WithMetadata(middleware.CertificatePasser),
 	}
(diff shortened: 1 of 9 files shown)


components/automate-gateway/gateway/services.go
```diff
@@ -321,6 +321,7 @@ func versionedRESTMux(grpcURI string, dopts []grpc.DialOption, toggles gwRouteFe
 func muxFromRegisterMap(grpcURI string, dopts []grpc.DialOption, localEndpoints map[string]registerFunc) (*runtime.ServeMux, error) {
 	opts := []runtime.ServeMuxOption{
 		runtime.WithIncomingHeaderMatcher(headerMatcher),
+		runtime.WithMarshalerOption("application/json+pretty", &runtime.JSONPb{OrigName: true, EmitDefaults: true, Indent: "  "}),
```
msorens: Just wondering what the Indent is for... does that add that much at each nested element?

srenatus:
> Update: with this code in place, any JSON data passed in a curl command does not make it to the back-end. 

Luckily, a few of our tests fail, too. 😄 I'll look into this tomorrow.

srenatus:
@msorens so, as it turns out, `curl` sends a `content-type` header set to `application/x-www-form-urlencoded` if you provide a request body (and no content-type). That in turn was eaten when I've retrieved the query param through `req.ParseForm()` and `req.Form`. So, now, we'll do something more correct: We retrieve `?pretty` through `req.URL.Query()`, disregarding the content-type header.

components/automate-gateway/gateway/services.go
srenatus: Yup! https://godoc.org/github.com/golang/protobuf/jsonpb#Marshaler
```
    // A string to indent each level by. The presence of this field will
    // also cause a space to appear between the field separator and
    // value, and for newlines to be appear between fields and array
    // elements.
    Indent string
```

PR #623, opened 2019-06-18
Assorted compliance-service test fixes

stevendanna:
These fixes are from #595 but are not specific to the changes on that branch.

Signed-off-by: Alex Pop <alexpop@users.noreply.github.com>


alexpop:
yey, they are in master, thank you!

Issue #529, opened 2019-06-06
Notifications for client run failures incorrectly identify failing cookbook when using 'ignore_failure' property.


brywilliamsCF:
# Description
If a client run contains a resource that fails but uses the 'ignore_failure' property, notifications sent for any later client run failures will incorrectly refer to the cookbook containing the first ignored resource and not the cookbook containing the resource that actually stopped the client run and triggered the notification.  Applies to Slack, ServiceNow, and Webhook notifications.

# Automate Version
20190513175357

# Replication Case
1. Create cookbook1 with a resource that will fail but use the 'ignore_failure' property.
2. Create cookbook2 with a resource that will fail.
3. Apply both cookbooks to a node.
4. cookbook1 will fail but the run will continue due to the use of 'ignore_failure'.
5. cookbook2 will fail, ending the run.
6. Automate will send a JSON notification.  The JSON key 'exception_title' will correctly refer to the failed resource from cookbook2, but the key 'cookbook' will be cookbook1.

vjeffrey:
@brywilliamsCF thanks for reporting the issue! I've labeled it as a bug and we will work to get the right people looking at this issue

PR #634, opened 2019-06-19
Notifications chef client runs skip ignored failure

lancewf:

Notifications from failed CCRs should skip resource failures that are marked 'ignore_failure' = true. The bug was found in the gateway CCR parsing code. When trying to parse the boolean flag the error variable was checked incorrectly. 

1. `start_all_services`
1. Create a requestbin URL to send notifications from http://requestbin.sjcmmsn.com/
1. Create a Slack, 'Chef client run failures' notification with the webhook pointing to the URL created in the line above in Automate (https://a2-dev.test/settings/notifications).
1. `JSON_FILE=/src/inspec/a2-api-integration/files/fixtures/converge/converge-failure-report_with_ignored_failure_run_converge.json send_chef_run_example lb`
1. Check the requestbin notification to ensure that there is a "attachments" -> "fields" -> "value" with "insights-ignored::default"'. This is incorrect because the "insights-ignored" failure was marked "ignore_failure" = true. 
1. Now lets apply the bug fix, `rebuild components/automate-gateway` 
1. Send the Same CCR again `JSON_FILE=/src/inspec/a2-api-integration/files/fixtures/converge/converge-failure-report_with_ignored_failure_run_converge.json send_chef_run_example lb`
1. Check the requestbin for the new notification and ensure that there is a "attachments" -> "fields" -> "value" with "insights-real::default"'.

https://github.com/chef/automate/issues/529

- [x] Code actually executed?


components/automate-gateway/gateway/datacollector.go
```diff
@@ -385,7 +385,7 @@ func getBoolIfExists(fieldName string, body []byte) bool {
 	fieldBytes, _, _, err := jsonparser.Get(body, fieldName)
 	if err == nil {
 		b, err := jsonparser.ParseBoolean(fieldBytes)
-		if err != nil {
+		if err == nil {
```
lancewf: This change fixes the bug.

components/automate-gateway/gateway/datacollector_test.go
```diff
@@ -169,7 +169,7 @@ func TestDataCollectorParseBytesToChefRunFailure(t *testing.T) {
 	assert.Equal(t, "", res1.Delta)
 	assert.Equal(t, "insights-test", res1.CookbookName)
 	assert.Equal(t, "0.1.1", res1.CookbookVersion)
-	assert.Equal(t, false, res1.IgnoreFailure.GetBoolValue())
+	assert.Equal(t, true, res1.IgnoreFailure.GetBoolValue())
```
lancewf: This is the only real change to this test. The other tests are the go simplification.

components/ingest-service/examples/converge-failure-report.json
```diff
@@ -3127,7 +3127,7 @@
       "before": {},
       "duration": "0",
       "delta": "",
-      "ignore_failure": false,
+      "ignore_failure": true,
```
lancewf: This allows the above test to check if the valve is converted correctly.

components/automate-gateway/gateway/datacollector.go
vjeffrey: could you add a comment saying why we do the err == nil check here?  i know you included some info about it in the pr, but i think our future selves might appreciate having a comment in the code there

inspec/a2-api-integration/controls/notifications-alert.rb
```diff
@@ -34,6 +34,24 @@
       end
     end
 
+    #converge-failure-with-ignored-failure-report_run_converge
+    let(:failed_ccr_message_with_ignored_failure) do
+      return @failed_ccr_json ||= begin
```
phiggins: Memoizing with an instance variable inside a `let` block is redundant, and it is dangerous in this case since the instance variable will be set by the first `let` to be referenced.

inspec/a2-api-integration/controls/notifications-alert.rb
```diff
@@ -180,6 +198,36 @@
...
+        expect(send_failed_ccr_with_ignored_failure.http_status).to eq(200)
+        sleep (5)
+
+        # Check if the requestbin received the alert
+        request = inspec.http("#{requestbin_url}/#{requestbin_name}")
+        after_request_count = JSON.parse(request.body, symbolize_names: true)[:request_count]
+        expect(after_request_count).to eq(before_request_count + 1)
+
+        # get the last sent notification
+        notification = JSON.parse(request.body, symbolize_names: true)[:requests][before_request_count]
```
phiggins: Nitpick: instead of accessing this by indexing into the array, you could replace `[before_request_count] with `.last`.

lancewf: I always forget how friendly Ruby is with lists.

PR #635, opened 2019-06-19
One more bash helper to generate auth resources from scratch

msorens:

This wraps up the final planned segment of the work started in PR #610 (Add new bash helper for a2 auth resources) by adding the final of a triumvirate of functions useful for exercising an A2 system.

This introduces `authGen` that, unlike `authLoad` from the last PR that requires enumerated data to populate a system, generates data on the fly, thus allowing creating arbitrarily large sets of resources.
This supports both creating and deleting such resources. Here's a simple example:

```
Create 500 policies starting with a seed of 101:
$ authGen policies create 500 101

Those policies need these members:
$ authGen tokens create 500 101

Now, just for illustration, delete the last 100 of those 500 policies:
$ authGen policies delete 100 501
```

### :+1: Definition of Done

See above.

NA

PR #610 

- [x] Code actually executed?

### DIFF
--- a/scripts/a2_collections
+++ b/scripts/a2_collections
@@ -1,12 +1,14 @@
 #!/bin/bash
 
 # Prerequisites:
-#    jq preprocessor
+#    jq post-processor
 #       https://stedolan.github.io/jq/
+#    jo pre-processor (for authGen function)
+#       https://github.com/jpmens/jo
 #    variable $TOK
 #       An admin token obtained from `get_admin_token` or `chef-automate admin-token`
 #    variable $TARGET_HOST
-#       A host running A2, typically `https://a2-dev.test`
+#       A host running A2, either `https://a2-dev.test` or a remote machine.
 #
 # These works either inside or outside Habitat Studio.
 
@@ -36,8 +38,8 @@
 #     a2 -o --show users name bob
 #     a2 --v2 -o --show -p proj1,proj2 roles
 #
-#     To run on a different environment you can adjust the env vars for a single invocation. Example:
-#     TARGET_HOST=https://a2-local-fresh-install-dev.cd.chef.co TOK=stV8b5NqP6qepD9I23OFn_U2OME= a2 -s policies
+# To run on a different environment you can adjust the env vars for a single invocation. Example:
+# TARGET_HOST=https://fresh-install-dev.cd.chef.co TOK=stVpD9I23OFn_U2OME= a2 -s policies
 
 authQuery () {
   # prerequisites
@@ -93,6 +95,9 @@ authQuery () {
 #    -d or --dry-run      show the curl commands but do not execute them
 #    -2 or --v2           use IAM V2 endpoints
 #
+# Example:
+#    authLoad teams captured/teams-from-acceptance.txt
+#
 authLoad() {
   # process options
   dry_run=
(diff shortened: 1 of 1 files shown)


scripts/a2_collections
```diff
@@ -137,3 +142,97 @@ authLoad() {
...
+  COUNT=$3
+  SEED=$4
+  ID_PREFIX=test-$RESOURCE
+  [[ ! $MODE =~ ^(create|delete)$ ]] && { echo "mode must be 'create' or 'delete'"; return; }
+  [[ ! $RESOURCE =~ ^(tokens|users|teams|policies)$ ]] && { echo "resource must be in: tokens, users, teams, policies"; return; }
+
+  if [[ $MODE == create ]]; then
+    echo "Creating $COUNT $RESOURCE..."
+    for (( i = 0; i < $COUNT; i++ )) 
+    do
```
stevendanna: Since this is dev code, I think it is fine as is, but you might consider extracting the body of this for loop into a `create_resource` function. Similarly, a delete_resource function for the loop in the else clause.

scripts/a2_collections
```diff
@@ -137,3 +142,97 @@ authLoad() {
...
+  MODE=$2
+  COUNT=$3
+  SEED=$4
+  ID_PREFIX=test-$RESOURCE
+  [[ ! $MODE =~ ^(create|delete)$ ]] && { echo "mode must be 'create' or 'delete'"; return; }
+  [[ ! $RESOURCE =~ ^(tokens|users|teams|policies)$ ]] && { echo "resource must be in: tokens, users, teams, policies"; return; }
+
+  if [[ $MODE == create ]]; then
+    echo "Creating $COUNT $RESOURCE..."
+    for (( i = 0; i < $COUNT; i++ ))
```
stevendanna: Note, like `i`, I believe the `$` isn't needed in arithmetic expressions. So this could be:

```suggestion
    for (( i = 0; i < COUNT; i++ )) 
```

similar changes could be made in other arithmetic expressions in this script.  Not sure if we have a style preference here.

scripts/a2_collections
```diff
@@ -137,3 +142,97 @@ authLoad() {
...
+    for (( i = 0; i < $COUNT; i++ )) 
+    do
+      local id_index=$(($i+$SEED))
+      local name="$ID_PREFIX $id_index"
+      local id=$ID_PREFIX-$id_index
+      case "$RESOURCE" in
+        tokens)  ;;
+        users)  ;;
+        teams) 
+          local teams=$(curl -sSkH "api-token: $TOK" "$TARGET_HOST/$auth_path/$RESOURCE")
```
stevendanna: This doesn't matter in this script, but we tend to prefer
```suggestion
          local teams
          team=$(curl -sSkH "api-token: $TOK" "$TARGET_HOST/$auth_path/$RESOURCE")
```

Using `local` and a command substitution assignment together means you lose the return value and if you are using `set -e`, you don't get the exit-on-failure behavior you might expect.  See

https://github.com/koalaman/shellcheck/wiki/SC2155

for more info.

msorens:
@srenatus Thank you, and I appreciate your point... however, I look at this as three small functions rather than one big script, put in the same file for convenience.

msorens:
@stevendanna Thanks for making me aware of shellcheck! I cleaned it up further with that.

srenatus:
> @srenatus Thank you, and I appreciate your point... however, I look at this as three small functions rather than one big script, put in the same file for convenience.

Oh I didn't worry about the number of lines -- that's a poor approximation for complexity 😉 

It's difficult to describe the line in the sand where bash stops being an appropriate tool. IMHO, if you start doing arithmetic, you've crossed it.

Note that I'm not arguing that this script wouldn't work, or do its job -- I just think that it introduces a maintainance burden and a tarpit of potentially lost time: getting bash right **is hard**, and the bugs are all subtle. This hightens the burden of entry for people less intimately acquainted to bash, at any rate. 0% of the issues shellcheck raises could ever happen if this was using a real scripting language; or a cute little go executable .

srenatus:
I guess my point can be summed up as _this is fine, let's not do it again_ 😜 (Why does noone follow the iron rule of shell scripts? "for every line you add, you've got to remove one somewhere else")

scripts/a2_collections
```diff
@@ -36,8 +38,8 @@
 #     a2 -o --show users name bob
 #     a2 --v2 -o --show -p proj1,proj2 roles
 #
-#     To run on a different environment you can adjust the env vars for a single invocation. Example:
-#     TARGET_HOST=https://a2-local-fresh-install-dev.cd.chef.co TOK=stV8b5NqP6qepD9I23OFn_U2OME= a2 -s policies
+# To run on a different environment you can adjust the env vars for a single invocation. Example:
+# TARGET_HOST=https://fresh-install-dev.cd.chef.co TOK=stVpD9I23OFn_U2OME= a2 -s policies
```
srenatus: Guess it doesn't matter that we're leaking admin tokens into the world, since the machine is on our VPN. 🤷‍♂

scripts/a2_collections
```diff
@@ -125,15 +130,116 @@ authLoad() {
       # v1 resources work by just re-using output of authQuery except for users, which need a password added
       # TODO: check that all v2 resources work
       if [[ "$RESOURCE" == "users" ]]; then
-        line=$(jq -c '. += {"password": "chefautomate"}' <<< $line)
+        line=$(jq -c '. += {"password": "chefautomate"}' <<< "$line")
```
srenatus: [nit]
```suggestion
        line=$(jq -c '. += { password: "chefautomate" }' <<< "$line")
```

scripts/a2_collections
```diff
@@ -125,15 +130,116 @@ authLoad() {
...
       fi
       if [[ "$show_cmd" == 1 ]] || [[ "$dry_run" == 1 ]]; then 
         echo "curl -sSkH \"api-token: \$token\" \"$host/$auth_path/$RESOURCE\" -d '$line'"
       fi
       if [[ "$dry_run" == "" ]]; then
```
srenatus: 💭 wasn't that `-n`? per @stevendanna's comment somewhere

scripts/a2_collections
```diff
@@ -125,15 +130,116 @@ authLoad() {
...
+# Purpose:
+#     Generate a list of a2 auth resources.
+#     This generates dummy data that is not particularly realistic
+#     but can be used to observe general performance characteristics.
+#     NB: When generating policies, it builds one policy per token,
+#     so you should also generate a set of tokens with the same parameters.
+#     This function also allows you to delete your generated resources just as easily.
+# 
+# Example:
+#     Create 500 policies starting with a seed of 101:
```
srenatus: [nit] I think PRNG when reading "seed", but it's fine.

scripts/a2_collections
```diff
@@ -125,15 +130,116 @@ authLoad() {
...
+  done
+}
+
+function create_resource {
+  local token=$1
+  local host=$2
+  local id_index=$(($3 + SEED))
+  local id=$ID_PREFIX-$id_index
+  local name="$ID_PREFIX $id_index"
+  local json=""
```
srenatus: [nit] I don't believe we need this initialization, i.e. this should do the trick, too:
```suggestion
  local json
```
but it doesn't hurt, I suppose.

PR #636, opened 2019-06-19
lib/platform: enforce timezones in pg connections

srenatus:

We've recently realised that some of our tests depend on the timezone used with PG.

It seems unfortunate to have no control over this. So, this will do two things:

1. for each PG connection made through our platform libs (which should be _any_ in production use), set the timezone for the session to UTC
2. for the copy-pasted auth unit tests using PG, manually fixes all `PG_URL` vars to use `timezone=UTC`.

### :+1: Definition of Done

✅ Tests pass when built with the postgres version that had UTC/UCT messed up. [These are here](https://buildkite.com/chef-oss/chef-automate-master-verify/builds?branch=sr%2Ftest-something-with-timezones). 

NA

https://git.postgresql.org/gitweb/?p=postgresql.git;a=commit;h=e3846a00c2f87402dcedf7f07950ab2d89cf5827

- [X] Code actually executed?


lib/io/fileutils/access_linux_test.go
```diff
@@ -1,4 +1,4 @@
-// +build !windows
+// +build !windows !darwin
```
msorens: i'm puzzled; I thought the same rule applies to test or non-test files. Why did you remove the directive in access_linux.go but kept it here in access_linux_test.go?

components/authn-service/constants/constants.go
```diff
@@ -20,4 +20,4 @@ const UnassignedProjectID = "(unassigned)"
 const AllProjectsExternalID = "*"
 
 // TestPgURL is the URL to connect to the authn test DB
-const TestPgURL = "postgresql://postgres@127.0.0.1:5432/authn_test?sslmode=disable"
+const TestPgURL = "postgresql://postgres@127.0.0.1:5432/authn_test?sslmode=disable&timezone=UTC"
```
msorens: Should this change be applied to docs/examples, too?
Specfically:
components/session-service/examples/config.yml
components/teams-service/examples/config.yml
components/teams-service/README.md

lib/platform/config.go
```diff
@@ -232,9 +232,11 @@ func (c *Config) GetPGConnInfoURI(user string, opts ...ConnInfoOpts) (*PGConnInf
 					user, c.GetPostgresql().GetIp(), c.GetPostgresql().GetCfg().GetPort(), strings.Join(opts, "&"))
```
msorens: Just from pattern matching, you have added timezone awareness in both the format string and the env in L258-267. However, here you have added it only to the env (L328), not the format string (L229). Is there a reason for the difference?

lib/io/fileutils/access_linux_test.go
srenatus: An oversight...

lib/platform/config.go
srenatus: There's no difference. I've added it to `opts` in L196, and it's put into the format string using `strings.Join(opts, "&")`.

components/authn-service/constants/constants.go
srenatus: Thanks for the pointer, these should be nuked 😃

srenatus:
will merge when green. deployment team will know how to hunt is down if this ends up being stupid for one reason or another :wink:

lib/platform/config.go
srenatus: ℹ️ Rolled back the URI change, it's only supported by the golang lib, not libpq. (pg_dump etc)

PR #637, opened 2019-06-19
make json well-formed for compliance export

rickmarry:

As users of the compliance json export endpoint, we need to have properly formed json returned in the response of the request.  

Currently, the response is designed to be made up of a json array of json objects.  The problem is that this array is not surrounded by open/close square braces, and the objects within it are not comma separated; both of which are required in order for a json array to be considered well-formed.

This PR addresses both shortcomings and as a result, the compliance json export is now well-formed.

### :+1: Definition of Done
The json that comes back is well formed.

- [ ] Code actually executed?

### DIFF
--- a/components/automate-gateway/gateway/services.go
+++ b/components/automate-gateway/gateway/services.go
@@ -538,6 +538,15 @@ func (s *Server) ReportExportHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	//when the type is json, we will be retrieving a json array of objects and since we will be getting them one at a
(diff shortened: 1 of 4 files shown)


phiggins:
I raised this concern in one of our meetings before but I'm not sure this is correct. We have multiple endpoints that export a JSON stream: https://en.wikipedia.org/wiki/JSON_streaming

If that's something we want to do across the board that's fine, but now our json export endpoints have different behavior and we've potentially changed the output that a customer could depend on.

PR #639, opened 2019-06-19
use start_time as well as end_time when filtering reports for export

rickmarry:
We need to consider start_time as well as end_time when performing export for compliance scans.

Currently, we ignore start_time and, instead, compute it as the [UTC] beginning of day for the same day as that of the end_time.  The rationale for this is that because all other apis for compliance reporting, use this approach to time span, so should export.  We have recently identified a real need to break the exports up into smaller time intervals, necessitating the use of start_time.

Once this PR merges, we will no longer compute start_time from end_time, we will use what is given.  From the UI perspective, nothing will change because now the UI will compute that start_time and send it to us.  The API users will benefit from this new flexibility.

- [ ] Code actually executed?

### DIFF
--- a/components/compliance-service/api/reporting/server/server.go
+++ b/components/compliance-service/api/reporting/server/server.go
@@ -163,7 +163,7 @@ func (srv *Server) Export(in *reporting.Query, stream reporting.ReportingService
 	}
 
 	// Step 1: Retrieving the latest report ID for each node based on the provided filters
-	esIndex, err := relaxting.GetEsIndex(formattedFilters, false, false)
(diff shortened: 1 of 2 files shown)


vjeffrey:
the change for the ui has merged: https://github.com/chef/automate/pull/647

vjeffrey:
we've found that the ui change that was merged is not doing quite the right thing:
Screen Shot 2019-06-21 at 09 12 36

the start time should be beg of day of end time, but it is always beg of day of today

vjeffrey:
i will be getting a fix for the required ui change up sometime today

PR #640, opened 2019-06-19
report-ids endpoint

rickmarry:
This endpoint returns the list of compliance run-ids for the given input filter.  All of the compliance supported filters are supported and both start_time as well as end_time are used when filtering the list.

This is useful for getting all of the runids that happened inside of a timespan.

- [ ] Code actually executed?


components/automate-gateway/api/compliance/reporting/reporting.proto
```diff
@@ -30,6 +30,21 @@ service ReportingService {
 			action: "compliance:reports:list"
 		};
 	};
+	// should cover /reportids
+    rpc ListReportIds(Query) returns (ReportIds) {
```
vjeffrey: can you add a comment here noting that this endpoint is *not* limited by the 10k es pagination thing

components/compliance-service/api/reporting/server/server.go
```diff
@@ -63,6 +63,48 @@ func (srv *Server) ListReports(ctx context.Context, in *reporting.Query) (*repor
 	return &reports, nil
 }
 
+// ListReports returns a list of reports based on query
+func (srv *Server) ListReportIds(ctx context.Context, in *reporting.Query) (*reporting.ReportIds, error) {
```
vjeffrey: i know the pr is still marked as wip...but i assume you'll be adding a test or two for this new fancy endpoint?

rickmarry: yep.. i got the tests in now.. will take this out of WIP

components/automate-gateway/api/compliance/reporting/reporting.proto
rickmarry: done

rickmarry: done

Issue #463, opened 2019-05-31
Add configurable gap to donut chart component


apriofrost:
# User Story
When there is multiple statuses represented by different status colors are presented in a donut chart, it becomes harder for the user to tell them apart intuitively. Adding a gap between the connecting edge can improve its readability.

# Definition of Done
Update the existing donut chart component in the UI library:
- [x] When there is only one color in the donut chart, remove the current gray line on the highest point of the circle.

- [x] When there is more than one color in the donut chart, add a gap between the color bands where they connect. 

- [x] The size of the gap should be configurable depending on the overall size of the chart (provided by the designer in the instance)

- [x] Apply the change on the Service Groups page with the gap size = 1px.

# Fina Look on the Service Groups Page
Left: Before  Right: After
![Snip20190530_13](https://user-images.githubusercontent.com/9401459/58658165-bb62c200-82d4-11e9-8b16-0cc32266463e.png)

PR #643, opened 2019-06-20
UI-463 Add config gap to donut chart component Signed-Off-By: Tara Bl…

tarablack01:
…ack <tblack@chef.io>

Screen Shot 2019-06-19 at 7 19 56 PM

When there is multiple statuses represented by different status colors are presented in a donut chart, it becomes harder for the user to tell them apart intuitively. Adding a gap between the connecting edge can improve its readability.

### :+1: Definition of Done
Update the existing donut chart component in the UI library:

When there is only one color in the donut chart, remove the current gray line on the highest point of the circle.

The size of the gap should be configurable depending on the overall size of the chart (provided by the designer in the instance)

Apply the change on the Service Groups page with the gap size = 1px.

### Replication Steps
Type the word, feat, on your keyboard // Turn on EAS Applications
run cmds:
[hab studio]: applications_populate_database
[hab studio]: make update-ui-lib
Service groups is under applications

https://github.com/chef/automate/issues/463

### DIFF
--- a/components/chef-ui-library/src/charts/chef-radial-chart/chef-radial-chart.tsx
+++ b/components/chef-ui-library/src/charts/chef-radial-chart/chef-radial-chart.tsx
@@ -66,7 +66,7 @@ let UID = 0;
  *     color: var(--chef-unknown);
  *   }
  * </style>
- * <chef-radial-chart style="width: 220px; height: 220px;">
+ * <chef-radial-chart gap-size=".05" style="width: 220px; height: 220px;">
  *   <span slot="innerText">Text for the center of the chart</span>
  *
  *   <chef-data-point value="4" class="failed">4 Failed</chef-data-point>
@@ -87,6 +87,11 @@ export class ChefRadialChart {
   @Prop() id: string;
 
   /**
+   * The width of the gap to apply between chart segments.
+   */
+  @Prop({ reflectToAttr: true }) gapSize = .03;
+
+  /**
    * Optionally hide tooltips. They are shown by default.
    */
   @Prop() tooltips = true;
@@ -105,7 +110,9 @@ export class ChefRadialChart {
 
   @Method() updateDataPoints() {
     const dataPoints = Array.from(this.el.querySelectorAll('chef-data-point'));
-    this.dataPoints = groupBy((d) => d.secondary ? 'secondary' : 'primary', dataPoints);
+    const dataPointsWithValues = [];
+    dataPoints.forEach((d) => (Number(d.value) > 0) ? dataPointsWithValues.push(d) : null);
+    this.dataPoints = groupBy((d) => d.secondary ? 'secondary' : 'primary', dataPointsWithValues);
   }
 
   componentWillLoad() {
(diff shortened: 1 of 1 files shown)


apriofrost:
looking great!!! 👍

components/chef-ui-library/src/charts/chef-radial-chart/chef-radial-chart.tsx
```diff
@@ -86,6 +86,11 @@ export class ChefRadialChart {
    */
   @Prop() id: string;
 
+  /**
+   * The width of the gap to apply between chart segments.
```
msorens: This comment should mention units:
```suggestion
   * The width of the gap (in radians) to apply between chart segments.
```

components/chef-ui-library/src/charts/chef-radial-chart/chef-radial-chart.tsx
```diff
@@ -86,6 +86,11 @@ export class ChefRadialChart {
...
+   */
+  @Prop({ reflectToAttr: true }) gapSize = .03;
```
msorens: I would like to see this converted to degrees instead of radians for two reasons:
1. Probably less folks think in radians than in degrees.
2. Much preferred to be able to specify gapSize as an integer instead of floating point.

Keep the public API as degrees; keep the conversion to radians internal to this file.

components/chef-ui-library/src/charts/chef-radial-chart/chef-radial-chart.tsx
```diff
@@ -105,15 +110,17 @@ export class ChefRadialChart {
 
   @Method() updateDataPoints() {
     const dataPoints = Array.from(this.el.querySelectorAll('chef-data-point'));
-    this.dataPoints = groupBy((d) => d.secondary ? 'secondary' : 'primary', dataPoints);
+    const dataPointsWithValues = [];
+    dataPoints.forEach((d) => (Number(d.value) > 0) ? dataPointsWithValues.push(d) : null);
+    this.dataPoints = groupBy((d) => d.secondary ? 'secondary' : 'primary', dataPointsWithValues);
```
scottopherson: Could shave off a line using `map`:

`const dataPointsWithValues = dataPoints.map(d => Number(d.value) > 0) ? d : null);`

components/chef-ui-library/src/charts/chef-radial-chart/chef-radial-chart.tsx
```diff
@@ -105,15 +110,17 @@ export class ChefRadialChart {
...
   }
 
   componentWillLoad() {
     this.updateDataPoints();
   }
 
   render() {
-    const pieChart = pie().sort(null).value((d) => d.value);
+    const pieChart = pie().sort(null).value((d) => d.value).padAngle(this.gapSize);
```
scottopherson: @msorens reasons sound reasonable to me:`this.gapSize * Math.PI / 180`

components/chef-ui-library/src/charts/chef-radial-chart/chef-radial-chart.tsx
tarablack01: @scottopherson If I use map it will keep the null dataPoints in the array.

apriofrost:
Looking great to me! I like using degrees, too! Good suggestion @msorens !

components/chef-ui-library/src/charts/chef-radial-chart/chef-radial-chart.tsx
scottopherson: @tarablack01 ah yea I see. Could go with `.filter` then:

`const dataPointsWithValues = dataPoints.filter(d => Number(d.value) > 0)`

PR #646, opened 2019-06-20
[A2-815] list rules with staged and current rules

srenatus:

> `ListRules`: Refactor or write separate APIs to return both a current rules list and staged rules list

This adds a `ListStagedOrAppliedRules` to the storage layer, and another grpc method to authz-service exposing it.

#### TODO:
- [ ] expose in the gateway, using an ENUM for applied/staged

### :+1: Definition of Done

> `ListRules`: Refactor or write separate APIs to return both a current rules list and staged rules list

😉 

- `rebuild components/authz-service`
- create a project

TODO

- [A2-815](https://chefio.atlassian.net/browse/A2-815)

- [X] Code actually executed?


api/interservice/authz/v2/project.proto
```diff
@@ -111,6 +111,8 @@ message ProjectRule {
     string name = 3;
     ProjectRuleTypes type = 4;
     repeated Condition conditions = 5;
+    bool deleted = 6;
```
srenatus: ℹ️Surfacing the deleted and staged/applied status of a rule in the interservice API. Nothing done for `automate-gateway` yet.

api/interservice/authz/v2/project.proto
```diff
@@ -173,7 +175,9 @@ message GetRuleResp {
     ProjectRule rule = 1;
 }
 
-message ListRulesReq {}
+message ListRulesReq {
+    bool include_staged = 1;
```
srenatus: ℹ️ Pass `ListRulesReq{IncludeStaged: true}` to also see staged rules.

api/interservice/authz/v2/project.proto
```diff
@@ -111,6 +111,8 @@ message ProjectRule {
     string name = 3;
     ProjectRuleTypes type = 4;
     repeated Condition conditions = 5;
+    bool deleted = 6;
+    string status = 7;
```
srenatus: _OMG why no enum?!!!_

I hear you -- but it's just so annoying to convert three different kinds of enums (storage layer, authz-service GRPC, automate-gateway GRPC), and I'm not convinced it's worth it. Happy to listen to arguments, but maybe we can just make our lives a little brighter and our PRs a little smaller. 😉 

Also, in the external API (automate-gateway), we can (and should) still use an enum.

components/authz-service/server/v2/rules_test.go
```diff
@@ -143,6 +143,7 @@ func TestCreateRule(t *testing.T) {
 							Operator:  api.ProjectRuleConditionOperators_MEMBER_OF,
```
srenatus: ℹ️I haven't added any tests on that layer yet.

components/authz-service/storage/postgres/migration/sql/49_add_status_field_to_rules.up.sql
```diff
@@ -0,0 +1,154 @@
+BEGIN;
```
srenatus: ℹ️ Updated SQL functions to return the "status" = "applied"/"staged".

Converted to use `json_build_object` in the process.

components/authz-service/storage/postgres/migration/sql/README.md
```diff
@@ -47,3 +47,4 @@
 - [`45_query_rules_for_project.up.sql`](45_query_rules_for_project.up.sql)
 - [`46_add_application_policies.up.sql`](46_add_application_policies.up.sql)
 - [`47_staged_rules_tables.up.sql`](47_staged_rules_tables.up.sql)
+- [`49_add_status_field_to_rules.up.sql`](49_add_status_field_to_rules.up.sql)
```
srenatus: ⚠️ before merge, this will need to be adjusted.

components/authz-service/storage/v2/memstore/memstore.go
```diff
@@ -243,6 +243,7 @@ func (s *State) GetPolicyChangeNotifier(ctx context.Context) (v2.PolicyChangeNot
 }
```
srenatus: ℹ️ Memstore adaptations. I anticipate that we all kind of have some of these in our current WIP PRs... :/

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -137,7 +137,7 @@ SELECT array_agg(policy_id) FROM pol_ids`,
 func (p *pg) ListPolicies(ctx context.Context) ([]*v2.Policy, error) {
 	projectsFilter, err := projectsListFromContext(ctx)
 	if err != nil {
-		return nil, p.processError(err)
+		return nil, err
```
srenatus: one of two drive-by cleanups, please see the commit

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1104,19 +1104,14 @@ func (p *pg) DeleteRule(ctx context.Context, id string) error {
...
 	}
 
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	tx, err := p.db.BeginTx(ctx, nil)
```
srenatus: one of two drive-by cleanups, please see the commit

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1127,30 +1122,28 @@ func (p *pg) GetStagedOrAppliedRule(ctx context.Context, id string) (*v2.Rule, e
...
-	}
-
 	return &rule, nil
 }
 
 func (p *pg) ListRules(ctx context.Context) ([]*v2.Rule, error) {
+	return p.listRulesUsingFunction(ctx, "SELECT query_rules($1)")
+}
+
+func (p *pg) ListStagedAndAppliedRules(ctx context.Context) ([]*v2.Rule, error) {
```
srenatus: ℹ️ new method for retrieving rules of each state -- as mentioned in A2-815.

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3501,14 +3535,106 @@ func TestListRules(t *testing.T) {
...
+			assert.Equal(t, &rule2, resp[0])
+		},
+	}
+
+	for name, test := range cases {
+		t.Run(name, test)
+		db.Flush(t)
+	}
+}
+func TestListStagedAndAppliedRules(t *testing.T) {
```
srenatus: The amount of copy-paste involved in creating these tests is deeply unsatisfying :grumpycat:

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -6438,9 +6567,10 @@ func insertAppliedRule(t *testing.T, db *testhelpers.TestDB, rule storage.Rule)
 			dbID, pq.Array(c.Value), c.Attribute.String(), c.Operator.String())
 		require.NoError(t, err)
 	}
+	rule.Status = "applied"
 }
 
-func insertStagedRule(t *testing.T, db *testhelpers.TestDB, rule storage.Rule) {
+func insertStagedRule(t *testing.T, db *testhelpers.TestDB, rule *storage.Rule) {
```
srenatus: ℹ️ We have to change the function type to pass a pointer to mutate the `rule` argument, to add the status below

components/authz-service/storage/postgres/migration/sql/49_add_status_field_to_rules.up.sql
```diff
@@ -0,0 +1,154 @@
...
+
+CREATE OR REPLACE FUNCTION
+  query_staged_and_applied_rules(_project_filter TEXT[])
```
bcmdarroch: does this query concatenate the rows from the two tables? 

for `GetStagedOrAppliedRule(id)` we return the staged rule if it exists, otherwise return the applied rule (due to an acceptance requirement that users should always see the latest staged change). we may want to make this query work the same way - returning only one version of each rule. 

although since i don't know of a specific case where we're using this query yet, this is probably ok for now.

components/authz-service/storage/v2/memstore/memstore.go
```diff
@@ -243,6 +243,7 @@ func (s *State) GetPolicyChangeNotifier(ctx context.Context) (v2.PolicyChangeNot
...
 
 func (s *State) CreateRule(_ context.Context, rule *storage.Rule) (*storage.Rule, error) {
+	rule.Status = "applied"
```
bcmdarroch: if we want to match postgres behavior, the status here would be `staged`. only on ApplyRule does everything staged get set to `applied`.

components/authz-service/storage/v2/memstore/memstore.go
bcmdarroch: If we wanna match it as close to postgres as possible, I think we'd want a staged cache and an applied cache. Is that worthwhile? Probably not.

api/interservice/authz/v2/project.proto
tylercloke: I really like the compromise of having an enum at the gateway and not in the domain layer. The enum unwrapping code is super super annoying.

tylercloke: I really like the compromise of having an enum at the gateway and not in the domain layer. The enum unwrapping code is super super annoying.

components/authz-service/server/v2/rules_test.go
```diff
@@ -470,6 +471,7 @@ func TestGetRule(t *testing.T) {
 				Type:       api.ProjectRuleTypes_NODE,
 				ProjectId:  projectID,
 				Conditions: apiConditions,
+				Status:     "applied",
```
tylercloke: Maybe this could be a constant? We have some at the DB level after the `DeleteRule` refactor PR merged.

components/authz-service/storage/v2/memstore/memstore.go
tylercloke: Heh probably not as much WIP as we should. Left a comment in Slack.

components/authz-service/storage/postgres/migration/sql/49_add_status_field_to_rules.up.sql
msorens: A couple points.
1. Looking at `GetStagedOrAppliedRule` I do not believe it does what you think, @bcmdarroch . Looks like the SQL underlying that call, `query_staged_rule`, works exactly the same as this new function, returning both staged and applied rules.
2. I do take issue with the name of this new function `query_staged_and_applied_rules` compared to the aforementioned function `query_staged_rule`. Please rename one or the other so they differ only in plurality. 😁

components/authz-service/storage/postgres/migration/sql/49_add_status_field_to_rules.up.sql
```diff
@@ -0,0 +1,154 @@
...
+  RETURNS setof json AS $$
+
+  -- fetch staged rules
+  SELECT json_build_object(
+    'id', r.id,
+    'project_id', r.project_id,
+    'name', r.name,
+    'type', r.type,
+    'deleted', r.deleted,
+    'status', 'staged',
```
msorens: Consider SQL constants for 'staged' and 'applied'

srenatus:
rebased. will adress review comments later.

components/authz-service/storage/postgres/migration/sql/49_add_status_field_to_rules.up.sql
srenatus: I think we've resolved that in standup:
> returning both staged and applied rules.

but we use `QueryRowContext`, so we're basically returning the staged one first (if it exists), so we're OK here, even if by accident. There might be a cleaner solution, but this one will do. 😉

components/authz-service/server/v2/rules_test.go
srenatus: yeah a constant in a different package :wink: this is only tests I'll add another package-level one.

components/authz-service/storage/postgres/migration/sql/49_add_status_field_to_rules.up.sql
srenatus: Really? As far as I understand it, constants in SQL are scoped per-function. So, this functions could have a constant, but there's not much to gain, really, since both strings only occurr once.... Also it doesn't look common in our code base to have SQL constants. 😃

components/authz-service/storage/v2/memstore/memstore.go
srenatus: Leaving this as-is to not break all the memstore-based tests. Let's nuke them. 😉

components/authz-service/storage/postgres/migration/sql/49_add_status_field_to_rules.up.sql
bcmdarroch: `query_staged_rules` checks both `applied` and `staged` but [only returns the applied rule if no staged version exists](https://github.com/chef/automate/blob/master/components/authz-service/storage/postgres/migration/sql/47_staged_rules_tables.up.sql#L57).

PR #651, opened 2019-06-20
[A2-878] Update Rule Refactor

blakestier:

Previously, the UpdateRule API directly updated the rules. Since any rule update takes time to completely apply to ingested resources, we're refactoring the rules API so that all changes to rules are stored in a `staged` rules table (preserving the rules in their current state in the `applied` table). Only when the user calls ApplyRules will the updated rules in the staging table be applied to the `applied` rules table.

### :+1: Definition of Done

UpdateRule now:

1) Creates a new "staged rule" for existing "applied" rule
2) Updates existing "staged rule" if already created
3) Does not update staged rules marked for deletion

TODO: what should memstore implementation look like?

#### Setup
```
build components/authz-service
start_all_services
export TOK=`chef-automate iam token create TEST --admin`
```
#### Creates a new "staged rule" for existing "applied" rule

1. insert an "applied" rule into DB (if we use curl, it will just go to staged)
```
chef-automate dev psql chef_authz_service

insert into iam_projects (id, name, type, projects) values ('foo-project', 'Foo Project', 'custom', ARRAY['foo-project']);

insert into iam_project_rules (id, project_id, name, type) values ('foo-rule', 'foo-project', 'foorule', 'node');

\q
```
2. update rule via curl
```
curl -kH "api-token: $TOK" -X PUT https://localhost/apis/iam/v2beta/rules/foo-rule -d "$(jq -n '{ id: "foo-rule", name: "my updated foo rule", type: "NODE", project_id: "foo-project", conditions: [{ attribute: "CHEF_SERVERS", operator: "MEMBER_OF", values: ["prod", "staging"]}]}')" | jq .

# returns rule with updated name and conditions
```
3. get updated rule via curl (always returns most updated version of the rule, aka whatever is staged)
```
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/rules/foo-rule | jq .

# returns rule with updated name and conditions
```

#### Updates existing "staged rule" if already created
1. curl create new rule
```
curl -kH "api-token: $TOK" -X POST https://localhost/apis/iam/v2beta/rules -d "$(jq -n '{ id: "bar-rule", name: "my bar rule", type: "EVENT", project_id: "foo-project", conditions: [{ attribute: "CHEF_ORGS", operator: "EQUALS", values: ["dev"]}]}')" | jq .
```
2. update rule via curl
```
curl -kH "api-token: $TOK" -X PUT https://localhost/apis/iam/v2beta/rules/bar-rule -d "$(jq -n '{ id: "bar-rule", name: "my updated bar rule", type: "EVENT", project_id: "foo-project", conditions: [{ attribute: "CHEF_ORGS", operator: "EQUALS", values: ["dev"]}, { attribute: "CHEF_SERVERS", operator: "EQUALS", values: ["prod"]}]}')" | jq .

# returns rule with updated name and conditions
```
3. get updated rule via curl 
```
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/rules/bar-rule | jq .

# returns rule with updated name and conditions
```

#### Does not update staged rules marked for deletion
1. delete applied rule via curl (marks staged rule as deleted)
```
curl -kH "api-token: $TOK" -X DELETE https://localhost/apis/iam/v2beta/rules/foo-rule | jq .
```
2. update rule via curl
```
curl -kH "api-token: $TOK" -X PUT https://localhost/apis/iam/v2beta/rules/foo-rule -d "$(jq -n '{ id: "foo-rule", name: "my updated foo rule", type: "NODE", project_id: "foo-project", conditions: [{ attribute: "CHEF_SERVERS", operator: "MEMBER_OF", values: ["prod", "staging"]}]}')" | jq .

{
  "error": "error creating rule with ID \"foo-rule\": rule marked for deletion",
  "message": "error creating rule with ID \"foo-rule\": rule marked for deletion",
  "code": 13,
  "details": []
}
```

- [ ] Code actually executed?

### DIFF
--- a/components/authz-service/server/v2/project.go
+++ b/components/authz-service/server/v2/project.go
@@ -355,7 +355,7 @@ func (s *state) UpdateRule(ctx context.Context, req *api.UpdateRuleReq) (*api.Up
 				"cannot change project_id for existing rule with ID %q ", req.Id)
 		}
 		return nil, status.Errorf(codes.Internal,
-			"error creating rule with ID %q: %s", req.Id, err.Error())
+			"error updating rule with ID %q: %s", req.Id, err.Error())
 	}
 
 	apiRule, err := fromStorageRule(resp)
@@ -376,7 +376,7 @@ func (s *state) GetRule(ctx context.Context, req *api.GetRuleReq) (*api.GetRuleR
 			"error retrieving rule with ID %q: %s", req.Id, err.Error())
 	}
 	if resp.Deleted {
-		return nil, status.Errorf(codes.NotFound, "could not find rule with ID %q", req.Id)
+		return nil, status.Errorf(codes.NotFound, "rule with ID %q marked for deletion", req.Id)
 	}
 
 	apiRule, err := fromStorageRule(resp)
--- a/components/authz-service/storage/postgres/migration/sql/49_update_rule_triggers.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/49_update_rule_triggers.up.sql
@@ -8,29 +8,28 @@ AS $$
 DECLARE
   rule_db_id INTEGER;
 BEGIN
-    IF (EXISTS (SELECT id FROM iam_project_rules ipr WHERE id=in_rule_id AND projects_match_for_rule(ipr.project_id, projects_filter))) THEN -- cannot 'update' a non-existed applied rule
-            IF (NOT EXISTS (SELECT id, project_id FROM iam_project_rules ipr WHERE id=in_rule_id AND project_id=in_project_id)) THEN
-                RAISE EXCEPTION 'incoming project does not match rule project %', in_rule_id USING -- do we need this? maybe we can just return null or something?
-                ERRCODE='PRJTR';
-            END IF;
-
-            INSERT INTO iam_staged_project_rules as ispr (id, project_id, name, type, deleted)
-            VALUES (in_rule_id, in_project_id, in_name, in_type, false)
-            ON CONFLICT (id) -- applied rule has already been updated, so update the staged version of it
-                DO UPDATE
-                        SET
-                        id = in_rule_id,
-                        project_id = in_project_id,
-                        name          = in_name,
-                        type   = in_type,
-                        deleted    = in_deleted
-            WHERE ispr.id = in_rule_id AND projects_match_for_rule(ispr.project_id, projects_filter)
-        RETURNING ispr.db_id INTO rule_db_id;
-        RETURN rule_db_id;
-    -- TODO: add conditions
+    IF (NOT EXISTS (SELECT id FROM iam_project_rules ipr WHERE id=in_rule_id AND projects_match_for_rule(ipr.project_id, projects_filter))) AND
+        (NOT EXISTS (SELECT id FROM iam_staged_project_rules ipr WHERE id=in_rule_id AND projects_match_for_rule(ipr.project_id, projects_filter))) THEN
+            RAISE EXCEPTION 'not found %', in_rule_id USING
+            ERRCODE='case_not_found'; -- do we need this more specific? can we just return 20000?
+    ELSIF (NOT EXISTS (SELECT id, project_id FROM iam_project_rules ipr WHERE id=in_rule_id AND project_id=in_project_id)) AND
+        (NOT EXISTS (SELECT id, project_id FROM iam_staged_project_rules ipr WHERE id=in_rule_id AND project_id=in_project_id)) THEN
+            RAISE EXCEPTION 'incoming project does not match rule project %', in_rule_id USING -- do we need this? maybe we can just return null or something?
+            ERRCODE='PRJTR';
     ELSE
-        RAISE EXCEPTION 'not found %', in_rule_id USING -- do we need this? maybe we can just return null or something?
-        ERRCODE='case_not_found';
+        INSERT INTO iam_staged_project_rules as ispr (id, project_id, name, type, deleted)
+        VALUES (in_rule_id, in_project_id, in_name, in_type, false)
+        ON CONFLICT (id) -- applied rule has already been updated, so update the staged version of it
+            DO UPDATE
+                    SET
+                    id = in_rule_id,
+                    project_id = in_project_id,
+                    name          = in_name,
+                    type   = in_type,
+                    deleted    = in_deleted
+        WHERE ispr.id = in_rule_id AND projects_match_for_rule(ispr.project_id, projects_filter)
+        RETURNING ispr.db_id INTO rule_db_id;
+    RETURN rule_db_id;
     END IF;
     RETURN rule_db_id;
 END;
@@ -42,10 +41,10 @@ BEGIN
         ERRCODE='PRJID';
 END$$ LANGUAGE plpgsql;
 
--- cannot update project it
+-- cannot update project id
 CREATE TRIGGER verify_project_id BEFORE UPDATE OF project_id ON iam_staged_project_rules
 FOR EACH ROW
-WHEN (OLD.project_id != NEW.project_id) 
+WHEN (OLD.project_id != NEW.project_id)
 EXECUTE PROCEDURE fn_project_id_check();
 
 CREATE OR REPLACE FUNCTION fn_deleted_rule_check() RETURNS TRIGGER AS $$
--- a/components/authz-service/storage/postgres/migration/sql/49_update_rule_triggers.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/49_update_rule_triggers.up.sql
@@ -35,18 +35,6 @@ BEGIN
 END;
 $$;
 
-CREATE OR REPLACE FUNCTION fn_project_id_check() RETURNS TRIGGER AS $$
-BEGIN 
-    RAISE EXCEPTION 'cannot change project_id: % -> %', OLD.project_id, NEW.project_id USING
-        ERRCODE='PRJID';
-END$$ LANGUAGE plpgsql;
-
--- cannot update project id
-CREATE TRIGGER verify_project_id BEFORE UPDATE OF project_id ON iam_staged_project_rules
-FOR EACH ROW
-WHEN (OLD.project_id != NEW.project_id)
-EXECUTE PROCEDURE fn_project_id_check();
-
 CREATE OR REPLACE FUNCTION fn_deleted_rule_check() RETURNS TRIGGER AS $$
 BEGIN 
     RAISE EXCEPTION 'rule with id % has been deleted', NEW.id USING
--- a/components/authz-service/storage/postgres/migration/sql/49_update_rule_triggers.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/49_update_rule_triggers.up.sql
@@ -9,6 +9,11 @@ DECLARE
   rule_db_id INTEGER;
 BEGIN
     IF (EXISTS (SELECT id FROM iam_project_rules ipr WHERE id=in_rule_id AND projects_match_for_rule(ipr.project_id, projects_filter))) THEN -- cannot 'update' a non-existed applied rule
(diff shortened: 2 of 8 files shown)


components/authz-service/server/v2/project.go
```diff
@@ -376,7 +376,7 @@ func (s *state) GetRule(ctx context.Context, req *api.GetRuleReq) (*api.GetRuleR
 			"error retrieving rule with ID %q: %s", req.Id, err.Error())
 	}
 	if resp.Deleted {
-		return nil, status.Errorf(codes.NotFound, "could not find rule with ID %q", req.Id)
+		return nil, status.Errorf(codes.NotFound, "rule with ID %q marked for deletion", req.Id)
```
blakestier: We updated the error in `UpdateRule` to be more transparent-- a rule marked for deletion is essentially "Not Found" (cannot be deleted, will not be displayed on the UI), so we backtracked and modified `CreateRule` as well.

components/authz-service/storage/postgres/migration/sql/49_update_rule_triggers.up.sql
```diff
@@ -0,0 +1,50 @@
+
```
blakestier: extra line?

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1024,32 +1024,25 @@ func (p *pg) UpdateRule(ctx context.Context, rule *v2.Rule) (*v2.Rule, error) {
...
+		rule.ID, rule.ProjectID, rule.Name, rule.Type.String(), false, pq.Array(projectsFilter))
+	var ruleDbID int
+	if err := row.Scan(&ruleDbID); err != nil {
 		if err == sql.ErrNoRows {
 			return nil, storage_errors.ErrNotFound
 		}
 		return nil, p.processError(err)
 	}
 
-	// If they tried to change the project_id, abort the transaction with an error.
```
blakestier: This is now bundled into the SQL query `update_rule`

components/authz-service/server/v2/project.go
tylercloke: I wonder if we should change the shape of our restful APIs to be more clear about when an endpoint is operating on just applied rules or not (`/applied-rules` vs `/rules`). I could see getting a 404 being a bit confusing if a staged rule exists. Don't need to decide here, just something to think about 🤔

components/authz-service/storage/postgres/migration/sql/49_update_rule_triggers.up.sql
```diff
@@ -0,0 +1,50 @@
...
+BEGIN;
+
+CREATE OR REPLACE FUNCTION update_rule(in_rule_id text, in_project_id text, in_name text, in_type text, in_deleted boolean, projects_filter TEXT[])
```
tylercloke: [super nit] It's a good practice in sql functions to start the variable names with an underscore (e.g. `_in_rule_id text`) to avoid potential future conflict with tables, columns, etc with the same name that might get added in the future. Probably pretty safe in this case with the `in_` prefixes though.

components/authz-service/storage/postgres/migration/sql/49_update_rule_triggers.up.sql
```diff
@@ -0,0 +1,50 @@
...
+$$;
+
+CREATE OR REPLACE FUNCTION fn_deleted_rule_check() RETURNS TRIGGER AS $$
+BEGIN 
+    RAISE EXCEPTION 'rule with id % has been deleted', NEW.id USING
+        ERRCODE='RDLTD'; 
+END$$ LANGUAGE plpgsql;
+
+-- cannot update a rule that is staged for deletion 
+CREATE TRIGGER verify_not_deleted BEFORE UPDATE ON iam_staged_project_rules
```
tylercloke: Do you know if `BEFORE UPDATE` triggers on deletion of rows? If so, this will break my PR as they both are written since the last thing the apply rules database code does is just straight up drop everything from `iam_staged_project_rules`. If it doesn't trigger on deletion, then we are fine :)

components/authz-service/storage/postgres/migration/sql/49_update_rule_triggers.up.sql
```diff
@@ -0,0 +1,50 @@
...
+        INSERT INTO iam_staged_project_rules as ispr (id, project_id, name, type, deleted)
+        VALUES (in_rule_id, in_project_id, in_name, in_type, false)
+        ON CONFLICT (id) -- applied rule has already been updated, so update the staged version of it
+            DO UPDATE
+                    SET
+                    id = in_rule_id,
+                    project_id = in_project_id,
+                    name          = in_name,
+                    type   = in_type,
+                    deleted    = in_deleted
```
tylercloke: Why are we passing deleted into the update function? They shouldn't be able to update to deleted, right? You should have to run `DeleteRule` for that I think. I see that it's hard-coded to false in the go code so maybe we can just drop it and not update the deleted column.

components/authz-service/storage/postgres/migration/sql/49_update_rule_triggers.up.sql
tylercloke: I think the answer is we are fine after googling. Worst case I'll see things break when I rebase and we can address then!

blakestier: I think UPDATE registers differently than DELETE 👍

components/authz-service/storage/postgres/postgres.go
```diff
@@ -77,6 +77,12 @@ func parsePQError(err *pq.Error) error {
 		return storage.ErrConflict
 	case "23503": // Foreign key violation
 		return storage.ErrForeignKey
+	case "20000": // not found
+		return storage.ErrNotFound
+	case "PRJTR": // custom code for when a user tries to change a rule's project
+		return storage.ErrChangeProjectForRule
+	case "RDLTD": // object is staged for deletion and cannot be updated
```
msorens: Suggestions for consistency and clarity on lines 74, 82, and 84:
``` 
// custom code: attempt to delete a non-deletable policy defined in migration 02
// custom code: attempt to change a rule's projects that are immutable
// custom code: attempt to update a rule that is staged for deletion
```

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1024,32 +1024,25 @@ func (p *pg) UpdateRule(ctx context.Context, rule *v2.Rule) (*v2.Rule, error) {
...
-	if projectID != rule.ProjectID {
-		return nil, storage_errors.ErrChangeProjectForRule
-	}
-
 	// Delete the existing conditions. Don't need to worry about not found case since a rule must have conditions.
```
msorens: s/not found/"not found"/ for clarity

components/authz-service/storage/postgres/migration/sql/49_update_rule_triggers.up.sql
```diff
@@ -0,0 +1,50 @@
...
+        WHERE ispr.id = in_rule_id AND projects_match_for_rule(ispr.project_id, projects_filter)
+        RETURNING ispr.db_id INTO rule_db_id;
```
msorens: Would help understanding this with indenting like this:
```
       ON CONFLICT (id) DO -- applied rule has already been updated, so update the staged version of it
            UPDATE SET
              id = in_rule_id,
              project_id = in_project_id,
              name = in_name,
              type = in_type,
              deleted = in_deleted
            WHERE ispr.id = in_rule_id AND projects_match_for_rule(ispr.project_id, projects_filter)
        RETURNING ispr.db_id INTO rule_db_id;
```
Matches indent level of "on conflict" with "returning"
Matches "where" with "update set".
Removes the ... unusual indents on the argument list.

components/authz-service/storage/postgres/migration/sql/49_update_rule_triggers.up.sql
```diff
@@ -0,0 +1,50 @@
...
+    RETURN rule_db_id;
+    END IF;
+    RETURN rule_db_id;
```
msorens: Delete this line; it is never reached.

components/authz-service/storage/postgres/migration/sql/49_update_rule_triggers.up.sql
```diff
@@ -0,0 +1,50 @@
...
+  rule_db_id INTEGER;
+BEGIN
+    IF (NOT EXISTS (SELECT id FROM iam_project_rules ipr WHERE id=in_rule_id AND projects_match_for_rule(ipr.project_id, projects_filter))) AND
+        (NOT EXISTS (SELECT id FROM iam_staged_project_rules ipr WHERE id=in_rule_id AND projects_match_for_rule(ipr.project_id, projects_filter))) THEN
+            RAISE EXCEPTION 'not found %', in_rule_id USING
+            ERRCODE='case_not_found'; -- do we need this more specific? can we just return 20000?
+    ELSIF (NOT EXISTS (SELECT id, project_id FROM iam_project_rules ipr WHERE id=in_rule_id AND project_id=in_project_id)) AND
+        (NOT EXISTS (SELECT id, project_id FROM iam_staged_project_rules ipr WHERE id=in_rule_id AND project_id=in_project_id)) THEN
+            RAISE EXCEPTION 'incoming project does not match rule project %', in_rule_id USING -- do we need this? maybe we can just return null or something?
+            ERRCODE='PRJTR';
```
msorens: Suggest delete all occurrences of `ipr` in L11, L12, L15, and L16. It is redundant.
Also, not used consistently in several places (example: L11 qualifies `project_id` but does not qualify `id`)

components/authz-service/storage/postgres/migration/sql/49_update_rule_triggers.up.sql
```diff
@@ -0,0 +1,50 @@
...
+RETURNS TEXT
+LANGUAGE plpgsql
+AS $$
+DECLARE
+  rule_db_id INTEGER;
+BEGIN
+    IF (NOT EXISTS (SELECT id FROM iam_project_rules ipr WHERE id=in_rule_id AND projects_match_for_rule(ipr.project_id, projects_filter))) AND
+        (NOT EXISTS (SELECT id FROM iam_staged_project_rules ipr WHERE id=in_rule_id AND projects_match_for_rule(ipr.project_id, projects_filter))) THEN
+            RAISE EXCEPTION 'not found %', in_rule_id USING
+            ERRCODE='case_not_found'; -- do we need this more specific? can we just return 20000?
```
msorens: Please move `USING` to the next line, so it reads `USING ERRCODE='case_not_found';`
Similarly for L17-18 and L39-40

components/authz-service/storage/postgres/migration/sql/49_update_rule_triggers.up.sql
```diff
@@ -0,0 +1,50 @@
...
+END;
+$$;
+
+CREATE OR REPLACE FUNCTION fn_deleted_rule_check() RETURNS TRIGGER AS $$
+BEGIN 
+    RAISE EXCEPTION 'rule with id % has been deleted', NEW.id USING
+        ERRCODE='RDLTD'; 
+END$$ LANGUAGE plpgsql;
+
+-- cannot update a rule that is staged for deletion
```
msorens: 🤔 What if we want to update a rule to simply be undeleted? I think this is a vital use case.

blakestier: I think at this point we are not offering that option. The UI will not display rules marked for deletion, for example. 

If we get to a place where we show drafts, we can remove this.

components/authz-service/storage/postgres/migration/sql/49_update_rule_triggers.up.sql
blakestier: I believe the SQL needed it due to ambiguity. I will verify and if that's the case, I'll use it elsewhere to be consistent.

Issue #444, opened 2019-05-29
Event Feed as a separate new service


lancewf:
As an Automate developer, I would like to be able to expand the event-feed to include events from other parts of the system.

The event feed is currently an aggregation between the config-mgmt-service and compliance-service. Most of the logic is spread between these two services and also the automate-gateway. The automate-gateway is the aggregator between these two services. When a request comes in for 100 of the most recent events, requests are made to the compliance-service and config-mgmt-service for 100 events. Then in the automate-gateway, these events are merged together and only 100 combined events are returned. 

The problem with the current system is adding new events would involve calling the third service to collect the events. This third service would have to store and implement the same event feed API and logic. This means the current approach is not easily expandable.

The suggested change is to pull the compliance-service's event feed code out of the compliance-service into its own service. Then have the config-mgmt-service's events be added this new service through the event-service. 

These changes would allow the below expansion of the event feed:
* Habitat Application events. 
* New event string with errors. This would allow Automate errors to be displayed in the event feed.
* notifications-service events. 
* Deployment-service events Automate upgrades and backups.  

The PR steps to accomplish this task would be:
1. Pull the compliance-service's event feed code out of the compliance-service into its own service called the event-feed-service
1. Push config-mgmt-service's events to the event-feed-service. Prevent the config-mgmt-service from returning events. 
1. Move all the automate-gateway event-feed code into the event-feed-service.

#### PRs
- [x] Create event-feed-service habitat package - https://github.com/chef/automate/pull/652
- [x] Add event-feed-service to automate-deployment-service - https://github.com/chef/automate/pull/671
- [x] Move event feed code from the compliance-service to the event-feed-service - https://github.com/chef/automate/pull/707
- [x] Direct events in the event-service to the event-feed-service - https://github.com/chef/automate/pull/732
- [x] Add the Event feed data lifecycle - https://github.com/chef/automate/pull/778
- [x] Migrate compliance-service data, backup, and redirect automate-gateway requests; delete compliance event feed - https://github.com/chef/automate/pull/784
- [x] Add project filtering for GetFeed - https://github.com/chef/automate/pull/4059
~~- [ ] Break up the "eventfeed-2-feeds" index into daily indexes~~
- [x] Add project filtering for GetFeedSummary - https://github.com/chef/automate/pull/4071
- [x] Add project filtering for GetFeedTimeline - https://github.com/chef/automate/pull/4072
[...]

PR #652, opened 2019-06-20
Event Feed initial service

lancewf:

Creating the event feed as a separate service. This is the initial check-in of the service before it is added to the deployment service. This is following the steps laid out in the Chicken and egg issue
https://github.com/chef/automate/blob/master/components/automate-deployment/docs/how-to-add-a-service.md#manifest-chicken-and-egg-issue. 

https://github.com/chef/automate/issues/444

- [x] Code actually executed?


components/event-feed-service/pkg/grpc/grpc.go
```diff
@@ -0,0 +1,46 @@
...
+
+// NewGRPCServer returns a server that provides our services:
+// * applications
+// * health
+func NewGRPCServer(connFactory *secureconn.Factory, c *config.EventFeed) *grpc.Server {
+	grpcServer := connFactory.NewServer()
+
+	log.Info("Creating Server and starting it")
+	// applicationsServer := server.New(c.GetStorage())
+	// applications.RegisterApplicationsServiceServer(grpcServer, applicationsServer)
```
vjeffrey: i was about to ask to remove these commented out bits, but given that we'll be iterating on this, i think it's fine to leave it for now

PR #655, opened 2019-06-20
[A2-813] Apply Staged Project Rules API

tylercloke:

Update `POST /apis/iam/v2beta/apply-rules` to apply all staged project rules and then trigger a `Refresh` of the OPA cache before kicking of the reindex.

Please see the commit titled `[SHAME]` for context around how the `PolicyRefresher` singleton is being routed around.

##### Build and generate

```
start_all_services
chef_load_nodes 100
```

##### Make projects and stage initial rules

```
# curl -kH "api-token: $TOK" -XPOST -d "$(jq -n '{ id: "project1", name: "my project 1" }')" https://localhost/apis/iam/v2beta/projects
# curl -kH "api-token: $TOK" -XPOST -d "$(jq -n '{ id: "project2", name: "my project 2" }')" https://localhost/apis/iam/v2beta/projects
# curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/rules -d "$(jq -n '{ id: "rule1", name: "my foo rule", type: "NODE", project_id: "project1", conditions: [{ operator: "EQUALS", attribute: "CHEF_ORGS", values: ["The Misfits"]}]}')" | jq .
# curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/rules -d "$(jq -n '{ id: "rule2", name: "my foo rule", type: "NODE", project_id: "project2", conditions: [{ operator: "MEMBER_OF", attribute: "CHEF_ORGS", values: ["Guardians of the Galaxy", "The Avengers"]}]}')" | jq .
```

Everything is still in unassigned because rules are staged only:

```
# curl -H "projects: (unassigned)" -sH "api-token: $TOK" https://localhost/api/v0/cfgmgmt/nodes?pagination.size=1000 --insecure | jq . | grep "The Misfits" | wc -l
3
# curl -H "projects: (unassigned)" -sH "api-token: $TOK" https://localhost/api/v0/cfgmgmt/nodes?pagination.size=1000 --insecure | jq . | grep "The Avengers" | wc -l
6
```

Can see things staged:

```
chef_authz_service=# select * from iam_project_rules;
 db_id | id | project_id | name | type
-------+----+------------+------+------
(0 rows)

chef_authz_service=# select * from iam_staged_project_rules;
 db_id |  id   | project_id |    name     | type | deleted
-------+-------+------------+-------------+------+---------
     2 | rule1 | project1   | my foo rule | node | f
     3 | rule2 | project2   | my foo rule | node | f
(2 rows)
```

##### Apply Initial Rules

```
# curl -kH "api-token: $TOK" -XPOST https://localhost/apis/iam/v2beta/apply-rules
```

It should be fast since we only have 100 nodes.

Can see everything moved from staged into applied:

```
chef_authz_service=# select * from iam_project_rules;
 db_id |  id   | project_id |    name     | type
-------+-------+------------+-------------+------
     1 | rule1 | project1   | my foo rule | node
     2 | rule2 | project2   | my foo rule | node
(2 rows)

chef_authz_service=# select * from iam_staged_project_rules;
 db_id | id | project_id | name | type | deleted
-------+----+------------+------+------+---------
(0 rows)
```

And can now curl nodes that have rules with projects:

```
# curl -H "projects: project1" -sH "api-token: $TOK" https://localhost/api/v0/cfgmgmt/nodes?pagination.size=1000 --insecure | jq . | grep "The Misfits" | wc -l
3
curl -H "projects: project2" -sH "api-token: $TOK" https://localhost/api/v0/cfgmgmt/nodes?pagination.size=1000 --insecure | jq . | grep "The Avengers" | wc -l
6
```

##### Update and Delete projects

Now delete rule1 and update rule2:

```
# curl -kH "api-token: $TOK" -XDELETE https://localhost/apis/iam/v2beta/rules/rule1
TODO
```

Can see the changes are staged in the DB:

```
chef_authz_service=# select * from iam_staged_project_rules;
 db_id |  id   | project_id |    name     | type | deleted
-------+-------+------------+-------------+------+---------
     1 | rule1 | project1   | my foo rule | node | t
(1 row)

chef_authz_service=# select * from iam_project_rules;
 db_id |  id   | project_id |    name     | type
-------+-------+------------+-------------+------
     1 | rule1 | project1   | my foo rule | node
     2 | rule2 | project2   | my foo rule | node
(2 rows)
```

And not live:

```
# curl -H "projects: project1" -sH "api-token: $TOK" https://localhost/api/v0/cfgmgmt/nodes?pagination.size=1000 --insecure | jq . | grep "The Misfits" | wc -l
3
TODO
```

Re-apply the rules:

```
# curl -kH "api-token: $TOK" -XPOST https://localhost/apis/iam/v2beta/apply-rules
```

See the updated results:

```
# curl -H "projects: project1" -sH "api-token: $TOK" https://localhost/api/v0/cfgmgmt/nodes?pagination.size=1000 --insecure | jq . | grep "The Misfits" | wc -l
0
# curl -H "projects: (unassigned)" -sH "api-token: $TOK" https://localhost/api/v0/cfgmgmt/nodes?pagination.size=1000 --insecure | jq . | grep "The Misfits" | wc -l
3
TODO
```

https://github.com/chef/automate/pull/651

- [ ] Code actually executed?

### DIFF
--- a/components/authz-service/server/server.go
+++ b/components/authz-service/server/server.go
@@ -71,7 +71,9 @@ func NewGRPCServer(ctx context.Context,
 		return nil, errors.Wrap(err, "could not initialize v1 server")
 	}
 
-	v2PolServer, err := v2.NewPostgresPolicyServer(ctx, l, e, migrationsConfig,
+	// TODO (tc) Refactor how singletons are shared between GRPC servers. Should we also
+	// be sharing a single v2 store / pg instance?
+	v2PolServer, policyRefresher, err := v2.NewPostgresPolicyServer(ctx, l, e, migrationsConfig,
 		dataMigrationsConfig, v1Server.Storage(), vChan)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not initialize v2 policy server")
@@ -88,7 +90,7 @@ func NewGRPCServer(ctx context.Context,
 	}
 
 	v2ProjectsServer, err := v2.NewPostgresProjectsServer(ctx, l, migrationsConfig,
-		dataMigrationsConfig, e, eventServiceClient, configManager)
+		dataMigrationsConfig, e, eventServiceClient, configManager, policyRefresher)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not initialize v2 projects server")
 	}
--- a/components/authz-service/server/v2/policy_refresher.go
+++ b/components/authz-service/server/v2/policy_refresher.go
@@ -306,7 +306,7 @@ func NewMockPolicyRefresher() PolicyRefresher {
 	return &mockPolicyRefresher{}
 }
 
-func (refresher *mockPolicyRefresher) Refresh(context.Context) error {
+func (*mockPolicyRefresher) Refresh(context.Context) error {
 	return nil
 }
 
--- a/components/authz-service/server/v2/policy_refresher.go
+++ b/components/authz-service/server/v2/policy_refresher.go
@@ -299,17 +299,3 @@ func (refresher *policyRefresher) getRuleMap(ctx context.Context) (map[string][]
 func pretty(vsn api.Version) string {
 	return fmt.Sprintf("IAM v%d.%d", int32(vsn.Major), int32(vsn.Minor))
 }
-
-type mockPolicyRefresher struct{}
-
-func NewMockPolicyRefresher() PolicyRefresher {
-	return &mockPolicyRefresher{}
-}
-
-func (*mockPolicyRefresher) Refresh(context.Context) error {
-	return nil
-}
-
-func (refresher *mockPolicyRefresher) RefreshAsync() error {
-	return nil
-}
--- a/components/authz-service/server/v2/policy_refresher.go
+++ b/components/authz-service/server/v2/policy_refresher.go
@@ -299,3 +299,17 @@ func (refresher *policyRefresher) getRuleMap(ctx context.Context) (map[string][]
 func pretty(vsn api.Version) string {
 	return fmt.Sprintf("IAM v%d.%d", int32(vsn.Major), int32(vsn.Minor))
 }
+
+type mockPolicyRefresher struct{}
+
+func NewMockPolicyRefresher() PolicyRefresher {
+	return &mockPolicyRefresher{}
+}
+
+func (refresher *mockPolicyRefresher) Refresh(context.Context) error {
+	return nil
+}
+
+func (refresher *mockPolicyRefresher) RefreshAsync() error {
+	return nil
+}
--- a/components/authz-service/server/v2/project.go
+++ b/components/authz-service/server/v2/project.go
@@ -30,12 +30,12 @@ import (
 	event "github.com/chef/automate/components/event-service/server"
 )
 
-// state the server state for projects
-type state struct {
+// ProjectState the server state for projects
+type ProjectState struct {
 	log                  logger.Logger
 	store                storage.Storage
 	engine               engine.ProjectRulesRetriever
-	projectUpdateManager *ProjectUpdateManager
+	ProjectUpdateManager *ProjectUpdateManager
 	policyRefresher      PolicyRefresher
 	applyRuleMux         sync.Mutex
 }
@@ -82,16 +82,16 @@ func NewProjectsServer(
 	pr PolicyRefresher,
 ) (api.ProjectsServer, error) {
 
-	return &state{
+	return &ProjectState{
 		log:                  l,
 		store:                s,
 		engine:               e,
-		projectUpdateManager: NewProjectUpdateManager(eventServiceClient, configManager),
+		ProjectUpdateManager: NewProjectUpdateManager(eventServiceClient, configManager),
 		policyRefresher:      pr,
 	}, nil
 }
 
-func (s *state) GetProject(ctx context.Context,
+func (s *ProjectState) GetProject(ctx context.Context,
 	req *api.GetProjectReq) (*api.GetProjectResp, error) {
 	p, err := s.store.GetProject(ctx, req.Id)
 	if err != nil {
@@ -110,7 +110,7 @@ func (s *state) GetProject(ctx context.Context,
 	return &api.GetProjectResp{Project: apiProject}, nil
 }
 
-func (s *state) CreateProject(ctx context.Context,
+func (s *ProjectState) CreateProject(ctx context.Context,
 	req *api.CreateProjectReq) (*api.CreateProjectResp, error) {
 	p, err := storage.NewProject(req.Id, req.Name, storage.Custom)
 	if err != nil {
@@ -137,7 +137,7 @@ func (s *state) CreateProject(ctx context.Context,
 	return &api.CreateProjectResp{Project: apiProject}, nil
 }
 
-func (s *state) UpdateProject(ctx context.Context,
+func (s *ProjectState) UpdateProject(ctx context.Context,
 	req *api.UpdateProjectReq) (*api.UpdateProjectResp, error) {
 	p, err := storage.NewProject(req.Id, req.Name, storage.Custom)
 	if err != nil {
@@ -163,7 +163,7 @@ func (s *state) UpdateProject(ctx context.Context,
 	return &api.UpdateProjectResp{Project: apiProject}, nil
 }
 
-func (s *state) ApplyRulesStart(
+func (s *ProjectState) ApplyRulesStart(
 	ctx context.Context, _ *api.ApplyRulesStartReq) (*api.ApplyRulesStartResp, error) {
 	// NOTE (tc): Only one call to ApplyRulesStart can happen at a time.
 	// This should be good enough to prevent race conditions for single node,
@@ -173,7 +173,7 @@ func (s *state) ApplyRulesStart(
 	s.applyRuleMux.Lock()
 	defer s.applyRuleMux.Unlock()
 
-	switch s.projectUpdateManager.State() {
+	switch s.ProjectUpdateManager.State() {
 	case config.NotRunningState:
 		break
 	case config.RunningState:
@@ -201,7 +201,7 @@ func (s *state) ApplyRulesStart(
 			"error refreshing policy cache. the rules were updated but the apply was not started, please try again.")
 	}
 
-	err = s.projectUpdateManager.Start()
+	err = s.ProjectUpdateManager.Start()
 	if err != nil {
 		s.log.Warnf("error starting project update: %s", err.Error())
 		return nil, status.Errorf(codes.Internal,
@@ -211,33 +211,33 @@ func (s *state) ApplyRulesStart(
 	return &api.ApplyRulesStartResp{}, nil
 }
 
-func (s *state) ApplyRulesCancel(
+func (s *ProjectState) ApplyRulesCancel(
 	context.Context, *api.ApplyRulesCancelReq) (*api.ApplyRulesCancelResp, error) {
 	s.log.Info("apply project rules: CANCEL")
-	err := s.projectUpdateManager.Cancel()
+	err := s.ProjectUpdateManager.Cancel()
 	if err != nil {
 		s.log.Errorf("Could not cancel project update: %v", err.Error())
 	}
 	return &api.ApplyRulesCancelResp{}, nil
 }
 
-func (s *state) ApplyRulesStatus(
+func (s *ProjectState) ApplyRulesStatus(
 	context.Context, *api.ApplyRulesStatusReq) (*api.ApplyRulesStatusResp, error) {
-	time, err := ptypes.TimestampProto(s.projectUpdateManager.EstimatedTimeComplete())
+	time, err := ptypes.TimestampProto(s.ProjectUpdateManager.EstimatedTimeComplete())
 	if err != nil {
 		s.log.Errorf("Could not convert EstimatedTimeComplete to protobuf Timestamp %v", err)
 		time = &tspb.Timestamp{}
 	}
 	return &api.ApplyRulesStatusResp{
-		State:                 s.projectUpdateManager.State(),
-		PercentageComplete:    float32(s.projectUpdateManager.PercentageComplete()),
+		State:                 s.ProjectUpdateManager.State(),
+		PercentageComplete:    float32(s.ProjectUpdateManager.PercentageComplete()),
 		EstimatedTimeComplete: time,
-		Failed:                s.projectUpdateManager.Failed(),
-		FailureMessage:        s.projectUpdateManager.FailureMessage(),
+		Failed:                s.ProjectUpdateManager.Failed(),
+		FailureMessage:        s.ProjectUpdateManager.FailureMessage(),
 	}, nil
 }
 
-func (s *state) ListProjects(
+func (s *ProjectState) ListProjects(
 	ctx context.Context, _ *api.ListProjectsReq) (*api.ListProjectsResp, error) {
 	ps, err := s.store.ListProjects(ctx)
 	if err != nil {
@@ -265,7 +265,7 @@ func (s *state) ListProjects(
 	return &resp, nil
 }
 
-func (s *state) ListProjectsForIntrospection(
+func (s *ProjectState) ListProjectsForIntrospection(
 	ctx context.Context, req *api.ListProjectsReq) (*api.ListProjectsResp, error) {
 
 	// Introspection needs unfiltered access.
@@ -297,7 +297,7 @@ func (s *state) ListProjectsForIntrospection(
 	return &resp, nil
 }
 
-func (s *state) DeleteProject(ctx context.Context,
+func (s *ProjectState) DeleteProject(ctx context.Context,
 	req *api.DeleteProjectReq) (*api.DeleteProjectResp, error) {
 	err := s.store.DeleteProject(ctx, req.Id)
 	switch err {
@@ -311,7 +311,7 @@ func (s *state) DeleteProject(ctx context.Context,
 	}
 }
 
-func (s *state) ListRulesForAllProjects(ctx context.Context,
+func (s *ProjectState) ListRulesForAllProjects(ctx context.Context,
 	req *api.ListRulesForAllProjectsReq) (*api.ListRulesForAllProjectsResp, error) {
 
 	ruleMap, err := s.engine.ListProjectMappings(ctx)
@@ -339,18 +339,18 @@ func (s *state) ListRulesForAllProjects(ctx context.Context,
 	}, nil
 }
 
-func (s *state) HandleEvent(ctx context.Context,
+func (s *ProjectState) HandleEvent(ctx context.Context,
 	req *automate_event.EventMsg) (*automate_event.EventResponse, error) {
 	s.log.Debugf("authz is handling your event %s", req.EventID)
 
 	response := &automate_event.EventResponse{}
 	if req.Type.Name == event.ProjectRulesUpdateStatus {
-		err := s.projectUpdateManager.ProcessStatusEvent(req)
+		err := s.ProjectUpdateManager.ProcessStatusEvent(req)
 		if err != nil {
 			return response, err
 		}
 	} else if req.Type.Name == event.ProjectRulesUpdateFailed {
-		err := s.projectUpdateManager.ProcessFailEvent(req)
+		err := s.ProjectUpdateManager.ProcessFailEvent(req)
 		if err != nil {
 			return response, err
 		}
@@ -359,7 +359,7 @@ func (s *state) HandleEvent(ctx context.Context,
 	return response, nil
 }
 
-func (s *state) CreateRule(ctx context.Context, req *api.CreateRuleReq) (*api.CreateRuleResp, error) {
+func (s *ProjectState) CreateRule(ctx context.Context, req *api.CreateRuleReq) (*api.CreateRuleResp, error) {
 	r, err := s.prepareStorageRule(req.Id, req.ProjectId, req.Name, req.Type, req.Conditions)
 	if err != nil {
 		return nil, status.Errorf(codes.InvalidArgument, "error processing request: %s", err.Error())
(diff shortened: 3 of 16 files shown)


components/authz-service/server/v2/project.go
```diff
@@ -158,9 +163,22 @@ func (s *state) UpdateProject(ctx context.Context,
...
 	s.log.Info("apply project rules: START")
-	err := s.projectUpdateManager.Start()
+
+	err := s.store.ApplyStagedRules(ctx)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal,
+			"error applying staged projects: %s", err.Error())
+	}
+
+	err = s.policyRefresher.Refresh(ctx)
```
srenatus: Uhm I don't get it. What hinders us from refreshing in the store's `ApplyStagedRules` method?

[Looks like we do that](https://github.com/chef/automate/pull/655/commits/b45ee71418d0c8325fc65162db54f088949f3889#diff-d28b0cc9c342bf1da873bddba25960f6R1347). I don't think I understand why this is necessary then.

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1274,6 +1269,94 @@ func (p *pg) ListRulesForProject(ctx context.Context, projectID string) ([]*v2.R
...
+					LEFT OUTER JOIN iam_staged_rule_conditions AS cond
+					ON rule_db_id=r.db_id
+					WHERE r.id=$1;`,
+			id, dbID,
+		)
+		if err != nil {
+			return p.processError(err)
+		}
+	}
+
```
srenatus: ❓ Shouldn't something else happen here, too? Like, do the re-tagging stuff in the domain services? Not saying that this should belong in this PR, just wondering how we'd architect that.

⚠️ There's _nothing_ ensuring that this GRPC method isn't called twice concurrently; the transaction will only ensure that one of the `n` calls succeeds -- so, if we do something to some other service, that case needs to be considered.

components/authz-service/server/v2/project.go
tylercloke: Because that method is async whereas this is sync. You run into race conditions if you don't use this method.

components/authz-service/server/v2/policy_refresher.go
```diff
@@ -299,3 +299,17 @@ func (refresher *policyRefresher) getRuleMap(ctx context.Context) (map[string][]
...
 	return fmt.Sprintf("IAM v%d.%d", int32(vsn.Major), int32(vsn.Minor))
 }
+
+type mockPolicyRefresher struct{}
+
+func NewMockPolicyRefresher() PolicyRefresher {
+	return &mockPolicyRefresher{}
+}
+
+func (refresher *mockPolicyRefresher) Refresh(context.Context) error {
```
srenatus: [nit] no need to name the receiver if it's not used
```suggestion
func (*mockPolicyRefresher) Refresh(context.Context) error {
```

components/authz-service/server/v2/policy_refresher.go
```diff
@@ -299,3 +299,17 @@ func (refresher *policyRefresher) getRuleMap(ctx context.Context) (map[string][]
 func pretty(vsn api.Version) string {
 	return fmt.Sprintf("IAM v%d.%d", int32(vsn.Major), int32(vsn.Minor))
 }
+
+type mockPolicyRefresher struct{}
```
srenatus: ❓ Can we defined this in `server/v2/policy_test.go` instead? Won't make it into prod code that way. (Not that it does much harm here, but... just for the sake of it)

components/authz-service/server/v2/project.go
```diff
@@ -158,13 +164,47 @@ func (s *state) UpdateProject(ctx context.Context,
...
-	s.log.Info("apply project rules: START")
-	err := s.projectUpdateManager.Start()
+	ctx context.Context, _ *api.ApplyRulesStartReq) (*api.ApplyRulesStartResp, error) {
+	// NOTE (tc): Only one call to ApplyRulesStart can happen at a time.
+	// This should be good enough to prevent race conditions for single node,
+	// in conjunction with the table locking that happens in store.ApplyStagedRules.
+	// Can still get into a weird state if we panic, but a re-apply will fix things up.
+	// We will be refactoring for multi-node using workflow tooling in the near future.
+	s.applyRuleMux.Lock()
+	defer s.applyRuleMux.Unlock()
```
srenatus: Interesting approach. I suppose we could also have gone with some process (goroutine) reading from a request channel, serving one request at a time, and pushing the response into a response channel. But this is fine, I suppose.

components/authz-service/server/v2/project.go
```diff
@@ -158,13 +164,47 @@ func (s *state) UpdateProject(ctx context.Context,
...
+			"error applying staged projects: %s", err.Error())
+	}
+
+	// TODO (tc): If we panic between here and manager Start, we will be in a state where the rules
+	// have been updated in the database but Refresh has not been kicked off.
+	// We will be refactoring with workflow to make this safer soon.
+	err = s.policyRefresher.Refresh(ctx)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal,
+			"error refreshing policy cache: %s\nthe rules were updated but the apply was not started, please try again", err.Error())
```
srenatus: [nit] Newlines? Let's not do that. Let's log if we need more info, but multi-line error messages are weird. 😉

components/authz-service/storage/v2/memstore/memstore.go
```diff
@@ -325,6 +325,11 @@ func (s *State) DeleteRule(ctx context.Context, id string) error {
 	return nil
 }
 
+func (s *State) ApplyStagedRules(ctx context.Context) error {
```
srenatus: [nit] 
```suggestion
func (*State) ApplyStagedRules(context.Context) error {
```

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1267,6 +1262,105 @@ func (p *pg) ListRulesForProject(ctx context.Context, projectID string) ([]*v2.R
...
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	tx, err := p.db.BeginTx(ctx, nil)
+	if err != nil {
+		return p.processError(err)
+	}
+
+	_, err = tx.ExecContext(ctx,
+		`LOCK TABLE iam_project_rules; LOCK TABLE iam_rule_conditions;
```
srenatus: ❓Correct me if I'm wrong, but _locking the table_ means there's no way for a concurrent request to create a new rule while this is running. The alternative would be to lock existing rows in `iam_staged_project_rules`, maybe? [There's a dozen ways to do locking](https://www.postgresql.org/docs/9.4/explicit-locking.html)... but if this does the trick, I suppose it's fine.

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1267,6 +1262,105 @@ func (p *pg) ListRulesForProject(ctx context.Context, projectID string) ([]*v2.R
...
+	rows, err := tx.QueryContext(ctx,
+		`INSERT INTO iam_project_rules (id, project_id, name, type)
+				SELECT s.id, s.project_id, s.name, s.type
+					FROM iam_staged_project_rules AS s
+					WHERE deleted=false
+				ON CONFLICT (id) DO UPDATE
+				SET name=excluded.name, type=excluded.type
+				RETURNING id, db_id;`)
+	defer func() {
+		if err := rows.Close(); err != nil {
```
srenatus: I believe this needs to go below the `err != nil`; if `err == nil` in line 1301, the defer here would run and call `Close()` on `rows == nil`

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1267,6 +1262,105 @@ func (p *pg) ListRulesForProject(ctx context.Context, projectID string) ([]*v2.R
...
+			return p.processError(err)
+		}
+	}
+
+	_, err = tx.ExecContext(ctx,
+		`DELETE FROM iam_project_rules WHERE id IN(
+			SELECT a.id FROM iam_project_rules AS a
+			LEFT OUTER JOIN iam_staged_project_rules AS s
+			ON s.id=a.id
+			WHERE s.deleted=true);`)
```
srenatus: [nit] sql style question, we could do this
```suggestion
			WHERE s.deleted)`)
```

but that is fine, too.

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -4384,6 +4384,64 @@ func TestDeleteRule(t *testing.T) {
 	}
 }
 
+func TestApplyStagedRules(t *testing.T) {
+	store, db, _ := testhelpers.SetupTestDB(t)
+	defer db.CloseDB(t)
+	defer store.Close()
+	// No project filter concerns in these tests so safe to re-use context.
```
bcmdarroch: meaning we don't pass projects to this API or we ignore them if passed?

tylercloke: We don't check the projects header at all for this API.

components/authz-service/storage/v2/postgres/postgres.go
tylercloke: We don't want people to be able to update or create anything while this code is in progress.

components/authz-service/storage/v2/memstore/memstore.go
```diff
@@ -339,6 +339,11 @@ func (s *State) DeleteRule(ctx context.Context, id string) error {
 	return nil
 }
 
+func (*State) ApplyStagedRules(context.Context) error {
+	// TODO
```
bcmdarroch: should this TODO be done before this gets merged?

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -4384,6 +4384,64 @@ func TestDeleteRule(t *testing.T) {
...
+			projID := "project-1"
+			insertTestProject(t, db, projID, "let's go jigglypuff - topsecret", storage.Custom)
+			ruleType := storage.Node
+			rule1 := insertStagedRuleWithMultipleConditions(t, db, projID, ruleType)
+			condition, err := storage.NewCondition(ruleType,
+				[]string{"chef-server-2"}, storage.ChefServer, storage.MemberOf)
+			rule2, err := storage.NewRule("new-id-2", projID, "name2", ruleType,
+				[]storage.Condition{condition})
+			require.NoError(t, err)
+			insertStagedRule(t, db, rule2)
```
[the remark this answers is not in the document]
bcmdarroch: (this says outdated cuz i commented on an older commit but still needs to be changed btw)

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1245,6 +1240,105 @@ func (p *pg) ListRulesForProject(ctx context.Context, projectID string) ([]*v2.R
...
+			return p.processError(err)
+		}
+	}
+
+	_, err = tx.ExecContext(ctx,
+		`DELETE FROM iam_project_rules WHERE id IN(
+			SELECT a.id FROM iam_project_rules AS a
+			LEFT OUTER JOIN iam_staged_project_rules AS s
+			ON s.id=a.id
+			WHERE s.deleted)`)
```
msorens: Outer join does not quite make sense here, but actually do not need a join at all. I believe this will suffice:
```
`DELETE FROM iam_project_rules
   WHERE id IN
     (SELECT id FROM iam_staged_project_rules WHERE deleted)`
```

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1245,6 +1240,105 @@ func (p *pg) ListRulesForProject(ctx context.Context, projectID string) ([]*v2.R
...
+	for id, dbID := range ids {
+		_, err = tx.ExecContext(ctx,
+			`DELETE FROM iam_rule_conditions WHERE rule_db_id=$1;`, dbID)
+		if err != nil {
+			return p.processError(err)
+		}
+
+		_, err = tx.ExecContext(ctx,
+			`INSERT INTO iam_rule_conditions (rule_db_id, value, attribute, operator)
+					SELECT $2, cond.value, cond.attribute, cond.operator FROM iam_staged_project_rules AS r
```
msorens: Suggest line break before "FROM..."

components/authz-service/server/v2/project.go
tylercloke: `notifyPolicyChange` is async. It doesn't wait until the OPA cache is updated. Thus, we need the `policyRefresher.Refresh` call, because we need the OPA cache to be updated before `projectUpdateManager.Start`.

components/authz-service/storage/v2/memstore/memstore.go
tylercloke: I think it can be removed.

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1245,6 +1240,105 @@ func (p *pg) ListRulesForProject(ctx context.Context, projectID string) ([]*v2.R
...
+	defer cancel()
+
+	tx, err := p.db.BeginTx(ctx, nil)
+	if err != nil {
+		return p.processError(err)
+	}
+
+	_, err = tx.ExecContext(ctx,
+		`LOCK TABLE iam_project_rules; LOCK TABLE iam_rule_conditions;
+			LOCK TABLE iam_staged_project_rules; LOCK TABLE iam_staged_rule_conditions; `,
```
msorens: Easier to read with 4 lines (one LOCK stmt per line).

components/authz-service/server/v2/project.go
```diff
@@ -158,13 +164,50 @@ func (s *state) UpdateProject(ctx context.Context,
...
+
+	switch s.projectUpdateManager.State() {
+	case config.NotRunningState:
+		break
+	case config.RunningState:
+		return nil, status.Error(codes.FailedPrecondition,
+			"cannot apply rules: apply already in progress")
+	default:
+		return nil, status.Error(codes.Internal,
+			"failed to parse apply state")
```
blakestier: super nit: I'd maybe include the word `rules` or `project rules` or something in addition to `apply` here as `apply` feels vague-- but I don't have a mental model rn of what the error message would be

components/authz-service/server/v2/project.go
blakestier: @tylercloke you mean the gocache . . .right? I still don't think we need to bump the OPA cache unless a project is deleted-- and we _should_ probably worry about that in addition to rules changing, but we don't currently. . . 😳

that said, I know we've woven the OPA cache bumping with the gocache so it's kinda moot.

components/authz-service/server/v2/project.go
```diff
@@ -158,13 +164,50 @@ func (s *state) UpdateProject(ctx context.Context,
...
+	}
+
+	// TODO (tc): If we panic between here and manager Start, we will be in a state where the rules
+	// have been updated in the database but Refresh has not been kicked off.
+	// We will be refactoring with workflow to make this safer soon.
+	err = s.policyRefresher.Refresh(ctx)
+	if err != nil {
+		s.log.Warnf("error refreshing policy cache: %s", err.Error())
+		return nil, status.Errorf(codes.Internal,
+			"error refreshing policy cache. the rules were updated but the apply was not started, please try again.")
```
blakestier: this seems like an extra long error message. I'd push the err.Error() here and the longer message into the logs.

components/authz-service/storage/v2/postgres/postgres.go
blakestier: I'm not sure what kind of lock this executes without a mode specified. Will it allow `READ` access? We want other services (compliance) to be able to access the existing rules, no?

blakestier:
I had a few questions and I feel really uncomfortable merging this without any automated tests, even though repro steps are 👍

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -4515,6 +4518,173 @@ func TestDeleteRule(t *testing.T) {
...
+			assertCount(t, 1, db.QueryRow(`SELECT count(*) FROM iam_staged_project_rules WHERE id=$1`, rule2.ID))
+			assertCount(t, 4, db.QueryRow(`SELECT count(*) FROM iam_staged_rule_conditions`))
+
+			err = store.ApplyStagedRules(ctx)
+			assert.NoError(t, err)
+			assertEmpty(t, db.QueryRow(`SELECT count(*) FROM iam_staged_project_rules`))
+			assertEmpty(t, db.QueryRow(`SELECT count(*) FROM iam_staged_rule_conditions`))
+			assertCount(t, 1, db.QueryRow(`SELECT count(*) FROM iam_project_rules WHERE id=$1`, rule1.ID))
+			assertCount(t, 1, db.QueryRow(`SELECT count(*) FROM iam_project_rules WHERE id=$1`, rule2.ID))
+			assertCount(t, 4, db.QueryRow(`SELECT count(*) FROM iam_rule_conditions`))
```
blakestier: I'd feel better if we made these assertions do more than count the total conditions (looked for conditions in the shape of the staged ones)

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1245,6 +1240,105 @@ func (p *pg) ListRulesForProject(ctx context.Context, projectID string) ([]*v2.R
...
+
+	// Upsert all staged rules into applied rules marked for update, returning the id and db_id
+	// of all rules affected so we can update their conditions below.
+	rows, err := tx.QueryContext(ctx,
+		`INSERT INTO iam_project_rules (id, project_id, name, type)
+				SELECT s.id, s.project_id, s.name, s.type
+					FROM iam_staged_project_rules AS s
+					WHERE deleted=false
+				ON CONFLICT (id) DO UPDATE
+				SET name=excluded.name, type=excluded.type
```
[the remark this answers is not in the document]
blakestier: We also have security against a rule with a changed type going into the staged table, fwiw

components/authz-service/server/v2/project.go
blakestier: (this is probably fine, just seems like something else should be handling such granular messaging)

components/authz-service/server/server.go
```diff
@@ -71,7 +71,9 @@ func NewGRPCServer(ctx context.Context,
 		return nil, errors.Wrap(err, "could not initialize v1 server")
 	}
 
-	v2PolServer, err := v2.NewPostgresPolicyServer(ctx, l, e, migrationsConfig,
+	// TODO (tc) Refactor how singletons are shared between GRPC servers. Should we also
```
srenatus: 💭 This is what the _rearchitect authz-service_ card is about, too. Suggestions welcome 😉

components/authz-service/server/v2/project.go
jaym: assuming i'm remembering things correctly from our conversation:
there are 2 parts to this dance. 1 is updating the database, the other is synchronizing the OPA store with the database. Updating the database will at some point in the future trigger replication. This is asynchronous. If you want to change the database and update the OPA store, you need to call refresh on the policy refresher thing because the database layer doesn't know about the OPA layer.

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1245,6 +1240,105 @@ func (p *pg) ListRulesForProject(ctx context.Context, projectID string) ([]*v2.R
...
+	ids := make(map[string]string)
+	for rows.Next() {
+		var id string
+		var dbID string
+		err = rows.Scan(&id, &dbID)
+		if err != nil {
+			return p.processError(err)
+		}
+		ids[id] = dbID
+	}
```
stevendanna: [errcheck] We are missing an error check here, but it is sneaky.  rows.Next() returns `false` on error so you need to check rows.Err() after the end of the for loop.

components/authz-service/server/v2/system_integration_test.go
```diff
@@ -317,23 +226,165 @@ func TestIntegrationListRulesForAllProjects(t *testing.T) {
...
+				&api_v2.ProjectRule{
+					Id:         createResp2.Rule.Id,
+					ProjectId:  createResp2.Rule.ProjectId,
+					Name:       "this has been updated",
+					Type:       createResp2.Rule.Type,
+					Conditions: newConditions,
+					Status:     "applied",
+				},
+			}
+			eexpectedAfterApplyRules2 := []*api_v2.ProjectRule{createResp3.Rule}
```
bcmdarroch: s/eexpected/expected

i've lost track of which of these responses should be missing the deleted rule

components/authz-service/server/v2/system_integration_test.go
```diff
@@ -118,91 +144,314 @@ func TestListRulesForAllProjects(t *testing.T) {
...
-func setupWithOPAV2(t *testing.T) testSetup {
+func rulesEqual(t *testing.T, expected, actual []*api_v2.ProjectRule) {
+	t.Helper()
+	if len(expected) != len(actual) {
+		t.FailNow()
+		fmt.Printf("expected len %d != actual len %d", len(expected), len(actual))
+	}
+
+	for i, expectedRule := range expected {
+		actualRule := actual[i]
```
msorens: While it might work coincidentally, relying on the DB to return rules in the same order makes the tests fragile. Consider turning one into a hash on the id so you could then iterate the other and find the right rule, then compare them.

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -4614,8 +4617,175 @@ func TestDeleteRule(t *testing.T) {
...
+			rule := insertAppliedRuleWithMultipleConditions(t, db, projID, storage.Node)
+			assertCount(t, 1, db.QueryRow(`SELECT count(*) FROM iam_project_rules
+				WHERE id=$1 AND project_id=$2 AND name=$3 AND type=$4`, rule.ID, rule.ProjectID, rule.Name, rule.Type.String()))
+
+			err := store.ApplyStagedRules(ctx)
+			assert.NoError(t, err)
+			assertCount(t, 1, db.QueryRow(`SELECT count(*) FROM iam_project_rules
+				WHERE id=$1 AND project_id=$2 AND name=$3 AND type=$4`, rule.ID, rule.ProjectID, rule.Name, rule.Type.String()))
+		}},
+		{"when there are n staged rules marked for update but no applied rules, it creates n staged rules", func(t *testing.T) {
```
msorens: Test name does not sound right; should it end with "... it creates n applied rules" ?

components/authz-service/server/v2/system_integration_test.go
```diff
@@ -50,49 +56,69 @@ func TestSystemPolicies(t *testing.T) {
 
 	for desc, test := range cases {
 		t.Run(desc, test)
+		ts.TestDB.Flush(t)
 	}
+
+	ts.Shutdown(t, ctx)
```
msorens: You mention below potential problems if this is not called... consider making this a `defer` call then to guarantee it gets called.

PR #656, opened 2019-06-21
[nodemanager] Avoid nil pointer dereference on bad job type

stevendanna:
Previously `updated` would be nil if the job type or status was not
recognized, leading to a nil pointer dereference later in this
function.


stevendanna:
This doesn't happen in practice, but we found it during development in #595 and it caused us some confusion.

PR #657, opened 2019-06-21
ensure we get the actual value for end time when calculating start of day for start_time

vjeffrey:
As noted in https://github.com/chef/automate/pull/639#issuecomment-504463446, a recent change we made was setting the start time on export to beginning of today, instead of beginning of day of end time.
I discovered we were getting the end time object, but not the value, and so moment was just setting start time to the beginning of today's date.
This change ensures that we're getting the actual value.

Screen Shot 2019-06-21 at 12 59 44
Screen Shot 2019-06-21 at 12 59 53

### :+1: Definition of Done
start time is set to beg of day of whatever value is the end time

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
+++ b/components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
@@ -178,7 +178,7 @@ export class StatsService {
     const endtime = filters.find(filter => filter['end_time']);
     const filtersCopy = filters.map(filter => {
       if (filter['start_time']) {
-        filter['start_time'] = moment(endtime).startOf('day');
+        filter['start_time'] = moment(endtime['end_time']).startOf('day');
       }
       return filter;
     });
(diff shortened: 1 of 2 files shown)


vjeffrey:
looking at what kind of test i can add for this..

components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
```diff
@@ -178,7 +178,7 @@ export class StatsService {
     const endtime = filters.find(filter => filter['end_time']);
     const filtersCopy = filters.map(filter => {
       if (filter['start_time']) {
-        filter['start_time'] = moment(endtime).startOf('day');
+        filter['start_time'] = moment(endtime['end_time']).startOf('day');
```
phiggins: Is `filters` an array of maps? This makes sense then.

phiggins:
> looking at what kind of test i can add for this..

There's tests for `formatFilters`, you could do a similar thing if you extracted the start_time logic to a function that wrapped `formatFilters` and then test that the result had the proper `start_time` set.

https://github.com/chef/automate/blob/8cc636c3853396c19e9c79d5f6e0b28b9780fff0/components/automate-ui/src/app/pages/%2Bcompliance/shared/reporting/stats.service.spec.ts#L118

PR #658, opened 2019-06-21
[deployment] Thread contexts through `hab` command I/O

### LINKED ISSUE #659: Support HAB .8x in deployment-service
While we are here:

- Use a context for all hab command invocations

Fixed by #658

ryancragun:

We have several I/O operations that interact with the hab supervisor through the `hab` executable. This builds on the unmerged work in https://github.com/chef/automate/pull/5 but uses contexts with deadlines instead of timeouts. This brings us one step closer to threading a single context with deadlines through the entire bootstrap chain.

This also adds support for `hab` > 0.79, which was previously broken because of behavior changes related to service unloading.

Fixes #659


components/automate-deployment/pkg/client/restore.go
```diff
@@ -187,60 +187,73 @@ func (r *DeploymentRestore) loadService() error {
...
-		deployed, err := r.target.DeployedServices(ctx)
-		if err == nil {
-			alive := false
-			for _, svc := range deployed {
-				if svc.Pkg.Name() == deploymentServiceName {
-					alive = true
-				}
-			}
+	deploymentServiceRunning := func() (bool, error) {
+		var (
```
jaym: does this all need to be forward declared?

components/automate-deployment/pkg/client/restore.go
```diff
@@ -187,60 +187,73 @@ func (r *DeploymentRestore) loadService() error {
...
-	for {
-		deployed, err := r.target.DeployedServices(ctx)
-		if err == nil {
-			alive := false
-			for _, svc := range deployed {
-				if svc.Pkg.Name() == deploymentServiceName {
-					alive = true
-				}
-			}
+	deploymentServiceRunning := func() (bool, error) {
```
jaym: i like this. below, it reads much nicer than what it was before as the intent becomes very clear

components/automate-deployment/pkg/client/restore.go
ryancragun: I'm was going to write a thesis on why forward declaring variables in the code is 10 million times better than relying on editors/IDE's/gopsl for type informtion because I don't have to context switch to see what things are, especially since the compiler doesn't care if you forward declare things, but instead I'll just leave this comment. :dancer:

components/automate-deployment/pkg/converge/compiler_test.go
```diff
@@ -75,7 +75,7 @@ func NoSupervisorUpgradeRequired(m *target.MockTarget) *target.MockTarget {
 	m.On("RenderAutomateUnitFile", "", testHabBinPkg, testHabLauncherPkg).Return(string(testAutomateUnitFileContent), nil)
```
jaym: this would be a good time to let this file die

ryancragun: I require no convincing

components/automate-deployment/pkg/target/hab_cmd.go
```diff
@@ -14,35 +18,50 @@ var stdHabOptions = []command.Opt{
 	command.Envvar("HAB_LICENSE", "accept-no-persist"),
 }
 
+const (
+	// HabTimeoutInstallPackage is the timeout for InstallPackage
+	// commands. Since package installs also install dependencies,
+	// a given package installation can often take considerable
+	// time.
+	HabTimeoutInstallPackage = 1200 * time.Second
```
jaym: i say we do 1 hour

components/automate-deployment/pkg/target/hab_cmd.go
```diff
@@ -120,37 +146,76 @@ func (c *habCmd) IsInstalled(pkg habpkg.VersionedPackage) (bool, error) {
...
+
+	cmdOpts := append(standardHabOptions(), command.Context(ctx), args)
+	output, err := c.executor.CombinedOutput("hab", cmdOpts...)
+
+	// NOTE: hab <= 0.79 returned 0 if the service was already unloaded.
+	// Beginning with 0.80, hab will exit 1 if the service was already unloaded.
+	// As we need to support mixed versions because of upgrades we'll do our
+	// best to handle backwards compatibility of this interface.
+	if err != nil {
+		regex := regexp.MustCompile("not loaded")
```
jaym: can we pull this out into a global variable so it fails to even start if someone messes it up

components/automate-deployment/pkg/target/local_target.go
```diff
@@ -334,25 +334,29 @@ func (t *LocalTarget) SetUserToml(name, config string) error {
...
-	return t.installPackage(svc, channel)
+func (t *LocalTarget) InstallService(ctx context.Context, svc habpkg.Installable, channel string) error {
+	return t.installPackage(ctx, svc, channel)
 }
 
 // unloadServiceWithHabVersion unloads the given service with a
 // particular version of hab and hab-sup. This is used during the
 // Habitat upgrade process to unload services before upgrading hab.
-func (t *LocalTarget) unloadServiceWithHabVersion(svc habpkg.VersionedPackage, binPkg habpkg.HabPkg, habSupPkg habpkg.HabPkg) error {
+func (t *LocalTarget) unloadServiceWithHabVersion(ctx context.Context, svc habpkg.VersionedPackage, binPkg habpkg.HabPkg, habSupPkg habpkg.HabPkg) error {
```
jaym: ummmm, any idea if it's possible to get rid of this? If not, we maybe should take your fix from above?

components/automate-deployment/pkg/target/local_target.go
```diff
@@ -787,20 +785,34 @@ func (t *LocalTarget) HabSupRestart(sortedServiceList []string) (bool, error) {
...
+				errC <- err
+				break
+			}
+
+			time.Sleep(500 * time.Millisecond)
 		}
-		time.Sleep(500 * time.Millisecond)
+	}
+
+	go unloadService()
```
jaym: I'm not sure about refactoring this to use a goroutine. if the context timed out before, the "long running" ServiceInfo call would bail, so I think it was already correct.
With the goroutine version, I think we end up with cases where we can leak the goroutine. For example, if the unloadService function gets an error, and then the context times out and we return `ctx.Err()`, i think the goroutine will never be able to write to the err channel because there is no listener.

components/automate-deployment/pkg/client/restore.go
jaym: fair enough

components/automate-deployment/pkg/target/local_target.go
ryancragun: Good catch! Whether or not we need to worry about unloading with a specific hab/hab-sup I can't speak to. As per using the same implementation under the hood, not really unless we either add a new function to the HabCmd interface or modify the existing one to require specifying them. I think the former is probably better long term, but for now I think just patching both functions is the least intrusive until we want to do another refactor.

stevendanna: I believed that to get rid of this we need to stop supporting upgrades from old versions of deployment-service. Namely, this is used so that we can unload services from hab-sup even if the running hab-sup is very old.

stevendanna:
I hit approve, but I did leave one question maybe worth double-checking

PR #661, opened 2019-06-22
automate-load-balancer += http2

srenatus:
In @msorens' performance investigations, he observed that some latency is introduced by the fact that a browser only open _n_ TCP connections to one origin (_n=6_ for Chrome, it seems). This is unfortunate, but since HTTP2 solves that, by pipelining multiple requests in a single connection, it seemed worthwhile to give that a try.

And indeed, it seems that the specific delay ("stalled" in the dev tools' details view) goes down significantly when using HTTP2.

So, let's do this 😉 

Note that for the problem at hand here, internal connections don't matter as much -- so this only enables HTTP2 at the edge.

✅ Added a simple test, which outputs nothing on success. But glancing at `sup.log`, we find
```
automate-load-balancer.default [25/Jun/2019:12:47:02 +0000]  "GET / HTTP/2.0" 200 "0.002" 2016 "-" "curl/7.63.0" "172.17.0.2:10161" "200" "0.000" 27
```
So, it's doing its thing ✔️

### DIFF
--- a/.expeditor/verify_private.pipeline.yml
+++ b/.expeditor/verify_private.pipeline.yml
@@ -109,6 +109,26 @@ steps:
   # or take a long time.
   - wait
 
+  - label: "[integration] automate-load-balancer"
+    command:
+      - . scripts/verify_setup
+      - |
+        hab studio run "source scripts/verify_studio_init &&
+          start_deployment_service &&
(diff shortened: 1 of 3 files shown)


srenatus:
Adding a test... ⏳

Issue #540, opened 2019-06-07
flaky test: compliance scanner


vjeffrey:
# User Story

```

1) Failure:
--
  | 30_jobs_docker_spec.rb#test_0005_works [/var/lib/buildkite-agent/builds/single-use-privileged-i-0634b28ea655af127-1/chef-oss/.golang/chef-automate-master-verify/src/github.com/chef/automate/components/compliance-service/api/tests/30_jobs_docker_spec.rb:669]:
  | expected | 0s
  | actual | 1m 30s
  | @@ -1 +1 @@
  | -1
  | +2
```

https://buildkite.com/chef-oss/chef-automate-master-verify/builds/1796#db69b150-b637-418c-837e-4e70f13322c5

we've seen this multiple times. 
i believe it's related to https://github.com/chef/automate/pull/377/files and https://github.com/chef/automate/pull/399 ....sorry :/

alexpop:
Let me have a look at this bad boy!

alexpop:
Looking into this issue, I noticed that for this nodes list query:
```ruby
MANAGER_GRPC nodes, :list, Nodes::Query.new(filters: [
      Common::Filter.new(key: "state", values: ["TERMINATED"]),
      Common::Filter.new(key: "environment", values: ["trou"]),
      Common::Filter.new(key: "group", values: ["mak", "do"])
    ])
```

the node manager is doing the following SQL query to get the list of nodes:
```sql
SELECT
  n.connection_error,
  n.id,
  n.last_contact,
  n.last_job,
  n.manager,
  n.name,
  n.platform,
  n.platform_version,
  n.report_id,
  n.source_state,
  n.status,
  n.target_config,
  n.last_scan,
  n.last_run,
  n.projects_data,
  COALESCE(('[' || string_agg('{"key":"' || t.key || '"' || ',"value": "' || t.value || '"}', ',') || ']'), '[]') :: JSON AS tags,
  COALESCE(array_to_json(array_remove(array_agg(DISTINCT m.manager_id), NULL)), '[]') AS manager_ids,
  COALESCE(array_to_json(array_remove(array_agg(p.project_id), NULL)), '[]') AS projects,
  COUNT(*) OVER () AS total_count
FROM nodes n
  LEFT JOIN nodes_tags nt ON n.id = nt.node_id
  LEFT JOIN tags t ON t.id = nt.tag_id
  LEFT JOIN node_managers_nodes m on n.id = m.node_id
[...]

PR #662, opened 2019-06-23
Fix node manager tags based filtering

alexpop:
Fixes https://github.com/chef/automate/issues/540

### DIFF
--- a/components/compliance-service/api/tests/25_nodes_spec.rb
+++ b/components/compliance-service/api/tests/25_nodes_spec.rb
@@ -630,7 +630,7 @@ describe File.basename(__FILE__) do
     )
     assert_equal(["M$", "test-auto-job-creation"], nodes_list["nodes"])
 
-    # Get the nodes tagged market:market% or market:missing-in-action%:
+    # Get the nodes tagged department:market% or department:missing-in-action%:
     nodes_list = MANAGER_GRPC manager, :search_nodes, Manager::NodeQuery.new(
       node_manager_id: "e69dc612-7e67-43f2-9b19-256afd385820",
       query: Manager::Query.new(
@@ -639,8 +639,9 @@ describe File.basename(__FILE__) do
         ]
       )
     )
+    assert_equal(["betamaniac"], nodes_list["nodes"])
 
-    # Get the nodes NOT tagged market:market% or market:missing-in-action%:
+    # Get the nodes NOT tagged department:market% or department:missing-in-action%:
     nodes_list = MANAGER_GRPC manager, :search_nodes, Manager::NodeQuery.new(
       node_manager_id: "e69dc612-7e67-43f2-9b19-256afd385820",
       query: Manager::Query.new(
--- a/components/compliance-service/api/tests/25_nodes_spec.rb
+++ b/components/compliance-service/api/tests/25_nodes_spec.rb
@@ -291,7 +291,7 @@ describe File.basename(__FILE__) do
     }
     assert_equal_json_sorted(expected_node2.to_json, node2.to_json)
 
-    # # Get node by id with all details
+    # Get node by id with all details
     node3 = MANAGER_GRPC nodes, :read, Nodes::Id.new(id: node_id3)
     node3['last_contact'] = nil
     expected_node3 = {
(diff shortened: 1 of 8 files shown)


components/compliance-service/api/tests/25_nodes_spec.rb
```diff
@@ -597,6 +597,72 @@ def cleanup
...
+
+    # Get the nodes tagged market:market% or market:missing-in-action%:
+    nodes_list = MANAGER_GRPC manager, :search_nodes, Manager::NodeQuery.new(
+      node_manager_id: "e69dc612-7e67-43f2-9b19-256afd385820",
+      query: Manager::Query.new(
+        filter_map: [
+          Common::Filter.new(key: "department", values: ["market", "missing-in-action"]),
+        ]
+      )
+    )
```
phiggins: Is there a missing assert here?

components/nodemanager-service/pgdb/filtering.go
```diff
@@ -102,16 +102,17 @@ func buildWhereFilter(mergeableFilters []*common.Filter, tableAbbrev string, fil
...
-		conditions = append(conditions, tagCondition)
+		havingFilter = fmt.Sprintf("HAVING (%s)", strings.Join(tagConditions, ") AND ("))
 	}
-
-	whereFilter = fmt.Sprintf("WHERE (%s)", strings.Join(conditions, " AND "))
-	logrus.Debugf("buildWhereFilter, whereFilter=%s", whereFilter)
-	return whereFilter, nil
+	if len(conditions) > 0 {
+		whereFilter = fmt.Sprintf("WHERE (%s)", strings.Join(conditions, " AND "))
+	}
```
phiggins: Nitpick: the result of this should be the same with or without the if condition here.

components/nodemanager-service/pgdb/nodes_integration_test.go
```diff
@@ -200,30 +202,49 @@ func (suite *NodesIntegrationSuite) TestGetNodesCanFilterByTags() {
...
 	suite.Require().NoError(err)
-
 	suite.Equal(2, len(fetchedNodes))
 	suite.Equal(&pgdb.TotalCount{Total: 2, Unreachable: 0, Reachable: 0, Unknown: 3}, count)
-	suite.Equal("Nacho Node", fetchedNodes[0].GetName())
-	suite.Equal("Taco Node", fetchedNodes[1].GetName())
+	if len(fetchedNodes) >= 2 {
+		suite.Equal("Nacho Node", fetchedNodes[0].GetName())
+		suite.Equal("No Nacho Node", fetchedNodes[1].GetName())
+	}
```
phiggins: I feel like instead of a conditional it'd be better to make this an explicit part of the test by changing the length test above to use `suite.Require()`.

components/nodemanager-service/pgdb/nodes_integration_test.go
```diff
@@ -200,30 +202,49 @@ func (suite *NodesIntegrationSuite) TestGetNodesCanFilterByTags() {
...
+		Key:    "tacos",
+		Values: []string{"yes"},
+	}
+	fetchedNodes, count, err = suite.Database.GetNodes("name", nodes.Query_ASC, 1, 100, []*common.Filter{filter1, filter2})
+	suite.Require().NoError(err)
+	suite.Equal(1, len(fetchedNodes))
+	suite.Equal(&pgdb.TotalCount{Total: 1, Unreachable: 0, Reachable: 0, Unknown: 3}, count)
+	if len(fetchedNodes) >= 1 {
+		suite.Equal("No Nacho Node", fetchedNodes[0].GetName())
+	}
```
phiggins: Same comment here, if the length test above fails we probably shouldn't even be running anymore tests.

components/compliance-service/api/tests/25_nodes_spec.rb
alexpop: Yes, good catch. Pushed another commit with the assert and also updated the comments for the two department tag filters.

components/nodemanager-service/pgdb/filtering.go
alexpop: Without the if condition, whereFilter becomes `WHERE ()` when no conditions exist. This leads to bad SQL:
```
nodemanager_service=# SELECT * FROM tags WHERE();
ERROR:  syntax error at or near ")"
LINE 1: SELECT * FROM tags WHERE();
```

components/nodemanager-service/pgdb/nodes_integration_test.go
alexpop: Yeah, great use of `suite.Require()`. Pushed a commit and removed the condition.

components/nodemanager-service/pgdb/nodes_integration_test.go
alexpop: Done, thank you

components/nodemanager-service/pgdb/filtering.go
```diff
@@ -38,31 +38,31 @@ func mergeFilters(mergeableFilters []*common.Filter) ([]common.Filter, error) {
 	return filters, nil
 }
 
-func handleTagFilters(tagFilters []common.Filter) (string, error) {
-	var tagConditions []string
+func handleTagFilters(tagFilters []common.Filter) ([]string, error) {
+	tagConditions := make([]string, 0)
```
stevendanna: Minor, we could pre-allocate this array here:

```suggestion
	tagConditions := make([]string, 0, len(tagFilters))
```

components/nodemanager-service/pgdb/filtering.go
```diff
@@ -38,31 +38,31 @@ func mergeFilters(mergeableFilters []*common.Filter) ([]common.Filter, error) {
 	return filters, nil
 }
 
-func handleTagFilters(tagFilters []common.Filter) (string, error) {
-	var tagConditions []string
+func handleTagFilters(tagFilters []common.Filter) ([]string, error) {
```
stevendanna: I'm wondering if this is a clearer function name we might use here.  `TagFiltersToSQLConditions` maybe?

components/nodemanager-service/pgdb/node-managers.go
```diff
@@ -303,7 +303,7 @@ func (db *DB) GetNodeManagers(sortField string, insortOrder manager.Query_OrderT
 	if err != nil {
 		return nil, 0, errors.Wrapf(err, "GetNodeManagers error validating node manager filters")
 	}
-	whereFilter, err := buildWhereFilter(filters, "nm", mgrFilterField)
+	whereFilter, _, err := buildWhereHavingFilter(filters, "nm", mgrFilterField)
```
stevendanna: It seems the second argument here is only ever "nm" or "n", might it be worth it to use some well-named package-level constants:

```golang
const (
    NodeTableAbbrev        = "n"
    NodeManagerTableAbbrev = "nm"
)
```

that is one of a number of small things we could probably do to make the code a bit easier to understand.

components/nodemanager-service/pgdb/filtering.go
phiggins: Ahh, right. Good call.

components/nodemanager-service/pgdb/filtering.go
alexpop: Thank you Steven, I pushed a commit with your suggestions.

components/nodemanager-service/pgdb/node-managers.go
alexpop: Yes, I created these constants for the table abbreviations:
```go
const (
  NodesTableAbbrev        = "n"
  TagsTableAbbrev         = "t"
  NodeManagersTableAbbrev = "nm"
)
```

PR #664, opened 2019-06-24
Ingestion improvements to applications-service

afiune:
In preparation to start work to add the `timewizard` from [JIRA#A2-665](https://chefio.atlassian.net/browse/A2-665),
I have started with a number of improvements to structure a little better
the ingestion mechanism inside the applications-service, doing so I have
discovered a number of issues that we need to cover before doing any
new modification in the database and ingestion.

Problems:

- Race condition: when different services that belong to the same service
group are sending messages to Automate at the same time we will insert
different service-group and deployment entries instead of a single one.
The main reason for this problem is the fact that we process 50 things at a
time through the worker pools and they don't coordinate between them.
- Define what are the fields that we can update.
- Define criteria for grouping services.
- Fix ingestion to accept **origin** updates.
- Fix ingestion to remove the **channel** when there is no update strategy:
new channel updates are registered but if a service changes to have
NO update strategy, we won't update that.
- Fix ingestion to update the new **package_ident** field in the database

Improvements:
- Create a new `ingest` go package to encapsulate all the ingestion logic.
- Moved some ingestion logic from the `nats` & `server` packages to the new 
`ingest` package.
- Added a `workerOutputQueue` to detect errors and count events processed,
failed and successful.
- Added a number of `Ingester` tests.

### :+1: Definition of Done

All above problems are fixed.

Run the integration tests for the applications-service:
```
$ hab studio enter
[2][default:/src:0]# applications_integration
```
https://chefio.atlassian.net/browse/A2-889
https://chefio.atlassian.net/browse/A2-665

- [x] Code actually executed?

### DIFF
--- a/components/applications-service/integration_test/ingestion_test.go
+++ b/components/applications-service/integration_test/ingestion_test.go
@@ -226,6 +226,28 @@ func TestIngestSigleServiceInsertAndUpdate(t *testing.T) {
 			"the service channel name is not the expected one")
 		assert.Equal(t, "us", svcList[0].Site,
 			"the service site name is not the expected one")
+
+		sgList := suite.GetServiceGroups()
+		assert.Equal(t, 1, len(sgList), "wrong number of service groups")
+
+		assert.Equal(t, "db.default", sgList[0].Name,
+			"the service_group name is not the expected one")
+		assert.Equal(t, "test/db", sgList[0].Package,
+			"the service_group package is not the expected one")
+		assert.Equal(t, "0.1.0/20200101121212", sgList[0].Release,
+			"the service_group release is not the expected one")
+		assert.Equal(t, "OK", sgList[0].HealthStatus,
+			"the service_group health status is not the expected one")
+		assert.Equal(t, int32(100), sgList[0].HealthPercentage,
+			"the service_group health percentage is not the expected one")
+		assert.Equal(t, "test-app", sgList[0].Application,
+			"the service_group application name is not the expected one")
+		assert.Equal(t, "development", sgList[0].Environment,
+			"the service_group environment name is not the expected one")
+		assert.Equal(t, int32(1), sgList[0].ServicesHealthCounts.Total,
+			"the total number of services in this service_group is not the expected one")
+		assert.Equal(t, int32(1), sgList[0].ServicesHealthCounts.Ok,
+			"the OK number of services in this service_group is not the expected one")
 	}
 
 	// 2) Update same service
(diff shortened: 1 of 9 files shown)


components/applications-service/pkg/ingest/ingest.go
```diff
@@ -77,9 +95,7 @@ func (i *Ingester) runWorkerLoop() {
 		case event := <-i.workerTaskQueue:
 			err := i.DBClient.IngestHealthCheckEvent(event)
 			if err != nil {
-				log.WithFields(log.Fields{
-					"error": err.Error(),
-				}).Error("Unable to ingest habitat event")
+				log.WithError(err).Error("Unable to ingest habitat event")
```
danielsdeleo: 🥇

components/applications-service/integration_test/suite_test.go
```diff
@@ -66,6 +71,17 @@ func NewSuite(database string) *Suite {
...
 
+	// A global Ingester instance to ingest any "message" into our system
+	//
+	// From any test you can directly call:
+	// ```
+	// res, err := suite.Ingester.IgestMessage(msg)
+	// ```
+	s.Ingester = ingest.New(c, s.StorageClient)
+
+	// Start processing messages as they are sended to the ingest events channel
```
danielsdeleo: `s/sended/sent/` If the spellchecker doesn't complain now, it could in the future.

components/applications-service/pkg/nats/subscriber.go
```diff
@@ -17,9 +17,11 @@ func (nc *NatsClient) Subscribe(eventsCh chan<- []byte) (stan.Subscription, erro
...
 
 		log.WithFields(log.Fields{
-			"protocol": "nats",
-			"data":     string(msg.Data),
-			"subject":  nc.subject,
+			"protocol":          "nats",
+			"message_data":      string(msg.Data),
+			"message_timestamp": msg.Timestamp,
+			"message_subject":   msg.Subject,
+			"message_sequence":  msg.Sequence,
```
danielsdeleo: 🥇

components/applications-service/integration_test/ingestion_test.go
```diff
@@ -5,4 +5,277 @@
...
+	svcList := suite.GetServices()
+	assert.Equal(t, len(events), len(svcList), "wrong number of services")
+
+	// We have a race condition where if we have different services sending messages
+	// to Automate at the same time that belongs to the same service-group we will
+	// insert different service-groups and deployments entries instead of a single one.
+	//
+	// The main reason of this problem is because we process 50 things at a time (worker pool)
+	//
+	// TODO @afiune fix and uncomment this test!
```
danielsdeleo: This is mentioned in the PR description... did you fix it or just find it?

afiune: I found it, I didn't know if the PR would get too big for review so I first added tests to discover the failures. Though I have already fixed a number of them and pushed my fixes, let me know what you think? 

I didn't fix the race condition, I am going to write a bug card to fix it in a different PR. 💯

components/applications-service/integration_test/suite_test.go
afiune: Changed! Thanks

components/applications-service/integration_test/ingestion_test.go
afiune: Jira bug card: https://chefio.atlassian.net/browse/A2-890

PR #670, opened 2019-06-24
[A2-814] ListRulesForProject returns latest rule updates

bcmdarroch:

Any user viewing a project's rules should be able to see the latest staged changes to a project. This PR refactors ListRulesForProject to check both the `applied` and `staged` rule tables and ensure that only the most up-to-date version of the rule is returned in the list.

This also means any applied rule that has been staged for deletion will not be returned in this list.

db setup
```
# first project with one rule is updated

insert into iam_projects (id, name, type, projects) values ('foo-project', 'Foo Project', 'custom', ARRAY['foo-project']);

insert into iam_project_rules (id, project_id, name, type) values ('foo-rule', 'foo-project', 'foorule', 'node');

insert into iam_rule_conditions (rule_db_id, attribute, value, operator) values ((SELECT db_id from iam_project_rules where id='foo-rule'), 'chef-server', ARRAY['foo'], 'equals');

insert into iam_staged_project_rules (id, project_id, name, type, deleted) values ('foo-rule', 'foo-project', 'updated foorule', 'node', false);

insert into iam_staged_rule_conditions (rule_db_id, attribute, value, operator) values ((SELECT db_id from iam_staged_project_rules where id='foo-rule'), 'chef-server', ARRAY['new foo'], 'equals');


# second project with one rule that has not been changed

insert into iam_projects (id, name, type, projects) values ('bar-project', 'Bar Project', 'custom', ARRAY['bar-project']);

insert into iam_project_rules (id, project_id, name, type) values ('bar-rule', 'bar-project', 'barrule', 'node');

insert into iam_rule_conditions (rule_db_id, attribute, value, operator) values ((SELECT db_id from iam_project_rules where id='bar-rule'), 'chef-server', ARRAY['bar'], 'equals');
```

curl
```
export TOK=`chef-automate iam token create TEST --admin`
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects/foo-project/rules | jq
{
  "rules": [
    {
      "id": "foo-rule",
      "project_id": "foo-project",
      "name": "updated foorule",
      "type": "NODE",
      "conditions": [
        {
          "attribute": "CHEF_SERVERS",
          "values": [
            "new foo"
          ],
          "operator": "EQUALS"
        }
      ]
    }
  ]
}

curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects/bar-project/rules | jq
{
  "rules": [
    {
      "id": "bar-rule",
      "project_id": "bar-project",
      "name": "barrule",
      "type": "NODE",
      "conditions": [
        {
          "attribute": "CHEF_SERVERS",
          "values": [
            "bar"
          ],
          "operator": "EQUALS"
        }
      ]
    }
  ]
}
```
note the `applied` version of the rule is not returned.

- [ ] Code actually executed?

### DIFF
--- a/components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
@@ -1,0 +1,54 @@
+BEGIN;
+
+CREATE OR REPLACE FUNCTION
+  query_rules_for_project(_project_id TEXT, _project_filter TEXT[])
+  RETURNS setof json AS $$
+
+  -- fetch staged rules
+   SELECT json_build_object(
+    'id', r.id,
+    'project_id', r.project_id,
+    'name', r.name,
+    'type', r.type,
+    'status', 'staged',
+    'conditions', json_agg(
+      json_build_object(
+        'value', rc.value,
+        'operator', rc.operator,
+        'attribute', rc.attribute
+      )
+    )
+  ) AS rule
+  FROM iam_staged_project_rules AS r
+  INNER JOIN iam_staged_rule_conditions AS rc ON rc.rule_db_id=r.db_id
+  WHERE projects_match_for_rule(project_id, _project_filter) AND r.deleted=false
+  GROUP BY r.id, r.project_id, r.name, r.type
+
+  UNION ALL
+
+  -- fetch applied rules
+  SELECT json_build_object(
+    'id', r.id,
+    'project_id', r.project_id,
+    'name', r.name,
+    'type', r.type,
+    'status', 'applied',
+    'conditions', json_agg(
+      json_build_object(
+        'value', rc.value,
+        'operator', rc.operator,
+        'attribute', rc.attribute
+      )
+    )
+  ) AS rule
+  FROM iam_project_rules AS r
+  INNER JOIN iam_rule_conditions AS rc
+  ON rc.rule_db_id=r.db_id
+  WHERE _project_id=project_id AND projects_match_for_rule(project_id, _project_filter)
+    -- return applied rule only if no staged changes exist
+    AND NOT EXISTS (SELECT 1 FROM iam_staged_project_rules WHERE id=r.id)
+  GROUP BY r.id, r.project_id, r.name, r.type;
+
+ $$ LANGUAGE sql;
+
+ COMMIT;
--- a/components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
@@ -22,7 +22,7 @@ CREATE OR REPLACE FUNCTION
   FROM iam_staged_project_rules AS r
   INNER JOIN iam_staged_rule_conditions AS rc 
   ON rc.rule_db_id=r.db_id
-  WHERE _project_id=r.project_id AND projects_match_for_rule(project_id, _project_filter) AND r.deleted=false
+  WHERE _project_id=r.project_id AND projects_match_for_rule(r.project_id, _project_filter) AND r.deleted=false
   GROUP BY r.id, r.project_id, r.name, r.type
 
   UNION ALL
@@ -45,7 +45,7 @@ CREATE OR REPLACE FUNCTION
   FROM iam_project_rules AS r
   INNER JOIN iam_rule_conditions AS rc
   ON rc.rule_db_id=r.db_id
-  WHERE _project_id=r.project_id AND projects_match_for_rule(project_id, _project_filter)
+  WHERE _project_id=r.project_id AND projects_match_for_rule(r.project_id, _project_filter)
     -- return applied rule only if no staged changes exist
     AND NOT EXISTS (SELECT 1 FROM iam_staged_project_rules WHERE id=r.id)
   GROUP BY r.id, r.project_id, r.name, r.type;
--- a/components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
@@ -20,8 +20,9 @@ CREATE OR REPLACE FUNCTION
     )
   ) AS rule
   FROM iam_staged_project_rules AS r
-  INNER JOIN iam_staged_rule_conditions AS rc ON rc.rule_db_id=r.db_id
-  WHERE _project_id=project_id AND projects_match_for_rule(project_id, _project_filter) AND r.deleted=false
+  INNER JOIN iam_staged_rule_conditions AS rc 
+  ON rc.rule_db_id=r.db_id
+  WHERE _project_id=r.project_id AND projects_match_for_rule(project_id, _project_filter) AND r.deleted=false
   GROUP BY r.id, r.project_id, r.name, r.type
 
   UNION ALL
@@ -44,7 +45,7 @@ CREATE OR REPLACE FUNCTION
   FROM iam_project_rules AS r
   INNER JOIN iam_rule_conditions AS rc
   ON rc.rule_db_id=r.db_id
-  WHERE _project_id=project_id AND projects_match_for_rule(project_id, _project_filter)
+  WHERE _project_id=r.project_id AND projects_match_for_rule(project_id, _project_filter)
     -- return applied rule only if no staged changes exist
     AND NOT EXISTS (SELECT 1 FROM iam_staged_project_rules WHERE id=r.id)
   GROUP BY r.id, r.project_id, r.name, r.type;
--- a/components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
@@ -21,7 +21,7 @@ CREATE OR REPLACE FUNCTION
   ) AS rule
   FROM iam_staged_project_rules AS r
   INNER JOIN iam_staged_rule_conditions AS rc ON rc.rule_db_id=r.db_id
-  WHERE projects_match_for_rule(project_id, _project_filter) AND r.deleted=false
+  WHERE _project_id=project_id AND projects_match_for_rule(project_id, _project_filter) AND r.deleted=false
   GROUP BY r.id, r.project_id, r.name, r.type
 
   UNION ALL
--- a/components/authz-service/storage/v2/postgres/postgres_test.go
+++ b/components/authz-service/storage/v2/postgres/postgres_test.go
@@ -3861,6 +3861,33 @@ func TestListRulesForProject(t *testing.T) {
 			assert.NoError(t, err)
 			assert.ElementsMatch(t, []*storage.Rule{}, resp)
 		}},
+		{"when multiple projects exist, only the requested project's rules are returned", func(t *testing.T) {
+			ctx := context.Background()
+			projID1 := "foo-project"
+			insertTestProject(t, db, projID1, "first project", storage.Custom)
+
+			condition1, err := storage.NewCondition(storage.Node,
+				[]string{"chef-server-1"}, storage.ChefServer, storage.MemberOf)
+			require.NoError(t, err)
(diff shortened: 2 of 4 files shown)


components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
```diff
@@ -0,0 +1,54 @@
+BEGIN;
+
+CREATE OR REPLACE FUNCTION
+  query_rules_for_project(_project_id TEXT, _project_filter TEXT[])
```
bcmdarroch: not sure if there's a way we can DRY up some of these query functions? there's a lot of overlap with [these queries](https://github.com/chef/automate/blob/master/components/authz-service/storage/postgres/migration/sql/50_add_status_field_to_rules.up.sql)

inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -1074,12 +1074,11 @@
         expect(resp.http_status.to_s).to match(/200|404/)
       end
 
-      # TODO uncomment once ListRulesForProject has been refactored with staged rules
-      # it "GET /iam/v2beta/projects/:id/rules returns the rules for the project" do
-      #   resp = automate_api_request("/apis/iam/v2beta/projects/#{CUSTOM_PROJECT_ID}/rules")
-      #   expect(resp.http_status).to eq 200
-      #   expect(resp.parsed_response_body[:rules]).to match_array([CUSTOM_RULE_1, CUSTOM_RULE_2])
-      # end
+      it "GET /iam/v2beta/projects/:id/rules returns the most up-to-date rules for the project" do
```
bcmdarroch: This should pass but we'll want to add further test coverage for applied vs. staged rules once `ApplyRules` is in-place

components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
srenatus: At this point, it would be quite wortwhile to figure this out 😉

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3700,13 +3700,15 @@ func TestListRulesForProject(t *testing.T) {
 
 			condition4, err := storage.NewCondition(ruleType,
 				[]string{"chef-server-2"}, storage.ChefServer, storage.MemberOf)
+			require.NoError(t, err)
```
srenatus: 👏

components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
```diff
@@ -0,0 +1,54 @@
...
+      json_build_object(
+        'value', rc.value,
+        'operator', rc.operator,
+        'attribute', rc.attribute
+      )
+    )
+  ) AS rule
+  FROM iam_staged_project_rules AS r
+  INNER JOIN iam_staged_rule_conditions AS rc ON rc.rule_db_id=r.db_id
+  WHERE projects_match_for_rule(project_id, _project_filter) AND r.deleted=false
```
msorens: This should not be working (unless you only have rules for one project in the DB).
So... suggest you add a test first that has two projects to see if it fails.
Then make the test pass. 😉 

```suggestion
  WHERE _project_id=r.project_id AND projects_match_for_rule(r.project_id, _project_filter) AND r.deleted=false
```

components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
```diff
@@ -0,0 +1,54 @@
...
+        'value', rc.value,
+        'operator', rc.operator,
+        'attribute', rc.attribute
+      )
+    )
+  ) AS rule
+  FROM iam_project_rules AS r
+  INNER JOIN iam_rule_conditions AS rc
+  ON rc.rule_db_id=r.db_id
+  WHERE _project_id=project_id AND projects_match_for_rule(project_id, _project_filter)
```
msorens: Let's be consistent in usage:
```suggestion
  WHERE _project_id=r.project_id AND projects_match_for_rule(r.project_id, _project_filter)
```

components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
```diff
@@ -0,0 +1,54 @@
...
+    'conditions', json_agg(
+      json_build_object(
+        'value', rc.value,
+        'operator', rc.operator,
+        'attribute', rc.attribute
+      )
+    )
+  ) AS rule
+  FROM iam_staged_project_rules AS r
+  INNER JOIN iam_staged_rule_conditions AS rc ON rc.rule_db_id=r.db_id
```
msorens: Line break before `ON`

components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
```diff
@@ -0,0 +1,54 @@
...
+  query_rules_for_project(_project_id TEXT, _project_filter TEXT[])
+  RETURNS setof json AS $$
+
+  -- fetch staged rules
+   SELECT json_build_object(
+    'id', r.id,
+    'project_id', r.project_id,
+    'name', r.name,
+    'type', r.type,
+    'status', 'staged',
```
blakestier: do we need the `deleted` field here?

blakestier: or no bc these are only deleted . . . hmm ok

components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
blakestier: I think we could dry them up. The primary difference seems to revolve around the `deleted` field and I think we could include that and drop it at the next level, potentially. We also might be able to wrap some of the inner logic into separate functions, so updates to the outer function aren't as extensive-looking.

components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
bcmdarroch: it'd be more explicit, but then i'd also need to return any empty column from the applied table below. the API's going to ignore it so i think we can ignore it here

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3791,6 +3791,103 @@ func TestListRulesForProject(t *testing.T) {
 			assert.NoError(t, err)
 			assert.Zero(t, len(resp))
 		}},
+		{"when there are staged changes for the project's rules, returns the latest changes to the rules", func(t *testing.T) {
```
blakestier: I think the phrasing around "latest" is maybe too ambiguous. What about 
"when there are only staged changes for the project's rules, returns the staged changes to the rules" 
and
"when there are staged changes for some of the project's rules, returns those staged versions and rules that have no staged changes"

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3791,6 +3791,103 @@ func TestListRulesForProject(t *testing.T) {
...
+
+			condition, err := storage.NewCondition(storage.Node,
+				[]string{"chef-server-1"}, storage.ChefServer, storage.MemberOf)
+			require.NoError(t, err)
+			rule, err := storage.NewRule("foo-rule", projID, "coo foo rule", condition.Type, []storage.Condition{condition})
+			assert.NoError(t, err)
+			insertAppliedRule(t, db, &rule)
+
+			// mark for deletion
+			insertStagedRule(t, db, &rule, true)
```
blakestier: hmm maybe we should add another helper `insertDeletedStagedRule` instead of needing the comment 🤔

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3791,6 +3791,103 @@ func TestListRulesForProject(t *testing.T) {
...
+			require.NoError(t, err)
+			insertStagedRule(t, db, &updatedRule, false)
+
+			appliedRule := insertAppliedRuleWithMultipleConditions(t, db, projID, condition.Type)
+
+			resp, err := store.ListRulesForProject(ctx, projID)
+			assert.NoError(t, err)
+			assert.ElementsMatch(t, []*storage.Rule{&updatedRule, appliedRule}, resp)
+		}},
+		{"when a project's rule has been marked for deletion, it will not be returned", func(t *testing.T) {
```
blakestier: "when a project's only rule has been marked for deletion, returns empty array"

blakestier: also it looks like we could use a ""when a project has two applied rules and one is staged for deletion, returns only the non-deleted one"

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -3791,6 +3791,103 @@ func TestListRulesForProject(t *testing.T) {
...
+			insertAppliedRule(t, db, &rule)
+
+			// mark for deletion
+			insertStagedRule(t, db, &rule, true)
+
+			resp, err := store.ListRulesForProject(ctx, projID)
+			assert.NoError(t, err)
+			assert.ElementsMatch(t, []*storage.Rule{}, resp)
+		}},
+		{"when multiple projects exist, only the requested project's rules are returned", func(t *testing.T) {
```
blakestier: I'm obsessing over the inconsistency of `__ are returned` vs `returns __` 😂 :i'msorry:

inspec/a2-api-integration/controls/iam_v2.rb
```diff
@@ -1094,6 +1088,33 @@
         expect(resp.parsed_response_body[:rule]).to eq(CUSTOM_RULE_1)
       end
 
+      it "GET /iam/v2beta/projects/:id/rules returns the most up-to-date rules for the project" do
```
blakestier: here, too, "up-to-date" is probably only meaningful to us now bc we're in the thick of it. I'd be more specific about staged vs applied (or something)

components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
bcmdarroch: Update: @blakestier and I timeboxed writing a smarter query, but ended up deciding the complexity and time it would take to write one were too much. 
[This stack overflow answer](https://stackoverflow.com/questions/11740256/refactor-a-pl-pgsql-function-to-return-the-output-of-various-select-queries) seemed like a promising lead if we ever come back to this.

components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
```diff
@@ -0,0 +1,55 @@
...
+        'value', rc.value,
+        'operator', rc.operator,
+        'attribute', rc.attribute
+      )
+    )
+  ) AS rule
+  FROM iam_staged_project_rules AS r
+  INNER JOIN iam_staged_rule_conditions AS rc 
+  ON rc.rule_db_id=r.db_id
+  WHERE _project_id=r.project_id AND projects_match_for_rule(project_id, _project_filter) AND r.deleted=false
```
msorens:
```suggestion
  WHERE _project_id=r.project_id AND projects_match_for_rule(r.project_id, _project_filter) AND r.deleted=false
```

components/authz-service/storage/postgres/migration/sql/51_change_project_rules_query.up.sql
```diff
@@ -0,0 +1,55 @@
...
+        'value', rc.value,
+        'operator', rc.operator,
+        'attribute', rc.attribute
+      )
+    )
+  ) AS rule
+  FROM iam_project_rules AS r
+  INNER JOIN iam_rule_conditions AS rc
+  ON rc.rule_db_id=r.db_id
+  WHERE _project_id=r.project_id AND projects_match_for_rule(project_id, _project_filter)
```
msorens:
```suggestion
  WHERE _project_id=r.project_id AND projects_match_for_rule(r.project_id, _project_filter)
```

PR #673, opened 2019-06-25
Run pg integration tests in automate lib

stevendanna:
These tests are not currently running on master:

    SKIP: TestDBInterface (0.00s)
        exporter_integration_test.go:61:
        Skipping exporter integration tests because pg_ctl was not
        found on path! To run these tests, you must have PostgreSQL
        installed

This has likely been the case since this library was extracted
from the deployment-service.


stevendanna:
Horray, they still pass!

Screen Shot 2019-06-25 at 1 23 14 PM

PR #678, opened 2019-06-25
Add timestamp fields to database tables

afiune:
We are adding timestamps to all our tables inside the applications-service.

### :+1: Definition of Done
Every table in the database has `created_at` and `updated_at` timestamps.

Additionally, we will start storing the timestamp of when habitat submitted
the event message. (new column in service table `last_event_occurred_at`)

Build and start the applications-service and run the integration tests,
you can also login to the database and list the data.

https://chefio.atlassian.net/browse/A2-892

- [x] Code actually executed?

### DIFF
--- a/components/applications-service/integration_test/storage_service_timestamp_test.go
+++ b/components/applications-service/integration_test/storage_service_timestamp_test.go
@@ -105,8 +105,8 @@ func TestStorageGetServiceFromUniqueFieldsExistUpdateLastEventOccurredAt(t *test
 		initialOccurredAtTimestamp = svc.LastEventOccurredAt
 	}
 
-	// Trigger an update but wait for a second before creating the new event
-	time.Sleep(1 * time.Second)
+	// Trigger an update but wait for half a second before creating the new event
+	time.Sleep(500 * time.Millisecond)
 	suite.IngestService(
 		NewHabitatEvent(
 			withSupervisorId("1q2w3e4r"),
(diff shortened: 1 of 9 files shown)


components/applications-service/pkg/storage/postgres/service_groups.go
```diff
@@ -295,13 +295,13 @@ func (db *Postgres) GetServiceGroupsHealthCounts() (*storage.HealthCounts, error
 
 // ServiceGroupExists returns the name of the service group if it exists
 func (db *Postgres) ServiceGroupExists(id string) (string, bool) {
-	var sg serviceGroup
-	err := db.SelectOne(&sg, "SELECT * FROM service_group WHERE id = $1", id)
+	var sgName string
+	err := db.SelectOne(&sgName, "SELECT name FROM service_group WHERE id = $1", id)
```
danielsdeleo: 💯

kmacgugan:
Can we add a test that ensures created_at timestamp doesn't get updated on update? Or just manually test that to make sure.

afiune:
@kmacgugan already tested it since I added the function, triggers, and procedures. I would love if you could also test it to give your +1 

I could also add a test but I am trying to slice my PRs so that they don't get too big.

afiune:
@kmacgugan BTW Welcome back!!!

tenor-221985525

components/applications-service/integration_test/storage_service_timestamp_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestStorageGetServiceFromUniqueFieldsNotExist(t *testing.T) {
+	svc, exist := suite.StorageClient.GetServiceFromUniqueFields("foo", "123")
+	assert.False(t, exist)
+	assert.Nil(t, svc)
+}
+
+func TestStorageGetServiceFromUniqueFieldsEmptyParameters(t *testing.T) {
```
danielsdeleo: This test doesn't seem to be testing much since the storage is empty. Maybe you could insert the hab event from the next test and then check that the following all return false:

```
suite.StorageClient.GetServiceFromUniqueFields("", "")
suite.StorageClient.GetServiceFromUniqueFields("postgres", "")
suite.StorageClient.GetServiceFromUniqueFields("", "1q2w3e4r")
```

components/applications-service/integration_test/storage_service_timestamp_test.go
```diff
@@ -0,0 +1,113 @@
...
+			withPackageIdent("core/postgres/0.1.0/20190101121212"),
+			withHealth("CRITICAL"),
+		),
+	)
+
+	svc, exist = suite.StorageClient.GetServiceFromUniqueFields("postgres", "1q2w3e4r")
+	if assert.True(t, exist) {
+		updatedOccurredAtTimestamp = svc.LastEventOccurredAt
+
+		// thea substraction from the initial and the updated timestamp should be greater than a second
```
danielsdeleo: typo

components/applications-service/integration_test/storage_service_timestamp_test.go
afiune: Will do

afiune: Though, now that I am doing this modification, it doesn't matter if we have something in the database or not since we have verifications that check if any of the parameters is empty, we just return immediately without running a query. Should I still modify it?

PR #682, opened 2019-06-26
Dan/apps stats api

danielsdeleo:

Creates a stats API for applications. It reports three items which we are interested in including in product telemetry:
* total number of supervisors
* total number of services
* total number of service groups

Including these in product telemetry is intended to help us understand the following things:
* Are customers moving from limited POC/demo deployments to production usage?
* What size and shape of load should we apply during performance and load testing?
* Should we consider alternative UI designs and interaction patterns to help users deal with larger scale data?

### :+1: Definition of Done

Counts of supervisors, services and service groups are available via the front-end API. Integration with browser-based telemetry will come later.

(in the studio) Build applications-service and automate-gateway. Start all services.

```
export H="api-token: $(get_admin_token)"
curl -k -H "$H" https://a2-dev.test/apis/beta/applications/stats
```

Counts should match what you can see in the UI. You can use `applications_populate_database` to create some things to count.

- [ ] Code actually executed?


api/external/applications/applications.proto
```diff
@@ -47,14 +43,21 @@ service ApplicationsService {
   };
 
   rpc GetServicesBySG(ServicesBySGReq) returns (ServicesBySGRes) {
-    // TODO (afiune, 4/2019): need to replace this once we have resources and such
```
afiune: Thanks! 🏅

components/applications-service/integration_test/ingestion_test.go
```diff
@@ -110,9 +113,15 @@ func TestIngestMultiServicesDifferentServiceGroups(t *testing.T) {
...
 
+	// make sure the stats endpoint stuff is also correct
+	assert.Equal(t, int32(len(events)), suite.GetServicesCountForStatsEndpoint(), "wrong number of services")
+
 	// Verify there are only three service_groups
 	sgList := suite.GetServiceGroups()
 	assert.Equal(t, 3, len(sgList), "wrong number of service groups")
+
+	// make sure the stats endpoint stuff is also correct
+	assert.Equal(t, int32(3), suite.GetServiceGroupsCountForStatsEndpoint(), "wrong number of service groups")
```
afiune: nice!

components/applications-service/pkg/storage/postgres/supervisors.go
```diff
@@ -0,0 +1,13 @@
...
+
+const (
+	selectSupervisorsTotalCount = `
+SELECT count(*)
+  FROM supervisor;
+`
+)
+
+func (db *Postgres) GetSupervisorsCount() (int32, error) {
+	count, err := db.DbMap.SelectInt(selectSupervisorsTotalCount)
```
afiune: These queries are very small, I would recommend only creating constants when the query needs to be written down with format so that it is easier to understand it.

```suggestion
	count, err := db.DbMap.SelectInt("SELECT count(*) FROM supervisor")
```

Same for the other two above. (`service` and `service_group`)

PR #684, opened 2019-06-26
Allow unpacking bootstrap bundle

jaym:
Adds a command to unpack the bundle, along with a switch to the
deploy command:

```
chef-automate deploy --bootstrap-bundle /path/to/bundle
chef-automate bootstrap bundle unpack /path/to/bundle
```

For now, unpacking is done by the CLI. We can change this so that
deployment-service does it in the future once we nail down how we
want this feature to work


components/automate-deployment/pkg/bootstrap/bootstrap.go
```diff
@@ -62,6 +64,20 @@ func FullBootstrap(ctx context.Context,
 		return err
 	}
 
+	if bootstrapBundlePath != "" {
```
stevendanna: My only concern with doing this here, is that the point of some of this indirection was to protect us against version/code mis-matches during deployment.  That is, the rest of this code is shelling out to the deployment-service command to handle the bootstrapping.  The thought was that we might want to be able to deploy a version of the deployment-service that doesn't line up with the version of automate-cli exactly.

I'm not sure if we really care about that anymore though.

jaym: we can probably move it into `DeployDeploymentService`. I'm not sure I see where we're calling the deployment service binary. These functions seem to just use local target

stevendanna: https://github.com/chef/automate/blob/8349e23d64d92d53cd1f53f766d81edd29377955/components/automate-deployment/pkg/bootstrap/bootstrap.go#L113-L116

stevendanna: Once we have hab installed, we turn everything else over to the dsCmd if possible.  As I said above, I'm not sure we want to do this still, but just wanted to point it out in case we do.

Also, I think we should probably make a specific Bootstrapper function for this and call that from FullBootstrap rather than doing it inline, just to remain consistent.

jaym: i see

components/automate-deployment/pkg/bootstrap/deployment_service_cmd.go
```diff
@@ -47,7 +47,11 @@ func (ds *DeploymentServiceCommand) CanBootstrap() bool {
 	return ds.HasCapability("bootstrap")
 }
 
-func (ds *DeploymentServiceCommand) DeployService(ctx context.Context, config *dc.ConfigRequest, m manifest.ReleaseManifest) error {
+func (ds *DeploymentServiceCommand) CanUnpackBootstrapBundle() bool {
+	return ds.HasCapability("bootstrap")
```
stevendanna: I think we should give this a new capability name, maybe "bootstrap-with-bundle"

jaym: oops

Issue #600, opened 2019-06-14
manual bulk node add: modifying tags on one node can delete tags from others


vjeffrey:
# User Story
To reproduce, first get some compliance nodes into Automate node manager with some matching tags like so:
```
curl -s --insecure -H "api-token: $token_val" https://chef-automate.test/api/v0/nodes/bulk-create -d '
{"nodes": [{
     "name_prefix": "my-ssh-node",
     "manager":"automate",
     "target_config": {
        "backend":"ssh",
        "hosts":["localhost","127.0.0.1"],
        "port": 22
      },
      "tags": [
        { "key":"bleep", "value":"bloop" },
        { "key":"bada", "value":"bing" }
      ]
    },
    {
     "name": "my-other-node",
     "manager":"automate",
      "target_config": {
        "backend":"ssh",
        "hosts":["localhost"],
        "port": 22
      },
      "tags": [
        { "key":"bleep", "value":"bloop" }
      ]
    }
  ]
}'
```

> And then go into Automate node manager, and change the e.g. value for the "bleep" tag on any of the 3 nodes that were added to something else, like "AAA", and the tags on the other nodes disappear afterwards, either all of them, or only the ones that were matching to the tag that was modified – we have observed both.

# Definition of Done
Modifying tags on one node has no effect on the tags of another node

# Related Resources
https://chefio.atlassian.net/browse/TRI-857?oldIssueView=true

alexpop:
Looking into this...

PR #695, opened 2019-06-27
Compliance API, fix node manager tag update

alexpop:
Fixes https://github.com/chef/automate/issues/600

I also sorted node tags by key as I was getting random integration test failures locally.

### DIFF
--- a/components/nodemanager-service/Makefile
+++ b/components/nodemanager-service/Makefile
@@ -186,6 +186,11 @@ run-db-tests:
 	@POSTGRES_URI="$(POSTGRES_URI)" \
 		go test -v $(shell go list ./pgdb/...) -cover -database
 
+run-one-db-test:
+	@POSTGRES_URI="$(POSTGRES_URI)" \
+		go test -v $(shell go list ./pgdb/...) -run TestRunNodesIntegrationSuite/TestBulkAddNodesAndTagsUpdate -cover -database
+
+
 test-compliance-ingestion-to-manager-conn: run-background
 	cd ../compliance-service && make run
 	go run ../compliance-service/examples/ingest/ingest_client.go --file ../compliance-service/ingest/examples/compliance-success-tiny-report.json
--- a/components/nodemanager-service/pgdb/nodes.go
+++ b/components/nodemanager-service/pgdb/nodes.go
@@ -97,11 +97,10 @@ WHERE nodes.id = ANY($1)
 RETURNING nodes.name;
 `
 
+// Delete all the tags references for node_id
+// Leaving the tags alone as they might be used by other nodes
 const deleteNodeTags = `
-DELETE FROM tags
-WHERE id IN (SELECT tag_id
-             FROM nodes_tags
-             WHERE node_id = $1);
+DELETE FROM nodes_tags WHERE node_id = $1;
 `
 
 const deleteNodesSecretsByNodeId = `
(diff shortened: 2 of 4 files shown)


components/nodemanager-service/pgdb/nodes.go
```diff
@@ -370,7 +369,15 @@ func (db *DB) BulkAddNodes(inNodes []*nodes.Node) (nodeIDs []string, err error)
 		if err != nil {
 			return nodeIDs, errors.Wrap(err, "BulkAddNodes unable to translate node to db struct")
 		}
-		for _, host := range inNode.GetTargetConfig().GetHosts() {
+
+		hosts := inNode.GetTargetConfig().GetHosts()
+		if inNode.GetTargetConfig().Host != "" {
+			hosts = append(hosts, inNode.GetTargetConfig().Host)
+		}
+		if len(hosts) < 1 {
```
alexpop: @vjeffrey had this feedback here, but got lost we I opened the PR:

_👍 code makes sense, but i wonder...should we be strict and reject the call (at the api level) is a host field is non empty on a bulk nodes call, b/c they should be sticking everything in hosts?_

alexpop: A host can be passed in the BulkAddNodes payload at the moment via:

```
target_config:
  host: string
```
and/or
```
target_config:
  hosts: []string
```
My change accepts both and only errors out if none are provided.

I could be convinced as well to only accept `target_config.hosts` in the BulkAddNodes call. This wouldn't break existing scripts as we ignored `target_config.host` for this call.

components/nodemanager-service/Makefile
```diff
@@ -186,6 +186,11 @@ run-db-tests:
 	@POSTGRES_URI="$(POSTGRES_URI)" \
 		go test -v $(shell go list ./pgdb/...) -cover -database
 
+run-one-db-test:
+	@POSTGRES_URI="$(POSTGRES_URI)" \
+		go test -v $(shell go list ./pgdb/...) -run TestRunNodesIntegrationSuite/TestBulkAddNodesAndTagsUpdate -cover -database
+
+
```
phiggins: Did you mean to leave this in here?

components/nodemanager-service/pgdb/nodes.go
```diff
@@ -370,7 +369,15 @@ func (db *DB) BulkAddNodes(inNodes []*nodes.Node) (nodeIDs []string, err error)
...
+			return nodeIDs, errors.New("BulkAddNodes unable find TargetConfig hosts")
```
phiggins: This error message reads weirdly and also doesn't provide a lot of specific information about how to fix it. How about something like: `BulkAddNodes requests require host or hosts in TargetConfig`.

components/nodemanager-service/pgdb/nodes.go
vjeffrey: So expanding on that bit above, if we weren't sharing the message definition between calls, would we have ever provided users with the ability to add hosts via a singular string host field and an array of strings hosts field for an endpoint used for bulk node add?  i would argue probably not - which is why I lean towards going with the strict approach

alexpop: After sleeping on it, to avoid muddying the waters, it makes sense to only process `target_config.hosts` for the bulk endpoint. I pushed another commit where the API errors out if `target_config.host` is used for the bulk endpoint.

components/nodemanager-service/Makefile
alexpop: Yeah... I figured having a target for running a single test would speed up the dev-test cycles for others as well. I don't like that the test is hardcoded inside. Further improvement could be to use git diff to pick up the test that's being worked on and only run that.

components/nodemanager-service/pgdb/nodes.go
alexpop: Thank you, always happy to make messages clearer. Pushed change

PR #696, opened 2019-06-27
auth services: add rows.Err() checks

srenatus:
Thanks to @stevendanna's attention to detail, this makes our auth services do what needs to be done when using `database/sql`'s [`(*Rows).Next`](https://golang.org/pkg/database/sql/#Rows.Next):

> Next prepares the next result row for reading with the Scan method. *It returns true on success, or false if there is no next result row **or an error happened while preparing it**.* Err should be consulted to distinguish between the two cases.

### DIFF
--- a/components/authz-service/storage/v2/postgres/postgres.go
+++ b/components/authz-service/storage/v2/postgres/postgres.go
@@ -148,11 +148,10 @@ func (p *pg) ListPolicies(ctx context.Context) ([]*v2.Policy, error) {
 
 	var pols []*v2.Policy
 	rows, err := p.db.QueryContext(ctx,
-		`SELECT query_policies from query_policies($1);`, pq.Array(projectsFilter))
+		"SELECT query_policies from query_policies($1)", pq.Array(projectsFilter))
 	if err != nil {
 		return nil, p.processError(err)
 	}
-
 	defer func() {
 		if err := rows.Close(); err != nil {
 			p.logger.Warnf("failed to close db rows: %s", err.Error())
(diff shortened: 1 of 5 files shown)


components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -148,11 +148,10 @@ func (p *pg) ListPolicies(ctx context.Context) ([]*v2.Policy, error) {
 
 	var pols []*v2.Policy
 	rows, err := p.db.QueryContext(ctx,
-		`SELECT query_policies from query_policies($1);`, pq.Array(projectsFilter))
+		"SELECT query_policies from query_policies($1)", pq.Array(projectsFilter))
```
blakestier: Whoa I feel like we have backticks everywhere for these. . . .

components/teams-service/storage/postgres/postgres.go
```diff
@@ -335,19 +337,21 @@ func (p *postgres) PurgeUserMembership(ctx context.Context, userID string) ([]uu
 	for rows.Next() {
 		var teamID uuid.UUID
 		if err := rows.Scan(&teamID); err != nil {
-			return nil, err // TODO
+			return nil, p.processError(err) // TODO
```
blakestier: Is this still a `TODO`?

components/authz-service/storage/v2/postgres/postgres.go
srenatus: Totally doesn't matter. I just prefer to use backticks only where they're the most handy choice -- when there's `"` in the string I'd otherwise have to escape. If there's nothing to escape, I'd pick plain `"`. I've just done that because I was touching code near these anyways. It doesn't matter 😃

PR #698, opened 2019-06-27
automate-ui: cleanup statusText usage

srenatus:
- removed from test specs where it didn't matter
- replaced with error.message where it did matter
- placed a comment on it where I didn't want to change anything

What is this? HTTP/2 doesn't send a "status text". Status text is the last segment after

    HTTP/x.y <CODE> <REASON>
 
in the server response, so, where with HTTP/1.1 we see

    HTTP/1.1 401 Unauthorized

with HTTP/2, we see

    HTTP/2 401

The corresponding [RFC7540 paragraph][1] is

> HTTP/2 does not define a way to carry the version or reason phrase that is included in an HTTP/1.1 status line.

[1]: https://greenbytes.de/tech/webdav/rfc7540.html#rfc.section.8.1.2.4.p.2


srenatus:
`reset_password.component.ts`  ooooh workflow. :shiver: ;)

srenatus:
WMWG (will merge when green)

PR #700, opened 2019-06-27
[ingest-service] Use cereal library for periodic jobs

stevendanna:
Moving this to the cereal library should allow:

- Job config updates to work in a multi-node setting, and
- Multiple nodes to run the job scheduler without the job running
  twice.

*Removed Features*

- The StopJobScheduler and StartJobScheduler were not implemented in
  the cereal library.  We could implement these features, but the
  superteam indicated that these endpoints were really only for
  testing.


components/ingest-service/habitat/plan.sh
```diff
@@ -10,10 +10,15 @@ pkg_binds=(
   [authz-service]="port"
   [event-service]="port"
   [nodemanager-service]="port"
+  # TODO(ssd) 2019-05-02: We could consider making these optional
```
ryancragun: Why make them optional? We already rely on the platform config for the pg URL so if we don't have the binds then ~scheduler~ ~manager~ pg backend isn't ever going to start up.

components/ingest-service/server/delete_nodes_scheduler.go
```diff
@@ -5,18 +5,25 @@ package server
...
 )
 
 // DeleteMarkedNodes - run the delete node task now
 func (server *JobSchedulerServer) DeleteMarkedNodes(ctx context.Context,
 	empty *ingest.DeleteMarkedNodesRequest) (*ingest.DeleteMarkedNodesResponse, error) {
-	jConfig := server.configManager.GetDeleteNodesSchedulerConfig()
-	deletedNodes, err := server.client.DeleteMarkedNodes(ctx, jConfig.Threshold)
+
+	var threshold string
+	err := server.jobManager.GetScheduledWorkflowParameters(ctx, "periodic_delete_nodes", "delete_nodes", &threshold)
```
ryancragun: Strange pattern here. Why not just return the string as a value instead of giving a pointer to a string? By using a pointer we have to worry about it being `nil`. I don't see much in the way of a performance gain either because strings efficiently use pointers under the hood.

components/ingest-service/config/old_job_config.go
```diff
@@ -0,0 +1,96 @@
...
+	},
+	MissingNodesForDeletion: JobConfig{
+		ID:        MissingNodesForDeletion,
+		Threshold: "30d",
+		Every:     "15m",
+		Running:   true,
+	},
+}
+
+func readOldJobConfigFromFile(configFile string) (*OldJobConfig, error) {
```
ryancragun: kinda strange that this doesn't return an error if the file doesn't exist. I would expect it to return the error and it would be up to the caller to know whether or not they wish to proceed if the file doesn't exist.

components/ingest-service/grpc/grpc.go
```diff
@@ -181,3 +207,14 @@ func Spawn(opts *serveropts.Opts) error {
 
 	return grpcServer.Serve(conn)
 }
+
+func pgURL(pgURL string, pgDBName string) (string, error) {
+	if pgURL == "" {
+		var err error
+		pgURL, err = platform.PGURIFromEnvironment(pgDBName)
```
ryancragun: This is some real magic!

components/ingest-service/server/jobs.go
```diff
@@ -1,124 +1,256 @@
...
+	taskName string
+}
+
+func NewSingleTaskWorkflow(taskName string) *SingleTaskWorkflow {
+	return &SingleTaskWorkflow{taskName}
+}
+
+func (s *SingleTaskWorkflow) OnStart(w cereal.WorkflowInstance, ev cereal.StartEvent) cereal.Decision {
+	var params string
+	err := w.GetParameters(&params)
```
ryancragun: wtf does the json parser do when you try unmarshal a struct to a string?

components/ingest-service/server/jobs.go
```diff
@@ -1,124 +1,256 @@
...
+	"github.com/chef/automate/api/interservice/ingest"
+	"github.com/chef/automate/components/ingest-service/backend"
+	"github.com/chef/automate/components/ingest-service/config"
+	"github.com/chef/automate/lib/cereal"
 )
 
-func (server *JobSchedulerServer) addUpdateJob(jc config.JobConfig) {
-	var cronFunc func()
+func InitializeJobManager(c *cereal.Manager, client backend.Client) error {
+	err := c.RegisterTaskExecutor("delete_expired", &DeleteExpiredMarkedNodesTask{client}, cereal.TaskExecutorOpts{})
```
ryancragun: I think it'd be nice for some package to export constants for these task/workflow names and to use the constants everywhere.

components/ingest-service/server/node_missing_scheduler.go
```diff
@@ -5,17 +5,32 @@ package server
...
 	log "github.com/sirupsen/logrus"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+
+	"github.com/chef/automate/api/interservice/ingest"
+	"github.com/chef/automate/lib/cereal"
+)
+
+const (
+	markNodesMissingWorkflowName = "missing_nodes"
```
ryancragun: Yes! This, but only public everywhere that the name is used.

components/ingest-service/server/jobs.go
stevendanna: We aren't doing that.  We are unmarshalling a string into a string.  A string is valid json. 

https://play.golang.org/p/1Ti7P1AMmEG

That said, this will probably change to a struct for some other reasons in my next push.

components/ingest-service/server/delete_nodes_scheduler.go
stevendanna: You need a pointer here because GetShceduledWorkflowParameters basically unmarshal's the parameters into the pointer you give it and you can't unmarshal into a value.

components/ingest-service/habitat/plan.sh
stevendanna: I _think_ my thought when writing this comment was that it seems kinda bad for ingest to fail just because we can't run scheduled tasks.  Ingest is one of the "try to keep it working under all circumstances" services since it is what takes in data.  Ideally, imho, whole feature might live somewhere else.

components/ingest-service/server/delete_nodes_scheduler.go
stevendanna: I'm thinking in another PR, I'm going to make a GetScheduledWorkflowByName() that returns a cereal.Schedule and then the caller can do the unmarshal themselves.  It is annoying as it is then a bit more code on the caller side, but would probably avoid this confusion here.

components/ingest-service/grpc/grpc.go
jaym: :)

components/ingest-service/server/delete_nodes_scheduler.go
ryancragun: Yes! What I was trying to describe (badly) in standup was more or less that kind of an approach, but in the end I wasn't sure if I could justify it. It adds more verbosity and more boilerplate but I think you might gain a bit more clarity.

components/ingest-service/server/jobs.go
ryancragun: Yeah, your explanation of it being a straight up byte array of some go type makes sense.

components/ingest-service/server/jobs.go
```diff
@@ -1,124 +1,280 @@
 package server
 
 import (
-	"github.com/chef/automate/components/ingest-service/backend"
-	"github.com/chef/automate/components/ingest-service/config"
+	"fmt"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
```
lancewf: There are two "logrus" import lines.

stevendanna: Ah, thanks, must have use `logrus` rather than `log` down below out of habit.

components/ingest-service/server/jobs.go
```diff
@@ -1,124 +1,280 @@
...
+
+func MigrateJobsSchedule(c *cereal.Manager, oldConfigFile string) error {
+	jc, err := config.OldJobConfigFromFile(oldConfigFile)
+	if err != nil {
+		log.WithError(err).Warn("failed to read old job config from disk, defaults will be used")
+	}
+
+	for idx, name := range config.JobList {
+		scheduleName := fmt.Sprintf("periodic_%s", name)
+		config, err := jc.ConfigForJob(idx)
```
lancewf: `jc` could be `nil` here should we still use this variable?

components/ingest-service/server/jobs.go
```diff
@@ -1,124 +1,280 @@
...
+			return errors.Wrapf(err, "failed to register workflow for %q", jobName)
+		}
+	}
+	return nil
+}
+
+func MigrateJobsSchedule(c *cereal.Manager, oldConfigFile string) error {
+	jc, err := config.OldJobConfigFromFile(oldConfigFile)
+	if err != nil {
+		log.WithError(err).Warn("failed to read old job config from disk, defaults will be used")
```
lancewf: Where the defaults coming from for fresh installs?

components/ingest-service/server/jobs.go
stevendanna: ConfigForJob handles the nil case.  It is a bit ugly, and most of the ugliness is stemming from attempting to handle the case of potentially having a partial on-disk config.

components/ingest-service/integration_test/suite_test.go
```diff
@@ -359,7 +358,27 @@ func createServices(s *Suite) {
 	// // To test the 'mark missing nodes for deletion' job
 	// res, err := suite.JobSchedulerServer.MarkMissingNodesForDeletion(ctx, &req)
 	// ```
-	s.JobSchedulerServer = server.NewJobSchedulerServer(s.ingest, s.JobScheduler, s.ConfigManager)
+	jobManager, err := cereal.NewManager(postgres.NewPostgresBackend(postgresUrl))
+	if err != nil {
+		logrus.WithError(err).Fatal("could not create job manager")
+	}
+
```
stevendanna: ⚠️ `defer jobManager.Stop()`!!

components/ingest-service/config/config_mgmt.go
```diff
@@ -4,117 +4,15 @@ import (
 	"io/ioutil"
 	"os"
 
-	"github.com/chef/automate/api/interservice/ingest"
-	project_update_lib "github.com/chef/automate/lib/authz"
-	base_config "github.com/chef/automate/lib/config"
```
lancewf: I am glad to see all this config going into Postgres.

components/ingest-service/integration_test/suite_test.go
lancewf: That can be added to `GlobalTeardown` on line 106.

components/automate-deployment/pkg/backup/spec.go
```diff
@@ -303,6 +303,12 @@ func DefaultSpecs(serviceNames []string) []Spec {
 					SrcPath: "/hab/svc/ingest-service/data",
 				},
 			},
+			SyncDbsV2: []DatabaseDumpOperationV2{
+				{
+					Name: "chef_ingest_service",
```
jaym: i had a dream about this being annoying because we would have to specify the tables. i'm glad that was just a nightmare

components/automate-deployment/pkg/server/server.go
```diff
@@ -450,7 +450,7 @@ func usesPlatformScaffolding(service *deployment.Service) bool {
 	switch service.Name() {
 	case "secrets-service", "pg-sidecar-service", "teams-service", "authz-service", "session-service", "automate-dex",
 		"authn-service", "compliance-service", "nodemanager-service", "notifications-service", "applications-service",
-		"automate-cs-bookshelf", "automate-cs-oc-erchef", "automate-cs-oc-bifrost":
+		"automate-cs-bookshelf", "automate-cs-oc-erchef", "automate-cs-oc-bifrost", "ingest-service":
```
jaym: sorry
https://github.com/chef/automate/projects/3#card-24267364

inspec/a2-api-integration/controls/ingest-job-scheduler.rb
```diff
@@ -93,6 +93,7 @@ def refresh_elasticsearch()
           ).http_status
         ).to eq 200
 
+        sleep 10
```
lancewf: Did you find that you needed this sleep? The `refresh_elasticsearch` should ensure all the data is updated.

.studio/ingest-service
```diff
@@ -14,6 +14,7 @@ function ingest_integration() {
   local rc
 
   # Go based integration tests
+  export PG_URL="postgresql://ingest@127.0.0.1:10145/chef_ingest_service?sslmode=verify-ca&sslcert=/hab/svc/ingest-service/config/service.crt&sslkey=/hab/svc/ingest-service/config/service.key&sslrootcert=/hab/svc/ingest-service/config/root_ca.crt"
```
afiune: Alternatively I did this trick in the applications service to get the PG URL: https://github.com/chef/automate/blob/master/.studio/applications-service#L158

afiune: And in the integration test: https://github.com/chef/automate/blob/master/components/applications-service/integration_test/suite_test.go#L47

Both did the trick! :100:

inspec/a2-api-integration/controls/ingest-job-scheduler.rb
stevendanna: I'll look into this, but IIRC, the sleep is now needed because the job worker might take a second to pick up the task when you trigger it.

Issue #704, opened 2019-06-27
Review Chef Server gather-logs and ensure feature parity in Chef Automate gather-logs


yzl:

yzl:
Chef Automate gather-logs appears to cover the ground that Chef Server gather-logs does. If we find files/command output we would like to include, we can do that as a separate ticket.

PR #708, opened 2019-06-27
Include chef-server files in the support bundle

yzl:
Include chef-server's `etc` files in the support bundle.

We already collect the service logs and the config files.

#704 

- [ ] Code actually executed?

### DIFF
--- a/components/automate-cli/cmd/chef-automate/gather-logs.go
+++ b/components/automate-cli/cmd/chef-automate/gather-logs.go
@@ -452,6 +452,12 @@ func runGatherLogsLocalCmd(outfileOverride string, logLines uint64) error {
 	g.AddCopiesFromPath("config", "/hab/svc")
 	g.AddCopiesFromPath("logs", "/hab/svc")
 
+	// A2-chef-server-related files
+	g.AddCopiesFromPath("etc", "/hab/svc/automate-cs-bookshelf")
+	g.AddCopiesFromPath("etc", "/hab/svc/automate-cs-nginx")
+	g.AddCopiesFromPath("etc", "/hab/svc/automate-cs-oc-bifrost")
+	g.AddCopiesFromPath("etc", "/hab/svc/automate-cs-oc-erchef")
+
 	// local status
 	g.AddCommand("df_h", "df", "-h")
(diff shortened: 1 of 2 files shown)


components/automate-cli/cmd/chef-automate/gather-logs.go
```diff
@@ -452,6 +452,12 @@ func runGatherLogsLocalCmd(outfileOverride string, logLines uint64) error {
 	g.AddCopiesFromPath("config", "/hab/svc")
 	g.AddCopiesFromPath("logs", "/hab/svc")
 
+	// A2-chef-server-related files
```
jaym: Am i missing something here:
```
[272][default:/src:0]# ls /hab/svc/automate-cs-bookshelf/etc
ls: cannot access '/hab/svc/automate-cs-bookshelf/etc': No such file or directory
[273][default:/src:2]# ls /hab/svc/automate-cs-nginx/etc
ls: cannot access '/hab/svc/automate-cs-nginx/etc': No such file or directory
[275][default:/src:0]# ls /hab/svc/automate-cs-oc-bifrost/etc
ls: cannot access '/hab/svc/automate-cs-oc-bifrost/etc': No such file or directory
[276][default:/src:2]# ls /hab/svc/automate-cs-oc-bifrost/etc
ls: cannot access '/hab/svc/automate-cs-oc-bifrost/etc': No such file or directory
```

yzl: `AddCopiesFromPath` isn't a join, so the path can be in any subdirectory under `/hab/svc`; in this case, `/etc/` is mostly under `/hab/svc/<service>/var/etc`:

```
[25][default:/src:126]# ls /hab/svc/automate-cs-bookshelf/var/
etc  logs
[26][default:/src:0]# ls /hab/svc/automate-cs-nginx/var/
logs  stdout  tmp
[27][default:/src:0]# ls /hab/svc/automate-cs-oc-bifrost/var/
etc  logs
[28][default:/src:0]# ls /hab/svc/automate-cs-oc-erchef/var/
etc  logs
```

jaym:
I'm thinking if we already have the logs, we should not include these files. The only thing they contain are secret things iirc.

PR #710, opened 2019-06-27
Add telemetry to count client-run based nodes and compliance stats

kmacgugan:
We want to get a better idea of how many client-run and compliance nodes customers are hooking up to automate. This will help us to better understand if licenses are being fully utilized. 

client-run nodes are counted when no filters are applied. IAMv2.1 will
still apply filters but we can take the maximum reported.
compliance stats include filters since start and end time are often employed.

Click on the client-runs tab. Look at the network traffic in the browser tools window and find the event with a payload reporting "clientRunPureCount".
Click on the compliance tab. Look at the network traffic in the browser tools window and find the event with a payload reporting "complianceCountsWithFilters"

- [ ] Code actually executed?

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/shared/reporting/report-data.service.ts
+++ b/components/automate-ui/src/app/pages/+compliance/shared/reporting/report-data.service.ts
@@ -52,8 +52,8 @@ export class ReportDataService {
         // Data will give us nodes, platforms, profiles and environments
         // We will also report filters so that we know when these counts are constrained.
         if (data !== null) {
-          var summaryStats = data.stats
-          this.telemetryService.track("complianceCountsWithFilters", {summaryStats, filters})
+          const summaryStats = data.stats;
+          this.telemetryService.track('complianceCountsWithFilters', {summaryStats, filters});
         }
       });
   }
--- a/components/automate-ui/src/app/pages/+compliance/shared/reporting/report-data.service.ts
+++ b/components/automate-ui/src/app/pages/+compliance/shared/reporting/report-data.service.ts
@@ -1,6 +1,7 @@
 import { Injectable } from '@angular/core';
 import { remove } from 'lodash';
 import { StatsService } from './stats.service';
+import { TelemetryService } from '../../../../services/telemetry/telemetry.service';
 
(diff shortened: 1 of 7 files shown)


kmacgugan:
Open to suggestions if there is a better place to put the compliance track call. Putting it in the service meant adding a mock telemetry service to several of the unit tests. 

Also open to debate is if we want to report all node counts even when there are filters. I chose to only report when there aren't filters because this is closer to the business questions people are asking.

components/automate-ui/src/app/pages/+compliance/shared/reporting/report-data.service.ts
```diff
@@ -39,14 +40,21 @@ export class ReportDataService {
...
   getReportingSummary(filters) {
     this.statsService.getSummary(filters)
       .subscribe(data => {
         this.reportingSummaryEmpty = this.isAllZeros(data.stats);
         this.reportingSummary = data;
+        // Data will give us nodes, platforms, profiles and environments
+        // We will also report filters so that we know when these counts are constrained.
+        if (data !== null) {
+          var summaryStats = data.stats
+          this.telemetryService.track("complianceCountsWithFilters", {summaryStats, filters})
```
vjeffrey: is the filter information useful? if not, should we send it when only start time and end time filters are applied?

kmacgugan: I wasn't sure, mostly because I am not super familiar with the compliance filters. Can start and end time be set to a super narrow window? That could skew things.

vjeffrey: start time is 10 days ago beg of day, end time is end of today utc. user can only modify end time, which is always end of today utc

vjeffrey: but the data that comes back for that specific endpoint is always "just today"

vjeffrey: (or, just whatever day is selected in the cal)

PR #711, opened 2019-06-28
Properly copy swagger files into docs component.

phiggins:
To build the API docs hab package, `automate-chef-io` needs a local copy of the `.swagger.json` files generated by compiling the protos. Expeditor has a script it runs after a PR is merged with changes to the swagger files, however it was incomplete in two ways:

* It only copied the swagger files from `automate-gateway` and not from `api`.
* It copied files from a nested directory structure into one directory, but the swagger files' file names were not unique so this resulted in copying some files over the top of each other.

I'm not sure how to test the expeditor bits so I tried to keep that as minimal as possible.


components/automate-chef-io/Makefile
```diff
@@ -11,6 +11,15 @@ clean:
...
 	pushd themes/chef && git fetch && git reset --hard origin/master && popd
 
+SWAGGER_DIR=data/docs/api_chef_automate
+
+clean_swagger_files:
+	rm -rf $(SWAGGER_DIR)/*
+
+sync_swagger_files: clean_swagger_files
+	pushd ../../api/external && find . -name '*.swagger.json' | cpio -p -dumv ../../components/automate-chef-io/$(SWAGGER_DIR) && popd
+	pushd ../automate-gateway/api/ && find . -name '*.swagger.json' | cpio -p -dumv ../../automate-chef-io/$(SWAGGER_DIR) && popd
```
phiggins: This was trickier than it seemed at first because of the need to work across osx and linux. Busybox's (that's what the hab studio uses, right?) cpio supports all of these cli flags so hopefully this will just work: https://busybox.net/downloads/BusyBox.html

components/automate-chef-io/Makefile
```diff
@@ -11,6 +11,15 @@ clean:
 sync: themes/chef
 	pushd themes/chef && git fetch && git reset --hard origin/master && popd
 
+SWAGGER_DIR=data/docs/api_chef_automate
+
+clean_swagger_files:
+	rm -rf $(SWAGGER_DIR)/*
```
srenatus: This looks scary -- can we play it safe with
```suggestion
	rm -rf ./$(SWAGGER_DIR)/*
```

components/automate-chef-io/Makefile
phiggins: `cp` won't copy the directory structure, will it?

srenatus: Oh! Wasn't aware of that detail 😓 it doesn't. Ignore me 😃

phiggins: One of the issues solved in this PR is that all of the swagger files were copied into one directory, and there are two `users.swagger.json`, for example.

phiggins: This solution is one of the few I could come up with that didn't get into `sed`ing the paths of files, etc, that got pretty gross quickly.

PR #712, opened 2019-06-28
[A2-292] Property-based tests for rules

msorens:

Defining property-based tests for rules, checking real DB operations--particularly including the effect of applying rules--while ignoring the ingest service and policy refresher for this first phase.

I have reduced the number of tests from gopter's default of 100 test runs because these tests are rather slow and caused timeouts in CI.

I have also reduced globally the maximum number of things generated for lists (typically generated via `gen.SliceOf` otherwise `ListRules` tests regularly exceed the capacity of the GRPC message size (4194304). I have even further constrained the maximum number of rule conditions locally to a smaller number of items for the same reason.

Here is the contract that these tests are concerned with. I am particularly focussing on staged vs applied rules in these tests.

```
( 1) newly created rules are staged
( 2) creating rules with non-unique IDs prohibited
( 3) fetching newly created rules are staged
( 4) applying rules makes staged rules applied
( 5) reports staged rules as staged
( 6) does not report staged rules as applied
( 7) reports applied rules as applied
( 8) updating newly created staged rules remain staged
( 9) updating applied rules become staged
(10) updating applied and staged rules remain staged
(11) deleting newly created staged rules deletes them immediately
(12) applied rules that are deleted remain available until rule changes applied
(13) deleting applied rules with staged updates are deleted immediately
```

### :+1: Definition of Done

Tests have good coverage and all pass

- [ ] Code actually executed?

### DIFF
--- a/components/authz-service/server/v2/policy_test.go
+++ b/components/authz-service/server/v2/policy_test.go
@@ -21,7 +21,6 @@ import (
 	uuid "github.com/chef/automate/lib/uuid4"
 
 	api_v2 "github.com/chef/automate/api/interservice/authz/v2"
-	"github.com/chef/automate/components/authz-service/config"
 	constants_v1 "github.com/chef/automate/components/authz-service/constants/v1"
 	constants_v2 "github.com/chef/automate/components/authz-service/constants/v2"
 	"github.com/chef/automate/components/authz-service/engine"
@@ -2981,11 +2980,9 @@ func setupV2WithMigrationState(t *testing.T,
 	polV2, _, err := v2.NewPoliciesServer(ctx, l, mem_v2, writer, pl, vChan)
 	require.NoError(t, err)
 
-	eventServiceClient := &testhelpers.MockEventServiceClient{}
-	configMgr, err := config.NewManager("/tmp/.authz-delete-me")
 	require.NoError(t, err)
 	projectsSrv, err := v2.NewProjectsServer(ctx, l, mem_v2,
-		rulesRetriever, eventServiceClient, configMgr, testhelpers.NewMockPolicyRefresher())
+		rulesRetriever, testhelpers.NewMockProjectUpdateManager(), testhelpers.NewMockPolicyRefresher())
 	require.NoError(t, err)
 
 	vSwitch := v2.NewSwitch(vChan)
--- a/components/authz-service/server/v2/project.go
+++ b/components/authz-service/server/v2/project.go
@@ -30,12 +30,12 @@ import (
 	event "github.com/chef/automate/components/event-service/server"
 )
 
-// ProjectState the server state for projects
-type ProjectState struct {
+// state holds the server state for projects
+type state struct {
 	log                  logger.Logger
 	store                storage.Storage
 	engine               engine.ProjectRulesRetriever
-	ProjectUpdateManager *ProjectUpdateManager
+	projectUpdateManager ProjectUpdateMgr
 	policyRefresher      PolicyRefresher
 	applyRuleMux         sync.Mutex
 }
@@ -50,7 +50,8 @@ func NewMemstoreProjectsServer(
 	pr PolicyRefresher,
 ) (api.ProjectsServer, error) {
 
-	return NewProjectsServer(ctx, l, memstore.New(), e, eventServiceClient, configManager, pr)
+	projectUpdateManager := NewProjectUpdateManager(eventServiceClient, configManager)
+	return NewProjectsServer(ctx, l, memstore.New(), e, projectUpdateManager, pr)
 }
 
 // NewPostgresProjectsServer instantiates a ProjectsServer using a PG store
@@ -69,7 +70,8 @@ func NewPostgresProjectsServer(
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to initialize v2 store state")
 	}
-	return NewProjectsServer(ctx, l, s, e, eventServiceClient, configManager, pr)
+	projectUpdateManager := NewProjectUpdateManager(eventServiceClient, configManager)
+	return NewProjectsServer(ctx, l, s, e, projectUpdateManager, pr)
 }
 
 func NewProjectsServer(
@@ -77,8 +79,7 @@ func NewProjectsServer(
 	l logger.Logger,
 	s storage.Storage,
 	e engine.ProjectRulesRetriever,
-	eventServiceClient automate_event.EventServiceClient,
-	configManager *config.Manager,
+	projectUpdateManager ProjectUpdateMgr,
 	pr PolicyRefresher,
 ) (api.ProjectsServer, error) {
 
@@ -86,7 +87,7 @@ func NewProjectsServer(
 		log:                  l,
 		store:                s,
 		engine:               e,
-		ProjectUpdateManager: NewProjectUpdateManager(eventServiceClient, configManager),
+		projectUpdateManager: projectUpdateManager,
 		policyRefresher:      pr,
 	}, nil
 }
--- a/components/authz-service/server/v2/project.go
+++ b/components/authz-service/server/v2/project.go
@@ -30,12 +30,12 @@ import (
 	event "github.com/chef/automate/components/event-service/server"
 )
 
-// state holds the server state for projects
-type state struct {
+// ProjectState holds the server state for projects
+type ProjectState struct {
 	log                  logger.Logger
 	store                storage.Storage
 	engine               engine.ProjectRulesRetriever
-	projectUpdateManager ProjectUpdateMgr
+	ProjectUpdateManager ProjectUpdateMgr
 	policyRefresher      PolicyRefresher
 	applyRuleMux         sync.Mutex
 }
@@ -87,7 +87,7 @@ func NewProjectsServer(
 		log:                  l,
 		store:                s,
 		engine:               e,
-		projectUpdateManager: projectUpdateManager,
+		ProjectUpdateManager: projectUpdateManager,
 		policyRefresher:      pr,
 	}, nil
 }
--- a/components/authz-service/server/v2/rules_property_test.go
+++ b/components/authz-service/server/v2/rules_property_test.go
@@ -90,7 +90,8 @@ func TestCreateRuleProperties(t *testing.T) {
 	params := gopter.DefaultTestParametersWithSeed(seed)
 	params.MinSize = 1 // otherwise, we'd get zero-length "conditions" slices
 	properties := gopter.NewProperties(params)
-	properties.Property("creates a rule in storage", prop.ForAll(
+
+	properties.Property("creates a staged rule", prop.ForAll(
 		func(reqs projectAndRuleReq) bool {
 			defer testDB.Flush(t)
 			_, err := cl.CreateProject(ctx, &reqs.CreateProjectReq)
@@ -107,7 +108,15 @@ func TestCreateRuleProperties(t *testing.T) {
 				return false // bad run
 			}
 
-			r, err := cl.GetRule(ctx, &api.GetRuleReq{Id: reqs.rules[0].Id})
+			rStaged, err := cl.GetRule(ctx, &api.GetRuleReq{Id: reqs.rules[0].Id})
+			if err != nil {
+				t.Error(err.Error())
+				return false
+			}
+
+			cl.ApplyRulesStart(ctx, &api.ApplyRulesStartReq{})
+
+			rApplied, err := cl.GetRule(ctx, &api.GetRuleReq{Id: reqs.rules[0].Id})
 			if err != nil {
 				t.Error(err.Error())
 				return false
@@ -117,10 +126,13 @@ func TestCreateRuleProperties(t *testing.T) {
 			// to replicate the mapping logic in tests.
 			// Instead, as we add ReadRule, and ListRules etc, methods, we can use this
 			// testing approach to write meaningful assertions.
-			return len(r.Rule.Conditions) == len(reqs.rules[0].Conditions) &&
-				r.Rule.Id == reqs.rules[0].Id &&
-				r.Rule.Name == reqs.rules[0].Name &&
-				r.Rule.ProjectId == reqs.rules[0].ProjectId
+			result := len(rStaged.Rule.Conditions) == len(reqs.rules[0].Conditions) &&
+				rStaged.Rule.Id == reqs.rules[0].Id &&
+				rStaged.Rule.Name == reqs.rules[0].Name &&
+				rStaged.Rule.ProjectId == reqs.rules[0].ProjectId &&
+				rStaged.Rule.Status == "staged" &&
+				rApplied.Rule.Status == "applied"
+			return result
 		},
 		gopter.CombineGens(createProjectReqGen, createRuleReqGen).Map(func(vals []interface{}) projectAndRuleReq {
 			p := vals[0].(api.CreateProjectReq)
@@ -132,6 +144,7 @@ func TestCreateRuleProperties(t *testing.T) {
 			}
 		}),
 	))
+
 	properties.Property("ensures IDs are unique", prop.ForAll(
 		func(reqs projectAndRuleReq) bool {
 			defer testDB.Flush(t)
--- a/components/authz-service/server/v2/rules_property_test.go
+++ b/components/authz-service/server/v2/rules_property_test.go
@@ -49,7 +49,7 @@ func TestCreateRuleProperties(t *testing.T) {
 					return reportErrorAndYieldFalse(t, err)
 				}
 
-				return respRule.Status == "" && // TODO: this is wrong; should be "staged"
+				return respRule.Status == "staged" &&
 					ruleMatches(reqs.rules[0], *respRule)
 			},
 			createProjectAndRuleGen,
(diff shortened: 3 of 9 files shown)


components/authz-service/server/v2/policy_test.go
```diff
@@ -2981,11 +2980,9 @@ func setupV2WithMigrationState(t *testing.T,
 	polV2, _, err := v2.NewPoliciesServer(ctx, l, mem_v2, writer, pl, vChan)
 	require.NoError(t, err)
 
-	eventServiceClient := &testhelpers.MockEventServiceClient{}
-	configMgr, err := config.NewManager("/tmp/.authz-delete-me")
 	require.NoError(t, err)
 	projectsSrv, err := v2.NewProjectsServer(ctx, l, mem_v2,
-		rulesRetriever, eventServiceClient, configMgr, testhelpers.NewMockPolicyRefresher())
+		rulesRetriever, testhelpers.NewMockProjectUpdateManager(), testhelpers.NewMockPolicyRefresher())
```
msorens: I wished to mock the project update manager itself, greatly simplifying the necessary test framework by allowing focussing just on the DB. That necessitated this change in a few places. 

All the changes herein other than rules_property_test.go are part of this adjustment.

components/authz-service/server/v2/project.go
```diff
@@ -30,12 +30,12 @@ import (
...
 )
 
-// ProjectState the server state for projects
+// ProjectState holds the server state for projects
 type ProjectState struct {
 	log                  logger.Logger
 	store                storage.Storage
 	engine               engine.ProjectRulesRetriever
-	ProjectUpdateManager *ProjectUpdateManager
+	ProjectUpdateManager ProjectUpdateMgr
```
msorens: Making it an interface to allow passing a mock.

components/authz-service/testhelpers/testhelpers.go
```diff
@@ -157,6 +158,44 @@ func (tf *TestFramework) Shutdown(t *testing.T, ctx context.Context) {
 	// this TestFramework's authz instance could write some bad state while the next test is running.
 }
 
+// SetupProjectsAndRulesWithDB is a simplified test framework
+// useful for integration tests with just the DB.
+func SetupProjectsAndRulesWithDB(t *testing.T) (
```
msorens: Reinstated this simpler helper that allows focussing just on the DB for these tests.

components/authz-service/testhelpers/testhelpers.go
```diff
@@ -358,3 +397,45 @@ func WaitForWithTimeout(t *testing.T, f func() bool, timeout time.Duration, mess
...
+func (*mockProjectUpdateManager) PercentageComplete() float64 {
+	return 0
+}
+
+func (*mockProjectUpdateManager) EstimatedTimeComplete() time.Time {
+	return time.Time{}
+}
+
+func (*mockProjectUpdateManager) State() string {
+	return "not_running" // manager.stage.State
```
msorens: By always passing "not_running" that removes the ingest service from the picture on this set of integration tests.

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -18,17 +21,511 @@ import (
 )
 
 const (
-	idRegex = "^[a-z0-9-]{1,64}$"
+	idRegex        = "^[a-z0-9-]{1,64}$"
+	conditionLimit = 10 // avoid the grpc limit of 4194304
```
msorens: As mentioned in the PR preamble, ListRules could exceed GRPC capacity if the number of rule conditions was not constrained -- which could be a problem in the real world.

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -18,17 +21,511 @@ import (
...
+				cl.ApplyRulesStart(ctx, &api.ApplyRulesStartReq{})
+
+				rApplied, err := cl.GetRule(ctx, &api.GetRuleReq{Id: reqs.rules[0].Id})
+				if err != nil {
+					return reportErrorAndYieldFalse(t, err)
+				}
+
+				return rApplied.Rule.Status == "applied" &&
+					ruleMatches(updateReq, *rApplied.Rule) &&
+					rStaged.Rule.Status == "" && // TODO: this is wrong; should be "staged"
```
msorens: I will be fixing UpdateRule to fill in "staged".

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -18,17 +21,511 @@ import (
...
+
+				cl.ApplyRulesStart(ctx, &api.ApplyRulesStartReq{})
+
+				_, err = cl.GetRule(ctx, &api.GetRuleReq{Id: reqs.rules[0].Id})
+				deletedWhenApplied := err != nil && strings.Contains(err.Error(), "could not find")
+
+				return deletedWhenApplied &&
+					ruleMatches(reqs.rules[0], *rDeleted.Rule) &&
+					rDeleted.Rule.Status == "applied" && // TODO: wrong!
+					!rDeleted.Rule.Deleted // TODO: wrong!
```
msorens: I will be fixing these two items in this PR.

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -82,88 +579,94 @@ func TestCreateRuleProperties(t *testing.T) {
...
-			defer testDB.Flush(t)
-			_, err := cl.CreateProject(ctx, &reqs.CreateProjectReq)
-			if err != nil {
-				t.Error(err.Error())
-				return false // bad run
-			}
+func createRules(ctx context.Context, cl api.ProjectsClient, rules []api.CreateRuleReq) ([]api.CreateRuleReq, error) {
+	for i, rule := range rules {
+		// make the id unique!
+		rules[i].Id = fmt.Sprintf("%s-index-%d", rules[i].Id, i)
```
msorens: As soon as I introduced multiple rule generation, I ran into the issue that gopter does not make any claims on uniqueness. This patch handles that.

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -18,17 +21,511 @@ import (
...
+			createProjectAndRuleGen,
+		))
+
+	properties.TestingRun(t)
+}
+
+func getGopterParams(seed int64) *gopter.Properties {
+	params := gopter.DefaultTestParametersWithSeed(seed)
+	params.MinSize = 1             // otherwise, we'd get zero-length "conditions" slices
+	params.MaxSize = 25            // otherwise, we may exceed GRPC capacity in ListRules
```
msorens: This is a "global" setting, constraining lists to  25 items. Too bad gopter does not provide more local control; that's why I introduced the conditionLimit you'll see, to locally constrain condition lists to a smaller number.

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -18,17 +21,511 @@ import (
...
-	testFW := testhelpers.NewTestFramework(t, ctx)
-	cl := testFW.Projects
-	testDB := testFW.TestDB
-	seed := testFW.Seed
+	ctx := context.Background()
+	cl, testDB, _, _, seed := testhelpers.SetupProjectsAndRulesWithDB(t)
+	properties := getGopterParams(seed)
+	createRuleReqGen, createProjectReqGen, createProjectAndRuleGen := getGenerators()
+
+	properties.Property("creates a staged rule and confirms it is staged",
```
srenatus: 💭 Is that a _property_? I'd propose something like
> all created rules are staged

Similarly below,
> applying a staged rule makes it applied

It's a bit silly, but describing _what the test does_ does read a bit odd to me 🤔

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -18,17 +21,511 @@ import (
...
+					return reportErrorAndYieldFalse(t, err)
+				}
+
+				return rApplied.Rule.Status == "applied" &&
+					ruleMatches(reqs.rules[0], *rApplied.Rule)
+			},
+			createProjectAndRuleGen,
+		))
+
+	properties.Property("ensure creating a rule fails if the rule ID is not unique",
```
srenatus: How about this?
> non-unique rule IDs cannot be created

or something similar, not _"ensuring..."_

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -18,17 +21,511 @@ import (
...
+
+				resp, err := cl.ListRules(ctx, &api.ListRulesReq{IncludeStaged: true})
+				if err != nil {
+					return reportErrorAndYieldFalse(t, err)
+				}
+
+				return rulesMatch(reqs.rules, resp)
+
+			},
+			createProjectAndRuleGen,
```
srenatus: I guess we could have done something fancier here -- like, create a bunch of rules, apply some  of them, assert that numbers match.

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -18,17 +21,511 @@ import (
...
+		))
+
+	properties.Property("updates an applied and staged rule and applies update confirming transition from staged to applied",
+		prop.ForAll(
+			func(reqs projectAndRuleReq) bool {
+				defer testDB.Flush(t)
+
+				_, err := cl.CreateProject(ctx, &reqs.CreateProjectReq)
```
srenatus: I think we've got these two function calls so often they might also be refactored into a little helper? I mean the "create project; create rule" steps.

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -18,17 +21,511 @@ import (
...
+					return reportErrorAndYieldFalse(t, err)
+				}
+
+				_, err = cl.DeleteRule(ctx, &api.DeleteRuleReq{Id: reqs.rules[0].Id})
+				if err != nil {
+					return reportErrorAndYieldFalse(t, err)
+				}
+
+				_, err = cl.GetRule(ctx, &api.GetRuleReq{Id: reqs.rules[0].Id})
+				deletedWhenStaged := err != nil && strings.Contains(err.Error(), "could not find")
```
srenatus: Could we not use `AssertCode` here? Instead of `strings.Contains()`?
```suggestion
				deletedWhenStaged := grpc_test.AssertCode(t, codes.NotFound, err)
```
or something like that? [AssertCode](https://github.com/chef/automate/blob/6f4f5c482549dc615a99f68d3f5e87626310b1c8/lib/grpc/grpctest/status.go#L39) checks if `err != nil`, too.

components/authz-service/server/v2/rules_property_test.go
srenatus: Have you had a chance to check if there already is a github issue for the lack of more local control?

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -18,17 +21,511 @@ import (
...
+		))
+
+	properties.TestingRun(t)
+}
+
+func getGopterParams(seed int64) *gopter.Properties {
+	params := gopter.DefaultTestParametersWithSeed(seed)
+	params.MinSize = 1             // otherwise, we'd get zero-length "conditions" slices
+	params.MaxSize = 25            // otherwise, we may exceed GRPC capacity in ListRules
+	params.MinSuccessfulTests = 25 // reduced from default 100 to avoid timeouts in CI!
```
srenatus: I guess we could expand the timeouts if this is insufficient for us.

Assuming this fails because there's no output for quite a while... I wonder if "output a `"."` for every 10 runs" would be another worthy upstream issue. 🤔

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -18,17 +21,511 @@ import (
...
+	return gopter.NewProperties(params)
+}
+
+func getGenerators() (gopter.Gen, gopter.Gen, gopter.Gen) {
```
srenatus: Interesting. Could these be package-level variables, too?

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -82,88 +579,94 @@ func TestCreateRuleProperties(t *testing.T) {
...
+	t.Error(err.Error())
+	return false
+}
 
-			r, err := cl.GetRule(ctx, &api.GetRuleReq{Id: reqs.rules[0].Id})
-			if err != nil {
-				t.Error(err.Error())
+func ruleMatches(req interface{}, actual api.ProjectRule) bool {
+	if _, ok := req.(api.CreateRuleReq); ok {
+		ruleReq, _ := req.(api.CreateRuleReq)
```
srenatus: That's the first argument in the `if` above, so you could do
```go
if ruleReq, ok := req.(api.CreateRuleReq); ok {
```
instead

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -82,88 +579,94 @@ func TestCreateRuleProperties(t *testing.T) {
...
+		return len(actual.Conditions) == len(ruleReq.Conditions) &&
+			actual.Id == ruleReq.Id &&
+			actual.Name == ruleReq.Name &&
+			actual.ProjectId == ruleReq.ProjectId
+	} else {
+		ruleReq, _ := req.(api.UpdateRuleReq)
```
srenatus: Not worth checking the `ok` here?

components/authz-service/server/v2/rules_property_test.go
srenatus: I think it would be neat to fix that in the generator definiton: feels like it could be [part of the `Map` call above](https://github.com/chef/automate/pull/712/files#diff-66769720a1732f29e7206a24759eea52R584).

msorens: Definitely; good catch!

components/authz-service/server/v2/rules_property_test.go
msorens: I did, and there was not. There is a `sliceOfN` variation but that is for a fixed size `N`, not up to `N` which I would rather have. Might submit a PR for that in the future...

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -18,17 +21,445 @@ import (
...
+
+				return rApplied.Rule.Status == "applied" &&
+					ruleMatches(updateReq, *rApplied.Rule) &&
+					rStaged.Rule.Status == "" && // TODO: this is wrong; should be "staged"
+					ruleMatches(updateReq, *rStaged.Rule)
+			},
+			createProjectAndRuleGen,
+		))
+
+	properties.Property("updating applied rules become staged",
```
bcmdarroch: s/updating/updates to

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -18,17 +21,445 @@ import (
...
+				return rInitialApplied.Rule.Status == "applied" &&
+					rStaged.Rule.Status == "" && // TODO: this is wrong; should be "staged"
+					rFinalApplied.Rule.Status == "applied" &&
+					ruleMatches(updateReq, *rStaged.Rule) &&
+					ruleMatches(updateReq, *rFinalApplied.Rule)
+			},
+			createProjectAndRuleGen,
+		))
+
+	properties.Property("updating applied and staged rules remain staged",
```
bcmdarroch: s/updating/updates to

components/authz-service/server/v2/rules_property_test.go
bcmdarroch: Will that require a new migration to fix the update query? If so, I've got a small PR (coming soon) fixing up some rule queries, so maybe we could roll your fix into that migration

components/authz-service/server/v2/rules_property_test.go
msorens: See if you think the new set of names capture the spirit more correctly.

components/authz-service/server/v2/rules_property_test.go
msorens: No, we are not actually returning an object from the DB layer, so just did it in postgres.go.

components/authz-service/testhelpers/testhelpers.go
```diff
@@ -157,6 +158,44 @@ func (tf *TestFramework) Shutdown(t *testing.T, ctx context.Context) {
...
+	projectUpdateManager := NewMockProjectUpdateManager()
+	projectsSrv, err := server.NewProjectsServer(
+		ctx, l, pg, &TestProjectRulesRetriever{}, projectUpdateManager, NewMockPolicyRefresher())
+	require.NoError(t, err)
+
+	serviceCerts := helpers.LoadDevCerts(t, "authz-service")
+	connFactory := secureconn.NewFactory(*serviceCerts)
+
+	// TODO(sr): refactor our constructors. Having to maintain the middleware in
+	// three places is tedious and error-prone.
```
blakestier: 😢 I think we're over 3 now . . .

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1097,6 +1097,17 @@ func (p *pg) DeleteRule(ctx context.Context, id string) error {
 		if err != nil {
 			return p.processError(err)
 		}
+		// Code is built around expectation that rules always have at least one condition,
```
srenatus: I'd prefer we fixed the other code. This is a bit too weird.

msorens: I'm fine with that; I agree it definitely is kludgy.

PR #713, opened 2019-06-28
Add time wizard to node service card (state duration)

afiune:

The time since the last service status change. 

e.g., `3 minutes ago, 2 hours 30 minutes ago, 3 days ago.`

Keep the time format the same with the current representation in Automate.
>@afiune We currently don't have a specific format but I am proposing to standardize
it by creating some reusable code inside a shared library called `timef`.

### Sorting

**(depending):** Ascending within each criticality category, from the most recent one to the oldest.

### :+1: Definition of Done
The time wizard is displayed in the UI.

### I present you... the time wizard!
tenor-222455429

image

- Enter the studio. `hab studio enter`
- Build the applications-service. `build components/applications-service`
- Build the automate-gateway. `build components/automate-gateway`
- Build and start the automate-ui. **(Use your favorite way to do so, mine is the following)**
```
[3][default:/src:130]# export DEVPROXY_URL=localhost
[4][default:/src:130]# build components/automate-ui-devproxy
[5][default:/src:130]# start_automate_ui_background
```
- Start all services. `start_all_services`

Now you are ready to start ingesting some messages, you can do so by populating
the database with our so famous helper `applications_populate_database` and/or
you can send individual messages with `applications_publish_supervisor_message`.

https://chefio.atlassian.net/browse/A2-665
https://chefio.atlassian.net/browse/A2-897

- [x] Code actually executed?


components/applications-service/pkg/server/server.go
```diff
@@ -233,17 +236,93 @@ func convertStorageServicesToApplicationsServices(svcs []*storage.Service) []*ap
...
+		durationPretty += fmt.Sprintf("%d mins", min)
+	}
+	if sec > 0 {
+		durationPretty += fmt.Sprintf("%d seconds", sec)
+	}
+
+	return
+}
+
+func timeDiff(a, b time.Time) (year, month, day, hour, min, sec int) {
```
jaym: does https://golang.org/pkg/time/#Time.Sub not work?

afiune: We need a more verbose output, I can use it and then split the string, I guess.

afiune: I am going to try to move this code into a library so that we can use it in other parts of the system. 💯

lib/timef/time_diff.go
```diff
@@ -0,0 +1,96 @@
+// time format, time calculation, all time related things.
+package timef
```
afiune: I chose the package name `timef` since I didn't want to use the native `time` package name and it is a short for `time_format`

components/applications-service/pkg/server/server.go
```diff
@@ -236,18 +236,17 @@ func convertStorageServicesToApplicationsServices(svcs []*storage.Service) []*ap
...
-			Application:         svc.Application,
-			Environment:         svc.Environment,
-			Fqdn:                svc.Fqdn,
+			SupervisorId: svc.SupMemberID,
+			Release:      svc.FullReleaseString(),
+			Group:        svc.Group,
+			HealthCheck:  convertHealthStatusToProto(svc.Health),
+			Status:       applications.ServiceStatus(applications.ServiceStatus_value[svc.Status]),
+			Application:  svc.Application,
+			Environment:  svc.Environment, Fqdn: svc.Fqdn,
```
danielsdeleo: This some auto-formatting mishap?

components/applications-service/pkg/storage/postgres/db.go
```diff
@@ -101,10 +101,21 @@ type service struct {
...
+	HealthUpdatedAt     time.Time `db:"health_updated_at"`
 	CreatedAt           time.Time `db:"-"`
 	UpdatedAt           time.Time `db:"-"`
+
+	// (internal) use it to know if the service needs an update or not
+	needUpdate bool `db:"-"`
+}
+
+// implement the PreInsert hook for the service struct/table
+// TODO @afiune should we do this instead of SQL functions/triggers
```
srenatus: 💭 I don't think so. This is a gorp-only thing; SQL triggers are independent of the library used to talk to the database (they're even independent of the programming language used). I think that's preferrable. Related: Do you think gorp is worth it? When we've been using that in auth land, we ended up noticing that straightforward usage of `database/sql` (with `github.com/lib/pq` of course) is not really more cumbersome, and exposes more PG features, most notably transactions.

afiune: I just say it because to implement a timestamp update here is one line and in postgres, you have to implement the trigger and function: https://github.com/chef/automate/blob/master/components/applications-service/pkg/storage/postgres/schema/sql/04_timestamps.up.sql#L17-L43

components/applications-service/pkg/server/server.go
afiune: yup.

afiune: I can try to bypass it and make app env and fqdn one single line... 🤔

afiune: Changed!

afiune:
@danielsdeleo Thank you so much for your feedback, I think I would like to understand further your comment and also would love to hear how would you propose the implementation should look like, do you think we should make the time calculation in the UI? Or am I misreading your feedback? 😅

components/applications-service/pkg/storage/postgres/db.go
srenatus: I'm aware of your triggers 😉 -- I'd say the disadvantage of a few extra lines is outweighted by not being tied to a specific library 🤔

afiune: Agree! It is either or, because with the triggers and functions you have to be aware of the state of the database and maintain them. Which could be challenging in the future with X number of functions and triggers and enums, etc. etc. 😉

I love your feedback btw, this is what I was hoping for in the review process! 💯

kmacgugan: Is there an advantage to using gorp? With a trigger it seems you could do it all within a transaction which might be better for error handling?

afiune:
I'm starting to modify the time format.

components/applications-service/pkg/storage/postgres/db.go
afiune: @kmacgugan I have chatted with @srenatus about the use of `gorp` and how the auth team have used it as well as the reasons why they are not using it anymore, the TL DR is that they prefer the straightforward `SELECT ...` calls and not using the package and its features, they didn't find any blocker, it was more a preference/choice.

An important point about this move is that they don't have a place they could check in the code for the state of functions, triggers, etc. in PG and they have to start the service to run all the migration scripts and check directly in the database how the definitions look, which is why I prefer having the code at hand to verify those details without the need of starting the service and database.

So, it is up to us to use it or not. I personally like the idea of thing simple things like these line to update a timestamp but when things get complicated, I would definitely want to code it as PG functions/triggers. So a mix sounds like the way to go but I would like to hear your thoughts.

srenatus: (Here's all the details 😄)

> So, we ended up not using many of gorp's features, so it didn't seem worth the dependency -- a few straightforward `SELECT ...` calls can easily be handled using the stdlib `database/sql` methods.
>
> I think at one point or another, I was annoyed by gorp for these reasons -- which may or may not have been resolved; or actually been my fault at using the library properly, or reading the docs:
>
> a) transactions -- it didn't seems straightforward to do that with gorp
> b) context handling -- I believe this was fixed by now -- any DB call should use the caller's context, we want db actions to be aborted if the client doesn't care anymore.

> It's a bit cumbersome to maintain PG functions -- migrations have to `CREATE OR REPLACE` them again and again.
>
> On the positive side, all it takes to figure out what's going on is a `chef-automate psql chef_authz_service` session, and you can query the same functions there that the go code queries. You don't have to copy-paste long queries from Go into the psql REPL. This is a weak argument, but still :sweat_smile:
>
> As a general thing, the more happens in the database, the less do we have to worry about concurrency and stuff; like, what happens if the same grpc method is executed two times concurrently.

Tl;dr: If you know the SQL, do the SQL -- we don't need fancy wrapping 😜

afiune: @srenatus well that tl;dr is confusing since it tells me to not use it for this change and from our chat, I asked if there are blockers and/or what was your recommendation and you said:
>No, not at all. If you like working with gorp, do it. :wink: (edited) 

I think I'd be better off removing this and going SQL all the way.

srenatus: 😃 sorry I wasn't explicit there -- whatever you think is best here is fine by me, I'm not going to block this! 😄 I stand by that other comment, _if you like gorp, use it_ -- what else!

The "tl;dr" was just summing up my personal preference, I guess. 😉

kmacgugan: Thanks for exploring this topic in depth! I think it is worth it to explore the trade offs before pulling in a new dep vs using SQL-fu.

srenatus: [pedantry] It's not really _new_ -- it's been in this code (I think), and other services in a2, for a loooong time. 😄

danielsdeleo:
@afiune yeah, I think the "most ideal" solution is to put the time calculation in the UI. You can imagine that if we did something like adding localization, we wouldn't want to have to change the API, just the UI. That said, this code is written and working and would be useful for CLIs and we're not planning on doing any localization any time soon. So my remaining concern is that we don't want people to build any tooling that uses the automate API that will depend on us using a specific format for this information; also I am discussing with @apriofrost that we may want a tooltip with the actual time that the state change occurred so folks can correlate this information with graphs in their monitoring tools if they are troubleshooting an outage.

So this is a long way of saying that I think we can:
* find some way to make it clear to users of the API that they should not rely on the exact formatting of the string description of the time delta
* include the actual time in the API response, and this is what API users can use, along with a possible tooltip in our UI.

components/applications-service/pkg/storage/postgres/db.go
afiune: Yeah, it is definitely not new. 💯

afiune:
@kmacgugan @srenatus I've removed the use of `gorp` hook. Let me know if there is anything I need to modify to get your approval. Thanks for all your feedback.

afiune:
I have created a card to review the time format in a further change https://chefio.atlassian.net/browse/A2-907

PR #715, opened 2019-06-28
[cereal] Add GetWorkflowScheduleByName

stevendanna:
This removes the other "Getters" in favor of a function that returns
the entire schedule to you.

Also, while I was here, I added a missing error check.

### DIFF
--- a/lib/cereal/postgres/postgres.go
+++ b/lib/cereal/postgres/postgres.go
@@ -40,6 +40,22 @@ const (
 			ON s.instance_name = res.instance_name AND s.workflow_name = res.workflow_name;
 		`
 
+	getScheduledWorkflowQuery = `
+WITH res AS (
+    SELECT DISTINCT ON (instance_name, workflow_name) *
+    FROM cereal_workflow_results
+    ORDER BY workflow_name, instance_name, end_at DESC
+)
+SELECT s.id, enabled, s.instance_name, s.workflow_name, s.parameters,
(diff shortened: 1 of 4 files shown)


lib/cereal/postgres/postgres.go
```diff
@@ -40,6 +40,22 @@ const (
 			ON s.instance_name = res.instance_name AND s.workflow_name = res.workflow_name;
 		`
 
+	getScheduledWorkflowQuery = `
+WITH res AS (
+    SELECT DISTINCT ON (instance_name, workflow_name) *
+    FROM cereal_workflow_results
+    ORDER BY workflow_name, instance_name, end_at DESC
```
jaym: So we need to LIMIT 1 or is the LEFT JOIN going to take care of this

stevendanna: This produces a table where each workflow + instance name result has exactly one row, that table is joined with the workflow schedules, where workflow+instance name is uniq and then where have a WHERE clause that I think guarantees we only get one row. 

Maybe it would be clearer if we did `SELECT * ... ORDER BY end_at DESC LIMIT 1`, but I kinda went for symmetry with the other query.

jaym: we should try doing an explain on the query, cause i think LIMIT 1 will use some sort of priority queue "sort" with 1 element vs sorting all the things. But who knows how database actually work

PR #716, opened 2019-06-28
session-service: fix bug causing SAML refresh to fail

srenatus:
...unless there happened to be some old data stored with the session.

Before, the "is this a real proper session?" logic had depended on any
key being set in data. For local users and LDAP users, there would be a
`refresh_token`; for SAML users, there wouldn't.

Before #596, SAML users would have their already-used `client_state` and
`relay_state` keys, and their sessions would be non-empty. With that
change, however, we're supporting multiple such keys, and to avoid
accumulating them, they're removed when used. This would lead to SAML
users having empty session data.

Now, we explicitly add an `alive: true` to the session data when a new
session is requested.

### :+1: Definition of Done

See demo script 👇 

- `start_all_services`, `rebuild components/session-service`.
- login using SAML
- wait for two minutes
- observe that you haven't been logged out

- if you're keen, check session data in postgres: 
```
[13][default:/src:0]# chef-automate dev psql chef_session_service -- -c "select encode(data, 'escape') from sessions"
                                                             encode
--------------------------------------------------------------------------------------------------------------------------------
 {"data":{"alive":true,"client_state_7dphJNj8-OcRMA==":"/","relay_state_7dphJNj8-OcRMA==":true},"deadline":1561815864283360262}
(1 row)
```

- Zendesk ticket 22666

- [X] Code actually executed?

### DIFF
--- a/components/session-service/server/server.go
+++ b/components/session-service/server/server.go
@@ -64,9 +64,10 @@ type Server struct {
 }
 
 const (
+	aliveKey                 = "alive"
 	refreshTokenKey          = "refresh_token"
-	relayStateKey            = "relay_state"  // random, passed to OIDC IdP
-	clientStateKey           = "client_state" // coming from the browser, passed back on success
+	relayStateKeyPrefix      = "relay_state_"  // random, passed to OIDC IdP
+	clientStateKeyPrefix     = "client_state_" // coming from the browser, passed back on success
 	numRelayStateRandomBytes = 10
 	redirectURIKey           = "redirect_uri"
 )
@@ -220,7 +221,7 @@ func (s *Server) callbackHandler(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, fmt.Sprintf("no state in request: %q", r.Form), http.StatusBadRequest)
 		return
 	}
-	knownRelayState, err := sess.GetBool(relayStateKey + "_" + state)
+	knownRelayState, err := sess.GetBool(relayStateKeyPrefix + state)
 	if err != nil {
 		s.log.Debugf("bad session data (relay state): %v", err)
 		http.Error(w, "bad session data (relay state)", http.StatusBadRequest)
@@ -271,7 +272,7 @@ func (s *Server) callbackHandler(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	clientState, err := sess.GetString(clientStateKey + "_" + state)
+	clientState, err := sess.GetString(clientStateKeyPrefix + state)
 	s.log.Debugf("retrieved clientState %q", clientState)
 	if err != nil {
 		s.log.Debugf("bad session data (client state): %v", err)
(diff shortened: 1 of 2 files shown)


components/session-service/server/server.go
```diff
@@ -64,9 +64,10 @@ type Server struct {
 }
 
 const (
+	aliveKey                 = "alive"
 	refreshTokenKey          = "refresh_token"
-	relayStateKey            = "relay_state"  // random, passed to OIDC IdP
-	clientStateKey           = "client_state" // coming from the browser, passed back on success
+	relayStateKeyPrefix      = "relay_state"  // random, passed to OIDC IdP
```
srenatus: drive-by: call the prefixes what they are

components/session-service/server/server.go
```diff
@@ -64,9 +64,10 @@ type Server struct {
 }
 
 const (
+	aliveKey                 = "alive"
 	refreshTokenKey          = "refresh_token"
-	relayStateKey            = "relay_state"  // random, passed to OIDC IdP
-	clientStateKey           = "client_state" // coming from the browser, passed back on success
+	relayStateKeyPrefix      = "relay_state_"  // random, passed to OIDC IdP
+	clientStateKeyPrefix     = "client_state_" // coming from the browser, passed back on success
```
srenatus: drive-by cleanup

components/session-service/server/server.go
```diff
@@ -454,17 +455,19 @@ func (s *Server) refreshHandler(w http.ResponseWriter, r *http.Request) {
 	// maybeExchangeRefreshTokenForIDToken will return the still-valid idToken.
 
 	// Check if the session exists
-	// Note 2018/01/05 (sr): Couldn't find a better way.
-	if keys, err := sess.Keys(); err != nil { // nolint: vetshadow
+	if alive, err := sess.GetBool(aliveKey); err != nil {
+		s.log.Error("error retrieving session alive key")
 		httpError(w, http.StatusInternalServerError)
 		return
-	} else if len(keys) == 0 {
```
srenatus: ⚡️ This is the assumption that was invalidated for SAML users in #596.

components/session-service/server/server.go
```diff
@@ -579,20 +582,27 @@ func (s *Server) newHandler(w http.ResponseWriter, r *http.Request) {
...
-	if err := sess.PutBool(w, relayStateKey+"_"+relayState, true); err != nil {
+	if err := sess.PutBool(w, relayStateKeyPrefix+relayState, true); err != nil {
 		s.log.Errorf("couldn't put relay state into session: %s", err)
 		httpError(w, http.StatusInternalServerError)
 		return
 	}
 	s.log.Debugf("stored relayState %s", relayState)
 	authCodeURL := s.client.AuthCodeURL(relayState)
+
+	if err := sess.PutBool(w, aliveKey, true); err != nil {
```
srenatus: This way, we don't rely on having something in session data to conclude that it's a valid session; either it has `alive:true`, or it isn't.

components/session-service/server/server_test.go
```diff
@@ -33,23 +32,23 @@ const bldrURLString = "https://builder.test/"
 
 func TestSafeCmpStrings(t *testing.T) {
 	t.Run("when the strings are the same, it returns true", func(t *testing.T) {
```
srenatus: drive-by cleanup

components/session-service/server/server.go
```diff
@@ -454,17 +455,19 @@ func (s *Server) refreshHandler(w http.ResponseWriter, r *http.Request) {
...
 
 	// Check if the session exists
-	// Note 2018/01/05 (sr): Couldn't find a better way.
-	if keys, err := sess.Keys(); err != nil { // nolint: vetshadow
+	if alive, err := sess.GetBool(aliveKey); err != nil {
+		s.log.Error("error retrieving session alive key")
 		httpError(w, http.StatusInternalServerError)
 		return
-	} else if len(keys) == 0 {
+	} else if !alive {
```
blakestier: maybe I've lost track but does this need to be an `else if`? There's only two states, right?

blakestier: not worth blocking on tho

srenatus: The if branch is `err! = nil`,  when the thing couldn't be retrieved

blakestier: hmm the var `alive` assignment is confusing me. So this reads `if err is not nil, do this, else if it's not alive, do this` and then the rest is if `alive`? No extra `else` required?

msorens: Correct @blakestier.
Would it clearer if we were to remove the `else`? (That's valid because the previous block includes a `return`.) Then it is:
```
if err is not nil do this and return
if !alive do that and return
// continue with "alive" case
```

PR #721, opened 2019-06-30
Add HA Chef Server configuration test

jaym:


integration/run_test
```diff
@@ -26,6 +26,8 @@ workdir="/go/src/github.com/chef/automate"
 
 docker_run_args=(
         "--detach"
+        "--env" "CI"
```
ryancragun: Hold up yo, does this mean that when we locally test it and a failure occurs that we'll auto-kill the container?

integration/tests/ha_chef_server.sh
```diff
@@ -0,0 +1,65 @@
...
+        "$cli_bin" deploy config.toml \
+            --product chef-server \
+            --hartifacts "$test_hartifacts_path" \
+            --override-origin "$HAB_ORIGIN" \
+            --manifest-dir "$test_manifest_path" \
+            --admin-password chefautomate \
+            --accept-terms-and-mlsa
+
+    docker exec -t $_frontend1_container_name \
+        "$cli_bin" bootstrap bundle create -o bootstrap.abb
```
ryancragun: is `-o` a required switch? :thinking: I thought we went with an optional path but I guess I'm wrong.

integration/tests/ha_chef_server.sh
```diff
@@ -0,0 +1,65 @@
...
+            --override-origin "$HAB_ORIGIN" \
+            --manifest-dir "$test_manifest_path" \
+            --admin-password chefautomate \
+            --bootstrap-bundle bootstrap.abb \
+            --accept-terms-and-mlsa
+
+}
+
+do_test_deploy() {
+    docker exec --env "PATH=/hab/bin:/bin" -t $_frontend2_container_name chef-server-ctl test
```
ryancragun: Kinda sorta wish we had `jayproxy` running in front of these and then ran pedant against the proxy

integration/run_test
jaym: it passes on the CI environment variable. so if you set it to true locally, then probably.

integration/tests/ha_chef_server.sh
jaym: it's not required. it means force overwrite if the file exists which i wrote out of habit from running it locally

integration/tests/ha_chef_server.sh
jaym: i think this is something we can work towards. We need to figure out how to get the correct for of chef-server-ctl installed. I think we can use the bundle for the secrets.

PR #723, opened 2019-07-01
[event-feed-service] Apply golangci-lint

stevendanna:
This project was already very close to being lint clean, so I've
applied the linter to it and added a studio function for running the
linter.

This unfortunately duplicates some of the functionality of we have in
Makefile.common_go, but I didn't want to wade into the task of fixing
the curious blend of Makefiles, shell scripts, and studio functions
that various parts of the repository uses.

We may want to have go_component_static_tests run the linter; however,
I'll defer that until all the projects that use
go_component_static_tests are passing the linter.

### DIFF
--- a/.studio/golang
+++ b/.studio/golang
@@ -19,7 +19,7 @@ document "go_component_unit" <<DOC
   Example: Run the unit tests for the authz-service with verbosity enabled
   ----------------------------------------------------
   $ go_component_unit -v config-mgmt-service
-  
+
   The command provides tab-completion, so if you type this...
 
   $ go_component_unit<space><tab><tab>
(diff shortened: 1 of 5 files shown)


.studio/golang
```diff
@@ -260,6 +260,53 @@ function go_component_static_tests() {
 # Adding auto tab complete
 complete -F _component_auto_complete go_component_static_tests
 
+document "go_component_lint" <<DOC
+  Runs the golangci linter for a component
+DOC
+function go_component_lint() {
+  [ "x$1" == "x" ] && error "Missing component name argument; try 'describe ${FUNCNAME[0]}'" && return 1
```
srenatus: [nit] except that `describe "go_component_lint"` doesn't exactly tell me what the positional argument is supposed to be. 😉

PR #726, opened 2019-07-01
dev/acceptance envs: remove saml lies and deal with the fallout

srenatus:

It's gotten confusing that all envs claim to do SAML when just a small subset of them do. (It's got to do with some limitation on Okta, not our code.)

- [X] only have those envs _claim_ to do SAML that actually do
- [X] only have those envs configure SAML where it has been set up in Okta
- [X] adjust the e2e cypress specs to work with both SAML and local-only login pages
- [x] fill out this preamble
- [x] verify the terraform stuff works
- [x] if time permits, drive the cypress e2e tests from tags, not the bash script

- [ ] Code actually executed?

### DIFF
--- a/.expeditor/buildkite/cypress.sh
+++ b/.expeditor/buildkite/cypress.sh
@@ -15,6 +15,10 @@ instances_to_test=( \
 
 for instance in ${instances_to_test[*]}
 do
+  # these machines have no SAML setup
+  if grep -Eq 'fresh-install|iamv2p1' <<< "$instance"; then
+    export CYPRESS_SKIP_SSO=true
+  fi
   echo "--- Executing Cypress tests against $instance"
   export CYPRESS_BASE_URL="https://$instance"
   export CYPRESS_RECORD_KEY="$CYPRESS_RECORD_KEY"
--- a/.expeditor/buildkite/cypress.sh
+++ b/.expeditor/buildkite/cypress.sh
@@ -4,21 +4,15 @@ set -euo pipefail
 
 cd /workdir/e2e
 
-instances_to_test=( \
-    "a2-iamv2-local-fresh-install-${CHANNEL}.cd.chef.co" \
-    "a2-iamv2-local-inplace-upgrade-${CHANNEL}.cd.chef.co" \
-    "a2-iamv2p1-local-fresh-install-${CHANNEL}.cd.chef.co" \
-    "a2-iamv2p1-local-inplace-upgrade-${CHANNEL}.cd.chef.co" \
-    "a2-local-fresh-install-${CHANNEL}.cd.chef.co" \
-    "a2-perf-test-single-local-inplace-upgrade-${CHANNEL}.cd.chef.co" \
-)
+data=$(curl --silent https://a2-${CHANNEL}.cd.chef.co/assets/data.json)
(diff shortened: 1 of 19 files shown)


terraform/test-environments/modules/chef_automate_install/variables.tf
```diff
@@ -111,6 +111,11 @@ variable "iam_version" {
   description = "Enable major/minor versions of A2 IAM."
 }
 
+variable "saml" {
```
bcmdarroch: 🙌 thanks for making this more visible

.expeditor/buildkite/cypress.sh
```diff
@@ -4,17 +4,15 @@ set -euo pipefail
...
-    "a2-iamv2p1-local-inplace-upgrade-${CHANNEL}.cd.chef.co" \
-    "a2-local-fresh-install-${CHANNEL}.cd.chef.co" \
-    "a2-perf-test-single-local-inplace-upgrade-${CHANNEL}.cd.chef.co" \
-)
+data=$(curl --silent https://a2-${CHANNEL}.cd.chef.co/assets/data.json)
+instances_to_test=$(jq -nr --arg-json data "$data" '$data[] | select(.tags[] | contains ("e2e")) | .fqdn')
 
 for instance in ${instances_to_test[*]}
 do
+  if !jq -enr --arg fqdn "$instance" --argjson data "$data" '$data[] | select(.fqdn == $fqdn) as $i | "saml" | IN($i.tags[])'; then
```
stevendanna: I think `IN` requires at least jq 1.6, which doesn't appear available via hab, but I'm not sure what is available where this script runs.

terraform/test-environments/scenarios/airgapped_single_local_inplace_upgrade.tf
```diff
@@ -87,4 +88,7 @@ module "airgapped_single_local_inplace_upgrade_deploy" {
   channel         = "${var.channel}"
   deployment_type = "local"
   upgrade         = "true"
+
+  # SAML
+  saml = "true"
```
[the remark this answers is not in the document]
srenatus: The flow we use ("bindings" in SAML, HTTP Post and HTTP Redirect) are all relying on the client (the user's browser) to be able to reach both A2 and the SAML IdP -- that's a given even for airgapped envs, I suppose. (At least for our purposes: someone could also have airgapped clients, but we don't.)

.expeditor/buildkite/cypress.sh
srenatus: Oh, that's unfortunate. I'll see if there's anything to about that...

stevendanna: Open PR is here: https://github.com/habitat-sh/core-plans/pull/2719

Although, another odd thing is that 1.5 is uploaded as version 1.10 so we might need to figure out if there is anything we can do there.

terraform/test-environments/scenarios/airgapped_single_local_inplace_upgrade.tf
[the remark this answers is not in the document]
stevendanna: 🤦‍♂ Yes, of course, thanks!

.expeditor/buildkite/cypress.sh
srenatus: ✔️ jq code rewritten

Issue #268, opened 2019-05-07
Projects Create Ingest Rule


susanev:
_Note_: This is currently blocked by #257 

[Jira subtask](https://chefio.atlassian.net/browse/A2-826)
[Jira parent task](https://chefio.atlassian.net/browse/A2-318) with video and more context of where we are headed.

Issue #267, opened 2019-05-07
Projects detail page updates


susanev:
[Jira subtask](https://chefio.atlassian.net/browse/A2-825)
[Jira parent task](https://chefio.atlassian.net/browse/A2-318) with video and more context of where we are headed.

PR #736, opened 2019-07-01
UI-267 UI-268 Project Rules Signed-Off-By: Tara Black<tblack@chef.io>

tarablack01:
Screen Shot 2019-07-01 at 4 17 47 PM
Screen Shot 2019-07-01 at 4 18 17 PM
Screen Shot 2019-07-01 at 4 19 09 PM
Screen Shot 2019-07-01 at 4 19 28 PM
Screen Shot 2019-07-01 at 4 19 41 PM

**A2-825** makes the visual updates to the projects detail pages. Most data should be mocked until the Auth team can come back and connect it to the backend.

Please leverage the already implemented designs of the full-page designs (A2-827) when adding members to a policy.(Ideally, this should re-use the same component as the create page (A2-826), with a few conditionals or variables to allow it to cover both create and update cases.)

**A2-826** This task makes the visual updates to the project create ingest rule page. Most data should be mocked until the Auth team can come back and connect it to the backend.

**Separate story for the chef-select boxes styling** https://chefio.atlassian.net/browse/A2-826

### :+1: Definition of Done

**PROJECT DETAILS PAGE**:

Empty State

When no ingest rules exist.. display the empty state pattern

Message: Create the first ingest rule to get started!

Create Rule primary button

Ingest Rules table

Use mocked data until Auth can come back and connect it to the backend

Use the `chef-toolbar` with always enabled primary Create Rule button

The table has 4 columns: Name, Resource Type, Conditions, and Edits.

Name displays the human created name of the rule as a link

Resource Type displays Node or Event

Conditions displays the count of conditions associated with that rule, the number displayed first followed by condition in the single case or conditions when more than one conditions.

Edits communicates either Edits pending when there are edits in staging or Applied when the edits have been applied.

The control menu at the end of the row is dynamic. If the rule is not currently being applied then it should have the option to Delete Rule.

If any of the rules are in the Edits pending state then text with a link back to projects should be displayed: Edits pending, update projects to apply edits.

Delete Rule

This should only appear in the control menu when an update is not in progress

Includes a confirmation with the text _Deleting rule "<rule name>" <strong>cannot be undone</strong>. and primary button Delete Rule

**PROJECTS CREATE RULE**:
Please leverage the already implemented designs of the full-page designs (A2-827) when adding members to a policy. (Ideally, this should re-use the same component as the create page (A2-825), with a few conditionals or variables to allow it cover both create and update cases.)

On create the heading should be <Project Name>: Rule, but once they have entered characters in the rule name it should dynamically change to update Rule to match the rule's name

For now, the only options in Resource Type are Node and Event.

Condition column headers are ATTRIBUTE, OPERATOR, and VALUE – please use `text-transform` to achieve the uppercase letters

Note: Once a resource type is selected, the attribute column header should change to ATTRIBUTE <resource type>

On create an initial (empty) condition should automatically be displayed. For additional conditions, the user will need to use the Add Condition button. Conditions should be added to the bottom of the list.When more than 1 condition, the word AND should be displayed at the end of the condition row (except for on the last condition)

The only current operators are equals and member of: the user enters a string for equals and a comma-separated list for member of, but there is no visual distinction required in the input field other than the help text, given below.

The delete button removes the condition without any kind of confirmation

Save Rule should only become enabled when a name, resource type and value for at least one condition are provided.

Help text...

equals: Case sensitive, must match exactly.

member of: Comma-separated list, at least one must match exactly.

make update-ui-lib // Update chef-ui library coomponents
chef-automate iam upgrade-to-v2 --beta2.1 // Add projects
Settings > Projects

https://chefio.atlassian.net/browse/A2-825
https://chefio.atlassian.net/browse/A2-826
https://github.com/chef/automate/issues/267
https://github.com/chef/automate/issues/268

- [x] Code actually executed?

### DIFF
--- a/components/automate-ui/src/app/entities/rules/rule.actions.ts
+++ b/components/automate-ui/src/app/entities/rules/rule.actions.ts
@@ -25,7 +25,7 @@ export interface RuleSuccessPayload {
   rule: Rule;
 }
 
-export class GetRules implements Action {
+export class GetRulesForProject implements Action {
   readonly type = RuleActionTypes.GET_ALL;
 
   constructor(public payload: { project_id: string }) { }
@@ -117,7 +117,7 @@ export class UpdateRuleFailure implements Action {
 }
 
 export type RuleActions =
-  | GetRules
+  | GetRulesForProject
   | GetRulesSuccess
   | GetRulesFailure
   | GetRule
--- a/components/automate-ui/src/app/entities/rules/rule.actions.ts
+++ b/components/automate-ui/src/app/entities/rules/rule.actions.ts
@@ -1,0 +1,134 @@
+import { HttpErrorResponse } from '@angular/common/http';
+import { Action } from '@ngrx/store';
+
+import { Rule } from './rule.model';
+
+export enum RuleActionTypes {
+  GET_ALL         = 'RULES::GET_ALL',
+  GET_ALL_SUCCESS = 'RULES::GET_ALL::SUCCESS',
+  GET_ALL_FAILURE = 'RULES::GET_ALL::FAILURE',
+  GET             = 'RULES::GET',
+  GET_SUCCESS     = 'RULES::GET::SUCCESS',
+  GET_FAILURE     = 'RULES::GET::FAILURE',
+  CREATE          = 'RULES::CREATE',
+  CREATE_SUCCESS  = 'RULES::CREATE::SUCCESS',
+  CREATE_FAILURE  = 'RULES::CREATE::FAILURE',
+  DELETE          = 'RULES::DELETE',
+  DELETE_SUCCESS  = 'RULES::DELETE::SUCCESS',
+  DELETE_FAILURE  = 'RULES::DELETE::FAILURE',
+  UPDATE          = 'RULES::UPDATE',
+  UPDATE_SUCCESS  = 'RULES::UPDATE::SUCCESS',
+  UPDATE_FAILURE  = 'RULES::UPDATE::FAILURE'
+}
+
+export interface RuleSuccessPayload {
+  rule: Rule;
+}
+
+export class GetRules implements Action {
+  readonly type = RuleActionTypes.GET_ALL;
+
+  constructor(public payload: { project_id: string }) { }
+}
+
+export interface GetRulesSuccessPayload {
+  rules: Rule[];
+}
+
+export class GetRulesSuccess implements Action {
+  readonly type = RuleActionTypes.GET_ALL_SUCCESS;
+
+  constructor(public payload: GetRulesSuccessPayload) { }
+}
+
+export class GetRulesFailure implements Action {
+  readonly type = RuleActionTypes.GET_ALL_FAILURE;
+
+  constructor(public payload: HttpErrorResponse) { }
+}
+
+export class GetRule implements Action {
+  readonly type = RuleActionTypes.GET;
+
+  constructor(public payload: { id: string }) { }
+}
+
+export class GetRuleSuccess implements Action {
+  readonly type = RuleActionTypes.GET_SUCCESS;
+
+  constructor(public payload: RuleSuccessPayload) { }
+}
+
+export class GetRuleFailure implements Action {
+  readonly type = RuleActionTypes.GET_FAILURE;
+
+  constructor(public payload: HttpErrorResponse) { }
+}
+
+export class CreateRule implements Action {
+  readonly type = RuleActionTypes.CREATE;
+  constructor(public payload: {project_id: string, rule: Rule}) { }
+}
+
+export class CreateRuleSuccess implements Action {
+  readonly type = RuleActionTypes.CREATE_SUCCESS;
+  constructor(public payload: RuleSuccessPayload) { }
+}
+
+export class CreateRuleFailure implements Action {
+  readonly type = RuleActionTypes.CREATE_FAILURE;
+  constructor(public payload: HttpErrorResponse) { }
+}
+
+export class DeleteRule implements Action {
+  readonly type = RuleActionTypes.DELETE;
+
+  constructor(public payload: { id: string }) { }
+}
+
+export class DeleteRuleSuccess implements Action {
+  readonly type = RuleActionTypes.DELETE_SUCCESS;
+
+  constructor(public payload: { id: string }) { }
+}
+
+export class DeleteRuleFailure implements Action {
+  readonly type = RuleActionTypes.DELETE_FAILURE;
+
+  constructor(public payload: HttpErrorResponse) { }
+}
+
+export class UpdateRule implements Action {
+  readonly type = RuleActionTypes.UPDATE;
+
+  constructor(public payload: { rule: Rule }) { }
+}
+
+export class UpdateRuleSuccess implements Action {
+  readonly type = RuleActionTypes.UPDATE_SUCCESS;
+
+  constructor(public payload: RuleSuccessPayload) { }
+}
+
+export class UpdateRuleFailure implements Action {
+  readonly type = RuleActionTypes.UPDATE_FAILURE;
+
+  constructor(public payload: HttpErrorResponse) { }
+}
+
+export type RuleActions =
+  | GetRules
+  | GetRulesSuccess
+  | GetRulesFailure
+  | GetRule
+  | GetRuleSuccess
+  | GetRuleFailure
+  | CreateRule
+  | CreateRuleSuccess
+  | CreateRuleFailure
+  | DeleteRule
+  | DeleteRuleSuccess
+  | DeleteRuleFailure
+  | UpdateRule
+  | UpdateRuleSuccess
+  | UpdateRuleFailure;
--- a/components/automate-ui/src/app/entities/rules/rule.reducer.ts
+++ b/components/automate-ui/src/app/entities/rules/rule.reducer.ts
@@ -23,7 +23,7 @@ const CREATE_ERROR = 'createError';
 const DELETE_STATUS = 'deleteStatus';
 const UPDATE_STATUS = 'updateStatus';
 
-export const Attributes = {
+export const ruleAttributes = {
   node: [
     {
       key: 'CHEF_ORGS',
@@ -69,54 +69,13 @@ export const Attributes = {
 export const ruleEntityAdapter: EntityAdapter<Rule> = createEntityAdapter<Rule>();
 
 export const RuleEntityInitialState: RuleEntityState = ruleEntityAdapter.getInitialState({
+  getAttributes: ruleAttributes,
   getAllStatus: EntityStatus.notLoaded,
   getStatus: EntityStatus.notLoaded,
   createStatus: EntityStatus.notLoaded,
   createError: null,
   deleteStatus: EntityStatus.notLoaded,
-  updateStatus: EntityStatus.notLoaded,
-  getAttributes: {
-    node: [
-      {
-        key: 'CHEF_ORGS',
-        value: 'Chef Organization'
-      },
-      {
-        key: 'CHEF_SERVER',
-        value: 'Chef Server'
-      },
-      {
-        key: 'CHEF_ENV',
-        value: 'Environment'
-      },
-      {
-        key: 'CHEF_ROLE',
-        value: 'Chef Role'
-      },
-      {
-        key: 'CHEF_TAG',
-        value: 'Chef Tag'
-      },
-      {
-        key: 'CHEF_POLICY_NAME',
-        value: 'Chef Policy Name'
-      },
-      {
-        key: 'CHEF_POLICY_GROUP',
-        value: 'Chef Policy Group'
-      }
-    ],
-    event: [
-      {
-        key: 'CHEF_ORGS',
-        value: 'Chef Organization'
-      },
-      {
-        key: 'CHEF_SERVER',
-        value: 'Chef Server'
-      }
-    ]
-  }
+  updateStatus: EntityStatus.notLoaded
 });
 
 export function ruleEntityReducer(
--- a/components/automate-ui/src/app/entities/rules/rule.reducer.ts
+++ b/components/automate-ui/src/app/entities/rules/rule.reducer.ts
@@ -1,0 +1,102 @@
+import { EntityState, EntityAdapter, createEntityAdapter } from '@ngrx/entity';
+import { HttpErrorResponse } from '@angular/common/http';
+import { set, pipe, unset } from 'lodash/fp';
+
+import { EntityStatus } from 'app/entities/entities';
+import { RuleActionTypes, RuleActions } from './rule.actions';
+import { Rule } from './rule.model';
+
+export interface RuleEntityState extends EntityState<Rule> {
+  getAllStatus: EntityStatus;
+  getStatus: EntityStatus;
+  createStatus: EntityStatus;
+  createError: HttpErrorResponse;
+  updateStatus: EntityStatus;
+  deleteStatus: EntityStatus;
+}
+
+const GET_ALL_STATUS = 'getAllStatus';
+const GET_STATUS = 'getStatus';
+const CREATE_STATUS = 'createStatus';
+const CREATE_ERROR = 'createError';
+const DELETE_STATUS = 'deleteStatus';
+const UPDATE_STATUS = 'updateStatus';
+
+export const ruleEntityAdapter: EntityAdapter<Rule> = createEntityAdapter<Rule>();
+
+export const RuleEntityInitialState: RuleEntityState = ruleEntityAdapter.getInitialState({
+  getAllStatus: EntityStatus.notLoaded,
+  getStatus: EntityStatus.notLoaded,
+  createStatus: EntityStatus.notLoaded,
+  createError: null,
+  deleteStatus: EntityStatus.notLoaded,
+  updateStatus: EntityStatus.notLoaded
+});
+
+export function ruleEntityReducer(
+  state: RuleEntityState = RuleEntityInitialState,
+  action: RuleActions): RuleEntityState {
+
+  switch (action.type) {
+    case RuleActionTypes.GET_ALL:
+      return set(GET_ALL_STATUS, EntityStatus.loading, state);
+
+    case RuleActionTypes.GET_ALL_SUCCESS:
+      return set(GET_ALL_STATUS, EntityStatus.loadingSuccess,
+        ruleEntityAdapter.addAll(action.payload.rules, state));
+
+    case RuleActionTypes.GET_ALL_FAILURE:
+      return set(GET_ALL_STATUS, EntityStatus.loadingFailure, state);
+
+    case RuleActionTypes.GET:
+      return set(GET_STATUS, EntityStatus.loading, state);
+
+    case RuleActionTypes.GET_SUCCESS:
+      return set(GET_STATUS, EntityStatus.loadingSuccess,
+        ruleEntityAdapter.addOne(action.payload.rule, state));
+
+    case RuleActionTypes.GET_FAILURE:
+      return set(GET_STATUS, EntityStatus.loadingFailure, state);
+
+    case RuleActionTypes.CREATE:
+      return set(CREATE_STATUS, EntityStatus.loading, state);
+
+    case RuleActionTypes.CREATE_SUCCESS:
+      return pipe(
+        set(CREATE_STATUS, EntityStatus.loadingSuccess),
+        unset(CREATE_ERROR)
+      )(ruleEntityAdapter.addOne(action.payload.rule, state)) as RuleEntityState;
+
+    case RuleActionTypes.CREATE_FAILURE:
+      return pipe(
+        set(CREATE_STATUS, EntityStatus.loadingFailure),
+        set(CREATE_ERROR, action.payload)
+      )(state) as RuleEntityState;
+
+    case RuleActionTypes.DELETE:
+      return set(DELETE_STATUS, EntityStatus.loading, state);
+
+    case RuleActionTypes.DELETE_SUCCESS:
+      return set(DELETE_STATUS, EntityStatus.loadingSuccess,
+        ruleEntityAdapter.removeOne(action.payload.id, state));
+
+    case RuleActionTypes.DELETE_FAILURE:
+      return set(DELETE_STATUS, EntityStatus.loadingFailure, state);
+
+    case RuleActionTypes.UPDATE:
+      return set(UPDATE_STATUS, EntityStatus.loading, state);
+
+    case RuleActionTypes.UPDATE_SUCCESS:
+      return set(UPDATE_STATUS, EntityStatus.loadingSuccess,
+        ruleEntityAdapter.updateOne({
+          id: action.payload.rule.id,
+          changes: action.payload.rule
+        }, state));
+
+    case RuleActionTypes.UPDATE_FAILURE:
+      return set(UPDATE_STATUS, EntityStatus.loadingFailure, state);
+
+    default:
+      return state;
+  }
+}
--- a/components/automate-ui/src/app/entities/rules/rule.reducer.ts
+++ b/components/automate-ui/src/app/entities/rules/rule.reducer.ts
@@ -7,6 +7,7 @@ import { RuleActionTypes, RuleActions } from './rule.actions';
 import { Rule } from './rule.model';
 
 export interface RuleEntityState extends EntityState<Rule> {
+  getAttributes: any;
   getAllStatus: EntityStatus;
   getStatus: EntityStatus;
   createStatus: EntityStatus;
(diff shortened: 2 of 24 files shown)


components/automate-ui/src/app/pages/project/details/project-details.component.ts
```diff
@@ -76,6 +109,39 @@ export class ProjectDetailsComponent implements OnDestroy {
...
+      ? 'Edits pending'
+      : 'Applied';
+  }
+
+  hasEditsPending(rules: Rule[]): boolean {
+    return _find(rules, ['edits', 'staging']) ? true : false;
+  }
+
+  // Note: This will be dealt with later, right now, we don't check if it's used
+  inUseMessage(): string {
```
bcmdarroch: How will this be used?

components/automate-ui/src/app/pages/project/details/project-details.component.html
```diff
@@ -52,5 +53,73 @@
...
+                <chef-th class="controls"></chef-th>
+              </chef-tr>
+            </chef-thead>
+            <chef-tbody>
+              <chef-tr *ngFor="let rule of rules$ | async">
+                <chef-td>
+                  <a [routerLink]="['/settings', 'projects', project?.id, 'rules', rule.id]">{{ rule.name }}</a>
+                </chef-td>
+                <chef-td>{{ rule.type | titlecase }}</chef-td>
+                <chef-td>{{ rule.conditions.length }} conditions</chef-td>
```
bcmdarroch: could you make `conditions` singular when there's just one condition?

susanev:
im going to take a closer look tomorrow but i did notice that the top-nav is showing up on the ingest rule page, it should not be there.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -0,0 +1,106 @@
+<chef-page
```
msorens: This page is a bit too tall, when viewed on my 15" Mac. If you compare to the previous use in policy-add-members.component.html, that page fits without any vertical scrolling. This page seems to have extra vertical whitespace that could be removed so it does not scroll on a standard 15" Mac (looks to be about 766 pixels high).

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  ngOnDestroy() {
+    this.isDestroyed.next(true);
+    this.isDestroyed.complete();
+  }
+
+  getHeading() {
+    return (this.ruleForm.value.name || !this.project)
+      ? `${this.ruleForm.value.name}`
+      : `${this.project.name}: Rule`;
```
msorens: You might not have a project for a tiny fraction of time, but I don't think that needs special casing here.
More significantly, the project name should always appear.
Taken together, getHeading should just return this:
```
return `${this.project.name}: ` + (this.ruleForm.value.name || "Rule")
```

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -0,0 +1,106 @@
...
+          <div class="condition-item"
+            [formGroupName]="i"
+            *ngFor="let condition of ruleForm.get('conditions').controls; let i = index">
+
+            <chef-form-field class="attribute-field">
+              <label>
+                <span class="label">{{ ruleForm.get('type').value | uppercase }} Attribute</span>
+                <select ngDefaultControl formControlName="attribute">
+                  <option value='CHEF_SERVER'>Chef Server</option>
+                  <option value='CHEF_ORGS'>Chef Organization</option>
```
msorens: This list should not be hard-coded. The items in the list need to update dynamically based on the resource type selection, as itemized on https://chefio.atlassian.net/browse/A2-318.

susanev: oh sorry this is my fault, i told @tarablack01 this was okay and the auth team could do that update. would you prefer that get addressed in this pr @msorens?

msorens:
There is enough complexity in here that warrants unit tests. Here are just a few ideas--not a complete list:

*Create/Details Page*
1. the state of the page title changing from "<project>: Rule" to "<project>: <rule name>" once a rule name exists;
2. the choices in the attribute list change based on the selection of resource type;
3. the save button is disabled until a condition is entered.
4. Attribute column heading defaults to "Attribute" but changes to "Attribute Node" or "Attribute Event" when those resource types selected.
5. A single empty condition appears by default.
6. When multiple conditions appear the word "AND" appears between them.

*List Page*
1. You see the "Create the first ingest rule to get started" message when no rules, and it goes away when some rules.
2. Condition count column in the table accurately reflects the number of conditions on child rules.
3. A rule has a delete options available except when rule is currently being applied.
4. There's a link back to project for rule in "edits pending" state.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
msorens: No preference--I am just going by the spec. 😁

components/automate-ui/src/app/entities/rules/rule.actions.ts
```diff
@@ -0,0 +1,134 @@
...
+  CREATE          = 'RULES::CREATE',
+  CREATE_SUCCESS  = 'RULES::CREATE::SUCCESS',
+  CREATE_FAILURE  = 'RULES::CREATE::FAILURE',
+  DELETE          = 'RULES::DELETE',
+  DELETE_SUCCESS  = 'RULES::DELETE::SUCCESS',
+  DELETE_FAILURE  = 'RULES::DELETE::FAILURE',
+  UPDATE          = 'RULES::UPDATE',
+  UPDATE_SUCCESS  = 'RULES::UPDATE::SUCCESS',
+  UPDATE_FAILURE  = 'RULES::UPDATE::FAILURE'
+}
```
scottopherson: I thought about creating a separate set of Rule actions/reducer/requests/etc for the projects-list rule updates stuff, but ended up just sticking them along with [the existing Project actions](https://github.com/chef/automate/pull/511/files#diff-77f2a4bd0e110812355e14cafa2cae8fR22). Though I think the actions I've added would belong in this set of actions you've added.

Should we tweak the naming to be a little more specific? Like `IngestRule` or `ProjectRule` or something? I only ask there's already an existing concept of `Rule` in the UI for "notification rules": https://github.com/chef/automate/blob/2de87fde778344f37a0e6d10a115175657775bcd/components/automate-ui/src/app/services/rules/rules.service.ts.

bcmdarroch: Yeah I'd second adding these to project actions, since rules can't exist without projects. 

On the API we use `ProjectRule`

susanev:
Alright, I did a more detailed pass, here are a few more things id love to get fixed.
*project detail page*
- _cannot be undone_ in the delete modal should be in strong tags

*ingest rule page*
- save should not be enabled when they have zero conditions
- delete buttons, should be using the tertiary button (@bcmdarroch mentioned this too)
- top-nav shouldn't be visible 
- the heading should always be <project name>: <rule name> currently if you edit the rule it updates the heading to only be the rule name
- is there any way to use text-transform for the EVENT and NODE before the ATTRIBUTE? that way screen readers don't read it letter by letter like an acronym.
- help text on the inputs should be 12px

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
bcmdarroch: here are the allowed attributes for each rule type ([from this story](https://chefio.atlassian.net/browse/A2-318)):
```
NODE ATTRIBUTES
Chef Organization (string)
Chef Server (string)
Environment (string)
Chef Role (string[])
Chef Tag (string[])
Chef Policy Name (string)
Chef Policy Group (string)

EVENT ATTRIBUTES
Chef Organization (string)
Chef Server (string)
```

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
tarablack01: @msorens The requirements is:
On create the heading should be <Project Name>: Rule, but once they have entered characters in the rule name it should dynamically change to update Rule to match the rule's name.

components/automate-ui/src/app/pages/project/details/project-details.component.ts
tarablack01: @bcmdarroch Isn't needed.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
tarablack01: @msorens That top nav needed to be hidden.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
tarablack01: @msorens I misread the requirement, thanks for the catch!

components/automate-ui/src/app/entities/rules/rule.reducer.ts
```diff
@@ -30,7 +74,49 @@ export const RuleEntityInitialState: RuleEntityState = ruleEntityAdapter.getInit
   createStatus: EntityStatus.notLoaded,
   createError: null,
   deleteStatus: EntityStatus.notLoaded,
-  updateStatus: EntityStatus.notLoaded
+  updateStatus: EntityStatus.notLoaded,
+  getAttributes: {
+    node: [
```
msorens: Please use the constant `Attributes` that you just defined above instead of duplicating the list here.

components/automate-ui/src/app/pages/project/details/project-details.component.html
```diff
@@ -78,7 +78,7 @@
                   <a [routerLink]="['/settings', 'projects', project?.id, 'rules', rule.id]">{{ rule.name }}</a>
                 </chef-td>
                 <chef-td>{{ rule.type | titlecase }}</chef-td>
-                <chef-td>{{ rule.conditions.length }} conditions</chef-td>
+                <chef-td>{{ getConditionsText(rule.conditions.length) }}</chef-td>
```
msorens: Please use the `pluralize` pipe to get the plural here. That way it obviates the need for a custom function every time.

```suggestion
                <chef-td>{{ rule.conditions.length | pluralize : 'condition' : 's' }}  </chef-td>
```

Other example in the code base: https://github.com/chef/automate/blob/98cfe6344fab6dd9636317e0ce2dcc2128341643/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html#L47

tarablack01:
@susanev check the UI for your requested changes, let me know if it all looks good.

susanev:
@tarablack01 sorry sorry, still a few more small things to fix...

in the [zeplin designs](https://zpl.io/aMWKm1N) you should be able to click on any element to see its style info in the right panel, let me know if you don't have access.

* i still cant get the 3-dot menus on ingest rules to open
* input label styles (rule name, resource type, conditions)
    * `font-size: 14px`
    * `font-weight: 600`
* for inputs rule name and resource type add an asterisk at the end of the label indicating required
* there should be 30px of space between inputs
* condition column label styles (attribute, operator, value)
    * `font-size: 12px`
    * `font-weight: 200`
* "All conditions must be true for the rule to be true" should be `font-size: 12px`

susanev:
@tarablack01 oh i just thought of something else with regards to suppressing the last delete icon button on the condition row. i dont think that will work for multiples because we still want them to be able to delete any one of them. so i think we should only hide that delete icon button when there is exactly one row.

components/automate-ui/src/app/pages/project/details/project-details.component.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+    });
+  });
+
+  describe('when there are rules', () => {
+    beforeEach(() => {
+      store.dispatch(new GetRulesSuccess({
+          rules: rules
+      }));
+      component.project = project;
+      component.rules = rules;
```
msorens: The `dispatch` here is not doing anything--you can see that by commenting it out and you'll see that all tests still pass. That means you are not actually testing the block beginning `this.store.select(projectFromRoute)...` in the component. That would be worthwhile but would require some rework on that block if you wanted to do it. 
For the tests you have here, then, I would suggest delete the `dispatch` above and also delete the first test `rules array should have two rules` because it is really a tautology. The remaining tests here are useful with just setting the properties on L169-170 above.
(These comments applies to the prior section on "no rules", too.)

components/automate-ui/src/app/pages/project/details/project-details.component.spec.ts
```diff
@@ -0,0 +1,199 @@
...
+      fixture.detectChanges();
+    });
+
+    it('rules array should have two rules', () => {
+      expect(component.rules.length).toBe(2);
+    });
+
+    it('show rules section when rules tab is selected', () => {
+      component.onTabChange({ target: { value: 'rules'} });
```
msorens: This falls under the category of what I call "phantom tests". If you delete L179 the test still passes. So a "pass" is telling us nothing; here's a suggestion on how to make it tell us "something".
```
it('shows/hides rules section when based on selection', () => {
      component.onTabChange({ target: { value: 'details'} });
      expect(component.showTab('rules')).toBeFalsy();
      component.onTabChange({ target: { value: 'rules'} });
      expect(component.showTab('rules')).toBeTruthy();
    });
```
And I would do the analogous test for the details section.

(These comments applies to the prior section on "no rules", too.)

(Reference on phantom tests: [The Phantom Menace in Unit Testing](https://www.red-gate.com/simple-talk/dotnet/software-testing/the-phantom-menace-in-unit-testing/))

components/automate-ui/src/app/pages/project/rules/project-rules.component.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+        })
+      ],
+      providers: [
+        FeatureFlagsService
+      ]
+    }).compileComponents();
+  }));
+
+  beforeEach(() => {
+    jasmine.addMatchers(customMatchers);
```
msorens: customMatchers are not being used in this file so can delete this line

components/automate-ui/src/app/pages/project/rules/project-rules.component.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+      component.project = project;
+      component.rule = <Rule>{};
+      component.isLoading = false;
+      fixture.detectChanges();
+    });
+
+    it('the heading should show <Project Name>: Rule', () => {
+      const heading = `${component.project.name}: Rule`;
+      const compHeading = component.getHeading();
+      expect(heading).toBe(compHeading);
```
msorens: Short enough that you could inline the `compHeading` variable:
```suggestion
      expect(heading).toBe(component.getHeading());
```
Same comment applies to next couple tests below.

components/automate-ui/src/app/pages/project/rules/project-rules.component.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+      let conditions = component.ruleForm.get('conditions').value.length;
+      expect(conditions).toBe(1);
+      component.deleteCondition(0);
+      conditions = component.ruleForm.get('conditions').value.length;
+      expect(conditions).toBe(0);
+    });
+
+    it('should return condition value of string', () => {
+      const conditionValue = component.getConditionValue('adasd');
+      expect(typeof conditionValue === 'string').toBeTruthy();
```
msorens: The return value of `getConditionValue` is string so this is really a tautology.
But the contract of this test could be satisfied by checking the actual value:
```suggestion
      expect(component.getConditionValue('adasd')).toBe('adasd');
```
The same expectation could be used for the singleton array in the next test.
Then, add a third test that includes an array with more than one value, expecting, e.g.
```
expect(component.getConditionValue(['one','two','three'])).toBe('one,two,three');
```

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -0,0 +1,110 @@
...
+          <label>
+            <span class="label">Resource Type</span>
+            <select id="resourceType" ngDefaultControl formControlName="type">
+              <option value='NODE'>Node</option>
+              <option value='EVENT'>Event</option>
+            </select>
+          </label>
+          <chef-error
+            *ngIf="ruleForm.get('type').hasError('required') && ruleForm.get('type').dirty"
+          >Resource Type is required</chef-error>
```
msorens: Suggest delete this `<chef-error>` element--it can never be triggered on a `<select>` list.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -0,0 +1,110 @@
...
+              <label>
+                <span class="label">{{ getAttributeLabel() }}</span>
+                <select ngDefaultControl formControlName="attribute">
+                  <option 
+                    *ngFor="let option of attributes[ruleForm.get('type').value.toLowerCase()]; let i = index"
+                    [value]="option.key"
+                  >{{ option.value }}</option>
+                </select>
+              </label>
+            </chef-form-field>
```
msorens: A couple points (probably @susanev should weigh in on these), but comparing to her designs:
(1) The attribute field is not supposed to get wider once you select a resource type.
(2) In the requirements video, the attribute, operator, and value fields have much less whitespace between them.
image

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -0,0 +1,110 @@
...
+            
+            <chef-form-field class="attribute-field">
+              <label>
+                <span class="label">Value</span>
+                <chef-input ngDefaultControl
+                  formControlName="values"
+                  [value]="getConditionValue(condition.controls.values.value)"
+                  (change)="updateConditionValues(condition)"
+                ></chef-input>
+              </label>
```
msorens: Need `<chef-error>` element here with a check for required and non-blank pattern:
```
<chef-error *ngIf="(ruleForm.get('values').hasError('required') || ruleForm.get('values').hasError('pattern')) && ruleForm.get('values').dirty">
Value is required.
</chef-error>
```
The problem, however, is that this field is in a list, so the name `values` won't work (and won't work in L79 above either). I am not sure how to do the validation for a list element like this.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -0,0 +1,110 @@
...
+          >Name is required</chef-error>
+        </chef-form-field>
+
+        <chef-form-field class="display3">
+          <label>
+            <span class="label">Resource Type</span>
+            <select id="resourceType" ngDefaultControl formControlName="type">
+              <option value='NODE'>Node</option>
+              <option value='EVENT'>Event</option>
+            </select>
```
msorens: Requirement on L35 (resource type cannot be changed once saved) needs to be implemented here.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -0,0 +1,220 @@
...
+    this.isDestroyed.next(true);
+    this.isDestroyed.complete();
+  }
+
+  getHeading() {
+    return `${this.project.name}: ` + (this.ruleForm.value.name || 'Rule');
+  }
+
+  getAttributeLabel() {
+    return `${this.ruleForm.get('type').value.toLowerCase()} attribute`;
```
msorens: Delete `.toLowerCase()` The only reason that is needed is for the unit test, so should do that transform in the test itself.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -0,0 +1,110 @@
...
+  confirm-btn-text="Save Rule"
+  [attr.heading]="getHeading()"
+  [attr.disable-confirm]="!ruleForm.valid"
+  [attr.page-loading]="isLoading"
+  [attr.confirm-loading]="saving"
+  (confirm)="saveRule()"
+  (close)="closePage()">
+  <div *ngIf="!isLoading">
+    <ng-container>
+      <form id="ruleForm" [formGroup]="ruleForm" (ngSubmit)="saveRule()">
```
msorens: Hmm... editing a rule to add, delete, or change condition(s) does not work. I can add a second condition to a rule with one condition and save it, but then when re-opening the rule the second condition does not show up. Similarly I can delete a condition from a rule with 4 conditions, but it comes back when I re-open. Same thing with editing a value.
So any kind of condition editing is not being saved.

Oh, probably the back-end wiring is not done, right? So this would be for another card.

msorens:
This is a challenging task! I've noted a number of fixes needed, but on the whole, looking quite reasonable, so do not get discouraged!!

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
susanev: oh the selects are fine for now, @vinay033 is going to do the task fo styling the select menus.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
susanev: yup, that should be part of the follow-up work that the auth team does

components/automate-ui/src/app/pages/project/rules/project-rules.component.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+      const compBackRoute = component.backRoute();
+      expect(backRoute).toEqual(compBackRoute);
+    });
+
+    it('the formgroup should have one condition', () => {
+      const conditions = component.ruleForm.get('conditions').value.length;
+      expect(conditions).toBe(1);
+    });
+
+    it('should create a condition', () => {
```
msorens:
```suggestion
    it('upon adding a condition the formgroup includes two conditions', () => {
```

components/automate-ui/src/app/pages/project/rules/project-rules.component.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+      expect(attributeLabel).toBe('node attribute');
+    });
+
+    it('should have attribute label with EVENT type', () => {
+      component.ruleForm.get('type').setValue('EVENT');
+      const attributeLabel = component.getAttributeLabel();
+      expect(attributeLabel).toBe('event attribute');
+    });
+
+    it('should not show add label with one condition', () => {
```
msorens:
```suggestion
    it('should not show "and" label with one condition', () => {
```

components/automate-ui/src/app/pages/project/rules/project-rules.component.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+      const showAndLabel = component.showAndLabel(0);
+      expect(showAndLabel).toBeFalsy();
+    });
+
+    it('should not show delete button with one condition', () => {
+      const showDelete = component.showDelete(0);
+      expect(showDelete).toBeFalsy();
+    });
+
+    it('should show add label with two conditions', () => {
```
msorens:
```suggestion
    it('should show "and" label with two conditions', () => {
```

components/automate-ui/src/app/pages/project/rules/project-rules.component.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+      const conditions = component.ruleForm.get('conditions').value.length;
+      expect(conditions).toBe(2);
+    });
+
+    it('should delete a condition', () => {
+      let conditions = component.ruleForm.get('conditions').value.length;
+      expect(conditions).toBe(1);
+      component.deleteCondition(0);
+      conditions = component.ruleForm.get('conditions').value.length;
+      expect(conditions).toBe(0);
```
msorens: Technically this test is valid, but the user can never delete a condition when there is only one (as you have implemented). So two things to consider:
1. Add a couple more conditions then delete one, so it reflects what a user could actually do.
2. Disallow `deleteCondition()` from deleting the last condition. But then there's the issue of should we display an error in that case, and if so where... so maybe ignore this second point. 🤷‍♂

components/automate-ui/src/app/pages/project/rules/project-rules.component.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+      expect(showDelete).toBeFalsy();
+    });
+
+    it('should show add label with two conditions', () => {
+      component.addCondition();
+      const showAndLabel = component.showAndLabel(0);
+      expect(showAndLabel).toBeTruthy();
+    });
+
+    it('should show delete button with two conditions', () => {
```
msorens:
```suggestion
    it('should show delete button on first condition with two conditions', () => {
```

Then should add one more test:
```
should show delete button on second condition with two conditions
```
(This correlates to @susanev 's comment about the delete on the last row needs to appear.)

msorens:
Got to pass buildkite so can merge this as is. All remaining change requests will start afresh in a new PR, coming soon!

PR #738, opened 2019-07-02
misc rules API fixes

bcmdarroch:

Saw a few issues I wanted to clean up.

First commit addresses an issue I discovered when certain tests passed only when I set the rule type to Node. 

Second fixes some misleading behavior I noticed when a project didn't exist:
```
[18][default:/src:0]# curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects/foo-project/rules | jq

{
  "rules": []
}
```

### :+1: Definition of Done

- [x]  ListRulesForProject returns NotFound if the project doesn't exist (rather than an empty list that might make the customer think the project does exist but it's empty)
- [x]  Attribute validation (i.e. ensuring Event rules don't include incorrect attributes) is now enforced on the rule level, so there's no need to include the rule type on the condition
- [x] when marking an applied rule for deletion, we no longer try to copy the applied rule's `db_id` into the staged rule table - this caused a flaky test when the applied rule's `db_id` conflicted with a different staged rule's `db_id`

- [ ] Code actually executed?


components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1195,6 +1195,14 @@ func (p *pg) ListRulesForProject(ctx context.Context, projectID string) ([]*v2.R
 		return nil, err
 	}
 
+	// verify project exists, otherwise we just return an empty list
+	// that could be misleading
+	var project v2.Project
+	row := p.db.QueryRowContext(ctx, `SELECT query_project($1, $2)`, projectID, pq.Array(projectsFilter))
```
srenatus: [nit] I know we use backticks everywhere, but we really don't need them 😄 They're convenient when you want `"` in your string, but we don't here 😉 
```suggestion
	row := p.db.QueryRowContext(ctx, "SELECT query_project($1, $2)", projectID, pq.Array(projectsFilter))
```

components/authz-service/storage/postgres/migration/sql/52_query_returns_condition_type.up.sql
```diff
@@ -0,0 +1,194 @@
...
+    'type', r.type,
+    'deleted', r.deleted,
+    'status', 'staged',
+    'conditions', json_agg(
+      json_build_object(
+        'value', rc.value,
+        'operator', rc.operator,
+        'attribute', rc.attribute,
+        -- condition type always matches the rule type
+        'type', r.type
```
srenatus: ❓ We could also change this in our Go code, couldn't we? If the conditions' types always match the rule type, it's weird that they've got such a field at all. What am I missing? 😃

components/authz-service/storage/postgres/migration/sql/52_query_returns_condition_type.up.sql
```diff
@@ -0,0 +1,194 @@
...
+  WHERE projects_match_for_rule(project_id, _project_filter)
+  GROUP BY r.id, r.project_id, r.name, r.type;
+
+$$ LANGUAGE sql;
+
+CREATE OR REPLACE FUNCTION
+  query_staged_or_applied_rule(_rule_id TEXT, _project_filter TEXT[])
+  RETURNS json AS $$
+
+  -- check for applied rule
```
msorens:
```suggestion
  -- check for staged rule
```
(You did not introduce this bug, but you've got the power to fix it anyway. 😁  )

components/authz-service/storage/postgres/migration/sql/52_query_returns_condition_type.up.sql
bcmdarroch: We use the condition type to enforce what attributes can be set: https://github.com/chef/automate/blob/master/components/authz-service/storage/v2/condition.go#L63

I agree though, if there's another way to enforce that, I'd prefer not to have a type on conditions

srenatus: ...but isn't that only for _creating new conditions_? I'm still not sure I see the point of _returning it_ from the database query `1+n` times, once in the rule object, and once in every single one of its conditions?

components/authz-service/storage/postgres/migration/sql/52_query_returns_condition_type.up.sql
bcmdarroch: decided to scrap this migration after all :o

components/authz-service/storage/postgres/migration/sql/52_query_returns_condition_type.up.sql
bcmdarroch: @srenatus what do you think about https://github.com/chef/automate/pull/738/commits/a9d02449e127977d3aad3d2a7b06290649599ff0?

components/authz-service/storage/v2/rule.go
```diff
@@ -72,8 +72,10 @@ func validateRuleInputs(id string,
 	}
 
 	for _, condition := range conditions {
-		if condition.Type != ruleType {
-			return errors.New("all condition types must match the parent rule type")
+		if ruleType == Event {
+			if condition.Attribute != Organization && condition.Attribute != ChefServer {
+				return errors.New("rules of type Event only accept Conditions with attributes Organization or ChefServer")
+			}
```
msorens: While this gets the job done, it is what I think of as "magic expressions" (akin to magic strings). Would be a bit cleaner to do something like this:
```
if err := conditionAllowed(ruleType, condition.Attribute); err != nil {
	return err
}
```
and `conditionAllowed` could be a tad more explicit with something like:
```
if ruleType == Event {
  if condition.Attribute != Organization && condition.Attribute != ChefServer {
	return errors.New("rules of type Event only accept Conditions with attributes Organization or ChefServer")
  }
} else if ruleType == Node {
  return nil // all values allowed for nodes
}
```

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1195,6 +1195,14 @@ func (p *pg) ListRulesForProject(ctx context.Context, projectID string) ([]*v2.R
 		return nil, err
 	}
 
+	// verify project exists, otherwise we just return an empty list
```
msorens: As soon as you have more than one SQL statement should wrap in a transaction. Granted these are both read operations, but it has a slim chance for unexpected results.

components/authz-service/storage/postgres/migration/sql/52_query_returns_condition_type.up.sql
srenatus: :100: much less weird. Thanks for doing the no-pain-no-gain kind of work there 👍

PR #739, opened 2019-07-02
Remove transactions from task workers

stevendanna:
Most of the design reviewers were concerned with the prospect of
long-live transactions. This PR removes long-lived transactions and
replaces it with a periodic `update` from the worker thread along
with a periodic garbage collection of expired tasks.

Now, when a task worker is lost, the workflow will receive a workflow
event indicated that the task was lost whereas previously the task
would potentially get picked up by another worker.

This is moving us from an "at least once" model to an "at most once"
model for a single enqueued task.  If a workflow wants to retry the 
lost task, it can re-enqueue the task.

We've also started moving our SQL functions from plain SQL to
plpgsql.

Signed-off-by: Jay Mundrawala <jmundrawala@chef.io>
Co-authored-by: Jay Mundrawala <jmundrawala@chef.io>

### DIFF
--- a/lib/cereal/postgres/postgres.go
+++ b/lib/cereal/postgres/postgres.go
@@ -213,7 +213,13 @@ func (pg *PostgresBackend) expireDeadWorkers() error {
 	if err != nil {
 		return err
 	}
-	defer rows.Close()
+
+	defer func() {
+		if err := rows.Close(); err != nil {
+			logrus.WithError(err).Error("failed to close db rows")
+		}
+	}()
+
 	for rows.Next() {
 		var workerID string
 		err := rows.Scan(&workerID)
@@ -661,6 +667,7 @@ func (pg *PostgresBackend) DequeueTask(ctx context.Context, workerID uuid4.UUID,
 
 	tid, task, err := pg.dequeueTask(ctx, workerID, taskName)
 	if err != nil {
+		cancel()
 		return nil, nil, err
 	}
 	taskc := &PostgresTaskCompleter{
--- a/lib/cereal/postgres/postgres.go
+++ b/lib/cereal/postgres/postgres.go
@@ -5,8 +5,6 @@ import (
 	"database/sql"
 	"time"
 
-	"github.com/chef/automate/lib/uuid4"
-
 	"github.com/golang-migrate/migrate"
 	"github.com/golang-migrate/migrate/database/postgres"
 	"github.com/golang-migrate/migrate/source"
@@ -19,9 +17,9 @@ import (
 )
 
 const (
-	enqueueTaskQuery   = `SELECT cereal_enqueue_task($1, $2, $3, $4, $5)`
+	enqueueTaskQuery   = `SELECT cereal_enqueue_task($1, $2, $3, $4)`
 	dequeueTaskQuery   = `SELECT * FROM cereal_dequeue_task($1)`
-	completeTaskQuery  = `SELECT cereal_complete_task($1, $2, $3, $4, $5)`
+	completeTaskQuery  = `SELECT cereal_complete_task($1, $2, $3, $4)`
 	getTaskResultQuery = `SELECT task_name, parameters, status, error, result FROM cereal_task_results WHERE id = $1`
 
 	enqueueWorkflowQuery  = `SELECT cereal_enqueue_workflow($1, $2, $3)`
@@ -96,8 +94,7 @@ type PostgresBackend struct {
 type PostgresTaskCompleter struct {
 	// tid is the Task's id in our postgresql database.  We need
 	// this to complete the correct task.
-	tid            int64
-	writebackToken string
+	tid int64
 
 	db *sql.DB
 
@@ -615,12 +612,8 @@ func (workc *PostgresWorkflowCompleter) EnqueueTask(task *backend.Task, opts bac
 		opts.StartAfter = time.Now()
 	}
 
-	writebackToken, err := uuid4.NewV4()
-	if err != nil {
-		return err
-	}
-	_, err = workc.enqueueTaskStmt.ExecContext(workc.ctx,
-		workc.wid, writebackToken.String(), opts.StartAfter, task.Name, task.Parameters)
+	_, err := workc.enqueueTaskStmt.ExecContext(workc.ctx,
+		workc.wid, opts.StartAfter, task.Name, task.Parameters)
 	if err != nil {
 		return errors.Wrap(err, "failed to enqueue task")
 	}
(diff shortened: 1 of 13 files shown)


lib/cereal/postgres/sql_schema.go
```diff
@@ -255,39 +264,84 @@ $$;
...
+
+CREATE OR REPLACE FUNCTION cereal_expire_dead_workers()
+RETURNS SETOF TEXT
+AS $$
+DECLARE
+    t cereal_tasks%rowtype;
+BEGIN
+    FOR t IN
+        SELECT cereal_tasks.*
+        FROM cereal_tasks LEFT JOIN cereal_busy_task_workers ON cereal_tasks.worker_id = cereal_busy_task_workers.id
```
jaym: skip locked?

lib/cereal/postgres/sql_schema.go
```diff
@@ -102,17 +102,26 @@ CREATE TABLE cereal_workflow_results (
...
     updated_at    TIMESTAMPTZ NOT NULL DEFAULT NOW(),
     start_after   TIMESTAMPTZ NOT NULL DEFAULT NOW(),
     task_name     TEXT NOT NULL,
-    parameters    BYTEA
+    parameters    BYTEA,
+    task_state    cereal_task_state NOT NULL DEFAULT 'queued',
+    worker_id     TEXT
+);
+
+CREATE TABLE cereal_busy_task_workers (
```
jaym: can a worker forget about a task and then still have a claim to it because of this table?

lib/cereal/postgres/postgres.go
```diff
@@ -626,87 +658,82 @@ func (pg *PostgresBackend) dequeueTask(tx *sql.Tx, taskName string) (int64, *bac
...
+func (taskc *PostgresTaskCompleter) workerPing() {
+	for {
+		select {
+		case <-taskc.ctx.Done():
+			return
+		case <-time.After(30 * time.Second):
+			logrus.Debugf("checkin for worker %s", taskc.workerID.String())
+			_, err := taskc.db.ExecContext(taskc.ctx, "UPDATE cereal_busy_task_workers SET last_checkin = NOW() WHERE id = $1", taskc.workerID.String())
+			if err != nil {
+				logrus.WithError(err).Error("failed to update worker check-in time")
```
ryancragun: The error might have the detail, but if not we'll probably want to add the workerID to the log output

lib/cereal/postgres/worker_pinger.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+type workerPinger struct {
+	workerID        uuid4.UUID
+	db              *sql.DB
+	wgStart         sync.WaitGroup
+	wgStop          sync.WaitGroup
+	stop            context.CancelFunc
+	once            sync.Once
```
stevendanna: Is this used?

lib/cereal/postgres/postgres_test.go
```diff
@@ -179,10 +180,177 @@ func TestTaskCompleteWhileWorkflowIsRunning(t *testing.T) {
...
+
+	// start work on task
+	_, taskCompleter, err := b1.DequeueTask(ctx, taskName)
+	require.NoError(t, err)
+
+	// task finishes, expect write back
+	err = taskCompleter.Succeed(nil)
+	require.NoError(t, err)
+
+	// Make sure we get a message indicating failure because of lost work
```
jaym: this comment is wrong

lib/cereal/postgres/sql_schema.go
```diff
@@ -255,39 +258,101 @@ $$;
...
-    WHERE t1.id = (
-        SELECT t2.id FROM cereal_tasks t2
-        WHERE t2.task_name = _task_name AND t2.try_remaining > 0 AND start_after < NOW()
-        ORDER BY t2.enqueued_at FOR UPDATE SKIP LOCKED LIMIT 1
-    ) RETURNING t1.id, t1.workflow_instance_id, t1.parameters
-$$ LANGUAGE SQL;
+DECLARE
+    r cereal_tasks%rowtype;
+BEGIN
+    FOR r IN
```
stevendanna: Now that we are back to updating a single table, I'm not sure this conversion to a `FOR` loop make sense anymore.

lib/cereal/postgres/task_cleaner.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func (w *taskCleaner) expireDeadTasks(ctx context.Context, expireOlderThanSeconds int64) error {
+	rows, err := w.db.QueryContext(ctx,
+		"SELECT * FROM cereal_expire_tasks($1)", expireOlderThanSeconds)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
```
stevendanna: Once it is merged we could use our `fileutils.LogClose` here:

```golang
fileutils.LogClose(rows, logrus.StandardLogger(), "failed to close db rows")
```

lib/cereal/postgres/task_pinger.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+		w.wgStop.Done()
+	}()
+}
+
+func (w *taskPinger) ping(ctx context.Context) (shouldExit bool, err error) {
+	logctx := logrus.WithFields(logrus.Fields{
+		"taskID": w.taskID,
+	})
+	logctx.Debug("checkin for worker")
```
stevendanna:
```suggestion
	logctx.Debug("checkin for task")
```

lib/cereal/postgres/task_pinger.go
```diff
@@ -0,0 +1,83 @@
...
+	_, err = w.db.ExecContext(ctx, "SELECT cereal_ping_task($1)", w.taskID)
+	if err != nil {
+		logctx.WithError(err).Error("failed to update worker check-in time")
```
stevendanna:
```suggestion
		logctx.WithError(err).Error("failed to update task check-in time")
```

lib/cereal/postgres/task_cleaner.go
```diff
@@ -0,0 +1,95 @@
...
+		taskTimeout:   300 * time.Second,
+	}
+	cleaner.wgStart.Add(1)
+	return cleaner
+}
+
+func (w *taskCleaner) Start(ctx context.Context) {
+	// Make sure this function is only called once. A second call will
+	// cause this to panic
+	w.wgStart.Done()
```
stevendanna: 💭 Turning this into a panic is a nice defense against what we see as a programming mistake, provided we have sufficient test coverage.  I suppose an alternative would be to use sync.Once to make subsequence calls to Start() meaningless.

lib/cereal/postgres/sql_schema.go
```diff
@@ -196,6 +198,7 @@ AS $$
         VALUES('cancel', _workflow_instance_id);
 $$ LANGUAGE SQL;
 
+-- TODO(jaym): This function looks wrong now
```
stevendanna: How so?  I'm thinking maybe it was wrong in a previous iteration but should be correct again now that we've reduced this down to just pining updated_at?

jaym: i've basically forgotten why we had abandon. I feel like it should just be complete

jaym: the way unclaimed tasks are selected also is wrong now

stevendanna: We have abandoned in case you complete a task when there are still outstanding tasks.  I agree this feels like it could just be part of complete.

PR #740, opened 2019-07-02
Don't write out credentials for non automate deploys

jaym:
For chef-server only, we don't need the credentials to be written as they wont be used anywhere


components/automate-deployment/pkg/client/deployer.go
```diff
@@ -1544,6 +1545,10 @@ func (d *deployer) saveDeploymentCreds() {
 		return
 	}
 
+	if !deployment.ContainsAutomateCollection(d.mergedCfg.GetDeployment()) {
```
ryancragun: Might be nice to leave a comment about why we're done early

PR #741, opened 2019-07-02
Cleanup auth collection in products.meta

jaym:
The auth collection already had authn-service, so automate
did not need to repeat it. Also, teams-service should belong
under auth.


jaym:
hmm. it should have complained

PR #742, opened 2019-07-02
[CI] Start shellchecking some scripts

stevendanna:
We have a LOT of bash and shell in this repository. shellcheck can
help catch errors and other shell gotcha's. This PR gets us started by
applying it to one small set of files.

### DIFF
--- a/.expeditor/update_habitat.sh
+++ b/.expeditor/update_habitat.sh
@@ -18,8 +18,8 @@ HAB_VERSION=$HAB_SUP_VERSION
 HAB_RELEASE=$(curl "https://bldr.habitat.sh/v1/depot/channels/core/stable/pkgs/hab/$HAB_VERSION/latest" | jq -r .ident.release)
 
 HAB_LAUNCH_IDENT_JSON=$(curl "https://bldr.habitat.sh/v1/depot/channels/core/stable/pkgs/hab-launcher/latest" | jq -r .ident)
-HAB_LAUNCH_VERSION=$(echo $HAB_LAUNCH_IDENT_JSON | jq -r .version)
-HAB_LAUNCH_RELEASE=$(echo $HAB_LAUNCH_IDENT_JSON | jq -r .release)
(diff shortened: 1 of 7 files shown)


.expeditor/update_habitat.sh
```diff
@@ -18,8 +18,8 @@ HAB_VERSION=$HAB_SUP_VERSION
 HAB_RELEASE=$(curl "https://bldr.habitat.sh/v1/depot/channels/core/stable/pkgs/hab/$HAB_VERSION/latest" | jq -r .ident.release)
 
 HAB_LAUNCH_IDENT_JSON=$(curl "https://bldr.habitat.sh/v1/depot/channels/core/stable/pkgs/hab-launcher/latest" | jq -r .ident)
-HAB_LAUNCH_VERSION=$(echo $HAB_LAUNCH_IDENT_JSON | jq -r .version)
-HAB_LAUNCH_RELEASE=$(echo $HAB_LAUNCH_IDENT_JSON | jq -r .release)
+HAB_LAUNCH_VERSION=$(echo "$HAB_LAUNCH_IDENT_JSON" | jq -r .version)
```
srenatus: [bikeshed] I prefer args, like,
```suggestion
HAB_LAUNCH_VERSION=$(jq -nr --argjson ident "$HAB_LAUNCH_IDENT_JSON" '$ident.version')
```

But it might be overkill and at any rate, out-of-scope for this PR.

scripts/repo_health
```diff
@@ -2,11 +2,11 @@
 
 set -e
 
-echo "Install latest dep"
+echo "Install latest dep and other tools"
```
srenatus: :+1:

.expeditor/update_habitat.sh
stevendanna: I think this too (having just learned of it this morning), but I'm gonna leave it out here since these are all more or less mechanical corrections.

PR #743, opened 2019-07-02
fix saml in dev and acceptance for inspec and cypress

srenatus:
Cleaning up the fall-out of #726. When testing that other PR, 

- I had assumed that the inspec stuff already knew how to deal with single-login method situations.
- Also, I hadn't considered that cypress tests actually attempt to _logout_, spot-checking the specs, I hadn't triggered one that had attempted that.

Well, now, we're good. 😄

### DIFF
--- a/.expeditor/buildkite/cypress.sh
+++ b/.expeditor/buildkite/cypress.sh
@@ -4,8 +4,8 @@ set -euo pipefail
 
 cd /workdir/e2e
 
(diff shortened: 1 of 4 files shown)


srenatus:
> Just one question--what does "bikeshed jq" mean?

I was alluding to _bikeshedding_ -- see https://en.wiktionary.org/wiki/bikeshedding -- to convey that what was there before, was fine, too, I just preferred my bikeshed in a different color. 🤷‍♂ 😉

PR #748, opened 2019-07-02
[A2-903] add db_id on policies, policy_members, policy_projects

blakestier:
### :+1: Definition of Done

- [ ] include the changes to `iam_statements` in there as well
- [ ]  rename the tables (`iam_policies_projects`)
- [ ] clean up the assertions in the test file (had to do too many replacement IMO, but then again, it’s done)

- [ ] Code actually executed?


components/authz-service/storage/postgres/migration/sql/52_add_db_id_members.up.sql
```diff
@@ -0,0 +1,156 @@
...
+    AS $$
+    SELECT
+        db_id
+    FROM
+        iam_members
+    WHERE
+        id = _id;
+$$
+LANGUAGE SQL;
+CREATE OR REPLACE FUNCTION projects_match (_projects TEXT[], _projects_filter TEXT[])
```
srenatus: ❓ why did this function have to be changed? I don't see any IDs in it... 🤔

components/authz-service/storage/postgres/migration/sql/52_add_db_id_members.up.sql
```diff
@@ -0,0 +1,156 @@
...
+                    SELECT
+                        stmt.id,
+                        stmt.effect,
+                        stmt.actions,
+                        stmt.resources,
+                        stmt.role,
+                        -- get each statement's projects by cross-referencing iam_policy_statements and iam_statement_projects
+                        (
+                            SELECT
+                                COALESCE(Json_agg(proj.id) FILTER (WHERE proj.id IS NOT NULL),
```
srenatus:
```suggestion
                                COALESCE(json_agg(proj.id) FILTER (WHERE proj.id IS NOT NULL),
```

components/authz-service/storage/postgres/migration/sql/52_add_db_id_members.up.sql
```diff
@@ -0,0 +1,156 @@
...
+            (
+            SELECT
+                array_agg(proj.id) FILTER (WHERE proj.id IS NOT NULL)
+            FROM iam_policy_projects AS pol_projs
+            LEFT OUTER JOIN iam_projects AS proj ON pol_projs.project_id = proj.id WHERE pol_projs.policy_id = pol.id) AS projects FROM iam_policies AS pol WHERE pol.id = _policy_id
+        GROUP BY
+            pol.id
+)
+SELECT
+    json_build_object('id', temp.id, 'name', temp.name, 'type', temp.type, 'statements', COALESCE(temp.statements, '{}'), 'members', COALESCE(temp.members, '{}'), 'projects', COALESCE(temp.projects, '{}')) AS POLICY
```
srenatus: This is what pg_formatter gives us? Ugh 😉

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -117,11 +117,14 @@ func (p *pg) PurgeSubjectFromPolicies(ctx context.Context, sub string) ([]string
...
-SELECT array_agg(policy_id) FROM pol_ids`,
-		sub)
+		WITH pol_db_ids AS (
+			DELETE FROM iam_policy_members
+			WHERE member_id=(SELECT db_id FROM iam_members WHERE name=$1)
+			RETURNING policy_id
+		)
+		SELECT array_agg(id)
+		FROM iam_policies
+		WHERE db_id IN (SELECT * FROM pol_db_ids);`, sub)
```
srenatus: Hmm I guess we could also do this

```sql
SELECT array_agg(id)
FROM iam_policies
WHERE db_id IN (
  DELETE FROM iam_policy_members
  WHERE member_id=(
    SELECT db_id
    FROM iam_members
    WHERE name=$1)
  RETURNING policy_id)
```
Hmmm I think I like yours better.

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -356,13 +359,13 @@ func (p *pg) associatePolicyWithProjects(ctx context.Context,
 	// TODO this might be simplified as we modify how projects are assigned
 	// Drop any existing associations.
 	_, err := q.ExecContext(ctx,
-		`DELETE FROM iam_policy_projects WHERE policy_id=$1;`, policyID)
+		`DELETE FROM iam_policy_projects WHERE policy_id=policy_db_id($1);`, policyID)
```
srenatus: hrm that backtick and semicolon habit 
```suggestion
		"DELETE FROM iam_policy_projects WHERE policy_id=policy_db_id($1)", policyID)
```
but this really doesn't matter 😉

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -6633,8 +6623,8 @@ func TestPurgeSubjectFromPolicies(t *testing.T) {
 			require.NoError(t, err)
 			assert.Equal(t, []string{polID}, ids)
 
-			assertEmpty(t, db.QueryRow(`SELECT count(*) FROM iam_policy_members WHERE member_id=$1`, member0.ID))
-			assertOne(t, db.QueryRow(`SELECT count(*) FROM iam_policy_members WHERE member_id=$1`, member1.ID))
+			assertEmpty(t, db.QueryRow(`SELECT count(*) FROM iam_policy_members WHERE member_id=member_db_id($1)`, member0.ID))
+			assertOne(t, db.QueryRow(`SELECT count(*) FROM iam_policy_members WHERE member_id=member_db_id($1)`, member1.ID))
```
srenatus: I wonder if we could use some more helpers, like
```go
func count(from, query string, args ...interface{}) *db.Row {
  return db.QueryRow(fmt.Sprintf("SELECT count(*) FROM %s WHERE %s", from, query), args...)
}
```
so we could turn these into
```go
assertOne(t, count("iam_policy_members", "member_id=member_db_id($1)", member1.ID)
```
🤔 (Generally, sprintf-ing SQL queries is a bad idea, but for tests, fine 🤷‍♂) -- ⚠️ I haven't tested this, the var args handling could be wrong, see https://golang.org/ref/spec#Passing_arguments_to_..._parameters for all the details.

blakestier: yeah fixing up the assertions is also on my list https://chefio.slack.com/archives/C61F9HHKK/p1562111196290600

I'll just put all this in the PR description instead of just hoping ppl stumble on it in slack lol

components/authz-service/storage/postgres/migration/sql/52_add_db_id_members.up.sql
blakestier: 🤔

components/authz-service/storage/v2/postgres/postgres.go
blakestier: it's a toss up, but I like that the `DELETE` leads the first one (which seems like the right one for a method with the word `Purge` in it)

components/authz-service/storage/v2/postgres/postgres.go
blakestier: it totally does-- on my list for later if time allows

blakestier: also one of them got mad when I tried it because a string was involved . . . hmmm can't recall

components/authz-service/storage/postgres/migration/sql/53_add_db_id_to_iam_statements.up.sql
```diff
@@ -0,0 +1,187 @@
...
+ALTER TABLE iam_policy_members
+    DROP COLUMN policy_temp_id;
+ALTER TABLE iam_policy_members
+    ADD UNIQUE (member_id, policy_id);
+
+ALTER TABLE iam_policy_projects RENAME COLUMN policy_id TO policy_temp_id;
+ALTER TABLE iam_policy_projects
+    ADD COLUMN policy_id INTEGER REFERENCES iam_policies (db_id) ON DELETE CASCADE;
+UPDATE
+    iam_policy_projects ipp
```
bcmdarroch: side-note: it'd be great if we could add a db diagram to one of our readmes once this db_id cleanup is done. [this site](https://dbdiagram.io/home) looked promising - i think we might be able to import our postgresql

blakestier: ooooo

blakestier:
I'm going to merge this as it's green and push the feedback into the next batch of db work

PR #749, opened 2019-07-02
[A2-890] Fix race condition in db with deployments

kmacgugan:
We allowed non-unique deployments to be inserted, since we ingest in batches this
could end up with duplicates. We applied a uniqueness constraint to prevent duplicates.
Same batches of new service-groups or new deployments are then retried when the unique constraint is hit.

Hard to repro outside of integration tests, but theres a test! 

- [ ] Code actually executed?

### DIFF
--- a/components/applications-service/integration_test/ingestion_test.go
+++ b/components/applications-service/integration_test/ingestion_test.go
@@ -189,19 +189,86 @@ func TestIngestMultiServicesSameServiceGroup(t *testing.T) {
 
 	// also check stats endpoint function
 	assert.Equal(t, int32(len(events)), suite.GetServicesCountForStatsEndpoint(), "wrong number of services")
-	// We have a race condition where if we have different services sending messages
-	// to Automate at the same time that belongs to the same service-group we will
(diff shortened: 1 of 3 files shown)


components/applications-service/integration_test/ingestion_test.go
```diff
@@ -174,19 +174,86 @@ func TestIngestMultiServicesSameServiceGroup(t *testing.T) {
...
-	//sgList := suite.GetServiceGroups()
-	//assert.Equal(t, 1, len(sgList), "there should be only one single service group")
+	sgList := suite.GetServiceGroups()
+	assert.Equal(t, 1, len(sgList), "there should be only one single service group")
+	// also check stats endpoint function
+	assert.Equal(t, int32(1), suite.GetServiceGroupsCountForStatsEndpoint(), "there should be only one single service group")
+}
+
+// This tests our ability to retry inserting service groups when there is an existing
+// deployment but two new service groups in the same batch of messages.
```
danielsdeleo: 🥇 thanks for the explanation!

Issue #747, opened 2019-07-02
Client Runs page's node export infinite loop


lancewf:
# User Story
As an Automate user, I want to be able to export client runs nodes. 

# Definition of Done
Be able to export client runs nodes when there are duplicate UUIDs.

# Demo Script / Repro Steps
1. Create 110 nodes where there are a few with the same UUIDs.
1. Open the Client Runs page and export the nodes into CSV. 
1. Ensure there are no duplicate nodes.

PR #750, opened 2019-07-03
Client Runs page's node export infinite loop

lancewf:
Preventing the Client Runs page's node export from going into an infinite loop. To cause the infinite loop when using the node export, a user needs to have nodes with the same name that has uppercase characters. This bug is caused by using the Elasticsearch SearchAfter API with non-matching curser values. When requesting the second page of values the last value from the first page's node name is used in the request. For the node-state index, the node name values are matched as lower-case, but the code was sending the values with uppercase. When this happens, SearchAfter API will return the first page again. 

I have not seen that this infinite loop timeout in any way. Therefore a context deadline of one minute was added to prevent this from happening again. 

1. `build components/config-mgmt-service && start_all_services`
1. Create 101 nodes with the same name but different UUIDs `for i in {1..101}; do send_chef_run_example; done`
1. Open https://a2-dev.test/client-runs and export the nodes

https://github.com/chef/automate/issues/747

- [x] Code actually executed?

### DIFF
--- a/components/config-mgmt-service/grpcserver/node_export.go
+++ b/components/config-mgmt-service/grpcserver/node_export.go
@@ -71,8 +71,14 @@ func (s *CfgMgmtServer) NodeExport(request *pRequest.NodeExport, stream service.
 	if err != nil {
 		return err
 	}
-
-	return s.exportNodes(stream.Context(), request, exporter)
+	streamCtx := stream.Context()
+	deadline, ok := streamCtx.Deadline()
+	if !ok {
+		deadline = time.Now().Add(time.Minute)
+	}
+	ctx, cancel := context.WithDeadline(streamCtx, deadline)
+	defer cancel()
+	return s.exportNodes(ctx, request, exporter)
 }
 
 func getExportHandler(outputType string, stream service.CfgMgmt_NodeExportServer) (exportHandler, error) {
--- a/components/config-mgmt-service/backend/elastic/nodes.go
+++ b/components/config-mgmt-service/backend/elastic/nodes.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"reflect"
+	"strings"
 	"time"
 
 	"github.com/olivere/elastic"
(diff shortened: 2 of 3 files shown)


components/config-mgmt-service/grpcserver/node_export.go
```diff
@@ -71,8 +71,14 @@ func (s *CfgMgmtServer) NodeExport(request *pRequest.NodeExport, stream service.
...
-
-	return s.exportNodes(stream.Context(), request, exporter)
+	streamCtx := stream.Context()
+	deadline, ok := streamCtx.Deadline()
+	if !ok {
+		deadline = time.Now().Add(time.Minute)
+	}
+	ctx, cancel := context.WithDeadline(streamCtx, deadline)
+	defer cancel()
+	return s.exportNodes(ctx, request, exporter)
```
lancewf: Adding a one-minute timeout for requests.

components/config-mgmt-service/grpcserver/node_export.go
```diff
@@ -71,8 +71,14 @@ func (s *CfgMgmtServer) NodeExport(request *pRequest.NodeExport, stream service.
...
 		return err
 	}
-
-	return s.exportNodes(stream.Context(), request, exporter)
+	streamCtx := stream.Context()
+	deadline, ok := streamCtx.Deadline()
+	if !ok {
+		deadline = time.Now().Add(time.Minute)
+	}
+	ctx, cancel := context.WithDeadline(streamCtx, deadline)
```
stevendanna: [💭 follow-up] We've used this pattern in a few places, I wonder if something like `WithDefaultTimeout(ctx, defaultTimeout)` might be a useful library function that basically does lines 74-79.

srenatus: Hmm counter-proposal:

Let's use [`context.WithTimeout`](https://golang.org/pkg/context/#WithTimeout), it is a simple method [doing the same thing](https://golang.org/src/context/context.go?s=14162:14239#L440). Also, [`context.WithDeadline`](https://golang.org/src/context/context.go?s=12173:12241#L373) already checks if the parent context has a deadline that is before the one we've passed, and ignores ours if that is the case. So, we can just do this:

```go
ctx := stream.Context()
ctx, cancel := context.WithTimeout(ctx, time.Minute)
defer cancel()
```

...and that doesn't warrant a new helper method. 😉

stevendanna: @srenatus I thought that at first too, but it is slightly different semantics.  Namely, your formulation will always produce a ctx that will time out in 1 minute.  The formulation in the PR will only produce such a timeout if the existing context doesn't already have a timeout.

srenatus: > Namely, your formulation will always produce a ctx that will time out in 1 minute.

Yup, and that only matters if the calling context has a timeout _longer than one minute_. :+1: If we need this functionality, sure, let's do it 😄

lancewf: What I was trying to do here was add a timeout only if there was not one already. This would allow for the gateway, load balancer, or clients to override the timeout.

lancewf:
I am going to merge this now. I will create a new PR If we decide that we need to change the context. 

Thank you all for the reviews.

PR #751, opened 2019-07-03
habitat: pin core/jq-static/1.6

srenatus:
This is to be reverted when we've dealt with core/jq-static/1.10.

I believe that if we get this merged, and then pull the erroneous version 1.10 off bldr (or make it private, whatever), we can revert this: `core/jq-static` will then resolve to 1.6, and everything will build OK.


srenatus:
we don't need this 💣

PR #762, opened 2019-07-03
[scripts] Add more missing shebang lines

stevendanna:
Not having these shebang lines makes running shellcheck less useful
and makes it harder to use the script in all contexts.

Note that scripts/copy_hartifacts is being fixed in #758

While touching these files, I've run shellcheck on them and fixed all
of the violations.

### DIFF
--- a/scripts/verify_build
+++ b/scripts/verify_build
@@ -14,8 +14,9 @@ export HAB_NOCOLORING=true
 export HAB_LICENSE="accept-no-persist"
 
 sudo -E hab pkg install core/ruby
-export PATH=$PATH:$(hab pkg path core/ruby)/bin
-sudo -E $(hab pkg path core/ruby)/bin/gem install toml
+export PATH
+PATH="$PATH:$(hab pkg path core/ruby)/bin"
+sudo -E "$(hab pkg path core/ruby)"/bin/gem install toml
 
 if [[ $EUID -eq 0 ]]; then
     key_cache="/hab/cache/keys"
@@ -35,7 +36,7 @@ curl "https://packages.chef.io/manifests/dev/automate/latest.json" > results/dev
 curl "https://packages.chef.io/manifests/current/automate/latest.json" > results/current.json
 curl "https://packages.chef.io/manifests/acceptance/automate/latest.json" > results/acceptance.json
 
-changed_components=($(./scripts/changed_components))
+mapfile -t changed_components < <(./scripts/changed_components)
 if [[ ${#changed_components[@]} -ne 0 ]]; then
     cat << EOF | buildkite-agent annotate --style "info"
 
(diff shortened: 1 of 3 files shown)


scripts/verify_build
```diff
@@ -35,7 +36,7 @@ curl "https://packages.chef.io/manifests/dev/automate/latest.json" > results/dev
 curl "https://packages.chef.io/manifests/current/automate/latest.json" > results/current.json
 curl "https://packages.chef.io/manifests/acceptance/automate/latest.json" > results/acceptance.json
 
-changed_components=($(./scripts/changed_components))
+mapfile -t changed_components < <(./scripts/changed_components)
```
srenatus: Nice! [TIL](https://wiki-dev.bash-hackers.org/commands/builtin/mapfile)

scripts/verify_build
```diff
@@ -14,8 +14,9 @@ export HAB_NOCOLORING=true
 export HAB_LICENSE="accept-no-persist"
 
 sudo -E hab pkg install core/ruby
-export PATH=$PATH:$(hab pkg path core/ruby)/bin
-sudo -E $(hab pkg path core/ruby)/bin/gem install toml
+export PATH
+PATH="$PATH:$(hab pkg path core/ruby)/bin"
+sudo -E "$(hab pkg path core/ruby)"/bin/gem install toml
```
srenatus: ❓ Would there be a way with `hab pkg exec core/ruby`, too, here?

scripts/verify_build
```diff
@@ -14,8 +14,9 @@ export HAB_NOCOLORING=true
 export HAB_LICENSE="accept-no-persist"
 
 sudo -E hab pkg install core/ruby
-export PATH=$PATH:$(hab pkg path core/ruby)/bin
-sudo -E $(hab pkg path core/ruby)/bin/gem install toml
+export PATH
+PATH="$PATH:$(hab pkg path core/ruby)/bin"
```
phiggins: Do you know why it recommends splitting the export and the assignment? I'm curious and I don't see any explanations of things on the shellcheck homepage.

stevendanna: I do!  It has to do with an interaction with `set -e` the "exit on error" option.  Consider:

```
> cat ./test.sh
#!/bin/bash

set -e

export PATH=$(false)
echo "got here"
> ./test.sh
got here
> cat test2.sh 
#!/bin/bash

set -e

export PATH
PATH=$(false)
echo "got here"
> ./test2.sh
>
```

When you do `export FOO=$(some_command)` or `local FOO=$(some_command)` you end up not seeing the failure some_command.

stevendanna: The shellcheck wiki entry for it is here: https://github.com/koalaman/shellcheck/wiki/SC2155

phiggins: Oh neat, thanks!

PR #768, opened 2019-07-03
Create ingest client v1

afiune:
Add stats of ingested events.

### :+1: Definition of Done
We can see stats from the ingester client.

Build and start the application-service, then ingest a message and see the new stats:
```

[ suppressed output ]

[2][default:/src:0]# applications_grpcurl chef.automate.api.applications.ApplicationsService.GetServicesStats
{
  "ingestStats": {

  }
}
[3][default:/src:0]# applications_publish_supervisor_message
INFO[0000] Connecting to NATS Server                     client_id=applications-publisher cluster_id=event-service mtls_enabled=true nats_url="nats://0.0.0.0:10140"
INFO[0000] Publishing message                            message="event_metadata:<supervisor_id:\"1234567890\" occurred_at:<seconds:1562178623 nanos:289956200 > application:\"demo\" environment:\"demo\" fqdn:\"localhost\" > service_metadata:<package_ident:\"core/redis/0.1.0/20190703183023\" service_group:\"redis.default\" > " type=HealthCheckEvent
[4][default:/src:0]# applications_grpcurl chef.automate.api.applications.ApplicationsService.GetServicesStats
{
  "totalServiceGroups": 1,
  "totalServices": 1,
  "totalSupervisors": 1,
  "ingestStats": {
    "totalEventsProcessed": 1,
    "totalEventsSuccessful": 1
  }
}
```

- [x] Code actually executed?


danielsdeleo:
Who are these stats for? We have this information already in the prometheus endpoint, which can be enabled with `chef-automate dev enable-prometheus` and then you can see everything in the graphs or curl the prometheus format stats endpoint on the applications service. If we are targeting automate developers here then maybe we need to make this functionality more obvious or easier to use?

On the other hand, I think it would be cool for us to track whether users are in general seeing ingest errors but then we need to think about how we slice up time to make the value that gets reported to telemetry useful to us.

afiune:
@danielsdeleo I'm just trying to make the stats endpoint more robust by adding the ingester numbers to the mix. As you mentioned, tracking/knowing when there are failures and how many things have we processed will help us understand the ingestion part of the service. 

I can also see a use case where we want users to report to us the stats to understand their system, in such cases we wouldn't want them to start the prometheus at a firt glance and instead, we could have a quick look by hitting the RPC func. 

If this is not making the stats better, let us talk about it before accepting this change. Thanks for your feedback! 🙂

danielsdeleo:
> I can also see a use case where we want users to report to us the stats to understand their system, in such cases we wouldn't want them to start the prometheus at a firt glance and instead, we could have a quick look by hitting the RPC func.

I think it would be best to have one place for operational stats like this instead of two. The stats endpoint I recently added is focused on telemetry, so we can understand whether people are actually using the system and how they are using it; or stated differently, the numbers are a proxy measurement of how the users of the system behave. The prometheus endpoint gives information about how the software is behaving, and is intended to be useful for operational concerns.

The metrics endpoint that feeds the prometheus server is always enabled, it can be accessed via `curl -k --cert /hab/svc/deployment-service/data/deployment-service.crt --key /hab/svc/deployment-service/data/deployment-service.key https://localhost:20133/metrics` 

I realize this is not easy to remember or all that intuitive, but rather than adding a different place to get operational metrics, I would prefer to focus on making it easier to use the prometheus endpoints we have. Maybe something like `chef-automate dev mtls-curl` or `chef-automate dev scrape-metrics` ?

afiune:
I can see how this could look like we want operations on the mix, which it is not a bad idea, being able to detect a big number of failures very fast and repeatedly but, to be honest, I wasn't thinking about operations, I was thinking to add these numbers to the metrics to understand the system, that is, to know how many messages go through and try to make decisions based on those numbers. 

In the future, we will need to ingest different kinds of messages, and I would also like to see those numbers here, a quick example of this is when we made some architectural decisions about the _health_check messages_ where we assume they will be very repetitive and self-healing so we want them to go through a different way of ingestion. I think with these numbers those decisions will be more clear for us to make.

Thanks for looking into this Dan, since it is an unplanned change, I understand this might impact us so I wouldn't want to push further if you don't think this is, in any way, useful.

components/applications-service/pkg/ingester/client.go
```diff
@@ -0,0 +1,14 @@
+package ingester
+
+type Client interface {
```
srenatus: 💭 I kind of miss `context.Context` here. Is there some reason it doesn't make sense to be passed along / made use of? (I'm -- haha -- lacking context, too, so this might be no concern at all.)

components/applications-service/integration_test/suite_test.go
```diff
@@ -117,6 +118,16 @@ func (s *Suite) DeleteDataFromStorage() {
 
 // IngestService ingests a single HealthCheckEvent message into the database
 func (s *Suite) IngestService(event *habitat.HealthCheckEvent) {
+	eventsProcessed := s.Ingester.EventsProcessed()
```
srenatus: I don't think this is concurrency-safe -- can only one of these ever happen at the same time? If this could be called twice concurrently, we might miss an event.

afiune: You are absolutely right! I can fix that for sure. 💯

components/applications-service/pkg/ingester/client.go
afiune: yup, we gotta fix this as well as in the storage layer! 😭

danielsdeleo:
Chatted with @afiune on this and decided to close. We have these same stats more or less in the prometheus endpoint now and would like to keep all operationally-focused metrics in there. The stats endpoint is meant to be called by the UI for exporting to our telemetry pipeline. Information about whether users are seeing errors in ingestion would be cool to have in telemetry but we would need to do some more work on the time bounding to make this a useful measurement in that context.

PR #775, opened 2019-07-03
Serve combined swagger docs with redoc

phiggins:
An implementation of some docs work for #248 

Combines the swagger files generated from our proto compilation and serves them with redoc inside the existing hugo static site. The in-progress docs are marked with draft status so they are visible locally but not in the "live" docs.


components/automate-chef-io/Makefile
```diff
@@ -1,38 +1,51 @@
...
+			mkdir -p "../../automate-chef-io/$(SWAGGER_DIR)/`dirname $$i`" && \
+			cp $$i "../../automate-chef-io/$(SWAGGER_DIR)/$$i" \
 		; done && \
 	popd
 
-serve: sync
+$(SWAGGER_RESULT_FILE): $(SWAGGER_FILES)
+
+generate_swagger: $(SWAGGER_RESULT_FILE)
+	swagger mixin --quiet -c=$(EXPECTED_CONFLICTS) $(SWAGGER_FILES) > $(SWAGGER_RESULT_FILE)
```
tduffield: This assumes you have swagger installed. As far as I can tell there is no documentation on how to install it currently available.

vjeffrey:
this is really cool, it's looking awesome!!!

components/automate-chef-io/Makefile
phiggins: I was just chatting with @vjeffrey about that, I forgot to update the setup instructions. Readme update incoming!

kagarmoe: Let's add 

```
brew tap go-swagger/go-swagger
brew install go-swagger
```
And a link to https://goswagger.io/install.html#homebrewlinuxbrew for other systems.

components/automate-chef-io/README.md
```diff
@@ -6,25 +6,31 @@ A new static website for Automate is based on [Hugo](https://gohugo.io), the Go-
...
     ```
-    $ brew install hugo
+    brew install hugo
     ```
 
-### Routine Setup
+1. [Install go-swagger](https://goswagger.io/install.html). If you're using homebrew:
 
-_These are the steps you'll need to run every time you launch a development environment._
+    ```
```
kagarmoe: We've installed swagger in the docker container, so we probably don't need these lines.

phiggins: What is the docker container? Is that the preferred way to run hugo? Does this mean we don't recommend installing hugo or go-swagger?

kagarmoe: Ok, let's go with what you have. We will explain the docker approach in another PR.

PR #777, opened 2019-07-03
Add nginx headers for automate-ui per owasp secure headers project

itmustbejj:
Signed-off-by: Josh Hudson <jhudson@chef.io>

Add secure headers to harden automate-ui as per owasp best practices More info at https://www.owasp.org/index.php/OWASP_Secure_Headers_Project#tab=Headers

### :+1: Definition of Done
strict-transport-security, x-content-type-options, x-xss-protection headers are properly set in Automate UI.

Visit Automate UI and inspect any response headers in Chrome.
Screen Shot 2019-07-03 at 4 29 43 PM

- [] Necessary tests added/updated?
- [] Necessary docs added/updated?
- [x] Code actually executed?

### DIFF
--- a/components/automate-load-balancer/habitat/config/nginx.conf
+++ b/components/automate-load-balancer/habitat/config/nginx.conf
@@ -145,6 +145,12 @@ http {
     #
     add_header Cache-Control "no-cache, no-store, must-revalidate";
 
+    # Implement secure headers in accordance w/ owasp best practices
+    # https://www.owasp.org/index.php/OWASP_Secure_Headers_Project#tab=Headers
+    add_header x-xss-protection "1; mode=block" always;
+    add_header Strict-Transport-Security "max-age=63072000; includeSubDomains" always;
+    add_header X-Content-Type-Options "nosniff" always;
+
     {{#if ../cfg.service.maintenance_mode ~}}
     error_page 503 /maintenance.html;
 
--- a/components/automate-load-balancer/habitat/config/nginx.conf
+++ b/components/automate-load-balancer/habitat/config/nginx.conf
@@ -141,7 +141,7 @@ http {
     #
     #     Note:
     # /!\ Any location block introducing its own add_header call MUST
-    #     replicate this line (see /dex/, /session/ below).
+    #     replicate these lines (see /dex/, /session/ below).
     #
     add_header Cache-Control "no-cache, no-store, must-revalidate";
 
@@ -217,6 +217,9 @@ http {
       proxy_set_header Connection ""; # Required to make persistent connections happen
       add_header X-Frame-Options sameorigin; # forbid other <iframe>ing
       add_header Cache-Control "no-cache, no-store, must-revalidate";
+      add_header x-xss-protection "1; mode=block" always;
+      add_header Strict-Transport-Security "max-age=63072000; includeSubDomains" always;
+      add_header X-Content-Type-Options "nosniff" always;
     }
 
     location /session/ {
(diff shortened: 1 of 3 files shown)


stevendanna:
This might allow us to remove this: https://github.com/chef/automate/blob/master/integration/helpers/ssl_filters.jq#L20

itmustbejj:
An alternate implementation would be to add the headers only to `location /` here:
https://github.com/chef/automate/blob/f45c491a307b953509f1c4841703e2ecc0237f22/components/automate-load-balancer/habitat/config/nginx.conf#L232-L237

components/automate-load-balancer/habitat/config/nginx.conf
```diff
@@ -145,6 +145,12 @@ http {
     #
     add_header Cache-Control "no-cache, no-store, must-revalidate";
 
+    # Implement secure headers in accordance w/ owasp best practices
+    # https://www.owasp.org/index.php/OWASP_Secure_Headers_Project#tab=Headers
+    add_header x-xss-protection "1; mode=block" always;
+    add_header Strict-Transport-Security "max-age=63072000; includeSubDomains" always;
+    add_header X-Content-Type-Options "nosniff" always;
```
srenatus: ❓ Do we need to copy these around like the `Cache-Control` headers, following this comment? https://github.com/chef/automate/blob/23fdf23efeaa27355e3072f827b0024fde90c0d7/components/automate-load-balancer/habitat/config/nginx.conf#L142-L145

itmustbejj: As these lines were added to a `server` block, I don't believe this would apply.

srenatus: I'm not claiming that this wasn't true but I suppose by that logic, we should be able to drop all those `add_header Cache-Control "...";` statements below then. 🤔

itmustbejj: How would you prefer to proceed? I'm open to moving these lines above the Cache-Control line on line 146 if that would help, or I could move limit the scope to the location for automate-ui, which also has its own Cache-Control.

stevendanna: The comment here is still valid. Consider:

```
[13][default:/src:0]# curl -kv https://localhost/api/ 2>&1 >/dev/null | grep '^<'
< HTTP/2 404 
< server: nginx
< date: Tue, 09 Jul 2019 10:33:23 GMT
< content-type: text/plain; charset=utf-8
< content-length: 10
< x-content-type-options: nosniff
< x-xss-protection: 1; mode=block
< strict-transport-security: max-age=63072000; includeSubDomains
< x-content-type-options: nosniff
< 
[14][default:/src:0]# curl -kv https://localhost/dex/ 2>&1 >/dev/null | grep '^<'
< HTTP/2 404 
< server: nginx
< date: Tue, 09 Jul 2019 10:33:25 GMT
< content-type: text/plain; charset=utf-8
< content-length: 19
< x-content-type-options: nosniff
< 
```

Thus, if we want these headers universally applied (I think we do?) then we need to copy them into any location block that uses `add_header`.

integration/helpers/ssl_filters.jq
```diff
@@ -17,7 +17,6 @@ map(select(
          (.id != "cert_caIssuers" or .port != "443") and        # Test uses self-signed cert
          (.id != "cert_chain_of_trust" or .port != "443") and   # Test uses self-signed cert
          (.id != "cipherlist_128Bit" or .port != "443") and     # For browser compatibility (untested)
-         (.id != "HSTS" or .port != "443") and                  # TODO
```
srenatus: 🎉

PR #779, opened 2019-07-04
Update missing db tables on service updates

afiune:
We were lacking some data update on service health events,
currently, we only update its health, update strategy and
package identifier, but it is missing to update the supervisor,
deployment and service-group data, specifically the application
name, environment, site, and fqdn.

### :+1: Definition of Done
This commit is updating the ingestion to update all missing tables on
any service update.

- Start the studio.
- Build and start the applications-service.
- Send a dummy service message with `applications_publish_supervisor_message --sup-id 1`
- Check how the service looks:
```
[4][default:/src:130]# applications_grpcurl chef.automate.api.applications.ApplicationsService.GetServices
{
  "services": [
    {
      "supervisorId": "1",
      "release": "core/redis/0.1.0/20190704080203",
      "group": "redis.default",
      "application": "demo",
      "environment": "demo",
      "fqdn": "localhost",
      "previousHealthCheck": "NONE",
      "currentHealthSince": "43 seconds"
    }
  ]
}
```
- Update all the fields that weren't possible to update before:
```
[5][default:/src:0]# applications_publish_supervisor_message  --sup-id 1 --fqdn foo --group test --application updated --environment new --site bar
```
- Verify the service again:
```
[6][default:/src:0]# applications_grpcurl chef.automate.api.applications.ApplicationsService.GetServices
{
  "services": [
    {
      "supervisorId": "1",
      "release": "core/redis/0.1.0/20190704080403",
      "group": "redis.test",
      "application": "updated",
      "environment": "new",
      "fqdn": "foo",
      "site": "bar",
      "previousHealthCheck": "NONE",
      "currentHealthSince": "2 minutes, 35 seconds"
    }
  ]
}
```
https://chefio.atlassian.net/browse/A2-891

- [x] Code actually executed?


components/applications-service/pkg/storage/postgres/hab_event.go
```diff
@@ -134,41 +134,167 @@ func (db *Postgres) IngestHealthCheckEventWithoutMetrics(event *habitat.HealthCh
...
+}
+
+func (db *Postgres) updateTables(
+	svc *service,
+	eventMetadata *habitat.EventMetadata,
+	svcMetadata *habitat.ServiceMetadata,
+	pkgIdent *packageIdent,
+	health string) error {
+
+	//updates := map[string]interface{}{}
```
danielsdeleo: leftover cruft?

components/applications-service/pkg/storage/postgres/hab_event.go
```diff
@@ -134,41 +134,167 @@ func (db *Postgres) IngestHealthCheckEventWithoutMetrics(event *habitat.HealthCh
...
+	return db.triggerDataUpdates(svc, deploy, sup, sg)
+}
+
+func (db *Postgres) triggerDataUpdates(
+	svc *service,
+	deploy *deployment,
+	sup *supervisor,
+	sg *serviceGroup) error {
+
+	// update only if there is something to update
```
danielsdeleo: This comment is a bit confusing. Does it refer to the the checks like `if svc.needUpdate {` we do for each of the update-able things?

components/applications-service/pkg/storage/postgres/hab_event.go
```diff
@@ -134,41 +134,167 @@ func (db *Postgres) IngestHealthCheckEventWithoutMetrics(event *habitat.HealthCh
...
+	if sup.Site != eventMetadata.GetSite() {
+		sup.Site = eventMetadata.GetSite()
+		sup.needUpdate = true
+	}
+
+	// TODO @afiune we could have this column in all tables as well
+	// when a supervisor needs update, update the timestamp of the last event received
+	//if sup.needUpdate {
+	//sup.LastEventOccurredAt = convertOrCreateTimestamp(eventMetadata.GetOccurredAt())
+	//}
```
danielsdeleo: I'm not sure what the value would be here, do we need to leave this TODO? Are we gonna do it?

components/applications-service/integration_test/ingestion_test.go
```diff
@@ -604,3 +598,48 @@ func TestIngestHealthUpdatedAtUpdatesOnlyOnHealthUpdate(t *testing.T) {
...
+		eventsProcessed = suite.Ingester.EventsProcessed()
+		event           = NewHabitatEvent(
+			withSupervisorId("4f1un3"),
+			withPackageIdent("core/db/0.1.0/20200101121212"),
+			withServiceGroup("db.default"),
+		)
+		bytes, err = proto.Marshal(event)
+	)
+
+	if assert.Nil(t, err) {
```
srenatus: If we use `require.Nil`, or even better, `require.NoError`, we don't need the if then wrapping.

components/applications-service/pkg/storage/postgres/hab_event.go
afiune: Updated

components/applications-service/pkg/storage/postgres/hab_event.go
afiune: I will remove it, what I wanted to do is to have the same field we have in the `service` table to know "What was the last Habitat message that modified `XYZ` table"?

You could know things like; oh look, the supervisor (eg. FQDN of the machine) was modified with a Habitat message that occurred at this time.

If we need this we can talk about it in standup and apply it, just for the record, we won't be saving the timestamp of hab messages unless the message modifies a field that belongs to the `service` table.

components/applications-service/integration_test/ingestion_test.go
afiune: Amazing feedback! Thank you, changed already! 💯

components/applications-service/pkg/storage/postgres/hab_event.go
afiune: Oh yeah, thanks for the reminder!! I thought about making the functions a bit smarter and simpler by passing an array of tables to be updated, I was in the middle of it and forgot to finish it! I'll finish it and push my idea.

afiune: here: https://github.com/chef/automate/pull/779/commits/c82b8ba0efba724c140721605397e9ef4abe7e1e check it out!

tenor-250846682

components/applications-service/pkg/storage/postgres/hab_event.go
```diff
@@ -134,41 +134,159 @@ func (db *Postgres) IngestHealthCheckEventWithoutMetrics(event *habitat.HealthCh
...
+}
+
+func (db *Postgres) updateTables(
+	svc *service,
+	eventMetadata *habitat.EventMetadata,
+	svcMetadata *habitat.ServiceMetadata,
+	pkgIdent *packageIdent,
+	health string) error {
+
+	db.updateService(svc, eventMetadata, svcMetadata, pkgIdent, health)
```
danielsdeleo: let's make sure the service always gets some update so the database has a record of when the last message was received for a service

afiune: Fixed! Thanks for the clarification @danielsdeleo - Since this is the only missing thing I am going to merge it after buildkite is green, having said that we are now always updating the service timestamp. 💯

PR #781, opened 2019-07-05
[scripts] Run shellcheck on scripts/*.sh and .buildkite folders

stevendanna:
Further expanding the use of shellcheck.

### DIFF
--- a/.buildkite/hooks/pre-exit
+++ b/.buildkite/hooks/pre-exit
@@ -7,5 +7,6 @@ echo "Attempting to clean up docker containers for buildkitejob=$BUILDKITE_JOB_I
 matching_containers=$(docker ps --filter "label=buildkitejob=$BUILDKITE_JOB_ID" --format "{{.ID}}")
 if [ -n "$matching_containers" ]; then
     echo "Killing docker containers $matching_containers"
+    # shellcheck disable=SC2086
     docker kill $matching_containers
 fi
--- a/scripts/spellcheck.sh
+++ b/scripts/spellcheck.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -o pipefail
+set -eo pipefail
 
 # This runs spell checking on all projects in A2.
 # For the front-end projects it checks these file types: .ts, .tsx, .html, .md.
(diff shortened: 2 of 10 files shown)


.buildkite/hooks/pre-exit
```diff
@@ -7,5 +7,6 @@ echo "Attempting to clean up docker containers for buildkitejob=$BUILDKITE_JOB_I
 matching_containers=$(docker ps --filter "label=buildkitejob=$BUILDKITE_JOB_ID" --format "{{.ID}}")
 if [ -n "$matching_containers" ]; then
     echo "Killing docker containers $matching_containers"
+    # shellcheck disable=SC2086
```
stevendanna: I think we want word-splitting here. To avoid needing it, we'd have to refactor this a bit.

scripts/spellcheck.sh
```diff
@@ -29,13 +29,18 @@ set -o pipefail
...
+fatal() {
+    echo "fatal error: $1" >&2
+    exit 2
+}
 
-command -v cspell >/dev/null 2>&1 || { echo >&2 "Cspell required; install with 'npm install -g cspell'.  Aborting."; exit 2; }
-test -d components && test -d scripts || { echo >&2 "Must start in a2 root directory.  Aborting."; exit 2; }
-test -f cspell.json || { echo >&2 "Oh dear, oh dear: config file (cspell.json) is missing.  Aborting."; exit 2; }
+command -v cspell >/dev/null 2>&1 || fatal "Cspell required; install with 'npm install -g cspell'"
+# shellcheck disable=SC2015
```
stevendanna: I think this is a correct usage of `A && B || C`

scripts/verify_studio_init.sh
```diff
@@ -10,6 +10,7 @@
 #
 SUP_LOG_FILE=/hab/sup/default/sup.log
 
+#shellcheck disable=SC1091
```
stevendanna: 🔜

scripts/spellcheck.sh
```diff
@@ -29,13 +29,18 @@ set -o pipefail
...
+test -d components && test -d scripts || fatal "Must start in a2 root directory."
```
srenatus: Do we need these? I don't suppose other scripts have similar checks...

Issue #782, opened 2019-07-05
ingest-service can produce bulk requests that are reject by automate-es-gateway


stevendanna:
`automate-es-gateway` proxies requests from application services such as ingest-service to Elasticsearch.  `automate-es-gateway` is an nginx instances with a default max client body size of 250m. 

`ingest-service` will currently attempt to batch up to 10,000 chefRun messages together.  On some customer sites, it appears that this can easily produce a bulk request that exceeds the 250m limit in es-gateway:

```
time="2019-07-05T09:05:36+01:00" level=error msg="bulkRunPublisherBundler Failed" bulkReq
uestsCount=2814 error="elastic: Error 413 (Request Entity Too Large)" message=ChefRun publish time=11.599948567s
```

While the client_max_body_size is configurable, simply increasing this is not without costs:

1) ingest-service will still need all of the inflight requests in memory, increasing its heap usage.
2) On some systems, we are unsure of the performance implications of very large (500+mb) bulk requests on elasticsearch.  

Thus, we think one or all of the following might be advisable:

- [ ] Make the batch sizes configurable, providing support with a crude lever to use to help customers with this problem.

- [ ] Make the batch limit not just a count but also based on the size of the final request.

- [ ] Determine whether we think it is OK to push client_max_body_size to some much larger value.

moutons:
on the same system most of the failures occurred with batch sizes between 1k and 2k, and as small as 
```
hab: ingest-service.default(O): time="2019-07-05T03:36:43+01:00" level=error 
msg="bulkRunPublisherBundler Failed" bulkRequestsCount=795 error="elastic: Error 413 
(Request Entity Too Large)" message=ChefRun publish time=2.327240034s
```

lancewf:
Thank you for writing this up with the possible solutions. The good news I see from this is that the bulk bundling feature is being pushed to its limits with the customer. If they are bundling 2,814 requests then their Automate is at the maximum ingestion limit (that is a bundle of 938 CCRs)! 

The first fix of making the batch sizes configurable should not be a problem. We could add the size variable to the ingest-service config. 

The second fix of limiting bulked messages on the size will take some testing to see how much it affects the performance.

stevendanna:
@lancewf Thanks for following up, that is my impression from reading the code as well.  If we started with the batch size configuration, that would give support a lever to help customers with while we investigate the right batching criteria.

lancewf:
@stevendanna Below is a PR to add the batch sizes to the Automate config. 
[...]

PR #783, opened 2019-07-05
Adding max number of bundled msg to the automate config

lancewf:

Allowing the maximum number of bundled CCR and Chef server action messages during ingestion to be adjusted with the Automate config. 

1. `build components/ingest-service && build components/automate-deployment && components/automate-cli && start_all_services`
1. Watch the ingest-service logs `sl | grep "ingest-service.default" &`
1. Create a patch.toml file with the below contents
```
[ingest.v1.sys.service]
  max_number_of_bundled_run_msgs = 555
  max_number_of_bundled_action_msgs = 777

[ingest.v1.sys.log]
level = "debug"
```
1. Update the max number of bundled messages, `/hab/pkgs/lancewf/automate-cli/0.1.0/20190705180559/bin/chef-automate config patch patch.toml`
1. What the logs for messaged like below. 
```
ingest-service.default(O): time="2019-07-05T18:12:22Z" level=debug msg="starting bulkActionPublisherBundler" maxNumberOfBundledActionMsgs=777
ingest-service.default(O): time="2019-07-05T18:12:22Z" level=debug msg="starting bulkRunPublisherBundler" maxNumberOfBundledRunMsgs=555
```

https://github.com/chef/automate/issues/782

- [x] Code actually executed?


api/config/ingest/config_request.go
```diff
@@ -26,6 +26,8 @@ func DefaultConfigRequest() *ConfigRequest {
 	c.V1.Sys.Service.Port = w.Int32(10122)
 	c.V1.Sys.Service.PurgeConvergeHistoryAfterDays = w.Int32(30)
 	c.V1.Sys.Service.PurgeActionsAfterDays = w.Int32(30)
+	c.V1.Sys.Service.MaxNumberOfBundledRunMsgs = w.Int32(10000)
```
stevendanna: 💭 I wonder if we want to lower this default for run messages when where.  At 10,000 messages, we'll hit the es-gateway limit at an expected message size of 25KB.  I think in practice this doesn't matter because most users don't see a request rate that can even get close to bundling that many messages.

components/ingest-service/pipeline/publisher/bulk_publisher.go
```diff
@@ -88,7 +94,7 @@ func bulkActionPublisherBundler(in <-chan message.ChefAction, client backend.Cli
 			// Only publish the collection of bundled messages if
 			// the inbox is empty or there are over 10,000 messages bundled
```
stevendanna: [nit] This comment still has the hard-coded value.

api/config/ingest/config_request.go
lancewf: @stevendanna Do you have a guess at what the average message size is for our users?

stevendanna: I'm not sure how representative this is, but at one customer, we recorded a few thousand API calls and saw the following percentiles in terms of data_collector client body sizes at the frontend lb:

```
     50%      90%      95%      99%    99.5%    99.9%
 71367.0 127934.0 140055.0 213987.0 237235.0 385558.2
```

I believe those are in bytes. So roughly 70k median request size.

(more info on where those numbers came from in the old history https://github.com/chef/a2/pull/4238)

lancewf: Does a default of 2,500 seem correct? That allows an average message size of 100KB.

stevendanna: That seems reasonable to me.

PR #785, opened 2019-07-08
Bump Node.js version 8.9.4 -> 10.15.2

msorens:

Continuing work towards the latest Angular upgrade, it is necessary to move to Node.Js version 10. Though Node.js is up to v12, v10 is the appropriate version for production applications, which should use either "Active" or "Maintenance" phase versions, and the state of the world today is shown here (from https://nodejs.org/en/about/releases/):

image

(So in a couple months it would be appropriate to bump the version further to v12, but not now.)

The latest release of node 10/Active is designated with codename "Dubnium" only, as far as I can tell, because it begins with "D", and is the follow-on to node 8/Active "Carbon" beginning with a "C", and so forth. You can see that the latest Dubnium release is quite recent (https://nodejs.org/en/download/releases/).

For those interested, see [What's New In Node 10](https://www.angularminds.com/blog/article/whats-new-in-node-js-version-10-features-and-improvements.html).

Also included in this PR are two other changes relevant to the impending Angular upgrade:

1. Needed to upgrade codelyzer, which is used for a small subset of linting rules.

2. Some time ago, we had already migrated from `@angular/http` to `@angular/common/http` per the Angular preparation instructions... but the actual package `@angular/http` -- now unused -- was still in the package.json file, and it was impeding the Angular upgrade.

⚠️ **In case I am not the one merging this, it should be accompanied by a notice in multiple Slack channels that if you build the ui locally you need to (a) upgrade your own version of node and (b) `npm install`.**

Update:
At the time of writing, the latest Dubnium version is 10.16.0. But we are limited by the packages that habitat knows about, too, and the latest there is 10.15.2 (thanks to @srenatus and @afiune!). The available habitat packages are listed at  https://bldr.habitat.sh/#/pkgs/core/node10.

### :+1: Definition of Done

UI still works.

1. Upgrade your local version of node.
```
nvm use 10.15.2
```

2. Update your packages corresponding to the node version change (order matters!).
```
cd automate/components/chef-ui-library
npm install
cd components/automate-ui
npm install
```

3. Restart your automate-ui server
```
make serve
```

- [x] Code actually executed?


srenatus:
💭 If we bump, we might as well bump to v12, I guess?

> should use either "Active" or "Maintenance" phase versions

looks like the sorce says
> Production applications should only use Active *LTS* or Maintenance *LTS* releases.

...and these are v8 and v10 respectively, at the moment.

So yeah let's bump this. I suppose we could look into getting the latest v10 into core-plans, then?

srenatus:
Fixed the title. To bump to 10.16.0, we'll need some action on https://github.com/habitat-sh/core-plans.

components/automate-ui/package.json
```diff
@@ -73,7 +72,7 @@
     "ajv": "^6.5.3",
     "angular2-template-loader": "^0.6.2",
     "axe-webdriverjs": "^2.0.1",
-    "codelyzer": "^4.4.4",
+    "codelyzer": "^4.5.0",
```
phiggins: The changes to these deps feels very minor, it might be worth putting them in their own PR to make this one smaller.

srenatus:
@phiggins I'm with you. I'll drop that commit and merge this when green.

srenatus:
> In case I am not the one merging this, it should be accompanied by a notice in multiple Slack channels that if you build the ui locally you need to (a) upgrade your own version of node and (b) npm install.

✔️

tduffield:
In the future, please also coordinate with release engineering when making major version upgrades of software languages. We'll need to make sure our pipelines are ready to receive such an upgrade :) cc/ @sdelano @vjeffrey

msorens:
Thanks for the note, @tduffield. I guess I had thought of this as a library update, like we've done for Go and TypeScript casually, but I can see that this is a bigger thing.

PR #786, opened 2019-07-08
Review and fix time format

afiune:
Current in Automate, we have a variety of ways to represent time. For example, `7/3/2019` on Compliance report page; `7:07 AM Wednesday, July 3` on Event Feed; `Wed, June 3, 2019 at 7:07 pm` on Settings-Node Integrations page. 

We want to start forming a standardized way to represent times by use cases. The first two use case we would like to address is :

- As an operator, when I am troubleshooting an outage in my monitoring tools, I would like to coordinate the event in the monitoring tool to the service health status in Automate by matching the time stamp. Because my monitoring tools use UTC (Coordinated Universal Time) time zone, having the time visualized in the same format in Automate would help me correlate the events with health status more easily.

- As an operator, [same scenario as above]. My monitoring tools use the local time zone, and I would like to know when a health status change relative to my local time so that I can correlate the events with the service health status.

### Our Proposal Solution

Our solution is limited by the fact that currently in Automate, we don’t show users how the time zone is defined or how it can be reconfigured. Therefore, we choose to display two different “time” formats in the UI as the default to cover the use cases mentioned above: one in the UTC format, one in the relative-to-now time format (see the screenshot below). The later is displayed in a tooltip when hoving over or focus on the former time element. In the future, the two fields could be configurable by the users depending on their specific needs.

image

### :+1: Definition of Done

1. Under the health check report, display “Changed from [previous different status] to [current status] [relative-time]”.
1. If there is no [previous different status], display “Since the service was loaded, [relative-time]” (Liz’s note: need to chat with Kimberly Garmoe about the grammar here and above).
1. For [previous different status] and [current status], display all health check status in lower case, except for “OK” (upper case).
1. For [relative-time]:
    1. When the time is less than a minute, display "a moment ago". For example, Changed from OK to warning a moment ago, Since the service was loaded, a moment ago.
    1. When the time is more than a minute, display only up to two units from the left starting from the one that's available: x year(s), x month(s), x day(s), x hour(s), x minute(s). For example, if it's more than a year, show Changed from warning to OK 1 year 3 months ago; if it's less than an hour, show Changed from OK to unknown 57 minutes ago.
    1. When hovering over or keyboard focused on the element, display a tooltip with the [UTC-time] in the style shown in the screenshot (note: the new tooltip component is going to be under development soon, probably when this issue is under development)
1. For [UTC-time]:
    1. Display it in the standardized RFC1123 format like this: Wed, 03 Jul 2019 17:08:53 UTC.
1. Style specification:
    1. Default or enabled state for the entire sentence:

```
font-family: Muli-Regular;
font-size: 12px;
color: #6F7878;
letter-spacing: 0.2px;
line-height: 18px;
```
1. 
    1. Hover and focus state for the [relative-time]: 
    1. Change the color of [relative-time] to `chef-primary-dark #222435`.

- Enter the studio.
- Build the applications-service and the automate-ui.
- Start all services.
- Populate the database by sending a single message through.
```
[1][default:/src:0]# applications_publish_supervisor_message  --sup-id "123abc" --fqdn app.example.com --site eu --name app --release 20101212010101
```
- Look at the UI and refresh a few times. (https://a2-dev.test/applications)
Screen Shot 2019-07-08 at 10 25 46 AM

- Send a change of health check to your service:
 ```
[1][default:/src:0]# applications_publish_supervisor_message  --health 1 --sup-id "123abc" --fqdn app.example.com --site eu --name app --release 20101212010101
```
- Look at the UI and refresh. (https://a2-dev.test/applications)
Screen Shot 2019-07-08 at 10 26 08 AM

### Timestamp Tooltip
timewizard

https://chefio.atlassian.net/browse/A2-907

- [ ] Code actually executed?

### CODE REVIEW

**Comment by afiune** [comment]
I am only missing the `hover` tooltip, all the data is coming from the backend and plugged all the way to the UI, we just need the ui-component to apply it.

**Comment by afiune** [comment]
I just finished the tooltip! 💥 

timewizard


### This change is ready!
tenor-141021449

**Comment by apriofrost** [comment]
Great work @afiune ! Overall this looks great to me! 
The style of the tooltip could use some updates but we will address it in another PR, probably by the UI team (https://github.com/chef/automate/issues/473 is the issue I going to put the detailed design spec in, still working on it). 
Another comment is about the time zone format. I saw in your PR it uses GMT. Could we update it to use UTC (the actual time is the same)? Reading this article (https://www.timeanddate.com/time/gmt-utc-time.html) It looks like we should stick to UTC since it doesn’t change for daylight saving time, or used by any territory that might use a daylight saving time. UTC is also the format that other monitoring tools use.

File: components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.scss [hunk 1]
```diff
@@ -17,6 +17,13 @@ ul {
   margin-top: 5px;
 }
 
+chef-tooltip {
+  display: inline-table;
+  background: $chef-primary-dark;
```
Shadae: 1.) If we remove this line &...

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
```diff
@@ -56,16 +56,18 @@
...
               *ngIf="service.channel === ''">NO CHANNEL</chef-badge>
           </div>
           <div class ="health-line">
             <chef-icon class="{{ service.health_check }}">{{ service.health_check | serviceStatusIcon }}</chef-icon>
-            {{ healthCheckStatus(service) }}
+            {{ healthCheckStatus(service.health_check) }}
           </div>
-          <div class ="timewizard-line">
-            Since {{ service.current_health_since }} ago
+          <chef-tooltip [attr.for]="'health-timestamp' + i">{{ formatTimestamp(service.health_updated_at) }}</chef-tooltip>
```
Shadae: 2.) change this to `<chef-tooltip style='--tooltip-color: #222435;' [attr.for]="'health-timestamp' + i">{{ formatTimestamp(service.health_updated_at) }}</chef-tooltip>`

components/applications-service/integration_test/services_test.go
```diff
@@ -486,6 +488,27 @@ func assertServicesEqual(t *testing.T, expected, actual []*applications.Service)
...
+			// By default the previous health check should be set to NONE
+			assert.Equal(t,
+				applications.HealthStatus_NONE,
+				actual[i].PreviousHealthCheck,
+				"The previous health check of a service is not the expected one")
+
+			// The health check should have been updated less than two seconds ago
+			healthUpdatedAt, err := ptypes.Timestamp(actual[i].HealthUpdatedAt)
+			assert.Nil(t, err)
+			//@afiune upgrate testify go package to be able to use 'assert.Lessf()'
```
phiggins: The version we're using has `InDelta` which is a little nicer https://godoc.org/github.com/stretchr/testify/assert#InDelta

components/applications-service/integration_test/services_test.go
```diff
@@ -486,6 +488,27 @@ func assertServicesEqual(t *testing.T, expected, actual []*applications.Service)
...
+			assert.Truef(t,
+				float64(2) > time.Now().Sub(healthUpdatedAt).Seconds(),
+				"The health check should have been updated less than two seconds ago")
+
+			// The current health since field should be empty since we are testing that
+			// the health was updated less than two seconds ago and our timef library doesn't
+			// display milliseconds therefore this field should always be empty
+			assert.Truef(t,
```
phiggins: You could use `Contains` here https://godoc.org/github.com/stretchr/testify/assert#Contains

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
```diff
@@ -56,16 +56,18 @@
...
+          <div class="timewizard-line">
+            {{ timewizardMessage(service.health_check, service.previous_health_check) }}
+            <a [id]="'health-timestamp' + i"> {{ service.current_health_since }} ago</a>
```
phiggins: Nitpick: should this be `'health-timestamp-'`?

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
afiune: I am changing this now but, out of curiosity, why do we prefer the inline style here instead of the css?

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.scss
afiune: changed

components/applications-service/integration_test/services_test.go
afiune: But what would it be the criteria? We would like to match three cases:

- `""`
- `"1 second"`
- `"2 seconds"`

phiggins: The second example, `assert.Contains(t, ["", "1 second"], actual[i].CurrentHealthSince)`, seems equivalent to what you're testing now.

afiune:
I modified everything I could, @Shadae using momentjs didn't show up the `UTC` label we wanted, also I found some info that indicates that, maybe, we should be displaying `GMT`

https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.3.1
>All HTTP date/time stamps MUST be represented in Greenwich Mean Time (GMT), without exception. For the purposes of HTTP, GMT is exactly equal to UTC (Coordinated Universal Time). This is indicated in the first two formats by the inclusion of "GMT" as the three-letter abbreviation for time zone, and MUST be assumed when reading the asctime format. HTTP-date is case sensitive and MUST NOT include additional LWS beyond that specifically included as SP in the grammar.

I used momentjs and I got the same result:

Screen Shot 2019-07-09 at 12 58 07 AM

scottopherson:
@afiune we can format dates however we please in the HTML content. I believe that spec only applies to HTTP communication; not the content being communicated.

We just need to decide what format we want to use for these tooltips. I'd vote for simply displaying the ISO format:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString
https://momentjs.com/docs/#/displaying/as-iso-string/

But we can pretty much do whatever with `moment.format()`.

Github displays a localized date string:
Screen Shot 2019-07-08 at 8 12 00 PM

scottopherson:
Perhaps we could punt on choosing the best formats by allowing the user to specify somewhere how they want dates to be displayed in the UI. Though that would be a larger feature discussion of course. 




It's actually a part of an even larger [i18n](https://angular.io/guide/i18n) discussion to be had but I digress.

susanev:
yes!! the ux team would like this to be a user setting in the future.

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
afiune: @Shadae With this change now the unit tests are failing:
```
src/app/page-components/services-sidebar/services-sidebar.component.html
--
  | L67 \|          <chef-tooltip style='--tooltip-color: #222435;' [attr.for]="'health-timestamp-' + i">{{ for...
  | ^ Inline style [  style='--tooltip-color: #222435;' ] cannot be use. (inline-style-disabled)
```

components/applications-service/integration_test/services_test.go
afiune: changed

components/applications-service/integration_test/services_test.go
afiune: changed

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
afiune: changed

afiune:
@scottopherson @susanev Thank you so much for your feedback, though your comments introduced more doubts than answers to the problem of time formatting 😅 

In my opinion, we (mainly UX) should talk as a group and decide what is the format we want to enforce in the UI, I like systems that let you change how you see times, but as we mentioned we are not close to having that flexibility in the UI, therefore using a single, common way of presenting time is the right thing to do here.

I have in the code three proposals:

1. Forcing UTC with custom RFC format. (If we like this proposal, we should move this format variable to a common place where other components in the UI can use it)
1. ISO format (as @scottopherson suggested it)
1. UTC format (displaying `GMT` from underlying javascript function)

Which option do we want to move forward with? @apriofrost @chef/automate-ui-team ?

afiune:
@scottopherson @susanev Thank you so much for your feedback, though your comments introduced more doubts than answers to the problem of time formatting 😅 

In my opinion, we (mainly UX) should talk as a group and decide what is the format we want to enforce in the UI, I like systems that let you change how you see times, but as we mentioned we are not close to having that flexibility in the UI, therefore using a single, common way of presenting time is the right thing to do here.

I have in the code three proposals:

1. Forcing UTC with custom RFC format. (If we like this proposal, we should move this format variable to a common place where other components in the UI can use it)
1. ISO format (as @scottopherson suggested it)
1. UTC format (displaying `GMT` from underlying javascript function)

Which option do we want to move forward with? @apriofrost @chef/automate-ui-team ?

apriofrost:
After discussion with @afiune , we've agreed on accepting the PR as it is even though it doesn't match exactly what's defined in the 4th point of "Definition of Done":

> 4. For [relative-time]:
>     - When the time is less than a minute, display "a moment ago". For example, Changed from OK to warning a moment ago, Since the service was loaded, a moment ago.
>     - When the time is more than a minute, display only up to two units from the left starting from the one that's available: x year(s), x month(s), x day(s), x hour(s), x minute(s). For example, if it's more than a year, show Changed from warning to OK 1 year 3 months ago; if it's less than an hour, show Changed from OK to unknown 57 minutes ago.

The difference is that the PR displays the "seconds" data in the UI. We've come to an agreement that it's acceptable because: 
1. We don't want to add too much UI logic to the backend/API at this point.
2. The length of the label looks fine with the limitation of just displaying two units.

We keep the description in the issue as it is so that the @chef/ux-team can refer to it for future iteration on unifying the time format across Automate.

danielsdeleo:
Verified locally also 👍

PR #787, opened 2019-07-08
scripts/changed_components.rb: fix pkg_deps call

srenatus:
It didn't matter, but when running this on master, the pkg_build_deps
construction would fail for anything that is looking at
`$PLAN_CONTEXT/...` etc.

    cat: /../.nvmrc: No such file or directory
    core/curl core/coreutils chef/mlsa core/nginx core/jq-static

It didn't matter because the script in question doesn't care for build
deps/

### DIFF
--- a/scripts/changed_components.rb
+++ b/scripts/changed_components.rb
@@ -14,8 +14,7 @@ class Hash
 end
 
 def get_hab_deps(source_dir)
-  plan_path = "#{source_dir}/habitat/plan.sh"
-  depString = `bash -c '. #{plan_path}; echo ${pkg_deps[*]}'`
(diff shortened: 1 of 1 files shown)


scripts/changed_components.rb
```diff
@@ -14,8 +14,7 @@ def tsort_each_child(node, &block)
 end
 
 def get_hab_deps(source_dir)
-  plan_path = "#{source_dir}/habitat/plan.sh"
-  depString = `bash -c '. #{plan_path}; echo ${pkg_deps[*]}'`
+  depString = `bash -c 'cd ${source_dir}; PLAN_CONTEXT=habitat . habitat/plan.sh; echo ${pkg_deps[*]}'`
```
stevendanna: ⚠️ Pre Coffee Review: Should this be `#{source_dir}`?

srenatus: Shoot I'll fix that...

PR #788, opened 2019-07-08
scripts/changed_components.rb: fix pkg_deps call again

srenatus:


srenatus:
https://buildkite.com/chef/chef-automate-master-verify-private/builds/2888#b26830f5-10fe-497b-9fdd-3e4750fd89d1/134-229 yay no errors

PR #806, opened 2019-07-08
updating codeowners

vjeffrey:
updating to ensure we ping ux for review when we
touch the ui, and changing all references to infra-automation-team
and compliance-team to superteam


susanev:
@apriofrost @jonong1972 what do y'all think of also getting notified for changes to `automate-cli`?

apriofrost:
@susanev I'm good with just the UI related files. I used to receive way too many notifications when I was the owner of the `hab-cli` related files. I'm worried about that would happen again. :)

PR #837, opened 2019-07-10
convert Cypress to Typescript

bcmdarroch:

Srry javascript we like type safety ;)

We decided it'd be a good idea to keep Cypress consistent with the rest of our UI code, which is all written in Typescript. 

### :+1: Definition of Done
- Cypress tests pass locally and on buildkite

- see [e2e README](https://github.com/chef/automate/blob/master/e2e/README.md#running-cypress-locally) for steps to run cypress test locally

- used [this helpful script](https://github.com/bahmutov/add-typescript-to-cypress) to get Cypress configured for TS

- [ ] Code actually executed?


e2e/cypress/plugins/index.js
```diff
@@ -1,17 +1,10 @@
...
-// You can read more here:
-// https://on.cypress.io/plugins-guide
-// ***********************************************************
+const cypressTypeScriptPreprocessor = require('./cy-ts-preprocessor')
 
 // This function is called when a project is opened or re-opened (e.g. due to
 // the project's config changing)
 
-module.exports = (on, config) => {
+module.exports = on => {
```
srenatus: [nit] Either we keep the `config` bit or we remove the comment below explaining it, the proposed halfway thing is confusing :wink:

e2e/package.json
```diff
@@ -15,5 +15,10 @@
   "license": "ISC",
   "dependencies": {
     "cypress": "^3.1.5"
+  },
+  "devDependencies": {
```
srenatus: ❔ I thought using this was a one-time-thing; so, does this need to be added?

e2e/cypress/index.d.ts
```diff
@@ -0,0 +1,9 @@
+declare namespace Cypress {
+  interface Chainable<Subject> {
+    login(url: string, username: string): Cypress.Chainable<any>
```
srenatus: ❓ Can we do better than `any` here? [Maybe `{}`?](https://mariusschulz.com/blog/typescript-2-2-the-object-type#object-vs-object-vs)

e2e/cypress/support/index.d.ts
```diff
@@ -0,0 +1,9 @@
+declare namespace Cypress {
```
srenatus: Do we need this file twice? 😃

e2e/cypress/integration/ui/settings/01_user_mgmt.spec.ts
```diff
@@ -1,10 +1,12 @@
+/// <reference types="cypress"/>
```
srenatus: https://github.com/cypress-io/cypress/issues/1152#issuecomment-509729392 this suggests that there's an alternative way that doesn't require this to be added to every file... but I have tried neither.

e2e/cypress/support/index.d.ts
bcmdarroch: unfortunately we weren't able to figure out a way to use `cy.login` inside the `adminLogin` command definition without duplicating the `index.d.ts` under in the `support` dir.

we could also do away with the `admin` login helper

bcmdarroch: the main issue is that adding cypress custom commands in typescript feels very brittle at this point. there's dozens of different ways to do it according to [this issue](https://github.com/cypress-io/cypress/issues/1065), and even [the example](https://github.com/omerose/cypress-support/blob/master/cypress/support/commands.ts) the cypress docs point to didn't "just work" as expected. only adding the `index.d.ts` finally got the spec files to recognize the custom commands.

this approach might be more stable in the longrun: https://medium.com/@NicholasBoll/cypress-io-scaling-e2e-testing-with-custom-commands-6b72b902aab

e2e/cypress/integration/ui/settings/01_user_mgmt.spec.ts
bcmdarroch: 😮 good find, that solved the extra `index.d.ts` issue too!

e2e/cypress/plugins/index.js
bcmdarroch: hmm yeah the plugin changed that for reasons unknown. not sure if we'll ever need `config` but i'll drop the comments for now

e2e/cypress/support/index.d.ts
bcmdarroch: actually looks like [this comment](https://github.com/cypress-io/cypress/issues/1152#issuecomment-509729392) solved the problem!

e2e/package.json
bcmdarroch: lol just some shameless advertising by the plugin i spose

.expeditor/buildkite/cypress.sh
```diff
@@ -18,5 +18,5 @@ do
   export CYPRESS_RECORD_KEY="$CYPRESS_RECORD_KEY"
 
   npm install # get dependencies defined in e2e/package.json
-  cypress run --record
+  npm run cypress:run --record
```
srenatus: That's a thing? Huh. I wouldn't have thought that this would work 🤔

bcmdarroch: lemme double-check, pretty sure it should work

bcmdarroch: oo good call that didn't work like i hoped 😅

srenatus: Sorry for spoiling the post-merge surprise 🎉

bcmdarroch: now that we're saving the failure screenshots and videos as buildkite artifacts, i wonder what we're really getting out of recording to the cypress dashboard anyway 💭

vjeffrey: my fly-by two cents is that we should drop the recording to cypress dashboard. only a few of us have access anyways and if we're already uploading things as bk artifacts, then, as you suggest, i don't think we're getting much out of it

srenatus: Good point -- however, we're **not** uploading artifacts in the deploy/dev test code (the file we're commening on).

vjeffrey: ooooh, darn -- is there a way to do so?

srenatus: Sure, we do it here: https://github.com/chef/automate/pull/837/files#diff-58a979e454f063e727ec94a03c56253dR22

srenatus: What I have no idea about is: _Where are BK artifacts stored? Do they cause much costs on the long run?_ (It's probably S3; but is it _our_ S3 or BK's S3? (Do they bill for artifacts?))

srenatus: Pinged releng about it: https://chefio.slack.com/archives/C074S2A94/p1562868350187500

bcmdarroch: i had added that 👍

bcmdarroch: i'll keep it recording for now but maybe after we've seen a few deploys save the recordings as expected, i'll go in and drop the recording stuff

bcmdarroch: depending on what releng says

e2e/package.json
bcmdarroch: actually looks like that was standing in for the `webpack-preprocsser` and `ts-loader` deps. so i just added those back in

e2e/cypress/plugins/index.js
```diff
@@ -1,17 +1,9 @@
-// ***********************************************************
-// This example plugins/index.js can be used to load plugins
-//
-// You can change the location of this file or turn off loading
-// the plugins file with the 'pluginsFile' configuration option.
-//
-// You can read more here:
-// https://on.cypress.io/plugins-guide
-// ***********************************************************
+const cypressTypeScriptPreprocessor = require('./cy-ts-preprocessor')
```
msorens: Would like to see a comment at the top of the file:
```
// This file was generated when running the js-to-ts conversion.
// See https://github.com/bahmutov/add-typescript-to-cypress
```

Did that conversion also create the above file cy-ts-preprocessor.js ?
If so please add the same comment there.
If not, would be good to know where those settings came from.

e2e/cypress/index.d.ts
```diff
@@ -0,0 +1,9 @@
+declare namespace Cypress {
+  interface Chainable<Subject> {
```
msorens: Please add comment:
```
// All added commands in commands.ts must be included in this interface.

bcmdarroch:
@msorens thanks for pointing that out, I saw those same failures locally as well. The problem is that those specs expect a few form values to be provided by environment variables. we should probably update those specs to skip if those variables aren't set

Issue #629, opened 2019-06-18
Add update strategy to the channel badge


apriofrost:
[jira link](https://chefio.atlassian.net/browse/A2-887)

PR #839, opened 2019-07-10
Add update strategy to the channel badge

afiune:
The update strategy tells users how the services will be updated among a group. Habitat supports three update strategies: `none`, `rolling`, and `at-once`. 

Showing the data along with the channel name helps users understand where the service will get an update from, and how it will update.

### :+1: Definition of Done

Include the update strategy for a service instance in the channel badge, in the format of `[channel name]: [update strategy name]` (there is a space after the `:`)

When there is NO channel and NO update strategy, display `NO CHANNEL: NONE`

#### Mockup
image

- Enter your studio.
- Build the `automate-gateway`, `applications-service`, and the `automate-ui`.
- Start all services. `start_all_services`.
- Publish a single service message with no channel and no strategy:
```
[5][default:/src:0]# applications_publish_supervisor_message --sup-id 2cc9c042-65cd-4812-81ed-176f98afaca2
```
- Look at the UI: https://a2-dev.test/applications
Screen Shot 2019-07-10 at 9 55 26 AM

- Update the same service to have a channel and strategy:
```
[6][default:/src:0]# applications_publish_supervisor_message --sup-id 2cc9c042-65cd-4812-81ed-176f98afaca2 --channel stable --strategy rolling
```
-- or
```
[6][default:/src:0]# applications_publish_supervisor_message --sup-id 2cc9c042-65cd-4812-81ed-176f98afaca2 --channel stable --strategy at-once
```
- Look back at the UI: https://a2-dev.test/applications
image

https://chefio.atlassian.net/browse/A2-887 & https://github.com/chef/automate/issues/629

- [x] Code actually executed?


components/applications-service/integration_test/global_test.go
```diff
@@ -413,3 +423,109 @@ func habServicesABCD() []*habitat.HealthCheckEvent {
 		),
 	}
 }
+
+// @afiune there are cases where the order of the services that are returned are not
+// exactly the same, we do care about order but in some degree, for example health check
+// status and fqdn, but we don't care about things like order of supervisor_id so we can
+// skip that check by not specifying it into the expected response
+func assertServicesEqual(t *testing.T, expected, actual []*applications.Service) {
```
srenatus: ❓ `assert.ElementsMatch(t, expected, actual)` doesn't do the trick?

afiune: First of all, thanks for looking into this PR! Second, this was just a move of code from A to B file https://github.com/chef/automate/pull/839/commits/06b87a19ab2b7aba63660214cc2da1100a6844bd
Third and last, no, it doesn't do the trick since we might have db IDs and timestamps that won't match exactly the same.

srenatus: Ah, I see. Thanks.

components/applications-service/cmd/applications-publisher/main.go
```diff
@@ -126,6 +129,25 @@ func main() {
 		clientID = "applications-publisher-" + strconv.FormatInt(t.UnixNano(), 10)
 	}
 
+	// If the channel was specified, add the update config
+	if len(channel) > 0 {
```
srenatus: It's probably fine, but if no channel is provided, but an update-strategy, the latter will just be ignored; it won't raise an error.

components/applications-service/pkg/storage/client.go
```diff
@@ -58,6 +88,7 @@ type Service struct {
 	Channel             string
 	Site                string
 	PreviousHealth      string    `db:"previous_health"`
+	UpdateStrategy      string    `db:"update_strategy"`
```
srenatus: I'm a little surprised about db struct tags here; in `pkg/storage/postgres`, I understand them. Here? Not really.

components/applications-service/pkg/storage/postgres/db.go
```diff
@@ -106,6 +106,7 @@ type service struct {
 	Channel             string    `db:"channel"`
 	FullPkgIdent        string    `db:"package_ident"`
 	PreviousHealth      string    `db:"previous_health"`
+	UpdateStrategy      string    `db:"update_strategy"`
```
srenatus: I suppose we could also use a PG ENUM here. But I know it's cumbersome to do all the translation... (we've got example code for this in authz-service, but using `database/sql`)

components/applications-service/pkg/storage/postgres/hab_event.go
```diff
@@ -7,6 +7,7 @@ import (
 
 	"github.com/chef/automate/api/external/applications"
 	"github.com/chef/automate/api/external/habitat"
+	"github.com/chef/automate/components/applications-service/pkg/storage"
 	"github.com/go-gorp/gorp"
```
srenatus: [nit] It's nice if the imports are sorted; first stdlib, then third-party deps, then internal deps. `goimports -local github.com/chef/automate` supports that.

components/automate-ui/src/app/entities/service-groups/service-groups.model.ts
```diff
@@ -19,6 +19,7 @@ export interface Service {
   fqdn: string;
   site: string;
   channel: string;
+  update_strategy: string;
```
srenatus: This could use a string literal type, I guess? `update_strategy: "NONE" | "ROLLING" | "AT_ONCE";`

components/applications-service/cmd/applications-publisher/main.go
afiune: @srenatus not sure if you noticed but, we default the strategy to `at-once`, but the important thing here is that channel, we want to replicate the behavior of habitat and, if in Habitat you don't specify a channel, then the strategy won't be populated. TL DR no channel, no strategy.

components/applications-service/pkg/storage/client.go
afiune: yeah, I might have to fix that in a next PR. I agree with what you are mentioning here!

components/applications-service/pkg/storage/postgres/db.go
afiune: What would it be the benefits of using an ENUM in postgres?

components/applications-service/pkg/storage/postgres/hab_event.go
afiune: I am supposedly running that command in my vim editor, not sure why it is not sorting them! 🤔

components/applications-service/cmd/applications-publisher/main.go
srenatus: Yeah, what I imagined to possibly happen is this:
When I call `applications-publisher --channel dev --update-strategy foobar`, I get an error. When I run `applications-publisher --update-strategy foobar`, I *don't*. In a perfect world, specifying `--update-strategy` without specifying `--channel` would perhaps be an error.

However, I don't know if our cli arg lib makes this simple; if it doesn't, I doubt it's worth it 🤷‍♂

components/automate-ui/src/app/entities/service-groups/service-groups.model.ts
afiune: FYI: It is `AT-ONCE`, but I like that idea!

srenatus: Yeah, you get the gist 😉

components/applications-service/cmd/applications-publisher/main.go
afiune: I see, well I can add this check for sure. At the end, this is just a publisher tool to populate our DB but we want it to be as similar as a habitat supervisor as possible.

components/automate-ui/src/app/entities/service-groups/service-groups.model.ts
afiune: tenor-70491942

components/applications-service/pkg/storage/postgres/hab_event.go
afiune: Is this how you want them¿?
```go
import (
	"fmt"
	"strings"
	"time"

	dblib "github.com/chef/automate/lib/db"

	"github.com/go-gorp/gorp"
	"github.com/golang/protobuf/ptypes"
	timestamp "github.com/golang/protobuf/ptypes/timestamp"
	"github.com/pkg/errors"
	"github.com/prometheus/client_golang/prometheus"
	"github.com/prometheus/client_golang/prometheus/promauto"

	"github.com/chef/automate/api/external/applications"
	"github.com/chef/automate/api/external/habitat"
	"github.com/chef/automate/components/applications-service/pkg/storage"
)
```

srenatus: 💭  I'd cluster the dblib with the rest in the last section, but it's also a bit of a nitpicky comment anyways -- :+1:

components/applications-service/cmd/applications-publisher/main.go
srenatus: Your call -- I have to little context to discern if this an issue. Probably not 😄

components/applications-service/pkg/storage/postgres/db.go
srenatus: The same as your storage layer's enum, I guess, just one layer down. Anyways, in authlahoma, we've stopped doing that, as it's cumbersome to deal with enums _on every layer_. 🙄

components/applications-service/pkg/storage/client.go
```diff
@@ -42,6 +42,36 @@ const (
 	None     = "NONE"
 )
 
+type UpdateStrategy int
+
+const (
+	NoneStrategy UpdateStrategy = iota
```
danielsdeleo: 💭 could this cause an issue if, say habitat adds a new kind of upgrade strategy and the user adopts the new upgrade strategy before they upgrade automate to a newer version that has this "enum" updated? If we really want this to be an enum then maybe `0` should be "Unknown" or something like that(?). "NONE" implies that the supervisor won't update following a package promotion, which could affect a use case like "find all the services that will update if I promote package X to the "production" channel.

afiune: No. When habitat has a new upgrade strategy, the habitat team has to also update the protobuf enum defined in the event messages, that is where automate and habitat won’t be in sync and we need to update our protobuf and code to accept the new strategy. 

Going deeper into this topic, any modification that habitat will do will also trigger a need for us to update our protobuf and code. Unless we code to accept anything that comes through, which has its trade offs. 

What would you like do to here? Remove the enum or continue having it?

afiune:
Talked with @apriofrost and @danielsdeleo about those cases where Habitat could introduce a new update strategy and Automate is not capable to understand it, an example is a customer that upgrade Habitat to the latest version with a new update strategy but not upgrade Automate, in such cases we would like to create a new `UNRECOGNIZED` label/enum that will cover any other strategy that Automate doesn't know about but Habitat is sending.

I will work on this in a separate PR since it will be simpler and smaller for review.

components/applications-service/pkg/storage/client.go
afiune: Another fix that you pointed out @srenatus https://github.com/chef/automate/pull/906

PR #840, opened 2019-07-10
[CI/nightly] Build ALL components during nightly build

stevendanna:
We periodically find that components no longer build because of
dependency changes. We typically find out that these components don't
build when developers are doing cleanup or maintenance work such as
updating dependencies. Only finding build problems during such tasks
increases the cost of starting such tasks and means that we discover
the problems long after the change that produced the problem, making
it harder to debug.

Currently, the nightly builds only rebuild whatever changed in the
latest commit to master, using the same changed components logic that
the rest of our verify pipelines use.

This changes the nightlies to build all components, which should allow
us to discover build issues more promptly, divorced from any
particular change.

- [ ] Code actually executed?

### DIFF
--- a/.studio/common
+++ b/.studio/common
@@ -227,22 +227,6 @@ function rebuild() {
   return $err
 }
 
-document "build_all" <<DOC
-  Build every hab package in components/
-DOC
-function build_all() {
-  log_line "Building all components..."
-  for d in /src/components/*/; do
-    [[ -f "$d/habitat/plan.sh" ]] || continue # skip what can't be built
-    if [[ "$d" = "/src/components/automate-chef-io" ]] && [[ -n "$AUTOMATE_OSS_BUILD" ]]; then
-        log_line "Skipping build of $d since AUTOMATE_OSS_BUILD is set."
-        continue
-    fi
-    log_line "Building $d"
-    build "$d" || return 1
-  done;
-}
-
 function verify_component() {
   if [[ "$1" == "deployment-service" || "$1" == "api" || "$1" == "config" ]]; then
     return 0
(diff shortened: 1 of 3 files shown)


stevendanna:
Test of the nightly logic here: https://buildkite.com/chef/chef-automate-master-nightly/builds/91#718e35ff-7388-4189-bbe4-b60f5d05bdb9

scripts/verify_build.sh
```diff
@@ -36,15 +36,30 @@ curl "https://packages.chef.io/manifests/dev/automate/latest.json" > results/dev
...
+
     cat << EOF | buildkite-agent annotate --style "info"
+This change rebuilds ALL components because BUILD_ALL=$BUILD_ALL:
+$(printf '* %s\n' "${changed_components[@]}")
+EOF
 
+else
+    mapfile -t changed_components < <(./scripts/changed_components.rb)
+    if [[ ${#changed_components[@]} -ne 0 ]]; then
+        cat << EOF | buildkite-agent annotate --style "info"
```
srenatus: [nit] Let's free that cat:
```suggestion
        buildkite-agent annotate --style "info" <<EOF
```
(I'm aware that this was just moved, feel free to ignore. 😉 )

.studio/common
```diff
@@ -227,22 +227,6 @@ function rebuild() {
   return $err
 }
 
-document "build_all" <<DOC
```
stevendanna: This function could not have been working for people since many of the packages it would have built do not build.  Thus, I've removed it since it implies people don't use it.

stevendanna:
It passed the nightly! https://buildkite.com/chef/chef-automate-master-nightly/builds/94

PR #842, opened 2019-07-10
automate-gateway: make unknown data cause an error

srenatus:
Before, if you (for example) POST

    { "foo": "bar", "baz": 300 }

where the protobuf message this is to be converted into only has a field

    string foo = 1;

this would silently be ignored.

Now, it's an error:

    {
      "error": "unknown field \"bar\" in request.CreateProjectReq",
      "message": "unknown field \"bar\" in request.CreateProjectReq",
      "code": 3,
      "details": []
    }


srenatus:
Turns out this affects a few of our inspec tests... might continue looking into this later...

srenatus:
To be discussed, reopened, eventually. I'll make a tech debt card.

PR #843, opened 2019-07-10
Update all chef-server pins to the latest

stevendanna:
This updates all pins for packages published by chef/chef-server and
introduces a script that can be used to update the pins in the
future. The script can be used as follows:

    NO_GIT=true ./.expeditor/update_chef_server.sh unstable

and can also be wired into expeditor. We have not added expeditor
configuration for it yet as I think we need to decide what channel we
want to follow. If we want to follow `stable` then the script will
need additional logic to ensure either that:

1) We never roll these pins backwards when currently pinned to
unstable, or

2) Updates all transitive dependencies pins as well.


stevendanna:
Merging this without configuring it with expeditor yet.

PR #847, opened 2019-07-10
[pg-gateway] Set server-fin timeout

stevendanna:
When a go process dies because of a signal or a panic, defer'd
database closes may not execute. In this case, HAProxy will keep the
client side connection in CLOSE-WAIT and the server-side connection in
ESTAB. Setting `server-fin` times out the server-side connection
quickly, avoiding a leaked process. Without this, the process would
not close for 5 minutes until the tunnel timeout is hit.

### DIFF
--- a/components/automate-pg-gateway/habitat/config/haproxy.conf
+++ b/components/automate-pg-gateway/habitat/config/haproxy.conf
@@ -14,18 +14,16 @@ defaults
     timeout tunnel 300s
     # Client inactivity connections for half-closed connections (i.e
     # the server has closed the connection and but the client has
-    # not).  The HAProxy documentation recommends setting this
-    # whenever `timeout tunnel` is also set.
-    timeout client-fin 50s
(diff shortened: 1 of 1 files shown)


components/automate-pg-gateway/habitat/config/haproxy.conf
```diff
@@ -17,6 +17,15 @@ defaults
...
     # whenever `timeout tunnel` is also set.
     timeout client-fin 50s
+    # NOTE(ssd) 2019-07-10: The HAProxy documentation says that this
+    # setting should not be needed; however, we have observed that
+    # when a postgresql client disconnects uncleanly, HAProxy does not
+    # immediately close the backend until the connection times out. If
+    # a service is failing in a loop, this can quickly lead to the
+    # exhaustion of available ports.
+    #
+    # This number was chosen with very little thought.
```
srenatus: 🚀

PR #848, opened 2019-07-10
[lib/cereal] manager.Stop() should close backend

stevendanna:
This allows users to defer the close to properly shutdown the db.


jaym:
to capture the slack discussion: it's weird that the manager takes ownership of a thing it didn't create. In reality, this isn't a problem as you'll have one of both and we can change it in the future if it matters

PR #851, opened 2019-07-10
Document how to configure external data stores.

### LINKED ISSUE #846: Finish documenting external datastores.

yzl:
# :nut_and_bolt: Description
Document the current tested state of configuring external data stores with Automate.

Fixes #846  

- [ ] Code actually executed?

### DIFF
--- a/components/automate-chef-io/content/docs/configuration.md
+++ b/components/automate-chef-io/content/docs/configuration.md
@@ -29,7 +29,7 @@ The `chef-automate init-config` command generates an annotated Chef Automate con
 
 #### Chef Automate FQDN
 
-To change the fully qualified domain name (FQDN) of your Chef Automate installation, create a `.toml` file that contains the partial configuration:
+To change the fully qualified domain name (FQDN) of your Chef Automate installation, create a TOML file that contains the partial configuration:
 
 ```TOML
 [global.v1]
@@ -51,7 +51,7 @@ The upgrade strategy determines when a Chef Automate installation is upgraded. T
 
 Changing the upgrade strategy from `none` to `at-once` will install the latest packages from your install channel.
 
-To change the upgrade strategy of your Chef Automate installation, create a `.toml` file that contains the partial configuration:
+To change the upgrade strategy of your Chef Automate installation, create a TOML file that contains the partial configuration:
 
 ```toml
 [deployment.v1.svc]
@@ -91,7 +91,7 @@ curl -X PUT -H "api-token: $TOK" -H "Content-Type: application/json" -d '{"name"
 
 #### Load Balancer Certificate and Private Key
 
-To change the load balancer certificate and private key of your Chef Automate installation, create a `.toml` file that contains the partial configuration:
+To change the load balancer certificate and private key of your Chef Automate installation, create a TOML file that contains the partial configuration:
 
 ```toml
 [[global.v1.frontend_tls]]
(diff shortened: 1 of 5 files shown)


components/automate-chef-io/content/docs/install.md
```diff
@@ -119,3 +126,126 @@ sudo chef-automate config patch config.toml
 ```
 
 and retry opening Chef Automate in your browser.
+
+### Configuring External Data Stores
```
yzl: I thought it was worth spelling this out here (instead of pointing to the configuration page) because if someone knows upfront they need an external data store, we want them to do their initial deploy that way. I'm not sure this is the best way to encourage that, so I'm willing to hear other ideas.

components/automate-chef-io/content/docs/install.md
```diff
@@ -31,6 +31,13 @@ sudo ./chef-automate init-config
 This will create a config.toml file with default values that should allow you to proceed,
 or you can edit it to specify desired FQDN, login name, and so on.
 
+If you have requirements around data size and/or redundancy, see [Configuring External
+Data Stores]({{< relref "#configuring-external-data-stores" >}}) for information on
+configuring Chef Automate to use an externally-deployed PostgreSQL database cluster
+and/or Elasticsearch cluster. If you have requirements around a highly-available
+deployment of Chef Automate, please reach out to a Customer Success or Professional
```
yzl: Are these the right folks to contact?

components/automate-chef-io/content/docs/configuration.md
```diff
@@ -698,6 +710,37 @@ Then run `chef-automate config patch </path/to/your-file.toml>` to deploy your c
 # name = "automate"
 ```
 
+#### Configuring An External PostgreSQL Database
+To use Chef Automate with an external PostgreSQL cluster, create a `.toml` file containing the following partial configuration.
```
stevendanna: 🤔I'm wondering if we prefer "TOML file" to  "`.toml` file"?

components/automate-chef-io/content/docs/configuration.md
```diff
@@ -698,6 +710,37 @@ Then run `chef-automate config patch </path/to/your-file.toml>` to deploy your c
...
+password = "<admin password>"
+[global.v1.external.postgresql.auth.password.dbuser]
+username = "<dbuser username>"
+password = "<dbuser password>"
+
+[global.v1.external.postgresql.backup]
+enable = true
+
+```
+
```
stevendanna: I _think_ we are OK not restarting the world in these directions because the restart-workaround code will restart all pgsql clients.  However, I think in general we might still want to do a full service restart so that database initialization code in the various services run again.

components/automate-chef-io/content/docs/install.md
stevendanna: I'm wondering if we want the the initial install instructions to be the _only_ instructions for now?

components/automate-chef-io/content/docs/configuration.md
yzl: We're not consistent, so I'm going to make us consistent by making it "TOML file".

components/automate-chef-io/content/docs/install.md
yzl: @stevendanna That makes sense to me, so I removed the ones on the configuration page.

PR #856, opened 2019-07-11
Fix fixing swagger copying

phiggins:
Followup to #711. Rewrote the bash bits without using `cpio` so hopefully it will play nicer with the environment expeditor runs in.

### DIFF
--- a/components/automate-chef-io/Makefile
+++ b/components/automate-chef-io/Makefile
@@ -19,8 +19,8 @@ clean_swagger_files:
 sync_swagger_files: clean_swagger_files
 	pushd ../../api/external && \
 		find . -name '*.swagger.json' | sed 's/^\.\///' | while read i ; do \
-			mkdir -p ../../components/automate-chef-io/$(SWAGGER_DIR)/`dirname $$i` && \
-			cp $$i ../../components/automate-chef-io/$(SWAGGER_DIR)/`dirname $$i` \
+			mkdir -p "../../components/automate-chef-io/$(SWAGGER_DIR)/`dirname $$i`" && \
+			cp $$i "../../components/automate-chef-io/$(SWAGGER_DIR)/$$i" \
 		; done && \
 	popd
 
--- a/components/automate-chef-io/Makefile
+++ b/components/automate-chef-io/Makefile
@@ -17,8 +17,19 @@ clean_swagger_files:
 	rm -rf ./$(SWAGGER_DIR)/*
 
 sync_swagger_files: clean_swagger_files
-	pushd ../../api/external && find . -name '*.swagger.json' | cpio -p -dumv ../../components/automate-chef-io/$(SWAGGER_DIR) && popd
-	pushd ../automate-gateway/api/ && find . -name '*.swagger.json' | cpio -p -dumv ../../automate-chef-io/$(SWAGGER_DIR) && popd
+	pushd ../../api/external && \
+		find . -name '*.swagger.json' | sed 's/^\.\///' | while read i ; do \
+			mkdir -p ../../components/automate-chef-io/$(SWAGGER_DIR)/`dirname $$i` && \
+			cp $$i ../../components/automate-chef-io/$(SWAGGER_DIR)/`dirname $$i` \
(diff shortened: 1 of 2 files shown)


components/automate-chef-io/Makefile
```diff
@@ -17,8 +17,19 @@ clean_swagger_files:
...
+			cp $$i ../../components/automate-chef-io/$(SWAGGER_DIR)/`dirname $$i` \
+		; done && \
+	popd
+
+	pushd ../automate-gateway/api/ && \
+		find . -name '*.swagger.json' | sed 's/^\.\///' | while read i ; do \
+			mkdir -p ../../automate-chef-io/$(SWAGGER_DIR)/`dirname $$i` && \
+			cp $$i ../../automate-chef-io/$(SWAGGER_DIR)/`dirname $$i` \
+		; done && \
+	popd
```
phiggins: If anyone would like to attempt to golf this down, `cd components/automate-chef-io` and `make sync_swagger_files`, then `git status` should show no changes to the swagger files (they are deleted and recopied on each invocation). Keep in mind this needs to run in a restricted env so things like `cpio` and `rsync` are probably not available.

components/automate-chef-io/Makefile
```diff
@@ -17,8 +17,19 @@ clean_swagger_files:
 	rm -rf ./$(SWAGGER_DIR)/*
 
 sync_swagger_files: clean_swagger_files
-	pushd ../../api/external && find . -name '*.swagger.json' | cpio -p -dumv ../../components/automate-chef-io/$(SWAGGER_DIR) && popd
-	pushd ../automate-gateway/api/ && find . -name '*.swagger.json' | cpio -p -dumv ../../automate-chef-io/$(SWAGGER_DIR) && popd
+	pushd ../../api/external && \
+		find . -name '*.swagger.json' | sed 's/^\.\///' | while read i ; do \
+			mkdir -p ../../components/automate-chef-io/$(SWAGGER_DIR)/`dirname $$i` && \
```
msorens: Safer to quote the argument in case of embedded spaces:
```suggestion
			mkdir -p "../../components/automate-chef-io/$(SWAGGER_DIR)/`dirname $$i`" && \
```

components/automate-chef-io/Makefile
```diff
@@ -17,8 +17,19 @@ clean_swagger_files:
...
+			cp $$i ../../components/automate-chef-io/$(SWAGGER_DIR)/`dirname $$i` \
```
msorens: What you have certainly works, but can do a bit less work this way:
```suggestion
			cp $$i "../../components/automate-chef-io/$(SWAGGER_DIR)/$$i" \
```

phiggins:
@msorens I tested your changes locally and they appear to work. Thanks!

components/automate-chef-io/Makefile
```diff
@@ -17,8 +17,19 @@ clean_swagger_files:
 	rm -rf ./$(SWAGGER_DIR)/*
 
 sync_swagger_files: clean_swagger_files
-	pushd ../../api/external && find . -name '*.swagger.json' | cpio -p -dumv ../../components/automate-chef-io/$(SWAGGER_DIR) && popd
-	pushd ../automate-gateway/api/ && find . -name '*.swagger.json' | cpio -p -dumv ../../automate-chef-io/$(SWAGGER_DIR) && popd
+	pushd ../../api/external && \
+		find . -name '*.swagger.json' | sed 's/^\.\///' | while read i ; do \
```
srenatus: Re: `s/^\.\///`, would the `./` in the path actually matter for the mkdir and cp calls below? 

```
$ mkdir -p foo/./bar/./././baz
$ find foo
foo
foo/bar
foo/bar/baz
$
```

components/automate-chef-io/Makefile
srenatus: 😃 _+1, insightful_

msorens:
@phiggins FYI: Looks like you applied my suggestions in the first chunk of code, but the exact same situation occurs in the second chunk of code.

PR #857, opened 2019-07-11
Consistency improvement and minor bug fixes for auth UI pages

msorens:

In the course of acceptance testing, I started noticing some inconsistent wordings, and error triggers. This PR is a general cleanup of a host of little things across policies, projects, users, teams, and tokens (we don't have any forms for roles, so nothing there).

#### Cross-resource concerns
1. Pulled out common validation regexes used in multiple places into a single constants file, helpers/auth/regex.ts.
2. Ensure all error displays do not end in a period.
3. Ensure wording the same for same phrases.
4. Ensure consistent grammar "A team is..." vs. "Team is....".
5. Ensure note for each Validator reminds the developer which template to sync with.
6. Ensure consistent formatting of <chef-error> elements.
7. Ensure "Required" error msg omits the phrase "non-blank" (as it really doesn't add value).

#### Specific resource changes
1. **CreateUser**: now triggers error if the full name is just whitespace.
2. **CreateUser**: now triggers error if the user name is empty. (It turns out that a regex disallowing an empty string is not good enough!)
3. **AddPolicyMember**: reset "active" error (error against state of the data, not a typing error) as soon as user edits the value again. That way, when pressing "Add" again, the error will be able to go from invisible to visible, thus appearing again.
4. **AddPolicyMember**: now triggers error if the member expression is just whitespace.
5. **AddPolicyMember**: now triggers error if the member expression is empty.
6. **CreateTeam**: now triggers error if ID is invalid.
7. **CreateTeam**: now triggers error if ID exceeds 64 characters.
8. **EditToken**: hooked up present--but unused--validator, allowing removing manual check.

- [x] Code actually executed?

### DIFF
--- a/components/automate-ui/src/app/helpers/auth/regex.ts
+++ b/components/automate-ui/src/app/helpers/auth/regex.ts
@@ -1,6 +1,10 @@
 export class Regex {
 
   public static readonly patterns = {
+
+    // all auth IDs use this pattern
+    ID: '[0-9a-z-]+',
+
     // NB: neither \S nor ^\s work inside the brackets in this regex language.
     NON_BLANK: '.*[^ ].*'
   };
--- a/components/automate-ui/src/app/helpers/auth/regex.ts
+++ b/components/automate-ui/src/app/helpers/auth/regex.ts
@@ -1,0 +1,8 @@
+export class Regex {
+
+  public static readonly patterns = {
+    // NB: neither \S nor ^\s work inside the brackets in this regex language.
+    NON_BLANK: '.*[^ ].*'
+  };
+
+}
--- a/components/automate-ui/src/app/pages/api-token/details/api-token-details.component.html
+++ b/components/automate-ui/src/app/pages/api-token/details/api-token-details.component.html
@@ -35,7 +35,7 @@
           <input chefInput formControlName="name" type="text" (keyup)="handleNameChange()">
         </chef-form-field>
       </form>
-      <chef-button primary [disabled]="disableSave || saveInProgress" (click)="saveNameChange()">
+      <chef-button primary [disabled]="!updateNameForm.valid || disableSave || saveInProgress" (click)="saveNameChange()">
         <chef-loading-spinner *ngIf="saveInProgress"></chef-loading-spinner>
         <ng-container *ngIf="!saveInProgress">
           <span *ngIf="disableSave">Saved</span>
--- a/components/automate-ui/src/app/pages/api-token/details/api-token-details.component.ts
+++ b/components/automate-ui/src/app/pages/api-token/details/api-token-details.component.ts
@@ -34,6 +34,7 @@ export class ApiTokenDetailsComponent implements OnInit, OnDestroy {
               fb: FormBuilder
               ) {
       this.updateNameForm = fb.group({
+        // Must stay in sync with error checks in api-token-details.component.html
         name: ['', Validators.required]
       });
     }
@@ -65,8 +66,7 @@ export class ApiTokenDetailsComponent implements OnInit, OnDestroy {
     }
 
     public handleNameChange(): void {
-      this.disableSave = this.updateNameForm.controls['name'].value === this.token.name ||
-                         this.updateNameForm.controls['name'].value.trim() === '';
+      this.disableSave = this.updateNameForm.controls['name'].value === this.token.name;
     }
 
     public saveNameChange(): void {
--- a/components/automate-ui/src/app/pages/api-token/list/api-token-list.component.ts
+++ b/components/automate-ui/src/app/pages/api-token/list/api-token-list.component.ts
@@ -58,10 +58,10 @@ export class ApiTokenListComponent implements OnInit {
     this.apiTokenCount$ = store.pipe(select(totalApiTokens));
 
     this.createTokenForm = fb.group({
-      // Make sure these stay in sync with error checks in create-object-modal.component.html
+      // Must stay in sync with error checks in create-object-modal.component.html
       name: ['', Validators.required],
-      id: ['', [
-        Validators.required, Validators.pattern(Regex.patterns.ID), Validators.maxLength(64)]]
+      id: ['',
+        [Validators.required, Validators.pattern(Regex.patterns.ID), Validators.maxLength(64)]]
     });
   }
 
--- a/components/automate-ui/src/app/pages/api-token/list/api-token-list.component.ts
+++ b/components/automate-ui/src/app/pages/api-token/list/api-token-list.component.ts
@@ -7,6 +7,7 @@ import { filter, takeUntil, map } from 'rxjs/operators';
 import { identity } from 'lodash/fp';
 
 import { NgrxStateAtom } from 'app/ngrx.reducers';
+import { Regex } from 'app/helpers/auth/regex';
 import { loading, EntityStatus } from 'app/entities/entities';
 import { Type } from 'app/entities/notifications/notification.model';
 import { CreateNotification } from 'app/entities/notifications/notification.actions';
@@ -19,8 +20,6 @@ import {
 } from 'app/entities/api-tokens/api-token.actions';
 import { CreateToken } from 'app/entities/api-tokens/api-token.actions';
 import { saveStatus, saveError } from 'app/entities/api-tokens/api-token.selectors';
-
-const ID_PATTERN = '[0-9a-z-]+';
 
 @Component({
   selector: 'app-api-tokens',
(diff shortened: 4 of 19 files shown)


components/automate-ui/src/app/helpers/auth/regex.ts
```diff
@@ -0,0 +1,12 @@
+export class Regex {
```
msorens: PR was getting big enough, but probably should rename this class to `Constants` so in addition to `patterns` could also have:
- `messages`, for things like "Only lowercase letters, numbers, and hyphens(-) are allowed." and "must be 64 characters or less"
- `numbers`, for things like the `64` that is used multiple times in `Validators.maxLength(64)`

components/automate-ui/src/app/pages/api-token/details/api-token-details.component.html
```diff
@@ -35,7 +35,7 @@
           <input chefInput formControlName="name" type="text" (keyup)="handleNameChange()">
         </chef-form-field>
       </form>
-      <chef-button primary [disabled]="disableSave || saveInProgress" (click)="saveNameChange()">
+      <chef-button primary [disabled]="!updateNameForm.valid || disableSave || saveInProgress" (click)="saveNameChange()">
```
msorens: Validator was defined in ts file but not used.

components/automate-ui/src/app/pages/api-token/details/api-token-details.component.ts
```diff
@@ -65,8 +66,7 @@ export class ApiTokenDetailsComponent implements OnInit, OnDestroy {
     }
 
     public handleNameChange(): void {
-      this.disableSave = this.updateNameForm.controls['name'].value === this.token.name ||
-                         this.updateNameForm.controls['name'].value.trim() === '';
```
msorens: This check is now handled automatically by Validator.

components/automate-ui/src/app/pages/api-token/list/api-token-list.component.ts
```diff
@@ -20,8 +21,6 @@ import {
 import { CreateToken } from 'app/entities/api-tokens/api-token.actions';
 import { saveStatus, saveError } from 'app/entities/api-tokens/api-token.selectors';
 
-const ID_PATTERN = '[0-9a-z-]+';
```
msorens: Magic value--boom!

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -72,7 +73,8 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
     fb: FormBuilder) {
 
     this.expressionForm = fb.group({
-      expression: ['', Validators.required]
+      // Must stay in sync with error checks in policy-add-members.component.html
+      expression: ['', [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]]
```
msorens: Yes, `required` is still needed even if `pattern` disallows an empty string.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
```diff
@@ -18,10 +18,14 @@ <h2>Add Member Expression</h2>
           <form [formGroup]="expressionForm">
             <chef-form-field>
               <label>Member expression *</label>
-              <input chefInput formControlName="expression" type="text">
+              <input chefInput formControlName="expression" type="text" (keyup)="resetErrors()" >
```
msorens: As soon as the value is edited, clear the error so it can then appear again if it needs to later.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -287,8 +293,8 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
       return;
     }
 
-    this.duplicateMember = member.name in this.membersAvailableMap;
-    if (this.duplicateMember) {
+    if (member.name in this.membersAvailableMap) {
+      this.duplicateMember = true;
```
msorens: Make code parallel to rest of function.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -298,9 +304,6 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
         return;
       }
     });
-    if (this.alreadyPolicyMember) {
-      return;
```
msorens: Unreachable code

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
```diff
@@ -18,10 +18,14 @@ <h2>Add Member Expression</h2>
...
               <chef-error *ngIf="unparsableMember">Invalid member expression.</chef-error>
               <chef-error *ngIf="duplicateMember">Member expression already in table.</chef-error>
               <chef-error *ngIf="alreadyPolicyMember">Member already in policy.</chef-error>
+                <chef-error *ngIf="(expressionForm.get('expression').hasError('required') || expressionForm.get('expression').hasError('pattern')) && expressionForm.get('expression').dirty">
```
srenatus: We don't incude `dirty` in the other `*ngIf`s -- is that a special thing here?

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
```diff
@@ -18,10 +18,14 @@ <h2>Add Member Expression</h2>
...
+                  Expression is required
+                </chef-error>
+
```
srenatus: [nit] indentation looks off, and this extra line can go

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -72,7 +73,8 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
     fb: FormBuilder) {
 
     this.expressionForm = fb.group({
-      expression: ['', Validators.required]
+      // Must stay in sync with error checks in policy-add-members.component.html
```
srenatus: 🤔 Can't we centralize the entire `['', [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]]` array, then?

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
msorens: Yes, it is--meant to add a pre-review comment there. The first three can only ever occur by someone actively pressing "Add", so they cannot show up by themselves on page load, and thus no need of `ngIf`. The latter is, like in all the other files, needing of the `dirty` flag so that the error does not show up upon page load.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
msorens: Indeed... another candidate for the Constants file I mentioned above; maybe next PR, though...?

components/automate-ui/src/app/pages/team-create/team-create.component.html
```diff
@@ -17,17 +17,23 @@
                 <span class="label">Team {{ versionedNameOrId }}</span>
                 <chef-input ngDefaultControl formControlName="teamId" type="text"></chef-input>
               </label>
-               <chef-error *ngIf="(teamCreateForm.get('teamId').hasError('required') || teamCreateForm.get('teamId').hasError('pattern')) && teamCreateForm.get('teamId').dirty">
-                  A non-blank team {{ versionedNameOrId }} is required
-                </chef-error>
+              <chef-error *ngIf="teamCreateForm.get('teamId').hasError('required') && teamCreateForm.get('teamId').dirty">
+                Team {{ versionedNameOrId }} is required
```
susanev: so... i think this is fine. but the form labels are _Team ID_ and _Name_ so its a bit weird that the error text uses _Team ID_ and _Team name_ and there is a disconnect there. if we can make those consistent, that would be rad.

components/automate-ui/src/app/pages/user-management/user-form/user-form.component.html
```diff
@@ -4,33 +4,38 @@
...
   </chef-form-field>
   <chef-form-field>
     <label>
       <span class="label">Username</span>
       <input chefInput formControlName="username" type="text">
     </label>
-    <chef-error *ngIf="createUserForm.get('username').hasError('pattern') && createUserForm.get('username').dirty">Invalid
-      username</chef-error>
+    <chef-error *ngIf="(createUserForm.get('username').hasError('required') || createUserForm.get('username').hasError('pattern')) && createUserForm.get('username').dirty">
+      Invalid username
```
susanev: can we be more specific here? this error isn't very helpful...

msorens: Split required and pattern to allow giving two specific error messages. 👍

components/automate-ui/src/app/pages/team-create/team-create.component.html
msorens: Fixed it both for team create and team edit. 👍

PR #865, opened 2019-07-11
[CI] Remove [integration] prefix from some jobs

stevendanna:
We know that everything in the `post-build` phase of the pipeline is
an integration test. All this does is make it hard to read what failed.

@srenatus gave me the idea. I think the gain in scan-ability is worth any
downside.

### DIFF
--- a/.expeditor/verify_private.pipeline.yml
+++ b/.expeditor/verify_private.pipeline.yml
@@ -109,7 +109,7 @@ steps:
   # or take a long time.
   - wait
 
-  - label: "[integration] automate-load-balancer"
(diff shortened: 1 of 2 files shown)


.expeditor/verify_private.pipeline.yml
```diff
@@ -618,7 +618,7 @@ steps:
           single-use: true
           privileged: true
 
-  - label: "[integration] ontop backup "
+  - label: "ontop backup "
```
msorens:
```suggestion
  - label: "ontop backup"
```

PR #870, opened 2019-07-12
[automate-867] apply rules integration test

bcmdarroch:

We've been using inspec for a while to meet our API integration test needs, but it is not always the most ideal tool for such tests.

Since we've had a good experience using Cypress to test the UI so far, we decided to try out Cypress for our API integration test needs, starting with testing out the flow for applying project rules to nodes.

### :+1: Definition of Done

- [x] using Cypress, we can run an API-only integration test to ensure rule changes get applied to nodes as expected
    - [x] when a project gets a new rule and all rule changes are applied, the nodes affected by those rules can be filtered by the rule's project 
    - [x] when a project's rules are updated and all rule changes are applied, the nodes affected by those rules react accordingly to the new filter rules
    - [x] when a project's is deleted and all rule changes are applied, the nodes affected by those rules can 
 *no longer* be filtered by the deleted rule's project 

TODO:
- [x] fix failing update rule test
- [x] skip if not on v2.1

- run cypress locally by [following these steps](https://github.com/chef/automate/blob/master/e2e/README.md#running-cypress-locally)

- [ ] Code actually executed?

### DIFF
--- a/.expeditor/buildkite/cypress.sh
+++ b/.expeditor/buildkite/cypress.sh
@@ -6,12 +6,20 @@ cd /workdir/e2e
 
 data=$(curl --silent "https://a2-${CHANNEL}.cd.chef.co/assets/data.json")
 instances_to_test=$(jq -nr --argjson data "$data" '$data[] | select(.tags | any(. == "e2e")) | .fqdn')
+versions=(v1 v2 v2.1)
 
 for instance in ${instances_to_test[*]}
 do
   if ! jq -enr --arg fqdn "$instance" --argjson data "$data" '$data[] | select(.fqdn == $fqdn) | .tags | any(. == "saml")'; then
     export CYPRESS_SKIP_SSO=true
   fi
+
+  for version in ${versions[*]}
+  do
+    if jq -enr --arg version "$version" --arg fqdn "$instance" --argjson data "$data" '$data[] | select(.fqdn == $fqdn) | .tags | any(. == $version)'; then
+      export CYPRESS_IAM_VERSION=$version
+    fi
+  done
 
   echo "--- Executing Cypress tests against $instance"
   export CYPRESS_BASE_URL="https://$instance"
@@ -21,6 +29,6 @@ do
 
   if ! npm run cypress:record; then
       buildkite-agent artifact upload "cypress/videos/*;cypress/videos/**/*;cypress/screenshots/*;cypress/screenshots/**/*"
-      return 1
+      exit 1
   fi
 done
--- a/e2e/cypress/integration/api/iam/01_projects_api.spec.ts
+++ b/e2e/cypress/integration/api/iam/01_projects_api.spec.ts
@@ -88,6 +88,18 @@ describe('projects API', () => {
             })
           }
 
+          let totalNodes = 0
+          cy.request({
+            headers: {
+              'api-token': admin_token,
+              'projects': '(unassigned)'
+            },
+            method: 'GET',
+            url: '/api/v0/cfgmgmt/nodes?pagination.size=10'
+          }).then((response) => {
+            totalNodes = response.body.length
+          })
+
           cy.fixture('converge/avengers1.json').then(node1 => {
             cy.fixture('converge/avengers2.json').then(node2 => {
               cy.fixture('converge/xmen1.json').then(node3 => {
@@ -104,8 +116,7 @@ describe('projects API', () => {
               })
             })
           })
-
-          cy.wait(5000)  // TODO replace with polling apply status request
+          waitForNodes(totalNodes)
 
           // confirm nodes are unassigned
           cy.request({
@@ -158,10 +169,10 @@ describe('projects API', () => {
           url: 'api/v0/ingest/events/chef/node-multiple-deletes',
           body: {
             node_ids: [
-              "7188b88b-2236-4e27-a875-d3a10a70c497",
-              "639844f4-2ce6-42ba-8c9d-853db69adff3",
-              "75c2376e-d07e-4d2b-ab43-d658c6250a62",
-              "da60e383-67f6-4501-b726-1f28e03bf6ea"
+              "f6a5c33f-bef5-433b-815e-a8f6e69e6b1b", 
+              "82760210-4686-497e-b039-efca78dee64b", 
+              "9c139ad0-89a5-44bc-942c-d7f248b155ba", 
+              "6453a764-2415-4934-8cee-2a008834a74a"
             ]
           }
         })
@@ -203,7 +214,8 @@ describe('projects API', () => {
           method: 'POST',
           url: '/apis/iam/v2beta/apply-rules'
         })
-        cy.wait(5000) // TODO replace with polling apply status request
+        // waitForSuccessfulApply()
+        cy.wait(5000) 
 
         // confirm rules are applied
         for (let project of [avengers_project, xmen_project]) {
@@ -265,7 +277,8 @@ describe('projects API', () => {
           method: 'POST',
           url: '/apis/iam/v2beta/apply-rules'
         })
-        cy.wait(5000) // TODO replace with polling apply status request
+        // waitForSuccessfulApply()
+        cy.wait(5000) 
 
         cy.request({
           headers: {
@@ -292,7 +305,8 @@ describe('projects API', () => {
           method: 'POST',
           url: '/apis/iam/v2beta/apply-rules'
         })
-        cy.wait(5000)  // TODO replace with polling apply status request
+        // waitForSuccessfulApply()
+        cy.wait(5000) 
 
         cy.request({
           headers: {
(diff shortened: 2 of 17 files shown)


e2e/cypress/integration/api/iam/01_projects_api.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+          url: '/apis/iam/v2beta/policies/administrator-access/members:add',
+          body: {
+            members: [`token:${token_id}`]
+          }
+        })
+
+        // TODO fix this type
+        const nodes = <any>[]
+
+        Cypress.Promise.all([
```
bcmdarroch: decided to go with [this suggestion](https://github.com/cypress-io/cypress/issues/2932#issuecomment-476608299) over cypress' recommendation

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
```diff
@@ -0,0 +1,166 @@
...
+          auth: { bearer: admin.id_token },
+          method: 'POST',
+          url: '/apis/iam/v2beta/policies/administrator-access/members:add',
+          body: {
+            members: [`token:${token_id}`]
+          }
+        })
+
+        // TODO fix this type
+        const nodes = <any>[]
```
bcmdarroch: my typescript is so rusty 😭

srenatus: No shame in digging into the handbook, really: https://www.typescriptlang.org/docs/handbook/basic-types.html

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
bcmdarroch: turned out this didn't work so lots of callbacks it is

.expeditor/buildkite/cypress.sh
```diff
@@ -6,13 +6,21 @@ cd /workdir/e2e
...
 
 for instance in ${instances_to_test[*]}
 do
   if ! jq -enr --arg fqdn "$instance" --argjson data "$data" '$data[] | select(.fqdn == $fqdn) | .tags | any(. == "saml")'; then
     export CYPRESS_SKIP_SSO=true
   fi
 
+  for version in ${versions[*]}
+  do
+    if jq -enr --arg version "$version" --arg fqdn "$instance" --argjson data "$data" '$data[] | select(.fqdn == $fqdn) | .tags | any(. == $version)'; then
```
bcmdarroch: there's no `version` key for the tags, so we have to check for each tag by value `v1, v2, or v2.1`

.expeditor/buildkite/cypress.sh
```diff
@@ -21,6 +29,6 @@ do
 
   if ! npm run cypress:record; then
       buildkite-agent artifact upload "cypress/videos/*;cypress/videos/**/*;cypress/screenshots/*;cypress/screenshots/**/*"
-      return 1
+      exit 1
```
bcmdarroch: to make it more clear this is a cypress failure case, not a bash failure

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
```diff
@@ -0,0 +1,370 @@
+const admin_token = 'ZNhnqJyDqgkXVk_bpobS3YhIMz0='
```
blakestier: is it possible to instantiate and fetch this token? feels funny to have it hard-coded like this

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
```diff
@@ -0,0 +1,370 @@
...
+  id: `avengers-project-${Cypress.moment().format('MMDDYYhhmm')}`,
+  name: "Test Avengers Project"
+}
+
+const xmen_project = {
+  id: `xmen-project-${Cypress.moment().format('MMDDYYhhmm')}`,
+  name: "Test X-men Project"
+}
+
+let avengersRule = {
```
blakestier: does this work as a `const`? I know we're using it as part of the UPDATE but does that actually modify the variable?

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
```diff
@@ -0,0 +1,370 @@
...
+      values: ["xmen"]
+    }
+  ]
+}
+
+describe('projects API', () => {
+
+  if (Cypress.env('IAM_VERSION') != 'v2.1') {
+    describe('applying project rules', () => {
+      it.skip('must be run on IAM v2.1')
```
blakestier: 😎

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
blakestier: mmmm prob can't make one without one 🤔 conundrum

terraform/test-environments/scenarios/single_local_iamv2p1_fresh_install.tf
```diff
@@ -35,6 +35,7 @@ module "single_local_iamv2p1_fresh_install" {
     X-Deployment-Type  = "local"
     X-Channel          = "${var.channel}"
     X-CI-Test          = "e2e"
+    X-IAM              = "v2.1"
```
blakestier: what's the `X-` part?

blakestier:
Looks good so far! Had some questions for my own learnding but I'm so excited you've done all this work (and there's a lot of it 😅 ). Happy to review when you're all done-- maybe worth a walk-through 🎉

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
```diff
@@ -319,3 +333,38 @@ describe('projects API', () => {
...
+//   cy
+//     .request({
+//       headers: {
+//         'api-token': admin_token      
+//       },
+//       method: 'GET',
+//       url: '/apis/iam/v2beta/apply-rules'
+//     })
+//     .then((resp: Cypress.ObjectLike) => {
+//       if (resp.body.percentage_complete == 1 && resp.body.state == "not_running" && !resp.body.failed)
```
blakestier: should the percentage complete be `100`? or is that an oddly named variable?

integration/tests/cypress_e2e.sh
```diff
@@ -13,13 +13,16 @@ do_deploy() {
     chef-automate license apply "$A2_LICENSE"
 
     case $IAM in
+        v1)
+            export CYPRESS_IAM_VERSION='v1'
```
srenatus: Adding this to each case is the same as just doing `export CYPRESS_IAM_VERSION=$IAM` outside the case, isn't it? 😉

terraform/test-environments/scenarios/single_local_iamv2p1_fresh_install.tf
srenatus: It's jus a convention; like, _these tags are ours_, I believe.

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
```diff
@@ -0,0 +1,370 @@
...
+      attribute: "CHEF_ORGS",
+      operator: "EQUALS",
+      values: ["xmen"]
+    }
+  ]
+}
+
+describe('projects API', () => {
+
+  if (Cypress.env('IAM_VERSION') != 'v2.1') {
```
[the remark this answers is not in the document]
bcmdarroch: Ooh good call on adding tslint, I'll do that

.expeditor/buildkite/cypress.sh
bcmdarroch: @blakestier thought the `return` made it look like the script itself had failed, rather than the cypress test.

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
bcmdarroch: I could but then I couldn't share the token value with all of the tests. Though I can try to figure something else out.

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
bcmdarroch: yeah we made this `let` so we could modify `avengersRule` itself. we could make this variable a const and just create a new `updatedAvengersRule`. that might be a bit cleaner

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
bcmdarroch: Yeah this might require some work on the ingest-service side. Right now percentage_complete returns as `0` or `1` (at least when dealing with very small data sets like in this test)

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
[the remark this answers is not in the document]
bcmdarroch: actually since today got away from me, I'm gonna add a small issue for adding TSlint to cypress

.expeditor/buildkite/cypress.sh
srenatus: Indeed! 
```
$ grep . script_*.sh
script_one.sh:#!/bin/bash
script_one.sh:return 1
script_two.sh:#!/bin/bash
script_two.sh:exit 1
$ ./script_one.sh
./script_one.sh: line 3: return: can only `return' from a function or sourced script
$ echo $?
1
$ ./script_two.sh
$ echo $?
1
$
```

integration/tests/cypress_e2e.sh
```diff
@@ -47,6 +49,6 @@ do_test_deploy() {
     npm install # get dependencies defined in e2e/package.json
     if ! npm run cypress:run; then
         buildkite-agent artifact upload "cypress/videos/*;cypress/videos/**/*;cypress/screenshots/*;cypress/screenshots/**/*"
-        return 1
+        exit 1
```
srenatus: The situation here is different, though -- **this defines a function**, and `return 1` is thus the right thing to do, I'd say. See:

```
$ cat script_three.sh
#!/bin/bash

foo() {
  return 1
}

foo
$ ./script_three.sh
$ echo $?
1
$
```

PR #871, opened 2019-07-12
Remove disabled webhook URLs from repository completely

stevendanna:
This removes old slack webhook URLs. These webhooks have been
disabled, but to make scanning easier, they've been replaced with fake
data that can be excluded in the credscan tool.

### DIFF
--- a/tools/credscan/credscan.go
+++ b/tools/credscan/credscan.go
@@ -102,6 +102,10 @@ var a2Config = config{
 	},
 	contentInclude: []pattern{
 		{
+			description: "slack hooks",
+			regex:       `https://hooks.slack.com`,
+		},
+		{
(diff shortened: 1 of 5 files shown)


tools/credscan/credscan.go
```diff
@@ -180,7 +192,7 @@ func main() {
 	// until we need more flexibility.
 	//
 	outbuf := &bytes.Buffer{}
-	gitGrepCmd := fmt.Sprintf("git grep -E \"%s\" | grep -v -E \"%s\"",
+	gitGrepCmd := fmt.Sprintf("git grep -a -E \"%s\" | grep -v -E \"%s\"",
```
srenatus: 💡

PR #873, opened 2019-07-12
[ci] Restructure tests to work around bug

stevendanna:
There is currently a bug related to a bad interaction between the
periodic cert check and our shutdown logic. The code is very
entangled and it is going to take some time to sort out safely. For
now, I think modifying the cert generation test may lead us to time
out less frequently.

### DIFF
--- a/integration/helpers/backup.sh
+++ b/integration/helpers/backup.sh
@@ -76,9 +76,16 @@ test_missing_checksums_file_fails() {
 
 test_can_regenerate_cert_after_restore() {
     log_info "Checking that automate can regnerate certs after restore"
-    log_info "Removing deployment-service certificate"
+    # TODO(ssd) 2019-07-12: We stop the deployment service first to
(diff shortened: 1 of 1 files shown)


integration/helpers/backup.sh
```diff
@@ -76,9 +76,16 @@ test_missing_checksums_file_fails() {
 
 test_can_regenerate_cert_after_restore() {
     log_info "Checking that automate can regnerate certs after restore"
-    log_info "Removing deployment-service certificate"
+    # TODO(ssd) 2019-07-12: We stop the deployment service first to
+    # work around a bug. The bug is related to the fact that the
+    # periodic cert check will notice the deleted cert, regenerate it,
+    # and restart us in the middle of the shutdown sequence, leading
+    # to problems.
+    log_info " - Stopping deployment-service certificate"
```
stevendanna:
```suggestion
    log_info " - Stopping deployment-service"
```

Issue #832, opened 2019-07-09
add db_id as primary key to auth databases


susanev:
## Overview

We'd like to keep the primary IDs of our databases immutable and safe from users.

Currently, roles and projects have a db_id column, but only roles uses it as the primary ID. Policies, teams, and tokens all use the id as the primary ID, and they have no db_id column.

Let's make our databases consistent, with each resource table (policies, roles, projects, teams, tokens) using an integer db_id as the primary ID. All relational tables should be updated as necessary.

Use db_id internally.

## Subtasks
- [x] update tokens table
- [x] update teams table
- [x] update members
- [x] add db_id to policy table and change related tables/constraints
- [x] add db_id to projects table and modify related tables/constraints
  - [x] `iam_projects <-> iam_[staged_]project_rules` #862 
  - [X] `iam_projects <-> iam_role_projects` #878
  - [x] `iam_projects <-> iam_policy_projects` #898 
  - [x] `iam_projects <-> iam_statement_projects` #905
  - [x] flip `PRIMARY KEY` constraint of `iam_projects` from `id` to `db_id` #913  
- [x] add db_id, policy_id to iam_statements/ remove iam_policy_statements #855
- [x] use statement db_id as key in `iam_statement_projects` also in #913
- [x] use `role` db_id in `iam_statements` #942
- [x] Make all columns that should have NULL constraint have NULL constraint #1052
The last one overlapped #820-- related content below:
- [x] Make `strict` use consistent in [*_db_id functions](https://github.com/chef/automate/pull/942#discussion_r306507012) #1055 

### Bonus tasks (read: I'm tired of this -- @srenatus )
- [x] Replace `iam_projects`'s `projects` column (type `text[]`) with a `iam_projects_projects` table having two columns of `db_id` (`INTEGER`) for properly assigning a project's projects
  - _Discussed in standup 07/25: we'd like to just drop the column today; and re-introduce it (properly) when we need the feature._
  - #1109 
- [x] Remove the `id` column of `iam_statements` and `iam_members`. They're internal artifacts we no longer need.
  - [x] `iam_members`: #1074 
  - [x] `iam_statements` **RFR** #1120

Note: there might already be statements using roles and that means there might be roles in statements that are not valid IDs, so just adding the constraint could cause migrations to fail.

solution: on migration, drop invalid roles if it's the only role in the statement and there are no actions then drop the statement entirely. we might also add a constraint that drops policies with empty statements

though it's not the nicest UX, we are in beta.

[...]

PR #878, opened 2019-07-12
[part of #832] use db_id between projects and roles

srenatus:

✔️ `iam_project_roles` now refers to `iam_projects` using the latter's `db_id` field, not `id`.

#832 

- [X] Code actually executed?

### DIFF
--- a/components/authz-service/storage/postgres/migration/sql/55_use_project_db_id_in_references_of_project_roles.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/55_use_project_db_id_in_references_of_project_roles.up.sql
@@ -40,4 +40,18 @@ CREATE OR REPLACE FUNCTION
 
 $$ LANGUAGE sql;
 
+CREATE FUNCTION
+  role_projects(_role_id TEXT)
+  RETURNS TEXT[] AS $$
+
+  SELECT COALESCE(array_agg(p.id) FILTER (WHERE p.id IS NOT NULL), '{(unassigned)}')
+  FROM iam_roles AS r
+  LEFT JOIN iam_role_projects AS rp
+  ON rp.role_id=r.db_id
+  LEFT JOIN iam_projects AS p
+  ON rp.project_id=p.db_id
+  WHERE r.id=_role_id
+
+$$ LANGUAGE sql;
+
 COMMIT;
--- a/components/authz-service/storage/postgres/migration/sql/55_use_project_db_id_in_references_of_project_roles.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/55_use_project_db_id_in_references_of_project_roles.up.sql
@@ -19,6 +19,29 @@ SET
 ALTER TABLE iam_role_projects
     DROP COLUMN project_temp_id;
 
+-- returns NULL if there's no associated projects; the call can decide
+-- (by coalesce'ing) if they want that to be {} or {(unassigned)}.
+CREATE OR REPLACE FUNCTION
+  role_projects(_role_id TEXT, OUT _project_ids TEXT[])
+  RETURNS TEXT[] AS $$
+
+  DECLARE
+  role_db_id INTEGER;
+  BEGIN
+      SELECT db_id INTO STRICT role_db_id
+      FROM iam_roles AS r
+      WHERE r.id=_role_id;
+      IF FOUND THEN
+          SELECT array_agg(p.id) INTO _project_ids
+          FROM iam_role_projects AS rp
+          LEFT JOIN iam_projects AS p
+          ON rp.project_id=p.db_id
+          WHERE rp.role_id=role_db_id;
+     END IF;
+  END;
+$$ LANGUAGE plpgsql;
+
+
 CREATE OR REPLACE FUNCTION
   query_role(_role_id TEXT)
   RETURNS json AS $$
@@ -28,29 +51,27 @@ CREATE OR REPLACE FUNCTION
     'name', r.name,
     'type', r.type,
     'actions', r.actions,
-    'projects', COALESCE(json_agg(p.id) FILTER (WHERE p.id IS NOT NULL), '[]')
+    'projects', COALESCE(role_projects(r.id), '{}')
     ) AS role
   FROM iam_roles AS r
-  LEFT JOIN iam_role_projects AS rp
-  ON rp.role_id=r.db_id
-  LEFT JOIN iam_projects AS p
-  ON rp.project_id=p.db_id
   WHERE r.id=_role_id
-  GROUP BY r.id, r.name, r.type, r.actions
 
 $$ LANGUAGE sql;
 
-CREATE FUNCTION
-  role_projects(_role_id TEXT)
-  RETURNS TEXT[] AS $$
+CREATE OR REPLACE FUNCTION
+  query_roles(_projects_filter TEXT[])
+  RETURNS setof json AS $$
 
-  SELECT COALESCE(array_agg(p.id) FILTER (WHERE p.id IS NOT NULL), '{(unassigned)}')
+  SELECT
+    json_build_object(
+      'id', r.id,
+      'name', r.name,
+      'type', r.type,
+      'actions', r.actions,
+      'projects', COALESCE(role_projects(r.id), '{}')
+    ) AS role
   FROM iam_roles AS r
-  LEFT JOIN iam_role_projects AS rp
-  ON rp.role_id=r.db_id
-  LEFT JOIN iam_projects AS p
-  ON rp.project_id=p.db_id
-  WHERE r.id=_role_id
+  WHERE projects_match(COALESCE(role_projects(r.id), '{(unassigned)}'),  _projects_filter);
 
 $$ LANGUAGE sql;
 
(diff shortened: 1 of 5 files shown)


components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -888,14 +882,8 @@ func checkIfRoleIntersectsProjectsFilter(ctx context.Context, q Querier,
...
-	// doesn't exist, this will return 0 rows which will bubble up to NotFoundErr when passed to processError.
-	row := q.QueryRowContext(ctx,
-		`SELECT COALESCE(array_agg(rp.project_id)
-				FILTER (WHERE rp.project_id IS NOT NULL), '{(unassigned)}') && $2 AS intersection
-			FROM iam_roles AS r
-			LEFT OUTER JOIN iam_role_projects AS rp ON rp.role_id=r.db_id
-			WHERE r.id = $1;`,
-		id, pq.Array(projectsFilter))
+	// doesn't exist, this will return a proper SQL "no rows" error when passed to processError.
+	row := q.QueryRowContext(ctx, "SELECT projects_match(role_projects($1), $2)", id, pq.Array(projectsFilter))
```
srenatus: 🤔 would make sense to coalesce this into `{(unassigned)}`, but the tests pass regardless...

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1372,7 +1358,7 @@ func (p *pg) CreateProject(ctx context.Context, project *v2.Project) (*v2.Projec
 	}
 
 	if project.Type == v2.Custom {
-		row := tx.QueryRowContext(ctx, ` WITH t as (SELECT id from iam_projects WHERE type='custom') SELECT COUNT(*) FROM t;`)
+		row := tx.QueryRowContext(ctx, "SELECT count(*) FROM iam_projects WHERE type='custom'")
```
srenatus: drive-by cleanup

components/authz-service/storage/postgres/migration/sql/55_use_project_db_id_in_references_of_project_roles.up.sql
```diff
@@ -40,4 +40,18 @@ CREATE OR REPLACE FUNCTION
 
 $$ LANGUAGE sql;
 
+CREATE FUNCTION
+  role_projects(_role_id TEXT)
+  RETURNS TEXT[] AS $$
+
+  SELECT COALESCE(array_agg(p.id) FILTER (WHERE p.id IS NOT NULL), '{(unassigned)}')
```
bcmdarroch: nifty 😋

components/authz-service/storage/postgres/migration/sql/55_use_project_db_id_in_references_of_project_roles.up.sql
```diff
@@ -0,0 +1,102 @@
...
+  BEGIN
+      SELECT db_id INTO STRICT role_db_id
+      FROM iam_roles AS r
+      WHERE r.id=_role_id;
+      IF FOUND THEN
+          SELECT array_agg(p.id) INTO _project_ids
+          FROM iam_role_projects AS rp
+          LEFT JOIN iam_projects AS p
+          ON rp.project_id=p.db_id
+          WHERE rp.role_id=role_db_id;
```
tylercloke: [nit] Some tabbing might make this a bit more readable.

components/authz-service/storage/postgres/migration/sql/55_use_project_db_id_in_references_of_project_roles.up.sql
```diff
@@ -0,0 +1,102 @@
...
+  role_projects(_role_id TEXT, OUT _project_ids TEXT[])
+  RETURNS TEXT[] AS $$
+
+  DECLARE
+  role_db_id INTEGER;
+  BEGIN
+      SELECT db_id INTO STRICT role_db_id
+      FROM iam_roles AS r
+      WHERE r.id=_role_id;
+      IF FOUND THEN
```
msorens: I am torn... a successful find sets `FOUND`, yes, but an unsuccessful find throws an exception. So you do not really need the `IF FOUND...`; it will work exactly the same without it.

However... on the one hand, it is more clear to have `IF FOUND...` because you know that's what the author really meant. But on the other hand, it makes one think (incorrectly) that a 'not found' case just falls through.
🤔 Not sure what is "better". What do you think about keeping the `IF FOUND` and just above the `END IF` add a comment `-- ELSE throw SQLException` ?

components/authz-service/storage/postgres/migration/sql/55_use_project_db_id_in_references_of_project_roles.up.sql
```diff
@@ -0,0 +1,102 @@
...
+          SELECT array_agg(p.id) INTO _project_ids
+          FROM iam_role_projects AS rp
+          LEFT JOIN iam_projects AS p
```
msorens: `LEFT JOIN` implies there is the possibility that there will be some null records in rp.project_id that does not have a corresponding p.db_id, but that should not be possible, so `INNER JOIN` would be more appropriate.

components/authz-service/storage/postgres/migration/sql/55_use_project_db_id_in_references_of_project_roles.up.sql
```diff
@@ -0,0 +1,102 @@
...
+      'actions', r.actions,
+      'projects', COALESCE(role_projects(r.id), '{}')
+    ) AS role
+  FROM iam_roles AS r
+  WHERE projects_match(COALESCE(role_projects(r.id), '{(unassigned)}'),  _projects_filter);
+
+$$ LANGUAGE sql;
+
+-- helper functions, now made strict: if they don't find anything,
+-- a "not found" exception is thrown
```
msorens: Should we be more complete with the comment? To the effect of "must match exactly one row; throws an exception if none (not found) or more than one (too many rows) are returned"

components/authz-service/storage/postgres/migration/sql/55_use_project_db_id_in_references_of_project_roles.up.sql
```diff
@@ -0,0 +1,102 @@
...
+            iam_projects
+        WHERE
+            id = t.project_temp_id);
+ALTER TABLE iam_role_projects
+    DROP COLUMN project_temp_id;
+
+-- returns NULL if there's no associated projects; the call can decide
+-- (by coalesce'ing) if they want that to be {} or {(unassigned)}.
+CREATE OR REPLACE FUNCTION
+  role_projects(_role_id TEXT, OUT _project_ids TEXT[])
```
msorens: Is there an advantage to having the `OUT` parameter instead of just returning a value?

components/authz-service/storage/postgres/migration/sql/55_use_project_db_id_in_references_of_project_roles.up.sql
srenatus: I've been thinking more about this one, too. At some point, I wanted it to return `{(unassigned)}`, but then I've reconsidered. And I think you're right, if we don't do that, we don't need the `IF FOUND` conditional. 👍

components/authz-service/storage/postgres/migration/sql/55_use_project_db_id_in_references_of_project_roles.up.sql
srenatus: We cannot `SELECT INTO` without a named parameter... we might not _strictly have to_ do this for the projects part of this, but I don't suppose it hurts 😉

components/authz-service/storage/postgres/migration/sql/55_use_project_db_id_in_references_of_project_roles.up.sql
```diff
@@ -0,0 +1,102 @@
...
+          LEFT JOIN iam_projects AS p
+          ON rp.project_id=p.db_id
+          WHERE rp.role_id=role_db_id;
+     END IF;
+  END;
+$$ LANGUAGE plpgsql;
+
+
+CREATE OR REPLACE FUNCTION
+  query_role(_role_id TEXT)
```
msorens: Just noticed: shouldn't this function also have the second parameter of `_projects_filter` ?

srenatus: If we wanted to unify how project filtering works, yes. However, I've considered that out of scope for this issue. Furthermore, I'm not convinced about what the best approach is just yet.

components/authz-service/storage/postgres/migration/sql/55_use_project_db_id_in_references_of_project_roles.up.sql
srenatus: I've given all my trust to `pg_format` 😉

srenatus:
ℹ️ will merge when green

components/authz-service/storage/postgres/migration/sql/55_use_project_db_id_in_references_of_project_roles.up.sql
```diff
@@ -0,0 +1,97 @@
...
+    SELECT
+        db_id INTO STRICT _db_id
+    FROM
+        iam_projects
+    WHERE
+        id = _id;
+END;
+$$
+LANGUAGE plpgsql;
+CREATE OR REPLACE FUNCTION project_id (_db_id iam_projects.db_id % TYPE, OUT _id iam_projects.id % TYPE)
```
blakestier: Are we using this anywhere? I can't seem to find it 🤔

msorens: I agree; don't see it being used anywhere. 🤷‍♂

PR #881, opened 2019-07-12
[cereal] Wire up workflow cancelation

jaym:
- Added a cancel workflow signal. Workflows can decide how to handle
this
- If a workflow finishes when outstanding tasks, the context for those
tasks will be canceled the next time they try to ping.
- I changed the task ping interval to 15 seconds so canceled tasks exit
earlier

### DIFF
--- a/lib/cereal/cereal.go
+++ b/lib/cereal/cereal.go
@@ -356,9 +356,6 @@ type WorkflowExecutor interface {
 	OnTaskComplete(w WorkflowInstance, ev TaskCompleteEvent) Decision
 
 	// OnCancel is called when a workflow instance is to be canceled.
-	//
-	// BUG(jaym): It's currently possible to receive a CancelEvent before
-	// OnStart for the postgres implementation.
 	OnCancel(w WorkflowInstance, ev CancelEvent) Decision
 }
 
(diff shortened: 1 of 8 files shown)


lib/cereal/postgres/task_pinger.go
```diff
@@ -9,28 +9,30 @@ import (
...
 
-func newTaskPinger(db *sql.DB, taskID int64) *taskPinger {
+func newTaskPinger(db *sql.DB, taskID int64, pingInterval time.Duration) *taskPinger {
 	pinger := &taskPinger{
-		taskID:          taskID,
-		db:              db,
-		checkinInterval: 30 * time.Second,
+		taskID:       taskID,
+		db:           db,
+		pingInterval: pingInterval,
```
ryancragun: :+1: on this name change

lib/cereal/cereal.go
```diff
@@ -619,37 +616,43 @@ LOOP:
...
-					logrus.WithError(err).Error("Failed to fail task completer")
-				}
-			}
-			err = taskCompleter.Succeed(jsonResults)
-			if err != nil {
-				logrus.WithError(err).Error("failed to mark task as successful")
+			logctx.WithError(err).Error("failed to mark task as failed")
+			return err
+		}
+	} else {
```
ryancragun: Do we need this `else`?

lib/cereal/integration/cancel_workflow_test.go
```diff
@@ -0,0 +1,69 @@
...
+	defer m.Stop()
+	err := m.EnqueueWorkflow(context.Background(), workflowName, instanceName, nil)
+	suite.Require().NoError(err, "Failed to enqueue workflow")
+	wgWorkflowStart.Wait()
+	wgTaskStart.Wait()
+	logrus.Debug("canceling workflow")
+	err = m.CancelWorkflow(context.Background(), workflowName, instanceName)
+	suite.Require().NoError(err, "Failed to cancel workflow")
+	wgTaskDone.Wait()
+	time.Sleep(2 * time.Second)
```
ryancragun: A hardcoded sleep in a test :(

lib/cereal/postgres/task_pinger.go
```diff
@@ -9,28 +9,30 @@ import (
...
 	}
 	pinger.wgStart.Add(1)
 	return pinger
 }
 
 // Start starts a goroutine that checks the task in on some interval.
 // This function is only allowed to be called once.
-func (w *taskPinger) Start(ctx context.Context) {
+func (w *taskPinger) Start(ctx context.Context, onTaskLost func()) {
```
ryancragun: Can we perhaps explain our intention with the onTaskLost func?

lib/cereal/cereal.go
jaym: i think so. the previous block doesn't return in all cases

lib/cereal/integration/cancel_workflow_test.go
jaym: you'll see this on all these tests. The sleep sort of makes sure things don't get unexpectedly called again. I'm open to ideas here. The postgres tests don't need this since they directly control all the flow. Here, we're relying on other machinery to run it.

Issue #690, opened 2019-06-26
Client Runs page suggestions missing cookbooks


lancewf:
# User Story

As an Automate User, I want all the cookbooks in all the nodes run lists to appear as suggestions for cookbooks in the Client Runs page's search bar. 

_The below image comes from a bank. It is showing the node details run list of a node that has the cookbook "db_chef_server_ha"._
<img width="322" alt="Screen Shot 2019-06-26 at 10 35 44 AM" src="https://user-images.githubusercontent.com/1679247/60205725-cee74700-9806-11e9-962c-dcd3ab3c64fa.png">

The problem is shown when the user goes to the Client Runs search bar and the "db_chef_server_ha" cookbook is not listed in the suggestions. 

The search bar's suggestions are filled from a CCR's "node.automatic.cookbooks" (https://github.com/chef/automate/blob/master/components/ingest-service/examples/converge-success-report.json#L2727).

The run list is filled from the "run_list" field on a CCR https://github.com/chef/automate/blob/master/components/ingest-service/examples/converge-success-report.json#L3035. 

We need to first ensure the CCR "node.automatic.cookbooks" field always contains cookbooks from the "run_list" field. If this is not the case then during ingestion we need to add the cookbooks from the run list to the list of cookbooks stored for each node. 

### References
Below is a link to the original issue. 
https://chefio.atlassian.net/browse/A2-893

# Definition of Done

Cookbooks from a nodes' run list are always suggestable in the Client Runs search bar. 

# Demo Script / Repro Steps

We do not know the exact problem, so we can not have reproducible steps. 
If a CCR's "node.automatic.cookbooks" does not always have all the cookbooks from the run list then the below steps will repro the problem. 
1. `start_all_services`
1. `send_chef_run_example`
1. Open https://a2-dev.test/client-runs
1. Search for Cookbook "insights-test"

lancewf:
I am having problems replicating this bug with cookbooks. I am seeing a similar problem with attributes. I am going to investigate the problem of the attributes because the bug might only occur when there are hundreds of cookbooks which I have not tested.

lancewf:
I have been able to replicate the bug with cookbooks by having over 105 cookbooks in a run_list. I also found where the problem is, this link https://github.com/chef/automate/blob/master/components/config-mgmt-service/backend/elastic/suggestions.go#L102 the ``SuggestionQuerySize`` is set to 100. If you increase this value to 200 cookbooks over 105 will be added to suggestions. 

[...]

PR #885, opened 2019-07-12
Missing Cookbook suggestions

lancewf:
When there are over 105 cookbooks (or any other array value) on one node the last cookbooks in alphabetic order are not added to the suggestions list. This prevents a user from filter from these missing cookbooks. 

1. Run automate with the current code `start_all_services`
1. `send_chef_run_example`
1. Open https://a2-dev.test/client-runs
1. Select the node and open the "Attributes" section. 
1. Select the "Default" filter and write down an attribute that is the closest to the end of the alphabetic that you can find (I found "yum").
1. Go back to the node list page (https://a2-dev.test/client-runs)
1. In the search bar select "Attribute" and type in the attribute, you found in the above step. 
1. No attributes should not be displayed in the suggestions. 
1. Run with the new code `rebuild components/config-mgmt-service`
1. On the node list page (https://a2-dev.test/client-runs) again type in the attribute, you found above and it should appear in the suggestions. 

https://github.com/chef/automate/issues/690

- [x] Code actually executed?


components/config-mgmt-service/backend/elastic/suggestions.go
```diff
@@ -21,6 +21,93 @@ func (es Backend) GetSuggestions(term string, text string, filters map[string][]
...
+// 										"boost":100,
+// 										"value":"kk"
+// 								 }
+// 							}
+// 					 }
+// 				]
+// 		 }
+// 	},
+// 	"size":0
+// }
```
lancewf: Adding this comment to make it easier to understand what the elasticsearch query is.

components/config-mgmt-service/backend/elastic/suggestions.go
```diff
@@ -100,6 +187,9 @@ func (es Backend) getArrayAggSuggestions(term string, text string, filters map[s
 	}
 	// multiplying the size by 10 as elasticsearch will sort array aggregations by doc_count. Will trim it back to size once we match it again in go
 	aggs := elastic.NewTermsAggregation().Field(term).Size(SuggestionQuerySize)
+	if len(text) >= 2 {
+		aggs = aggs.Include(".*" + text + ".*")
+	}
```
lancewf: If the search box has two or more characters typed, then filter the aggregation with that value. Before when the aggregation was not filtered, the first 100 cookbooks would be used and if there were over 100 cookbooks then the last cookbooks would never be returned.

components/config-mgmt-service/integration_test/suggestions_test.go
```diff
@@ -44,6 +44,280 @@ func TestSuggestionsWithAnInvalidTypeReturnsError(t *testing.T) {
 	}
 }
 
+func TestSuggestionsLargeArrayValues(t *testing.T) {
```
lancewf: This test was failing before the above change.

vjeffrey:
haven't reviewed the code yet, but pulled it down and followed the instructions and it all worked beautifully! ⭐️

components/config-mgmt-service/backend/elastic/suggestions.go
```diff
@@ -90,16 +177,20 @@ func (es Backend) getArrayAggSuggestions(term string, text string, filters map[s
 	typeQuery := elastic.NewTypeQuery(IndexNodeState)
 	filters["exists"] = []string{"true"}
 	boolQuery := newBoolQueryFromFilters(filters)
+	textLower := strings.ToLower(text)
```
lancewf: Terms are stored as lower case in elasticsearch.

components/config-mgmt-service/integration_test/suggestions_test.go
```diff
@@ -44,6 +44,355 @@ func TestSuggestionsWithAnInvalidTypeReturnsError(t *testing.T) {
...
+		"yum.epel-debuginfo.managed",
+		"yum.main.logfile",
+		"runit.executable",
+		"windows.allow_reboot_on_failure",
+		"nginx.set_misc.version",
+		"nginx.passenger.packages.debian",
+		"nginx.auth_request.url",
+		"yum.epel-testing-debuginfo.make_cache",
+		"zum.epel-debuginfo.gpgcheck",
+	}
```
phiggins: You could save lines of code and make this easier to read/understand (maybe?) if you looped the number of times and built strings like `z1`, `z2`, etc. I'm thinking it'd maybe be easier to understand a failure if you got back just the wrong string rather than an arbitrary one from your collection.

components/config-mgmt-service/integration_test/suggestions_test.go
```diff
@@ -44,6 +44,355 @@ func TestSuggestionsWithAnInvalidTypeReturnsError(t *testing.T) {
...
+			"zum.set_misc.version",
+			"zum.passenger.packages.debian",
+			"zum.auth_request.url",
+			"zum.epel-testing-debuginfo.make_cache",
+			"zum.epel-debuginfo.gpgcheck",
+			"zum.5",
+			"zum.6",
+			"zum.7",
+			"zum.9",
+			"zum",
```
phiggins: Same concern here about the word list.

components/config-mgmt-service/integration_test/suggestions_test.go
```diff
@@ -44,6 +44,355 @@ func TestSuggestionsWithAnInvalidTypeReturnsError(t *testing.T) {
...
+	}
+
+	suite.IngestNodes([]iBackend.Node{node})
+	defer suite.DeleteAllDocuments()
+	res, err := cfgmgmt.GetSuggestions(context.Background(), &request)
+	assert.Nil(t, err)
+
+	actualSuggestions := extractTextFromSuggestionsResponse(res, t)
+
+	assert.True(t, contains(actualSuggestions, "zum"), "Suggestions does not contain 'zum'")
```
phiggins: There is an `assert.Contains` function that might be useful here: https://godoc.org/github.com/stretchr/testify/assert#Contains

Issue #853, opened 2019-07-10
re-architect authz-service


susanev:
## Overview

authz-service has grown into a fuzzy thing that becomes harder to understand with every kludge we add.

### Examples
* the service constructors are difficult to instantiate – leading to manual instantiations, and copied-around, diverging interceptor setups: [server/v2/project_test.go](https://github.com/chef/automate/blob/6f4f5c482549dc615a99f68d3f5e87626310b1c8/components/authz-service/server/v2/project_test.go#L400), [server/v2/policy_test.go](https://github.com/chef/automate/blob/99e34252c5492e0730306207f5e56ec435446ab1/components/authz-service/server/v2/policy_test.go#L2973)
* the version of IAM used is known to the main server struct, but its components aren't aware of this, leading to a multiple ways of threading through the IAM version (for example, the authz v2 grpc handler has a way of knowing which version is used through the versionswitch construct, which allows the authz grpc handler to be used as interceptor; the method for updating the engine store is different; the internal v2 and v2.1 GRPC endpoints are the same, so that component just doesn't care) – we should find a way to make all this simple and stupid.
* as a side-effect of this, the opa engine exposes [quite odd methods](https://github.com/chef/automate/blob/master/components/authz-service/engine/engine.go), which almost all do the same thing (with a recent PR, it's mostly 3 variants of each of SetPolicies, IsAuthorized, and Introspection)

## Acceptance Criteria

* There are no GRPC middleware setup in any of the test files – it’s an internal concern of the service constructor that they’re using.
* The IAM version used is represented in the service state – the “version switch” interceptor is reading it from there, and nothing is reading the version switch’s Version field.
* There’s no goroutine+channel construct used anywhere to bypass any odd “who knows about what” issues.

### Subtasks

- [ ] remove goroutine+channel handling version changes and instead the service state always knows the version
- [ ] decide and document the shape of the tests we want for all levels of testing (unit, property, integration at service level, integration at app level)
- [ ] use service constructors in tests
- [ ] refactor OPA methods to be more clear and DRY
- [ ] break up giant postgres files and refactor where possible

susanev:
this work has been deprioritized, so closing.

PR #886, opened 2019-07-12
Handle empty statement projects for v2 and v2.1

msorens:

In 2.1 policy statements must have a non-empty project.
Any policies created in v1, when migrated, will get "all projects" (`*`) assigned during migration.
Any policies created in v2.0, however, do not go through any migration. Therefore they must be created with the same "all projects" (`*`) silently appended if no projects are specified.

Bug #1: Creating a policy in v2.0 has an empty projects list on statements.
Bug #2: Creating a policy in v2.1 with no statement projects generates an invalid policy, i.e. a policy with no projects.

### :+1: Definition of Done
1. V2.0 policies with an empty projects list are assigned the default "all projects" (`*`).
2. Attempt to create v2.1 policies with an empty projects list are rejected.

In v2.0, create a policy with no statement projects, e.g. :
```
curl -sSkH "api-token: $TOK" -d \
'{"id":"team-view-pol", "name":"team viewer", "members":["user:local:coolbean"], \
"statements": [{"role": "team-viewer"}]}' $TARGET_HOST/apis/iam/v2beta/policies
```
Read it back and observe it has the default `*` assigned. Here's one way to do that:
```
authQuery -o -2 policies id team-view-pol | jq
```
Of course, if you explicitly include all projects in the statement (`projects: ["*"]`), that should return an identical result.

In v2.1, if you attempt to create a policy with no statement projects you will now receive an error, e.g.:
```
{
  "error": "error parsing policy \"team-view-pol3\": parse statement: project statements must include projects",
  "message": "error parsing policy \"team-view-pol3\": parse statement: project statements must include projects",
  "code": 3,
  "details": []
}
```

- [x] Code actually executed?


components/authz-service/server/v2/policy.go
```diff
@@ -48,9 +50,10 @@ func NewMemstorePolicyServer(
 	l logger.Logger,
 	e engine.V2pXWriter,
 	pl storage_v1.PoliciesLister,
+	vSwitch *VersionSwitch,
 	vChan chan api.Version) (PolicyServer, PolicyRefresher, error) {
 
-	return NewPoliciesServer(ctx, l, memstore.New(), e, pl, vChan)
+	return NewPoliciesServer(ctx, l, memstore.New(), e, pl, vSwitch, vChan)
```
msorens: Of course, vSwitch and vChan are related but they serve different functions, so I thought it best to keep them cleanly separate.

components/authz-service/server/v2/policy.go
```diff
@@ -859,8 +869,18 @@ func statementFromAPI(statement *api.Statement) (storage.Statement, error) {
 		statement.Resources = []string{"*"}
 	}
 
-	// map external representation of "all projects" to actual ID for that meta-project
 	projects := make([]string, len(statement.Projects))
+
+	if len(statement.Projects) == 0 {
+		if s.isBeta2p1() {
```
msorens: Everything above here was to be able to have this function exist--now we can distinguish v2/v2.1 and take appropriate action.

components/authz-service/server/v2/policy_test.go
```diff
@@ -467,6 +469,159 @@ func TestCreatePolicy(t *testing.T) {
 	}
 }
 
+func TestCreatePolicy2p1(t *testing.T) {
```
msorens: Add some explicit 2.1 tests where it matters: create and update with no projects.

components/authz-service/server/v2/policy_test.go
```diff
@@ -69,7 +69,7 @@ func TestCreatePolicy(t *testing.T) {
 			assert.Equal(t, []string{"cfgmgmt:nodes:*"}, resp.Statements[0].Actions)
 			assert.Equal(t, []string{"cfgmgmt:delete", "cfgmgmt:list"}, resp.Statements[0].Resources)
 			assert.Empty(t, resp.Statements[0].Role)
-			assert.Empty(t, resp.Statements[0].Projects)
+			assert.Equal(t, []string{constants_v2.AllProjectsExternalID}, resp.Statements[0].Projects)
```
msorens: This affirms the change that no projects provided in 2.0 now is assigned a default "all projects".

components/authz-service/server/v2/policy_test.go
```diff
@@ -919,19 +1074,17 @@ func TestUpdatePolicy(t *testing.T) {
 		}},
 		{"successfully updates policy statements", func(t *testing.T) {
 			storedPol, _ := addSomePoliciesToStore(t, store, prng)
-			// change original statement effect to allow
```
msorens: Comments in this test were outdated (read: wrong)... I repurposed the test to handle both explicit and implicit "all" cases.

components/authz-service/server/v2/policy_test.go
```diff
@@ -943,15 +1096,19 @@ func TestUpdatePolicy(t *testing.T) {
 
 			require.NoError(t, err)
 
+			// These individual statement checks are included in the final assertPoliciesMatch,
+			// but useful to check these key elements explicitly.
 			s0 := pol.Statements[0]
 			assert.Equal(t, api_v2.Statement_ALLOW, s0.Effect)
 			assert.Equal(t, []string{"compliance:profiles"}, s0.Resources)
 			assert.Equal(t, []string{"compliance:profiles:upload"}, s0.Actions)
+			assert.Equal(t, []string{constants_v2.AllProjectsExternalID}, s0.Projects)
```
msorens: Implicit or explicit look the same in the end.

components/authz-service/server/v2/policy_test.go
```diff
@@ -2870,6 +3027,13 @@ func getPolicyFromStore(t *testing.T, store *cache.Cache, id string) storage.Pol
 	return *pol
 }
 
+func assertProjectsMatch(t *testing.T, storageProject string, apiProject string) {
+	if storageProject == constants_v2.AllProjectsID {
+		storageProject = constants_v2.AllProjectsExternalID
+	}
+	assert.Equal(t, storageProject, apiProject, "statement projects differ")
+}
```
msorens: Needed just this one little tweak to handle the internal/external representation differences.

components/authz-service/server/v2/policy_test.go
```diff
@@ -2938,11 +3104,17 @@ type testSetup struct {
...
+		// Returning MigrationStatus of "Success" means we've migrated successfully to IAM v2
 		func(s storage.MigrationStatusProvider) error { return s.Success(context.Background()) })
 }
 
+func setupV2p1WithWriter(t *testing.T,
+	writer engine.V2pXWriter) testSetup {
+	return setupV2WithMigrationState(t, nil, writer, nil, nil, make(chan api_v2.Version, 1),
+		// Returning MigrationStatus of "SuccessBeta2.1" means we've migrated successfully to IAM v2.1
+		func(s storage.MigrationStatusProvider) error { return s.SuccessBeta1(context.Background()) })
+}
```
msorens: Thanks, @srenatus, for the framework that made adding v2.1 just plug-n-play, as it were!

components/authz-service/server/server.go
```diff
@@ -73,8 +73,8 @@ func NewGRPCServer(ctx context.Context,
 
 	// TODO (tc) Refactor how singletons are shared between GRPC servers. Should we also
 	// be sharing a single v2 store / pg instance?
-	v2PolServer, policyRefresher, err := v2.NewPostgresPolicyServer(ctx, l, e, migrationsConfig,
-		dataMigrationsConfig, v1Server.Storage(), vChan)
+	v2PolServer, policyRefresher, err := v2.NewPostgresPolicyServer(
+		ctx, l, e, migrationsConfig, dataMigrationsConfig, v1Server.Storage(), switcher, vChan)
```
blakestier: New addition is the `switcher` (line change gave me pause)

components/authz-service/server/v2/policy.go
```diff
@@ -140,7 +146,7 @@ func (s *policyServer) CreatePolicy(
 
 	// API requests always create custom policies.
 
-	pol, err := policyFromAPI(
+	pol, err := s.policyFromAPI(
```
blakestier: I'm curious why the change to have this be a function on `server` . . . maybe it's too late for me to be reviewing? 🙃

msorens: No, that's a fair question! It is (almost) turtles all the way down. In other words, at the bottom, we need to access the `vSwitch` which is on `s`. Follow that back up to here: `s.vSwitch` => s.isBeta2p1 => s.statementFromAPI => s.policyFromAPI.

components/automate-cli/pkg/diagnostics/integration/iam_v2.go
```diff
@@ -106,6 +112,26 @@ func CreateIAMV2Diagnostic() diagnostics.Diagnostic {
...
+	}
+
+	resp, err := apiClient.PoliciesClient().GetPolicyVersion(ctx, &policies_req.GetPolicyVersionReq{})
+	if err != nil {
+		return 0, status.Wrap(err, status.APIError, "Failed to retrieve IAM version")
+	}
+	return resp.Version.Major, nil
+}
+
 func init() {
```
srenatus: I'm afraid `init()` is rarely the correct move 😉 
I'd propse we do the following:
- add an `iam-v2` tag to diagnostic tests that require IAM v2 (here: https://github.com/chef/automate/pull/886/files#diff-78d181d2f500dd5af4a70ec53130c49cR52)
- use diagnostic filters for these, like https://github.com/chef/automate/blob/master/integration/tests/upgrade_current_master.sh#L4
- upgrade-to-v2 in the `integration/tests/*.sh` scripts that run the iam v2 diagnostic tests

components/authz-service/server/v2/policy_test.go
bcmdarroch: If the `implicit` way (as in, no projects in the statement means it applies to all projects) still works like before, why do we need to start enforcing projects on all v2.1 statements?

msorens: But it has never been legal to have a statement with no projects.

srenatus:
This failure is **unrelated to diagnostics**: https://buildkite.com/chef/chef-automate-master-verify-private/builds/3158#9da55f3b-f5ce-4db4-8e61-f8f52a268eea/579-1323

components/authz-service/server/v2/policy_test.go
bcmdarroch: Do you mean according to the acceptance requirements?

components/authz-service/server/v2/policy.go
```diff
@@ -902,3 +922,8 @@ func (s *policyServer) setVersionForInterceptorSwitch(v api.Version) {
 		s.vChan <- v
 	}
 }
+
+func (s *policyServer) isBeta2p1() bool {
```
srenatus: This falls apart when we're using IAM v2 endpoints with authz-service running on v1. This is what our inspec tests have been doing for a while now.

srenatus:
Team: If you look at the tests changes, you'll note that for v2, this will now include `"*"` in the statements' bodies: [a2-iam-v2-integration/iam_v2](https://github.com/chef/automate/blob/ms/empty-projects-fix/inspec/a2-iam-v2-integration/controls/iam_v2.rb#L74-L99).

It's a by-product of this approach to resolving the issue. I guess we could fiddle around a bit to remove this again if you're not on v2.1; but we already have accepted that there's `projects` keys in the return messages... so, adding this doesn't seem too bad.

What I find confusing, however, is that only the statements' `projects` is set. 🤔

srenatus:
Also, with disabling the iam-v1 token diagnostic for the iam v2 run, something has become apparent (to me): We have no upgrade test safety whatsoever for iam v2 or iam v2.1 from our diagnostics framework. I think I'll duplicate the upgrade tests...?

components/authz-service/server/v2/policy.go
srenatus: `vSwitch` is two things:
- an UnaryServerInterceptor, denying v1 reqs when using v2 and vice-versa (this PreconditionFailed response drives the gateway IAM version switching)
- a struct keeping an api.Version

It would be better, I think, as outlined in #853, to have a version field in the server struct, and have that server implement the Interceptor.

`vChan` is a channel we're using to flip the version in that struct; based on a successful migration (upgrade-to-v2) *or* read on server startup (ie. when it's already been switched to IAM v2). It's a **shortcut**; it shouldn't be needed. We're creating a portal between two pieces of the service (the migration logic and the policies server), and have the former throw something through it at the latter. It's not how things should be, this is not something inherently calling for the use of a channel.

components/automate-cli/pkg/diagnostics/integration/iam_v2.go
srenatus: ✔️ we've squashed that away now.

srenatus:
Still working on getting the added diagnostics test to _fail properly_, and then to fix it.

components/authz-service/server/v2/policy.go
blakestier: 🤔 Is that a smell? 🤷‍♂

components/authz-service/storage/postgres/datamigration/sql/09_make_projects_wildcard_an_actual_project.up.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
+INSERT INTO iam_projects (id, name, type) VALUES ('*', 'Wildcard project', 'chef-managed');
```
blakestier: we do have the ~~ALL PROJECTS~~ project . . . cannot recall why we didn't use `*` all the way down . . .

components/automate-cli/pkg/diagnostics/integration/auth_policies.go
```diff
@@ -20,7 +20,7 @@ type authPoliciesSave struct {
 func CreateAuthPoliciesDiagnostic() diagnostics.Diagnostic {
 	return diagnostics.Diagnostic{
 		Name: "auth-policies",
-		Tags: diagnostics.Tags{"auth"},
+		Tags: diagnostics.Tags{"auth", "iam-v1"},
```
blakestier: Oof shouldn't we have a test where we *do* migrate the v1 policies? 😩

srenatus: This is not the v1-to-v2 migration, this is used to (among other things) test an actual upgrade of the A2 version installed.

Also, we've got "did these policies get migrated?" tests; but they only assert that a migrated policy is there, I don't think they assert that they're _working_. *However*, implicitly, I believe this is tested for admin tokens 🤔

integration/tests/iam_v2_upgrade_diagnostics.sh
```diff
@@ -10,8 +10,8 @@ test_upgrades=true
 # do_test_deploy runs _before_ the upgrade, so we upgrade _from a state with
 # IAMv2 enabled_, and the diagnostics are run twice: before and after upgrade.
 do_test_deploy() {
-    log_info "run chef-automate iam upgrade-to-v2 --skip-policy-migration"
-    chef-automate iam upgrade-to-v2 --skip-policy-migration || return 1
+    log_info "run chef-automate iam upgrade-to-v2"
```
blakestier: oh this is what i meant in my prev comment 😂

components/authz-service/storage/postgres/datamigration/sql/09_make_projects_wildcard_an_actual_project.up.sql
blakestier: (I'm not clear on why we need this migration)

components/automate-deployment/pkg/server/api_token.go
```diff
@@ -71,6 +75,21 @@ func generateAdminToken(ctx context.Context,
 		Subjects: []string{fmt.Sprintf("token:%s", response.Id)},
 		Resource: "*",
 	})
+	if err != nil {
```
blakestier: should this just make the above token a member of the admin policy? does that exist yet? 🤔

components/authz-service/storage/postgres/datamigration/sql/09_make_projects_wildcard_an_actual_project.up.sql
srenatus: Frankly, I've made @msorens' approach work. I don't know! 💦

srenatus:
@blakestier as per discussion in the deployment-team channel -- since we're testing upgrades here, too (trying to do a much as we can reasonably do in one test), the first deployment is happening with code in `dev`. *So*, our changes to the deployment-service (through diagnostics) aren't there when we need them. I've thus split them off into their own PR: #946. When that's merged, we can rebase this one, and see where we get with it.

components/authz-service/storage/postgres/datamigration/sql/09_make_projects_wildcard_an_actual_project.up.sql
blakestier: Er, something is amiss 🤔

components/authz-service/server/v2/policy.go
```diff
@@ -859,8 +869,18 @@ func statementFromAPI(statement *api.Statement) (storage.Statement, error) {
...
+		if s.isBeta2p1() {
+			// no projects is an error for v2.1
+			return v2.Statement{}, errors.New("policy statements must include projects")
+		}
+		// no projects OK for v2.0 but set to "all" to allow 2.1 upgrade to work when needed
+		projects = append(projects, constants.AllProjectsID)
+	}
+
+	// map external representation of "all projects" to actual ID for that meta-project
 	for i, project := range statement.Projects {
```
blakestier: I wonder if we now should be iterating over `projects` and not `statement.Projects` . . . in which case, we'd want to append the `AllProjectsExternalID`

components/authz-service/storage/postgres/datamigration/sql/09_make_projects_wildcard_an_actual_project.up.sql
blakestier: https://github.com/chef/automate/pull/886/files#r304990794 maybe?

bcmdarroch: I'm also confused. [This migration](https://github.com/chef/automate/blob/78ca2f0114d92ccdba09bc2c627fa9afd010460d/components/authz-service/storage/postgres/datamigration/sql/04_v2_set_projects_default.up.sql#L12) makes me think we're already populating empty statements with `~~All Projects~~`

integration/tests/iam_v2_upgrade_diagnostics.sh
```diff
@@ -0,0 +1,25 @@
...
+test_diagnostics_filters="~iam-v1"
+test_upgrades=true
+
+# do_test_deploy runs _before_ the upgrade, so we upgrade _from a state with
+# IAMv2 enabled_, and the diagnostics are run twice: before and after upgrade.
+do_test_deploy() {
+    log_info "run chef-automate iam upgrade-to-v2 --skip-policy-migration"
+    chef-automate iam upgrade-to-v2 --skip-policy-migration || return 1
+
+    # ensure service startup works with IAM v2:
```
tylercloke: 😍

PR #852, opened 2019-07-10
Add RPC func and CLI cmd for disconnected services

afiune:
### :nut_and_bolt: Description
CLI command & RPC function to list disconnected services.

### :+1: Definition of Done
Users can list services that haven't received habitat events in a period of time. (threshold)

### :athletic_shoe: Demo Script / Repro Steps
- Start the studio
- Build and start the applications-service
- Populate the database or ingest a single message 
- (wait 5 min) Use the new CLI command to list disconnected services
```
chef-automate applications list-disconnected-services --threshold-minutes 5
```

#### When there are no disconnected services:
```
[45][default:/src:0]# /bin/chef-automate applications list-disconnected-services -m 20
There are no disconnected services with a threshold of 20 minute(s)
```

### CLI Display Format (default: json)
```
[39][default:/src:0]# /bin/chef-automate applications list-disconnected-services  --format json | jq .
{
  "services": [
    {
      "supervisor_id": "a24632dd-61fe-4a31-96e5-efb461a8f64a",
      "release": "core/app/0.1.0/20101212010101",
      "group": "app.default",
      "application": "demo",
      "environment": "demo",
      "fqdn": "app.example.com",
      "site": "eu",
      "previous_health_check": "NONE",
      "current_health_since": "22 minutes 59 seconds",
      "health_updated_at": "2019-07-11T07:27:50.033585Z"
    }
  ]
}
[40][default:/src:0]# /bin/chef-automate applications list-disconnected-services  -f pretty
services: <
  supervisor_id: "a24632dd-61fe-4a31-96e5-efb461a8f64a"
  release: "core/app/0.1.0/20101212010101"
  group: "app.default"
  application: "demo"
  environment: "demo"
  fqdn: "app.example.com"
  site: "eu"
  previous_health_check: NONE
  current_health_since: "23 minutes 10 seconds"
  health_updated_at: <
    seconds: 1562830070
    nanos: 33585000
  >
>
```
### RPC Function
```
[41][default:/src:0]# applications_grpcurl chef.automate.api.applications.ApplicationsService.GetDisconnectedServices -d '{"threshold_minutes":10}'
{
  "services": [
    {
      "supervisorId": "a24632dd-61fe-4a31-96e5-efb461a8f64a",
      "release": "core/app/0.1.0/20101212010101",
      "group": "app.default",
      "application": "demo",
      "environment": "demo",
      "fqdn": "app.example.com",
      "site": "eu",
      "previousHealthCheck": "NONE",
[...]

PR #889, opened 2019-07-15
Use convention name for context variables

afiune:
As per recommendation https://github.com/chef/automate/pull/852#discussion_r302424873


srenatus:
Thank you! 🎈

PR #890, opened 2019-07-15
Manual conversion of protobufs in gateway

afiune:
Currently every **"conversion"** done by `proto. CallDomainService()` does a multi conversion from:
```
Protobuf => JSON => Protobuf
```

The icing of the cake is that **[grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway)** does already a conversion from `JSON => Protobuf`, and therefore, we end up with:
```
JSON => Protobuf => JSON => Protobuf
```

tenor-74810411

### :+1: Definition of Done
Reduction of complexity and increase the speed of RPC functions in the gateway (the ones that belong to the `applications-service`) by avoiding unnecessary conversions.

List of functions:
- [x] GetServiceGroupsHealthCounts
- [x] GetServiceGroups
- [x] GetServices
- [x] GetServicesBySG
- [x] GetServicesStats
- [x] GetDisconnectedServices
- [x] GetVersion

N/A - All the code should compile and integrations tests should pass as before.

N/A - **Chore**

- [x] Code actually executed?

### DIFF
--- a/components/automate-gateway/handler/applications.go
+++ b/components/automate-gateway/handler/applications.go
@@ -52,18 +52,14 @@ func (a *Applications) GetServiceGroups(
 // GetServices returns a list of services
 func (a *Applications) GetServices(
 	ctx context.Context,
-	in *applications.ServicesReq) (*applications.ServicesRes, error) {
+	request *applications.ServicesReq) (*applications.ServicesRes, error) {
 
-	inDomain := &applications.ServicesReq{}
-	out := &applications.ServicesRes{}
-	f := func() (proto.Message, error) {
-		return a.client.GetServices(ctx, inDomain)
-	}
-	err := protobuf.CallDomainService(in, inDomain, f, out)
-	if err != nil {
-		return nil, err
-	}
-	return out, nil
+	log.WithFields(log.Fields{
+		"request": request.String(),
+		"func":    nameOfFunc(),
+	}).Debug("rpc call")
+
+	return a.client.GetServices(ctx, request)
 }
 
 // GetServicesBySG returns a list of services within a service-group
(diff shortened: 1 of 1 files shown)


components/automate-gateway/handler/applications.go
```diff
@@ -24,18 +26,24 @@ func NewApplicationsHandler(applicationsClient applications.ApplicationsServiceC
...
-	in *applications.ServiceGroupsHealthCountsReq) (*applications.HealthCounts, error) {
+	request *applications.ServiceGroupsHealthCountsReq) (*applications.HealthCounts, error) {
 
-	inDomain := &applications.ServiceGroupsHealthCountsReq{}
-	out := &applications.HealthCounts{}
-	f := func() (proto.Message, error) {
-		return a.client.GetServiceGroupsHealthCounts(ctx, inDomain)
-	}
-	err := protobuf.CallDomainService(in, inDomain, f, out)
+	log.WithFields(log.Fields{
```
srenatus: We shouldn't need this in the gateway: any GRPC call will be logged if
- its status is not "OK", or
- we're in debug mode

[see here for the logging middleware doing that](https://github.com/chef/automate/blob/a9aa5bf7f586090a532b71947e546446c64182e4/components/automate-gateway/gateway/server.go#L267-L287)

components/automate-gateway/handler/applications.go
```diff
@@ -24,18 +26,24 @@ func NewApplicationsHandler(applicationsClient applications.ApplicationsServiceC
...
 		return nil, err
 	}
-	return out, nil
+	return &applications.HealthCounts{
+		Total:    response.Total,
+		Ok:       response.Ok,
+		Warning:  response.Warning,
+		Critical: response.Critical,
+		Unknown:  response.Unknown,
+	}, nil
```
srenatus: I believe under certain circumstances, the compiler will do this for us: you could try
```go
return (*applications.HealthCounts)(response), nil
```
It's only allowed if the field types match, I believe. (I.e. if the compiler lets you do it, it's fine 😄 )

afiune: awesome! yes, the field types match, if fact, they are the same so I just found a better way to do this. 💯

Issue #405, opened 2019-05-22
convert compliance searchbar to the same format as client runs searchbar


vjeffrey:
# User Story

In an effort to drive parity across the app, we'd like to make the compliance searchbar be the same as the client runs searchbar.  This should be mostly ui work.

# Definition of Done
- compliance searchbar is same format as client runs one

PR #891, opened 2019-07-15
Implemented new compliance searchbar

vinay033:

In an effort to drive parity across the app, we'd like to make the compliance search bar be the same as the client runs search bar.
### :+1: Definition of Done
Compliance search bar is the same format as the client runs.
Screenshot from 2019-07-15 14-46-45
Screenshot from 2019-07-15 14-47-04

https://github.com/chef/automate/issues/405


vinay033:
> @vinay033 the placeholder text color and border color appear to be different colors on the client runs vs compliance search bars. Any chance you can align those to be the same color? if so, the compliance colors are the ones we would like to use. If it makes more sense for that change to be a separate pr then feel free to ignore this comment.
> 
> Otherwise, looks and works great, thank you!!

I have added changes for this, please have a look.

PR #894, opened 2019-07-15
[tls] Configure various nginx instances to use TLS 1.3 as well

stevendanna:
I'm unsure why this was previously only allowing TLS 1.2, seems like
we want to allow 1.3 if the client has it available.


srenatus:
Are there clients that we don't control? If not, we might as well `s/1.2/1.3/g`?

stevendanna:
> Are there clients that we don't control? If not, we might as well s/1.2/1.3/g?

For automate-load-balancer we definitely don't control the client. For most of the others, we do control the client, but our clients include Ruby, Go, Erlang, Elixir and other C applications that I don't think I'll be able to test in the near term.

PR #899, opened 2019-07-15
[lib/cereal] Handle error cases in scheduler code

stevendanna:
We had two cases with some TODOs around error handling in this
code.

1. We now attempt to disable scheduled workflows with invalid
   recurrence rules.

2. I've attempted to comment the existing code to indicate why we
   think the current error & sleep logic is correct. Namely, we only
   sleep when we think that not sleeping would result in an immediate
   retry of the same schedule workflow.

### DIFF
--- a/lib/cereal/workflow_scheduler.go
+++ b/lib/cereal/workflow_scheduler.go
@@ -37,7 +37,7 @@ func (w *workflowScheduler) run(ctx context.Context) {
 			if err != nil {
 				logrus.WithError(err).Error("failed to schedule workflows")
 			}
-			logrus.Infof("Recurring workflow scheduler sleep for %fs", nextSleep.Seconds())
+			logrus.Debugf("Recurring workflow scheduler sleep for %fs", nextSleep.Seconds())
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


lib/cereal/workflow_scheduler.go
```diff
@@ -86,36 +82,45 @@ func (w *workflowScheduler) scheduleWorkflow(ctx context.Context) (time.Duration
...
-	// rule types here to support the different use cases.
 	recurrence, err := rrule.StrToRRule(s.Recurrence)
 	if err != nil {
-		// TODO(ssd) 2019-05-13: Perhaps we should disable this rule so that it doesn't keep producing errors
-		// We need to do this otherwise we'll just keep trying to enqueue this and nothing else
-		return maxWakeupInterval, errors.Wrap(err, "could not parse recurrence rule for workflow, skipping")
+		logrus.WithError(err).Error("scheduled workflow has invalid recurrence rule, attempting to disable it")
+		err2 := completer.DisableSchedule(s)
+		if err2 != nil {
+			return maxWakeupInterval, errors.Wrap(err, "failed to disable workflow with invalid recurrence rule")
```
jaym: err2?

stevendanna: I've gone back and forth, hence the confusion at this site at the moment.  Since we are logging the first error at 87, let's return err2 here but not log it since the returned error will result in a log line.

PR #901, opened 2019-07-15
[lib/cereal] use sequence for task_results table

jaym:
task_results generates its own task_results id instead of relying
on the one from tasks. There is now a special field that ties together a
task result and task (in case we ever need it for debugging)

### DIFF
--- a/lib/cereal/postgres/sql_schema.go
+++ b/lib/cereal/postgres/sql_schema.go
@@ -119,6 +119,7 @@ CREATE TABLE cereal_tasks (
 
 CREATE TABLE cereal_task_results (
     id BIGSERIAL PRIMARY KEY,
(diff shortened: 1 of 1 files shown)


lib/cereal/postgres/sql_schema.go
```diff
@@ -119,6 +119,7 @@ CREATE TABLE cereal_tasks (
 
 CREATE TABLE cereal_task_results (
     id BIGSERIAL PRIMARY KEY,
+    task_id BIGINT NOT NULL,
```
stevendanna: Should we also add a unique constraint?

jaym: i think i like that. i'm guessing we wouldn't need an index since this table should be pretty small?

PR #903, opened 2019-07-15
[automate-cli] Restore the latest backup if no backup ID is given

ryancragun:
Attempt to restore the latest complete backup if no backup ID is given

### DIFF
--- a/components/automate-cli/cmd/chef-automate/backup.go
+++ b/components/automate-cli/cmd/chef-automate/backup.go
@@ -8,6 +8,7 @@ import (
 	"os/user"
 	"path"
 	"path/filepath"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


components/automate-cli/cmd/chef-automate/backup.go
```diff
@@ -577,6 +578,38 @@ will modify the existing data to match the contents of the backup.
...
+		}
+
+		parsedBackups = append(parsedBackups, parsed{task: b, date: date})
+	}
+
+	sort.Slice(parsedBackups, func(i, j int) bool {
+		return parsedBackups[i].date.After(parsedBackups[j].date)
+	})
+
+	return parsedBackups[0].task, nil
```
ryancragun: Need to handle out of range access here in case there are no completed backups

PR #906, opened 2019-07-16
Separate storage client and postgres implementation

afiune:
Creates a new `composedService` struct to separating the `storage` go
package that holds **only the client interface and its structs** that define
what a storage client should look like, the actual `postgres` implementation 
should NOT have a dependency on `db:"XYZ"` tags.

### :+1: Definition of Done
The `storage` and `postgres` go packages are decoupled.

N/A - Everything should work after this change is merged, build and run tests as usual.

N/A

- [x] Code actually executed?

### DIFF
--- a/components/applications-service/pkg/storage/postgres/service_internal_test.go
+++ b/components/applications-service/pkg/storage/postgres/service_internal_test.go
@@ -4,8 +4,38 @@ import (
 	"regexp"
 	"testing"
 
+	"github.com/chef/automate/components/applications-service/pkg/storage"
 	"github.com/stretchr/testify/assert"
 )
+
+func TestConvertComposedServiceToStorage(t *testing.T) {
+	expected := &storage.Service{}
+	assert.Equal(t, expected, convertComposedServiceToStorage(&composedService{}))
+
+	expected = &storage.Service{ID: 1, SupMemberID: "foo"}
+	assert.Equal(t, expected, convertComposedServiceToStorage(&composedService{ID: 1, SupMemberID: "foo"}))
+
+	expected = &storage.Service{Application: "abcd", PreviousHealth: "MORE-OR-LESS-OK"}
+	assert.Equal(t, expected, convertComposedServiceToStorage(&composedService{
+		Application: "abcd", PreviousHealth: "MORE-OR-LESS-OK", Status: "",
+	}))
+}
+
+func TestConvertComposedServicesToStorage(t *testing.T) {
+	expected := []*storage.Service{
+		&storage.Service{},
+		&storage.Service{ID: 1, SupMemberID: "foo"},
+		&storage.Service{Application: "abcd", PreviousHealth: "MORE-OR-LESS-OK"},
+	}
+	subject := convertComposedServicesToStorage([]*composedService{
+		&composedService{},
+		&composedService{ID: 1, SupMemberID: "foo"},
+		&composedService{Application: "abcd", PreviousHealth: "MORE-OR-LESS-OK", Status: ""},
+	})
+	for _, e := range expected {
+		assert.Contains(t, subject, e)
+	}
+}
 
 func TestBuildWhereConstraintsFromFiltersNoFilterReturnsEmptyString(t *testing.T) {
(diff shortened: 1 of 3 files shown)


components/applications-service/pkg/storage/postgres/services.go
```diff
@@ -238,6 +262,21 @@ func (db *Postgres) getServiceFromUniqueFields(name, member string) (*service, b
 	return &svc, true
 }
 
+// converts an array of composedService to an array of storage.Service
+func convertComposedServicesToStorage(svcs []*composedService) []*storage.Service {
```
srenatus: Wouldn't it be nice if there was a way to avoid this? There isn't. 😄

components/applications-service/pkg/storage/postgres/service_internal_test.go
```diff
@@ -4,9 +4,39 @@ import (
...
+		&storage.Service{},
+		&storage.Service{ID: 1, SupMemberID: "foo"},
+		&storage.Service{Application: "abcd", PreviousHealth: "MORE-OR-LESS-OK"},
+	}
+	subject := convertComposedServicesToStorage([]*composedService{
+		&composedService{},
+		&composedService{ID: 1, SupMemberID: "foo"},
+		&composedService{Application: "abcd", PreviousHealth: "MORE-OR-LESS-OK", Status: ""},
+	})
+	for _, e := range expected {
```
srenatus: ❓ How is this different from `assert.ElementsMatch(t, expected, subject)`? I might need more coffee to see it.

components/applications-service/pkg/storage/postgres/service_internal_test.go
```diff
@@ -4,9 +4,39 @@ import (
...
+func TestConvertComposedServiceToStorage(t *testing.T) {
+	expected := &storage.Service{}
+	assert.Equal(t, expected, convertComposedServiceToStorage(&composedService{}))
+
+	expected = &storage.Service{ID: 1, SupMemberID: "foo"}
+	assert.Equal(t, expected, convertComposedServiceToStorage(&composedService{ID: 1, SupMemberID: "foo"}))
+
+	expected = &storage.Service{Application: "abcd", PreviousHealth: "MORE-OR-LESS-OK"}
+	assert.Equal(t, expected, convertComposedServiceToStorage(&composedService{
+		Application: "abcd", PreviousHealth: "MORE-OR-LESS-OK", Status: "",
```
srenatus: There's little value in providing `Status: ""`, since that's what it is *if unset* anways. Unless there's a reason to be explicit here.

components/applications-service/pkg/storage/postgres/services.go
afiune: well the way I thought it in the first place is that since both structs could be similar, by adding the `db:""` tags you could just re-use it. But I did cross a boundary there so, this is what you get to do the things right. 😅

components/applications-service/pkg/storage/postgres/service_internal_test.go
afiune: I have no idea but it didn't work 😂 - I might need more ☕️ as well

components/applications-service/pkg/storage/postgres/services.go
srenatus: Yup, I was complaining about the language, not your code. 😉

components/applications-service/pkg/storage/postgres/service_internal_test.go
afiune: Yes, mainly I am testing that when there are set values it converts them and when there is NO value, it also respects them. Thoughts?

srenatus: So, better be explicit :+1: that's ok.

components/applications-service/pkg/storage/postgres/services.go
afiune: I didn't take it to the code but the language as you said, I just tried to work around the limitation but, wasn't a good idea 😂

Issue #769, opened 2019-07-03
Search/Filter API


danielsdeleo:
# Description

Update applications APIs so that we can search/filter by the following:
* Service Name, e.g., "nginx"
* Package name, e.g., "core/nginx"
* Origin, e.g., "core"
* Group Name (service group name suffix after the '.')
* version, e.g., "1.2.3"
* Release (sem-ver/buildstamp)
* Environment
* Application (note that Deployment == Environment + Application)
* Channel
* Site
* Health Check Message

When filter criteria are present in a request to the service groups API, only services that meet the filter criteria are shown:

If all services in a service group match the filter criteria, the data presented in the row is identical to what is shown without filtering.

If no services in a service group match the filter criteria, the service group is not included in the response.

When some services in a service group match the filter criteria and some do not, the service group is presented as if it only contains the matching services. For example if a service group contains 3 services with site "Seattle" and 5 services with site "London," then when the filter query specifies `site=="Seattle"`, the mini-donut chart, "M of N OK" text, and release information should only show the three services which have `site=="Seattle"`.

Likewise the services API (the one that provides data for the right side panel) only returns services that match the filter criteria.In the previous example, only the three services where `site` is "Seattle" would be returned.

## Excluded:

* UI work will occur last
* An API to provide search suggestions for the search bar will be created later

# Demo:

curl the service groups API, maybe use jq to make it easy to see that all the things are returned. Add search parameters to the request and show that only matching things are returned. For completeness, the demo should show cases where:
* all services match the filter criteria and the response is unchanged compared to an unfiltered response
* no services match the filter criteria and the response contains no services/service groups
* some services match the filter criteria and the number of results and associated fields (health counts for the mini-donuts, etc.) are modified to match
* Some services match the filter and the services sidebar is filtered appropriately to show the matching services

afiune:
I have an idea to tackle this work, that is to bring back the main service-group select query.

The idea behind here is that we can put back the main select that we moved to be a view
so that you can inject the filters through WHERE statements, these statements can be composed
in the same manner we are building the WHERE statements in the `services.go` file at:
=> https://github.com/chef/automate/blob/master/components/applications-service/pkg/storage/postgres/services.go#L241-L258

I have an open PR to show the idea and to talk through this solution, also please feel free to take over this branch and continue the development if this sounds like the right way to go.

PR #907, opened 2019-07-16
IDEA: Bring back main service-group select query

afiune:
My idea here is that we can put back the main big select that we moved to be a view
so that you can inject the filters through WHERE statements, these statements can be composed
in the same manner we are building the where statements in the services.go file at:
=> https://github.com/chef/automate/blob/master/components/applications-service/pkg/storage/postgres/services.go#L241-L258

Githug Issue: https://github.com/chef/automate/issues/769

### DIFF
--- a/components/applications-service/pkg/storage/postgres/service_groups.go
+++ b/components/applications-service/pkg/storage/postgres/service_groups.go
@@ -21,6 +21,56 @@ import (
 // OK        | Else if all services are 'Ok'
 // ---------------------------------------------------------------------
 const (
+	// @afiune IDEA: Bring back the main query. The explanation why is below.
+	mainSelectServiceGroupHealth = `
+SELECT *
+  ,(CASE WHEN health_critical > 0 THEN '1_CRITICAL'
+         WHEN health_unknown  > 0 THEN '2_UNKNOWN'
+         WHEN health_warning  > 0 THEN '3_WARNING'
+         ELSE '4_OK' END ) as health
+FROM (
+  SELECT  sg.id
+    ,sg.deployment_id
+    ,sg.name as name
+    ,COUNT(s.health) FILTER (WHERE s.health = 'OK') AS health_ok
+    ,COUNT(s.health) FILTER (WHERE s.health = 'CRITICAL') AS health_critical
+    ,COUNT(s.health) FILTER (WHERE s.health = 'WARNING') AS health_warning
+    ,COUNT(s.health) FILTER (WHERE s.health = 'UNKNOWN') AS health_unknown
+    ,COUNT(s.health) AS health_total
+    ,round((COUNT(s.health) FILTER (WHERE s.health = 'OK')
+          / COUNT(s.health)::float) * 100) as percent_ok
+    ,(SELECT array_agg(DISTINCT CONCAT (s.package_ident))
+        FROM service AS s
+       WHERE s.group_id = sg.id) AS releases
+    ,d.app_name as app_name
+    ,d.environment as environment
+  FROM service_group AS sg
+  JOIN service AS s
+       ON s.group_id = sg.id
+  JOIN deployment as d
+       ON sg.deployment_id = d.id
+
+  -- @afiune my idea here is that we can put back the main big select that we moved to be a view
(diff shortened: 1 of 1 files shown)


components/applications-service/pkg/storage/postgres/service_groups.go
```diff
@@ -21,6 +21,56 @@ import (
 // OK        | Else if all services are 'Ok'
 // ---------------------------------------------------------------------
 const (
+	// @afiune IDEA: Bring back the main query. The explanation why is below.
```
srenatus: So, I realise it's not beautiful to duplicate, **but**: If we get rid of the nesting in "the big query", can't we **use the view, and _restrict `SELECT`s of that view_ in the same way**?

I think this might be largely equivalent (except for column order):
```sql
CREATE OR REPLACE VIEW service_group_health AS
  SELECT  sg.id AS id
         ,sg.deployment_id AS deployment_id
         ,sg.name AS name
         ,COUNT(*) FILTER (WHERE s.health = 'OK') AS health_ok
         ,COUNT(*) FILTER (WHERE s.health = 'CRITICAL') AS health_critical
         ,COUNT(*) FILTER (WHERE s.health = 'WARNING') AS health_warning
         ,COUNT(*) FILTER (WHERE s.health = 'UNKNOWN') AS health_unknown
         ,COUNT(*) AS health_total
         ,round(COUNT(*) FILTER (WHERE s.health='OK') * 100 / COUNT(*)) AS percent_ok
         ,(CASE WHEN (COUNT(*) FILTER (WHERE s.health='OK')) > 0 THEN '1_CRITICAL'
                WHEN (COUNT(*) FILTER (WHERE s.health='UNKNOWN')) > 0 THEN '2_UNKNOWN'
                WHEN (COUNT(*) FILTER (WHERE s.health='WARNING')) > 0 THEN '3_WARNING'
                ELSE '4_OK' END ) AS health
         ,(SELECT array_agg(DISTINCT CONCAT (s.package_ident))
             FROM service AS s
            WHERE s.group_id = sg.id) AS releases
         ,d.app_name AS app_name
         ,d.environment AS environment
        FROM service_group AS sg
        JOIN service AS s
             ON s.group_id = sg.id
        JOIN deployment AS d
             ON sg.deployment_id = d.id
       GROUP BY sg.id, sg.deployment_id, sg.name, d.app_name, d.environment;
```

I believe you should then be able to

```sql
SELECT * FROM service_group_health sgh WHERE sgh.id IN (SELECT DISTINCT group_id FROM service WHERE origin = 'custom' AND name = 'nginx')
```

etc. _I might be missing something_.

Notes:
- `COUNT(s.health) FILTER (WHERE s.health='yadda')` is the same as `COUNT(*) FILTER (WHERE s.health='yadda')` -- `COUNT(s.health)` would be different from `COUNT(*)` if `s.health` coud be NULL, but it cannot with that filter
- for the percentage, is better to multiply _first_ and divide _second_ ([see here](https://dba.stackexchange.com/a/69132))

afiune: You are right Stephan, this could be another approach but, now that you mention it, I think we can't use this approach because we still need to reduce services counts, that is, to display service-groups with information coming from the filteres services. I personally don't think is the right thing to do to hide or reduce the service-groups but I think this is what we want to have for iteration number one.

Example:
service group **redis.default** has four services:
1) 2 services with site equals to `us`
1) 2 services with site equals to `eu`

If I do a filter `site=eu`, then these things should happen:
1) display only service groups that has services with the site tag equal to `eu`
1) reduce the service group health summary to only count such services. (in this case the service group should report only two services instead of four.)

The approach that @danielsdeleo had was to use [`WITH` statements](https://www.postgresql.org/docs/9.1/queries-with.html), which now I think might be the best approach, though, we still need to move the view back to the code.

srenatus: Do we need to use _one view_ (or function or whatever) to support those different kinds of queries? I suppose there could be a view that has less (or no) aggregation happening, and another one that does aggregate services; queriers could use one or the other, depending on what they need to know/filter.

PR #905, opened 2019-07-15
[AUTOMATE-832] Use db_id in iam_statement_projects

tylercloke:
### :nut_and_bolt: Description

Use project's `db_id` in the `iam_statement_projects` table instead of `id`

### :athletic_shoe: Demo Script / Repro Steps

Setup:

```
start_all_services
chef-automate iam upgrade-to-v2 --beta2.1
export TOK=`chef-automate iam token create my-token --admin`
```

Make projects:

```
curl -kH "api-token: $TOK" -XPOST -d "$(jq -n '{ id: "project1", name: "my project 1" }')" https://localhost/apis/iam/v2beta/projects
curl -kH "api-token: $TOK" -XPOST -d "$(jq -n '{ id: "project2", name: "my project 2" }')" https://localhost/apis/iam/v2beta/projects
```

Make file `policy.json`:

```
{
  "name": "Test Policy 1",
  "id": "test-policy-1",
  "members": [ "user:local:*", "token:*" ],
  "statements": [
    {
      "effect": "ALLOW",
      "actions": [ "iam:*" ],
      "projects": [ "project1", "project2" ]
    }
  ]
}
```

Create policy:

```
curl -kH "api-token: $TOK" -XPOST -d @policy.json https://localhost/apis/iam/v2beta/policies
```

Run migration:

```
rebuild components/authz-service
```

`project_id` now refers to project's `db_id`:

```
chef_authz_service=# select * from iam_statement_projects;
             statement_id             | project_id
--------------------------------------+------------
 9e7fefc0-c7dc-40e6-b807-9a4a56f81cf2 |          1
 e7486529-2cff-42da-8803-040f0e05f513 |          1
 04f39a5d-b672-4f01-a53e-95e172ed11cc |          1
 a174a052-6ca4-4999-8cd7-47982830a08d |          1
 d0c16669-4254-4a60-8617-2dbdf0e97b3f |          1
 756a0faa-9b5f-4ee0-a026-fd55bbfee509 |          5
 756a0faa-9b5f-4ee0-a026-fd55bbfee509 |          6
 6506d801-393b-445d-9181-1eff4f97fdb7 |          2
(8 rows)
```

### :chains: Related Resources

- #832 

### :white_check_mark: Checklist

- [ ] Necessary tests added/updated?
- [ ] Necessary docs added/updated?
- [ ] Code actually executed?
- [ ] Vetting performed (unit tests, lint, etc.)?

PR #913, opened 2019-07-16
[part of #832] flip primary key for iam projects

srenatus:
Last step of #832's subtask for `iam_projects` references. ~~Branched-off of #905, and depends on that to go in first.~~

Also rolled in updating iam_statement_projects to use  iam_statements db_id as well.

- #832 
  - #905 

- [X] Code actually executed?

### DIFF
--- a/components/authz-service/storage/postgres/migration/sql/59_use_db_id_as_primary_key_in_iam_projects.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/59_use_db_id_as_primary_key_in_iam_projects.up.sql
@@ -1,0 +1,26 @@
+BEGIN;
+
+ALTER TABLE iam_projects
+    DROP CONSTRAINT iam_projects_pkey; -- 'id' primary key
+ALTER TABLE iam_projects
+    ADD PRIMARY KEY (db_id); -- implies unique
+ALTER TABLE iam_projects
+    DROP CONSTRAINT iam_projects_db_id_key CASCADE; -- 'db_id' unique
+ALTER TABLE iam_projects
+    ADD UNIQUE (id);
+
+-- recreate all links
+-- Note: we've done this before; but in the separated steps, we've
+-- relied on the unique
+ALTER TABLE iam_project_rules
+    ADD CONSTRAINT iam_projects_db_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id);
+ALTER TABLE iam_staged_project_rules
+    ADD CONSTRAINT iam_staged_project_rules_project_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id);
+ALTER TABLE iam_role_projects
+    ADD CONSTRAINT iam_role_projects_project_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id);
+ALTER TABLE iam_policy_projects
+    ADD CONSTRAINT iam_policy_projects_project_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id);
+ALTER TABLE iam_statement_projects
+    ADD CONSTRAINT iam_statement_projects_project_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id);
+
+COMMIT;
--- a/components/authz-service/storage/postgres/migration/sql/59_use_db_id_as_primary_key_in_iam_projects.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/59_use_db_id_as_primary_key_in_iam_projects.up.sql
@@ -13,14 +13,14 @@ ALTER TABLE iam_projects
 -- Note: we've done this before; but in the separated steps, we've
 -- relied on the unique
 ALTER TABLE iam_project_rules
-    ADD CONSTRAINT iam_projects_db_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id);
+    ADD CONSTRAINT iam_projects_db_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id) ON DELETE CASCADE;
 ALTER TABLE iam_staged_project_rules
-    ADD CONSTRAINT iam_staged_project_rules_project_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id);
+    ADD CONSTRAINT iam_staged_project_rules_project_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id) ON DELETE CASCADE;
 ALTER TABLE iam_role_projects
-    ADD CONSTRAINT iam_role_projects_project_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id);
+    ADD CONSTRAINT iam_role_projects_project_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id) ON DELETE CASCADE;
 ALTER TABLE iam_policy_projects
-    ADD CONSTRAINT iam_policy_projects_project_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id);
+    ADD CONSTRAINT iam_policy_projects_project_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id) ON DELETE CASCADE;
 ALTER TABLE iam_statement_projects
-    ADD CONSTRAINT iam_statement_projects_project_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id);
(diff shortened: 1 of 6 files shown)


components/authz-service/storage/postgres/migration/sql/59_use_db_id_as_primary_key_in_iam_projects.up.sql
```diff
@@ -0,0 +1,26 @@
+BEGIN;
+
+ALTER TABLE iam_projects
+    DROP CONSTRAINT iam_projects_pkey; -- 'id' primary key
+ALTER TABLE iam_projects
+    ADD PRIMARY KEY (db_id); -- implies unique
```
tylercloke: TIL: This is a [unique table constraint](https://www.postgresql.org/docs/8.1/ddl-constraints.html) (see section 5.3.3.) vs a table constraint (when you use the `UNIQUE` keyword).

Interestingly, when doing more research on table vs column unique constraint -- looks like the difference doesn't actually matter to postgres?

> We say that the first two constraints are column constraints, whereas the third one is a table constraint because it is written separately from any one column definition. Column constraints can also be written as table constraints, while the reverse is not necessarily possible, since a column constraint is supposed to refer to only the column it is attached to. (PostgreSQL doesn't enforce that rule, but you should follow it if you want your table definitions to work with other database systems.)

https://stackoverflow.com/questions/7455839/postgresql-difference-between-table-constraint-and-column-constraint

components/authz-service/storage/postgres/migration/sql/59_use_db_id_as_primary_key_in_iam_projects.up.sql
```diff
@@ -0,0 +1,26 @@
...
+ALTER TABLE iam_project_rules
+    ADD CONSTRAINT iam_projects_db_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id);
+ALTER TABLE iam_staged_project_rules
+    ADD CONSTRAINT iam_staged_project_rules_project_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id);
+ALTER TABLE iam_role_projects
+    ADD CONSTRAINT iam_role_projects_project_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id);
+ALTER TABLE iam_policy_projects
+    ADD CONSTRAINT iam_policy_projects_project_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id);
+ALTER TABLE iam_statement_projects
+    ADD CONSTRAINT iam_statement_projects_project_id_fkey FOREIGN KEY (project_id) REFERENCES iam_projects(db_id);
```
tylercloke: These aren't quite equlivant since they are missing `DEFERRABLE`.

Before migration:

```
    "iam_staged_project_rules_project_id_fkey" FOREIGN KEY (project_id) REFERENCES iam_projects(db_id) ON DELETE CASCADE DEFERRABLE
```

After migration:

```
    TABLE "iam_staged_rule_conditions" CONSTRAINT "iam_staged_rule_conditions_rule_db_id_fkey" FOREIGN KEY (rule_db_id) REFERENCES iam_staged_project_rules(db_id) ON DELETE CASCADE
```

However, after [reading up a bit](https://begriffs.com/posts/2017-08-27-deferrable-sql-constraints.html) it's not clear to me that this is a bad thing:

NOT DEFERRABLE | INITIALLY IMMEDIATE | INITIALLY DEFERRED
-- | -- | --
CHECK | row | row | row
NOT NULL | row | row | row
UNIQUE | row | statement | transaction
PRIMARY KEY | row | statement | transaction
FOREIGN KEY | statement | statement | transaction

This makes our transactions more flexible and I'm not sure for our use case we lose any safety.

tylercloke: Actually, not sure which it defaults to, so not sure things are _actually_ different 🤔 They must be if the final output is different though.

components/authz-service/storage/postgres/migration/sql/59_use_db_id_as_primary_key_in_iam_projects.up.sql
msorens: But, column constraints and table constraints are not actually different. When dealing with a _single_ column they are just two different ways to write exactly the same thing. When trying to apply uniqueness across _multiple_ columns, you must write it as a table constraint, but really only because of syntactical limitations.

components/authz-service/testhelpers/testhelpers.go
```diff
@@ -233,8 +233,13 @@ func SetupTestDB(t *testing.T) (storage.Storage, *TestDB, *prng.Prng, *migration
...
 
 func (d *TestDB) Flush(t *testing.T) {
-	_, err := d.Exec(`DELETE FROM iam_policies CASCADE; DELETE FROM iam_members CASCADE;
-		DELETE FROM iam_roles CASCADE; DELETE FROM iam_projects CASCADE; DELETE FROM iam_role_projects CASCADE;`)
+	_, err := d.Exec(`DELETE FROM iam_policies CASCADE;
+		DELETE FROM iam_members CASCADE;
+		DELETE FROM iam_roles CASCADE;
+		DELETE FROM iam_role_projects CASCADE;
+		DELETE FROM iam_project_rules;
+		DELETE FROM iam_staged_project_rules;
```
msorens: Why is `CASCADE` not needed on the two project rule tables?

components/authz-service/storage/postgres/migration/sql/59_use_db_id_as_primary_key_in_iam_projects.up.sql
msorens: I need some education and/or a good link explaining the usage of `DEFERRABLE` and `INITIALLY DEFERRED`. From what I have seen so far just using `DEFERRABLE`, as we have done previously in our code, does _not_ mean we are making the constraint deferred, only that it could be deferred if some other statement (e.g. `SET CONSTRAINTS...`) asked for it.

components/authz-service/testhelpers/testhelpers.go
srenatus: Oversight; also, I've just added these to see if it makes tests pass. I'll pick this up properly today, I hope.

components/authz-service/storage/postgres/migration/sql/59_use_db_id_as_primary_key_in_iam_projects.up.sql
srenatus: Thanks for paying attention to detail. I don't think we _need_ deferrable there; but I'd like to hear what @blakestier thinks about that.

`DEFERRABLE` means _"in a transaction, it's OK to be violated in the middle of it, as long as it's good on commit"_. 

This seems like a good in-depth intro: https://begriffs.com/posts/2017-08-27-deferrable-sql-constraints.html

blakestier: We can probably skip the deferrable, if we'd like. I was leaving it wherever it already was because we had other migrations that depended on the deferability, but given that those have already run by the time we hit this, it doesn't matter. Also, we can add deferability after the fact should we require it.

blakestier: mmm looks like we need them for upgrade on other tables, but not this one

components/authz-service/testhelpers/testhelpers.go
tylercloke: Don't think there is anything that actually would get cascade deleted via those tables atm anyway but I can add them in.

tylercloke: My question is why do we need those at all? They should be getting cascade deleted via `iam_projects`.

tylercloke: The lack of cascade deletion was what was making integration tests fail :) Added to the FK constraints.

components/authz-service/storage/postgres/migration/sql/60_use_db_id_as_primary_key_in_iam_statement_projects.up.sql
```diff
@@ -0,0 +1,165 @@
+BEGIN;
+
+ALTER TABLE iam_statement_projects
+    DROP CONSTRAINT iam_statement_projects_project_id_fkey;
```
srenatus: Wait, is that the right constraint to drop? The `project_id` part of this isn't altered, but the `statement_id` part -- or am I mistaken?

tylercloke: Ah yeah that's incorrect. I'll fix that up.

PR #915, opened 2019-07-16
[lib/cereal] Log correct error

jaym:


lib/cereal/postgres/postgres.go
```diff
@@ -644,7 +644,7 @@ func (pg *PostgresBackend) DequeueTask(ctx context.Context, taskName string) (*b
 	tid, task, err := pg.dequeueTask(tx, taskName)
 	if err != nil {
 		if errR := tx.Rollback(); errR != nil {
```
srenatus: I believe this could be written as
```go
if err != nil {
	if err := tx.Rollback(); err != nil {
		logrus.WithError(err).Warn("failed to rollback dequeue transaction")
	}
	cancel()
	return nil, nil, err
```
and the inner `err` would merely _shadow_, not overwrite, the outer `err`.

But this is clearer, I suppose.

ryancragun: I might be wrong but doesn't the linter complain if you shadow outer variables?

PR #922, opened 2019-07-16
[automate-cli] Ensure same writer isn't reused across retries

stevendanna:
If we reuse the io.Writer, then we create a corrupt package, thus the
retries have to happen outside the creation of the io.Writer.

### DIFF
--- a/components/automate-deployment/pkg/airgap/bundle_creator.go
+++ b/components/automate-deployment/pkg/airgap/bundle_creator.go
@@ -459,20 +459,23 @@ func (creator *InstallBundleCreator) downloadWithRetries(p habpkg.VersionedPacka
 				return nil // We don't log here since nearly every key is cached.
 			}
 
-			return creator.withRetry(func(tryCount int, ri retryInfo) error {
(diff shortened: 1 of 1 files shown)


components/automate-deployment/pkg/airgap/bundle_creator.go
```diff
@@ -438,61 +438,42 @@ func (creator *InstallBundleCreator) downloadPackageAndKeys(p habpkg.VersionedPa
...
+			progress.DownloadComplete(name, false)
+			if creator.keyCache.IsCached(header.KeyName) {
+				return nil // We don't log here since nearly every key is cached.
 			}
-			return err
-		}
-		progress.DownloadComplete(string(keyname), false)
-		return nil
+
+			return creator.withRetry(func(tryCount int, ri retryInfo) error {
```
stevendanna: I'm considering removing this inner retry.  I'm not sure which is more confusing, both downloads sharing the same retry count, or this key download potentially being retried tries*tries times.  I'm thinking sharing the same retry count is better.

Issue #867, opened 2019-07-11
Stage Project Ingest Rule Changes and Update Events


tylercloke:
# Problem

In order to facilitate updates of project ingest rules in a timely manner, we must apply them to the system in bulk. For the best user experience, we shall allow the user to stage these changes and persist them to the automate install before they take effect on the the incoming and historical data. This story defines that flow of information and the API that supports that use case.

# What it looks like when we're done

## Staging Changes

* A user should be able to stage changes to projects and rules in the system before applying them.

* If they leave their computer, get logged out, or their browser crashes, the changes they've staged should still be applicable

* If multiple users are staging changes, each user should be able to see the changes of the other user

* Staged changes should not be tied to a user, however, so we don't know who made what change

* All staged changes should be applied at the same time

## Applying Staged Changes

* Once a user is satisfied with the changes they've made to the projects and the rules, they should be able to apply the changes

* To effect the changes, the auth system needs to trigger an event in the event system that the project ingest rules have changed

### Waiting for success

* While the changes are taking effect, the user will see the status of the in-flight changes

* The auth system will wait for success messages from the in-flight changes in other services

* The user will receive a status via the user interface when the changes are complete

* A user will be unable to apply future changes without first canceling the in-flight changes

### Staging More Changes

* While changes are in flight, a user can continue to stage future changes to the ingest rules

* Before applying these rules, a user must first cancel the in-flight changes

# Subtasks

- [x] new table to persist desired changes to project rules
- [x] YJ Projects: Apply API
- [x] API to apply changes
- [x] API to fetch a project's rules including staged changes
- [x] API to fetch rules diff
- [x] API to check the state of currently running background job of applying rules
- [x] API to cancel currently running background job of applying rules
- [x] refactor UpdateRule
- [x] refactor DeleteRule
- [x] Refactor Rules API integration tests to include staging APIs @bcmdarroch #870 
- [x] Rule object contains staged or applied status @tylercloke 
- [x] Project API returns rule edits status @tylercloke https://github.com/chef/automate/pull/925

## Subtask: Project API returns rule edits status

We need to be able to show the rule edits status for a given project. This value appears in the Ingest Rules column of the Project List page.

There are three possible statuses:

Edits pending (see https://zpl.io/bJp9KvD)

* If any of the project’s rules have staged changes

No rules (see https://zpl.io/bA7ZvqP)

* If the project has no staged or applied rules

Applied (see https://zpl.io/2GG9m9m)

* If none of the project’s rules have staged changes

[...]

PR #925, opened 2019-07-16
[automate-867] Add overall status of rules for ListRulesForProject

tylercloke:

https://github.com/chef/automate/issues/867

#### Setup

```
start_all_services
```

##### No Rules: NO_RULES

```
curl -kH "api-token: $TOK" -XPOST -d "$(jq -n '{ id: "project1", name: "my project 1" }')" https://localhost/apis/iam/v2beta/projects
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects/project1/rules
{"rules":[],"status":"NO_RULES"}
```

##### Rules Staged: EDITS_PENDING

```
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/rules -d "$(jq -n '{ id: "rule1", name: "my foo rule", type: "NODE", project_id: "project1", conditions: [{ operator: "EQUALS", attribute: "CHEF_ORGS", values: ["The Misfits"]}]}')" | jq .
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects/project1/rules
{
  "rules": [
    {
      "id": "rule1",
      "project_id": "project1",
      "name": "my foo rule",
      "type": "NODE",
      "conditions": [
        {
          "attribute": "CHEF_ORGS",
          "values": [
            "The Misfits"
          ],
          "operator": "EQUALS"
        }
      ],
      "status": "STAGED"
    }
  ],
  "status": "EDITS_PENDING"
}
```

##### All Rules Applied: RULES_APPLIED

```
curl -kH "api-token: $TOK" -XPOST https://localhost/apis/iam/v2beta/apply-rules
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects/project1/rules
{
  "rules": [
    {
      "id": "rule1",
      "project_id": "project1",
      "name": "my foo rule",
      "type": "NODE",
      "conditions": [
        {
          "attribute": "CHEF_ORGS",
          "values": [
            "The Misfits"
          ],
          "operator": "EQUALS"
        }
      ],
      "status": "APPLIED"
    }
  ],
  "status": "RULES_APPLIED"
}
```

##### Some Staged, Some Applied: EDITS_PENDING

```
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/rules -d "$(jq -n '{ id: "rule2", name: "my foo rule", type: "NODE", project_id: "project1", conditions: [{ operator: "EQUALS", attribute: "CHEF_ORGS", values: ["yet another"]}]}')" | jq .
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects/project1/rules
{
  "rules": [
    {
      "id": "rule2",
      "project_id": "project1",
      "name": "my foo rule",
      "type": "NODE",
      "conditions": [
        {
          "attribute": "CHEF_ORGS",
          "values": [
            "yet another"
          ],
          "operator": "EQUALS"
        }
      ],
      "status": "STAGED"
    },
    {
      "id": "rule1",
      "project_id": "project1",
      "name": "my foo rule",
      "type": "NODE",
      "conditions": [
        {
          "attribute": "CHEF_ORGS",
          "values": [
            "The Misfits"
          ],
          "operator": "EQUALS"
        }
      ],
      "status": "APPLIED"
    }
  ],
  "status": "EDITS_PENDING"
}
```

- [ ] Code actually executed?

### TODO

- [ ] Update these [tests](https://github.com/chef/automate/pull/870) once they merge (will be done either in that PR or follow up PR)


components/authz-service/storage/v2/memstore/memstore.go
```diff
@@ -312,24 +312,39 @@ func (s *State) ListRules(_ context.Context) ([]*storage.Rule, error) {
...
 
-	return rules, nil
+	var rulesStatus v2.ProjectRulesStatus
+	if len(rules) == 0 {
+		rulesStatus = v2.NoRules
+	} else {
+		if anyStagedRules {
+			rulesStatus = v2.EditsPending
+		} else {
+			rulesStatus = v2.Applied
```
blakestier: I suppose you could skip this branch of the conditional and set the var to `v2.Applied` at the start (on 336)

tylercloke: That seems less clear to me on what the logic is personally. Curious what others think.

bcmdarroch: Go convention seems to be drop the `else` where possible

tylercloke: Ok I can do it that way, I think it's a bit harder to parse though

components/authz-service/server/v2/rules_test.go
```diff
@@ -16,6 +16,7 @@ import (
 )
 
 const applied = "applied"
+const staged = "staged"
```
srenatus: I'm aware your following the example of the code that was there already -- but it's a bit silly what we're doing here, isn't it? What do we try to protect against? I can only imagine two things:
1. We're renaming "applied" to "foobared" some time in the future. Would we then change this to `const applied = "foobared"`? I don't think so.
2. We're afraid of having typos in the code below, someone fat-fingering this into "aplied". What would happen? The tests where this is typo'ed in would probably fail, and the typo would be noticed. I'm having a hard time imagining a situation where a typo (which has had to slip through code review, too) turns a test into a phantom test, or some other harmful thing...

🤷‍♂ Of course it doesn't hurt either. I'm just wondering.

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1219,23 +1219,38 @@ func (p *pg) ListRulesForProject(ctx context.Context, projectID string) ([]*v2.R
...
+	var rulesStatus v2.ProjectRulesStatus
+	if len(rules) == 0 {
+		rulesStatus = v2.NoRules
+	} else {
+		if anyStagedRules {
+			rulesStatus = v2.EditsPending
+		} else {
+			rulesStatus = v2.Applied
+		}
+	}
```
srenatus: Let's flatten this:
```go
ruleStatus := v2.Applied
if len(rules) == 0 {
    rulesStatus = v2.NoRules
}
if anyStagedRules {
    ruleStatus = v2.EditsPending
}
```
I guess some similar comment applies to the memstore part. 😉

components/authz-service/storage/v2/project_rules_status.go
```diff
@@ -0,0 +1,37 @@
...
+
+const (
+	// The project has rules but all are applied
+	Applied ProjectRulesStatus = iota
+	// The project has rules and at least one is staged
+	EditsPending
+	// The project has no staged or applied rules
+	NoRules
+	// RulesStatusError occurs in some error situation
+	RulesStatusError
```
srenatus: [nit] If we changed the order here to have `RulesStatusError` be _first_, we could just `return nil, 0, err` in the postgres code. But this is fine, too.

components/authz-service/server/v2/rules_test.go
tylercloke: I agree in general. I followed the pattern because I didn't wanna update all the places applied was already used without a convo because I assume it was an intentional decision. I also think it's very unlikely that string would change but I guess this is helpful if it does. Regardless, probably not worth updating at this point 😅

components/authz-service/storage/v2/postgres/postgres.go
tylercloke: Very similar ;) Sounds like I'm in the vast minority on this one so I'll go ahead and update it in both places!

components/authz-service/storage/v2/project_rules_status.go
tylercloke: I guess that's technically true. Mostly makes the tests easier to read on failure. Otherwise, I don't think you'd really see the enum in it's integer context. Easy enough to fix though.

Issue #924, opened 2019-07-16
Large node objects can cause ingest-service to become overloaded permaturely


thomascate:
## Describe the bug
When testing with 3MB node objects I have encountered situations where a server starts rejecting messages while mostly idle. In this state, ingest-service is growing in memory usage, however won't consume more than 3-4 cores worth of CPU. 

The root cause of this appears to be that bulkRunPublisher only spawns a single thread, and the rest of the ingest pipeline is waiting for that thread to write to Elasticsearch. In my test scenario I used 3x Elasticsearch servers each using a 5k IOPs io1 ebs volume, and confirmed they weren't having load issues.

From the logs you can see publish messages are occurring sequentially, and taking increasingly long to finish.
>Jul 16 21:36:24 ip-10-1-1-179.us-west-2.compute.internal hab[2562]: ingest-service.default(O): time="2019-07-16T21:36:24Z" level=debug msg=bulkRunPublisherBundler bulkRequestsCount=3 message=ChefRun publish_time=774.793593ms
>Jul 16 21:36:28 ip-10-1-1-179.us-west-2.compute.internal hab[2562]: ingest-service.default(O): time="2019-07-16T21:36:28Z" level=debug msg=bulkRunPublisherBundler bulkRequestsCount=24 message=ChefRun publish_time=3.995722096s
>Jul 16 21:36:50 ip-10-1-1-179.us-west-2.compute.internal hab[2562]: ingest-service.default(O): time="2019-07-16T21:36:50Z" level=debug msg=bulkRunPublisherBundler bulkRequestsCount=135 message=ChefRun publish_time=21.856502108s
>Jul 16 21:37:35 ip-10-1-1-179.us-west-2.compute.internal hab[2562]: ingest-service.default(O): time="2019-07-16T21:37:35Z" level=debug msg=bulkRunPublisherBundler bulkRequestsCount=303 message=ChefRun publish_time=45.824964324s
>Jul 16 21:38:15 ip-10-1-1-179.us-west-2.compute.internal hab[2562]: ingest-service.default(O): time="2019-07-16T21:38:15Z" level=debug msg=bulkRunPublisherBundler bulkRequestsCount=303 message=ChefRun publish_time=39.510887008s

If you look at the "Transforming to bulk" messages, its buffer starts growing immediately, and grows until OOM killer takes effect and kills the ingest-service pid.

[...]

PR #927, opened 2019-07-17
Parallelization of bulk publisher for high loads

lancewf:
image

### 📋 TODO
- [ ] Add context check to fail the message out of the pipeline - this will be adding in future PRs
- [x] Add ``numberOfParallelBulkPublishers`` to the Automate config

Created a message distributor to send messages to multiple bulk publishers. The distributor will create a set number of bulk publishers that the messages will be sent to. The messages are sent to the first publisher until it is filled. Then the overflow messages are sent to the second publisher until it full and so on with the other publisher until they are all full. When they are all full the distributor will wait and try to send the messages again. This algorithm can be easily changed by updating the ``distributeMessage`` function. 

The reason behind filling one of the publishers before sending messages to the other ones is to increase the size of the bulk messages send. With a round-robin or equal distribution approach, all the bulk publishers will be sending small messages with the overhead of more connections to elasticsearch. For example, we would rather send 100 messages with one elasticsearch connection than 4 connections with 25 messages apiece. Once the load on the system is high enough that the bottleneck is publishing the bulk messages. The second bulk publisher's inbox will fill up and the second bulk publisher will start sending bulk messages in parallel. 

**Testing is needed** The distributor may only be useful with multiple elasticsearch nodes. 

Update the ``numberOfParallelBulkPublishers`` to add more publishers sending messages to Elasticsearch. 

https://github.com/chef/automate/issues/924

- [ ] Code actually executed?


components/ingest-service/pipeline/publisher/msg_distributor.go
```diff
@@ -0,0 +1,86 @@
...
+	out := make(chan message.ChefRun, 100)
+	pipes := buildActivePipes(childPipeBuilder, numProcessors, childPipeInboxSize)
+	mergeOutChannels(pipes, out)
+
+	go func() {
+		for msg := range in {
+			messageProcessed := false
+			for !messageProcessed {
+				for _, pipe := range pipes {
+					if !pipe.isFull() {
```
jaym: have you considered using something like this so you don't need to check for space:
```
select {
case pipe.in <- msg:
   messageProcessed = true
default:
}
```

jaym: this is fun: https://stackoverflow.com/questions/19992334/how-to-listen-to-n-channels-dynamic-select-statement

lancewf: Thank you. I will give that a try.

lancewf: The `select` worked great! Thanks Jay!

components/ingest-service/pipeline/publisher/msg_distributor.go
```diff
@@ -0,0 +1,87 @@
...
+			sendMessage(pipeInChannels, msg)
+		}
+		close(out)
+	}()
+
+	return out
+}
+
+func sendMessage(pipeInChannels []chan message.ChefRun, msg message.ChefRun) {
+	for true {
```
stevendanna: [nit]
```suggestion
	for {
```

stevendanna: Not sure how we feel about this style, but I think if you wanted you could do:

```golang
for !distributeMessage(pipeInChannels, msg) {
	// All pipes are full. Wait and try again
	time.Sleep(time.Millisecond * 10)
}

File: components/ingest-service/pipeline/publisher/msg_distributor.go [hunk 3]
```diff
@@ -0,0 +1,87 @@
...
+		case pipeInChannel <- msg:
+			return true
+		default:
+		}
+	}
+	return false
+}
+
+func mergeOutChannels(pipeOutChannels []<-chan message.ChefRun) chan message.ChefRun {
+	mergedOut := make(chan message.ChefRun, 100)
```
stevendanna: 💡 As a follow up to this PR, I think it might make sense to do some arithmetic on the maximum number of message we might end up with sitting in channel buffers if we get maximally backed up and then tune or remove some of these buffers accordingly.

components/ingest-service/pipeline/publisher/msg_distributor.go
```diff
@@ -0,0 +1,87 @@
...
+// If all the channels are full then false is returned
+func distributeMessage(pipeInChannels []chan message.ChefRun, msg message.ChefRun) bool {
+	for _, pipeInChannel := range pipeInChannels {
+		select {
+		case pipeInChannel <- msg:
+			return true
+		default:
+		}
+	}
+	return false
```
stevendanna: Would it make sense to log a warning here?  If we hit here, it means we potentially are no longer keeping up with requests right?

components/ingest-service/pipeline/publisher/msg_distributor.go
lancewf: Yeah, I thought about doing that change, but it does not seem as readable. The "distributeMessage" function does not say when to stop the for-loop. 
What about?
```suggestion
for messageProcessed := distributeMessage(pipeInChannels, msg); !messageProcessed; messageProcessed = distributeMessage(pipeInChannels, msg) {
   // All pipes are full. Wait and try again
   time.Sleep(time.Millisecond * 10)
}
```

components/ingest-service/pipeline/publisher/msg_distributor.go
lancewf: That is right. I will add a warning log.

components/ingest-service/pipeline/publisher/msg_distributor.go
stevendanna: Eh, I'd just leave it, it was pretty clear as is, I was just musing aloud :D

thomascate:
I did some load testing with this PR and a large Elasticsearch cluster and was able to achieve much higher throughput than before through the frontend. At peak I saw 60k 3MB nodes checking on a c5.4xlarge, with CPU being the ultimate bottleneck. It looks like this moves the bottleneck from waiting on publishing to the Automate servers hardware. I would recommend a bit higher than 6 though, from the testing we did 20 looks to be a good number.

components/ingest-service/pipeline/publisher/msg_distributor.go
afiune: I don't think you need the `default:` line

components/ingest-service/pipeline/chef_run.go
```diff
@@ -35,7 +36,8 @@ func NewChefRunPipeline(client backend.Client, authzClient iam_v2.ProjectsClient
 		processor.BuildRunProjectTagger(authzClient),
 		publisher.BuildNodeManagerPublisher(nodeMgrClient),
 		processor.BuildRunMsgToBulkRequestTransformer(client),
-		publisher.BuildBulkRunPublisher(client, maxNumberOfBundledRunMsgs),
+		publisher.BuildMsgDistributor(publisher.BuildBulkRunPublisher(
```
afiune: Super nit but, I can read this better if you ident it this way:
```go
publisher.BuildMsgDistributor(
  publisher.BuildBulkRunPublisher(client, maxNumberOfBundledRunMsgs),
  numberOfParallelBulkPublishers,
  maxNumberOfBundledRunMsgs,
),
```

components/ingest-service/pipeline/publisher/msg_distributor.go
```diff
@@ -0,0 +1,90 @@
...
+
+func msgDistributor(in <-chan message.ChefRun,
+	childPipeBuilder message.ChefRunPipe, numProcessors, childPipeInboxSize int) <-chan message.ChefRun {
+	pipeInChannels, out := buildChildPipes(childPipeBuilder, numProcessors, childPipeInboxSize)
+
+	go func() {
+		for msg := range in {
+			sendMessage(pipeInChannels, msg)
+		}
+		close(out)
```
afiune: Im not so sure about it but, should this close func call belongs in a defer statement? 🤔 (I'm just starting to think about what happens when one of these goroutines fail for any reason.)

components/ingest-service/pipeline/publisher/msg_distributor.go
```diff
@@ -0,0 +1,90 @@
...
+
+	return mergedOut
+}
+
+func buildChildPipes(pipeBuilder message.ChefRunPipe, numProcessors,
+	childPipeInboxSize int) ([]chan message.ChefRun, chan message.ChefRun) {
+	inChannels := make([]chan message.ChefRun, numProcessors)
+	outChannels := make([]<-chan message.ChefRun, numProcessors)
+	for index := range inChannels {
+		in := make(chan message.ChefRun, childPipeInboxSize)
```
afiune: Since this inbox size is coming all the way from the config file in habitat, should we do some parameter check for any `-1` or `0` (negative int or zero)?

components/ingest-service/pipeline/publisher/msg_distributor_test.go
```diff
@@ -0,0 +1,122 @@
+package publisher
```
afiune: nit: By convention, this file should be called `msg_distributor_internal_test.go` since it is testing the internals of the package.

afiune: Thouuuuugh, thanks for adding tests!!!!! 💯

components/ingest-service/pipeline/publisher/msg_distributor.go
lancewf: @afiune You need the `default:` line to allow this line `pipe.in <- msg` to not wait for room in the channel. 

The below code is doing. If the channel is not full send the message and set `messageProcessed` to true. If the channel is full exit the select. 
```
case pipe.in <- msg:
   messageProcessed = true
default:
```

components/ingest-service/pipeline/chef_run.go
lancewf: That does look better. Thanks.

components/ingest-service/pipeline/publisher/msg_distributor.go
lancewf: I think you might be correct. I don't want to make this change right now though, because this has already been thoroughly tested. This is a problem with all the processors in the pipeline and should be tested separately. 

Good catch!

components/ingest-service/pipeline/publisher/msg_distributor.go
lancewf: That is a good idea. I will want to do that in another PR. We have not fully published these parameters yet (meaning no it the docs). So, only the CS teams should be playing around with them for now.

api/config/ingest/config_request.go
```diff
@@ -29,6 +29,7 @@ func DefaultConfigRequest() *ConfigRequest {
 	c.V1.Sys.Service.MaxNumberOfBundledRunMsgs = w.Int32(2500)
 	c.V1.Sys.Service.MaxNumberOfBundledActionMsgs = w.Int32(10000)
 	c.V1.Sys.Service.NumberOfRunMsgsTransformers = w.Int32(9)
+	c.V1.Sys.Service.NumberOfRunMsgPublishers = w.Int32(2)
```
lancewf: We are currently going to default to two publishers. This should be good for a local Elasticsearch.

components/ingest-service/pipeline/publisher/bulk_publisher.go
```diff
@@ -12,8 +13,11 @@ import (
 )
 
 func BuildBulkRunPublisher(client backend.Client, maxNumberOfBundledRunMsgs int) message.ChefRunPipe {
+	count := 0
 	return func(in <-chan message.ChefRun) <-chan message.ChefRun {
-		return bulkRunPublisherBundler(in, client, maxNumberOfBundledRunMsgs)
+		name := fmt.Sprintf("pub-%d", count)
+		count++
+		return bulkRunPublisherBundler(in, client, maxNumberOfBundledRunMsgs, name)
```
lancewf: Adding a different name for each publisher to be able to see in the logs how many publishers are being used.

components/ingest-service/pipeline/publisher/msg_distributor.go
```diff
@@ -0,0 +1,94 @@
+package publisher
```
lancewf: This is the main addition.

components/ingest-service/pipeline/publisher/msg_distributor.go
afiune: Clever!!! ⭐️

jeremymv2:
picasso

PR #929, opened 2019-07-17
ci: shellcheck integration tests

srenatus:
I was waiting for a thing to finish, when it came over me.

I know some of these might be controversial, and the effort might outweight the benefit -- but I'd still prefer we shellcheck'ed all the things. 😉

### DIFF
--- a/.studiorc
+++ b/.studiorc
@@ -41,12 +41,12 @@ export ELASTICSEARCH_URL="http://localhost:10141"
 export POSTGRESQL_URL="localhost:5432"
 # Ingest
 export INGEST_URL="https://localhost:10122"
-export INGEST_PORT=$(echo $INGEST_URL | awk -F: '{ print $NF }')
-export INGEST_HOST=$(echo $INGEST_URL | awk -F: '{ print $2 }' | tr -d '/')
+export INGEST_PORT="localhost"
+export INGEST_HOST=10122
 # Automate Gateway service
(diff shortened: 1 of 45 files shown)


integration/tests/deep_upgrade.sh
```diff
@@ -26,24 +29,32 @@ do_create_config() {
 }
 
 do_deploy() {
-    cat $DEEP_UPGRADE_PATH > $test_manifest_path
```
stevendanna: 🤣 Unsure why I did it this way

.studiorc
```diff
@@ -41,12 +41,12 @@ export ELASTICSEARCH_URL="http://localhost:10141"
 export POSTGRESQL_URL="localhost:5432"
 # Ingest
 export INGEST_URL="https://localhost:10122"
-export INGEST_PORT=$(echo $INGEST_URL | awk -F: '{ print $NF }')
-export INGEST_HOST=$(echo $INGEST_URL | awk -F: '{ print $2 }' | tr -d '/')
+export INGEST_PORT="localhost"
```
srenatus: I've found this a bit silly, so I've ended it 😄

stevendanna: Uh oh, are these reversed?  Should this be 10122 and the next one "localhost"?

srenatus: 😅 oh no!

I'll fix that.

srenatus: https://github.com/chef/automate/pull/933

PR #934, opened 2019-07-17
Use the postgres11 client tools for backup/restore

jaym:
The HA backend cluster is using postgres 11. The 9.6 tools do not like
this and error out with a version mismatch.

### DIFF
--- a/components/automate-deployment/habitat/plan.sh
+++ b/components/automate-deployment/habitat/plan.sh
@@ -26,6 +26,9 @@ pkg_deps=(
   core/rsync
   core/tar
   chef/mlsa
+  # deployment-service uses the postgres11 client to backup/restore postgres.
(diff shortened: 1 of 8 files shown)


components/automate-deployment/habitat/plan.sh
```diff
@@ -26,6 +26,7 @@ pkg_deps=(
   core/rsync
   core/tar
   chef/mlsa
+  core/postgresql11-client
```
stevendanna: I feel like a comment here or _somewhere_ reminding us why we are on this version might be good.

integration/services/ha_backend/setup.sh
```diff
@@ -2,6 +2,8 @@
 
 set -e
 
+echo "Listing stuff"
```
ryancragun: My english teach would give you an f for using "stuff" ;)

PR #936, opened 2019-07-17
[cereal] Periodically cleanup workflow results

stevendanna:
Currently, the workflow results table grows without bound. This
imposes a limit of 10,000 rows that we attempt to enforce
periodically.

To avoid having to think too hard about the implications of DELETE and
our default transaction level, we take a postgresql advisory lock.

Note that we "over delete" to try to avoid deleting on every
interation once we hit the limit.

### DIFF
--- a/lib/cereal/postgres/task_cleaner.go
+++ b/lib/cereal/postgres/task_cleaner.go
@@ -10,6 +10,11 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+const (
+	defaultMaxWorkflowResults            = 10000
+	defaultWorkflowResultsDeletionMargin = 1000
+)
+
 type taskCleaner struct {
 	db      *sql.DB
 	wgStart sync.WaitGroup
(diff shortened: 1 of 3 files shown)


lib/cereal/postgres/task_cleaner.go
```diff
@@ -93,3 +113,52 @@ func (w *taskCleaner) expireDeadTasks(ctx context.Context, expireOlderThanSecond
...
+
+	tx, err := w.db.BeginTx(ctx, nil)
+	if err != nil {
+		return err
+	}
+
+	// NOTE(ssd) 2019-07-17: We are using a pg advsiory lock here,
+	// because this might run concurrently and we want the
+	// predictability of only running this deleting once.
+	row := tx.QueryRowContext(ctx, "SELECT pg_try_advisory_xact_lock(23320, 4090)")
```
jaym: how did you pick these numbers

stevendanna: `echo $RANDOM`

PR #940, opened 2019-07-17
[lib/cereal] Handle a few more TODOs

stevendanna:
- Return, thus canceling the transaction and rolling back the workflow
  on Enqueue Task failures.

- Fail rather than Complete workflow on jsonify error

- Fail workflow when EnqueueTask and Complete have been called in the
  same workflow step.

I wrote tests for 2 out of 3 of these cases. We really need a good way
to inject postgresql failures.

### DIFF
--- a/lib/cereal/cereal.go
+++ b/lib/cereal/cereal.go
@@ -229,8 +229,10 @@ func (w *workflowInstanceImpl) EnqueueTask(taskName string, parameters interface
 
 func (w *workflowInstanceImpl) Complete(opts ...CompleteOpts) Decision {
 	if len(w.tasks) > 0 {
-		panic("cannot call EnqueueTask and Complete() in same workflow step!")
+		logrus.Errorf("cannot call EnqueueTask and Complete in same workflow step! failing workflow")
+		return Decision{failed: true, err: errors.New("EnqueueTask and Complete called in same workflow step")}
 	}
+
 	d := Decision{
 		complete: true,
 	}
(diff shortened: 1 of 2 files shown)


lib/cereal/cereal.go
```diff
@@ -769,29 +771,24 @@ func (m *Manager) processWorkflow(ctx context.Context, workflowNames []string) b
...
 		s.Begin("enqueue_task")
 		for _, t := range decision.tasks {
 			err := completer.EnqueueTask(&t, backend.TaskEnqueueOpts{})
 			if err != nil {
-				// TODO(ssd) 2019-05-15: What do we
-				// want to do here? I think we need to
-				// rollback and have the workflow run
-				// again.
 				logrus.WithError(err).Error("failed to enqueue task!")
+				return true
```
stevendanna: We've defer'd a close on the completer, so this should end up rolling back our transaction and the workflow step should get retried.  This assumes the most likely errors here are pg errors.

lib/cereal/integration/workflow_error_test.go
```diff
@@ -41,7 +41,74 @@ func (suite *CerealTestSuite) TestWorkflowFail() {
...
+	instanceName := randName("instance")
+
+	doneChan := make(chan struct{})
+	m := suite.newManager(
+		WithWorkflowExecutor(
+			workflowName,
+			&workflowExecutorWrapper{
+				onStart: func(w cereal.WorkflowInstance, ev cereal.StartEvent) cereal.Decision {
+					close(doneChan)
+					return w.Continue(doneChan) // channels can't be marshalled
```
jaym: so clever

PR #945, opened 2019-07-18
[lib/cereal] Ignore missing task results

stevendanna:
The risk with this error return location and other similar locations
is that by simply returning the error, we will be aborting the
transaction, effectively placing the same workflow event on the queue.

This means that if the error is _not_ a transient error, we are
completely stuck.

Here, I've handled at least one non-transient error. Any other error
will continue to be retries. I think that some of this might be made
easier if we moved away from QueryRowContext since then Scan errors
and other errors wouldn't be intermingled.

### DIFF
--- a/lib/cereal/postgres/postgres.go
+++ b/lib/cereal/postgres/postgres.go
@@ -597,23 +597,44 @@ func (pg *PostgresBackend) DequeueWorkflow(ctx context.Context, workflowNames []
 
 	if event.Type == backend.TaskComplete {
 		event.CompletedTaskCount++
-		row := tx.QueryRowContext(ctx, getTaskResultQuery, taskResultID)
-		tr := backend.TaskResult{}
-		err = row.Scan(
-			&tr.TaskName, &tr.Parameters, &tr.Status, &tr.ErrorText, &tr.Result)
+		tr, err := getTaskResult(ctx, tx, taskResultID)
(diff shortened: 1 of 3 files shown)


lib/cereal/postgres/postgres.go
```diff
@@ -566,6 +561,29 @@ func (pg *PostgresBackend) DequeueWorkflow(ctx context.Context, workflowNames []
...
+	tr := backend.TaskResult{}
+	err := row.Scan(
+		&tr.TaskName, &tr.Parameters, &tr.Status, &tr.ErrorText, &tr.Result)
+	if err != nil {
+		if err == sql.ErrNoRows {
+			// This is unlikely to be a transient problem
+			// that aborting and retrying will fix. Thus,
+			// we move on and allow the workflow to handle
+			// the nil result.
+			return nil, nil
```
jaym: instead of nil, can we return a task result that returns an error

stevendanna: Oh, I like that idea.  Since then, if the user didn't need the task result, fine, but if they did they will definitely get an error when asking for it and don't have to do some nil checks on whatever they dereferenced.

stevendanna: Updated this to return an error, I thought about making it a completely different type, but I think this works for now.

PR #947, opened 2019-07-18
[lib/cereal] Temporarily pause processing of wayward workflows

stevendanna:
The processWorkflow event loop has a number of places where errors can
occur. In many cases, the most likely cause of the error is some
transient network problem. In these cases, logging the error, and
retrying may make sense.

Unfortunately, for the case of an unknown workflow event, no amount of
retrying will help. The event types are part of our source code, if we
don't know the event now, we won't know it 10 milliseconds from now
either.

This is especially problematic because it is likely that we'll
continue to select this same workflow event, stopping work on all
other workflows.

This PR attempts to address the problem by temporarily marking such
workflows as "wayward" and removing them from the list of workflows to
select. After some timeout, we start processing them again, in the
hope that the problematic event might have been processed by someone
else.

An alternative would be to drop the event completely. I think that
might make it problematic to add new workflow events in the future.
If a new workflow client is connected and uses some newly introduced
event, we likely want that client to be able to process the event.

### DIFF
--- a/lib/cereal/cereal.go
+++ b/lib/cereal/cereal.go
@@ -373,12 +373,51 @@ type TaskExecutor interface {
 	Run(ctx context.Context, task Task) (result interface{}, err error)
 }
 
+// waywardWorkflowList is a map of wayward workflows to the time they
+// entered the map. A wayward workflow is a workflow that has
+// experienced some fundamental processing error in the past. To avoid
+// stalling the processing of other workflows, we temporarily skip
+// events for any waywardWords.
+type waywardWorkflowList map[string]time.Time
+
+// waywardWorkflowTimeout is amount of time we will keep a given
+// workflow in our wayward workflow list.
+var waywardWorkflowTimeout = 60 * time.Second
+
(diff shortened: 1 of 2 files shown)


lib/cereal/cereal.go
```diff
@@ -373,12 +373,51 @@ type TaskExecutor interface {
 	Run(ctx context.Context, task Task) (result interface{}, err error)
 }
 
+// waywardWorkflowList is a map of wayward workflows to the time they
+// entered the map. A wayward workflow is a workflow that has
+// experienced some fundamental processing error in the past. To avoid
+// stalling the processing of other workflows, we temporarily skip
+// events for any waywardWords.
```
stevendanna:
```suggestion
// events for any wayward workflows.
```

stevendanna:
Added the log line.  Good point on this skipping all workflows of a given type.  Since I think that the cases this accounts for are going to be VERY, I'm OK with that for now.  If we expand its use, we should look into whether we .can improve it to be instances only.

PR #948, opened 2019-07-18
diagnostics: don't freak out if admin token v2 policy already exists

srenatus:
This makes me think that we're creating one v1 policy per run of
diagnostics, and never clean them up.

At least in IAM v2, we won't do that, but graciously ignore
"AlreadyExists" errors.

Follow-up to the discussion of #946.

### DIFF
--- a/components/automate-deployment/pkg/server/api_token_test.go
+++ b/components/automate-deployment/pkg/server/api_token_test.go
@@ -192,6 +192,47 @@ func TestGenerateAdminToken(t *testing.T) {
 		require.NotNil(t, resp)
 	})
 
+	t.Run("when API token succeeds but policy creation fails due to 'already exists', v2 policy creation succeeds", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


components/automate-deployment/pkg/server/api_token_test.go
```diff
@@ -192,6 +192,45 @@ func TestGenerateAdminToken(t *testing.T) {
...
+		}
+
+		mockAuthZ.CreatePolicyFunc = func(
+			_ context.Context, req *authz.CreatePolicyReq) (*authz.CreatePolicyResp, error) {
+
+			assert.Equal(t, []string{testSubjectString}, req.Subjects)
+			assert.Equal(t, "*", req.Action)
+			assert.Equal(t, "*", req.Resource)
+
+			return nil, status.Error(codes.AlreadyExists, "policy with id \"diagnostics-admin-token\" already exists")
```
srenatus: :headdesk: this test is wrong

PR #953, opened 2019-07-18
diagnostics: add iam v2 upgrade test

srenatus:
Spin-off of #886. Should be good without the rest 🤞


blakestier:
Commits merged with https://github.com/chef/automate/pull/886

PR #954, opened 2019-07-18
[deployment] Uninstall handles the case where /hab is mounted as a filesystem.

### LINKED ISSUE #764: Uninstall fails if `/hab` is mounted as a filesystem
# User Story

As an Automate user in a production installation, I install Automate in a separate `/hab` partition mounted on my server.   However, uninstall fails:

```
Removing Chef Automate
  stopping Automate services
  removing Automate from system startup configuration
  removing Habitat user and configuration
  removing Automate data stores
  removing Habitat package cache
ERRO[0207] failed to remove hab files                    error="unlinkat //hab: device or resource busy" path=/hab
UninstallError: An issue occurred when attempting to uninstall Chef Automate: failed to delete hab files at /hab: unlinkat //hab: device or resource busy
```

# Definition of Done

Uninstall should handle the case where `/hab` is installed as a separate filesystem gracefully.

# Demo Script / Repro Steps

1.  Create a sever with a separate `/hab` partition
2.  Perform a normal install
3.  Perform an uninstall

yzl:
If /hab is mounted as a filesystem, `chef-automate uninstall` removes only the contents and not the /hab directory.

Fixes #764

1. Build the deployment service

1. Start a docker container with a volume mounted to /hab: `$ docker run -d --name devtest --mount source="myvol,target=/hab" -v <full-path-to-your-automate-dir>:/a2 chefes/centos-systemd:latest`

1. Start a shell in the container: `$ docker exec -it <container-id> bash`

1. The contents of /proc/mounts in the container should include a line like: `/dev/sda1 /hab ext4 rw,relatime,stripe=1024,data=ordered 0 0`

1. In the container, do the following to install Automate
```
cd /a2
mkdir -p /hab/sup/default
CHEF_AUTOMATE_SKIP_SYSTEMD=true ./components/automate-deployment/bin/linux/chef-automate deploy --hartifacts ./results/ --override-origin <your-origin> --channel dev --skip-preflight
```
1. Run `chef-automate uninstall`. The output should be similar to

```
[root@49ace3e2632a a2]# CHEF_AUTOMATE_SKIP_SYSTEMD=true ./components/automate-deployment/bin/linux/chef-automate uninstall                                                                          
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! WARNING !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
This command will uninstall Chef Automate and any other applications deployed
with Habitat on this system. All associated data will be permanently destroyed.

If you wish to preserve your data, create a backup before uninstalling.
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! WARNING !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

The FQDN of your Chef Automate installation is "49ace3e2632a".
Type the FQDN in to continue.
FQDN: 49ace3e2632a

Removing Chef Automate
  stopping Automate services
  removing Automate from system startup configuration
  removing Habitat user and configuration
  removing Automate data stores
  removing Habitat package cache

Chef Automate uninstalled
```

1. Run `ls -l` on /hab to see that the directory remains but is empty.

- [x] Code actually executed?

### DIFF
--- a/components/automate-deployment/pkg/target/local_target.go
+++ b/components/automate-deployment/pkg/target/local_target.go
@@ -974,7 +974,34 @@ func (t *LocalTarget) DestroyPkgCache() error {
 	ctxLog := logrus.WithFields(logrus.Fields{
 		"path": habPath,
 	})
-	err := os.RemoveAll(habPath)
+
+	// If /hab is a mount point, remove the contents only.
+	mounts, err := ioutil.ReadFile("/proc/mounts")
+	if err != nil {
+		ctxLog.WithError(err).Error("failed to read file /proc/mounts")
+		return errors.Wrap(err, "failed to read file /proc/mounts")
+	}
+
+	// The trailing space avoids false matches on, e.g. /habababab.
+	habString := habPath + " "
+	if strings.Contains(string(mounts), habString) {
+		dirContents, err := filepath.Glob(filepath.Join(habPath, "*"))
+		if err != nil {
+			ctxLog.WithError(err).Errorf("failed to get contents of %s", habPath)
+			return errors.Wrapf(err, "failed to get contents of %s", habPath)
+		}
+		for _, file := range dirContents {
+			err = os.RemoveAll(file)
+			if err != nil {
+				ctxLog.WithError(err).Errorf("failed to remove hab file: %s", file)
+				return errors.Wrapf(err, "failed to remove remove hab file: %s", file)
+			}
+		}
+
+		return nil
+	}
+
+	err = os.RemoveAll(habPath)
 	if err != nil {
 		ctxLog.WithError(err).Error("failed to remove hab files")
 		return errors.Wrapf(err, "failed to delete hab files at %s", habPath)


components/automate-deployment/pkg/target/local_target.go
```diff
@@ -974,7 +974,34 @@ func (t *LocalTarget) DestroyPkgCache() error {
 	ctxLog := logrus.WithFields(logrus.Fields{
 		"path": habPath,
 	})
-	err := os.RemoveAll(habPath)
+
+	// If /hab is a mount point, remove the contents only.
+	mounts, err := ioutil.ReadFile("/proc/mounts")
```
stevendanna: If interested, we also have

https://github.com/chef/automate/blob/master/lib/proc/procparse.go#L71

which will return you a list of the mount point, but the existing code is also very clear, so probably not worth changing it.

components/automate-deployment/pkg/target/local_target.go
```diff
@@ -974,7 +974,34 @@ func (t *LocalTarget) DestroyPkgCache() error {
...
+	if err != nil {
+		ctxLog.WithError(err).Error("failed to read file /proc/mounts")
+		return errors.Wrap(err, "failed to read file /proc/mounts")
+	}
+
+	// The trailing space avoids false matches on, e.g. /habababab.
+	habString := habPath + " "
+	if strings.Contains(string(mounts), habString) {
+		dirContents, err := filepath.Glob(filepath.Join(habPath, "*"))
```
stevendanna: ℹ️ There is also `ioutils.ReadDir(habPath)` that could be used here.

Issue #884, opened 2019-07-12
CLI command to delete disconnected services


afiune:
Jira Ticket: https://chefio.atlassian.net/browse/A2-910

afiune:
Just like https://github.com/chef/automate/issues/858 but delete the disconnected services from the database.

PR #965, opened 2019-07-18
EAS Apps view: delete disconnected services

danielsdeleo:

Allow users to delete disconnected services from the EAS Applications view via CLI command. Disconnected services are services that have not had a health check message received within a user-specified threshold time.

Build automate-cli, applications-service, and automate-gateway; deploy all services.

In the studio, you can run something like `applications_populate_database ; sleep 60 ; applications_populate_database` to create sample data. Look at that applications view right after this completes, you should have 22 `billing-svc.kansas-prod` services. Then run:

```
chef-automate applications list-disconnected-services -m 1
```

Refresh the page and you should have 11 of them. You can/should also try the `-f JSON` and `-y` options.

- https://github.com/chef/automate/issues/884

- [x] Code actually executed?


api/external/applications/applications.proto
```diff
@@ -67,6 +67,17 @@ service ApplicationsService {
     option (chef.automate.api.iam.policy).action = "applications:serviceGroups:list";
   };
 
+  rpc DeleteDisconnectedServices(DisconnectedServicesReq) returns (ServicesRes ) {
+    option (google.api.http) = {
+      post: "/beta/applications/delete_disconnected_services"
```
afiune: Any reason why we didn’t do a DELETE request instead of POST?

components/applications-service/pkg/storage/postgres/services.go
```diff
@@ -229,6 +235,77 @@ func (db *Postgres) GetDisconnectedServices(thresholdMinutes int32) ([]*storage.
...
+// elapsed since the last event is greater than the specified thresholdMinutes.
+// After deleting the services, it also removes all supervisor, service_group,
+// and deployment records that are not referenced by any service record because
+// the application treats these things as emergent properties of groups of
+// services so they shouldn't exist if there are no associated services. This
+// also prevents the need for a second cleanup operation to delete these
+// things.
+func (db *Postgres) DeleteDisconnectedServices(thresholdMinutes int32) ([]*storage.Service, error) {
+
+	tx, err := db.DbMap.Begin()
```
afiune: Also why changing the way we were doing transactions here? Did you find a limitation or just preference?

api/external/applications/applications.proto
afiune: For context: https://github.com/chef/automate/pull/852#discussion_r302424709

kmacgugan:
This looks good, but I am wondering if we could add a test that exercises the whole delete sequence. Like deleting the supervisor and deployment after there are no services for them.

api/external/applications/applications.proto
danielsdeleo: I haven't re-read the spec recently, but according to my memory a DELETE request is supposed to refer to one specific resource in a REST context, like if we had something like `DELETE /services/:service_id` that would be correct. Since we are sending a command to delete anything that matches a specification, it's more of an RPC that should be a POST.

components/applications-service/pkg/storage/postgres/services.go
danielsdeleo: I wanted to stick to what gorp exposes so you only need the docs for one library to understand what's going on.

danielsdeleo:
> This looks good, but I am wondering if we could add a test that exercises the whole delete sequence. Like deleting the supervisor and deployment after there are no services for them.

🤔 hmmmm, I think it could be done if we expose an RPC for it. Which might be useful as an admin/debug thing. I'll think about it some more.

api/external/applications/applications.proto
danielsdeleo: Yeah the RFC says:

> The DELETE method requests that the origin server delete the resource
   identified by the Request-URI.

in this case there is not any specific resource identified by this URI.

cf. https://tools.ietf.org/html/rfc2616#section-9.7

afiune: Makes sense!

danielsdeleo:
Gonna look into getting test coverage for deleting orphaned service group/deployment/supervisor rows in a follow up; want to get this in so we can ship it more quickly with our promotion schedule.

PR #971, opened 2019-07-19
[deployment-service] Check NotBefore when validating certificates

stevendanna:
I've only really seen this case come up in somewhat contrived cases.
Namely, I found this a bit ago when trying to test what happens on cert expiration.

But, I suppose it could happen in some real-world VM scenarios as well.

### DIFF
--- a/components/automate-deployment/pkg/certauthority/certauthority.go
+++ b/components/automate-deployment/pkg/certauthority/certauthority.go
@@ -82,10 +82,21 @@ type ServiceCertData struct {
 
 const caNameFmt = "Chef Automate %s"
 
-// expiryMargin is how far before the NotAfter time on an x509
-// certificate we will mark the certificate as needing to be
-// regenerated.
-var expiryMargin = 28 * (time.Hour * 24)
+var (
+	// expiryMargin is how far before the NotAfter time on an x509
+	// certificate we will mark the certificate as needing to be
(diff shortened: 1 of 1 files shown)


components/automate-deployment/pkg/certauthority/certauthority.go
```diff
@@ -82,10 +82,21 @@ type ServiceCertData struct {
...
+var (
+	// expiryMargin is how far before the NotAfter time on an x509
+	// certificate we will mark the certificate as needing to be
+	// regenerated.
+	expiryMargin = 28 * (time.Hour * 24)
+
+	// notBeforeMargin is how far before a NotBefore time on an
+	// x509 certificate we will allow without marking the
+	// certificate as needing to be regenerated. The thinking
+	// behind this margin is that if the NotBefore time near, then
```
stevendanna:
```suggestion
	// behind this margin is that if the NotBefore time is near, then
```

components/automate-deployment/pkg/certauthority/certauthority.go
```diff
@@ -82,10 +82,21 @@ type ServiceCertData struct {
...
+	// it likely is just worth it it to wait rather than
```
stevendanna:
```suggestion
	// it likely is worth it it to wait rather than
```

Issue #75, opened 2019-03-05
Support for inspec-aws multi-region tests


davymcaleer:
# User Story

As an Automate User, I want to be able to use the updated multi-region capable [inspec-aws](https://github.com/inspec/inspec-aws) resources to be able to run compliance scans against all of my regions in a single compliance run.

# Definition of Done
We believe this should already be able to handle a version of InSpec that uses the new [inspec-aws](https://github.com/inspec/inspec-aws) resource pack, however it should be tested and we should be allowing Automate customers to make use of the improved multi-region support and ensure the CIS certified AWS Profile can be run correctly within A2

# Demo Script / Repro Steps
Run a compliance scan using the updated CIS Profile for Level 2 AWS Foundation Benchmark

vjeffrey:
talked with @alexpop about this:

need a new version of inspec for the inspec-aws profile, then can go ahead with this work

vjeffrey:
new inspec version/fixes for the plugins are being discussed. should be getting a fix through for that soon

vjeffrey:
https://github.com/inspec/train-aws/pull/8#pullrequestreview-256487053

vjeffrey:
@skpaterson is there a specific version of inspec we should pull in to move forward on this work?

vjeffrey:
https://github.com/chef/automate/compare/ap/use-inspec-4 <-- older wip branch from alex

vjeffrey:
https://github.com/chef/automate/pull/973 bumps the version of inspec used by automate to 4.7.18 https://github.com/chef/automate/pull/973

once that merges, we'll need to coordinate the release of AWS CIS Benchmark v1.2.0 with the automate change for going from creating one node per region -> one node per account

PR #973, opened 2019-07-19
bump inspec version to 4.7.18

vjeffrey:
just bumping the version of inspec used.
ran an aws api scan locally with this version to ensure everything works as expected (https://github.com/chef/automate/issues/75)

also: i'm not certain why expeditor didn't auto-open a pr for us, but it looks like that hasn't happened since we moved repos, so maybe need to update a repo reference over in inspec's ci pipeline.


vjeffrey:
hmm. the version bump is definitely causing some failures in our tests. i'm going to mark this as wip while i take a look at these.

stevendanna:
If you modify this: https://github.com/chef/automate/blob/master/.expeditor/config.yml#L162-L163 follow the 4.x series, expeditor will keep us updated

phiggins:
Somewhat unrelated to this PR, but I looked for other occurrences of `3.9.0` and found this sorta inaccurate comment. https://github.com/chef/automate/blob/master/api/config/compliance/config_request.go#L34-L35

The value does come from `default.toml`, but that file gets its value from the `plan.sh`. This comment makes it sound like there is another place to update in PRs like this one.

.studio/chef-server-collection
```diff
@@ -93,8 +93,8 @@ EOH
     "reporter": "chef-server-automate",
     "fetcher": "chef-server",
     "profiles": [{
-      "name": "apache-baseline",
-      "compliance": "test/apache-baseline"
+      "name": "ssl-baseline",
```
lancewf: I am just curious why the change from apache-baseline to ssl-baseline.

vjeffrey: that profile is using some deprecated resources that are causing the inspec check and exec calls to fail

PR #980, opened 2019-07-20
Add linting for nascent e2e typescript

msorens:

This is a follow-on to the recent conversion of e2e Cypress tests from
JavaScript to TypeScript:

1. Instruments the build to lint all-the-things and fail the build if lint fails.
2. Fix the outstanding 460 lint issues (!) from the initial port.
3. Drive-by: put package-lock.json under source control so that, as with our other UI code,
prevents version inconsistencies between developers.

### :+1: Definition of Done

image

You can now lint e2e files on the command-line:
```
cd e2e
npm install
npm run lint
```

https://github.com/chef/automate/pull/837

- [x] Code actually executed?

### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -289,6 +289,18 @@ steps:
       executor:
         docker:
 
+  - label: "[unit] e2e"
+    command:
+      - cd e2e
+      - npm install && npm run lint
+    timeout_in_minutes: 5
+    retry:
+      automatic:
+        limit: 1
+    expeditor:
+      executor:
+        docker:
+
   - label: "[unit] compliance-service"
     command:
       - scripts/install_golang.sh
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -289,14 +289,11 @@ steps:
       executor:
         docker:
 
-  - label: "[unit] e2e"
+  - label: "[static] e2e"
     command:
       - cd e2e
       - npm install && npm run lint
     timeout_in_minutes: 5
-    retry:
-      automatic:
-        limit: 1
     expeditor:
       executor:
         docker:
--- a/e2e/cypress/integration/ui/02_global_filter.spec.ts
+++ b/e2e/cypress/integration/ui/02_global_filter.spec.ts
@@ -1,51 +1,51 @@
 interface CreateProject {
-  id: string
-  name: string
+  id: string;
+  name: string;
 }
 
 describe('global projects filter', () => {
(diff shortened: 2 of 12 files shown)


.expeditor/verify.pipeline.yml
```diff
@@ -289,6 +289,18 @@ steps:
       executor:
         docker:
 
+  - label: "[unit] e2e"
```
msorens: OK, it is weird calling it "unit" tests... except that's where we have been putting linting (as prior ones have been unit-plus-lint). Could just call this one `[static]` I suppose. Other ideas...?

e2e/package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "cypress:open": "cypress open",
     "cypress:run": "cypress run",
-    "cypress:record": "cypress run --record"
+    "cypress:record": "cypress run --record",
+    "lint": "tslint -p tsconfig.json"
```
msorens: Adds the `npm run lint` command to use both from the command line and from buildkite.

e2e/tslint.json
```diff
@@ -0,0 +1,134 @@
+{
```
msorens: This was imported en masse from the same file in automate-ui (so we use a consistent rule set), less the extensions that are only needed in an Angular project.

.expeditor/verify.pipeline.yml
srenatus: We used to differentiate `[static]` and `[unit]` for services, e.g. authz-service, when we still ran them separately. When they were merged (because there wasn't that much value in having them separate, and a gigantic number of tests is hard to work with, I guess), static became part of unit. So yeah, let's  call this `[static]` 😃

.expeditor/verify.pipeline.yml
```diff
@@ -289,6 +289,18 @@ steps:
...
+    command:
+      - cd e2e
+      - npm install && npm run lint
+    timeout_in_minutes: 5
+    retry:
```
srenatus: Is this really needed?

msorens: Roger that; deleted the retry.

srenatus:
Re: https://chefio.slack.com/archives/C6JED4R8W/p1563775357005300

Would you mind removing that line from `.gitignore`?

e2e/cypress/integration/ui/02_global_filter.spec.ts
```diff
@@ -70,7 +73,7 @@ describe('global projects filter', () => {
   //     })
   //   cy.logout()
   // })
-})
+});
```
bcmdarroch: sigh i was enjoying not needing to remember semi-colons 😅

PR #996, opened 2019-07-22
Start linting our studio code

stevendanna:
- Enables all of shellcheck's "error" level linters for the studio. 
- Fixes SC2068 which was required to enable error level linters
- Fixes SC2155 which looked pretty easy to solve
- Fix a few miscellaneous issues that were in the vicinity of the other issues.
- In files otherwise touched by this change, remove unnecessary `;`s

### DIFF
--- a/scripts/repo_health.sh
+++ b/scripts/repo_health.sh
@@ -53,5 +53,7 @@ shellcheck -s bash -ax \
   scripts/*.sh \
   integration/**/*.sh
 
+shellcheck -s bash -ax -S error .studiorc .studio/*
+
 echo "Checking for possible credentials in the source code"
 go run ./tools/credscan
--- a/.studio/automate-gateway
+++ b/.studio/automate-gateway
@@ -27,7 +27,8 @@ function gateway_integration() {
   log_line "The automate-gateway version"
   gateway_version || return 1
 
-  local viewer_token="$(get_api_token_with_policy viewer-access)"
+  local viewer_token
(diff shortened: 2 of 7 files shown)


scripts/repo_health.sh
```diff
@@ -53,5 +53,7 @@ shellcheck -s bash -ax \
   scripts/*.sh \
   integration/**/*.sh
 
+shellcheck -s bash -ax -S error .studiorc .studio/*
```
srenatus: ❓ Worth a comment that this is not in the list above?

stevendanna: Good call. I'm about to change this in a follow-up PR to actually just be the list of checks that we fail since that will make it a sort of todo list:

```
> shellcheck -s bash -ax --format=gcc .studiorc .studio/* | awk '{print $NF}' | sort | uniq -c
   1 [SC1010]
   6 [SC2004]
   3 [SC2012]
   5 [SC2034]
   4 [SC2046]
   1 [SC2048]
 110 [SC2086]
   1 [SC2089]
   1 [SC2090]
   3 [SC2119]
   3 [SC2120]
   2 [SC2124]
   2 [SC2128]
   5 [SC2154]
   1 [SC2162]
  37 [SC2164]
   9 [SC2181]
   4 [SC2207]
   1 [SC2231]
   5 [SC2236]
```

So, I promise I'll comment this in the next PR, right now I'm impatient and don't want to wait for tests again.

PR #1001, opened 2019-07-22
automate-ui/chef-ui-library: cleanup habitat build

srenatus:
We've been installing and uninstalling angular-cli **once for each** of chef-ui-library and automate-ui. I've broken that for loop into two separate blocks for clarity.

Also, the previous code would end up running `npm install` in chef-ui-library twice: once intentionally, in its loop iteration, and once via `npm run install`, which is executed by `npm install` in the loop iteration for `automate-ui`.

I've wanted to cut down what is installed there -- _Do we really need pupetteer, and thereby chromium, to build the stencil components?_ -- but I gave up halfway 😉

-----
A quick, utterly unscientific experiment on my computer showed that this change saves us **one minute** in the habitat build of `automate-ui`:

```
$ time hab studio run 'build components/automate-ui' # on master
   automate-ui: Build time: 8m45s
      533.83 real         0.37 user         0.23 sys
```
vs
```
$ time hab studio run 'build components/automate-ui' # this PR
   automate-ui: Build time: 7m42s
      471.49 real         0.36 user         0.15 sys
```
Worth it? I think so. 🤔


components/automate-ui/habitat/plan.sh
```diff
@@ -37,41 +37,49 @@ do_unpack() {
...
-      for b in node_modules/.bin/*; do
-        fix_interpreter "$(readlink -f -n "$b")" core/coreutils bin/env
-      done
+  echo "Building $CACHE_PATH/automate-ui"
+  pushd "$CACHE_PATH/automate-ui"
+    npm_install --production
 
-      # Compile the Angular application
-      npm run build:prod
+    # Angular CLI isn't included in production deps so we need to install it manually.
```
msorens: Suggest s/manually/explicitly/

components/automate-ui/habitat/plan.sh
```diff
@@ -37,41 +37,49 @@ do_unpack() {
...
-        --fetch-retries 5
-      # Angular CLI isn't included in production deps so we need to install it manually.
-      npm install \
-          --unsafe-perm \
-          --loglevel error \
-          --fetch-retries 5 \
-          @angular/cli
+  echo "Building $CACHE_PATH/chef-ui-library"
+  pushd "$CACHE_PATH/chef-ui-library"
+    npm_install
```
msorens: Hmm, with your use of `--production` below, got to thinking should this one should include it, too...? 🤔

components/automate-ui/habitat/plan.sh
```diff
@@ -37,41 +37,49 @@ do_unpack() {
   cp ${PLAN_CONTEXT}/environments/environment.prod.ts $CACHE_PATH/automate-ui/src/environments/environment.prod.ts
 }
 
+npm_install() {
+  # --unsafe-perm lets our package.json install task copy files when running
```
msorens: Wow, that's a mouthful... Maybe clearer like this without losing any info...? 🤷‍♂  
```suggestion
  # --unsafe-perm lets our package.json install files when running
```
or this
```suggestion
  # --unsafe-perm lets our package.json install and copy files when running
```

components/automate-ui/habitat/plan.sh
srenatus: Nope it doesn't seem so, it would end up not having a stencil executable if we did that. There might be another way, I know too little about this stuff.

components/automate-ui/habitat/plan.sh
srenatus: Just copied, but yeah we can change it.

components/automate-ui/habitat/plan.sh
srenatus: If this was a github suggestion, I might have just clicked "apply"... but frankly, the semantic difference between the two is not that great, so I'll leave this as-is (or rather as-was-before). 😃

components/automate-ui/habitat/plan.sh
srenatus: I'm not sure this is not getting "wronger" with the suggestions. There's a difference between "npm install" and "npm run install", and I'd suppose the latter was refered to by "install task".

phiggins: "our package.json install task" refers to a specific thing, and copying files is what this flag affects. 

How about something like 

> --unsafe-perm enables our package.json install task to copy files when running as superuser during the hab package building.

?

components/automate-ui/habitat/plan.sh
phiggins: The `--production` flag prevents installing things listed in `devDependencies`, which had broken package building at some point. It might be worth benchmarking running without `--production` vs with it + explicitly installing `@angular/cli`? Explicitly installing `@angular/cli` might not even be necessary anymore...

phiggins: Oops, I didn't read far enough before making this comment. 😅

components/automate-ui/habitat/plan.sh
msorens: Ah, thanks @phiggins. Yes, that added "to" helps for sure!

srenatus:
rebased, will merge when green

PR #1002, opened 2019-07-22
deps: bump dex to v2.17.0

srenatus:
- https://github.com/dexidp/dex/releases/tag/v2.16.0
- https://github.com/dexidp/dex/releases/tag/v2.17.0

Nothing immediately relevant for us, but why not stay up to date?

Also, the next release is going to include https://github.com/dexidp/dex/pull/1486 which will let us simplify some code in local-user-service. Anticipating that, let's update now. 

ℹ️ This pulls in `github.com/coreos/go-oidc@v2`, as a dependency of Dex. This is a consequence of us using an in-memory dex server for testing, and thus having to eat its dependencies. You might argue that we should not have done that; but on the other hand, v2 of that library is what is maintained, and I do think we care about that. So, pulling this in is a good thing. 😃 

ℹ️ Also updates `golang.org/x/oauth2`, because it's better to do that once in a while -- changes are here: https://github.com/golang/oauth2/compare/0448841f0cbe9d174c6c1cedd177f583337b8e2c...0f29369cfe4552d0e4bcddc57cc75f4d7e672a33


srenatus:
💥 l-u-s needs its mocks updated, and session-service revealed some go-oidc dependency issue -- we'll need to bump that, too, I suppose.

PR #1003, opened 2019-07-22
Continue linting the studio

stevendanna:
This PR resolves some more lint errors and changes our CI run 
to only exclude those rules with known problems, ensuring the 
situation doesn't get worse over time.

### DIFF
--- a/.studio/notifications-service
+++ b/.studio/notifications-service
@@ -120,7 +120,7 @@ function notifications_update_component() {
   pushd /src/components/notifications-service/server >/dev/null;
     mix local.hex --force
     mix local.rebar --force
-    MIX_ENV=habitat mix do deps.get, compile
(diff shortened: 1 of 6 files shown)


.studio/notifications-service
```diff
@@ -120,7 +120,7 @@ function notifications_update_component() {
   pushd /src/components/notifications-service/server >/dev/null;
     mix local.hex --force
     mix local.rebar --force
-    MIX_ENV=habitat mix do deps.get, compile
+    MIX_ENV=habitat mix "do" deps.get, compile
```
srenatus: Hm? OK.

stevendanna: `do` is a shell keyword so things get confused. 🤷‍♂ I didn't look too far into it.

srenatus: Yeah suspected something like that. 🚤

PR #1010, opened 2019-07-22
[deployment] Add guard rails around upgrading deployments done with airgap bundles

### LINKED ISSUE #955: It is unclear to people that upgrading an install done with an airgap bundle requires an airgap bundle.
Ideas to make it clearer:

1. return an error if you are in airgap mode and do `upgrade run` without an airgap bundle
2. return info in API response to upgrade and print it out
3. If we don't already say this in the docs, we should.

yzl:
Supply a useful error when a user does not provide an airgap bundle when upgrading a deployment that was created with one.

Fixes #955

1.  Build an airgap bundle: `chef-automate airgap bundle create ./today.aib`.
2. Run `chef-automate airgap bundle create ./today.aib` to deploy.
3. Once the deploy has completed successfully, run `chef-automate upgrade run`. This will fail with an error `To upgrade a deployment created with an airgap bundle, use --airgap-bundle to specify a bundle to use for the upgrade.`.

- [x] Code actually executed?

### DIFF
--- a/components/automate-chef-io/content/docs/airgapped-installation.md
+++ b/components/automate-chef-io/content/docs/airgapped-installation.md
@@ -96,6 +96,7 @@ To send data from your Chef Server or Chef Clients to Chef Automate 2, the proce
 See [Configure Data Collection]({{< relref "data-collection.md" >}}) for more information.
 
 ### Upgrades
+To upgrade an airgapped install, you must supply an airgap bundle.
 
 On an internet-connected host, follow the steps in [Create an Airgap
 Installation Bundle]({{< relref "#create-an-airgap-installation-bundle" >}}) to upgrade your
--- a/components/automate-cli/cmd/chef-automate/upgrade.go
+++ b/components/automate-cli/cmd/chef-automate/upgrade.go
@@ -60,6 +60,11 @@ func runUpgradeCmd(cmd *cobra.Command, args []string) error {
 	}
 
 	offlineMode := upgradeRunCmdFlags.airgap != ""
+
+	if airgap.AirgapInUse() && !offlineMode {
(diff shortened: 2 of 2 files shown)


components/automate-deployment/pkg/server/server.go
```diff
@@ -1161,6 +1161,12 @@ func (s *server) shouldFetchManifest() bool {
 // the configured channel. If neither are available, an error is
 // returned.
 func (s *server) nextManifest() (*manifest.A2, error) {
+
+	if airgap.AirgapInUse() && s.deployment.Config.Deployment.GetV1().GetSvc().GetUpgradeStrategy().GetValue() != "none" {
+		logrus.Error("Airgapped deployments require the configuration setting upgrade_strategy=\"none\"")
+		return nil, errors.New("Airgapped deployments require the configuration setting upgrade_strategy=\"none\"")
```
jaym: will this break people?
why does it have to be set to none again?

yzl: The upgrade_strategy should be "none" for an airgapped installation so that the converger won't keep trying to reach the Internet to see if there is a new manifest available.

The people this might break are the ones using an airgap bundle to deploy when they aren't actually behind an airgap; they've been taking upgrades when they shouldn't have been able to. People using airgap bundles behind an airgap won't be worse off, and they might be better off because they'll get an error in the logs that says what to fix instead of a bunch of timeouts.

I think if enough things break, people using airgap bundles without being behind an airgap who take an upgrade containing this code could end up in trouble, so I will do something to write an error to the logs without throwing an actual error.

jaym: i'm not sure if this is correct. airgap has it's own manifest provider that will only look at what is on disk. Can you double check

yzl: You're right, and this case is a non-issue. I'll revert the change and update the PR.

PR #1017, opened 2019-07-22
Fix integration tests to run locally

jaym:
do_build was messed up
- the build changed components script was using the wrong ruby
- we need to be careful about the umasks we use when calling hab pkg install

### DIFF
--- a/integration/helpers/build.sh
+++ b/integration/helpers/build.sh
@@ -2,7 +2,7 @@
 
 build_changed() {
     hab pkg install core/ruby
-    PATH=$PATH:$(hab pkg path core/ruby)/bin
+    PATH=$(hab pkg path core/ruby)/bin:$PATH
     export PATH
     sudo "$(hab pkg path core/ruby)/bin/gem" install toml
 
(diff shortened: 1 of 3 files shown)


integration/helpers/build.sh
```diff
@@ -2,7 +2,7 @@
 
 build_changed() {
     hab pkg install core/ruby
-    PATH=$PATH:$(hab pkg path core/ruby)/bin
+    PATH=$(hab pkg path core/ruby)/bin:$PATH
```
phiggins: I don't know if this could cause the same problem, but we also construct PATH this way around in this script: https://github.com/chef/automate/blob/3757e9fdfec19d435860d46506e6f51611a3bea4/scripts/verify_build.sh#L18

stevendanna:
@bcmdarroch 

> Do we have steps for running integration tests locally documented anywhere?

The test suite was designed to run locally with the same command found in the Buildkite definition.  Thus for most of the integration test suites, you can run:

`integration/run_test integration/tests/TEST_DEFINITION`

PR #1018, opened 2019-07-22
Dan/improve disconnected svcs tests

danielsdeleo:

In https://github.com/chef/automate/pull/965 we added a CLI command to delete disconnected services. When the services are deleted, we also delete any supervisors, service groups, and/or deployments that no longer are referenced by any services. The original PR didn't have tests to ensure these objects were removed.  In this PR we update the tests to use the telemetry stats endpoint to verify the relevant rows are also deleted.

This is a test-focused PR, but the change to the stats endpoint can be verified by building the applications service, running `applications_populate_database` and then hitting the stats endpoint with `applications_grpcurl chef.automate.api.applications.ApplicationsService.GetServicesStats`. You can run the integration tests in the studio with `applications_integration`

- [ ] Code actually executed?


components/applications-service/pkg/storage/postgres/deployments.go
```diff
@@ -8,6 +8,10 @@ import (
 const (
 	deleteDeploymentsWithoutServices = `
 DELETE FROM deployment WHERE NOT EXISTS (SELECT 1 FROM service WHERE service.deployment_id = deployment.id )
+`
+	selectDeploymentsTotalCount = `
```
srenatus: [nit] I'm not sure this query _needs_ linebreaks. But this is fine 😉

api/external/applications/applications.proto
```diff
@@ -87,6 +87,37 @@ service ApplicationsService {
...
+//  deployment    | id
+//  deployment    | app_name
+//  deployment    | environment
+//  deployment    | created_at
+//  deployment    | updated_at
+//  service_group | id
+//  service_group | name
+//  service_group | deployment_id
+//  service_group | created_at
+//  service_group | updated_at
```
danielsdeleo: oops, this is left over from a different way I was going to solve the problem.

PR #1020, opened 2019-07-23
Project details and rules completion

msorens:

This is a follow-on to PR-736, implementing the suggestions/corrections remaining from that PR, plus other issues uncovered.

_Please follow the commits to understand specific changes._

Broadly, this PR covers some architectural changes and some behavioral changes.
The former are covered with my review comments and commit comments; the latter are summarized here:

1. Add an asterisk on name and type indicating required. 🤔 Don't we need to say somewhere `(*) Indicates required fields` ??
2. Tightened spacing per UX design between labels and input boxes.
3. Change the label in the error pop-up to match the `Rule Name` label.
4. Fix the width of the error pop-up to match the `Rule Name` field.
5. Increase spacing per UX design between fields.
6. Add an error pop-up for the `Resource Type` field; this will trigger if you tab your way through the element. 
7. Put that error pop-up immediately beneath the input (instead of beneath the `Resource types cannot be changed...` line). This required moving that line of text outside of the `<chef-form-field>`. Don't know if that will affect a screen reader or not.
8. (no arrow) Adjust the `Rule Name` error pop-up to also appear if only whitespace is typed. (That makes it behave the way every other input field works.)
9. (no arrow) Adjust font sizes/weights of everything except the title and description per UX spec.

image

1. Add error pop-up for `attribute`. This works independently for each row. Note that always says "Attribute is required" rather than "Event Attribute is required" or "Node Attribute is required". Could add that if desired for consistency. Widths of these pop-ups are a compromise: ideally should be equal to field width, but those elements are too narrow to show the error text. 🤷‍♂ 
2. Add error pop-up for `operator`.
3. Add error pop-up for `value`.
4. Recall in the previous section I mentioned I thought it was desirable to put the error pop-up immediately under the input, not separated by the help line of text, as you see here. Alas, the earlier fix was not viable in this table structure.
5. Displays trash can for all rows--but only when there is more than one row.
6. Adjusted size per UX design spec.

image

1. Correct enabling control menu: now opens on staged rules.

image

1. Disable changing rule type when editing rule.

image

Still to be done:
Adding the ID as discussed in slack starting here: https://chefio.slack.com/archives/C61F9HHKK/p1564077123212600

### :+1: Definition of Done

Code ready for proper wiring.

Rebuild automate-ui.
Go to Settings >> Projects
Create a project.
Create one or more rules for the project with one or more conditions.
Edit a rule.
Add/delete conditions.

_Note that there is no actual storage persistence so you will see some "issues" with data not being saved._

https://github.com/chef/automate/pull/736

- [x] Code actually executed?

### DIFF
--- a/components/automate-ui/src/app/entities/rules/rule.model.ts
+++ b/components/automate-ui/src/app/entities/rules/rule.model.ts
@@ -1,7 +1,8 @@
 export type RuleStatus = 'applied' | 'staged';
 export type RuleType = 'node' | 'event';
+export type ConditionOperator = 'MEMBER_OF' | 'EQUALS';
 
-interface KVPair {
+export interface KVPair {
   key: string;
   value: string;
 }
@@ -22,6 +23,6 @@ export interface Rule {
 
 export interface Condition {
   attribute: string;
-  values: any;
-  operator: string;
+  operator: ConditionOperator;
+  values: string[];
 }
--- a/components/automate-ui/src/app/entities/rules/rule.model.ts
+++ b/components/automate-ui/src/app/entities/rules/rule.model.ts
@@ -3,7 +3,7 @@ export interface Rule {
   project_id?: string;
   name: string;
   type: string;
-  edits: string;
+  status: string;
   conditions: Condition[];
 }
 
--- a/components/automate-ui/src/app/entities/rules/rule.model.ts
+++ b/components/automate-ui/src/app/entities/rules/rule.model.ts
@@ -1,10 +1,11 @@
 export type RuleStatus = 'applied' | 'staged';
+export type RuleType = 'node' | 'event';
 
 export interface Rule {
   id?: string;
   project_id?: string;
   name: string;
-  type: string;
+  type: RuleType;
   status: RuleStatus;
   conditions: Condition[];
 }
--- a/components/automate-ui/src/app/entities/rules/rule.model.ts
+++ b/components/automate-ui/src/app/entities/rules/rule.model.ts
@@ -2,7 +2,7 @@ export type RuleStatus = 'applied' | 'staged';
 export type RuleType = 'node' | 'event';
 export type ConditionOperator = 'MEMBER_OF' | 'EQUALS';
 
-export interface KVPair {
+interface KVPair {
   key: string;
   value: string;
 }
--- a/components/automate-ui/src/app/entities/rules/rule.model.ts
+++ b/components/automate-ui/src/app/entities/rules/rule.model.ts
@@ -1,6 +1,16 @@
 export type RuleStatus = 'applied' | 'staged';
 export type RuleType = 'node' | 'event';
 
+interface KVPair {
+  key: string;
+  value: string;
+}
+
+// Using array of pairs instead of a hashmap to allow ordering
+export type RuleTypeMappedObject = {
+  [K in 'node' | 'event']: KVPair[]
+};
+
 export interface Rule {
   id?: string;
   project_id?: string;
--- a/components/automate-ui/src/app/entities/rules/rule.model.ts
+++ b/components/automate-ui/src/app/entities/rules/rule.model.ts
@@ -1,9 +1,11 @@
+export type RuleStatus = 'applied' | 'staged';
+
 export interface Rule {
   id?: string;
   project_id?: string;
   name: string;
   type: string;
-  status: string;
+  status: RuleStatus;
   conditions: Condition[];
 }
 
--- a/components/automate-ui/src/app/pages/project/details/project-details.component.scss
+++ b/components/automate-ui/src/app/pages/project/details/project-details.component.scss
@@ -25,6 +25,11 @@ chef-tab-selector {
 }
 
 form {
+
+  #update-name {
+    width: 240px;
+  }
+
   input {
     font-size: 1em;
     width: 240px;
--- a/components/automate-ui/src/app/pages/project/details/project-details.component.spec.ts
+++ b/components/automate-ui/src/app/pages/project/details/project-details.component.spec.ts
@@ -26,7 +26,7 @@ describe('ProjectDetailsComponent', () => {
       id: 'rule-1',
       project_id: 'uuid-1',
       name: 'Rule 1',
-      type: 'NODE',
+      type: 'node',
       status: 'staged',
       conditions: [
         {
@@ -40,7 +40,7 @@ describe('ProjectDetailsComponent', () => {
       id: 'rule-2',
       project_id: 'uuid-1',
       name: 'Rule 2',
-      type: 'EVENT',
+      type: 'event',
       status: 'applied',
       conditions: [
         {
--- a/components/automate-ui/src/app/pages/project/details/project-details.component.spec.ts
+++ b/components/automate-ui/src/app/pages/project/details/project-details.component.spec.ts
@@ -11,7 +11,7 @@ import { IAMType } from 'app/entities/policies/policy.model';
 import { GetProjectSuccess } from 'app/entities/projects/project.actions';
 import { GetRulesSuccess } from 'app/entities/rules/rule.actions';
 import { projectEntityReducer } from 'app/entities/projects/project.reducer';
-import { ProjectDetailsComponent } from './project-details.component';
+import { ProjectDetailsComponent, ProjectTabNames } from './project-details.component';
 
 describe('ProjectDetailsComponent', () => {
   let component: ProjectDetailsComponent;
@@ -146,16 +146,25 @@ describe('ProjectDetailsComponent', () => {
       expect(component.rules.length).toBe(0);
     });
 
-    it('show rules section when rules tab is selected', () => {
-      component.onTabChange({ target: { value: 'rules'} });
-      expect(component.showTab('rules')).toBeTruthy();
+   it('defaults to showing rules section', () => {
+      expect(component.showTab(ProjectTabNames.Rules)).toBeTruthy();
+      expect(component.showTab(ProjectTabNames.Details)).toBeFalsy();
     });
 
-    it('should not display rule table when no rules', () => {
+    it('shows/hides sections when based on selection', () => {
+      component.onTabChange({ target: { value: ProjectTabNames.Details } });
+      expect(component.showTab(ProjectTabNames.Rules)).toBeFalsy();
+      expect(component.showTab(ProjectTabNames.Details)).toBeTruthy();
+      component.onTabChange({ target: { value: ProjectTabNames.Rules } });
+      expect(component.showTab(ProjectTabNames.Rules)).toBeTruthy();
+      expect(component.showTab(ProjectTabNames.Details)).toBeFalsy();
+    });
+
+    it('does not display rule table', () => {
       expect(component.showRulesTable()).toBeFalsy();
     });
 
-    it('the create your first rule message should display', () => {
+    it('displays create-your-first-rule message', () => {
       expect(component.showFirstRuleMessage()).toBeTruthy();
     });
   });
@@ -174,20 +183,29 @@ describe('ProjectDetailsComponent', () => {
       expect(component.rules.length).toBe(2);
     });
 
-    it('show rules section when rules tab is selected', () => {
-      component.onTabChange({ target: { value: 'rules'} });
-      expect(component.showTab('rules')).toBeTruthy();
+    it('defaults to showing rules section', () => {
+      expect(component.showTab(ProjectTabNames.Rules)).toBeTruthy();
+      expect(component.showTab(ProjectTabNames.Details)).toBeFalsy();
     });
 
-    it('should display rule table', () => {
+    it('shows/hides sections when based on selection', () => {
+      component.onTabChange({ target: { value: ProjectTabNames.Details } });
+      expect(component.showTab(ProjectTabNames.Rules)).toBeFalsy();
+      expect(component.showTab(ProjectTabNames.Details)).toBeTruthy();
+      component.onTabChange({ target: { value: ProjectTabNames.Rules } });
+      expect(component.showTab(ProjectTabNames.Rules)).toBeTruthy();
+      expect(component.showTab(ProjectTabNames.Details)).toBeFalsy();
+    });
+
+    it('displays rule table', () => {
       expect(component.showRulesTable()).toBeTruthy();
     });
 
-    it('the create your first rule message should not display', () => {
+    it('does not display create-your-first-rule message', () => {
       expect(component.showFirstRuleMessage()).toBeFalsy();
     });
 
-    it('show a link back to project when a rule is in "edits pending" state', () => {
+    it('shows a link back to project when a rule is in "edits pending" state', () => {
       expect(component.showProjectLink()).toBeTruthy();
     });
 
--- a/components/automate-ui/src/app/pages/project/details/project-details.component.spec.ts
+++ b/components/automate-ui/src/app/pages/project/details/project-details.component.spec.ts
@@ -6,7 +6,6 @@ import { MockComponent } from 'ng2-mock-component';
 
 import { ChefPipesModule } from 'app/pipes/chef-pipes.module';
 import { NgrxStateAtom } from 'app/ngrx.reducers';
-import { customMatchers } from 'app/testing/custom-matchers';
 import { FeatureFlagsService } from 'app/services/feature-flags/feature-flags.service';
 import { IAMType } from 'app/entities/policies/policy.model';
 import { GetProjectSuccess } from 'app/entities/projects/project.actions';
@@ -128,7 +127,6 @@ describe('ProjectListComponent', () => {
       }
     }));
 
-    jasmine.addMatchers(customMatchers);
     fixture = TestBed.createComponent(ProjectDetailsComponent);
     component = fixture.componentInstance;
     fixture.detectChanges();
--- a/components/automate-ui/src/app/pages/project/details/project-details.component.spec.ts
+++ b/components/automate-ui/src/app/pages/project/details/project-details.component.spec.ts
@@ -11,7 +11,7 @@ import { GetProjectSuccess } from 'app/entities/projects/project.actions';
 import { projectEntityReducer } from 'app/entities/projects/project.reducer';
 import { Project } from 'app/entities/projects/project.model';
 import { Rule } from 'app/entities/rules/rule.model';
-import { ProjectDetailsComponent } from './project-details.component';
+import { ProjectDetailsComponent, ProjectTabName } from './project-details.component';
 
 describe('ProjectDetailsComponent', () => {
   let component: ProjectDetailsComponent;
@@ -145,11 +145,14 @@ describe('ProjectDetailsComponent', () => {
     });
 
     it('shows/hides sections when based on selection', () => {
-      component.onTabChange({ target: { value: 'details' } });
+      let tabName: ProjectTabName = 'details';
+      component.onTabChange({ target: { value: tabName } });
+      expect(component.showTab(tabName)).toBeTruthy();
       expect(component.showTab('rules')).toBeFalsy();
-      expect(component.showTab('details')).toBeTruthy();
-      component.onTabChange({ target: { value: 'rules' } });
-      expect(component.showTab('rules')).toBeTruthy();
+
+      tabName = 'rules';
+      component.onTabChange({ target: { value: tabName } });
+      expect(component.showTab(tabName)).toBeTruthy();
       expect(component.showTab('details')).toBeFalsy();
     });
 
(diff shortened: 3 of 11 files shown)


components/automate-ui/src/app/pages/project/details/project-details.component.ts
```diff
@@ -89,6 +83,13 @@ export class ProjectDetailsComponent implements OnDestroy {
       });
   }
 
+  ngOnInit(): void {
+    this.projectForm = this.fb.group({
+      // Must stay in sync with error checks in project-details.component.html
```
srenatus: Is _now_ a good time to refactor this? 😉

msorens: 🤔 I'm not understanding your suggestion. That comment is reflecting upon an unavoidable consequence of how sometimes you have to refer to something in the code and in the template.

srenatus: Sorry, I've misread that. Skimming this, I thought it was another occurrance of the "must stay in sync with other thing" where we had agreed that it would be good to make it a constant.

components/automate-ui/src/app/pages/project/details/project-details.component.ts
```diff
@@ -118,28 +128,26 @@ export class ProjectDetailsComponent implements OnDestroy {
...
 
-  showDeleteRule(rule: Rule): boolean {
-    return rule.edits !== 'staging';
+  showDeleteRule(): boolean {
+    return true; // TODO: return false when *project* status is "updating..."
   }
 
   showProjectLink(): boolean {
-    return _find(this.rules, ['edits', 'staging']) ? true : false;
+    return find(['status', RuleStatus.Staged], this.rules) ? true : false;
```
srenatus: If we don't care about _what is returned_, let's use [`some`](https://gist.github.com/jfmengels/6b973b69c491375117dc#_somepredicate-collection) instead:
```suggestion
    return some(['status', RuleStatus.Staged], this.rules);
```

msorens: Nice!

msorens: `some` has the added benefit of returning a boolean, too!

components/automate-ui/src/app/pages/project/details/project-details.component.ts
```diff
@@ -20,19 +20,29 @@ import {
...
 
+export enum ProjectTabNames {
+  Rules = 'rules',
+  Details = 'details'
+}
+
+export enum RuleStatus {
+  Applied = 'applied',
+  Staged = 'staged'
+}
```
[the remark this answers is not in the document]
msorens: String literal types had slipped my mind... much obliged! I like that much better.

components/automate-ui/src/app/pages/project/details/project-details.component.ts
srenatus: yup that was my point actually 😉

components/automate-ui/src/app/entities/rules/rule.model.ts
```diff
@@ -1,14 +1,28 @@
...
+
+export interface KVPair {
+  key: string;
+  value: string;
+}
+
+// Using array of pairs instead of a hashmap to allow ordering
+export type RuleTypeMappedObject = {
+  [K in 'node' | 'event']: KVPair[]
+};
```
msorens: Tightened up type constraints a lot, with the help of these. 

The first few are string literal types. (Thanks @srenatus for reminding me about those!)

The last one is a `mapped type` that allows defining a complex typed constant in rule.reducer.ts. (More info: http://www.typescriptlang.org/docs/handbook/advanced-types.html#mapped-types)

components/automate-ui/src/app/pages/project/details/project-details.component.scss
```diff
@@ -25,6 +25,11 @@ chef-tab-selector {
 }
 
 form {
+
+  #update-name {
+    width: 240px;
+  }
```
msorens: Keeps the error pop-up width constrained to match the input field width.

components/automate-ui/src/app/pages/project/details/project-details.component.spec.ts
```diff
@@ -23,30 +22,30 @@ describe('ProjectListComponent', () => {
     type: <IAMType>'CHEF_MANAGED'
   };
   const rules = [
-    {
+    <Rule>{
```
msorens: Makes sure all the content is valid.

components/automate-ui/src/app/pages/project/details/project-details.component.ts
```diff
@@ -1,38 +1,40 @@
...
 import { Project } from 'app/entities/projects/project.model';
 import { GetProject, UpdateProject } from 'app/entities/projects/project.actions';
 import { GetRulesForProject, DeleteRule } from 'app/entities/rules/rule.actions';
-import { Rule } from 'app/entities/rules/rule.model';
+import { Rule, RuleStatus } from 'app/entities/rules/rule.model';
 import {
   allRules
 } from 'app/entities/rules/rule.selectors';
 
+type ProjectTabNames = 'rules' | 'details';
```
msorens: Another string literal type for the tabs on the page--suggest we use this for all the pages with tabs.

components/automate-ui/src/app/pages/project/details/project-details.component.ts
```diff
@@ -118,28 +120,26 @@ export class ProjectDetailsComponent implements OnDestroy {
...
-      ? 'Edits pending'
-      : 'Applied';
+  getEditStatus(rule: Rule): string {
+    return rule.status === 'staged' ? 'Edits pending' : 'Applied';
   }
 
-  showDeleteRule(rule: Rule): boolean {
-    return rule.edits !== 'staging';
+  showDeleteRule(): boolean {
+    return true; // TODO: return false when *project* status is "updating..."
```
msorens: This comes in a later sub-task.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -11,29 +11,29 @@
...
-              <option value='EVENT'>Event</option>
+            <span class="label">Resource Type <span aria-hidden="true">*</span></span>
+            <select id="create-type-dropdown" ngDefaultControl formControlName="type">
+              <option value='node'>Node</option>
+              <option value='event'>Event</option>
             </select>
           </label>
           <chef-error
-            *ngIf="ruleForm.get('type').hasError('required') && ruleForm.get('type').dirty"
+            *ngIf="ruleForm.get('type').hasError('required') && ruleForm.get('type').touched"
```
msorens: We use `dirty` on plain inputs, but it turns out that `touched` is much better for `<select>` elements. That way, tabbing through it without selecting a value activates the error display.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -11,29 +11,29 @@
...
           >Resource Type is required</chef-error>
-          <small class="help">Resource types cannot be changed after the rule is created.</small>
         </chef-form-field>
+        <small class="help">Resource types cannot be changed after the rule is created.</small>
```
msorens: Moved outside `<chef-form-field>` to get (presumably) desired behavior of the error pop-up immediately under the field. Note that physically in this file the `<chef-error>` was immediately under the field, even before my change. But it did not render in that order!

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -51,53 +51,57 @@
...
-                  <option 
+                  <option
                     *ngFor="let option of attributes[ruleForm.get('type').value.toLowerCase()]; let i = index"
                     [value]="option.key"
                   >{{ option.value }}</option>
                 </select>
               </label>
+              <chef-error
+                *ngIf="ruleForm.get('conditions').controls[i].get('attribute').hasError('required') && ruleForm.get('conditions').controls[i].get('attribute').touched"
+              >Attribute is required</chef-error>
```
msorens: Recipe on how to get an indexed error pop-up.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -51,53 +51,57 @@
...
-                <select ngDefaultControl
-                  formControlName="operator"
-                  (change)="updateConditionValues(condition)"
-                >
-                  <option value='EQUALS'>equals</option>
-                  <option value='MEMBER_OF'>member of</option>
+                <select ngDefaultControl formControlName="operator">
+                  <option *ngFor="let operator of operators"
+                    [value]="operator.key"
+                  >{{ operator.value }}</option>
```
msorens: I prefer not to have duplicate source of truth, so moved these values to the backing code.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -51,53 +51,57 @@
...
-            
+
             <chef-form-field class="attribute-field">
               <label>
                 <span class="label">Value</span>
                 <chef-input ngDefaultControl
                   formControlName="values"
-                  [value]="getConditionValue(condition.controls.values.value)"
-                  (change)="updateConditionValues(condition)"
+                  [value]="condition.controls.values.value"
```
msorens: Little change; long story. First, code originally was adjusting the actual value to be either a string or a string array based on operator (`equals` or `member of`). But it only did that if you flip-flopped the operator; did not do it if you changed the value without flipping the operator... I corrected that, and added some tests, and then realized did not want all that anyway, because the API always wants an array for `values`.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -51,53 +51,57 @@
...
               </small>
-
             </chef-form-field>
+
             <chef-form-field class="attribute-field">
               <label>
                 <span *ngIf="showAndLabel(i)" class="label">AND</span>
               </label>
-              <chef-button tertiary *ngIf="showDelete(i)" (click)="deleteCondition(i)">
+              <chef-button tertiary *ngIf="showDelete()" (click)="deleteCondition(i)">
```
msorens: No longer row specific; only cares about whether there is more than one row or not. If so, all rows get a button.

components/automate-ui/src/app/pages/project/rules/project-rules.component.spec.ts
```diff
@@ -175,47 +163,74 @@ describe('ProjectRulesComponent', () => {
       component.ruleForm.get('conditions').setValue([
         {
           attribute: 'CHEF_ORGS',
-          operator: 'EQUAL',
+          operator: <ConditionOperator>'EQUALS',
```
msorens: I found a flaw in TypeScript! No, just a flaw in my understanding. Adding the type cast here does not do anything! That is, it would be perfectly happy with `<ConditionOperator>'foobar'` and I do not see why it does not cause a compilation error. However, to me it is still valuable to put the type cast here: it tells me loud and clear where this comes from (and if you hover over `ConditionOperator` it tells you the allowed value, too.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -35,12 +33,24 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
   public ruleId: string;
   public ruleForm: FormGroup;
   public rule: Rule = <Rule>{};
-  public conditions: any[];
+  public conditions: FormGroup[];
```
msorens: Squashed another `any` !

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -100,8 +111,9 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
 
   ngOnInit(): void {
     this.ruleForm = this.fb.group({
-      name: [this.rule.name || '', Validators.required],
-      type: [this.rule.type || '', Validators.required],
+      // Must stay in sync with error checks in project-rules.component.html
+      name: ['', [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]],
+      type: [{ value: this.rule.type || '', disabled: this.editingRule } , Validators.required],
```
msorens: This little gem had not been used in the code base before AFAIK. Can't take the credit, though: Angular helped me find what to do. Initially I just put `[disabled] = "editingRule"` in the template but at runtime Angular patiently explained that could cause "changed after checked" errors and should use this pattern instead. This initializes the element with the `disabled` property given here.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -131,9 +143,10 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
 
   createCondition(attribute = '', operator = '', values = ''): FormGroup {
     return this.fb.group({
+      // Must stay in sync with error checks in project-rules.component.html
       attribute: [attribute, Validators.required],
       operator: [operator, Validators.required],
-      values: [values, Validators.required]
+      values: [values, [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]]
```
msorens: All input fields should have both `required` and `pattern`/`NON_BLANK` validators. One would think that a pattern should cover both cases, but no, for some reason it does not, even if the pattern explicitly disallows an empty value.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -152,17 +165,19 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
...
 
   populateConditions() {
     this.conditions = [];
 
     if (this.rule.conditions && this.rule.conditions.length !== 0) {
       this.rule.conditions.forEach(c => {
-        this.conditions.push(this.createCondition(c.attribute, c.operator, c.values));
+        this.conditions.push(
+          // Convert values array to display string
+          this.createCondition(c.attribute, c.operator, c.values.join(', ')));
```
msorens: Remember my story about conditions earlier? Fairly straightforward now. Here's one direction, taking the storage object and converting it to a string for display.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -171,36 +186,44 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
...
+    const ruleValue = this.ruleForm.value;
+    const conditions: Condition[] = [];
+    ruleValue.conditions.forEach(c => {
+      conditions.push(<Condition>{
+        attribute: c.attribute,
+        operator: c.operator,
+          // Convert values string to storage format
+        values: c.operator === <ConditionOperator>'EQUALS'
+          ? [c.values.trim()]
+          : c.values.split(/,\s*/).map(v => v.trim())
```
msorens: ... And this is the other direction: translating the displayed string value back to an appropriate array for storage.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -171,36 +186,44 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
...
-    }
-  }
-
-  getConditionValue(value): string {
-    return (typeof value === 'string') ? value : value.join(', ');
-  }
-
   createRule() {
-    this.store.dispatch(new CreateRule({project_id: this.project.id, rule: this.ruleForm.value}));
+    // TODO: Generate the id!
```
msorens: Coming soon; probably another PR...

components/automate-ui/src/app/pages/project/details/project-details.component.ts
```diff
@@ -118,28 +120,26 @@ export class ProjectDetailsComponent implements OnDestroy {
...
   }
 
   showProjectLink(): boolean {
-    return _find(this.rules, ['edits', 'staging']) ? true : false;
+    return some(['status', <RuleStatus>'staged'], this.rules);
```
srenatus: ❓ The `<RuleStatus>` is not inferred? Hmm!

components/automate-ui/src/app/pages/project/details/project-details.component.ts
bcmdarroch: Oo I'll update team details

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -1,25 +1,23 @@
...
+import { Regex } from 'app/helpers/auth/regex';
+import {
+  Rule, RuleTypeMappedObject, Condition, ConditionOperator, KVPair
+} from 'app/entities/rules/rule.model';
 import {
-  GetRule,
-  GetRulesForProject,
-  CreateRule,
-  UpdateRule
+  GetRule, GetRulesForProject, CreateRule, UpdateRule
```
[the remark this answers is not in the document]
bcmdarroch: I thought it depended on the length of the import line

susanev:
@msorens 
- the `AND` label and delete button do not seem to be in the correct place, should be vertically centered on the value input
- can we please get periods at the end of those error messages
- im now thinking we should just hide that conditions section until a resource type is selected because its kinda weird that you can't pick an attribute at all until you do that. if that's easy to do would be great to do now, but if not we can save it until later.

and.. not directly related. but im seeing that there is a `line-height` on all our inputs and that's cutting off the descenders, can we remove that?

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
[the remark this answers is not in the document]
msorens: Yes, not sure if we have an absolute convention, but extending on what @bcmdarroch said, my rule of thumb is, if it is a short list, go horizontal, and once it gets too long a line, then go vertical. I think other folks tend to do this as well...?  🤷‍♂

components/automate-ui/src/app/pages/project/details/project-details.component.ts
msorens: I investigated this and other use cases of type casting. Will post some notes in Slack on Monday...

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -171,36 +186,44 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
...
-    const updatedRule = this.ruleForm.value;
-    updatedRule.id = this.rule.id;
-    updatedRule.project_id = this.rule.project_id;
+    const updatedRule = this.convertToRule();
     this.store.dispatch(new UpdateRule({ rule: updatedRule }));
     this.store.dispatch(new GetRulesForProject({ project_id: this.rule.project_id }));
   }
 
+  convertToRule(): Rule {
+    const ruleValue = this.ruleForm.value;
```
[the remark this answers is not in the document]
msorens: I agree in general; went this way to be consistent with rest of the functions in the file.

components/automate-ui/src/app/pages/project/details/project-details.component.spec.ts
```diff
@@ -145,11 +145,14 @@ describe('ProjectDetailsComponent', () => {
     });
 
     it('shows/hides sections when based on selection', () => {
-      component.onTabChange({ target: { value: 'details' } });
+      let tabName: ProjectTabName = 'details';
+      component.onTabChange({ target: { value: tabName } });
```
srenatus: 🤔 This makes we wonder we can't give `onTabChange` a better type instead. Instead of `onTabChange(event)` we could go with
```typescript
onTabChange(event: { target: { value: ProjectTabName } })
```
or something like that (haven't tested it). I can see how its called with a more generic event, but since this is a) erased in transpilation, and b) only used this way, we could tighten the types here, I think.

srenatus:
[typo] `s/robutstness/robustness/` in your last commit message

components/automate-ui/src/app/pages/project/details/project-details.component.ts
```diff
@@ -139,7 +139,9 @@ export class ProjectDetailsComponent implements OnInit, OnDestroy {
   }
 
   showProjectLink(): boolean {
-    return some(['status', <RuleStatus>'staged'], this.rules);
+    const statusPropertyName = 'status';
```
srenatus: 💭I can see why you did this, but I think this is one step too far. 😉

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -40,13 +40,17 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
   public editingRule = false;
   private isDestroyed: Subject<boolean> = new Subject<boolean>();
 
+  // These constants ensure type safety
+  private equals_op: ConditionOperator = 'EQUALS';
+  private member_of_op: ConditionOperator = 'MEMBER_OF';
```
srenatus: Hmmm I think I'd rather type `public opteratos` and use these as-is.

susanev:
@msorens few more things...
- if you refresh the page when on the page to create/update an ingest rule the top-nav comes back
- there should be 18px of space between the _Add Condition_ button and the attribute label, currently, there is only 8px of space.
- (we can save this for later if we want too) the 3-dot menus on the projects list page load considerably slower than everything else

components/automate-ui/src/app/pages/project/details/project-details.component.spec.ts
msorens: Right you are--that worked just fine. 👍

Issue #599, opened 2019-06-14
flaky test: compliance TestReportingListSuggestions


vjeffrey:
# User Story

Noticed this one being flaky in the pipeline:
```

FAIL: TestReportingListSuggestions (0.69s) | 0s
-- | --
  | --- PASS: TestReportingListSuggestions/Projects:_"environment"_suggestions,_user_has_access_to_all_projects (0.00s)
  | --- FAIL: TestReportingListSuggestions/Projects:_"environment"_suggestions,_user_has_access_to_one_project_with_reports (0.00s)
  | Error Trace:	reporting_server_list_suggestions_test.go:540
  | Error:      	lengths don't match: 2 != 0
  | Test:       	TestReportingListSuggestions/Projects:_"environment"_suggestions,_user_has_access_to_one_project_with_reports
  | --- FAIL: TestReportingListSuggestions/Projects:_"environment"_suggestions,_user_has_access_to_some_projects_with_reports (0.00s)
  | Error Trace:	reporting_server_list_suggestions_test.go:540
  | Error:      	lengths don't match: 4 != 0
  | Test:       	TestReportingListSuggestions/Projects:_"environment"_suggestions,_user_has_access_to_some_projects_with_reports
  | --- PASS: TestReportingListSuggestions/Projects:_"environment"_suggestions,_user_has_access_to_projects_without_reports (0.00s)
  | --- FAIL: TestReportingListSuggestions/Projects:_"environment"_suggestions,_user_has_access_to_one_project_with_reports_and_unassigned_reports (0.00s)
  | Error Trace:	reporting_server_list_suggestions_test.go:540
  | Error:      	lengths don't match: 3 != 5
  | Test:       	TestReportingListSuggestions/Projects:_"environment"_suggestions,_user_has_access_to_one_project_with_reports_and_unassigned_reports
  | --- PASS: TestReportingListSuggestions/Projects:_"environment"_suggestions,_user_has_access_to_some_projects_with_reports_and_unassigned_reports (0.00s)
  | --- FAIL: TestReportingListSuggestions/Projects:_"environment"_suggestions,_user_has_access_to_projects_without_reports_and_unassigned_reports (0.00s)
  | Error Trace:	reporting_server_list_suggestions_test.go:540
  | Error:      	lengths don't match: 1 != 5
  | Test:       	TestReportingListSuggestions/Projects:_"environment"_suggestions,_user_has_access_to_projects_without_reports_and_unassigned_reports
[...]

PR #1022, opened 2019-07-23
Retry inserts in integration tests.

### LINKED ISSUE #1008: flaky test: TestListReports
## Describe the bug
There's a flaky test that has been particularly annoying these days:
```
FAIL: TestListReports (0.06s)
    require.go:794:
            Error Trace:    reporting_server_list_reports_test.go:39
            Error:          Received unexpected error:
                            elastic: Error 409 (Conflict)
            Test:           TestListReports
```

## To Reproduce
- run the `compliance_integration` tests multiple times

## Expected behavior
Expect the tests to pass every time

## Screenshots
N/A

## Additional context
https://buildkite.com/chef/chef-automate-master-verify-private/builds/3507#b5eabd1e-6008-4dff-aa2e-d18397745f58/829

phiggins:
Attempt to fix #1008

### DIFF
--- a/components/compliance-service/ingest/ingestic/ingestic.go
+++ b/components/compliance-service/ingest/ingestic/ingestic.go
@@ -135,11 +135,10 @@ func (backend *ESClient) InsertInspecSummary(ctx context.Context, id string, end
 		Refresh("false").
 		Do(ctx)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "InsertInspecSummary")
 	}
 
-	err = backend.setDailyLatestToFalse(ctx, data.NodeID, id, index)
-	return err
+	return backend.setDailyLatestToFalse(ctx, data.NodeID, id, index)
 }
 
 func (backend *ESClient) InsertInspecReport(ctx context.Context, id string, endTime time.Time, data *relaxting.ESInSpecReport) error {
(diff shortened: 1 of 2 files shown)


phiggins:
The context @vjeffrey mentioned above:

> I think the problems with the flaky integration tests are happening because we do a lot of async elastic operations for performance reasons and I think we have real concurrency bugs. Some of the failures we've been seeing lately are because of how the code handles the daily_latest: a new document is inserted and then every other document belonging to that node is updated to have daily_latest set to false, both of these asynchronously. When this is done in a tight loop, like in the tests, there is a lot of opportunity for conflicts.
> If we were using a relational database I'd say we should order by a date column rather than using that extra flag, but I don't really know enough about how we're doing that to say whether it's possible here.
> I have a PR to add some retries in the test code but I'm not really comfortable modifying the actual service code for this because I'm not sure how I'd know if I was introducing a performance regression.

phiggins:
Running these tests in a loop on my local machine I am still seeing failures, but they are all failing in `TestReportingListSuggestions` which is most likely related to these ES issues but would be a totally separate debugging session.

vjeffrey:
ya, the flaky reporting list suggestions test is documented here: https://github.com/chef/automate/issues/599

Issue #864, opened 2019-07-11
reports with bad data structures == inability to view node details + failure to build notif


vjeffrey:
original report + escalation from customer: https://chefio.atlassian.net/browse/A2-909
https://getchef.zendesk.com/agent/tickets/22761

## Describe the bug
Customer was using a custom profile that inherits controls from multiple CIS profiles. For whatever reason, the report generated by that profile resulted in a couple instances of:
```
   "refs": [
        {
          "ref": []
        }
      ],
```
(it's usually just `"refs":[]`)

this resulted in an inability to view node details, and the following error from notifications: "Could not build notifications InSpec event: json: cannot unmarshal array into Go value of type string"

## To Reproduce
At this point, I do not have a profile that will repro the behaviour.  But there is a scan report attached in the original jira ticket (I can't post here b/c customer info) that can be used to repro. I used the "make send-ingest-report" functionality to send the report in and observed the error. I then took out the bad refs structures mentioned above and no more errors.

## Expected behavior
Report goes through and we ignore the bad structure.
In a perfect world the bad structure in the report would never happen, and we would reject reports with said bad structure. However I'm not sure that's realistic, and I know it's not friendly to our users (they didn't put the bad structure there...they just made a custom profile using our profiles).

## Screenshots
<img width="1407" alt="Screen Shot 2019-07-11 at 09 13 36" src="https://user-images.githubusercontent.com/10341541/61062851-316e4480-a3bc-11e9-9fda-a3c95ff6f647.png">

## Versions (please complete the following information):
- Browser: chrome 75.0.3770.100 (Official Build) (64-bit)
- Automate Build Number: 20190711141902

vjeffrey:
tested in inplace-upgrade-dev. everything is working

PR #1039, opened 2019-07-23
Notifications not working InSpec Ref bug

lancewf:

Only two changed files
* components/notifications-client/builder/compliance_test.go
* components/notifications-service/server/priv/notifications.proto

There is a problem with custom InSpec reports where they sometimes have an array for the 'controls.refs.ref' instead of a string like below. 
```
 "controls": [
 	{
 		"title": "Checking for /etc/passwd",
 		"desc": "Checking for /etc/passwd desc",
 		"impact": 0.6,
 		"refs": [
 			{
 				"ref": []
 			}
 		],
```

For the notification-service we are going to remove the control.refs.ref. It is not being used and it can not be a generic because of the bug found with this PR https://github.com/chef/a2/pull/4364. Which I found because this test was failing. https://buildkite.com/chef/chef-automate-master-verify-private/builds/3631#f04c1c01-f7e6-495d-9a8d-d786ff38cb76 ( I am glad there was a test). 

1. `start_all_services`
1. Watch the compliance logs with `sl | grep "compliance-service.default" &`
1. Send an InSpec report with the ref bug `JSON_FILE=/src/components/notifications-client/builder/testdata/inspec-ref-bug-report.json send_inspec_example`
1. Ensure the error log `Could not build notifications InSpec event: json: cannot unmarshal array into Go value of type string` is displayed. 
1. Install the update with `rebuild components/compliance-service`
1. Resend the InSpec report with `JSON_FILE=/src/components/notifications-client/builder/testdata/inspec-ref-bug-report.json send_inspec_example`
1. Ensure the error log is not displayed. 

#864

- [x] Code actually executed?


components/notifications-client/builder/testdata/inspec-ref-bug-report.json
```diff
@@ -0,0 +1,230 @@
...
+        }
+      ],
+      "controls": [
+        {
+          "title": "Checking for /etc/passwd",
+          "desc": "Checking for /etc/passwd desc",
+          "impact": 0.6,
+          "refs": [
+            {
+              "ref": []
```
lancewf: This is the bug we are accounting for in the InSpec report.

components/notifications-service/server/priv/notifications.proto
```diff
@@ -96,7 +96,7 @@ message SourceLocation {
 }
 
 message Refs {
-  string ref = 1;
+  reserved 1; //string ref = 1; because the 'ref' can be an array also
```
lancewf: Removing the `ref` field fixes the problem.

Issue #1005, opened 2019-07-22
Teams: Add empty state to Users tab on Team detail pages


susanev:
## Overview

Currently, when a team has zero users we show an empty table, instead, we would like to hide the table and show an empty message with a call-to-action to add users.

## Design Details

[Designs on Zeplin](https://zpl.io/VQvRE59)

* When a team has zero users...
    * Hide the table and command bar
    * Add an empty message of _Add some users to get started!_
        * font-size: 18px; font-weight: 400
        * semantically a paragraph
        * 77px from chef-page-header to paragraph
        * 18px between paragraph and primary button
    * Primary button: _Add Users_

susanev:
Closing as it will be addressed in #794

Issue #794, opened 2019-07-08
Teams: General UI Fixes


susanev:
_Note: Please leverage the implementation from Projects and Tokens_

## Overview

In IAMv2 we are making updates to Teams, and this task aligns those updates to the UI. These design updates should only be displayed on v2, not v1.

## Design Details

[UX Designs in Zeplin](https://app.zeplin.io/project/5ba808703eb54372d860c737/dashboard?seid=5c65a8e7488d3b3c17ad9257)

* Team list view has columns Name and ID, in that order
* Create Team action results in a modal leveraging design already implemented for Projects and Tokens
* Team detail updates to heading, and summary
* Remove edit button on team detail pages
* Details tab added where the user can change the team name
* There should be 38px of space between the name input and the Save button
* Admins team name is changed to Admins

* When a team has zero users...
[Designs on Zeplin](https://zpl.io/VQvRE59)
    * Hide the table and command bar
    * Add an empty message of _Add some users to get started!_
        * font-size: 18px; font-weight: 400
        * semantically a paragraph
        * 77px from chef-page-header to paragraph
        * 18px between paragraph and primary button
    * Primary button: _Add Users_

srenatus:
![image.png](https://images.zenhubusercontent.com/5c0b885543d5c44ffa2f8003/089bf6bd-0721-4e9b-ba3b-4099a1552aad)

Looks like the name of the team with ID `admins` is still `admins`. Purposely?

Issue #1006, opened 2019-07-22
Teams: Align _Add User_ with Policy _Add Member_


susanev:
## Overview

To improve consistency and the user experience this task aligns the _Add User_ flow in Teams with the _Add Member_ flow in Policies.

## Design Details

[Designs in Zeplin](https://app.zeplin.io/project/5ba808703eb54372d860c737?seid=5cb5e8d2e6267e081e743067)

### Team Detail Page
* Change _Add User_ button to _Add Users_

### Add Users to Team Full-page 
* Heading: _Add Users to `<team name>`_
* Subheading: _Only local users can be added to teams._
* Table headings: Name, and Username
* Top of table displays the count of the number of users selected
* The table has name, and username columns; the links for the name should open in a new tab (because it breaks the user out of their flow)
* After adding users, send back to the team detail page with on-page user notification of _Added 1 user._ or _Added `<n>` users._
* On the case where there are no local users or all the local users have already been added to the team hide the table and display _There are no more local users to add, create some more!_ with _users_ being a link to`/users` that opens in a new tab

#### Buttons
* Primary button: _Add User_ when disabled, _Add User_ when one selected and enabled, Add _`<n>` Users_ when more than one user is selected and enabled; after user-action on the button the label becomes _Adding User..._ or _Adding Users..._ with the loading icon
* Tertiary button: _Cancel_

srenatus:
@tylercloke: ℹ️Assigned you since you've pull it from RFA.

srenatus:
The card says
>Automate only displays local users.

the UI says
> Only local users can be added to teams.

I'm assuming this has been discussed. (Just because I think the latter, what the UI displays, is a tad better :wink:)

susanev:
@srenatus yes, discussed. 😞. i will get better about updating things after discussions.

PR #1041, opened 2019-07-23
[automate-794] Teams: General UI Fixes

bcmdarroch:
Issue https://github.com/chef/automate/issues/794
We've brought the Teams (mostly) up to date with our current auth page styling.

### :+1: Definition of Done
1. New empty list state
Screen Shot 2019-07-23 at 12 03 45 PM
2. new heading and tabs on details page
Screen Shot 2019-07-23 at 12 04 08 PM
3. name updates are now part of the details tab
Screen Shot 2019-07-23 at 12 04 14 PM
4. new teams are now created in a modal, not on a different page like before
Screen Shot 2019-07-23 at 12 04 30 PM
5. that long admin team description has been migrated to `admins`
Screen Shot 2019-07-23 at 12 04 55 PM

1. log in to the UI
2. create a team with the modal
3. on the new team's details page, add users and update the team's name
4. play around the teams UI and make sure we didn't miss anything!

- took care of [this issue](https://github.com/chef/automate/issues/1005) as part of this work

- [ ] Code actually executed?


components/automate-ui/src/app/entities/teams/team.reducer.ts
```diff
@@ -96,7 +99,10 @@ export function teamEntityReducer(state: TeamEntityState = TeamEntityInitialStat
     }
 
     case TeamActionTypes.CREATE_FAILURE: {
-      return set('createStatus', EntityStatus.loadingFailure, state) as TeamEntityState;
+      return pipe(
+        set('createError', action.payload),
```
bcmdarroch: might rename these to `save` to correspond with the tokens selectors

components/automate-ui/src/app/page-components/user-table/user-table.component.scss
```diff
@@ -11,3 +13,13 @@ chef-table {
     justify-content: flex-end;
   }
 }
+
+.team-users-empty-state {
```
bcmdarroch: we'd like to make this `empty-state` rule shareable across components, but waiting on @chef/ui-team for some guidance. just adding it to `ui.component.scss` didn't cut it

components/teams-service/storage/postgres/datamigration/sql/03_update_admin_team_name.up.sql
```diff
@@ -0,0 +1,9 @@
+BEGIN;
+
+UPDATE teams
```
bcmdarroch: this change will persist if the user resets to v1, but we decided this was ok since we already have the editor and viewer teams persisting after reset.

components/automate-ui/src/app/page-components/user-table/user-table.component.html
```diff
@@ -1,42 +1,47 @@
 <app-authorized [allOf]="getPermissionsPath" [overrideVisible]="overridePermissionsCheck">
+  <ng-container *ngIf="(users$ | async)?.length === 0">
+    <div class="team-users-empty-state">
+      <p>Add some users to get started!</p>
```
bcmdarroch: tiny issue with this where the empty state message will flash again very briefly after adding some users to the team. rather than fixing that in this PR, we plan to pull in this issue as a follow-up https://github.com/chef/automate/issues/1006 which should make the problem go away

components/automate-ui/e2e/deprecated-routes.e2e-spec.ts
```diff
@@ -13,7 +13,6 @@ describe('Deprecated routes redirect to the correct new route for backwards comp
     ['/admin/tokens/my-object',                '/settings/tokens/my-object'],
     ['/admin/teams',                           '/settings/teams'],
     ['/admin/teams/my-object',                 '/settings/teams/my-object'],
-    ['/admin/add-team',                        '/settings/add-team'],
```
tylercloke: It would be nice if our create modals were in a route fragment like `/teams#create-team` similar as we do for tabs in the details pages. Navigating to that route would go to the list page and open the create modal. Then we could route this deprecated route to that and not break backwards compatibility.

components/teams-service/storage/postgres/datamigration/sql/03_update_admin_team_name.up.sql
tylercloke: If they are in v2.x mode, any new datamigrations should happen teams service restart (which happens on upgrade):

https://github.com/chef/automate/blob/master/components/teams-service/storage/postgres/postgres.go#L45-L49

susanev:
@bcmdarroch @blakestier y'all mind updating the docs team flow and screenshots?
https://automate.chef.io/docs/teams/

There is also mention of an _Add Team_ button, which is now _Create Team_.

components/teams-service/storage/postgres/datamigration/sql/03_update_admin_team_name.up.sql
blakestier: @tylercloke We're a little confused on your comment. (Also we've moved this to be a permanent migration anyway so maybe it's moot)

components/automate-ui/src/app/pages/team-management/team-management.component.ts
```diff
@@ -39,23 +53,88 @@ export class TeamManagementComponent implements OnInit {
...
+          if (state === EntityStatus.loadingFailure) {
+            const pendingCreateError = new Subject<boolean>();
+            this.store.pipe(
+              select(createError),
+              filter(identity),
+              takeUntil(pendingCreateError))
+              .subscribe((error) => {
+                pendingCreateError.next(true);
+                pendingCreateError.complete();
+                if (error.status === 409) {
```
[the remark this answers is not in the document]
blakestier: Hmm yep we copy-pasta'd, so fixing this in 3 places 😂

components/automate-cli/pkg/adminmgmt/adminmgmt.go
```diff
@@ -88,7 +88,7 @@ func CreateUserOrUpdatePassword(ctx context.Context,
 // found or created.
 func CreateAdminTeamIfMissing(ctx context.Context,
 	apiClient client.APIClient, dryRun bool) (string, bool, error) {
-	descr := "Members of the admins team, by default, have access to all parts of the API."
+	descr := "admins"
 	return EnsureTeam(ctx, "admins", descr, apiClient, dryRun)
```
srenatus:
```suggestion
	return EnsureTeam(ctx, "admins", "admins", apiClient, dryRun)
```

components/automate-ui/src/app/pages/team-management/team-management.component.html
```diff
@@ -3,35 +3,46 @@
...
+      [createForm]="createTeamForm"
+      (close)="closeCreateModal()"
+      (createClicked)="createTeam()">
+    </app-create-object-modal>
+
     <app-delete-object-modal
-      [visible]="modalVisible"
+      [visible]="deleteModalVisible"
       objectNoun="team"
       [objectName]="(iamMajorVersion$ | async) === 'v1' ? teamToDelete?.id : teamToDelete?.name"
```
srenatus: Hmm it's used so often, we could introduce helper functions, I guess: `isV1()`, `isV2()`.

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -182,23 +195,34 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
     }));
   }
 
-  updateTeam(): void {
-    const newTeam = <Team>Object.assign(
-      {}, this.team, { name: this.editForm.get('teamName').value });
-    this.store.dispatch(new UpdateTeam(newTeam));
+  public handleNameChange(): void {
+    this.disableSave = this.updateNameForm.controls['name'].value === this.team.name;
```
srenatus: I'd prefer we used the formgroup's means to determine if this is dirty or not instead of re-creating that logic using an instance variable.

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -182,23 +195,34 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
   }
 
-  setEditMode(isEditMode: boolean): void {
-    this.editForm.patchValue(
-      {
-        teamName: this.team.name
+  public saveNameChange(): void {
+    this.saveInProgress = true;
+    const teamToSave = <Team>Object.assign({}, this.team);
```
srenatus: [style] How about this?
```typescript
const name: string = this.updateNameForm.controls['name'].value.trim();
this.store.dispatch(new UpdateTeam({ ...this.team, name });
```

components/automate-ui/src/app/pages/team-management/team-management.component.ts
```diff
@@ -19,12 +27,18 @@ export class TeamManagementComponent implements OnInit {
   public sortedTeams$: Observable<Team[]>;
   public loading$: Observable<boolean>;
   public teamToDelete: Team;
-  public modalVisible = false;
+  public deleteModalVisible = false;
+  public createModalVisible = false;
+  public createTeamForm: FormGroup;
+  public creatingTeam = false;
+  public conflictErrorEvent = new EventEmitter<boolean>();
   public iamMajorVersion$: Observable<string>;
```
srenatus: We've got a type `IAMMajorVersion` in `entities/policy.model`. Let's use that, it's more specific than `string`.

components/teams-service/storage/memstore/memstore.go
```diff
@@ -303,7 +303,7 @@ func (m *memstore) teamIncludesUser(teamID uuid.UUID, userID string) bool {
 
 func (m *memstore) EnsureAdminsTeam(ctx context.Context) error {
 	if _, err := m.StoreTeam(ctx, storage.AdminsTeamName,
-		"Members of the admins team, by default, have access to all parts of the API."); err != nil {
+		"admins"); err != nil {
```
srenatus: [style] I'm not sure the newline is warranted now

components/teams-service/storage/postgres/migration/sql/08_update_admin_team_name.up.sql
```diff
@@ -0,0 +1,9 @@
+BEGIN;
+
+UPDATE teams
+    SET
+        description = 'admins'
+    WHERE
+        name = 'admins';
```
srenatus: ❓ This will affect IAMv1, and we're ok with that. Just stating the obvious 😉

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -66,6 +72,25 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
+              break;
+            }
+            default: { // If the user passed an invalid fragment or #users
+              this.tabValue = 'users';
+            }
+          }
+        } else {
+          // Default to definition in the case of no fragment.
+          this.tabValue = 'users';
+        }
```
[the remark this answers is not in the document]
srenatus: _+1_ on the code simplification; _+/- 0_ on the named constants 😉

components/automate-chef-io/content/docs/teams.md
```diff
@@ -33,7 +33,7 @@ export TOK=`chef-automate admin-token`
 
 #### Create a Team from Chef Automate
 
-As an administrative user, you can create a team in the UI from the **Settings** tab. Select _Teams_ in the sidebar then use the **Add Team** button:
+As an administrative user, you can create a team in the UI from the **Settings** tab. Select _Teams_ in the sidebar then use the **Create Team** button:
```
mjingle:
```suggestion
As an administrative user, you can create a team in the UI from the **Settings** tab. Select _Teams_ in the sidebar, and then use the **Create Team** button:
```

mjingle: Does the "Add Local Team" image below need to be updated also?

components/automate-ui/src/app/pages/team-management/team-management.component.ts
bcmdarroch: good catch, we're using string in some other places too

components/teams-service/storage/postgres/migration/sql/08_update_admin_team_name.up.sql
bcmdarroch: yup we quadrupled checked with @susanev 😂

components/automate-ui/src/app/pages/team-details/team-details.component.spec.ts
```diff
@@ -61,7 +61,9 @@ describe('TeamDetailsComponent', () => {
         MockComponent({ selector: 'chef-option' }),
         MockComponent({ selector: 'chef-heading' }),
         MockComponent({ selector: 'chef-subheading' }),
-        MockComponent({ selector: 'chef-tab-selector' }),
+        MockComponent({ selector: 'chef-tab-selector',
+          inputs: ['value', 'routerLink', 'fragment']
+        }),
         TeamDetailsComponent
       ],
```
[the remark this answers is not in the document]
bcmdarroch: What functionality would you suggest testing?

components/automate-ui/src/app/page-components/user-table/user-table.component.scss
bcmdarroch: seems like the best move would be to create an empty state component with the styles we want, so I'll make a card for that this week

components/automate-ui/e2e/deprecated-routes.e2e-spec.ts
bcmdarroch: We're going to leave this for now, on the assumption there aren't that many (if any) customers that have bookmarked these old routes and we'd rather stay consistent with the other list pages, which don't use fragments. (though that could change)

components/automate-ui/src/app/pages/team-details/team-details.component.ts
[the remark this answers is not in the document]
msorens: I do agree with @srenatus per his comments in my PR--better to use string literal types. See my implementation here: https://github.com/chef/automate/pull/1020/files#diff-513bbcc13778e318726a2b9bef418a2cR23

components/automate-ui/e2e/deprecated-routes.e2e-spec.ts
tylercloke: How come we don't think customers have this specific route bookmarked? I'm fine with it but in general let's not break backwards compatibility. Would be nice to add the fragment functionality regardless but there are a bunch of places to do that so I'll follow up with @susanev if that's behavior we want and we can make a card.

components/automate-ui/src/app/pages/team-details/team-details.component.spec.ts
```diff
@@ -76,17 +90,125 @@ describe('TeamDetailsComponent', () => {
...
     expect(component).toBeTruthy();
   });
 
+  it('defaults to showing users section', () => {
+    expect(component.tabValue).toBe('users');
+  });
+
+  it('show users section when users tab is selected', () => {
+    component.onSelectedTab({ target: { value: 'users' } });
+    expect(component.tabValue).toBe('users');
```
msorens: Just realized this on my PR and the same on your PR: this test passes, yes, but if you change `users` to `foo` on both L120 and L121 it will still pass, and that feels just slightly bad. Here's a fix:
```
    const tabName: TeamTabNames = 'users';
    component.onSelectedTab({ target: { value: tabName } });
    expect(component.tabValue).toBe(tabName);
```
The first line will throw a compile error if `users` is switched to `foo`.

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -32,18 +33,24 @@ import {
 
 const TEAM_DETAILS_ROUTE = /^\/settings\/teams/;
 
+type TeamTabNames = 'users' | 'details';
```
msorens: I violated a sort-of-standard notion that type names should be singular in my PR, and you've got the same here. Because when you instantiate something of this type, `const t: TeamTabNames` does not read right. That says "here's a constant t that is a TeamTabNames". Let's change this to singular.

components/automate-ui/src/app/pages/team-details/team-details.component.html
```diff
@@ -9,86 +9,72 @@
...
-      </div>
-    </section>
+      </section>
+    </ng-container>
+    <ng-container *ngIf="tabValue === 'details'">
+      <section class="page-body">
+        <form [formGroup]="updateNameForm">
+          <chef-form-field>
+            <label *ngIf="isV1">Description *</label>
+            <label *ngIf="!isV1">Name *</label>
```
msorens: Need to play nice with screen readers.
```suggestion
            <label *ngIf="!isV1">Name <span aria-hidden="true">*</span></label>
```

Same change on both L65 and L66.

components/automate-ui/src/app/pages/team-details/team-details.component.html
```diff
@@ -9,86 +9,72 @@
...
+            <input chefInput formControlName="name" type="text" (keyup)="keyPressed()">
```
msorens: Needs an associated error element:
```
<chef-error *ngIf="(updateNameForm.get('name').hasError('required') || updateNameForm.get('name').hasError('pattern')) && updateNameForm.get('name').dirty">
  {{ descriptionOrName() }} is required.
</chef-error>
```
Note that because the way @susanev 	wants these error messages to be (`<displayed-name> is required`), you might want to have a property `descriptionOrName()` as `return isV1? 'Description' : 'Name';`. With that in place, then L65-66 above can be simplified to a single line:
```
<label>{{ descriptionOrName() }} <span aria-hidden="true">*</span></label>
```

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -52,9 +59,25 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
 
   constructor(private store: Store<NgrxStateAtom>,
-    private fb: FormBuilder
+    public fb: FormBuilder,
+    private router: Router
   ) {
-    this.createForms(this.fb);
+    this.updateNameForm = fb.group({
+      // Must stay in sync with error checks in team-details.component.html
+      name: ['Loading...', Validators.required]
```
msorens: Needs to also check for all whitespace. (This matches the suggestion I made on the template side.)
```suggestion
      name: ['Loading...', [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]]
```

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -182,23 +199,34 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
   }
 
-  setEditMode(isEditMode: boolean): void {
-    this.editForm.patchValue(
-      {
-        teamName: this.team.name
+  public saveNameChange(): void {
+    this.saveSuccessful = false;
+    this.saving = true;
+    const name: string = this.updateNameForm.controls['name'].value.trim();
```
msorens: Suggest use non-strings where equivalent:
```suggestion
    const name: string = this.updateNameForm.controls.name.value.trim();
```

components/automate-ui/src/app/pages/team-management/team-management.component.spec.ts
```diff
@@ -65,6 +91,163 @@ describe('TeamManagementComponent', () => {
...
+
+    it('on create error, modal is closed with failure banner', () => {
+      spyOn(component.conflictErrorEvent, 'emit');
+      component.openCreateModal('v1');
+      component.createV1TeamForm.controls['name'].setValue(team.id);
+      component.createV1TeamForm.controls['description'].setValue(team.name);
+      component.createV1Team();
+
+      const error = <HttpErrorResponse>{
+        status: 500,
```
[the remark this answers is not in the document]
bcmdarroch: Do you happen to know why we wrote our own enum? We can't import these statuses from some library?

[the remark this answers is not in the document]
msorens: Yes, we could... and should! I was just lazy at the time; did not want to hunt up to see if there was an npm package or other easy way to do it.

PR #1047, opened 2019-07-23
add telemetry for node integration type and scan job payload

vjeffrey:
Just adding a couple telemetry calls to track:
- node integration type
- scan job creation data

### :+1: Definition of Done
telemetry calls for the two above mentioned items

build components/automate-ui
create node integration, see telemetry call in dev tools
create scan job, see telemetry call in dev tools


vjeffrey:
Screen Shot 2019-07-23 at 15 56 15
Screen Shot 2019-07-23 at 15 56 59

PR #942, opened 2019-07-18
[automate-832] Use role db_id in iam_statements (final part)

tylercloke:
### :nut_and_bolt: Description

Updated the `iam_statements` table to have a foreign key to role's via `db_id`. Also, since things are now properly relational, if you delete a role, policies will now remove the role from their statement. If removing the role would result in the statement having no role or actions, then the statement is removed. If that results in an empty custom policy, the policy is removed.

I also added a `insertTestStatement` helper and updated the postgres tests to use it.

**Also found and fixed a few bugs:**
* Resources and actions in the iam_statements didn't have `NOT NULL` constraints and could have resulted in null values. I've populated any NULLs with `'{}'` and added `NOT NULL` constraints.
* Fixed a bug where errors on db transactions would cause an infinite loop and crash authz-service.

##### Setup

```
start_all_services
rebuild components/authz-service
export TOK=`chef-automate iam token create my-token --admin`
```

##### Create a role and some policies that use it

```
curl -kH "api-token: $TOK" -XPOST -d '{"id":"my-custom-role", "name": "pika p", "actions":["iam:projects:list"]}' https://localhost/apis/iam/v2beta/roles
```

`policy1.json` contains a single statement with actions and the role:

```
{
  "name": "Test Policy 1",
  "id": "test-policy-1",
  "members": [ "user:local:*", "token:*" ],
  "statements": [
    {
      "effect": "ALLOW",
      "actions": [ "iam:*" ],
      "role": "my-custom-role"
    }
  ]
}
```

`policy2.json` contains a single statement with no actions and the role:

```
{
  "name": "Test Policy 2",
  "id": "test-policy-2",
  "members": [ "user:local:*", "token:*" ],
  "statements": [
    {
      "effect": "ALLOW",
      "role": "my-custom-role"
    }
  ]
}
```

`policy3.json` contains a multiple statement but one has no actions and the role:

```
{
  "name": "Test Policy 3",
  "id": "test-policy-3",
  "members": [ "user:local:*", "token:*" ],
  "statements": [
    {
      "effect": "ALLOW",
      "role": "my-custom-role"
    },
      {
      "effect": "ALLOW",
      "actions": [ "iam:*" ]
    }
  ]
}
```

Create them all:

```
curl -kH "api-token: $TOK" -XPOST -d @policy1.json https://localhost/apis/iam/v2beta/policies
curl -kH "api-token: $TOK" -XPOST -d @policy2.json https://localhost/apis/iam/v2beta/policies
[...]

PR #1052, opened 2019-07-23
Add NOT NULL constraints in all the places we are missing it and delete / update any invalid data

tylercloke:

Follow up to https://github.com/chef/automate/pull/942. Let's update all the places we are missing the `NOT NULL` constraint and delete any data that is invalid that might have snuck in.

Also, add trigger to clean up any statements that are projectless and consequently any custom policies that are statementless.

Lastly, have both triggers (project and role deletion) properly update notify_policy_change since policies changed as a result of the triggers.

- [ ] Code actually executed?

### DIFF
--- a/components/authz-service/storage/postgres/migration/sql/62_add_not_null_constraints.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/62_add_not_null_constraints.up.sql
@@ -11,10 +11,9 @@ ALTER TABLE iam_project_rules ALTER COLUMN project_id SET NOT NULL;
 
 -- if any iam_projects are somehow missing themselves in their projects array,
 -- update it to '{THEIR_ID}'.
-UPDATE iam_projects p1
-  SET projects = ARRAY[p2.id]
-  FROM iam_projects p2
-  WHERE p1.id = p2.id AND (p1.projects IS NULL OR p1.projects = '{}');
+UPDATE iam_projects
+  SET projects=ARRAY[id]
+  WHERE projects IS NULL OR projects = '{}';
 ALTER TABLE iam_projects ALTER COLUMN projects SET NOT NULL;
 
 DELETE FROM iam_role_projects WHERE project_id IS NULL;
@@ -87,7 +86,7 @@ EXECUTE PROCEDURE purge_statements_with_no_projects();
 -- what becomes tricky is adding the additional logic to completely remove the statement if the affected
 -- policy would also have no actions. so instead of splitting that logic up, let's handle it all in a trigger.
 
-CREATE OR REPLACE FUNCTION purge_statements_with_no_actions_or_role() RETURNS TRIGGER AS $$
+CREATE OR REPLACE FUNCTION purge_statements_missing_both_actions_and_role() RETURNS TRIGGER AS $$
   BEGIN
     -- for statements that will still have actions, simply remove the role since those
     -- statements are still valid
--- a/components/authz-service/storage/postgres/migration/sql/62_add_not_null_constraints.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/62_add_not_null_constraints.up.sql
@@ -1,0 +1,66 @@
+DELETE FROM iam_policy_members WHERE member_id IS NULL;
+ALTER TABLE iam_policy_members ALTER COLUMN member_id SET NOT NULL;
+DELETE FROM iam_policy_members WHERE policy_id IS NULL;
+ALTER TABLE iam_policy_members ALTER COLUMN policy_id SET NOT NULL;
+
+DELETE FROM iam_policy_projects WHERE policy_id IS NULL;
+ALTER TABLE iam_policy_projects ALTER COLUMN policy_id SET NOT NULL;
+
+DELETE FROM iam_project_rules WHERE project_id IS NULL;
+ALTER TABLE iam_project_rules ALTER COLUMN project_id SET NOT NULL;
+
+-- if any iam_projects are somehow missing themselves in their projects array,
+-- update it to '{THEIR_ID}'.
+UPDATE iam_projects p1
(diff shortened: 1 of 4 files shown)


components/authz-service/storage/postgres/migration/sql/62_add_not_null_constraints.up.sql
```diff
@@ -0,0 +1,111 @@
...
+$$
+LANGUAGE plpgsql;
+
+-- run what is in our cleanup function once
+DELETE FROM iam_statements s
+  WHERE NOT EXISTS (SELECT 1 FROM iam_statement_projects p WHERE s.db_id = p.statement_id AND p.project_id IS NOT NULL);
+
+DELETE FROM iam_policies p
+    WHERE NOT EXISTS (SELECT 1 FROM iam_statements s WHERE p.db_id = s.policy_id)
+    AND p.type != 'chef-managed';
```
blakestier: This seems fine since we don't, in principle, want to delete chef-managed policies, but it does make me wonder what the point of them is if they're missing all the things.

bcmdarroch: In theory they should be protected from edits so they should always have all the things 🤞

components/authz-service/storage/postgres/migration/sql/62_add_not_null_constraints.up.sql
```diff
@@ -0,0 +1,111 @@
...
+EXECUTE PROCEDURE purge_statements_with_no_projects();
+
+-- ADD notify_policy_change TO OUR ROLE'S TRIGGER! Everything else below is the same as 61 schema migration:
+
+-- i wanted to use the above FKEY constraint with ON DELETE to clean up any statement where the
+-- role getting removed would result in a NULL role_id in iam_statements. that is doable, but
+-- what becomes tricky is adding the additional logic to completely remove the statement if the affected
+-- policy would also have no actions. so instead of splitting that logic up, let's handle it all in a trigger.
+
+CREATE OR REPLACE FUNCTION purge_statements_with_no_actions_or_role() RETURNS TRIGGER AS $$
```
blakestier: this function makes me think we'd purge a statement missing _either_ actions or role . . . maybe `purge_statements_missing_both_actions_and_role` but also, we're not strictly purging here, we're doing other work (deleting the role from statements). 

Could we have the role deletion take place outside of the trigger, then call the trigger _ON DELETE_ of roles?

components/authz-service/storage/postgres/migration/sql/62_add_not_null_constraints.up.sql
blakestier: Yeah. Just that if we're concerned that someone might have edited them, then we might as well either blow 'em away or fix 'em. But better safe than sorry 🤷‍♂

components/authz-service/storage/postgres/migration/sql/62_add_not_null_constraints.up.sql
```diff
@@ -0,0 +1,111 @@
+DELETE FROM iam_policy_members WHERE member_id IS NULL;
+ALTER TABLE iam_policy_members ALTER COLUMN member_id SET NOT NULL;
+DELETE FROM iam_policy_members WHERE policy_id IS NULL;
+ALTER TABLE iam_policy_members ALTER COLUMN policy_id SET NOT NULL;
```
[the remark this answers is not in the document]
tylercloke: Nope, all of these columns were missing the NOT NULL constraint.

components/authz-service/storage/postgres/migration/sql/62_add_not_null_constraints.up.sql
```diff
@@ -0,0 +1,111 @@
...
+
+DELETE FROM iam_staged_rule_conditions WHERE rule_db_id IS NULL;
+ALTER TABLE iam_staged_rule_conditions ALTER COLUMN rule_db_id SET NOT NULL;
+
+DELETE FROM iam_statement_projects WHERE statement_id IS NULL;
+ALTER TABLE iam_statement_projects ALTER COLUMN statement_id SET NOT NULL;
+DELETE FROM iam_statement_projects WHERE project_id IS NULL;
+ALTER TABLE iam_statement_projects ALTER COLUMN project_id SET NOT NULL;
+
+-- move code to update policy_change_tracker and notify into a PG function so we can
```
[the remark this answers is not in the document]
tylercloke: I moved the below from statements in a go function into a SQL function so we can call it directly from both the go code and the SQL code.

components/authz-service/storage/postgres/migration/sql/62_add_not_null_constraints.up.sql
tylercloke: Ah yeah that name isn't great. Good catch! It could take place outside the trigger. Don't see a clear advantage either way. Will update function name for starters.

components/authz-service/storage/postgres/migration/sql/62_add_not_null_constraints.up.sql
```diff
@@ -0,0 +1,111 @@
...
+
+DELETE FROM iam_project_rules WHERE project_id IS NULL;
+ALTER TABLE iam_project_rules ALTER COLUMN project_id SET NOT NULL;
+
+-- if any iam_projects are somehow missing themselves in their projects array,
+-- update it to '{THEIR_ID}'.
+UPDATE iam_projects p1
+  SET projects = ARRAY[p2.id]
+  FROM iam_projects p2
+  WHERE p1.id = p2.id AND (p1.projects IS NULL OR p1.projects = '{}');
```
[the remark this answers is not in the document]
tylercloke: That's a little simpler xD

PR #1055, opened 2019-07-24
authz-service/sql: make helper methods strict

srenatus:

The helper methods are strict now. There's no customer impact besides a bit of warm, fuzzy feeling for our developers knowing their database contents are consistent.

Also refactored some of the error handling, using the foreign key exception to signal missing roles or projects in policy creations etc.

See tests pass.

- #832

### DIFF
--- a/components/authz-service/storage/postgres/datamigration/sql/01_v2_data_migrations.up.sql
+++ b/components/authz-service/storage/postgres/datamigration/sql/01_v2_data_migrations.up.sql
@@ -9,12 +9,6 @@ UPDATE iam_members
         name = 'team:local:operators';
 
 UPDATE iam_statements
-    SET
-        role_id = role_db_id('editor')
-    WHERE
-        role_id = role_db_id('operator');
-
-UPDATE iam_statements
     SET resources = array_replace(resources, 'team:local:operator', 'team:local:editor');
 
 UPDATE iam_roles
--- a/components/authz-service/storage/postgres/migration/sql/63_make_db_id_lookup_functions_strict.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/63_make_db_id_lookup_functions_strict.up.sql
@@ -32,5 +32,23 @@ BEGIN
 END;
 $$
 LANGUAGE plpgsql;
+CREATE OR REPLACE FUNCTION role_db_id (
+    _id iam_roles.id % TYPE,
+    OUT _db_id iam_roles.db_id % TYPE
+)
+    RETURNS iam_roles.db_id % TYPE
+    AS $$
+BEGIN
+    SELECT
+        db_id INTO _db_id
+    FROM
+        iam_roles
+    WHERE
+        id = _id;
+    IF NOT FOUND THEN
+        RAISE foreign_key_violation USING MESSAGE='role not found: ' || _id;
+    END IF;
+END;
+$$
+LANGUAGE plpgsql;
 COMMIT;
-
--- a/components/authz-service/storage/postgres/migration/sql/63_make_db_id_lookup_functions_strict.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/63_make_db_id_lookup_functions_strict.up.sql
@@ -1,5 +1,4 @@
 BEGIN;
-
 -- lookup by name, the UUID column can go away
 CREATE OR REPLACE FUNCTION member_db_id (
     _name iam_members.name % TYPE,
@@ -17,6 +16,21 @@ BEGIN
 END;
 $$
 LANGUAGE plpgsql;
-
+CREATE OR REPLACE FUNCTION policy_db_id (
+    _id iam_policies.id % TYPE,
+    OUT _db_id iam_policies.db_id % TYPE
+)
+    RETURNS iam_policies.db_id % TYPE
+    AS $$
+BEGIN
+    SELECT
+        db_id INTO STRICT _db_id
+    FROM
+        iam_policies
+    WHERE
+        id = _id;
+END;
+$$
+LANGUAGE plpgsql;
 COMMIT;
 
--- a/components/authz-service/storage/postgres/migration/sql/63_make_db_id_lookup_functions_strict.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/63_make_db_id_lookup_functions_strict.up.sql
@@ -83,4 +83,27 @@ END;
 $$
 LANGUAGE plpgsql;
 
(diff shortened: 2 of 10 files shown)


components/authz-service/storage/postgres/datamigration/sql/01_v2_data_migrations.up.sql
```diff
@@ -8,12 +8,6 @@ UPDATE iam_members
     WHERE
         name = 'team:local:operators';
 
-UPDATE iam_statements
-    SET
-        role_id = role_db_id('editor')
-    WHERE
-        role_id = role_db_id('operator');
-
```
srenatus: These weren't doing anything -- we've introduced a `db_id` so we could rename them without having to update references. Before the strict `role_db_id` function, the WHERE would have selected `role_id=NULL`, matching nothing.

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -612,23 +586,24 @@ func (p *pg) insertOrReusePolicyMemberWithQuerier(ctx context.Context, policyID
...
-	return err
+			VALUES (policy_db_id($1), member_db_id($2)) ON CONFLICT DO NOTHING`, policyID, member.Name)
+	return errors.Wrapf(err, "failed to upsert member link: member=%s, policy_id=%s", member.Name, policyID)
 }
 
 func (p *pg) getPolicyMembersWithQuerier(ctx context.Context, id string, q Querier) ([]v2.Member, error) {
 	rows, err := q.QueryContext(ctx,
 		`SELECT m.id, m.name FROM iam_policy_members AS pm
-			INNER JOIN iam_members AS m ON pm.member_id=m.db_id
+			JOIN iam_members AS m ON pm.member_id=m.db_id
```
srenatus: Every JOIN that is not LEFT/RIGHT is INNER. Just stumbled upon this when updating the piece above.

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -1660,8 +1654,44 @@ func TestCreatePolicy(t *testing.T) {
...
+				Projects: []string{projID0, projID1},
+			}
+
+			assertNoPolicyChange(t, store, func() {
+				resp, err := store.CreatePolicy(ctx, &pol)
+				require.Error(t, err)
+				assert.Nil(t, resp)
+				_, ok := err.(*storage_errors.ErrForeignKey)
+				require.True(t, ok, "expected foreign key error")
+				assert.Equal(t, "project not found: not-real-project", err.Error())
```
srenatus: This is new -- before, we would have only said "your projects contain one that doesn't exist". Now, we're explicit.

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -6836,9 +6881,9 @@ func genRole(t *testing.T, id string, name string, actions []string, projects []
 }
 
 func insertTestPolicy(t *testing.T, db *testhelpers.TestDB, policyName string) string {
-	row := db.QueryRow(fmt.Sprintf("INSERT INTO iam_policies "+
```
srenatus: What a shocking find 😉

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -612,23 +612,24 @@ func (p *pg) insertOrReusePolicyMemberWithQuerier(ctx context.Context, policyID
 	// updating either iam_members id or name columns which is the entire table. Also, we are currently
 	// not deleting any of the rows, but reusing them per name string.
 
-	_, err := q.ExecContext(ctx, `INSERT INTO iam_members (id, name)  VALUES ($1, $2) ON CONFLICT DO NOTHING;`,
+	_, err := q.ExecContext(ctx,
+		 `INSERT INTO iam_members (id, name) VALUES ($1, $2) ON CONFLICT DO NOTHING;`,
```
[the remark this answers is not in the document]
bcmdarroch: actually looks like you got it on your other PR 😌

components/authz-service/storage/postgres/migration/sql/63_make_db_id_lookup_functions_strict.up.sql
```diff
@@ -0,0 +1,22 @@
+BEGIN;
+
+-- lookup by name, the UUID column can go away
+CREATE OR REPLACE FUNCTION member_db_id (
+    _name iam_members.name % TYPE,
+    OUT _db_id iam_members.db_id % TYPE
```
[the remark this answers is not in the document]
srenatus: Lazy-people feature: it'll resolve to the mentioned column's _type_ at time of the function creation. If that column type changes, it will **not** adapt automatically.

components/authz-service/storage/postgres/migration/sql/63_make_db_id_lookup_functions_strict.up.sql
```diff
@@ -0,0 +1,109 @@
...
+
+CREATE OR REPLACE FUNCTION role_db_id (
+    _id iam_roles.id % TYPE,
+    OUT _db_id iam_roles.db_id % TYPE
+)
+    RETURNS iam_roles.db_id % TYPE
+    AS $$
+BEGIN
+    SELECT
+        db_id INTO _db_id
```
msorens: Is the reason not to use strict here so that you can customize the error message?

srenatus: Not quite -- you can customize the error message with `INTO STRICT` as well:

```
BEGIN
    SELECT * INTO STRICT myrec FROM emp WHERE empname = myname;
    EXCEPTION
        WHEN NO_DATA_FOUND THEN
            RAISE EXCEPTION 'employee % not found', myname;
        WHEN TOO_MANY_ROWS THEN
            RAISE EXCEPTION 'employee % not unique', myname;
END;
```
[See here](https://www.postgresql.org/docs/9.6/plpgsql-statements.html)

Both would have the same effect (except that we don't need to worry about the _too many rows_ case, since we've got indices on those fields).

**However**, [the docs](https://www.postgresql.org/docs/9.6/plpgsql-control-structures.html) also say that having `EXCEPTION` blocks makes this more expensive in processing:

> Tip: A block containing an EXCEPTION clause is significantly more expensive to enter and exit than a block without one. Therefore, don't use EXCEPTION without need.

Asking in the PG slack's #beginners channel only yielded that `IF NOT FOUND THEN` was more common.

PR #1060, opened 2019-07-24
Use full path to chef-server-ctl during migration

jaym:
A1 migration will fail for people with /bin linked to /usr/bin
because we binlink chef-server-ctl into /bin before the migration is
complete. This prevents us from calling the correct chef-server-ctl.
I hope this commit fixes that


jaym:
i feel like this is incomplete. workflow should also be problematic

stevendanna:
@jaym We could do the same thing for `automate-ctl` but I think we dodged a bullet on that one as we called our new binary `workflow-ctl` to avoid confusion.

jaym:
ah. that makes me feel better.

Issue #1000, opened 2019-07-22
Service Groups with different application names are grouped together


gscho:
## Describe the bug
When loading a new habitat service with event streaming enabled, the automate beta UI will group that service into any existing service group that matches the name, ignoring the `----event-stream-application` setting.

## To Reproduce
- Start one supervisor with event streams enabled and `--event-stream-application=app-1`
- Load up a redis server: `hab svc load core/redis`
- Start another supervisor on a different VM with event streams enabled and `--event-stream-application=app-2`
- Load up the same service as the first supervisor: `hab svc load core/redis`
- The automate UI will show 2 instances of redis.default grouped under `app-1`

## Expected behavior
I would expect there to be 2 groupings of redis.default in the UI. One showing `app-1` as the App name and the other showing `app-2`.

## Screenshots
TBD

## Versions (please complete the following information):
- OS: TBD
- Supervisor version 0.83.0
- Browser: firefox
- Automate Build Number: TBD

## Additional context
N/A

danielsdeleo:
Working on this one, see the referenced PR for more

danielsdeleo:
I didn't get the buildstamp of the first build with the patch applied, but the patch is present in the current dev build with buildstamp `20190731171254`. Any version after this will contain the fix. If we are able to keep to our normal promotion schedule, then the fix should be released to the acceptance channel on August 5 or 6.

gscho:
Thanks @danielsdeleo !

PR #1064, opened 2019-07-24
Dan/fix incorrect service grouping

danielsdeleo:

A customer reported that services are not assigned to the correct groups when they have different application settings. I was able to reproduce pretty easily by running:

```
applications_publish_supervisor_message --application ONE --sup-id 1111
applications_publish_supervisor_message --application TWO --sup-id 2222
applications_publish_supervisor_message --application THREE --sup-id 3333
```

On insert of a new service, the problem was that we were only looking at service groups based on their name, whereas we needed to be checking both the name and deployment id to get the correct row back.

After applying a fix for that problem, I investigated how the system would behave if the user has existing data that is incorrect due to this insertion bug. We recently added some code to modify the application and environment names if they change, but I've discovered that it was subtly incorrect: the existing code updates the deployment and/or service group values to match what was given in the latest health check message, but this changes the app name/environment/service group name for all services associated with the deployment/service group. I've therefore changed the logic so that when a health check message indicates that a service should belong to a different service group/deployment than what is currently in the database, we create a new deployment/service group, if necessary, then move the service over to the correct deployment and service group. I have confirmed that if there is incorrect data in the database resulting from the #1000 bug, it will be repaired as new health checks come in and the update logic gets applied.

To repro bug #1000, run this in the studio with the packages from the dev channel:

```
applications_publish_supervisor_message --application ONE --sup-id 1111
applications_publish_supervisor_message --application TWO --sup-id 2222
applications_publish_supervisor_message --application THREE --sup-id 3333
```

In the UI you will see all 3 services in one row in the UI which indicates they are all in the same application.

To repro the fix, build applications service and wait for it to get deployed. Then re-run the `applications_publish_supervisor_message` step from above and you should see 3 service groups in the UI.

#1000 

- [x] Code actually executed?

### DIFF
--- a/components/applications-service/integration_test/ingestion_test.go
+++ b/components/applications-service/integration_test/ingestion_test.go
@@ -7,7 +7,6 @@ package integration_test
 
 import (
 	"context"
-	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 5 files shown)


components/applications-service/integration_test/ingestion_test.go
```diff
@@ -672,3 +673,60 @@ func TestIngestConcurrencySafe(t *testing.T) {
...
+	require.NoError(t, err)
+	suite.Ingester.IngestMessage(bytes3)
+	eventsProcessed++
+	suite.WaitForEventsToProcess(eventsProcessed)
+
+	// TODO: replace printf with test to assert we have 2 services w/ 2 supervisors in 2 service groups in 2 deployments
+	svcList, err = suite.StorageClient.GetServices("name", true, 1, 5, nil)
+	require.NoError(t, err)
+	for _, svc := range svcList {
+		fmt.Printf("%+v\n", svc)
```
danielsdeleo: Prior to fixing the update logic, this prints (look at app and env):

```
&{ID:48 SupMemberID:1 Origin:core Name:db Version:0.1.0 Release:20200101121212 Status: Health:OK Group:db.default Fqdn: Application:newApp Environment:newEnv Channel: Site:us PreviousHealth:NONE UpdateStrategy:NONE LastEventOccurredAt:2019-07-24 20:36:38.118062 +0000 +0000 HealthUpdatedAt:2019-07-24 20:36:38.147417 +0000 +0000}
&{ID:47 SupMemberID:2 Origin:core Name:db Version:0.1.0 Release:20200101121212 Status: Health:OK Group:db.default Fqdn: Application:newApp Environment:newEnv Channel: Site:us PreviousHealth:NONE UpdateStrategy:NONE LastEventOccurredAt:2019-07-24 20:36:38.16734 +0000 +0000 HealthUpdatedAt:2019-07-24 20:36:38.14151 +0000 +0000}
```

After fixing, it prints:

```
&{ID:51 SupMemberID:1 Origin:core Name:db Version:0.1.0 Release:20200101121212 Status: Health:OK Group:db.default Fqdn: Application:app Environment:test-env Channel: Site:us PreviousHealth:NONE UpdateStrategy:NONE LastEventOccurredAt:2019-07-24 21:31:46.952817 +0000 +0000 HealthUpdatedAt:2019-07-24 21:31:46.986629 +0000 +0000}
&{ID:50 SupMemberID:2 Origin:core Name:db Version:0.1.0 Release:20200101121212 Status: Health:OK Group:db.default Fqdn: Application:newApp Environment:newEnv Channel: Site:us PreviousHealth:NONE UpdateStrategy:NONE LastEventOccurredAt:2019-07-24 21:31:46.997118 +0000 +0000 HealthUpdatedAt:2019-07-24 21:31:46.977513 +0000 +0000}
```

components/applications-service/pkg/storage/postgres/hab_event.go
```diff
@@ -172,43 +172,135 @@ func (db *Postgres) IngestHealthCheckEventWithoutMetrics(event *habitat.HealthCh
...
+	// the one we found. This will correctly handle the case where a service is
+	// in the same service group and deployment as it is now, or the case that
+	// the service was "moved" to a different deployment and service group that
+	// already exist.
+	//
+	// If there is no existing deploymentServiceGroup that matches the metadata,
+	// then the service has been moved to a new deployment/service group. In that
+	// case, we need to create them.
+	// After that, we need to check if the old deployment and service group have
+	// no more related services, and delete them if so.
```
danielsdeleo: The changes here are worth a careful read 🙏

components/applications-service/pkg/storage/postgres/hab_event.go
```diff
@@ -172,43 +172,135 @@ func (db *Postgres) IngestHealthCheckEventWithoutMetrics(event *habitat.HealthCh
...
 
-	deploy, err := db.getDeployment(svc.DeploymentID)
-	if err != nil {
-		return errors.Wrap(err, "unable to update tables")
+	// When the service belongs to a deployment+service_group that exist (either
+	// the same one as before, or a different one):
+	if err == nil {
+		// The service could be in the same service_group and deployment as it was
+		// before, but it could also be in a new one.
+		// TODO/FIXME: updateService needs to adjust the deployment_id and group_id
```
danielsdeleo: TODONE

components/applications-service/pkg/storage/postgres/hab_event.go
```diff
@@ -395,7 +493,15 @@ func (db *Postgres) insertNewService(
 		}
 
 		// 2) Service Group
-		gid, exist := db.getServiceGroupID(svcMetadata.GetServiceGroup())
+		// FIXME that's your problem right there :P
+		// we need to find service group by name and deployment id (done)
+		// TODO after this:
+		// * regression test
+		// * see how it behaves when running with existing bad data
+		// * does updateService work? (seems to, but be more systematic)
```
danielsdeleo: the answer is no, see above.

components/applications-service/pkg/storage/postgres/hab_event.go
```diff
@@ -395,7 +493,15 @@ func (db *Postgres) insertNewService(
 		}
 
 		// 2) Service Group
-		gid, exist := db.getServiceGroupID(svcMetadata.GetServiceGroup())
+		// FIXME that's your problem right there :P
+		// we need to find service group by name and deployment id (done)
+		// TODO after this:
+		// * regression test
+		// * see how it behaves when running with existing bad data
```
danielsdeleo: need to confirm, but I think fixing the update service logic will fix it

components/applications-service/pkg/storage/postgres/hab_event.go
kmacgugan: If it is TODONE can we remove the comment?

components/applications-service/integration_test/service_groups_sorting_test.go
```diff
@@ -218,11 +218,13 @@ func TestGetServiceGroupsSortedPercent(t *testing.T) {
 	response, err := suite.ApplicationsServer.GetServiceGroups(ctx, request)
 	assert.Nil(t, err)
 
-	if assert.Equal(t, 4, len(response.ServiceGroups)) {
+	if assert.Equal(t, 6, len(response.ServiceGroups)) {
```
danielsdeleo: Sorting behavior shouldn't be changed but these test were relying on the bug behavior so they had to be changed.

components/applications-service/pkg/storage/postgres/db.go
```diff
@@ -111,13 +107,6 @@ type service struct {
 	HealthUpdatedAt     time.Time `db:"health_updated_at"`
 	CreatedAt           time.Time `db:"-"`
 	UpdatedAt           time.Time `db:"-"`
-
-	// (internal) use it to know if the service needs an update or not
-	needUpdate bool `db:"-"`
-}
-
-func (svc *service) NeedUpdate() bool {
-	return svc.needUpdate
```
danielsdeleo: We update the service every time so we can find when it disconnects, so this was just always returning true in practice.

components/applications-service/pkg/storage/postgres/db.go
```diff
@@ -144,13 +133,6 @@ type serviceGroup struct {
 	DeploymentID int32     `db:"deployment_id"`
 	CreatedAt    time.Time `db:"-"`
 	UpdatedAt    time.Time `db:"-"`
-
-	// (internal) use it to know if the service-group needs an update or not
-	needUpdate bool `db:"-"`
-}
-
-func (sg *serviceGroup) NeedUpdate() bool {
-	return sg.needUpdate
```
danielsdeleo: Service group and deployment shouldn't be changed ever, we should just move services between them.

components/applications-service/pkg/storage/postgres/hab_event.go
```diff
@@ -172,62 +171,134 @@ func (db *Postgres) IngestHealthCheckEventWithoutMetrics(event *habitat.HealthCh
...
+ WHERE sg.name = $1 AND d.app_name = $2 AND d.environment = $3;
+`
+
+	maybeCleanupServiceGroup = `
+DELETE FROM service_group WHERE service_group.id = $1 AND (NOT EXISTS (SELECT 1 FROM service WHERE service.group_id = service_group.id ))
+`
+	maybeCleanupDeployment = `
+DELETE FROM deployment WHERE deployment.id = $1 AND (NOT EXISTS (SELECT 1 FROM service WHERE service.deployment_id = deployment.id ))
+`
+)
```
danielsdeleo: oops, need to move these things to not-here.

components/applications-service/pkg/storage/postgres/hab_event.go
```diff
@@ -172,62 +171,134 @@ func (db *Postgres) IngestHealthCheckEventWithoutMetrics(event *habitat.HealthCh
...
+//
+// If we did find a matching deployment and service group, and it's not what is
+// currently set on the service, we update the foreign key relations to the
+// other deployment and service group when we apply updates for the service's
+// other fields.
+//
+// If we changed the service's deployment and service group, then we run a
+// cleanup query to ensure the prior deployment and service group are deleted
+// if they no longer have associated services.
+func (db *Postgres) updateServiceAndRelations(
```
danielsdeleo: open to suggestions on how to break this function down into more readable parts

Issue #71, opened 2019-04-03
use vpn-protected testing instance to replace current testing instances


vjeffrey:
# User Story

i.e. the machines referenced here: https://github.com/chef/a2/blob/master/components/automate-gateway/integration/license_usage_nodes_test.go#L81 and here: https://github.com/chef/a2/blob/master/components/automate-gateway/integration/nodes_test.go#L19

those references are sprinkled through the code in a few other places

# Definition of Done
use vpn protected instances for testing

some WIP towards making this happen https://github.com/chef/a2/compare/vj/cleanup-cmp

PR #1066, opened 2019-07-25
modify tests to use inspec-scanner user and key in target testing

vjeffrey:
a few days ago i made some changes to use the vpn protected testing targets. this was using password access to ssh to the machines.
we are now transitioning to key usage, using the inspec-scanner user that is set up on those instances (https://github.com/chef/automate/blob/master/terraform/test-environments/scenarios/inspec_target_rhel7.tf#L69)

### :+1: Definition of Done
all tests run against target machines use key access

```
CHEF_USERNAME=username scripts/get_secrets.sh
source dev/secrets-env.sh
run nodes test: go_test components/automate-gateway/integration/nodes_test.g
run e2e scanner test
```

https://github.com/chef/automate/issues/71


vjeffrey:
grrr: 
/hab/pkgs/core/coreutils/8.30/20190115012313/bin/env: ‘RSA’: No such file or directory
 
i can repro the error locally...just not clear yet on how to fix it..

vjeffrey:
running in debug mode is see this: `"--env", "HAB_STUDIO_SECRET_AUTOMATE_ACCEPTANCE_TARGET_KEY=-----BEGIN RSA PRIVATE KEY--...`, which, combined with the error above, seems to support my theory that whitespace in an env var value is making this choke. 
:/

vjeffrey:
proposed workaround from @stevendanna : try base64 encoding the entire key, then decoding in tests

scripts/get_secrets.sh
```diff
@@ -59,20 +59,20 @@ vault kv get -field=license secret/a2/delivery_license | base64 --decode >compon
 
 # Automate acceptence env secrets
 target_host=$(vault kv get -field=data secret/a2/testing/target_host)
-target_username=$(vault kv get -field=data secret/a2/testing/target_username)
-target_password=$(vault kv get -field=data secret/a2/testing/target_password)
+target_user=$(vault kv get -field=data secret/a2/testing/target_user)
+target_key=$(vault kv get -field=data secret/a2/testing/target_key | base64)
```
stevendanna: I think you are going to need to upload a base64 encoded version because in CI getting these secrets is handled by buildkite

vjeffrey: ohhhh, good to know! 👍

vjeffrey:
https://github.com/habitat-sh/habitat/issues/6760

PR #1067, opened 2019-07-25
ensure ui for compliance suggestions is displaying all 100 results

vjeffrey:
in https://github.com/chef/automate/pull/1065 we did some work to bump the results returned from the compliance api from 10 -> 100, which worked beautifully.
but when testing the change later, i noticed i was only getting 25 results in the ui, despite the api returning 100.
this fixes that!

### :+1: Definition of Done
100 (or as many as available if less than 100) suggestion results returned

```
build components/automate-ui
load 100 compliance nodes (or more than 25)
suggest on node name, expect to see more than 25 nodes
```


vjeffrey:
Screen Shot 2019-07-24 at 23 46 47
Screen Shot 2019-07-24 at 23 47 11

Issue #943, opened 2019-07-18
Node Manager service crashes when calling /nodes/id/{id}


trickyearlobe:
## Describe the bug

When querying node data via API `GET /nodes/id/{id}` as documented in Swagger at `https://$AUTOMATE_FQDN/api/v0/openapi/ui/#/NodesService/Read`

* using a valid node id (guid), the nodemanager-service crashes and restarts.
* using an invalid id, I get the expected 404 Not Found

## To Reproduce

Get a list of valid node ID's

```
curl -s -H "api-token: $AUTOMATE_TOKEN" -H "Content-Type: application/json" -XPOST https://$AUTOMATE_FQDN/api/v0/nodes/search |jq
```

Assuming a node ID of eea7d742-9db9-49bb-800b-49f9d6aad47

```
curl -s -H "api-token: $AUTOMATE_TOKEN" -H "Content-Type: application/json" -XGET https://$AUTOMATE_FQDN/api/v0/nodes/id/eea7d742-9db9-49bb-800b-49f9d6aad47
```

## Expected behavior

Expected to get a JSON document with some data about the node

## Stack Trace

```
Jul 18 07:54:23 ip-172-31-5-8.eu-west-2.compute.internal hab[9230]: authz-service.default(O): time="2019-07-18T07:54:23Z" level=info msg="Authorization Query" action=read resource="nodes:eea7d742-9db9-49bb-800b-49f9d6aad47a" result=true subject="[token:531a569b-3b1b-4ccc-b12f-56e08918ca92]"
Jul 18 07:54:23 ip-172-31-5-8.eu-west-2.compute.internal hab[9230]: nodemanager-service.default(O): time="2019-07-18T07:54:23Z" level=info msg="read node with : id:\"eea7d742-9db9-49bb-800b-49f9d6aad47a\" "
Jul 18 07:54:23 ip-172-31-5-8.eu-west-2.compute.internal hab[9230]: nodemanager-service.default(O): panic: runtime error: invalid memory address or nil pointer dereference
Jul 18 07:54:23 ip-172-31-5-8.eu-west-2.compute.internal hab[9230]: nodemanager-service.default(O): [signal SIGSEGV: segmentation violation code=0x1 addr=0x108 pc=0xa35592]
Jul 18 07:54:23 ip-172-31-5-8.eu-west-2.compute.internal hab[9230]: nodemanager-service.default(O):
[...]

PR #1069, opened 2019-07-25
fix node read for nodes with nil target config

vjeffrey:
Part of the node read code in nodemanager service parses the node target config. Ingested nodes (nodes that were not manually added via node integration or node add) do not have a target config, and an oversight in the code resulted in the service exploding when trying to access a field in the nonexistent target config.

### :+1: Definition of Done
Able to read nodes that were only ingested.

build components/nodemanager-service
ingest some nodes
read ingested nodes

https://github.com/chef/automate/issues/943

### DIFF
--- a/components/compliance-service/api/tests/test_ingest_to_mgr_conn.rb
+++ b/components/compliance-service/api/tests/test_ingest_to_mgr_conn.rb
@@ -26,6 +26,11 @@ describe File.basename(__FILE__) do
     # those nodes should not have been added to the manual node manager, as they were ingested nodes, not manually added nodes
     manually_managed_nodes = MANAGER_GRPC nodes, :list, Nodes::Query.new(filters: [Common::Filter.new(key: "manager_id", values: ["e69dc612-7e67-43f2-9b19-256afd385820"])])
     assert_equal(original_manually_managed_nodes.total, manually_managed_nodes.total)
+
+    # ensure we can read an ingested node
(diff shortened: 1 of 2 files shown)


components/compliance-service/api/tests/test_ingest_to_mgr_conn.rb
```diff
@@ -26,6 +26,11 @@
     # those nodes should not have been added to the manual node manager, as they were ingested nodes, not manually added nodes
     manually_managed_nodes = MANAGER_GRPC nodes, :list, Nodes::Query.new(filters: [Common::Filter.new(key: "manager_id", values: ["e69dc612-7e67-43f2-9b19-256afd385820"])])
     assert_equal(original_manually_managed_nodes.total, manually_managed_nodes.total)
+
+    # ensure we can read an ingested node
+    nodes_list['nodes'].each { |node|
+      node_read = MANAGER_GRPC nodes, :read, Nodes::Id.new(id: node.id)
+    }
```
phiggins: It seems kind of strange to me that none of our other tests caught this. Do we not test ingested nodes anywhere else?

vjeffrey: ya i was surprised by that as well - 
we test compliance/reporting/nodes at the list and read level
we test nodes api at the list and read level for nodes added manually or via integration
but for nodes api + ingested nodes we were only testing list

PR #1070, opened 2019-07-25
ci: use buildkite log markers in verify_build

stevendanna:
- Use buildkite log markers to fold output and provide per-step
  timestamps

- Use HAB_STUDIO_SECRET_HAB_INTERACTIVE=false to turn of progress bars
  inside the build.

### DIFF
--- a/scripts/verify_build.sh
+++ b/scripts/verify_build.sh
@@ -9,24 +9,32 @@ if [[ "$CI" != "true" ]]; then
     echo "WARNING: This script is intended to run in buildkite.  It will likely fail outside of buildkite."
 fi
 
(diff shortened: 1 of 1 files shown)


srenatus:
image

😍

stevendanna:
You beat me to the screen shot :D.  I also figured out why we are still getting progress bars in a lot of cases. I applied the fix to the build here, but we can probably apply it in more places as well?

srenatus:
> I applied the fix to the build here, but we can probably apply it in more places as well?

🤔 make file targets? like, lint vs tests... but I don't think that'll be too informative.

PR #1073, opened 2019-07-25
log on timeout of hab command

stevendanna:
The idea here is that we log diagnostics output when hab commands time
out.  So far, we've only seen this in CI on hab 0.82.0

### DIFF
--- a/components/automate-deployment/pkg/target/hab_cmd.go
+++ b/components/automate-deployment/pkg/target/hab_cmd.go
@@ -2,8 +2,11 @@ package target
 
 import (
 	"context"
+	"os"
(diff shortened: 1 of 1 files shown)


stevendanna:
Immediately upon pushing this, I realized that the issues we've seen on hab 0.82.0 is almost surely the same musl+rust+malloc bug that has plagued us in the past.  We can bring back this trick if we ever want it, but for now, let's close this.

stevendanna:
We no longer think the hang is the musl+malloc bug but may be related to a low-memory condition.  Let's merge this to try to get some more information.

PR #1074, opened 2019-07-25
cleanup: remove member id column

srenatus:

~Branched off of #1055, use [this view to review](https://github.com/chef/automate/compare/385d0b9fb7852bfaa3aab78ef55b76f8c3571483...sr/cleanup/remove-member-id-column), please.~ ✔️ other PR merged, this one rebased

Since members are now referenced by their `db_id`, we no longer need a UUID for a member. Their names are unique as well. This change drops that column, and deals with the fallout.

See tests pass.

- #832 
  - #1055

### DIFF
--- a/components/authz-service/storage/postgres/migration/sql/64_drop_uuid_of_iam_members.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/64_drop_uuid_of_iam_members.up.sql
@@ -1,0 +1,141 @@
+BEGIN;
+
+ALTER TABLE iam_members DROP COLUMN id;
+
+CREATE OR REPLACE FUNCTION query_policy (_policy_id TEXT, _projects_filter TEXT[])
+    RETURNS json
+    AS $$
+    WITH temp AS (
+        SELECT
+            pol.db_id,
+            pol.id,
+            pol.name,
+            pol.type,
+            -- get policy's statements using temporary table
+            ( WITH statement_rows AS (
+                    SELECT
+                        stmt.db_id,
+                        stmt.id,
+                        stmt.effect,
+                        stmt.actions,
+                        stmt.resources,
+                        (
+                          SELECT COALESCE((SELECT id FROM iam_roles WHERE db_id=stmt.role_id), '') AS role
+                        ),
+                        (
+                            SELECT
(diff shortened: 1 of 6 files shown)


srenatus:
> This looks ready to me, although I see it's still a draft.

I'd like to rebase this after #1055 is in. Currently, this branch contains both PRs' commits

components/authz-service/storage/postgres/migration/sql/64_drop_uuid_of_iam_members.up.sql
```diff
@@ -0,0 +1,141 @@
...
+                    stmt.actions,
+                    stmt.resources,
+                    stmt.role_id)
+            SELECT
+                array_agg(statement_rows) FILTER (WHERE statement_rows.id IS NOT NULL)
+            FROM statement_rows) AS statements,
+            -- get policy members
+            (
+            SELECT
+                array_agg(mem)
```
srenatus: ℹ️ In this function and the other one, I have removed a `FILTER (WHERE mem.id IS NOT NULL)` here. 🤷‍♂ No idea why that was even there in the first place.

msorens: Thank you for the delta-comment here!

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -1327,12 +1327,11 @@ func TestCreatePolicy(t *testing.T) {
 				Actions:   actions0,
 			}
 			member0 := genMember(t, "user:local:albertine")
-			member1 := genMember(t, "user:local:albertine")
 
 			pol := storage.Policy{
 				ID:         polID,
 				Name:       name,
-				Members:    []storage.Member{member0, member1},
+				Members:    []storage.Member{member0, member0},
```
msorens: Why duplicate the member?

srenatus: Before, this was member0 and member1, which both sharing a name, but having a different ID. Since there are no IDs anymore, this should be approximating the same test case.

PR #1076, opened 2019-07-25
Add workflow for domain service projects update

jaym:


components/authz-service/server/v2/project_update_manager.go
```diff
@@ -523,3 +524,233 @@ func getEstimatedTimeCompleteInSec(event *automate_event.EventMsg) (time.Time, e
...
+		workflowName:   workflowName,
+		instanceName:   instanceName,
+		domainServices: domainServices,
+	}
+
+	return updateManager, nil
+}
+
+func (m *CerealProjectUpdateManager) Cancel() error {
+	err := m.manager.CancelWorkflow(context.Background(), m.workflowName, m.instanceName)
```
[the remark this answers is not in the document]
jaym: the interface doesn't take a context atm. we could change that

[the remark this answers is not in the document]
srenatus: I don't know what's needed. Just thought it curious to see so many `context.Background()` sprinkled around. Guess if we replace them with `context.TODO()` and change the interface later, that would be fine, too.

components/authz-service/server/v2/project.go
```diff
@@ -45,12 +45,19 @@ func NewMemstoreProjectsServer(
...
-	projectUpdateManager := NewProjectUpdateManager(eventServiceClient, configManager)
+	//projectUpdateManager := NewProjectUpdateManager(eventServiceClient, configManager)
+	projectUpdateManager, err := RegisterCerealProjectUpdateManager(projectUpdateCerealManager,
+		"ProjectUpdate", "Runner", []string{
+			"ingest",
+			"compliance",
+		})
+	if err != nil {
+		return nil, err
+	}
```
lancewf: I am not sure I understand why there are two ways to create the `projectUpdateManager`. It looks like this one (`NewMemstoreProjectsServer`) is not being used.

components/authz-service/server/v2/project_update_manager.go
```diff
@@ -523,3 +524,233 @@ func getEstimatedTimeCompleteInSec(event *automate_event.EventMsg) (time.Time, e
...
+	}
+	return err
+}
+
+func (m *CerealProjectUpdateManager) Start() error {
+	params := map[string]interface{}{}
+
+	for _, svc := range m.domainServices {
+		params[svc] = project_update_tags.DomainProjectUpdateWorkflowParameters{
+			ProjectUpdateID: "project-update-id",
```
lancewf: Do we need the `ProjectUpdateID` anymore?

components/authz-service/server/v2/project_update_manager.go
```diff
@@ -523,3 +524,233 @@ func getEstimatedTimeCompleteInSec(event *automate_event.EventMsg) (time.Time, e
...
+func RegisterCerealProjectUpdateManager(manager *cereal.Manager, workflowName string, instanceName string, domainServices []string) (ProjectUpdateMgr, error) {
+	domainServicesWorkflowExecutor := NewWorkflowExecutor(domainServices)
+
+	if err := manager.RegisterWorkflowExecutor(workflowName, domainServicesWorkflowExecutor); err != nil {
+		return nil, err
+	}
+
+	updateManager := &CerealProjectUpdateManager{
+		manager:        manager,
+		workflowName:   workflowName,
```
lancewf: Could the `workflowName` be hardcoded here?

components/authz-service/server/v2/project.go
```diff
@@ -70,7 +77,15 @@ func NewPostgresProjectsServer(
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to initialize v2 store state")
 	}
-	projectUpdateManager := NewProjectUpdateManager(eventServiceClient, configManager)
+	//projectUpdateManager := NewProjectUpdateManager(eventServiceClient, configManager)
+	projectUpdateManager, err := RegisterCerealProjectUpdateManager(projectUpdateCerealManager,
+		"ProjectUpdate", "Runner", []string{
```
lancewf: "Runner" is the `instanceName` variable for the workflow, should that be a unique ID for each Automate node?

lancewf: I don't think I know what the `instanceName` does.

components/authz-service/server/v2/project_update_manager.go
```diff
@@ -523,3 +524,233 @@ func getEstimatedTimeCompleteInSec(event *automate_event.EventMsg) (time.Time, e
...
+
+		if subWorkflow.Err() != nil {
+			errMsg = fmt.Sprintf("%s; %s: %s", errMsg, d, subWorkflow.Err().Error())
+		}
+	}
+
+	return errMsg
+}
+
+func (m *CerealProjectUpdateManager) PercentageComplete() float64 {
```
lancewf: What we were doing before for PercentageComplete was using the domain service with the longest estimated time to complete's PercentageComplete. What might be easier would be to use the domain service with the lowest PercentageComplete. If there were three domain services with 10%, 35%, and 95% all running at the same time, we would return the entire process is 10% complete. Here we are assuming that the 35% and 95% tasks will complete before the 10% one. If the task does not complete before the others then their percentage will become the new PercentageComplete. 
I could not think of a way of adding the domain percentage completes together because we do not know how much work each of them has to do. The average does not really make sense to me because some of the tasks are going to be completed a lot faster than the others. For example, say we have three domain tasks that have been running for a minute. Two of the tasks are complete but the other one takes an hour. I would think the percentage would be 1/60 = 1.6% complete, but the below calculation would give 1/3+1/3+1/60/3 = 67.2%.

components/compliance-service/habitat/default.toml
```diff
@@ -22,4 +22,4 @@ manual_polling_interval = 1440
 [agent]
 workers = 10
 buffer_size = 1000
-remote_inspec_version = "REPLACE-FROM-PLAN.SH"
+remote_inspec_version = "4.7.24"
```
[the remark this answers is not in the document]
lancewf: ➕

components/authz-service/server/v2/project.go
jaym: unique id would be problematic. By using a constant instance name, we guarantee only one of these runs at a time

jaym: singleton is probably a better name

components/authz-service/server/v2/project_update_manager.go
jaym: Up to you. I can wire one up if it's helpful

components/authz-service/server/v2/project.go
jaym: it might be a test thing, not sure. But if you have ideas on how you'd like to see this bit organized, i could make it happen. Right now, the domain services and authz need to agree on all of these values. Also, I need to not repeat this twice

lancewf: Oh, I see the `NewProjectUpdateManager` was already being called twice. 

Disregard my question.

components/authz-service/server/v2/project_update_manager.go
lancewf: I don't think we would need it anymore because the ProjectUpdateID was used before to ensure what update the cancel, start, and status requests were talking about. This was because with using the event-service old requests for prior project updates could have appeared during the next project update. 

If I am thinking of how the cereal service is working correctly, then that coordination problem could not happen.

jaym: Ya, all the work that happens is getting tied to 1 workflow id

components/authz-service/server/v2/project_update_manager.go
jaym: if percentage should be tied to time, then # complete tasks doesn't make sense. I'll change it back to how it was before

lib/authz/project_rules_update_workflow.go
```diff
@@ -0,0 +1,265 @@
...
+	projectTagUpdaterStatusTaskName string
+}
+
+type DomainProjectUpdateWorkflowParameters struct {
+	ProjectUpdateID string
+}
+
+type DomainProjectUpdateWorkflowPayload struct {
+	ProjectUpdateID             string
+	ESJobID                     []string
```
lancewf:
```suggestion
	ESJobIDs                     []string
```

lib/authz/project_rules_update_workflow.go
```diff
@@ -0,0 +1,265 @@
...
+			payload.ConsecutiveJobCheckFailures++
+			if payload.ConsecutiveJobCheckFailures > maxNumberOfConsecutiveFails {
+				return w.Fail(err)
+			}
+			w.EnqueueTask(m.projectTagUpdaterStatusTaskName, statusParameters{payload.ESJobID}, cereal.StartAfter(m.nextCheck())) // nolint: errcheck
+			return w.Continue(payload)
+		}
+		mergedJobStatus := JobStatus{}
+		if err := ev.Result.Get(&mergedJobStatus); err != nil {
+			// TODO (jaym): Should we cancel here?
```
lancewf: I agree you should fail here.

lib/authz/project_rules_update_workflow.go
```diff
@@ -0,0 +1,265 @@
...
+				"Failed to deserialize %s result", m.startProjectTagUpdaterTaskName)
+			return w.Fail(err)
+		}
+		if len(esJobIDs) <= 0 {
+			err := errors.New("0 ES jobs started")
+			logrus.WithError(err).Errorf(
+				"Failed to deserialize %s result", m.startProjectTagUpdaterTaskName)
+			return w.Fail(err)
+		}
+		w.EnqueueTask(m.projectTagUpdaterStatusTaskName, statusParameters{esJobIDs}, cereal.StartAfter(m.nextCheck())) // nolint: errcheck
```
lancewf: Why is another "start project update" queued after a "start project update" is completed.

lancewf:
I really enjoyed reading through this solution. I see this will be a great pattern for Automate. 
One question I had is where or how is the `DomainProjectUpdateWorkflowExecutor` marked as complete so the next update can be started? 

Below is the diagram I used to understand the interactions between the services more. 
image

jaym:
There's a little bit of magic you need to peal back to understand that. The way it works is that each service gets a `DomainProjectUpdateWorkflowExecutor` to manage the logic of starting the update and completing it. These are all run by authz. The magical part is that these are tied into 1 workflow my the MultiWorkflow workflow executor. You can see it's create in multiworkflow.go takes in workflow executors. So, each DomainProjectUpdateWorkflowExecutor will think the instance it is processing is complete when the `mergedJobStatus` says it is completed. However, this is more of a virtual instance, and it wont really be complete until all the domain services complete. Only at this point will authz be allowed to ask it to do it again.

jaym:
replaced by #1346

PR #1080, opened 2019-07-25
ci: move back to docker-based studio, now with labels

stevendanna:
This moves us back to the linux executor with a docker based studio.
We are moving back because of

https://github.com/buildkite-plugins/docker-buildkite-plugin/issues/127

which is still present in the build that we have.

### DIFF
--- a/scripts/verify_build.sh
+++ b/scripts/verify_build.sh
@@ -63,7 +63,8 @@ done
 if [[ "$build_commands" != "" ]]; then
     # We override HAB_CACHE_KEY_PATH to ensure we only see the key we
     # generated in this build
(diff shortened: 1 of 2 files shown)


scripts/verify_build.sh
```diff
@@ -63,7 +63,8 @@ done
 if [[ "$build_commands" != "" ]]; then
     # We override HAB_CACHE_KEY_PATH to ensure we only see the key we
     # generated in this build
-    HAB_ORIGIN=chef HAB_CACHE_KEY_PATH=$RESOLVED_RESULTS_DIR DO_CHECK=true hab studio run "source .studiorc; set -e; $build_commands"
+    export HAB_DOCKER_OPTS="--label buildkitejob=$BUILDKITE_JOB_ID"
```
stevendanna: I can't tell if we want `--tty` or not anymore.

jaym: probably leave it
change 1 thing at a time

PR #1083, opened 2019-07-26
Add support for filtering service groups by app, env and name

kmacgugan:
To avoid a query with substring we added a new column for name suffix to service group.
Technically the name suffix is actually the name of the service group, and the alternative
would be to compose the service.service_group name upon query.

App, Env and service group name all result in complete service group returns
this allows us to continue to use the view for these filters.

- [ ] Code actually executed?

### DIFF
--- a/components/applications-service/pkg/storage/postgres/service_groups.go
+++ b/components/applications-service/pkg/storage/postgres/service_groups.go
@@ -51,46 +51,31 @@ const (
 FROM service_group_health AS service_group_health_counts
 `
 
-	selectServiceGroupHealthWithPageSort = `
+	selectServiceGroupHealth = `
 SELECT * FROM service_group_health AS service_group_health
- ORDER BY %s
- LIMIT $1
-OFFSET $2
 `
 	selectServiceGroupHealthFilterCRITICAL = `
-SELECT * FROM service_group_health AS service_group_health
- WHERE health_critical > 0
- ORDER BY %s
- LIMIT $1
-OFFSET $2
+   health_critical > 0
 `
 	selectServiceGroupHealthFilterUNKNOWN = `
-SELECT * FROM service_group_health AS service_group_health
- WHERE health_unknown  > 0
+   health_unknown  > 0
    AND health_critical = 0
- ORDER BY %s
- LIMIT $1
-OFFSET $2
 `
 	selectServiceGroupHealthFilterWARNING = `
-SELECT * FROM service_group_health AS service_group_health
- WHERE health_warning  > 0
+   health_warning  > 0
    AND health_critical = 0
    AND health_unknown  = 0
- ORDER BY %s
- LIMIT $1
-OFFSET $2
 `
 	selectServiceGroupHealthFilterOK = `
-SELECT * FROM service_group_health AS service_group_health
- WHERE health_ok > 0
+   health_ok > 0
    AND health_critical = 0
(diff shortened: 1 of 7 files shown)


components/applications-service/pkg/storage/postgres/service_groups.go
```diff
@@ -337,6 +352,14 @@ func queryFromStatusFilter(text string) (string, error) {
 	}
 }
 
+func queryFromFieldFilter(field string, envs []string) (string, error) {
```
danielsdeleo: Forgot to rename the `env` arg perhaps?

components/applications-service/pkg/storage/postgres/service_groups.go
```diff
@@ -337,6 +352,14 @@ func queryFromStatusFilter(text string) (string, error) {
...
+	formattedVals := strings.Join(envs[:], "','")
+	formattedVals = "'" + formattedVals + "'"
+	fmt.Println(formattedVals)
```
danielsdeleo: leftover debug?

components/applications-service/pkg/storage/postgres/service_groups.go
```diff
@@ -337,6 +352,14 @@ func queryFromStatusFilter(text string) (string, error) {
...
+	query := ` ` + field + ` in (` + formattedVals + `)`
```
danielsdeleo: Do we want to uppercase `in` for style pedantry?

components/applications-service/pkg/storage/postgres/service_groups.go
```diff
@@ -337,6 +352,14 @@ func queryFromStatusFilter(text string) (string, error) {
 	}
 }
 
+func queryFromFieldFilter(field string, envs []string) (string, error) {
+	formattedVals := strings.Join(envs[:], "','")
```
danielsdeleo: do we need to restrict the characters we allow here to prevent SQL nastiness or does that happen elsewhere?

components/applications-service/pkg/storage/postgres/service_groups.go
```diff
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/chef/automate/components/applications-service/pkg/storage"
+	"github.com/chef/automate/components/secrets-service/utils"
```
afiune: I would suggest that, instead of depending on a component's code, we should move this code to a library. I have verified and there is more than one place where we define and use these functions so, it makes more sense for us to maintain a single codebase.  

https://github.com/chef/automate/search?q=IsSqlSafe&unscoped_q=IsSqlSafe

danielsdeleo: Agree 100%. We already noticed this and are planning to do it in a separate PR so we don't have features for one team mixed with a refactor that affects a lot of teams in one patch.

stevendanna: I agree.  We should avoid cross-component dependencies if we don't need them.  I'd say either copy the code or put it in lib.  I think there is maybe a good argument for a pgutils.  Some other functions that might be helpful there:

https://github.com/chef/automate/blob/master/lib/platform/pg/queries.go#L91-L129
https://github.com/chef/automate/blob/master/lib/cereal/postgres/postgres.go#L890-L902

and probably a number of others around the code base.

PR #1092, opened 2019-07-26
license-control-service: remove use of bindata library

stevendanna:
The use of this library seemed like overkill in this service as we
were getting data from a single file. In this change, I now directly
generate the file with a go template.


components/license-control-service/pkg/keys/keys.go
```diff
@@ -1,13 +1,11 @@
 package keys
 
 // Generate key bindata
-//go:generate curl --fail --silent --output data/keys.json http://license-acceptance.chef.co/keys
-//go:generate go-bindata -pkg $GOPACKAGE -o keys.bindata.go data/...
+//go:generate go run ../../tools/gen-keys https://license-generation-service.chef.co/keys keys.bindata.go
```
stevendanna: Note if you try this out locally, you need to be on the VPN.

components/license-control-service/tools/gen-keys/main.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"encoding/json"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"text/template"
+	"time"
+
+	"github.com/sirupsen/logrus"
```
srenatus: 💭 Does such a small utility need logrus? Fine. 😉

yzl: Not directly related to the PR, but...do we *have* a standard for what logging framework to use when?

srenatus: I don't think so. But also, I guess I shouldn't scold anyone for using what we use everywhere everywhere 😉 :ignoreme:

PR #1096, opened 2019-07-28
Make cereal service

jaym:
This change adds cereal as a service. Having cereal be a service will allow us to run workflows that span across services more easily. 
For project updates, it also avoids introducing cyclic dependencies as IPC for managing the project update workflow will happen through the cereal service. This means that authz will not have to talk to ingest to manage (start/status/cancel) a workflow. Instead, all services will connect to cereal service. The domain services (ingest, compliance, event-feed) will be pulling tasks from cereal service. authz will manage the entire workflow of project update.


components/cereal-service/pkg/server/server.go
```diff
@@ -0,0 +1,632 @@
...
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	type chanMsg struct {
+		Msg *cereal.DequeueWorkflowRequest
+		Err error
+	}
+
+	in := make(chan chanMsg)
+	go func() {
```
stevendanna: I'm not 100% following why these goroutines are necessary.

jaym: the send and recv functions don't take a context. Their underlying context expires when you return from the function that is part of the grpc interface that was generated. If you want to do a read with a context (for example one that times out), you need to do it in another goroutine. it's horrible

stevendanna: Ah, I missed the fact that the calling function is passing a context that has been given a timeout so I was confused about why we were going through all of this trouble when the context would only be canceled by the client going away anyway, but now that I see the Timeout it makes a lot more sense.

components/automate-deployment/pkg/server/server.go
```diff
@@ -450,7 +450,8 @@ func usesPlatformScaffolding(service *deployment.Service) bool {
 	switch service.Name() {
 	case "secrets-service", "pg-sidecar-service", "teams-service", "authz-service", "session-service", "automate-dex",
 		"authn-service", "compliance-service", "nodemanager-service", "notifications-service", "applications-service",
-		"automate-cs-bookshelf", "automate-cs-oc-erchef", "automate-cs-oc-bifrost", "ingest-service", "license-control-service":
+		"automate-cs-bookshelf", "automate-cs-oc-erchef", "automate-cs-oc-bifrost", "ingest-service", "license-control-service",
+		"cereal-service":
```
yzl: I merged #1126, so this can go into a package.meta.

components/cereal-service/habitat/hooks/run
```diff
@@ -0,0 +1,10 @@
+#!{{pkgPathFor "core/bash"}}/bin/bash -e
+
+exec 2>&1
+
+# Call the script to block until user accepts the MLSA via the package's config
+{{pkgPathFor "chef/mlsa"}}/bin/accept {{cfg.mlsa.accept}}
+
+pg-helper ensure-service-database "cereal_service"
```
ryancragun: Did we stop prefacing db's with `chef_`? Or am I imagining that we do that?

components/cereal-service/package.meta
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "chef/cereal-service",
+  "uses_platform_scaffolding": true
```
ryancragun: :heart_eyes:

lib/cereal/integration/schedule.go
```diff
@@ -227,7 +232,9 @@ func (suite *CerealTestSuite) TestExpiringSchedule() {
 
 	found := false
 	schedules, err := m.ListWorkflowSchedules(context.Background())
+	suite.Require().NoError(err)
 	for _, s := range schedules {
+		spew.Dump(s)
```
ryancragun: don't think we want this

api/interservice/cereal/cereal.proto
```diff
@@ -0,0 +1,210 @@
...
+message CreateWorkflowScheduleResponse {}
+
+message ListWorkflowSchedulesRequest { string domain = 1; }
+
+message Schedule {
+  string domain = 1;
+  string instance_name = 2;
+  string workflow_name = 3;
+  bytes parameters = 4;
+  string recurrence = 5;
```
ryancragun: Interesting. I made a struct for Recurrence in the branch that I'm working on but I didn't realize I could encode an rrule as a string and pass it this way. I think I like this a lot better.

components/cereal-service/habitat/hooks/run
jaym: i think you're right

api/config/cereal/config_request.proto
```diff
@@ -0,0 +1,37 @@
...
+      chef.automate.infra.config.Mlsa mlsa = 1;
+      chef.automate.infra.config.TLSCredentials tls = 2;
+      Service service = 3;
+      chef.automate.infra.config.Log log = 4;
+
+      message Service {
+        google.protobuf.StringValue host = 1;
+        google.protobuf.Int32Value port = 2
+            [ (chef.automate.api.port) =
+                  {name : "service", default : 10101, protocol : "grpc"} ];
```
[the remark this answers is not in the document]
jaym: that's how clang formatted it

components/cereal-service/pkg/server/server.go
```diff
@@ -0,0 +1,632 @@
...
+		"instance_name": req.InstanceName,
+	})
+	logctx.Info("enqueuing workflow")
+	if err := s.backend.EnqueueWorkflow(ctx, &backend.WorkflowInstance{
+		InstanceName: req.InstanceName,
+		WorkflowName: req.WorkflowName,
+		Parameters:   req.Parameters,
+	}); err != nil {
+		logctx.WithError(err).Error("failed to enqueue workflow")
+		if err == libcereal.ErrWorkflowInstanceExists {
```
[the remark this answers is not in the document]
jaym: they don't have a meaning in this context

components/cereal-service/integration/integration_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"os"
+	"testing"
+	"time"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/chef/automate/components/cereal-service/pkg/server"
+	"github.com/chef/automate/lib/cereal"
+
+	grpcceral "github.com/chef/automate/api/interservice/cereal"
```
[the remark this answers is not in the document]
jaym: there's a package clash between lib/cereal and the protobuf generated stuff interservice/cereal

api/interservice/cereal/cereal.proto
jaym: ya, i think we should all standardize on the string representation. compliance already uses it and you can describe anything you'd want with it

components/cereal-service/pkg/server/server.go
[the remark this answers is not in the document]
ryancragun: Why? I'm not sure I understand why we wouldn't return any other error encountered.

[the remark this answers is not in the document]
jaym: Here's a list of the errors:
```
	ErrNoTasks                  = errors.New("no tasks in queue")
	ErrNoWorkflowInstances      = errors.New("no workflow instances in queue")
	ErrWorkflowScheduleExists   = errors.New("workflow schedule already exists")
	ErrWorkflowInstanceExists   = errors.New("workflow instance already exists")
	ErrNoDueWorkflows           = errors.New("no due workflows")
	ErrNoScheduledWorkflows     = errors.New("no workflows are scheduled")
	ErrInvalidSchedule          = errors.New("workflow schedule is not valid")
	ErrWorkflowInstanceNotFound = errors.New("workflow instance not found")
	ErrWorkflowScheduleNotFound = errors.New("workflow schedule not found")
	ErrWorkflowNotComplete      = errors.New("workflow instance is still running")
	ErrTaskLost                 = errors.New("task lost before reporting its status")
```
ErrNoTasks is only returned when polling for tasks
ErrNoWOrkflowInstances is returned when polling for workflow instances
ErrWorkflowScheduleExists is returned when trying to put in a schedule
ErrNoDueWorkflows is used in the context of scheduling workflow instances
ErrNoScheduledWorkflows is used in the context of scheduling workflow instances (this can probably be folded up into another error)
ErrInvalidSchedule is used in the context of creating schedules (i guess scheduling too if you messed with the database)
ErrWorkflowInstanceNotFound is only returned when you're looking for a workflow instance
ErrWorkflowScheduleNotFound is returned when you're looking for a schedule
ErrWorkflowNotComplete is used when you try to get the result of an incomplete workflow instance
ErrTaskLost is only used in the context of processing tasks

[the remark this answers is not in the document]
jaym: oh wait, i think i see what you're saying

components/cereal-service/integration/integration_test.go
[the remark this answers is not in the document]
ryancragun: Roger, and it's super pedantic and I'm sorry, but we could probably come up with a name that doesn't spell it wrong or clash. `cereal_api`?

PR #1097, opened 2019-07-29
deps: bump grpc-gateway: 1.9.4 -> 1.9.5

srenatus:
⚠️ This time, we've got to expect trouble: the GRPC status
`FailedPreconditon` is now -- correctly, as it seems -- translated to HTTP
status `400 Bad Request`.

I think we might rely on this being translated to `412 Precondition
Failed` in a few places, at least from grepping around.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -481,7 +481,7 @@
   version = "v1.2.0"
 
 [[projects]]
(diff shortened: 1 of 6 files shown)


srenatus:
Looks good in CI, we're probably OK 😃

PR #1098, opened 2019-07-29
auth-stuff: reduce clutter in pipeline defs

srenatus:
- OPA doesn't need a studio for its tests
- local-user-service doesn't need a studio for its tests, combine static and unit again.
- installing the CLI `opa` in the Makefile is probably not needed
- remove `privileged: true` from some deployment-service steps that don't need them

### DIFF
--- a/scripts/install_hab_pkg.sh
+++ b/scripts/install_hab_pkg.sh
@@ -5,5 +5,5 @@ export HAB_NONINTERACTIVE=true
 export HAB_NOCOLORING=true
 export HAB_LICENSE="accept-no-persist"
 
-ident="${1:-ident required}"
+ident="${1?ident required}"
 hab pkg install -b "$ident"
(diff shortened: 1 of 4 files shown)


scripts/install_hab_pkg.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -euo pipefail
+
+export HAB_NONINTERACTIVE=true
+export HAB_NOCOLORING=true
+export HAB_LICENSE="accept-no-persist"
+
+ident="${1:-ident required}"
```
stevendanna: Should we use `?` here instead? 
```suggestion
ident="${1?ident required}"
```

srenatus: my bad! thanks

srenatus:
_DC-oh-no!_

PR #1100, opened 2019-07-29
license-control-service: store state in postgresql

stevendanna:

The goal of this PR is to move the license state in the license-control-service
into postgresql so that in a multi-node world license updates are seen on all
nodes.

### :+1: Definition of Done

If 2 license-control-services are configured against the same backend, an Update call on one machine should be seen on the other machine.

- [x] Tests added/updated?
- [x] Decide if we want to keep history of license updates
- [x] Do we want to cache the parsed license or are we OK doing the parse on each read?
- [x] Docs added/updated?


api/interservice/license_control/license-control.proto
```diff
@@ -4,7 +4,7 @@ package chef.automate.domain.license_control;
 option go_package = "github.com/chef/automate/api/interservice/license_control";
 
 import "google/protobuf/timestamp.proto";
-import "github.com/chef/automate/lib/license/license.proto";
+import "lib/license/license.proto";
```
stevendanna: This change fixes reflection

components/license-control-service/pkg/server/grpc_test.go
```diff
@@ -1,56 +1,28 @@
 package server_test
 
-import (
-	"context"
-	"os"
-	"testing"
+// func TestHealthCheck(t *testing.T) {
```
stevendanna: I can re-enable this once I get pg wired into the tests again

api/interservice/license_control/license-control.proto
srenatus: We've got a test helper for that somewhere; having run into the same bit before. 
👉 https://github.com/chef/automate/blob/master/lib/grpc/grpctest/reflection.go#L16

stevendanna:
> Do we want to cache the parsed license or are we OK doing the parse on each read?

I think I'm going to YAGNI this for now.  Happy to discuss it if others think it is critical.

components/license-control-service/pkg/storage/storage.go
```diff
@@ -0,0 +1,261 @@
+package storage
```
stevendanna: This new package makes it a bit easier to inject a different storage system.  I added this when I though I might just do unit tests without pg, but then I integrated pg into the unit tests anyway.  However, I like the isolation this gives still.

components/license-control-service/pkg/keys/keys.go
```diff
@@ -6,15 +6,50 @@ package keys
...
+	ConfiguredAt time.Time
+}
+
+func NewLicenseParser(pub PublicKeysData) *LicenseParser {
+	return &LicenseParser{
+		keyData: LoadPublicKeys(pub),
+	}
+}
+
+func (l *LicenseParser) Parse(licenseData string) (*lic.License, error) {
```
stevendanna: I needed this in both `server` and `backend` which is why this code moved here.  Putting it in `keys` is a _little_ odd but it isn't completely unrelated and I couldn't deal with another thing called "license" in this service.

components/license-control-service/pkg/storage/storage.go
```diff
@@ -0,0 +1,261 @@
...
+	// returend.
+	GetLicense(context.Context) (string, error)
+	// Cleanup removes the state kept by this backend. After
+	// Cleanup is called, GetLicense should return a
+	// NoLicenseError.
+	Cleanup(context.Context) error
+}
+
+var _ CurrentBackend = &PGBackend{}
+var _ UpgradeableBackend = &FileBackend{}
```
srenatus: Interesting. I'm used to `var _ UpgradeableBackend = (*FileBackened)(nil)`. Not much of a difference, though. 😄

components/license-control-service/pkg/storage/storage.go
```diff
@@ -0,0 +1,261 @@
...
+
+func NewCurrentBackend(pgURL string, migrationPath string, legacyFilePath string) CurrentBackend {
```
srenatus: [irrelevant] 
```suggestion
func NewCurrentBackend(pgURL, migrationPath, legacyFilePath string) CurrentBackend {
```

components/license-control-service/pkg/storage/storage.go
```diff
@@ -0,0 +1,261 @@
...
+	pgURL             string
+	migrationPath     string
+	legacyFileBackend UpgradeableBackend
+
+	db          *sql.DB
+	initialized bool
+}
+
+func (p *PGBackend) Init(ctx context.Context, l *keys.LicenseParser) error {
+	err := p.doInit(ctx, l)
```
srenatus: Hmm could use `sync.Once`, too, I guess? But probably doesn't matter since we ensure this is only called once? I guess every handler could call init, and it would only happen on the first run, and we wouldn't have to worry about "not initialized" conditions? 🤔

components/license-control-service/pkg/storage/storage.go
```diff
@@ -0,0 +1,261 @@
...
+
+func (p *PGBackend) GetLicense(ctx context.Context) (string, keys.LicenseMetadata, error) {
+	if !p.initialized {
+		return "", keys.LicenseMetadata{}, &NotInitializedError{backend: "sql"}
+	}
+	return p.doGetLicense(ctx)
+}
+
+func (p *PGBackend) doGetLicense(ctx context.Context) (string, keys.LicenseMetadata, error) {
+	row := p.db.QueryRowContext(ctx, "SELECT configured_at, data FROM licenses WHERE active=TRUE")
```
srenatus: [nit]
```suggestion
	row := p.db.QueryRowContext(ctx, "SELECT configured_at, data FROM licenses WHERE active")
```

components/license-control-service/pkg/storage/storage.go
stevendanna: Oh, that is a good idea -- just to be sure that there is likely no runtime cost to this.

components/license-control-service/pkg/storage/storage.go
stevendanna: Yeah, to be honest I'm not sure how much any of it is needed.  The goal was to make sure the callers always called Init() before any of the other methods.

I suppose a different way to do that would be to define

```
type UnintializedCurrentBackend interface {
    Init(context.Context, *keys.LicenseParser) (CurrentBackend, error)
}
```

and then the caller could only get a CurrentBackend via Init()...

components/license-control-service/pkg/storage/storage.go
srenatus: Guess it's probably superfluous, when `NewCurrentBackend(...) CurrentBackend` returns `&PGBackend{...}`, it should be sufficiently checked...

components/license-control-service/pkg/storage/storage.go
srenatus: 🤔 That's a nice idea. 💡 

But really, I wonder how much we try to foresee our own future mistakes -- I'm not sure we actually build precautions in the right places...

stevendanna: Right, in this case, you still need tests that will fail if this error is hit and if you have that test, you can just have a test that fails on the inevitable error that you'll get if GetLicense is called before Init().

I think I'll remove it, it'll also clean up a lot of code.

components/automate-deployment/pkg/server/server.go
```diff
@@ -450,7 +450,7 @@ func usesPlatformScaffolding(service *deployment.Service) bool {
 	switch service.Name() {
 	case "secrets-service", "pg-sidecar-service", "teams-service", "authz-service", "session-service", "automate-dex",
 		"authn-service", "compliance-service", "nodemanager-service", "notifications-service", "applications-service",
-		"automate-cs-bookshelf", "automate-cs-oc-erchef", "automate-cs-oc-bifrost", "ingest-service":
+		"automate-cs-bookshelf", "automate-cs-oc-erchef", "automate-cs-oc-bifrost", "ingest-service", "license-control-service":
```
jaym: sorry

components/license-control-service/habitat/hooks/run
```diff
@@ -1,9 +1,15 @@
 #!{{pkgPathFor "core/bash"}}/bin/bash
 
+set -e
+
 exec 2>&1
 
 # Call the script to block until user accepts the MLSA via the package's config
 {{pkgPathFor "chef/mlsa"}}/bin/accept {{cfg.mlsa.accept}}
 
+cp -r {{pkg.path}}/migrations {{pkg.svc_static_path}}
```
jaym: can we point to the migrations in pkg.path instead of copying?

components/license-control-service/pkg/server/server.go
```diff
@@ -274,90 +142,106 @@ func (s *LicenseControlServer) Update(ctx context.Context, req *lc.UpdateRequest
...
-	if err := s.persistLicenseToken(ctx, licData); err != nil {
-		log.WithFields(
-			log.Fields{
-				"license_id": license.Id,
-				"error":      err,
-			},
-		).Error("Unable to persist license")
+//Telemetry endpoint to return telemetry configuration
+func (s *LicenseControlServer) Telemetry(ctx context.Context, req *lc.TelemetryRequest) (*lc.TelemetryResponse, error) {
+	return &lc.TelemetryResponse{TelemetryEnabled: s.TelemetryEnabled, TelemetryUrl: s.TelemetryURL}, nil
```
jaym: is this on disk state that we will need to care about?

components/license-control-service/pkg/storage/storage.go
```diff
@@ -0,0 +1,248 @@
...
+		_, err := l.Parse(license)
+		if err != nil {
+			logrus.WithError(err).Warn("invalid license found on disk")
+			// this key is never going to be valid, so we
+			// just start up.
+			return nil
+		}
+
+		err = p.SetLicense(ctx, license)
+		if err != nil {
```
jaym: will this fail if the license is already in the database?

jaym: looks like no

components/license-control-service/pkg/server/server.go
stevendanna: This is something that needs discussion.  The *current* method of setting this is via the config:

https://automate.chef.io/docs/telemetry/#opting-out-of-telemetry

however, there is a legacy on-disk state for telemetry that the configuration looks for at server startup.

components/automate-deployment/pkg/server/server.go
ryancragun: It's almost time to invert this list :trollface:

components/license-control-service/pkg/storage/storage.go
```diff
@@ -0,0 +1,248 @@
...
+// have a configured License.
+type NoLicenseError struct{ backend string }
+
+func (n *NoLicenseError) Error() string {
+	return fmt.Sprintf("No license found in %s storage backend", n.backend)
+}
+
+// RetriableBackendError is the error returned by a backend when an
+// update has failed but may work if retried.
+type RetriableBackendError struct{ err error }
```
ryancragun: For some reason I thought we had a linter that would explode unless these were named in the `ErrRetriableBackend` style. Guess not!

components/license-control-service/pkg/storage/storage.go
```diff
@@ -0,0 +1,248 @@
...
+		return nil
+	case *NoLicenseError:
+		return p.migrateFromFileBackend(ctx, l)
+	default:
+		return err
+	}
+}
+
+// TODO(ssd) 2019-07-29: This migration would be problematic in
+// multi-node, but this should only happen on single node clusters.
```
ryancragun: why? I assume it's because to migrate to HA you'll have to do some other things that would make this upgrade occur on a single-node, but would be nice to know the reasoning.

components/license-control-service/pkg/storage/storage.go
srenatus: 👉 `FooError` for types, `ErrFoo` for values: https://github.com/golang/go/wiki/Errors#naming

I _wish_ we had a linter for that 😃

PR #1105, opened 2019-07-29
Quick styling updates to local_target.go

afiune:


components/automate-deployment/pkg/target/local_target.go
```diff
@@ -341,7 +340,7 @@ func (t *LocalTarget) SetUserToml(name, config string) error {
 		}
 	}()
 
-	_, err = file.Write([]byte(config))
+	_, err = file.WriteString(config)
```
stevendanna: Note to self: No AtomicWriter here?

afiune: That would be the best but, this will only happen if someone runs the installation twice at the same time in the same machine, right?

stevendanna: The fileutils.AtomicWriter also protects against partial write to a file in the case of a crash/powerloss/etc.

PR #1114, opened 2019-07-30
Accept Habitat Healthchecks via "Plain" NATS Messaging

danielsdeleo:

This enables Automate to receive health check events from habitat via "plain" NATS messaging instead of NATS-streaming (though nats-streaming continues to work; this patch makes automate subscribe to messages from both systems).

Messages sent via "plain" NATS messaging are not persisted in transit, so this change increases the possibility that messages will be lost, for example when the event service restarts. While not ideal, this is considered acceptable because the behavior of the overall system is generally convergent: habitat is sending a complete picture of the state of a service as understood by Automate with each health check message, which is sent by default every 30s. Automate does not keep a historical record of service states. Therefore, once Automate's ingestion system becomes available after a service interruption, it should recover a full picture of the state of all services within 30s.

In the future if we introduce new features that need more reliable delivery, we will reintroduce nats-streaming or use some other means to achieve more reliable delivery.

 There are two reasons we are making this change:
1. There is currently not a library available for rust that supports nats-streaming and also works correctly with TLS in all the configurations we need to support (such as self-signed certs). Supporting TLS is more important than reliable delivery of health check messages.
2. Past performance testing showed reason to suspect that persistence of in-transit messages reduced system throughput. For health check messages, which are currently the only message type and likely to be the vast majority of events in the future, we believe improved throughput will be worth the reduced delivery reliability.

https://github.com/chef/automate/issues/1124

Since we are just changing the internals of the messaging protocol and the data generators have been updated, the change can only really be observed via debug logs.

Start all services in the studio, then build applications-service and event-gateway. Patch the config with the following to enable debug logging for applications service:

```toml
[applications]
  [applications.v1]
    [applications.v1.sys]
      [applications.v1.sys.log]
        level = "debug"
```

Run the following to publish a message via the front-end event-gateway:
```sh
install_if_missing "$(app_service_origin)/applications-service" applications-publisher;
t=$(get_api_token)
applications-publisher --auth-token $t
```

Use `sl` to tail the log. You should see a log message like this:

```
[2019-07-30T22:41:32+00:00] applications-service.default(O): time="2019-07-30T22:41:32Z" level=debug msg="Message received" message_data="\n0\n\n1234567890\x1a\v\b\x9c\x8f\x83\xea\x05\x10\xc4\xd2\xf9$\"\x04demo*\x04demo:\tlocalhost\x120\n\x1fcore/redis/0.1.0/20190730224132\x1a\rredis.default" message_subject=habitat.event.healthcheck protocol="nats messaging"
```

The `protocol="nats messaging"` means the message was sent via "plain" NATS instead of NATS-streaming.

### DIFF
--- a/components/applications-service/pkg/nats/nats.go
+++ b/components/applications-service/pkg/nats/nats.go
@@ -39,7 +39,8 @@ type NatsClient struct {
 	durableID string
 	subject   string
 	certs.TLSConfig
-	conn               stan.Conn
+	msgConn            *natsc.Conn
+	streamConn         stan.Conn
 	retries            int
 	InsecureSkipVerify bool
 	DisableTLS         bool
@@ -95,9 +96,10 @@ func NewAndConnect(url, id string, tlsConfig certs.TLSConfig) (*NatsClient, erro
 // of retries and instead continue until we succeed. (when the server is up)
 func (nc *NatsClient) Connect() error {
 	var (
-		conn  stan.Conn
-		err   error
-		tries = uint64(0)
+		msgConn    *natsc.Conn
+		streamConn stan.Conn
+		err        error
+		tries      = uint64(0)
 	)
 
 	tlsConf, err := nc.natsTLSConfig()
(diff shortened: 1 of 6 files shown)


components/applications-service/pkg/nats/nats.go
```diff
@@ -15,7 +15,12 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-const connectionRetries = 5
+const (
+	connectionRetries                      = 5
+	deprecatedNATSStreamHealthCheckChannel = "habitat"
+	natsMessagingHealthcheckSubject        = "habitat.event.healthcheck"
```
danielsdeleo: @gpeers this is the NATS messaging subject we'll be listening to.

Issue #1101, opened 2019-07-29
refresh on <chef-page> brings back the top nav


susanev:
## Overview
When on a `<chef-page>` page if you refresh the top-nav comes back, and it should not.

## Repro Steps
* `chef-automate iam upgrade-to-v2` if not already on iam v2.0
* navigate to _Settings_ > _Policies_ > _any policy detail page_ > _Members_ tab
* use the _Add Members_ button
* refresh the page, and notice that the top-nav comes back

PR #1119, opened 2019-07-31
Automate-1101 Top nav shows on page refresh for page components Sign…

tarablack01:
…ed-Off-By: Tara Black <tblack@chef.io>

When on a <chef-page> page if you refresh the top-nav comes back, and it should not.

https://github.com/chef/automate/issues/1101

### :+1: Definition of Done
Top nav should not display for page component pages

chef-automate iam upgrade-to-v2 if not already on iam v2.0
navigate to Settings > Policies > any policy detail page > Members tab
use the Add Members button
refresh the page, and notice that the top-nav comes back

### DIFF
--- a/components/automate-ui/src/app/ui.component.ts
+++ b/components/automate-ui/src/app/ui.component.ts
@@ -63,9 +63,7 @@ export class UIComponent implements OnInit {
   ngOnInit(): void {
     this.router.events.pipe(
         filter(event => event instanceof ActivationStart)
-    ).subscribe((event: any) => {
-      this.renderNavbar = !event.snapshot.data.hideNavBar;
-    });
+    ).subscribe((event: any) => this.renderNavbar = !event.snapshot.data.hideNavBar);
 
     this.store.dispatch(new GetIamVersion());
   }
--- a/components/automate-ui/src/app/ui.component.ts
+++ b/components/automate-ui/src/app/ui.component.ts
@@ -47,7 +47,7 @@ export class UIComponent implements OnInit {
     private router: Router
   ) {
     this.notifications$ = store.select(notificationState);
-    
+
     // ActivationEnd specifically needs to be here in the constructor to catch early events.
     this.router.events.pipe(
       filter(event => {
--- a/components/automate-ui/src/app/ui.component.ts
+++ b/components/automate-ui/src/app/ui.component.ts
@@ -1,5 +1,5 @@
 import { Component, OnInit } from '@angular/core';
-import { ActivatedRoute, Router, ActivationStart, RoutesRecognized } from '@angular/router';
+import { ActivationStart, ActivationEnd, Router } from '@angular/router';
 import { Store } from '@ngrx/store';
 import { Observable } from 'rxjs';
 import { filter } from 'rxjs/operators';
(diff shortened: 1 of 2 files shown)


susanev:
@tarablack01 tried this out, and it seems the top-nav is missing from most pages now...

missing-top-nav

susanev:
@tarablack01 it seems the top-nav is missing from some pages now where it should be
if i go directly to `/event-feed`, `/client-runs`, `/applications`, or `user-details/admin` im not seeing any top nav

missing-top-nav

components/automate-ui/src/app/ui.component.ts
```diff
@@ -44,20 +44,25 @@ export class UIComponent implements OnInit {
...
+        filter(event => {
+          return event instanceof ActivationStart;
+        })
     ).subscribe((event: any) => {
       this.renderNavbar = event.snapshot.data.hideNavBar
         ? false
         : true;
     });
 
+    this.renderNavbar = this.route.snapshot.data.hideNavBar;
```
msorens: Two issues. First is polarity:
```suggestion
    this.renderNavbar = !this.route.snapshot.data.hideNavBar;
```
Second, I don't have a solution for as yet... but the `data` is empty; it is not returning the static route data that it seems like it should be. That is why no pages are getting the top nav bar, as @susanev observed.

components/automate-ui/src/app/ui.component.ts
```diff
@@ -1,5 +1,5 @@
 import { Component, OnInit } from '@angular/core';
-import { Router, ActivationStart } from '@angular/router';
+import { ActivatedRoute, Router, ActivationStart, RoutesRecognized } from '@angular/router';
```
msorens: RoutesRecognized is not used; delete.

tarablack01: @msorens Good catch! Removed

components/automate-ui/src/app/ui.component.ts
tarablack01: @msorens Tested more and value is available in constructor with first ActivationEnd, updated.

tarablack01:
@susanev @msorens Updated code to check the first NavigationEnd event for data value in constructor function.

components/automate-ui/src/app/ui.component.ts
```diff
@@ -47,15 +47,25 @@ export class UIComponent implements OnInit {
     private router: Router
   ) {
     this.notifications$ = store.select(notificationState);
+
+    this.router.events.pipe(
```
msorens: Since typically these would go in `ngOnInit` consider adding a comment above this line saying
```
// ActivationEnd specifically needs to be here in the constructor to catch early events.

Comment (tarablack01) [reply components/automate-ui/src/app/ui.component.ts]: @msorens added comment

File: components/automate-ui/src/app/ui.component.ts [hunk 4]
```diff
@@ -47,15 +47,25 @@ export class UIComponent implements OnInit {
...
+      filter(event => {
+        return event instanceof ActivationEnd;
+      })
```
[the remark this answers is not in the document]
tarablack01: @msorens Left from testing, changed back to single statement.

PR #1128, opened 2019-07-31
[auth] make rules RESTful

bcmdarroch:
We updated the paths of the Rules API to make them more RESTful and make it more obvious that rules are a sub-resource of projects, and cannot exist outside the context of a project.

In addition to updating the paths, we also updated the error handling so that on every rule API, we check that the project provided in the path exists before completing the operation on the rule. This way you must provide both the correct project ID and rule ID in order to successfully complete an operation on that rule.

NOTE: The API only cares if the project ID in the path is correct. It will ignore any `project_id` provided in the body of the request. We can't remove `project_id` from the message definition, so we will need to make sure we document the correct way to pass the project in any Rule request.

### :+1: Definition of Done

- [x] Rule paths match what's defined in the [frontend requests file](https://github.com/chef/automate/blob/master/components/automate-ui/src/app/entities/rules/rule.requests.ts)
- [x] each Rule API reacts with a NotFound error when the project in the path is incorrect for the requested rule

prep: 
```
start_all_services
rebuild components/automate-gateway && rebuild components/authz-service
export TOK=`chef-automate iam token create test --admin`

# create projects
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects -d "$(jq -n '{ id: "foo", name: "my foo project"}')" | jq .   
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects -d "$(jq -n '{ id: "bar", name: "my bar project"}')" | jq .
```

with the correct project in the path:
```
# create rule
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects/foo/rules -d "$(jq -n '{ id: "rule1", name: "my foo rule", type: "NODE", project_id: "foo", conditions: [{ operator: "EQUALS", attribute: "CHEF_ORGS", values: ["The Misfits"]}]}')" | jq .

# update rule
curl -kH "api-token: $TOK" -XPUT https://localhost/apis/iam/v2beta/projects/foo/rules/rule1 -d "$(jq -n '{ name: "my foo rule 2", type: "NODE", project_id: "foo", conditions: [{ operator: "EQUALS", attribute: "CHEF_ORGS", values: ["The Avengers"]}]}')" | jq .

# get rule
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects/foo/rules/rule1 | jq .

# get project's rules
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects/foo/rules | jq .

# delete rule
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects/foo/rules/rule1 | jq .
```

with the incorrect project in the path:
```
# create rule (again)
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects/foo/rules -d "$(jq -n '{ id: "rule1", name: "my foo rule", type: "NODE", project_id: "foo", conditions: [{ operator: "EQUALS", attribute: "CHEF_ORGS", values: ["The Misfits"]}]}')" | jq .

# update rule
curl -kH "api-token: $TOK" -XPUT https://localhost/apis/iam/v2beta/projects/bar/rules/rule1 -d "$(jq -n '{ name: "my foo rule 2", type: "NODE", project_id: "foo", conditions: [{ operator: "EQUALS", attribute: "CHEF_ORGS", values: ["The Avengers"]}]}')" | jq .

# get rule
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects/bar/rules/rule1 | jq .

# get project's rules
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects/bar/rules | jq .

# delete rule
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects/foo/bar/rule1 | jq .
```

- [x] Tests added/updated?
- [x] Docs added/updated?


components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1508,8 +1517,8 @@ func (p *pg) singleRowResultOrNotFoundErr(result sql.Result) error {
 	return nil
 }
 
-func (p *pg) getMapOfRuleAssociations(ctx context.Context, q Querier, id string) (map[string]bool, error) {
-	assocRow := q.QueryRowContext(ctx, `SELECT query_rule_table_associations($1);`, id)
+func (p *pg) getMapOfRuleAssociations(ctx context.Context, q Querier, id string, projectID string) (map[string]bool, error) {
+	assocRow := q.QueryRowContext(ctx, `SELECT query_rule_table_associations($1, $2);`, id, projectID)
```
srenatus: [nit]
```suggestion
	assocRow := q.QueryRowContext(ctx, "SELECT query_rule_table_associations($1, $2)", id, projectID)
```

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -1165,16 +1165,25 @@ func (p *pg) ListRulesForProject(ctx context.Context, projectID string) ([]*v2.R
 		return nil, v2.RulesStatusError, err
 	}
 
-	tx, err := p.db.BeginTx(ctx, nil)
-	if err != nil {
-		return nil, v2.RulesStatusError, p.processError(err)
+	// in our other APIs we use a a postgres query to do filtering
+	// however in this case, we can't automatically assume NoRows means NotFound
+	// because we want to differentiate between a project that is not in the project filter
+	// and a project that has no rules
```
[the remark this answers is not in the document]
blakestier: A project that has no rules will still return an empty array instead of NotFound-- unless the user doesn't have permission to see that project, in which case they will see NotFound.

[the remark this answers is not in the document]
bcmdarroch: we do want a project that has no rules to return an empty array.
the issue we found was that when we do the filtering check [here](https://github.com/chef/automate/blob/master/components/authz-service/storage/postgres/migration/sql/54_use_project_db_id_in_references_of_project_rules.up.sql#L72), we found we return an empty array for both cases when the project has no rules and the project is not found.

maybe you have some ideas how to keep it in the query? i agree, i don't like the special casing

components/automate-gateway/api/iam/v2beta/rules.proto
```diff
@@ -15,56 +15,45 @@ import "components/automate-grpc/protoc-gen-policy/iam/annotations.proto";
...
 
   rpc ListRulesForProject (ListRulesForProjectReq) returns (ListRulesForProjectResp) {
     option (google.api.http).get = "/iam/v2beta/projects/{id}/rules";
-    option (chef.automate.api.policy).resource = "auth:projects:{id}:rules";
-    option (chef.automate.api.policy).action = "read";
-    option (chef.automate.api.iam.policy).resource = "iam:projects:{id}:rules";
-    option (chef.automate.api.iam.policy).action = "iam:rules:list";
+    option (chef.automate.api.policy).resource = "auth:projects:{id}";
+    option (chef.automate.api.policy).action = "get";
+    option (chef.automate.api.iam.policy).resource = "iam:projects:{id}";
```
msorens: Should use `project_id` for consistency, no?

components/automate-gateway/api/iam/v2beta/rules.proto
```diff
@@ -15,56 +15,45 @@ import "components/automate-grpc/protoc-gen-policy/iam/annotations.proto";
...
-  // TODO (tc): Right now we don't have plans to use this in the UI
-  // and it doesn't return the staged rules like all the other query APIs.
-  // Should probably delete it or update it to return staged rules.
-  rpc ListRules (ListRulesReq) returns (ListRulesResp) {
-    option (google.api.http).get = "/iam/v2beta/rules";
-    option (chef.automate.api.policy).resource = "auth:rules";
-    option (chef.automate.api.policy).action = "read";
-    option (chef.automate.api.iam.policy).resource = "iam:rules";
-    option (chef.automate.api.iam.policy).action = "iam:rules:list";
+    option (chef.automate.api.iam.policy).resource = "iam:projects:{project_id}";
```
msorens: This does not quite make sense. This endpoint is returning all rules, not rules for a project. Maybe it no longer makes sense to keep this...?

components/authz-service/server/v2/project.go
```diff
@@ -401,13 +401,17 @@ func (s *ProjectState) UpdateRule(ctx context.Context, req *api.UpdateRuleReq) (
...
+	case nil: // continue
+	case storage_errors.ErrNotFound:
+		return nil, status.Errorf(codes.NotFound, "rule with ID %q not found", req.Id)
+	case storage_errors.ErrChangeProjectForRule:
+		return nil, status.Errorf(codes.FailedPrecondition,
+			"cannot change project_id for existing rule with ID %q ", req.Id)
+	default:
+		switch err.(type) {
+		case *storage_errors.ForeignKeyError:
+			return nil, status.Error(codes.NotFound, err.Error())
```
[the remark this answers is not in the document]
srenatus: It's because we've been mixing error constants and error types. The types of error constants are not informative, and can't be used in a `switch err.(type)` construct.

Yes, turning our error constants into error types would help, but
1. what we have here doesn't hurt _that much_
2. if we do it, let's do it in a separate PR (the nested construct here exists in other places, too)

components/automate-gateway/api/iam/v2beta/rules.proto
blakestier: @msorens this endpoint just returns a single rule, right? If not, I'm def confused!

blakestier: (We did remove `ListRules`, if that's what you're thinking of?)

components/automate-gateway/api/iam/v2beta/rules.proto
bcmdarroch: We thought that this fits with the convention of using `id` to identify the relevant resource in any API. (in this case, it's the `id` of the project). In the APIs with two IDs, we chose to use `project_id` to differentiate from the rule `id`, since the rule is specific resource being operated on. We also went this way to minimize cascading backend changes. 

That being said, in writing this PR, @blakestier and I saw that it would make a lot of sense to bring all of these APIs under the project API (in which case we might just always use `id` for the project as the primary resource). I'll add a tech debt card to do so.

PR #1132, opened 2019-08-01
Remove unnecessary copies from pkg.path

stevendanna:
These services were copying files that are only used for reading out
of package path and into the service's static path. No one I talked to
had a reason for why we were doing this.

We think maybe one of the following let us here:

1) permissions problems on files. We now test with a restrictive
default umask and it is unclear to me what permissions problem we
could have hit that the `cp` would have avoided.

2) Following a "rule of thumb" that the service should only depend on
files in `/hab/svc/SERVICE-NAME`. While I think this is a good rule
of thumb, for static data (this data could have been compiled into the
binary), I don't see the harm in reading it directly from the package
path.

Removing the copy comes with the advantage of removing another step
from service startup that could fail.

There is still calls to `cp` in our run scripts, but they are in
services that I'm not inclined to touch at the moment:

    components/automate-cs-nginx/habitat/hooks/run
    15:cp {{cfg.required_recipe.path}} {{pkg.svc_var_path}}/required_recipe

    components/automate-cs-oc-erchef/habitat/hooks/run
    56:cp {{pkg.svc_config_path}}/dark_launch_features.json {{pkg.svc_data_path}}

    components/automate-workflow-server/habitat/hooks/run
    26:  cp -a {{pkg.path}}/{{cfg.ssh_git.git_repo_template}} {{pkg.svc_files_path}}/

### DIFF
--- a/components/applications-service/habitat/config/config.toml
+++ b/components/applications-service/habitat/config/config.toml
@@ -20,7 +20,7 @@ cluster_id = "{{events.cfg.cluster_id}}"
 
 [postgres]
 database = "{{cfg.storage.database}}"
-schema_path = "{{pkg.svc_static_path}}/schema"
+schema_path = "{{pkg.path}}/schema"
 
 [tls]
 cert_path = "{{pkg.svc_config_path}}/service.crt"
--- a/components/applications-service/habitat/hooks/run
+++ b/components/applications-service/habitat/hooks/run
@@ -1,5 +1,7 @@
 #!{{pkgPathFor "core/bash"}}/bin/bash
 
+set -e
+
 exec 2>&1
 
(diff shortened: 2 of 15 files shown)


stevendanna:
> Do we need cleanups of what was copied before?

We could. It's kinda a bummer but I suppose it is probably good to avoid any confusion.

srenatus:
ℹ️ Added a commit cleaning out the old files. It got green on BK, so the worst thing that could have happened is that a path was wrong, `rm -rf` didn't care, and we end up not having cleaned up something. It's probably fine. I'd appreciate another set of eyes on this before we merge it.

srenatus:
Huh I thought this was merged already. Well. Let's get to it, then. 😄

PR #1139, opened 2019-08-01
compliance-service: prealloc linter

stevendanna:
This fixes all the prealloc violations and simplifies our LINTERARGS
based on changes to the main linter configuration.

The only remaining differences in compliance-service from the standard
configuration is the nakedret violations. Fixing those looks _mostly_
straightforward but there are a few with some subtle behavior so it is
probably best done as its own PR.


stevendanna:
Built on #1138

srenatus:
Rebased

PR #1141, opened 2019-08-01
Ensure `environment.prod.ts` is used w/ prod builds

scottopherson:

`angular.json` production target was not configured correctly to load the `environment.prod.ts` config and was instead using the nearly identical default environment config. This would result in `devMode` being executed within production builds.

The `prod` config target in `angular.json` is not used, `production` is the target that's used in production builds. All other production settings appear to be correct; minification, aot-compiliation, etc has been working as intended.

### :+1: Definition of Done

`automate-ui` does not exec `devMode` within production builds.

`npm run build:prod`

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/automate-ui/habitat/plan.sh
```diff
@@ -32,9 +32,6 @@ do_unpack() {
   mkdir -p $CACHE_PATH/chef-ui-library
   rsync --archive --exclude node_modules $PLAN_CONTEXT/.. $CACHE_PATH/automate-ui
   rsync --archive --exclude node_modules $PLAN_CONTEXT/../../chef-ui-library $CACHE_PATH
-
-  # Copy our alternate environment file into place
-  cp ${PLAN_CONTEXT}/environments/environment.prod.ts $CACHE_PATH/automate-ui/src/environments/environment.prod.ts
```
scottopherson: Does anyone know why `environment.prod.ts` was moved from `src/environments` to `habitat/environments`, only to be copied back to `src/environments` during a hab build?

srenatus: no idea

msorens: Strangely, that is how it started life. Apparently when `automate-ui` was created in the old a2 repository, that's where it was: https://github.com/chef/a2/blob/9b70a49d9c3d31613294420a1951da7e4172ae87/components/automate-ui/habitat/environments/environment.prod.ts

phiggins: If I'm remembering right, it was because the file in `habitat/` was a template that had some values injected into it before being copied into the right place. I don't know when/why that changed. There is probably more build related cruft that doesn't need to happen anymore.

This comment appears in a few places and I'm not sure it's relevant anymore but it is referring to the old path and should get updated/deleted: https://github.com/chef/automate/blob/30e0f2eb3ff2f042e1e65058b91228c47250f692/components/automate-ui/src/environments/environment.ts#L1-L12

PR #1145, opened 2019-08-01
Return 400 if no statements passed to policy create or update

tylercloke:

If a user tries to create or update a policy with no statements, return a 400. Added 400 cases in cypress. Have a lot of tests for other cases in inspec but porting them to cypress was out of the scope of this PR.

```
start_all_services
# run cypress tests. see readme there.
```

- [x] Tests added/updated?
- [ ] Docs added/updated?


components/authz-service/server/v2/policy_test.go
```diff
@@ -999,6 +999,14 @@ func TestUpdatePolicy(t *testing.T) {
 		{"fails with InvalidArgument when missing policy ID", func(t *testing.T) {
 			req := api_v2.UpdatePolicyReq{
 				Name: "testPolicy1",
+				Statements: []*api_v2.Statement{
```
srenatus: [nit] Guess we could have introduced a variable and re-used that instead of repeating the same block again and again

e2e/cypress/support/commands.ts
```diff
@@ -119,3 +119,47 @@ function LoginHelper(username: string) {
...
+
+Cypress.Commands.add('cleanupPoliciesByIDPrefix', (idToken: string, idPrefix: string) => {
+  cy.request({
+    auth: { bearer: idToken },
+    method: 'GET',
+    url: '/apis/iam/v2beta/policies',
+    failOnStatusCode: false
+  }).then((resp) => {
+    const body = resp.body;
+    for (const policy of body.policies) {
```
srenatus: For-loops are so golang, let's do something fancier 😉 
```typescript
resp.body.find(pol => pol.id.startsWith(idPrefix))
  .forEach(pol => cy.request({
    auth: { bearer: idToken },
    method: 'DELETE',
    url: `/apis/iam/v2beta/policies/${policy.id}`,
    failOnStatusCode: false
  }));
```
⚠️ untested, probably wrong 😄

api/interservice/authz/v2/policy.proto
```diff
@@ -34,7 +34,7 @@ message CreatePolicyReq {
             }
         }
        }];
-    repeated Statement statements = 4;
+    repeated Statement statements = 4 [(validate.rules).repeated.min_items = 1];
```
blakestier: 😍

components/authz-service/server/v2/policy_test.go
blakestier: That sounds worth it to me for the sake of test maintainability 🤷‍♂

e2e/cypress/support/commands.ts
tylercloke: This pattern is in a few places so I'll fix in my other PR after I rebase.

PR #1146, opened 2019-08-01
Refactor duplicated utils to common lib

kmacgugan:
Move error utils to lib/errorutils
Move pg/sql utils to lib/pgutils

- [ ] Tests added/updated?
- [ ] Docs added/updated?


api/external/secrets/secrets.go
```diff
@@ -166,8 +167,13 @@ func mapToKvs(m map[string]string) []*Kv {
 // will not add an error and be considered valid.
 func requiredField(value string, message string, errorList []*error) []*error {
 	if value == "" {
+<<<<<<< HEAD
 		newErrors := append(make([]*error, 0), errorList...)
 		err := errors.New(message)
+=======
+		newErrors := append(make([]*error, 0), errors...)
+		err := errorutils.ProcessInvalid(nil, message)
+>>>>>>> Refactor duplicated utils to common lib
```
stevendanna: I think you've conflicted with my recent cleanup to just rip out this dependency from this location

kmacgugan: Whoops, I missed this one. I saw your cleanup, and am wondering about InvalidSecretError. I wanted to take it out of the FormatErrorMsg func so that func can be shared but am not sure where to put it or if we can remove it.

stevendanna: I didn't go that route just because I looked at a lot of that error handling code and wasn't convinced it is really the kind of code we want to share via lib.  It is really only duplicated because we happen to copy/pasta services out of compliance.  That said, as I'm not the one typically working in these services, I'm happy with whatever y'all choose.

components/compliance-service/habitat/default.toml
```diff
@@ -22,4 +22,4 @@ manual_polling_interval = 1440
 [agent]
 workers = 10
 buffer_size = 1000
-remote_inspec_version = "REPLACE-FROM-PLAN.SH"
+remote_inspec_version = "4.7.24"
```
danielsdeleo: Just curious, why's this change here?

lancewf: This change should not be merged.

kmacgugan: fixing it, I am not sure how it got there, maybe from generated stuff?

lancewf: Yeah, it happens when you build the compliance-service hab package. I am really not sure why.

components/compliance-service/dao/pgdb/filtering.go
```diff
@@ -5,11 +5,12 @@ import (
...
 
-	"github.com/pkg/errors"
-	"github.com/sirupsen/logrus"
-
 	"github.com/chef/automate/components/compliance-service/api/common"
 	"github.com/chef/automate/components/compliance-service/utils"
+	"github.com/chef/automate/lib/errorutils"
+	"github.com/chef/automate/lib/pgutils"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
```
afiune: I think we are trying to separate the internal go packages and the external one, also ordering them by name cc @srenatus @msorens since they have been fixing the multiple times with PRs like this one https://github.com/chef/automate/pull/1159

PR #1151, opened 2019-08-02
ci: shellcheck rendered hooks

srenatus:

We have a bunch of bash scripts that are not shellchecked at the moment, because they're habitat templates. This PR introduces shellcheck into that, by shellchecking **the rendered hook scripts**. Naturally, they fail validation, so the findings don't break the build, but are merely reported and ignored.

At some later time, we can remove the `|| return 0` -- when the issues have been resolved.

👉 [These are the current findings](https://buildkite.com/chef/chef-automate-master-verify-private/builds/4052#d9d4e702-f302-49da-94f5-3f87df8e08b1/484)

### DIFF
--- a/integration/tests/shellcheck_rendered_hooks.sh
+++ b/integration/tests/shellcheck_rendered_hooks.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+#shellcheck disable=SC2034
+test_name="shellcheck hooks"
+test_deploy_inspec_profiles=()
+test_skip_diagnostics=true
+
+do_setup() {
+    do_setup_default
+
+    # We are defaulting to a umask of 077 to test
+    # installations on systems that are super locked down.
+    # Briefly override that strict default so we can install
+    # packages that non-root users can use (like the hab user
+    # for health checks or this script).
+    local previous_umask
+    previous_umask=$(umask)
+    umask 022
+
+    hab pkg install -b core/shellcheck
+
+    umask "$previous_umask"
+}
+
+do_test_deploy() {
+    shellcheck -s bash -x /hab/svc/*/hooks/*
+}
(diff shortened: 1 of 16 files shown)


srenatus:
> Are these /integration/ tests? 🤔 💭 I feel they are more like lint or syntax.... Any reason to put it inside the integration folder?

We cannot shellcheck the hooks unless they're are **rendered** -- otherwise, stuff like `cp {{pkg.path}}/foo {{ pkg.svc_static_path}}/whatever` will make shellcheck freak out. So, to still have some shellcheck coverage on our hook, these tests work on the rendered files. The only way I know get them is to have automate start up properly. Hence, we need the entire thing running. Since it shares the framework for getting it up with other integration tests, I've chosen that.

afiune:
@srenatus now that is the reason I was missing, thanks for the clarification!!! 👍

integration/tests/shellcheck_rendered_hooks.sh
```diff
@@ -0,0 +1,28 @@
...
+#shellcheck disable=SC2034
+test_name="shellcheck hooks"
+test_deploy_inspec_profiles=()
+test_skip_diagnostics=true
+
+do_setup() {
+    do_setup_default
+
+    # We are defaulting to a umask of 077 to test
+    # installations on systems that are super locked down.
```
msorens: Suggest:
```suggestion
    # installations on systems that are completely locked down.
```

srenatus: Unfortunately, I've copy-pasted that bit from the other places. I propose we turn it into a helper and fix the comment there.

srenatus:
> but unless we have a plan to fix the issues and start enforcing shellcheck I'm not sure how much value this will bring.

I'm all ears. Besides fixing that in this PR, I have no idea. Might do it on the side, but no promises 😉

srenatus:
@ryancragun Tried to fix all findings. I'd appreciate another quick glance. 😉

PR #1147, opened 2019-08-01
[auth] Update Condition Attribute enum values

bcmdarroch:
### :nut_and_bolt: Description: What code changed, and why?
The allowed values for a rule's Condition Attribute were a mix of plural and singular and didn't match how they're meant to be shown in the UI. Now they've been updated to match this list provided by superteam:
```
*Node Attributes*
Chef Organization
Chef Server
Environment
Chef Role
Chef Tag
Chef Policy Name
Chef Policy Group

*Event Attributes*
Chef Organization
Chef Server
````

### :athletic_shoe: How to Build and Test the Change
```
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects -d "$(jq -n '{ id: "bar", name: "my bar project"}')" | jq .

# no longer works
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects/bar/rules -d "$(jq -n '{ id: "rule2", name: "my foo rule", type: "NODE", conditions: [{ operator: "EQUALS", attribute: "CHEF_ORGS", values: ["The Misfits"]}]}')" | jq .

# works
curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects/bar/rules -d "$(jq -n '{ id: "rule2", name: "my foo rule", type: "NODE", conditions: [{ operator: "EQUALS", attribute: "CHEF_ORGANIZATION", values: ["The Misfits"]}]}')" | jq .
```

### :white_check_mark: Checklist

- [x] Tests added/updated?
- [x] Docs added/updated?

### :camera: Screenshots, if applicable

PR #1156, opened 2019-08-02
[automate-854] bring project details to life

bcmdarroch:
Mock data -> real live rule data from the backend! 🎉 

Thanks to @tarablack01's hard work, there wasn't too much to do here aside from tweak a few things to get everything working correctly! 🙌 

updating rule paths to be restful: https://github.com/chef/automate/pull/1147

### :+1: Definition of Done
- [x] Rule list on the Project Details page (https://a2-dev.test/settings/projects/my-new-project/rules) is populated by the GetRulesForProject API
- [x] Rule Details page (https://a2-dev.test/settings/projects/my-new-project/rules/new-rule) is populated by the GetRule API
- [x] Create Rule page creates a real rule using the CreateRule API
- [x] Update Rule page updates the rule using the UpdateRule API
- [x] Delete Rule deletes the rule using the DeleteRule
- [ ] Cypress integration tests: will add in follow-up PR

1. Navigate to https://a2-dev.test/settings/projects
2. Create a new project
3. Create a new rule for that project, it will be set to `Edits pending`
4. Open the rule's details page
5. Go back to https://a2-dev.test/settings/projects and click `Update Projects`
6. Back on your project details page, the new rule you've made will be set to `Applied`
7. Update the rule
8. Delete the rule

- [x] Tests added/updated?
- [x] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/pages/project/details/project-details.component.html
+++ b/components/automate-ui/src/app/pages/project/details/project-details.component.html
@@ -80,7 +80,7 @@
                 </chef-td>
                 <chef-td>{{ rule.id }}</chef-td>
                 <chef-td>{{ rule.type | titlecase }}</chef-td>
-                <chef-td>{{ rule.conditions.length | pluralize : 'condition' : 's' }}</chef-td>
+                <chef-td>{{ rule.conditions.length | pluralize : 'condition' : 'conditions' }}</chef-td>
                 <chef-td>{{ getEditStatus(rule) }}</chef-td>
                 <chef-td class="controls">
                   <app-authorized [allOf]="['/iam/v2beta/projects/{id}', 'delete', project?.id]">
--- a/components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
+++ b/components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
@@ -21,7 +21,10 @@ import {
 import {
   getRuleAttributes, getStatus, updateStatus, ruleFromRoute, createError, createStatus
 } from 'app/entities/rules/rule.selectors';
-import { projectFromRoute } from 'app/entities/projects/project.selectors';
+import {
+  getStatus as getProjectStatus,
+  projectFromRoute
+} from 'app/entities/projects/project.selectors';
 import { Project } from 'app/entities/projects/project.model';
 import { GetProject } from 'app/entities/projects/project.actions';
 
@@ -72,15 +75,17 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
 
       combineLatest(
         this.store.select(getStatus),
-        this.store.select(updateStatus)
+        this.store.select(updateStatus),
+        this.store.select(getProjectStatus)
       ).pipe(
         takeUntil(this.isDestroyed),
-        map(([gStatus, uStatus]) => {
+        map(([gStatus, uStatus, gpStatus]) => {
           const routeId = this.route.snapshot.paramMap.get('ruleid');
           this.isLoading =
             routeId
             ? (gStatus !== EntityStatus.loadingSuccess) ||
-              (uStatus === EntityStatus.loading)
+              (uStatus === EntityStatus.loading) ||
+              (gpStatus !== EntityStatus.loadingSuccess)
             : false;
         })).subscribe();
 
@@ -92,12 +97,20 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
         map(([pId, rId]) => {
           const project_id = pId as string;
           const rule_id = rId as string;
+          this.store.dispatch(new GetProject({ id: project_id }));
           this.store.dispatch(new GetRule({
             id: rule_id,
             project_id: project_id
           }));
-          this.store.dispatch(new GetProject({ id: project_id }));
         })).subscribe();
+
+      this.store.select(projectFromRoute).pipe(
+        filter(identity),
+        takeUntil(this.isDestroyed),
+        map((state) => {
+          this.project = <Project>Object.assign({}, state);
+        })
+      ).subscribe();
 
       this.store.select(ruleFromRoute).pipe(
         filter(identity),
@@ -105,14 +118,6 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
         map((state) => {
           this.rule = <Rule>Object.assign({}, state);
           this.editingRule = true;
-        })
-        ).subscribe();
-
-      this.store.select(projectFromRoute).pipe(
-        filter(identity),
-        takeUntil(this.isDestroyed),
-        map((state) => {
-          this.project = <Project>Object.assign({}, state);
         })
         ).subscribe();
 
@@ -133,9 +138,9 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
       name: [
         this.rule.name || '',
         [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]],
-      id: ['',
-        [Validators.required, Validators.pattern(Regex.patterns.ID), Validators.maxLength(64)]],
-      type: [{ value: this.rule.type || '', disabled: this.editingRule } , Validators.required],
+      id: [{ value: this.rule.id || '' },
+      [Validators.required, Validators.pattern(Regex.patterns.ID), Validators.maxLength(64)]],
+      type: [{ value: this.rule.type || '', disabled: this.editingRule }, Validators.required],
       conditions: this.fb.array(this.populateConditions())
     });
   }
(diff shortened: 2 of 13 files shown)


components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -84,21 +84,20 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
...
-        takeUntil(this.isDestroyed))
-        .subscribe((id: string) => {
-          this.store.dispatch(new GetProject({ id }));
-        });
+      combineLatest(
+        this.store.select(routeParams).pipe(pluck('id'), filter(identity)),
+        this.store.select(routeParams).pipe(pluck('ruleid'), filter(identity))
+      ).pipe(
+        takeUntil(this.isDestroyed),
+        map(([pId, rId]) => {
```
msorens: You can include types with destructured objects. So if you use this...
```suggestion
        map(([project_id, rule_id]: [string, string]) => {
```
you can then delete L93-94.

components/automate-ui/src/app/pages/project/details/project-details.component.html
```diff
@@ -80,7 +80,7 @@
                 </chef-td>
                 <chef-td>{{ rule.id }}</chef-td>
                 <chef-td>{{ rule.type | titlecase }}</chef-td>
-                <chef-td>{{ rule.conditions.length | pluralize : 'condition' : 's' }}</chef-td>
+                <chef-td>{{ rule.conditions.length | pluralize : 'condition' : 'conditions' }}</chef-td>
```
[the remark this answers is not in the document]
srenatus: 💭 _Oh my! The hours of development time saved by this feature...._ (not-joking: I believe this is too much. The simple and stupid thing might have been preferrable here.)

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -131,7 +130,9 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
 
     this.ruleForm = this.fb.group({
       // Must stay in sync with error checks in project-rules.component.html
-      name: ['', [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]],
+      name: [
+        this.rule.name || '',
```
srenatus: This is odd. Why do we need this?

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -208,20 +221,36 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
   }
 
   createRule() {
```
msorens: If I am understanding your changes here (L224-268), you removed the common rule creation solely because of the differences of `id` and `type`. However, when a rule is edited it is not possible to change the `id` or `type`, so this additional code should not be necessary. If you uncovered a case to the contrary, there may be some other issue to resolve, perhaps.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -285,11 +307,15 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
...
-    if (!this.modifyID && !this.isNavigationKey(event)) {
+    if (!this.modifyID && !this.isNavigationKey(event) && !this.editingRule) {
       this.conflictError = false;
       this.ruleForm.controls.id.setValue(
         IdMapper.transform(this.ruleForm.controls.name.value.trim()));
     }
+
+    if (this.editingRule) {
+      this.ruleForm.markAsDirty();
+    }
```
msorens: 🤔 If you need to set dirty manually here, I think there is something broken in `chef-page`.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
bcmdarroch: Since this component is used for updating and creating, the value needs to either start as a blank string or the existing rule name. This was confusing so we changed the form to initialize as a create form or an update form depending on the value of `editingRule`

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -72,48 +75,49 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
...
+        })).subscribe();
 
-      this.store.select(ruleFromRoute).pipe(
+      this.store.select(projectFromRoute).pipe(
         filter(identity),
         takeUntil(this.isDestroyed),
         map((state) => {
-          this.rule = <Rule>Object.assign({}, state);
-          this.editingRule = true;
+          this.project = <Project>Object.assign({}, state);
```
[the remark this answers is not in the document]
bcmdarroch: slick ✨

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
bcmdarroch: I think the issue is that the conditions are actually an array of forms themselves. Making changes to the attributes doesn't appear to bubble up to the `ruleForm`. Not sure if `chef-page` is related

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
bcmdarroch: thanks for pointing this out, a second look-through helped me find the deeper problem!

When I made these changes, `type` and `id` weren't being set in the update request. On second look, I realized that's because `this.ruleForm.value` only contained `{name: "my new rule", conditions: Array(1)}`. Replacing that variable with `this.ruleForm.controls.<field>.value` seemed to do the trick, allowing us to keep the `convertToRule` as it was 🙌

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -72,48 +75,47 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
 
       combineLatest(
         this.store.select(getStatus),
-        this.store.select(updateStatus)
+        this.store.select(updateStatus),
+        this.store.select(getProjectStatus)
       ).pipe(
         takeUntil(this.isDestroyed),
-        map(([gStatus, uStatus]) => {
+        map(([gStatus, uStatus, gpStatus]: [string, string, string]) => {
```
msorens: For future reference, just realized one could also just use `string[]`, too. 😁

bcmdarroch:
@susanev might take care of those in our follow-up Cypress PR! we'll see where I get with this CSS in 10 mins 😅

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
srenatus: 💭 `string[]` is wider than `[string, string, string]`, though -- `["a", "b", "c", "d"]` is in the former, not the latter. 🤷‍♂

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -72,48 +75,47 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
...
+      combineLatest(
+        this.store.select(routeParams).pipe(pluck('id'), filter(identity)),
+        this.store.select(routeParams).pipe(pluck('ruleid'), filter(identity))
+      ).pipe(
+        takeUntil(this.isDestroyed),
+        map(([project_id, rule_id]: string[]) => {
+          this.store.dispatch(new GetProject({ id: project_id }));
+          this.store.dispatch(new GetRule({
+            id: rule_id,
+            project_id: project_id
```
srenatus: [style] 
```suggestion
            project_id
```

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -285,11 +296,15 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
   }
 
   public handleNameInput(event: KeyboardEvent): void {
-    if (!this.modifyID && !this.isNavigationKey(event)) {
+    if (!this.modifyID && !this.isNavigationKey(event) && !this.editingRule) {
```
srenatus: Hrm I wish we could put more logic like this into the formgroup itself; instead of handling it in the component's state... 🤔

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -299,6 +314,14 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
     this.conflictError = false;
   }
 
+  handleConditionChange(): void {
+    if (this.editingRule) {
+      // changes on the condition forms do not bubble up to the ruleForm
+      // so we explicitly set it to dirty here
```
srenatus: Aren't nested formgroups a thing? I would, too, prefer if we didn't have to manage this manually...

bcmdarroch: yeah I would have expected it to work too. Maybe worth digging into the formgroup docs

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
bcmdarroch: we can get that in our follow-up

PR #1158, opened 2019-08-03
Check apply rules only in IAM v2.1

msorens:

Previously we were checking the apply-rules status at frequent intervals
regardless of the IAM version (which resulted in the noise of useless network calls) and regardless of user access to the endpoint (which resulted in the more significant noise of many 403 errors).

This PR adds awareness of both to clean up both those artifacts.

### :+1: Definition of Done

Network calls only when configured for IAM v2.1 and only with proper access.

rebuild automate-ui
Open the network tab in Dev Tools

Note that you will see calls to introspect access every 5 minutes:
image
And you will see calls to check apply-rules status only if introspection showed access allowed and only in IAM v2.1:
image

### DIFF
--- a/components/automate-ui/src/app/entities/projects/project.effects.ts
+++ b/components/automate-ui/src/app/entities/projects/project.effects.ts
@@ -2,8 +2,11 @@ import { Injectable } from '@angular/core';
 import { HttpErrorResponse } from '@angular/common/http';
 import { Actions, Effect, ofType } from '@ngrx/effects';
 import { interval as observableInterval, of as observableOf } from 'rxjs';
-import { catchError, mergeMap, map, filter, switchMap } from 'rxjs/operators';
+import { catchError, mergeMap, map, filter, switchMap, withLatestFrom } from 'rxjs/operators';
+import { identity } from 'lodash/fp';
+import { Store } from '@ngrx/store';
 
+import { NgrxStateAtom } from '../../ngrx.reducers';
 import { HttpStatus } from 'app/types/types';
 import { CreateNotification } from 'app/entities/notifications/notification.actions';
 import { Type } from 'app/entities/notifications/notification.model';
@@ -40,6 +43,7 @@ import {
 import {
   ProjectRequests
 } from './project.requests';
+import { iamMajorVersion, iamMinorVersion } from 'app/entities/policies/policy.selectors';
 
 const POLLING_INTERVAL_IN_SECONDS = 5;
 
@@ -47,7 +51,8 @@ const POLLING_INTERVAL_IN_SECONDS = 5;
 export class ProjectEffects {
   constructor(
     private actions$: Actions,
-    private requests: ProjectRequests
+    private requests: ProjectRequests,
+    private store: Store<NgrxStateAtom>
   ) { }
 
   @Effect()
@@ -196,7 +201,14 @@ export class ProjectEffects {
           observableOf(new GetApplyRulesStatusFailure(error))))));
 
   @Effect()
-  getLatestApplyRulesStatus$ = observableInterval(1000 * POLLING_INTERVAL_IN_SECONDS)
-    .pipe(map(() => new GetApplyRulesStatus()));
+  getLatestApplyRulesStatus$ = observableInterval(1000 * POLLING_INTERVAL_IN_SECONDS).pipe(
+    withLatestFrom(this.store.select(iamMajorVersion).pipe(filter(identity))),
+    withLatestFrom(this.store.select(iamMinorVersion).pipe(filter(identity))),
+    switchMap(([[_, iamMajorVersion], iamMinorVersion]) => observableOf(iamMajorVersion === 'v2' && iamMinorVersion === 'v1')),
+    filter(identity),
+    mergeMap(() => {
+      this.requests.getApplyRulesStatus().pipe(
+        map(() => new GetApplyRulesStatus()));
+    }));
 }
 
--- a/components/automate-ui/src/app/entities/projects/project.effects.ts
+++ b/components/automate-ui/src/app/entities/projects/project.effects.ts
@@ -11,10 +11,7 @@ import { HttpStatus } from 'app/types/types';
 import { CreateNotification } from 'app/entities/notifications/notification.actions';
 import { Type } from 'app/entities/notifications/notification.model';
 import { iamMajorVersion, iamMinorVersion } from 'app/entities/policies/policy.selectors';
-import { allPerms } from 'app/entities/userperms/userperms.selectors';
-
 import { ProjectRequests } from './project.requests';
-import { ProjectConstants } from './project.model';
 
 import {
   GetProjectsSuccess,
@@ -204,11 +201,8 @@ export class ProjectEffects {
   getLatestApplyRulesStatus$ = observableInterval(1000 * POLLING_INTERVAL_IN_SECONDS).pipe(
     withLatestFrom(this.store.select(iamMajorVersion).pipe(filter(identity))),
     withLatestFrom(this.store.select(iamMinorVersion).pipe(filter(identity))),
-    withLatestFrom(this.store.select(allPerms)),
-    filter(([[[_, major], minor], perms]) => {
-      const permObject = perms[ProjectConstants.APPLY_RULES_ENDPOINT];
-      const allowed = permObject && permObject.get;
-      return allowed && major === 'v2' && minor === 'v1';
+    filter(([[_, major], minor]) => {
+      return major === 'v2' && minor === 'v1';
     }),
     switchMap(() =>
       this.requests.getApplyRulesStatus().pipe(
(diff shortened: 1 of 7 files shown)


components/automate-ui/src/app/entities/projects/project.effects.ts
```diff
@@ -196,7 +202,19 @@ export class ProjectEffects {
...
+    withLatestFrom(this.store.select(iamMinorVersion).pipe(filter(identity))),
+    withLatestFrom(this.store.select(allPerms)),
+    switchMap(([[[_, major], minor], perms]) => {
+      const permObject = perms[ProjectConstants.APPLY_RULES_ENDPOINT];
+      const allowed = permObject && permObject.get;
+      if (allowed && major === 'v2' && minor === 'v1') {
+        return this.requests.getApplyRulesStatus().pipe(
+          map(() => new GetApplyRulesStatus()));
+      } else {
+        return observableOf(new NoopAction());
```
srenatus: ❓ Can we not just _do nothing_? I'd think somehow we should be able to pipe `(allowed && major === 'v2' && minor === 'v1')` and filter based on that....?

msorens: Great suggestion; see latest commit!

components/automate-ui/src/app/entities/projects/project.effects.ts
```diff
@@ -196,7 +201,19 @@ export class ProjectEffects {
...
+    filter(([[[_, major], minor], perms]) => {
+      const permObject = perms[ProjectConstants.APPLY_RULES_ENDPOINT];
+      const allowed = permObject && permObject.get;
+      return allowed && major === 'v2' && minor === 'v1';
+    }),
+    switchMap(() =>
+      this.requests.getApplyRulesStatus().pipe(
+        map((resp) => new GetApplyRulesStatusSuccess(resp)),
+        catchError((error: HttpErrorResponse) =>
+          observableOf(new GetApplyRulesStatusFailure(error))))));
```
msorens: The above block is where the action is happening: every interval collect the data we need from the store--version and permissions--then use a filter to trigger or halt the remaining action, which is to request the rules status.

components/automate-ui/src/app/entities/userperms/userperms.effects.ts
```diff
@@ -47,4 +50,10 @@ export class UserPermEffects {
     mergeMap((action: GetUserParamPerms) => this.requests.fetchParameterized(action.payload).pipe(
       map((resp: UserPermsResponsePayload) => new GetUserParamPermsSuccess(resp.endpoints)),
       catchError((error: HttpErrorResponse) => of(new GetUserParamPermsFailure(error))))));
+
+  @Effect()
+  fetchApplyRulesStatusAllowed$ = interval(1000 * POLLING_INTERVAL_IN_SECONDS).pipe(
+    mergeMap(() => this.requests.fetchSome({ paths: [ProjectConstants.APPLY_RULES_ENDPOINT] }).pipe(
+      map((resp: UserPermsResponsePayload) => new GetSomeUserPermsSuccess(resp.endpoints)),
+      catchError((error: HttpErrorResponse) => of(new GetSomeUserPermsFailure(error))))));
 }
```
msorens: In order to make the go/no-go decision on checking-rule-status we need the specific permission already in the store. This periodically updates that data.
(Elsewhere there is a similar block of code already in place to periodically update the version.)

components/automate-ui/src/app/ui.component.ts
```diff
@@ -58,7 +60,9 @@ export class UIComponent implements OnInit {
         : true;
     });
 
+    // Initial calls for polled events
     this.store.dispatch(new GetIamVersion());
+    this.store.dispatch(new GetSomeUserPerms({ paths: [ProjectConstants.APPLY_RULES_ENDPOINT] }));
```
msorens: ... And this does the initial call at application start (required because `observableInterval` emits at the _end_ of the interval).
Notice the other initial call already there to get the version.

components/automate-ui/src/app/entities/userperms/userperms.effects.ts
```diff
@@ -18,6 +19,8 @@ import {
 } from './userperms.actions';
 import { UserPermsRequests } from './userperms.requests';
 
+const POLLING_INTERVAL_IN_SECONDS = 300; // every 5 minutes
```
srenatus: How about either
```suggestion
const POLLING_INTERVAL_IN_MINUTES = 5;
```
or
```suggestion
const POLLING_INTERVAL_IN_SECONDS = 300;
```
? Informing our readers that 300s = 5m is underestimatinig them 😉

components/automate-ui/src/app/entities/userperms/userperms.effects.ts
srenatus: I believe these kinds of blocks are what got us to introduce `waitForAngular(false)` in our `automate-ui` "e2e" tests. (Back when we did the periodical license status check.) As a consequence, writing tests that aren't flakey has become more difficult.

Do you think spreading these is OK, since we're not that invested into those tests anymore, or should we perhaps revisit the alternatives?

msorens: Interesting point... I suspect the "damage" was done by the first `interval` call, so adding more would not make it worse. 🤷‍♂ But probably not a big concern since, as you say, we're likely moving away from protractor towards Cypress.

components/automate-ui/src/app/entities/projects/project.effects.ts
```diff
@@ -196,7 +201,19 @@ export class ProjectEffects {
...
   @Effect()
-  getLatestApplyRulesStatus$ = observableInterval(1000 * POLLING_INTERVAL_IN_SECONDS)
-    .pipe(map(() => new GetApplyRulesStatus()));
+  getLatestApplyRulesStatus$ = observableInterval(1000 * POLLING_INTERVAL_IN_SECONDS).pipe(
+    withLatestFrom(this.store.select(iamMajorVersion).pipe(filter(identity))),
+    withLatestFrom(this.store.select(iamMinorVersion).pipe(filter(identity))),
+    withLatestFrom(this.store.select(allPerms)),
+    filter(([[[_, major], minor], perms]) => {
+      const permObject = perms[ProjectConstants.APPLY_RULES_ENDPOINT];
+      const allowed = permObject && permObject.get;
```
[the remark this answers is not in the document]
msorens: I agree, the last is a bit much for all-in-one-line.
As to the former, the reason I did not go with that (or with the equivalent `const allowed = perms[ProjectConstants.APPLY_RULES_ENDPOINT].get;`) is that it does not allow for the case of the permObject being undefined.
While rather unlikely, it could happen if for some odd reason the dispatch-on-startup that populates this particular perm got delayed. That is why I have the extra guard in L210.

components/automate-ui/src/app/entities/projects/project.effects.ts
```diff
@@ -203,8 +205,11 @@ export class ProjectEffects {
   getLatestApplyRulesStatus$ = observableInterval(1000 * POLLING_INTERVAL_IN_SECONDS).pipe(
     withLatestFrom(this.store.select(iamMajorVersion).pipe(filter(identity))),
     withLatestFrom(this.store.select(iamMinorVersion).pipe(filter(identity))),
-    switchMap(([[_, major], minor]) => {
-      if (major === 'v2' && minor === 'v1') {
+    withLatestFrom(this.store.select(allPerms)),
+    switchMap(([[[_, major], minor], perms]) => {
```
bcmdarroch: what's the `_` stand for here?

components/automate-ui/src/app/entities/userperms/userperms.effects.ts
```diff
@@ -47,4 +50,10 @@ export class UserPermEffects {
     mergeMap((action: GetUserParamPerms) => this.requests.fetchParameterized(action.payload).pipe(
       map((resp: UserPermsResponsePayload) => new GetUserParamPermsSuccess(resp.endpoints)),
       catchError((error: HttpErrorResponse) => of(new GetUserParamPermsFailure(error))))));
+
+  @Effect()
+  fetchApplyRulesStatusAllowed$ = interval(1000 * POLLING_INTERVAL_IN_SECONDS).pipe(
+    mergeMap(() => this.requests.fetchSome({ paths: [ProjectConstants.APPLY_RULES_ENDPOINT] }).pipe(
+      map((resp: UserPermsResponsePayload) => new GetSomeUserPermsSuccess(resp.endpoints)),
```
bcmdarroch: If the other PR https://github.com/chef/automate/pull/1162 is ensuring all users have access to the apply-rules status endpoint, this extra introspection call seems unnecessary.

components/automate-ui/src/app/entities/projects/project.effects.ts
tylercloke: It just means we are ignoring that value I think.

components/automate-ui/src/app/entities/userperms/userperms.effects.ts
srenatus: > so adding more would not make it worse. 

Debatable! The amount of work required to fix this is increased. 📈

components/automate-ui/src/app/entities/projects/project.effects.ts
bcmdarroch: I meant what would that value be if it weren't ignored?

msorens: The four params are timer_tick, major, minor, and perms; it is not two values for major.

msorens:
Extra perms check removed. ✅

PR #1159, opened 2019-08-04
Periodic spelling sweep

msorens:

This is a periodic sweep of spelling corrections and exceptions across Automate.
Items to correct (or to create an exception for) are identified by running ./scripts/spellcheck.sh from the automate repo root.

All files herein are showing corrections except for cspell.json, which is the exceptions list.

As is typical, there are just a few words exposed to the user, so it makes us look just a wee bit better. All the rest are for the benefit of developers.

(Eventually the spell checker will be wired up to the build itself, but have not gotten to that yet.)


components/notifications-service/server/priv/rules.proto
```diff
@@ -62,7 +62,7 @@ message URLValidationResponse {
     OK = 0;  // we made the request and it succeeded with a 20x response
     ERROR = 1; // We made the request and it failed for any reason.
     INVALID_URL = 4; // the url was not valid.
-    NOTIFICATIONS_UNAVAIALBLE = 98; // we could not make the request due to capacity
+    NOTIFICATIONS_UNAVAILABLE = 98; // we could not make the request due to capacity
```
srenatus: Technically, this is breaking backwards compatibility.

components/applications-service/integration_test/ingestion_test.go
```diff
@@ -10,16 +10,17 @@ import (
...
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	"github.com/chef/automate/api/external/applications"
+	"github.com/chef/automate/api/external/habitat"
+	"github.com/chef/automate/components/applications-service/pkg/storage"
 )
 
-func TestIngestSigleService(t *testing.T) {
+func TestIngestSingleService(t *testing.T) {
```
afiune: that was me! 🤦‍♂

components/notifications-service/server/priv/rules.proto
afiune: Yea, I would not change this enum here, this has to be a separate change... 🤕

components/ingest-service/cmd/ingest-service/commands/serve.go
```diff
@@ -112,8 +112,8 @@ func init() {
 	serveCmd.Flags().String("postgresql-database", "chef_ingest_service", "PG Database name")
 	serveCmd.Flags().Int32("converge-history-days", -1, "Number of days to keep converge history for. A number less than or equal to 0 means data should never be deleted")
 	serveCmd.Flags().Int32("actions-days", -1, "Number of days to keep actions for. A number less than or equal to 0 means data should never be deleted")
-	serveCmd.Flags().Int("max-number-of-bundled-run-msgs", 2500, "The maximum number of run messages to bundle togeter during ingestion")
-	serveCmd.Flags().Int("max-number-of-bundled-action-msgs", 10000, "The maximum number of action messages to bundle togeter during ingestion")
+	serveCmd.Flags().Int("max-number-of-bundled-run-msgs", 2500, "The maximum number of run messages to bundle together during ingestion")
+	serveCmd.Flags().Int("max-number-of-bundled-action-msgs", 10000, "The maximum number of action messages to bundle together during ingestion")
```
lancewf: Sorry, this one was me 🤦‍♂

PR #1160, opened 2019-08-04
Auth input field consistency sweep

msorens:

Another sweep of consistency checks for auth UI code covering:

- All required fields are marked with an asterisk--and in an `aria-hidden` span.
- All input fields disallow whitespace-only values.
- Extraneous leading/trailing whitespace automatically trimmed.
- Field name in the error pop-up matches the displayed field name exactly.

N/A

### :+1: Definition of Done

All auth inputs have identical behavior with respect to the items noted above.

rebuild automate-ui
Navigate through the auth pages to see that any and all input fields behave consistently.

### DIFF
--- a/components/automate-ui/src/app/pages/team-management/team-management.component.ts
+++ b/components/automate-ui/src/app/pages/team-management/team-management.component.ts
@@ -111,7 +111,7 @@ export class TeamManagementComponent implements OnInit, OnDestroy {
 
   public createV2Team(): void {
     this.createTeamCommon({
-      id: this.createTeamForm.controls.id.value.trim(),
+      id: this.createTeamForm.controls.id.value,
       name: this.createTeamForm.controls.name.value.trim(),
       projects: [],
       guid: null
@@ -120,7 +120,7 @@ export class TeamManagementComponent implements OnInit, OnDestroy {
 
   public createV1Team(): void {
     this.createTeamCommon({
-      id: this.createV1TeamForm.controls.name.value.trim(),
+      id: this.createV1TeamForm.controls.name.value,
       name: this.createV1TeamForm.controls.description.value.trim(),
       projects: [],
       guid: null
--- a/components/automate-ui/src/app/pages/user-details/user-details.component.html
+++ b/components/automate-ui/src/app/pages/user-details/user-details.component.html
@@ -25,7 +25,7 @@
             <form [formGroup]="editForm">
               <chef-form-field class="display3">
                 <label>
-                  <span class="label">Full Name</span>
+                  <span class="label">Full Name <span aria-hidden="true">*</span></span>
                   <chef-input ngDefaultControl formControlName="fullName"></chef-input>
                 </label>
                 <chef-error *ngIf="editForm.get('fullName').hasError('required') && editForm.get('fullName').dirty">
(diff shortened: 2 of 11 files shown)


components/automate-ui/src/app/pages/team-management/team-management.component.ts
```diff
@@ -111,7 +111,7 @@ export class TeamManagementComponent implements OnInit, OnDestroy {
 
   public createV2Team(): void {
     this.createTeamCommon({
-      id: this.createTeamForm.controls.id.value.trim(),
+      id: this.createTeamForm.controls.id.value,
```
msorens: IDs are already constrained to have no whitespace, so `trim` is superfluous.

components/automate-ui/src/app/pages/user-details/user-details.component.html
```diff
@@ -25,10 +25,10 @@
             <form [formGroup]="editForm">
               <chef-form-field class="display3">
                 <label>
-                  <span class="label">Full Name</span>
+                  <span class="label">Full Name <span aria-hidden="true">*</span></span>
                   <chef-input ngDefaultControl formControlName="fullName"></chef-input>
                 </label>
-                <chef-error *ngIf="editForm.get('fullName').hasError('required') && editForm.get('fullName').dirty">
+                <chef-error *ngIf="(editForm.get('fullName').hasError('required') || editForm.get('fullName').hasError('pattern')) && editForm.get('fullName').dirty">
```
msorens: The `pattern` here refers to the regex disallowing whitespace-only.

susanev:
@msorens i had intentionally left the asterisks off of the rule attribute, operator, and values because the pattern is a bit different, its not an individual required item. we can chat more tomorrow if you feel strongly they need to be marked.

im also curious if these kinds of _sweeps_ should be across automate in the future, there aren't too many form inputs currently and it would be rad to get them to all be consistent.

msorens:
@susanev 
1. No strong feelings on my part; I was simply trying to channel _you_ consistently. So just call out any of the things here that you do not want.
2. Happy to do all A2 but was hesitant to undertake that without discussion first. 😁

susanev:
cool, yah then lets take them off for the rule attribute, operator, value, thank you!

prolly a good call on a bigger discussion, maybe a good topic for an automate standup?

msorens:
Asterisks on condition fields--removed. ✅

components/automate-ui/src/app/pages/user-details/user-details.component.ts
```diff
@@ -128,12 +129,16 @@ export class UserDetailsComponent implements OnInit, OnDestroy {
   }
 
   private createForms(fb: FormBuilder): void {
+    // Must stay in sync with error checks in user-details.component.html
     this.editForm = fb.group({
-      fullName: ['', Validators.required]
+      fullName: ['', [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]]
```
srenatus: I thought we had talked about this -- would be it a bad idea to put `[Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]` into the constants collection?

components/automate-ui/src/app/pages/user-details/user-details.component.ts
```diff
@@ -173,7 +178,7 @@ export class UserDetailsComponent implements OnInit, OnDestroy {
   }
 
   public updateFullName(): void {
-    const name = this.editForm.get('fullName').value;
+    const name = this.editForm.get('fullName').value.trim();
```
srenatus: Does this interact in some way with the new "mark as pristine" feature? Could this look like a dirty field because there was some space trimmed...?

msorens: Yes, but no. 😉 That is, the pristine/dirty machinations occur _while_ the user is editing. Once the user presses "save", _then_ the field is immediately trimmed before being sent to the back-end, so we store a correct value. (I believe that in every instance in our code the form is then closed so the now out-of-sync on-screen value evaporates.)

But, in one sense, this is actually creating a case that the pristine/dirty mechanism should have caught... but the consequence is negligible: if the user just adds some trailing whitespace we are simply re-saving the thing with the same name as before.

Issue #442, opened 2019-05-29
Add Status Filters to Compliance Node List View


jonong1972:
# User Story

As a Automate User, I would like to filter out the failed/passed/skipped nodes on the compliance node tab so that I may find my failed nodes easier.

# Definition of Done
Page: `/compliance/reports/nodes`

The nodes table will have our standard filter buttons above the nodes table.
Filters:
Total Nodes | Failed Nodes | Passed Nodes | Skipped Nodes

# Demo Script / Repro Steps
1. A user can navigate to goes to Compliance > Nodes Tab
2. User clicks on a status filter
3. System then filters the data table.

vjeffrey:
related: https://github.com/chef/automate/issues/328

PR #1161, opened 2019-08-05
Automate-442 Add status filters to compliance node list view Signed-O…

tarablack01:
…ff-By: Tara Black <tblack@chef.io>

As an Automate User, I would like to filter out the failed/passed/skipped nodes on the compliance node tab so that I may find my failed nodes easier.

https://github.com/chef/automate/issues/442

### :+1: Definition of Done
Page: /compliance/reports/nodes

The nodes table will have our standard filter buttons above the nodes table.

Filters:
Total Nodes | Failed Nodes | Passed Nodes | Skipped Nodes

A user can navigate to goes to Compliance > Nodes Tab
User clicks on a status filter
System then filters the data table.

Screen Shot 2019-08-04 at 8 54 25 PM

### DIFF
--- a/components/automate-ui/package.json
+++ b/components/automate-ui/package.json
@@ -18,8 +18,7 @@
     "test": "ng test --watch=false --code-coverage --source-map=false --configuration=no_auth",
     "test:watch": "ng test --watch=true --browsers='Chrome' --source-map=true --configuration=no_auth",
     "e2e": "ng e2e",
-    "lint": "npm run lint:ts && npm run lint:html && npm run lint:sass",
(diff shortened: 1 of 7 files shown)


susanev:
@jonong1972 tested and yup filters remain selected when you navigate to other pages.

vjeffrey:
"it seems the filter counts are only counting compliance nodes on the current page"
on the api side, it looks like we only return the total nodes. the filtered node counts are, in this pr, counted by filtering the list by passed/failed/skipped, which is why it's only applying to this page.

we have two options:
- make the singular api call actually be 4 calls: one for all, one for each of the status filters
- return total, passed, failed, skipped from the api side

should be a quick change on the backend to return totals per status along with total. we can get it done in the next couple days.

vjeffrey:
i have a WIP branch to include all the totals here. just needs some cleanup and tests. ill do that tomorrow.

PR #1167, opened 2019-08-05
Add API to list workflow instances

jaym:
Added an API to the cereal backend to list workflow instances.

```golang
type Driver interface {
  ListWorkflowInstances(ctx context.Context, opts ListWorkflowOpts) ([]*WorkflowInstance, error)
}

type ListWorkflowOpts struct {
  WorkflowName *string
  InstanceName *string
  IsRunning    *bool
}
```

It allows finding a workflow instance by (workflow name, instance name),
or allows listing all workflows, or all workflows of a certain workflow
name, or all running/not running workflows.

Example usages:

```golang
// List all workflow instances
instances, err = b1.ListWorkflowInstances(ctx, backend.ListWorkflowOpts{})

// List all completed workflow instances
isRunning = false
instances, err = b1.ListWorkflowInstances(ctx, backend.ListWorkflowOpts{
    IsRunning: &isRunning,
})

// List all running workflow instances
isRunning = true
instances, err = b1.ListWorkflowInstances(ctx, backend.ListWorkflowOpts{
    IsRunning: &isRunning,
})

// List all workflow instances of type workflowName
instances, err = b1.ListWorkflowInstances(ctx, backend.ListWorkflowOpts{
    WorkflowName: &workflowName,
})

// List all workflow instances of type workflowName that are currently
// running
instances, err = b1.ListWorkflowInstances(ctx, backend.ListWorkflowOpts{
    WorkflowName: &workflowName,
    IsRunning:    &isRunning,
})

// Find the workflow instance (workflowName, instanceName)
instances, err = b1.ListWorkflowInstances(ctx, backend.ListWorkflowOpts{
    WorkflowName: &workflowName,
    InstanceName: &instanceName
})

```

### DIFF
--- a/lib/cereal/postgres/postgres.go
+++ b/lib/cereal/postgres/postgres.go
@@ -29,28 +29,61 @@ const (
 	failWorkflowQuery     = `SELECT cereal_fail_workflow($1, $2)`
 	continueWorkflowQuery = `SELECT cereal_continue_workflow($1, $2, $3, $4, $5)`
 
+	// listWorkflowInstanceQuery is quite a tricky query. It's intent is to select
+	// one run of each workflow instance. This means it has to consult the actively
+	// running workflow instances from the cereal_workflow_instances table, and the
+	// completed workflow results from cereal_workflow_results.
+	// Parameters:
+	// $1 : workflow_name OR NULL: only workflow_names of the given value will match.
+	//                             all will match on NULL
+	// $2 : instance_name OR NULL: only instance_names of the given value will match.
+	//                             all will match on NULL
+	// $3 : is_running OR NULL: if true, only running instances will match. if false,
+	//                          only completed instances will match. if NULL, all will
+	//                          match.
 	listWorkflowInstancesQuery = `
+-- Find at most one of each (instance_name, workflow_name) from results,
+-- picking at the most recent result if multiple exist
 WITH res AS (
 	SELECT DISTINCT ON (instance_name, workflow_name) *
     FROM cereal_workflow_results
 	ORDER BY instance_name, workflow_name, end_at DESC
 )
+-- Select the workflow instance data, giving priority to that coming
+-- from the cereal_workflow_instances table
 SELECT COALESCE(a.id, res.id) id, 
        COALESCE(a.status, 'completed') status,
        COALESCE(a.instance_name, res.instance_name) instance_name,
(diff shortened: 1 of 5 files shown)


lib/cereal/postgres/postgres.go
```diff
@@ -497,6 +520,52 @@ func (pg *PostgresBackend) GetWorkflowInstanceByName(ctx context.Context, instan
...
+
+		if workflowInstance.Status == backend.WorkflowInstanceStatusCompleted {
+			workflowInstance.IsRunning = false
+			workflowInstance.Payload = nil
+			if errText.Valid {
+				workflowInstance.Err = errors.New(errText.String)
+			}
+		} else {
+			workflowInstance.IsRunning = true
+			workflowInstance.Result = nil
```
jaym: we need to set Err = ""

lib/cereal/postgres/postgres.go
```diff
@@ -29,6 +29,29 @@ const (
...
+	ORDER BY instance_name, workflow_name, end_at DESC
+)
+SELECT COALESCE(a.id, res.id) id, 
+       COALESCE(a.status, 'completed') status,
+       COALESCE(a.instance_name, res.instance_name) instance_name,
+	   COALESCE(a.workflow_name, res.workflow_name) workflow_name,
+	   COALESCE(a.parameters, res.parameters) parameters,
+	   a.payload payload,
+	   res.result result,
+	   res.error error
```
ryancragun: This is almost a SQL koan or something. Writing a query that joins two tables and optionally supports three different where options! There is a dragon here where we could join two workflows that share the same workflow name and instance name but we match the result of a previous instance to the currently running instance! That's right, a running instance that's already completed!

You handle it when you're scanning through the rows which is dope, but I think we should write a comment about that side effect of this query so future us (me) isn't confused about it.

tools/cereal-scaffold/main.go
```diff
@@ -411,3 +430,40 @@ func runScheduleTest(_ *cobra.Command, args []string) error {
...
+
+	if listInstanceOpts.WorkflowName != "" {
+		opts.WorkflowName = &listInstanceOpts.WorkflowName
+	}
+
+	instances, err := b.ListWorkflowInstances(context.Background(), opts)
+	if err != nil {
+		return err
+	}
+	spew.Dump(instances)
```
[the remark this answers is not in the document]
jaym: this is all debug code

PR #1115, opened 2019-07-31
Instrument chef-form-field to watch its input

msorens:
### :nut_and_bolt: Description: What code changed, and why?

Rather than having to manually instrument every input to watch for
its value being edited so that it matches its original value, this bakes that into the component. Here's the details embedded in the code itself:
```
 // A FormFieldComponent can optionally track whether an input field (FormControl)
  // being edited is the same as, or different from, the field's initial value.
  // That is, it knows whether the user has made a change or not,
  // and marks the pristine/dirty state accordingly, so that information may be used
  // in the parent form to e.g., disable the `Save` button, per our UX guidelines.
  // Example:
  // <form [formGroup]="updateForm">
  //   <chef-form-field
  //       [watch]="updateForm.controls.name.value"
  //       [control]="updateForm.controls.name">
  //     <label>Name <span aria-hidden="true">*</span></label>
  //     <chef-input formControlName="name" type="text"></chef-input>
  //   </chef-form-field>
  // </form>
  // <chef-button primary [disabled]="!updateForm.valid || !updateForm.dirty">
  //
  // You need to supply these two parameters:
  @Input() watch: string; //  dynamic value to watch
  @Input() control: FormControl; // control to set back to pristine when needed
```

Note that this functionality is optional. 
Any existing `<chef-form-field>` elements will function exactly as they did before.

### :chains: Related Resources

NA

### :+1: Definition of Done

The new functionality is wired into editing:
- a project name (Settings >> Projects >> `<project>` >> Details) 
- a team name (Settings >> Teams >> `<project>` >> Details) 

### :athletic_shoe: How to Build and Test the Change

Rebuild automate-ui.
Edit a project or team name as indicated above.
Edit the value.
➡️ Observe `Save` becomes enabled.
Restore the original value.
➡️ Observe `Save` becomes disable.

### :white_check_mark: Checklist

- [x] Tests added/updated?
- [x] Docs added/updated?

### :camera: Screenshots, if applicable

PR #1171, opened 2019-08-06
Add `FormControlDirective ` to track original value of form controls

scottopherson:

Adds a `FormControlDirective` that matches any form inputs utilizing
`[formControl]` or `[formControlName]` bindings. This directive keeps a reference to the control's initial value and ensures that a dirty control is marked as pristine if/whenever the value changes back to its initial value.

https://github.com/chef/automate/pull/1115

### :+1: Definition of Done

Form controls are marked as pristine whenever its value is equal to its initial value.

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/components/chef-components.module.ts
+++ b/components/automate-ui/src/app/components/chef-components.module.ts
@@ -14,6 +14,7 @@ import { CalendarComponent } from './calendar/calendar.component';
 import { CreateObjectModalComponent } from './create-object-modal/create-object-modal.component';
 import { DeleteObjectModalComponent } from './delete-object-modal/delete-object-modal.component';
 import { ErrorDirective } from './error/error.directive';
+import { FormControlDirective } from './form-control/form-control.directive';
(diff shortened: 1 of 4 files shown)


scottopherson:
> The one thing that is not covered by this work is those same input elements appearing in a table. 

@msorens what you mean by that exactly? This PR doesn't work for inputs within a table?

msorens:
That's correct @Scott. Here's the one use of it we have:
image
Go to Settings >> Projects
Create a project, Save
Create a rule. Enter a name. Enter a type. Then you will get to the Conditions table. Fill out just the one row is sufficient.
Save.
Then go back into the rule to edit it by clicking on its name.
Changing the items on the condition and then back again does not disable the Save.

Hmm... when I tested it earlier, the name field was working correctly, but just now as I was documenting the steps, even that is not working.

msorens:
OK, I resolved the above issue with the rules page. It is coding on the project-rules page now interfering with this nice, new shiny handler. Once addressed, this PR is working perfectly!

PR #1175, opened 2019-08-06
Run compliance reindex calls async

alexpop:
Switched to async ElasticSearch reindex calls to avoid gateway timeouts on long running reindexes.

### DIFF
--- a/components/compliance-service/reporting/relaxting/migrations.go
+++ b/components/compliance-service/reporting/relaxting/migrations.go
@@ -359,22 +359,22 @@ func (backend ES2Backend) getScanDateRange(indexPrefix string) (*time.Time, *tim
 	return &earliestScanDate, &mostRecentScanDate, nil
 }
 
-func (backend ES2Backend) reindex(src, dest, reindexScript, srcDocType string) (*elastic.BulkIndexByScrollResponse, bool, error) {
+func (backend ES2Backend) reindex(src, dest, reindexScript, srcDocType string) (bool, error) {
 	myName := "reindex"
 
 	client, err := backend.ES2Client()
 	if err != nil {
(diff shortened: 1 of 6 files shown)


components/compliance-service/reporting/relaxting/migrations.go
```diff
@@ -359,22 +359,22 @@ func (backend ES2Backend) getScanDateRange(indexPrefix string) (*time.Time, *tim
 	return &earliestScanDate, &mostRecentScanDate, nil
 }
 
-func (backend ES2Backend) reindex(src, dest, reindexScript, srcDocType string) (*elastic.BulkIndexByScrollResponse, bool, error) {
```
lancewf: We found that the `elastic.BulkIndexByScrollResponse` was not being used and we do not have access to it when starting the reindex with `DoAsync`

components/compliance-service/reporting/relaxting/util.go
```diff
@@ -73,6 +73,22 @@ func (backend *ES2Backend) ES2Client() (*elastic.Client, error) {
...
+	tasksGetTaskResponse, err := elastic.NewTasksGetTaskService(esClient).
+		TaskId(taskID).
+		WaitForCompletion(false).
+		Do(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	if tasksGetTaskResponse.Task == nil {
+		return false, fmt.Errorf("ReindexStatus: task %s not found", taskID)
```
lancewf: If the Task from the `taskID` is not found in Elasticsearch, we are throwing an error. This could happen if Elasticsearch is restarted.

lancewf: We should not be throwing an error here. I have found for some Elasticsearch instances that the `tasksGetTaskResponse.Task` is `nil`, but this does not mean there is an error.

PR #1178, opened 2019-08-06
include passed, skipped, failed counts on nodes response

vjeffrey:
In order to implement status filter buttons on the compliance reporting nodes list, we need to return the totals passed skipped and failed nodes in the api response. This code change makes a call to a function we already had in our code (which collects the node status counts for the overview page) and then adds those numbers to the response on the api.

https://github.com/chef/automate/pull/1161

### :+1: Definition of Done
total passed, skipped, and failed nodes returned on the api response for `compliance/reporting/nodes`

rebuild components/compliance-service
load more than 100 nodes (chef_load_compliance_scans)
check the api response

- [x] Tests added/updated?
- [ ] Docs added/updated?

Screen Shot 2019-08-06 at 15 17 28


components/compliance-service/habitat/default.toml
```diff
@@ -22,4 +22,4 @@ manual_polling_interval = 1440
 [agent]
 workers = 10
 buffer_size = 1000
-remote_inspec_version = "REPLACE-FROM-PLAN.SH"
+remote_inspec_version = "4.7.24"
```
vjeffrey: i'll take this out of the pr ...also we need to figure out a way to not always run into this problem

vjeffrey:
looks like i still have a few more tests that need updating. ill come back to this later this evening to update those tests

PR #1182, opened 2019-08-06
[automate-809] Projects for Team Creation and List View

tylercloke:

* Added optional projects to `app-create-object-modal`
* Wired projects into the Team List view
  * On team create, when in IAM v2.1 mode, it'll show the full list of projects you've selected in your project filter to add to the team. If none are selected, it'll default to `(unassigned)`.
  * If `(unassigned)` is the only thing selected in the projects filter, then the projects dropdown will be grayed out and you will always create with `(unassigned)`.
* Added empty list case to Team List view (since projects can filter all teams)
* Added the project list when in IAM v2.1 mode to the Team Details page. This was mostly done so I could assert something easily in testing. The rest of the project related changes for Team Details will land in a follow up PR.

Get into IAM v2.1 mode and create a few projects.

```
start_all_services
curl -kH "api-token: $TOK" -XPOST -d "$(jq -n '{ id: "project1", name: "my project 1" }')" https://localhost/apis/iam/v2beta/projects
curl -kH "api-token: $TOK" -XPOST -d "$(jq -n '{ id: "project2", name: "my project 2" }')" https://localhost/apis/iam/v2beta/projects
```

Navigate to `/settings/teams` and create teams with one project, no projects (unassigned), many projects, etc.

Downgrade to IAM v2.0 or v1.0 and don't see any of the updates.

- [x] Tests added/updated?
- [ ] Docs added/updated?

Screen Shot 2019-08-06 at 11 53 43 AM
Screen Shot 2019-08-06 at 12 24 18 PM
Screen Shot 2019-08-06 at 2 50 56 PM
Screen Shot 2019-08-06 at 4 36 41 PM
Screen Shot 2019-08-06 at 4 36 51 PM

### DIFF
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
@@ -6,7 +6,7 @@
           <chef-form-field>
             <label>
               <span class="label">Name <span aria-hidden="true">*</span></span>
-              <input chefInput formControlName="name" type="text" (keyup)="handleNameInput($event)" id="name-input">
+              <input chefInput formControlName="name" type="text" (keyup)="handleNameInput($event)" id="name-input" data-cy="create-name">
             </label>
             <chef-error *ngIf="(createForm.get('name').hasError('required') || createForm.get('name').hasError('pattern')) && createForm.get('name').dirty">
               Name is required.
@@ -18,7 +18,7 @@
           <chef-form-field>
             <label>
               <span class="label">ID <span aria-hidden="true">*</span></span>
-              <input chefInput formControlName="id" type="text" (keyup)="handleIDInput($event)" id="id-input">
+              <input chefInput formControlName="id" type="text" (keyup)="handleIDInput($event)" id="id-input" data-cy="create-id">
             </label>
             <chef-error *ngIf="createForm?.controls['id'].hasError('maxlength')">
               ID must be 64 characters or less.
@@ -38,10 +38,10 @@
         <div *ngIf="!modifyID" class="margin">
           <div id="id-fields">
             <span class="key-label">ID:&nbsp;</span>
-            <span>{{ this.createForm?.value.id }}</span>
+            <span data-cy="id-label">{{ this.createForm?.value.id }}</span>
           </div>
           <chef-toolbar>
-            <chef-button tertiary (click)="modifyID = true" id="edit-button-object-modal">Edit ID</chef-button>
+            <chef-button tertiary (click)="modifyID = true" id="edit-button-object-modal" data-cy="edit-button">Edit ID</chef-button>
           </chef-toolbar>
         </div>
         <div *ngIf="showProjectsDropdown">
@@ -52,7 +52,7 @@
           </app-projects-dropdown>
         </div>
         <div id="button-bar">
-          <chef-button primary id="create-button-object-modal"
+          <chef-button primary id="create-button-object-modal" data-cy="save-button"
             [disabled]="!createForm?.valid || creating || conflictError" (click)="createObject()">
             <chef-loading-spinner *ngIf="creating"></chef-loading-spinner>
             <span *ngIf="!creating">Create {{ objectNoun | titlecase }}</span>
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
@@ -44,6 +44,13 @@
             <chef-button tertiary (click)="modifyID = true" id="edit-button-object-modal">Edit ID</chef-button>
           </chef-toolbar>
         </div>
+        <div *ngIf="showProjectsDropdown">
+          <app-projects-dropdown
+            [projects]="projectsAssignable"
+            (onSelection)="projectsDropdownAction($event)"
+            [disabled]="projectsAssignable.length === 0">
+          </app-projects-dropdown>
+        </div>
         <div id="button-bar">
           <chef-button primary id="create-button-object-modal"
             [disabled]="!createForm?.valid || creating || conflictError" (click)="createObject()">
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
@@ -46,9 +46,9 @@
         </div>
         <div *ngIf="showProjectsDropdown">
           <app-projects-dropdown
-            [projects]="projectsAssignable"
+            [projects]="assignableProjects"
             (onSelection)="projectsDropdownAction($event)"
-            [disabled]="projectsAssignable.length === 0">
+            [disabled]="assignableProjects.length === 0">
           </app-projects-dropdown>
         </div>
         <div id="button-bar">
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
@@ -1,6 +1,7 @@
 import { Component, EventEmitter, Input, Output, OnInit } from '@angular/core';
 import { FormGroup } from '@angular/forms';
 import { IdMapper } from 'app/helpers/auth/id-mapper';
+import { Project } from 'app/entities/projects/project.model';
 
 @Component({
   selector: 'app-create-object-modal',
@@ -11,15 +12,18 @@ export class CreateObjectModalComponent implements OnInit {
   @Input() visible = false;
   @Input() creating = false;
   @Input() objectNoun: string;
+  @Input() showProjectsDropdown = false;
+  @Input() projectsAssignable: Project[] = [];
   @Input() createForm: FormGroup; // NB: The form must contain 'name' and 'id' fields
   @Input() conflictErrorEvent: EventEmitter<boolean>; // TC: This element assumes 'id' is the
                                                       // only create field that can conflict.
   @Output() close = new EventEmitter();
-  @Output() createClicked = new EventEmitter();
+  @Output() createClicked = new EventEmitter<Project[]>();
 
   // Whether the edit ID form is open or not.
   public modifyID = false;
   public conflictError = false;
+  private projectsSelected: Project[] = [];
 
   ngOnInit(): void {
     this.conflictErrorEvent.subscribe((isConflict: boolean) => {
@@ -50,10 +54,14 @@ export class CreateObjectModalComponent implements OnInit {
   }
 
   createObject(): void {
-    this.createClicked.emit();
+    this.createClicked.emit(this.projectsSelected);
   }
 
   private isNavigationKey(event: KeyboardEvent): boolean {
     return event.key === 'Shift' || event.key === 'Tab';
+  }
+
+  projectsDropdownAction(projects) {
+    this.projectsSelected = projects;
   }
 }
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
@@ -13,7 +13,7 @@ export class CreateObjectModalComponent implements OnInit {
   @Input() creating = false;
   @Input() objectNoun: string;
   @Input() showProjectsDropdown = false;
-  @Input() projectsAssignable: Project[] = [];
+  @Input() assignableProjects: Project[] = [];
   @Input() createForm: FormGroup; // NB: The form must contain 'name' and 'id' fields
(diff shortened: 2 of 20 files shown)


components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
```diff
@@ -50,10 +54,14 @@ export class CreateObjectModalComponent implements OnInit {
...
   createObject(): void {
-    this.createClicked.emit();
+    this.createClicked.emit(this.projectsSelected);
   }
 
   private isNavigationKey(event: KeyboardEvent): boolean {
     return event.key === 'Shift' || event.key === 'Tab';
   }
+
+  projectsDropdownAction(projects) {
```
msorens: I was going to say this should be typed like so...
```suggestion
  projectsDropdownAction(projects: Project[]): void {
```
But what you have here are _not_ full `Project` objects--they lack a type field.
But you then proceed (L65) to treat them as full Project objects in any case.
So please check that the type field is correctly getting filled in somewhere, too.

components/automate-ui/src/app/pages/team-management/team-management.component.ts
```diff
@@ -83,10 +91,27 @@ export class TeamManagementComponent implements OnInit, OnDestroy {
...
+          this.isMinorV1 = minorVersion === 'v1';
+        });
+
+    this.store.select(projectsAssignable)
+      .subscribe((assignable: ProjectsFilterOption[]) => {
+        this.dropdownProjects = assignable.map(p => {
+          return <Project>{
+            id: p.value,
+            name: p.label
+          };
```
msorens: Essentially same point I mentioned earlier. To be a `Project` object, it must also have a type field. But, of course, the UI does not need to care about that field. So is it worth having a `RenderedProject` object that is a subset without the type? 🤷‍♂ I am on the fence.

components/automate-ui/src/app/services/projects-filter/projects-filter.selectors.ts
```diff
@@ -1,9 +1,27 @@
 import { createSelector } from '@ngrx/store';
 
+import { ProjectConstants } from 'app/entities/projects/project.model';
+
 export const projectsFilterState = state => state.projectsFilter;
 
 export const options = createSelector(projectsFilterState, state => state.options);
 
+export const projectsAssignable = createSelector(projectsFilterState, state => {
+  let projectOptions = state.options.filter(p => p.checked);
```
msorens: Do we know the type of `p` to add in here...?

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
```diff
@@ -44,6 +44,13 @@ <h2 slot="title">Create {{ objectNoun | titlecase }}</h2>
             <chef-button tertiary (click)="modifyID = true" id="edit-button-object-modal">Edit ID</chef-button>
           </chef-toolbar>
         </div>
+        <div *ngIf="showProjectsDropdown">
+          <app-projects-dropdown
+            [projects]="projectsAssignable"
```
bcmdarroch: [nit] s/projectsAssignable/assignableProjects (sounds a bit more natural)

tylercloke: 🤖🤖🤖🤖🤖I'll change that in my follow up PR so I don't have to wait for CI again!

components/automate-ui/src/app/pages/team-management/team-management.component.ts
bcmdarroch: Seems more straightforward to treat it as a full project, if we just have to pass around one extra field. Though if we're not getting `type` back from the API that would change my answer 🤔

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
```diff
@@ -6,7 +6,7 @@ <h2 slot="title">Create {{ objectNoun | titlecase }}</h2>
           <chef-form-field>
             <label>
               <span class="label">Name <span aria-hidden="true">*</span></span>
-              <input chefInput formControlName="name" type="text" (keyup)="handleNameInput($event)" id="name-input">
+              <input chefInput formControlName="name" type="text" (keyup)="handleNameInput($event)" id="name-input" data-cy="create-name">
```
bcmdarroch: this (and other `data-cy` additions) can cause problems in the acceptance cypress tests, since this attribute won't be available in acceptance until this changes gets promoted.

components/automate-ui/src/app/pages/team-management/team-management.component.spec.ts
```diff
@@ -196,7 +197,7 @@ describe('TeamManagementComponent', () => {
       component.openCreateModal();
       component.createTeamForm.controls['name'].setValue(team.name);
       component.createTeamForm.controls['id'].setValue(team.id);
-      component.createV2Team();
+      component.createV2Team([]);
```
bcmdarroch: why does this need an empty array?

e2e/cypress/integration/ui/settings/02_team_mgmt.spec.ts
```diff
@@ -1,25 +1,47 @@
...
+  const project2ID = `${cypressPrefix}-project2-${now}`;
+  const project2Name = `${cypressPrefix} project2 ${now}`;
+  const generatedTeamID = teamName.split(' ').join('-');
+  const unassigned = '(unassigned)';
+
+  let iamVersion = <string><Object>Cypress.env('IAM_VERSION');
+  // assume 2.1 if not in CI. if you wish something different start cypress with
+  // IAM_VERSION set to what you are testing.
+  if (iamVersion === undefined) {
+    iamVersion = 'v2.1';
```
bcmdarroch: I'm not sure we should assume `v2.1` if run locally. I've been trying to figure out a way to have the test check for us if the environment variable isn't set (unfortunately not having much luck). So in the meantime it probably makes more sense to set this to `v2`, since that's most likely what a local environment will be set to

e2e/cypress/integration/ui/settings/02_team_mgmt.spec.ts
```diff
@@ -41,11 +63,251 @@ describe('team management', () => {
...
+      cy.contains(customTeamID).should('exist');
+      cy.contains(generatedTeamID).should('not.exist');
+      cy.go('back');
+    });
+  });
+
+  describeProjectsEnabled('team create modal with projects (IAM v2.1 only)', () => {
+    const unusedProject = 'cypress-unused';
+    const STORE_OPTIONS_KEY = 'projectsFilter.options';
+    const dropdownNameUntilEllipsisLen = 25;
```
bcmdarroch: so specific 😅

e2e/cypress/integration/ui/settings/02_team_mgmt.spec.ts
```diff
@@ -41,11 +63,251 @@ describe('team management', () => {
...
+        });
+
+        localStorage.setItem(STORE_OPTIONS_KEY, JSON.stringify([
+          {label: unassigned, value: unassigned, checked: true},
+          {label: project1Name, value: project1ID, checked: true},
+          {label: project2Name, value: project2ID, checked: true},
+          {label: unusedProject, value: unusedProject, checked: false}
+        ]));
+
+        // relog to apply localstorage state
```
bcmdarroch: why do we have to relog here? can you not set the state after logging in?

tylercloke: We just need to update localstorage before it loads the page, so it gets applied. If we update after the teams page loads it'll be too late.

components/automate-ui/src/app/pages/team-management/team-management.component.spec.ts
tylercloke: It's just the array of projects. Nothing else in here is testing project specific stuff so I left it empty.

e2e/cypress/integration/ui/settings/02_team_mgmt.spec.ts
tylercloke: Fine by me! Will update.

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
tylercloke: Fixed!

e2e/README.md
```diff
@@ -55,6 +55,17 @@ npm run cypress:run
 You can test against dev/acceptance environments as well. Just be
 sure to leave off the trailing `/`.
 
+### IAM Version
+
+Some tests will conditionally run in CI depending on what `CYPRESS_IAM_VERSION` is set to.
+To target those specific sets of tests set that environment variable when starting cypress:
+
+```bash
+CYPRESS_IAM_VERSION="v2.1" npm run cypress:run
```
bcmdarroch: one more thing, I'd include the other possible values for IAM_VERSION
`export CYPRESS_IAM_VERSION= v1 | v2 | v2.1`

and maybe a more emphatic `Your dev environment's IAM version MUST match the value of CYPRESS_IAM_VERSION for the tests to pass locally`

e2e/cypress/integration/ui/settings/02_team_mgmt.spec.ts
```diff
@@ -21,85 +21,106 @@ describe('team management', () => {
...
+        .invoke('attr', 'major-version')
+        .then((obj: Cypress.ObjectLike) => {
+          switch (<string><Object>obj) {
+            case 'v2': {
+              cy.get('#table-container chef-th').contains('ID');
+              cy.get('#table-container chef-th').contains('Name');
+
+              const systemTeams = ['admins', 'editors', 'viewers'];
+              systemTeams.forEach(name => {
+              cy.get('#table-container chef-tr').contains(name)
```
srenatus: If we do this we shouldn't need the `.parent().parent()`:
```typescript
    cy.get('#table-container').contains('chef-tr', name)
      .find('chef-control-menu')
```
[see here for another one of these](https://github.com/chef/automate/blob/322969e75683aa85f0850ddac5b97121873bdab9/e2e/cypress/integration/ui/settings/03_team_add_users.spec.ts#L112-L113)

...but I'm not sure what the `as('control-menu')` does here 🤔

e2e/cypress/integration/ui/settings/02_team_mgmt.spec.ts
```diff
@@ -109,19 +130,23 @@ describe('team management', () => {
     const dropdownNameUntilEllipsisLen = 25;
 
     context('when only the unassigned project is selected', () => {
-      before(() => {
+      beforeEach(() => {
         localStorage.setItem(STORE_OPTIONS_KEY, JSON.stringify([
```
srenatus: 💭 I know this was here before in some way already; but generally, I think it would be preferrable if the cypress tests knew nothing about localStorage. We want projects to be selected? So we should open the picker and select them. It's probably more work, but it would improve our test robustness a lot. We wouldn't have to redo them if we decide that localStorage is not the right tool or something like that...

e2e/cypress/integration/ui/settings/02_team_mgmt.spec.ts
```diff
@@ -109,19 +130,23 @@ describe('team management', () => {
...
-
-        // relog to apply localstorage state
         cy.adminLogin('/settings/teams').then(() => {
           const admin = JSON.parse(<string>localStorage.getItem('chef-automate-user'));
           adminToken = admin.id_token;
+          cy.cleanupTeamsByDescriptionPrefix(adminToken, cypressPrefix);
+          cy.cleanupProjectsByIDPrefix(adminToken, cypressPrefix);
         });
+        cy.restoreStorage();
+        // Necessary to pick up localstorage changes
```
srenatus: Another hack we wouldn't need if we selected projects by clicking 😉

tylercloke: I'm gonna merge but doing a follow up PR immediately after! I'll roll these into that PR.

e2e/cypress/integration/ui/settings/02_team_mgmt.spec.ts
tylercloke: Heh agreed. Didn't write that one. That's an old test. I'll update it in follow up PR.

e2e/cypress/integration/ui/settings/02_team_mgmt.spec.ts
tylercloke: That does sound better. Will take a crack in follow up PR I'm working on.

e2e/cypress/integration/ui/settings/02_team_mgmt.spec.ts
srenatus: Yup, with the construction I've proposed, we shouldn't have to assume how many calls to `parents()` we need.

PR #1187, opened 2019-08-07
automate-cli: fix IAM version formatting

srenatus:
Before:

    Checking IAM major:V1  policies for admin policy with admins team.

After, for each IAM version:

    Checking IAM v2.0 policies for admin policy with admins team.
    Checking IAM v2.1 policies for admin policy with admins team.
    Checking IAM v1.0 policies for admin policy with admins team.

Using "IAM v1.0" instead of "IAM v1" is in line with the output of

    chef-automate iam version


susanev:
i think the adding of `.0` was an intentional decision, but now i cant remember why...

srenatus:
ok i'll flip this back then

PR #1190, opened 2019-08-07
Suppress password error upon open

msorens:

Recent code changes inadvertently made the error-pop up for the auth password field when creating a user open immediately upon the form opening:

image

This PR fixes that.

Also did a quick drive-by to DRY up some validator code.

### :+1: Definition of Done

Opening the form does not show that pop-up error.

Rebuild automate-ui.
Go to Settings >> Users >> Create User.

### DIFF
--- a/components/automate-ui/src/app/helpers/auth/validator.ts
+++ b/components/automate-ui/src/app/helpers/auth/validator.ts
@@ -1,0 +1,35 @@
+import { FormControl, FormGroup } from '@angular/forms';
+
+// Custom validators for use on Angular Forms.
+export class ChefValidators {
+
+  // Typical use of matchFieldValidator is confirming a doubly-entered password
+  public static matchFieldValidator(otherField: string) {
+    return (control: FormControl): { [key: string]: any } => {
+      if (!control.root || !(<FormGroup>control.root).controls) {
+        return null;
+      }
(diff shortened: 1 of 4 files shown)


components/automate-ui/src/app/helpers/auth/validator.ts
```diff
@@ -0,0 +1,35 @@
+import { FormControl, FormGroup } from '@angular/forms';
```
msorens: Since I was looking at the validators I took this opportunity for a quick cut and paste to DRY up some code. (So no new code here; this file was extracted from the deleted bits of code you'll find below.)

components/automate-ui/src/app/pages/user-management/user-form/user-form.component.html
```diff
@@ -25,7 +25,7 @@
       <span class="label">Password <span aria-hidden="true">*</span></span>
       <input chefInput formControlName="password" type="password">
     </label>
-    <chef-error *ngIf="createUserForm.get('password').hasError('required') || (createUserForm.get('password').hasError('pattern')) && createUserForm.get('password').dirty">
+    <chef-error *ngIf="(createUserForm.get('password').hasError('required') || createUserForm.get('password').hasError('pattern')) && createUserForm.get('password').dirty">
```
msorens: This is the bug fix--logic error introduced by errant parenthesis.

PR #1194, opened 2019-08-08
[automate-854] Subpage routing for projects

msorens:

This adds a bit of functionality so that when you navigate to different tabs within a project, it is reflected in the URL:

image

While I was in there, I went through the other auth pages that already handle fragments (team-details and policy-details) and made all 3 use the same names, same structure, and same code patterns. (So the first commit did the task at hand; all the remaining commits did an orderly cleanup.)

There's a lot of whitespace changes, as I did some indentation updates, too. So probably easier to see the changes by following the commits.

NA

### :+1: Definition of Done
Selecting a tab updates the URL to show the tab as a fragment.
1. Specifying `#rules` goes to the rules tab.
2. Specifying `#details` goes to the details tab.
3. Specifying no fragment goes to the rules tab (the first tab).
4. Specifying any other fragment (i.e. invalid fragment) goes to the rules tab (the first tab).

rebuild automate-ui
Go to Settings >> Projects.
Add a project if necessary.
Select the project.
Select the tabs.

### DIFF
--- a/components/automate-ui/src/app/pages/policy/details/policy-details.component.html
+++ b/components/automate-ui/src/app/pages/policy/details/policy-details.component.html
@@ -24,7 +24,7 @@
       </table>
         <chef-tab-selector (change)="onSelectedTab($event)" [value]="tabValue">
           <chef-option value='definition'>Definition</chef-option>
-          <chef-option value='members' routerLink="./" fragment="members">Members</chef-option>
+          <chef-option value='members'>Members</chef-option>
         </chef-tab-selector>
       </chef-page-header>
     <section class="page-body">
--- a/components/automate-ui/src/app/pages/policy/details/policy-details.component.html
+++ b/components/automate-ui/src/app/pages/policy/details/policy-details.component.html
@@ -22,59 +22,57 @@
           </tr>
         </tbody>
       </table>
-        <chef-tab-selector (change)="onSelectedTab($event)" [value]="tabValue">
-          <chef-option value='definition'>Definition</chef-option>
-          <chef-option value='members'>Members</chef-option>
-        </chef-tab-selector>
-      </chef-page-header>
-    <section class="page-body">
-      <ng-container *ngIf="tabValue === 'definition'">
-        <chef-toolbar>
-          <chef-clipboard value={{policyJSON}} label="Copy Definition" icon=""></chef-clipboard>
-        </chef-toolbar>
-        <section id="json-bg">
-          <chef-snippet [code]="policy | json" lang="json"></chef-snippet>
-        </section>
-      </ng-container>
-      <ng-container *ngIf="tabValue === 'members'">
-        <div *ngIf="(members$ | async)?.length === 0" class="empty-state">
-          <p>Add some members to get started!</p>
+      <chef-tab-selector (change)="onSelectedTab($event)" [value]="tabValue">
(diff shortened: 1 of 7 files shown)


components/automate-ui/src/app/pages/policy/details/policy-details.component.html
```diff
@@ -22,59 +22,57 @@
...
+    <section class="page-body" *ngIf="tabValue === 'definition'">
+      <chef-toolbar>
+        <chef-clipboard value={{policyJSON}} label="Copy Definition" icon=""></chef-clipboard>
+      </chef-toolbar>
+      <section id="json-bg">
+        <chef-snippet [code]="policy | json" lang="json"></chef-snippet>
+      </section>
+    </section>
+    <section class="page-body" *ngIf="tabValue === 'members'">
+      <div *ngIf="(members$ | async)?.length === 0" class="empty-state">
```
srenatus: ❓ does this _need_ the `?`?

components/automate-ui/src/app/pages/policy/details/policy-details.component.html
```diff
@@ -22,59 +22,57 @@
...
-          </chef-tbody>
-        </chef-table>
-      </ng-container>
+      </chef-toolbar>
+      <chef-table *ngIf="(members$ | async)?.length > 0">
+        <chef-thead>
+          <chef-tr>
+            <chef-th>ID</chef-th>
+            <chef-th>Type</chef-th>
+            <chef-th></chef-th>
```
srenatus: Hm. Odd. Is this really the way to go? (I'm aware that my previous comment (re: `?`) and this one relate to copied/moved code, but still...)

components/automate-ui/src/app/pages/project/details/project-details.component.html
```diff
@@ -20,12 +20,12 @@
           </tr>
         </tbody>
       </table>
-      <chef-tab-selector [value]="selectedTab" (change)="onTabChange($event)">
+      <chef-tab-selector [value]="tabValue" (change)="onSelectedTab($event)">
```
blakestier: I think we use `onTabChange` in other places as well . . . curious about the name change

components/automate-ui/src/app/pages/policy/details/policy-details.component.html
msorens: It does. Anytime we want to deference a property on an observable brought into the template via an async pipe, that needs to be there because the object will be undefined for some indeterminate period before it resolves.

Tried to find a good reference for async pipes that covers it, but only found a couple that sort of explain it:
https://blog.angular-university.io/angular-reactive-templates/
https://ultimatecourses.com/blog/angular-ngif-async-pipe

components/automate-ui/src/app/pages/policy/details/policy-details.component.html
msorens: Not clear on what you're finding odd--the fact that the column has no heading for accessibility purposes, or something else...?

components/automate-ui/src/app/pages/project/details/project-details.component.html
msorens: On auth pages we use `onSelectedTab`; that is precisely why I changed it.
Also, take a look at my commit message on the commit--I did a google search of how prevalent the two are. 😉

Issue #1185, opened 2019-08-07
Auth form input issues


srenatus:
## Problem
Humans are often frustrated by filling out forms that throw errors before they have been able to provide any information and by the content being hidden on the page.

## Overivew

A few auth input issues to fix, (1) stop presenting an error on the password field before the user has a chance to provide one, (2) stop the browser from showing the auto-complete box and (3) give the user a chance to provide an ID before presenting them with an error.

#### To Reproduce
- Go to Settings > Users
- Click _Create User_
- Observe the form:

![image](https://user-images.githubusercontent.com/870638/62605304-1844af00-b8fa-11e9-8c60-0fe376d5bddb.png)

#### Expected behavior

There's no error below _Password_

#### Details
* Set auto-complete to `off` on inputs
* Reference: [Removing the browser's built-in auto-complete feature](https://github.com/chef/automate/pull/1095)

![auth-complete-fun](https://user-images.githubusercontent.com/5489125/62654616-0de3de80-b915-11e9-9d6b-795f7b4a6691.png)

#### Expected behavior

There is no auto-complete box.

#### To Reproduce
- Navigate to `Settings > Teams > Create Team`, `Settings > API Tokens > Create Token`, or `Settings > Projects > Create Projects`
- Use the _Edit ID_ button
- Observe an error

![id-error](https://user-images.githubusercontent.com/5489125/62654716-3f5caa00-b915-11e9-8174-1fb2474ea45b.png)

#### Expected behavior

There's no error below _ID_

## Subtasks
- [x] Fix password error #1190 @msorens 
- [ ] Turn off auto-complete *RFR* #1195
- [ ] Fix ID error *RFR* #1195

## Notes
ℹ️ Both in `dev` and `acceptance` as of 08/06.

susanev:
@srenatus i hijacked this issue and added a few more issues to it, hope you don't mind ✌️

PR #1195, opened 2019-08-08
[issue 1185] autocomplete="off" and dirty check in id input

srenatus:

1. set `autocomplete="off"` on some more inputs
2. ensure that clicking "edit ID" before having entered a name doesn't make the error pop up

(See the linked issue for details.)

#### ➕ wrapped into this PR

image
image

☝️ Fixed the uniqueness error messages, as discussed with @susanev in slack.

- #1185
- a little overlap with #1156, but not too much

### :+1: Definition of Done

NA

💁‍♂ See #1185's repro steps please.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
@@ -6,7 +6,7 @@
           <chef-form-field>
             <label>
               <span class="label">Name <span aria-hidden="true">*</span></span>
-              <input chefInput formControlName="name" type="text" (keyup)="handleNameInput($event)" id="name-input" data-cy="create-name">
+              <input chefInput formControlName="name" type="text" (keyup)="handleNameInput($event)" id="name-input" data-cy="create-name" autocomplete="off">
             </label>
             <chef-error *ngIf="(createForm.get('name').hasError('required') || createForm.get('name').hasError('pattern')) && createForm.get('name').dirty">
               Name is required.
@@ -18,7 +18,7 @@
           <chef-form-field>
             <label>
               <span class="label">ID <span aria-hidden="true">*</span></span>
-              <input chefInput formControlName="id" type="text" (keyup)="handleIDInput($event)" id="id-input" data-cy="create-id">
+              <input chefInput formControlName="id" type="text" (keyup)="handleIDInput($event)" id="id-input" data-cy="create-id" autocomplete="off"/>
             </label>
             <chef-error *ngIf="createForm?.controls['id'].hasError('maxlength')">
               ID must be 64 characters or less.
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
@@ -20,13 +20,13 @@
               <span class="label">ID <span aria-hidden="true">*</span></span>
               <input chefInput formControlName="id" type="text" (keyup)="handleIDInput($event)" id="id-input" data-cy="create-id" autocomplete="off"/>
             </label>
-            <chef-error *ngIf="createForm?.controls['id'].hasError('maxlength')">
+            <chef-error *ngIf="createForm.get('id').hasError('maxlength') && createForm.get('id').dirty">
               ID must be 64 characters or less.
             </chef-error>
-            <chef-error *ngIf="createForm?.controls['id'].hasError('required')">
+            <chef-error *ngIf="createForm.get('id').hasError('required') && createForm.get('id').dirty">
               ID is required.
             </chef-error>
-            <chef-error *ngIf="createForm?.controls['id'].hasError('pattern')">
+            <chef-error *ngIf="createForm.get('id').hasError('pattern') && createForm.get('id').dirty">
               Only lowercase letters, numbers, hyphens, and underscores are allowed.
             </chef-error>
             <chef-error *ngIf="conflictError">
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
@@ -30,7 +30,7 @@
               Only lowercase letters, numbers, hyphens, and underscores are allowed.
             </chef-error>
             <chef-error *ngIf="conflictError">
-              IDs must be unique--{{objectNoun}} with ID "{{createForm?.controls['id'].value}}" already exists.
+              {{ objectNoun | titlecase }} ID "{{createForm.get('id').value}}" already exists.
             </chef-error>
           </chef-form-field>
           <span class="detail">{{ objectNoun | titlecase }} IDs are unique, permanent, and cannot be changed later.</span>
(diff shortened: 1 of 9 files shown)


components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
```diff
@@ -18,19 +18,19 @@ <h2 slot="title">Create {{ objectNoun | titlecase }}</h2>
           <chef-form-field>
             <label>
               <span class="label">ID <span aria-hidden="true">*</span></span>
-              <input chefInput formControlName="id" type="text" (keyup)="handleIDInput($event)" id="id-input">
+              <input chefInput formControlName="id" type="text" (keyup)="handleIDInput($event)" id="id-input" autocomplete="off"/>
```
bcmdarroch: @susanev are there any inputs where we want `autocomplete` on? we could set this by default on the `chef-input` component. (and then use `<chef-input>` here instead of `<input chefInput>`)

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
```diff
@@ -18,19 +18,19 @@ <h2 slot="title">Create {{ objectNoun | titlecase }}</h2>
...
             </label>
-            <chef-error *ngIf="createForm?.controls['id'].hasError('maxlength')">
+            <chef-error *ngIf="createForm.get('id').hasError('maxlength') && createForm.get('id').dirty">
               ID must be 64 characters or less.
             </chef-error>
-            <chef-error *ngIf="createForm?.controls['id'].hasError('required')">
+            <chef-error *ngIf="createForm.get('id').hasError('required') && createForm.get('id').dirty">
```
bcmdarroch: did you change these to get rid of the `?` ?

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
susanev: none that i am aware of.

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
srenatus: The added dirty is the main part, the rest is just making things a bit more consistent.

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
bcmdarroch: @srenatus in that case, if you want to go a bit further, you could update [chef-input](https://github.com/chef/automate/blob/master/components/chef-ui-library/src/atoms/chef-input/chef-input.tsx) to set `auto-complete=off` (then turn these into chef-inputs, since i'm not sure why they're using the `chefInput` directive in the first place)

srenatus: Oh my, that is spot on. Let's do that instead :) :+1:

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
msorens: Good eye for detail @bcmdarroch !

In this case, it is OK to delete the safe navigation operator (https://angular.io/guide/template-syntax#the-safe-navigation-operator----and-null-property-paths) because `createForm`, if you trace it back to the consumers, is instantiated in the class constructor, so it is guaranteed to exist when the template is being processed.

components/chef-ui-library/src/atoms/chef-input/chef-input.tsx
```diff
@@ -95,7 +97,8 @@ export class ChefInput {
         placeholder={ this.placeholder }
         onChange={ this.handleChange.bind(this) }
         disabled={this.disabled}
-        aria-disabled={this.disabled}></input>
+        aria-disabled={this.disabled}
+        autocomplete="off"/>
```
msorens: It is a logical assumption to assume that `<foo></foo>` and `<foo/>` are equivalent. I know that in earlier HTML versions, some elements did not behave quite correctly if they were self-closing. From what I've read on HTML 5, though, that seems to no longer be an issue, so making this self-closing should work OK. Can anyone else confirm?

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
```diff
@@ -6,7 +6,7 @@ <h2 slot="title">Create {{ objectNoun | titlecase }}</h2>
           <chef-form-field>
             <label>
               <span class="label">Name <span aria-hidden="true">*</span></span>
-              <input chefInput formControlName="name" type="text" (keyup)="handleNameInput($event)" id="name-input" data-cy="create-name">
+              <input chefInput formControlName="name" type="text" (keyup)="handleNameInput($event)" id="name-input" data-cy="create-name" autocomplete="off">
```
msorens: With the change to the library component all of these explicit additions still need to be removed, right...?

components/chef-ui-library/src/atoms/chef-input/chef-input.tsx
srenatus: We've had self-closing inputs in `automate-ui` in a few places, so I've figured this would be OK.

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
srenatus: As far as I've understood @scottopherson, `<chef-input>` and `<input chefInput>` are not the same thing. @scottopherson can you chime in, please?

components/chef-ui-library/src/atoms/chef-input/chef-input.tsx
scottopherson: Elements like `input`, `img`, `link`, etc have always been fine to be self-closing, even before html5. However I believe html5 made the forward slash optional: `` vs ``.

scottopherson: I'd add `@Prop() autocomplete = 'off'` so that the value can be configured from the user of `chef-input`.

```
@Prop() autocomplete = 'off';
       
    ...
    aria-disabled={this.disabled}
    autocomplete={this.autocomplete} />
```

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
msorens: Sorry, was not more clear: What I meant was these items should be changed to `<chef-input>` so you can then remove the `autocomplete`.

srenatus: Hmm that also wasn't clear to me based on @scottopherson's answer -- **is** `<chef-input>` the clear winner we'll want to use everywhere?

msorens: No, not a clear winner. Probably should find out what is not supported with `<chef-input>` but my understanding is that we should use it if we can use it.

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
srenatus: Done ✔️

components/chef-ui-library/src/atoms/chef-input/chef-input.tsx
srenatus: ✔️ done

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
srenatus: @msorens feel free to take over, but I'd rather get this in quick and do a cleanup refactoring later -- @susanev wanted to get this into the next promotion :)

Issue #752, opened 2019-07-03
"ZD 18087" - Compliance: Sort by Platform does not work for redhat with release numbers


vinay033:
# User Story
When trying to sort by the platform in the compliance section, the release numbers for Redhat do not get properly parsed/sorted

# Definition of Done
Sorting by a column will sort correctly

# Demo Script / Repro Steps
have numerous Redhat versions to sort, attempt to sort by platform

tmonk42:
Same problem with sorting Ubuntu versions. Not RedHat specific.

susanev:
@tmonk42 yah, we will be fixing it for all platforms.
in progress: https://github.com/chef/automate/pull/1196

susanev:
@tmonk42 this has merged and should make it into the 09/09 release.

PR #1196, opened 2019-08-08
added changes for sort the platform with name and release version

vinay033:

When trying to sort by the platform in the compliance section, the release numbers of any platform do not get properly parsed/sorted

**Expected Results**: Sorting by a platform column will sort correctly
https://github.com/chef/automate/issues/752
### :+1: Definition of Done
I have added one sorting function which can help to sort the platform with name and release version.
please find the attached video link below
https://chef.zoom.us/recording/share/WTWPt3eJlLCcjQXsqaX-Zg-WSgN39E0lvjcHDVL5COewIumekTziMw?startTime=1565259570000

### DIFF
--- a/components/compliance-service/api/reporting/server/server.go
+++ b/components/compliance-service/api/reporting/server/server.go
@@ -305,11 +305,16 @@ func csvExport(stream reporting.ReportingService_ExportServer) exportHandler {
 
 // ListNodes returns a list of nodes based on query
 func (srv *Server) ListNodes(ctx context.Context, in *reporting.Query) (*reporting.Nodes, error) {
+	formattedFilters := formatFilters(in.Filters)
+	var platform_sort_field = "platform.name.lower"
+	if len(formattedFilters["platform"]) == 1 {
+		platform_sort_field = "platform.release.lower"
+	}
 	var nodes reporting.Nodes
 	var SORT_FIELDS = map[string]string{
 		"name":                                   "node_name.lower",
 		"environment":                            "environment.lower",
-		"platform":                               "platform.name.lower",
+		"platform":                               platform_sort_field,
 		"status":                                 "status",
 		"latest_report.status":                   "status",
 		"latest_report.end_time":                 "end_time",
@@ -321,7 +326,7 @@ func (srv *Server) ListNodes(ctx context.Context, in *reporting.Query) (*reporti
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
-	formattedFilters := formatFilters(in.Filters)
+	
 	formattedFilters, err = filterByProjects(ctx, formattedFilters)
 	if err != nil {
 		return nil, errorutils.FormatErrorMsg(err, "")
--- a/components/compliance-service/api/reporting/server/server.go
+++ b/components/compliance-service/api/reporting/server/server.go
@@ -306,15 +306,15 @@ func csvExport(stream reporting.ReportingService_ExportServer) exportHandler {
 // ListNodes returns a list of nodes based on query
 func (srv *Server) ListNodes(ctx context.Context, in *reporting.Query) (*reporting.Nodes, error) {
 	formattedFilters := formatFilters(in.Filters)
-	var platform_sort_field = "platform.name.lower"
-	if len(formattedFilters["platform"]) == 1 {
-		platform_sort_field = "platform.release.lower"
-	}
+	// var platform_sort_field = "platform.name.lower"
+	// if len(formattedFilters["platform"]) == 1 {
+	// 	platform_sort_field = "platform.release.lower"
+	// }
 	var nodes reporting.Nodes
 	var SORT_FIELDS = map[string]string{
 		"name":                                   "node_name.lower",
 		"environment":                            "environment.lower",
-		"platform":                               platform_sort_field,
+		"platform":                               "platform.full",
 		"status":                                 "status",
 		"latest_report.status":                   "status",
 		"latest_report.end_time":                 "end_time",
--- a/components/compliance-service/api/reporting/server/server.go
+++ b/components/compliance-service/api/reporting/server/server.go
@@ -306,10 +306,6 @@ func csvExport(stream reporting.ReportingService_ExportServer) exportHandler {
 // ListNodes returns a list of nodes based on query
 func (srv *Server) ListNodes(ctx context.Context, in *reporting.Query) (*reporting.Nodes, error) {
 	formattedFilters := formatFilters(in.Filters)
-	// var platform_sort_field = "platform.name.lower"
-	// if len(formattedFilters["platform"]) == 1 {
-	// 	platform_sort_field = "platform.release.lower"
-	// }
 	var nodes reporting.Nodes
 	var SORT_FIELDS = map[string]string{
 		"name":                                   "node_name.lower",
--- a/components/compliance-service/api/reporting/server/server.go
+++ b/components/compliance-service/api/reporting/server/server.go
@@ -326,7 +326,6 @@ func (srv *Server) ListNodes(ctx context.Context, in *reporting.Query) (*reporti
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
-	
 	formattedFilters, err = filterByProjects(ctx, formattedFilters)
 	if err != nil {
 		return nil, errorutils.FormatErrorMsg(err, "")
--- a/components/compliance-service/reporting/relaxting/nodes.go
+++ b/components/compliance-service/reporting/relaxting/nodes.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"sort"
+	"strings"
 
 	"github.com/golang/protobuf/ptypes"
 
@@ -175,11 +177,30 @@ func (backend *ES2Backend) GetNodes(from int32, size int32, filters map[string][
 		if err != nil {
 			return nil, emptyTotals, errors.Wrapf(err, "%s error retrieving node count totals: ", myName)
 		}
+		// sort nodes by the platform name and release
+		if sortField == "platform.name.lower" {
+			sortNodes(nodes, sortField, sortAsc)
+		}
 		return nodes, TotalNodeCounts{Total: int32(searchResult.TotalHits()), Passed: nodeSummary.Compliant, Failed: nodeSummary.Noncompliant, Skipped: nodeSummary.Skipped}, nil
 	}
 
 	logrus.Debugf("%s Found no nodes\n", myName)
 	return nodes, emptyTotals, nil
+}
+
+func sortNodes(nodes []*reportingapi.Node, sortField string, sortAsc bool) {
+	sort.Slice(nodes, func(i, j int) bool {
+		if sortAsc {
+			if strings.Compare(nodes[i].Platform.Name, nodes[j].Platform.Name) == 0 {
+				return nodes[i].Platform.Release < nodes[j].Platform.Release
+			}
+			return nodes[i].Platform.Name < nodes[j].Platform.Name
+		}
+		if strings.Compare(nodes[i].Platform.Name, nodes[j].Platform.Name) == 0 {
+			return nodes[i].Platform.Release > nodes[j].Platform.Release
+		}
+		return nodes[i].Platform.Name > nodes[j].Platform.Name
+	})
 }
 
 func convertToRSControlSummary(summ reporting.NodeControlSummary) *reportingapi.ControlSummary {
--- a/components/compliance-service/reporting/relaxting/nodes.go
+++ b/components/compliance-service/reporting/relaxting/nodes.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"sort"
-	"strings"
 
 	"github.com/golang/protobuf/ptypes"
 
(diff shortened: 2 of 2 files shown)


components/compliance-service/reporting/relaxting/nodes.go
```diff
@@ -174,13 +176,32 @@ func (backend *ES2Backend) GetNodes(from int32, size int32, filters map[string][
...
+			sortNodes(nodes, sortField, sortAsc)
+		}
 		return nodes, TotalNodeCounts{Total: int32(searchResult.TotalHits()), Passed: nodeSummary.Compliant, Failed: nodeSummary.Noncompliant, Skipped: nodeSummary.Skipped}, nil
 	}
 
 	logrus.Debugf("%s Found no nodes\n", myName)
 	return nodes, emptyTotals, nil
 }
 
+func sortNodes(nodes []*reportingapi.Node, sortField string, sortAsc bool) {
```
vjeffrey: would love to get a unit test for this function!

vjeffrey:
👍 on the changes, but would like to get a unit test in the mix

phiggins:
I liked your video, that's a cool way to demo the change. :)

It's not clear to me from the video or the linked issue what exactly is the current incorrect behavior. Is the incorrect behavior that nodes are being sortied by the platform name and not the version, but then the combined name + version is displayed in the UI?

The `GetNodes` function being modified in this PR does sorting and paging already, so with the change in this PR, the user could get some unexpected behavior where the currently viewed page is sorted correctly but going to the next page would show nodes I wouldn't expect to see. 

For example, if my nodes were like this and I was using a page size of 3:

| id | platform name | platform version |
| -- | -- | -- | 
| 1 | ubuntu | 1 |
| 2 | ubuntu | 3 |
| 3 | ubuntu | 2 |
| 4 | ubuntu | 3 |
| 5 | ubuntu | 2 |
| 6 | ubuntu | 1 |

My call to get the first page of nodes with `GetNodes` might return me nodes 1-3, which would then be sorted into order [1, 3, 2], showing me "ubuntu 1", "ubuntu 2", "ubuntu 3". My call to get the second page of nodes with `GetNodes` would return nodes 4-6, which would be sorted into order [6, 5, 4], which would show up in the UI as "ubuntu 1", "ubuntu 2", "ubuntu 3".

The correct thing to do would be for `GetNodes` to sort by the platform name and platform version in elastic, so that in the previous example my request for the first page might return me nodes 1, 3, and 6, sorted into the order [1, 6, 3] so that I would see "ubuntu 1", "ubuntu 1", "ubuntu 2".

There's already an example in `GetNodes` for adding a secondary sort: https://github.com/chef/automate/blob/462bfdca321c8926017ee5f48f46dc3ba927f891/components/compliance-service/reporting/relaxting/nodes.go#L83-L86

I think the correct behavior could be achieved by adding a secondary sort for platform version if the requested sort is platform name.

components/compliance-service/reporting/relaxting/nodes.go
vinay033: sure, I will add the test case.

vjeffrey: actually, now that i read through pete's comments, probably no need for a test here, since as he pointed out this should be done in elasticsearch for pagination concerns

vinay033:
> I liked your video, that's a cool way to demo the change. :)
> 
> It's not clear to me from the video or the linked issue what exactly is the current incorrect behavior. Is the incorrect behavior that nodes are being sortied by the platform name and not the version, but then the combined name + version is displayed in the UI?
> 
> The `GetNodes` function being modified in this PR does sorting and paging already, so with the change in this PR, the user could get some unexpected behavior where the currently viewed page is sorted correctly but going to the next page would show nodes I wouldn't expect to see.
> 
> For example, if my nodes were like this and I was using a page size of 3:
> 
> id	platform name	platform version
> 1	ubuntu	1
> 2	ubuntu	3
> 3	ubuntu	2
> 4	ubuntu	3
> 5	ubuntu	2
> 6	ubuntu	1
> My call to get the first page of nodes with `GetNodes` might return me nodes 1-3, which would then be sorted into order [1, 3, 2], showing me "ubuntu 1", "ubuntu 2", "ubuntu 3". My call to get the second page of nodes with `GetNodes` would return nodes 4-6, which would be sorted into order [6, 5, 4], which would show up in the UI as "ubuntu 1", "ubuntu 2", "ubuntu 3".
> 
> The correct thing to do would be for `GetNodes` to sort by the platform name and platform version in elastic, so that in the previous example my request for the first page might return me nodes 1, 3, and 6, sorted into the order [1, 6, 3] so that I would see "ubuntu 1", "ubuntu 1", "ubuntu 2".
> 
> There's already an example in `GetNodes` for adding a secondary sort:
> 
> https://github.com/chef/automate/blob/462bfdca321c8926017ee5f48f46dc3ba927f891/components/compliance-service/reporting/relaxting/nodes.go#L83-L86
> 
> I think the correct behavior could be achieved by adding a secondary sort for platform version if the requested sort is platform name.

@phiggins @vjeffrey  please check the below link of issue reproduce video
https://chef.zoom.us/recording/play/_52hVslkk7GFNURJv7w8mbqOTVKTtPztGAttgS56ZTeSWV6WUZWy6pKUqMv1_k-n?autoplay=true&startTime=1565612861000&_ga=2.182682982.1369754075.1565612989-168973823.1565612989

if user selecting the particular platform filter like `ubuntu`, `centos` or `debian` and then apply sorting then nothing is happen.

phiggins:
Hi @vinay033,

The problem here is that "platform" means multiple things in this context. Sometimes it means just "platform name" and sometimes it means "platform name + platform version". This screenshot from around 0:38 in the video in your last comment illustrates the problem. 

Screen Shot 2019-08-12 at 11 30 00 AM

In this screenshot, you've selected the filter "platform: centos" and are using the sorting controls on the platform column. You're correct, once you've filtered by platform the sorting controls on the platform column do not appear to do anything because all of the filtered nodes have the platform name "centos" and platform version is not used for sorting.

When the request for nodes comes into the backend, the sort parameter is "platform" and that gets translated into "platform name" here:

https://github.com/chef/automate/blob/462bfdca321c8926017ee5f48f46dc3ba927f891/components/compliance-service/api/reporting/server/server.go#L312

The request to get the nodes is made to elasticsearch passing in that value here:

https://github.com/chef/automate/blob/462bfdca321c8926017ee5f48f46dc3ba927f891/components/compliance-service/reporting/relaxting/nodes.go#L76-L81

The code in this PR is doing the sort for platform version too late, after fetching the node data from elasticsearch. Instead, when the sort parameter is "platform", the code should tell elasticsearch to sort by platform name and platform version, where it currently is only telling elasticsearch to sort by platform name.

vinay033:
> Hi @vinay033,
> 
> The problem here is that "platform" means multiple things in this context. Sometimes it means just "platform name" and sometimes it means "platform name + platform version". This screenshot from around 0:38 in the video in your last comment illustrates the problem.
> 
> Screen Shot 2019-08-12 at 11 30 00 AM
> 
> In this screenshot, you've selected the filter "platform: centos" and are using the sorting controls on the platform column. You're correct, once you've filtered by platform the sorting controls on the platform column do not appear to do anything because all of the filtered nodes have the platform name "centos" and platform version is not used for sorting.
> 
> When the request for nodes comes into the backend, the sort parameter is "platform" and that gets translated into "platform name" here:
> 
> https://github.com/chef/automate/blob/462bfdca321c8926017ee5f48f46dc3ba927f891/components/compliance-service/api/reporting/server/server.go#L312
> 
> The request to get the nodes is made to elasticsearch passing in that value here:
> 
> https://github.com/chef/automate/blob/462bfdca321c8926017ee5f48f46dc3ba927f891/components/compliance-service/reporting/relaxting/nodes.go#L76-L81
> 
> The code in this PR is doing the sort for platform version too late, after fetching the node data from elasticsearch. Instead, when the sort parameter is "platform", the code should tell elasticsearch to sort by platform name and platform version, where it currently is only telling elasticsearch to sort by platform name.

@phiggins I have added changes according to review comment, please have a look.

components/compliance-service/api/reporting/server/server.go
```diff
@@ -305,11 +305,16 @@ func csvExport(stream reporting.ReportingService_ExportServer) exportHandler {
...
+	var platform_sort_field = "platform.name.lower"
+	if len(formattedFilters["platform"]) == 1 {
+		platform_sort_field = "platform.release.lower"
+	}
 	var nodes reporting.Nodes
 	var SORT_FIELDS = map[string]string{
 		"name":                                   "node_name.lower",
 		"environment":                            "environment.lower",
-		"platform":                               "platform.name.lower",
+		"platform":                               platform_sort_field,
```
phiggins: I think the logic above is unnecessary, I think you could get the same effect by just changing this line to

```
"platform":                               "platform.release.lower",
```

vinay033: I was tried the same changes but in case of multiple platforms this changes not working because our requirement is only for in case of single platform, in case of multiple platforms we need to sort according to name only.

phiggins: > in case of multiple platforms we need to sort according to name only.

Why would it be better to sort by platform name instead of platform name and platform version? If multiple platforms are selected, sorting by platform name is just as wrong as in all the other cases.

phiggins: To be clear, I don't think it is correct to ever sort by just platform name. If I have nodes like this:

| id | platform name | platform version |
| -- | -- | -- | 
| 1 | ubuntu | 1 |
| 2 | ubuntu | 2 |
| 3 | centos | 1 |
| 4 | centos | 2 |

The correct order for sorting by platform would result in the order [3,4,1,2] displaying "centos 1", "centos 2", "ubuntu 1", "ubuntu 2". If they are only sorted by platform name, the centos nodes will appear before the ubuntu nodes, but the centos and ubuntu nodes will be sorted in arbitrary order.

vinay033:
```
"platform":                               "platform.release.lower",
```
if I add changes like above and If I do not add any platform filter then sort by behavior is not working properly, if you have any suggestion on that please mentioned here for the same line of code. 
I also tried changes like
```
"platform.name":                          	  "platform.name.lower",
"platform.release":                          	  "platform.release.lower",
```

phiggins: We probably do not want to change the `"platform"` value in the left column, that is coming from the UI and changing it would probably require a more significant update.

We already have this code that adds a secondary sort for node_name: https://github.com/chef/automate/blob/462bfdca321c8926017ee5f48f46dc3ba927f891/components/compliance-service/reporting/relaxting/nodes.go#L83-L86

What if you added a second condition above that like:

```
        // Sorting by platform requires sorting by platform name and version
	if sortField == "platform.name.lower" {
		searchSource.Sort("platform.release.lower", sortAsc)
	}
```

This couples `GetNodes` and `ListNodes` but they are already coupled and this does not seem too bad.

vinay033: I tried with the changes
```
        if sortField == "platform.name.lower" {
		searchSource.Sort("platform.release.lower", sortAsc)
	}
```
but not worked.

phiggins: It didn't work at all, meaning it didn't return the proper results? Or something else? I wonder if our existing test data would be enough to reproduce this in a regression test.

vinay033: hi @phiggins have you tested with the below code changes because of these changes not work as expected. 
```
        if sortField == "platform.name.lower" {
		searchSource.Sort("platform.release.lower", sortAsc)
	}
```
actually, the ZD card requirement was only for the single platform, that's the reason I had did the same changes in A1, can we use these changes for now or if you want to sort multiple platforms same as your above example then we can discuss it more.

susanev: @vinay033 can you link me to the ZD card? i want to make sure it aligns with the ux design direction from @jonong1972.

susanev: found it: https://chefio.atlassian.net/browse/MSYS-828

phiggins: I have not tested this. I haven't spent the time to get setup with all the required data. The problem is that the existing code does not handle sorting by platform properly at all, regardless of whatever special case is applied.

Ideally we would have a test case for this but I know getting test data setup is not straightforward so I don't know how easily doable this is. If it is this hard to demonstrate what the problem is, it is going to be hard to verify this isn't broken in the future.

vinay033:
> @vinay033, met with @jonong1972 today. the ux + product ask is that it sorts by platform name + version as @phiggins has described. if that's technically possible, wed love for that to happen.

sure.

alexpop:
@vinay033 to avoid sorting on two fields(`platform.name` and `platform.release`) in ElasticSearch, you can sort by the field that combines them together (`platform.full`):

Here's the output of a search I've done in elasticsearch:

```json
            {
                    "node_name": "redhat(2)-alpha-nginx(f)-apache(s)-failed",
                    "platform": {
                        "release": "6.11",
                        "name": "redhat",
                        "full": "redhat 6.11"
                    }
            },
            {
                    "node_name": "windows(1)-zeta-apache(s)-skipped",
                    "platform": {
                        "release": "7",
                        "name": "windows",
                        "full": "windows 7"
                    }
            }
```

vinay033:
> @vinay033 to avoid sorting on two fields(`platform.name` and `platform.release`) in ElasticSearch, you can sort by the field that combines them together (`platform.full`):
> 
> Here's the output of a search I've done in elasticsearch:
> 
> ```json
>             {
>                     "node_name": "redhat(2)-alpha-nginx(f)-apache(s)-failed",
>                     "platform": {
>                         "release": "6.11",
>                         "name": "redhat",
>                         "full": "redhat 6.11"
>                     }
>             },
>             {
>                     "node_name": "windows(1)-zeta-apache(s)-skipped",
>                     "platform": {
>                         "release": "7",
>                         "name": "windows",
>                         "full": "windows 7"
>                     }
>             }
> ```

yes, it works, I have updated the changes, please have a look.

components/compliance-service/api/reporting/server/server.go
```diff
@@ -305,11 +305,16 @@ func csvExport(stream reporting.ReportingService_ExportServer) exportHandler {
 
 // ListNodes returns a list of nodes based on query
 func (srv *Server) ListNodes(ctx context.Context, in *reporting.Query) (*reporting.Nodes, error) {
+	formattedFilters := formatFilters(in.Filters)
+	// var platform_sort_field = "platform.name.lower"
+	// if len(formattedFilters["platform"]) == 1 {
+	// 	platform_sort_field = "platform.release.lower"
+	// }
```
phiggins: Please delete these comments.

jonong1972:
Yup, just tested it. Looks good. thanks all.

PR #1200, opened 2019-08-08
Add domain support to cereal service

jaym:
Using the grpc backend now requires specifying a domain to namespace workflows and tasks


components/cereal-service/pkg/server/server.go
```diff
@@ -154,7 +192,10 @@ func (s *CerealService) DequeueWorkflow(req cereal.Cereal_DequeueWorkflowServer)
 	}
 	defer completer.Close() // nolint: errcheck
 
+	domain, workflowName := unnamespace(evt.Instance.WorkflowName)
+	evt.Instance.WorkflowName = workflowName
 	logctx = logctx.WithFields(logrus.Fields{
+		"domain-recv":     domain,
```
ryancragun: [ultra-nit] underscore or s/-recv//

Issue #1042, opened 2019-07-23
Client Runs search bar: filter suggestions content overflow


apriofrost:
## Describe the bug
When a filter with long value appears in the search bar suggestions, the text flows over the edge of its container.

## To Reproduce
Go to "Client Runs" tab, select the "policy revision" filter. A suggestion shows up overflowing the limits of its container.

## Expected behavior
Either increase the size of the suggestions container or wrap the text to a new line.
Set a limitation of the text being displayed, could be 260 characters max.

## Screenshots
![image](https://user-images.githubusercontent.com/9401459/61740050-93c32f80-ad42-11e9-8daf-a9f4de7dc466.png)

## Versions (please complete the following information):
- OS: [Mac OS]
- Browser [Chrome]
- Automate Build Number [20190722200451]

PR #1211, opened 2019-08-11
Automate-1042 Client Runs search bar: filter suggestions content over…

tarablack01:
When a filter with long value appears in the search bar suggestions, the text flows over the edge of its container.

https://github.com/chef/automate/issues/1042

### :+1: Definition of Done
Either increase the size of the suggestions container or wrap the text to a new line.
Set a limitation of the text being displayed, could be 260 characters max.

1. build components/automate-ui && start_all_services && chef_load_nodes
2. Go to https://a2-dev.test/infrastructure/client-runs and inspect

Screen Shot 2019-08-11 at 11 27 25 AM

### DIFF
--- a/components/automate-ui/src/app/page-components/search-bar/search-bar.component.scss
+++ b/components/automate-ui/src/app/page-components/search-bar/search-bar.component.scss
@@ -98,6 +98,8 @@
     border-bottom: 1px solid lightGrey;
 
     &.list-item {
+      overflow-wrap: break-word;
+      
       &.selected {
         color: $white;
         cursor: pointer;
--- a/components/automate-ui/src/app/page-components/search-bar/search-bar.component.scss
+++ b/components/automate-ui/src/app/page-components/search-bar/search-bar.component.scss
@@ -99,7 +99,7 @@
 
     &.list-item {
       overflow-wrap: break-word;
-      
+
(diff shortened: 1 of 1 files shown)


tarablack01:
> @tarablack01 looks like the no results box is showing now as the initial state before the user enters anything.. @apriofrost this isn't what you wanted, right?
> 
> client-runs-filters

@susanev great catch! I had removed the check to style forgot to add it back! Updated now.

apriofrost:
The screenshot looks good to me! Thanks for catching the issue @susanev !

susanev:
there was another issue that came up too, with the empty suggestion showing a box, i think @lancewf hit this issue before.. not sure how it was resolved

empty-suggestion-issue

lancewf:
@susanev I wish I remember how that was fixed before. That problem was fixed for the Client Runs page before. I am not seeing the bug on master. I will test out this branch and see if I can reproduce it.

lancewf:
@susanev I am able to reproduce this bug on master with only the 'error' type. This is not a bug in the UI code. The problem is in the API request. The API is returning `[{"score":1}]` where it should return `[{"score":1, "text": "some suggestion"}]`

I will take a look at this to see if I can easily fix the problem.

PR #1212, opened 2019-08-11
Code tightening (prepping for Angular 8)

msorens:
In an effort to make the upcoming PR for the Angular 8 upgrade smaller, I undertook to clean up many of the deprecation, lint, compiler, and test errors that are due to more stringent checks. So all of the changes herein are compatible with our existing Angular/rxjs/TypeScript versions. Some of them correct errors in the code base that went unreported and, being somewhat innocuous, unnoticed.

Follow the commits if you want to see each type of change in isolation.
However, I have also added pre-review comments on the first occurrence of each category so it is easy to review from the file changes, too.

### :+1: Definition of Done

Compiler, lint, and unit tests all still report no issues.

Rebuild automate-ui.
Run unit tests.
Run lint.

### DIFF
--- a/components/automate-ui/src/app/entities/api-tokens/api-token.effects.ts
+++ b/components/automate-ui/src/app/entities/api-tokens/api-token.effects.ts
@@ -8,9 +8,10 @@ import { identity } from 'lodash/fp';
 
 import { NgrxStateAtom } from 'app/ngrx.reducers';
 import { HttpStatus } from 'app/types/types';
-import { iamMajorVersion } from 'app/entities/policies/policy.selectors';
 import { CreateNotification } from 'app/entities/notifications/notification.actions';
 import { Type } from 'app/entities/notifications/notification.model';
+import { iamMajorVersion } from 'app/entities/policies/policy.selectors';
+import { IAMMajorVersion } from 'app/entities/policies/policy.model';
 import {
   ApiTokenActionTypes,
   GetAllTokens,
@@ -51,10 +52,11 @@ export class ApiTokenEffects {
     this.actions$.pipe(ofType<GetAllTokens>(ApiTokenActionTypes.GET_ALL)),
     this.store$.select(iamMajorVersion).pipe(filter(identity))])
     .pipe(
-      mergeMap(([_action, version]) =>
+      mergeMap(([_action, version]: [GetAllTokens, IAMMajorVersion]) =>
         this.requests.getAll(version).pipe(
           map(resp => new GetAllTokensSuccess(versionizeTokens(resp))),
-          catchError((error: HttpErrorResponse) => of(new GetAllTokensFailure(error))))));
+          catchError((error: HttpErrorResponse) => of(new GetAllTokensFailure(error))))
+      ));
 
   @Effect()
   getAllTokensFailure$ = this.actions$.pipe(
@@ -69,7 +71,7 @@ export class ApiTokenEffects {
     this.actions$.pipe(ofType<GetToken>(ApiTokenActionTypes.GET)),
     this.store$.select(iamMajorVersion).pipe(filter(identity))])
     .pipe(
-      mergeMap(([{ payload: { id } }, version]) =>
+      mergeMap(([{ payload: { id } }, version]: [GetToken, IAMMajorVersion]) =>
         this.requests.get(id, version).pipe(
           map(resp => new GetTokenSuccess(versionizeToken(resp))),
           catchError((error: HttpErrorResponse) => of(new GetTokenFailure(error))))));
@@ -87,7 +89,7 @@ export class ApiTokenEffects {
     this.actions$.pipe(ofType<UpdateToken>(ApiTokenActionTypes.UPDATE)),
     this.store$.select(iamMajorVersion).pipe(filter(identity))])
     .pipe(
-      mergeMap(([{ payload: { token } }, version]) =>
+      mergeMap(([{ payload: { token } }, version]: [UpdateToken, IAMMajorVersion]) =>
         this.requests.update(token, version).pipe(
           map(resp => new UpdateTokenSuccess(versionizeToken(resp))),
           catchError((error: HttpErrorResponse) => of(new UpdateTokenFailure(error))))));
@@ -105,7 +107,7 @@ export class ApiTokenEffects {
     this.actions$.pipe(ofType<CreateToken>(ApiTokenActionTypes.CREATE)),
     this.store$.select(iamMajorVersion).pipe(filter(identity))])
     .pipe(
-      mergeMap(([{ payload: { id, name } }, version]) =>
+      mergeMap(([{ payload: { id, name } }, version]: [CreateToken, IAMMajorVersion]) =>
         this.requests.create(id, name, version).pipe(
           map(resp => new CreateTokenSuccess(versionizeToken(resp))),
           catchError((error: HttpErrorResponse) => of(new CreateTokenFailure(error))))));
@@ -133,7 +135,7 @@ export class ApiTokenEffects {
     this.actions$.pipe(ofType<ToggleTokenActive>(ApiTokenActionTypes.TOGGLE)),
     this.store$.select(iamMajorVersion).pipe(filter(identity))])
     .pipe(
-      mergeMap(([{ payload: { id, active } }, version]) =>
+      mergeMap(([{ payload: { id, active } }, version]: [ToggleTokenActive, IAMMajorVersion]) =>
         this.requests.toggleActive(id, active, version).pipe(
           map(resp => new ToggleTokenActiveSuccess(versionizeToken(resp))),
           catchError((error: HttpErrorResponse) => of(new ToggleTokenActiveFailure(error))))));
@@ -159,7 +161,7 @@ export class ApiTokenEffects {
     this.actions$.pipe(ofType<DeleteToken>(ApiTokenActionTypes.DELETE)),
     this.store$.select(iamMajorVersion).pipe(filter(identity))])
     .pipe(
-      mergeMap(([{ payload }, version]) =>
+      mergeMap(([{ payload }, version]: [DeleteToken, IAMMajorVersion]) =>
         this.requests.delete(payload.id, version).pipe(
           map(() => new DeleteTokenSuccess(payload)),
           catchError((error: HttpErrorResponse) => of(new DeleteTokenFailure(error))))));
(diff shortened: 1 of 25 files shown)


components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.ts
```diff
@@ -212,24 +212,24 @@ export class ProfileOverviewComponent implements OnInit, OnDestroy {
...
 
   onFilesSelected(event) {
-    this.sendFiles(Array.from(event.target.files)).subscribe(
-      fileUploads => { this.fileUploads = fileUploads; },
-      null,
-      () => {
+    this.sendFiles(Array.from(event.target.files)).subscribe({
+      next: fileUploads => { this.fileUploads = fileUploads; },
+      error: null,
+      complete: () => {
```
msorens: Deprecated variant of subscribe:
```
subscribe is deprecated: Use an observer instead of a complete callback
```
That message is misleading; `subscribe` is _not_ wholly deprecated. The developers thought there were too many signatures for its own good (see https://github.com/ReactiveX/rxjs/issues/4159), so instead of using this (with three arguments)...
```
subscribe(next, err, complete) 
```
one needs to pass in a single argument (an "observer") like this:
```
subscribe({next: f1, err: f2, complete: f3})
```
That format also happens to make it simpler if you do not need to supply all three callbacks.

components/automate-ui/src/app/pages/client-runs/client-runs.component.ts
```diff
@@ -312,12 +312,14 @@ export class ClientRunsComponent implements OnInit, OnDestroy {
...
+    combineLatest([
       this.nodeCounts$.pipe(distinctUntilKeyChanged('total')),
       this.numberOfSearchBarFilters$.pipe(distinctUntilChanged())
-    )
+    ])
     .pipe(takeUntil(this.isDestroyed))
-    .subscribe(([nodeCounts, filterCount]) => {
+    .subscribe((values: any[]) => {
+      const nodeCounts: NodeCount = values[0];
+      const filterCount: number = values[1];
```
msorens: This is a slightly peculiar instance of the `combineLatest` change. This is the first occurrence that chains it to `subscribe` instead of `map`. But `subscribe` did not take kindly when I tried the usual fix, which would have been to add types to the destructured elements:
```
subscribe(([nodeCounts, filterCount]: [NodeCount, number])
```
So I found the above workaround was necessary.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -73,13 +73,13 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
...
-      combineLatest(
+      combineLatest([
         this.store.select(getStatus),
         this.store.select(updateStatus),
         this.store.select(getProjectStatus)
-      ).pipe(
+      ]).pipe(
         takeUntil(this.isDestroyed),
-        map(([gStatus, uStatus, gpStatus]: string[]) => {
+        map(([gStatus, uStatus, gpStatus]) => {
```
msorens: These are not technically strings; they are of type `EntityStatus` which is a constrained set of strings. So I could have just done this:
```
map(([gStatus, uStatus, gpStatus]: EntityStatus[]) => {
```
But, all the arguments are the same type (see my earlier note where there were different types), so the types are correctly inferred, and there is thus no need to add an explicit type here; it would be redundant.

components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
```diff
@@ -67,8 +67,8 @@ export class ServiceGroupsComponent implements OnInit, OnDestroy {
   private selectedSortField$: Observable<string>;
   private healthSummary$: Observable<HealthSummary>;
   private currentPage$: Observable<number>;
-  private currentFieldDirection: SortDirection;
-  private currentSortField: string;
+  public currentFieldDirection: SortDirection;
+  public currentSortField: string;
```
msorens: Interesting to note that these were used in some unit tests--even though they were private!--with no complaint in our current environment. But the upgrade will get more particular, requiring this fix.

components/automate-ui/src/app/services/http/http-client-auth.interceptor.spec.ts
```diff
@@ -36,7 +36,7 @@ describe('HttpClientAuthInterceptor', () => {
 
     it('when a 401 response is intercepted logs out the session', done => {
       spyOn(chefSession, 'logout');
-      httpClient.get('/endpoint').subscribe(null, done);
+      httpClient.get('/endpoint').subscribe({ error: done });
```
msorens: Here's a good example showing why the "observer object" is better than separate arguments ("next, error, complete" in that order). This test wants to ignore the `next` callback, so it gave a null. It then wants to provide an `error` callback, because the http call is expected to fail, and that is the jasmine `done`. So in the revised code, we just provide the `error` callback--no guesswork involved.

components/automate-ui/src/app/services/profiles/upload.service.spec.ts
```diff
@@ -8,26 +8,29 @@ class MockChefSessionService {
...
     it('returns the correct object', () => {
-      expect(this.service.getProgress('file', 'percent', 'status', 'response'))
+      expect(service.getProgress('file', 25, 1, 'response'))
         .toEqual({
           'name': 'file',
-          'percent': 'percent',
-          'status': 'status',
+          'percent': 25,
+          'status': 1,
           'response': 'response'
```
msorens: Well, so much for strict typing... these types were plain wrong, yet our current env does not complain.

components/automate-ui/src/app/services/projects-filter/projects-filter.requests.ts
```diff
@@ -3,18 +3,19 @@ import { HttpClient } from '@angular/common/http';
 import { Observable } from 'rxjs';
 
 import { environment as env } from 'environments/environment';
+import { Project } from 'app/entities/projects/project.model';
 
 export interface AuthorizedProjectsResponse {
-  projects: string[];
+  projects: Project[];
```
msorens: Whoa! Another interesting type that should never have passed muster.

components/automate-ui/src/app/entities/roles/role.requests.ts
```diff
@@ -6,7 +6,7 @@ import { environment as env } from 'environments/environment';
 import { Role } from './role.model';
 
 export interface RolesResponse {
-  policies: Role[];
+  roles: Role[];
```
srenatus: haha

srenatus: This makes me think we're using this wrong. Why did this not blow up somewhere?

components/automate-ui/src/app/entities/userperms/userperms.selectors.ts
```diff
@@ -8,7 +8,7 @@ export const userpermsState = createFeatureSelector<PermEntityState>(NGRX_REDUCE
 
 export const permList = createSelector(
   userpermsState,
-  (state) => map((id) => get(['byId', id], state),
+  (state) => map((id: string) => get(['byId', id], state),
```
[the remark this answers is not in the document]
srenatus: ❓ Is there a way to lint for something like this?

components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.ts
srenatus: 🤔 So, I think you mean `error` in the last code block. But from what you say, it's not required to supply something that's not need -- so, I wonder why we'd put `error: null` into line 217.

components/automate-ui/src/app/pages/+compliance/+scanner/containers/nodes-add/nodes-add.component.ts
```diff
@@ -66,11 +66,12 @@ export class NodesAddComponent implements OnInit {
...
 
     // Switch selectable secrets based on selected "backend" value (ssh, winrm)
-    this.secrets$ = combineLatest(
+    this.secrets$ = combineLatest([
       this.fetchSecrets(),
       this.backendValue
-    ).pipe(
-      map(([secrets, backend]) => secrets.filter(s => s.type === backend || s.type === 'sudo'))
+    ]).pipe(
+      map(([secrets, backend]: [any[], string]) =>
```
srenatus: how about
```suggestion
      map(([secrets, backend]: [{ type: string }[], string]) =>
```

seems to be an OK thing to do according to the playground: https://www.typescriptlang.org/play/index.html#code/MYewdgzgLgBBCmwBO8oQFwwN4ygTwAd5NokBLMAcxgF8BtAXRgF4Y6AoGL7XQ4mAEQQIACwG12DIA

components/automate-ui/src/app/pages/client-runs/client-runs.component.ts
srenatus: > But subscribe did not take kindly when I tried the usual fix [...]

What happened? Did it throw you out? 😉

components/automate-ui/src/app/entities/api-tokens/api-token.effects.ts
```diff
@@ -47,14 +48,15 @@ export class ApiTokenEffects {
...
 
   @Effect()
-  getAllTokens$ = combineLatest(
+  getAllTokens$ = combineLatest([
     this.actions$.pipe(ofType<GetAllTokens>(ApiTokenActionTypes.GET_ALL)),
-    this.store$.select(iamMajorVersion).pipe(filter(identity)))
+    this.store$.select(iamMajorVersion).pipe(filter(identity))])
     .pipe(
-      mergeMap(([_action, version]) =>
+      mergeMap(([_action, version]: [GetAllTokens, IAMMajorVersion]) =>
```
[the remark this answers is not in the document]
msorens: Yes, definitely a bummer... Unfortunately, post-upgrade has the same problem here.

components/automate-ui/src/app/entities/userperms/userperms.selectors.ts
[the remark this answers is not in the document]
msorens: Post-upgrade, it shows up either as a lint error or compiler error; I don't recall which.

components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.ts
msorens: Yep, you're correct; that was just mechanical on my part, but don't need the `error: null`.

PR #1214, opened 2019-08-12
Add cereal committed messages

jaym:
Relying on closing streams for completion was turning out to be a little bit magical. The clients now wait for a message that says what they have done has been committed (or not). For tasks, we also no longer log an error every time we successfully complete a task because the stream got closed and we're unsure if we got canceled or we finished


lib/cereal/grpc/grpc.go
```diff
@@ -80,9 +82,30 @@ func (c *workflowCompleter) EnqueueTask(task *backend.Task, opts backend.TaskEnq
 	c.tasks = append(c.tasks, t)
 	return nil
 }
+func (c *workflowCompleter) finish(err error) error {
```
stevendanna:
```suggestion

func (c *workflowCompleter) finish(err error) error {
```

lib/cereal/grpc/grpc.go
```diff
@@ -350,22 +347,24 @@ func (g *GrpcBackend) DequeueTask(ctx context.Context, taskName string) (*backen
...
-				cancel()
-				doneChan <- errors.New("canceled")
-				return
-			}
+		// This goroutine will read the next, and what must be the last
+		// message.
+		var errOut error
+		msg, err := s.Recv()
+		if err != nil {
+			logrus.WithError(err).Debug("received error: canceling task context")
```
stevendanna: I'm a little confused by these debug messages because we always cancel the task context at line 366 regardless of what case we hit.

jaym: fair enough. there was a time when this wasn't true, ill fix up the messages

lib/cereal/grpc/grpc.go
ryancragun: I looked at this for like 15 seconds trying to figure out wtf you changed. Whitespace :man_facepalming:

lib/cereal/grpc/grpc.go
```diff
@@ -80,9 +82,30 @@ func (c *workflowCompleter) EnqueueTask(task *backend.Task, opts backend.TaskEnq
 	c.tasks = append(c.tasks, t)
 	return nil
 }
+func (c *workflowCompleter) finish(err error) error {
+	defer c.s.CloseSend() // nolint: errcheck
```
ryancragun: Curious why we need to close the sender end of the stream twice?

lib/cereal/grpc/grpc.go
```diff
@@ -350,22 +347,24 @@ func (g *GrpcBackend) DequeueTask(ctx context.Context, taskName string) (*backen
...
+			logrus.WithError(err).Debug("received error: canceling task context")
+			errOut = err
+		} else if c := msg.GetCancel(); c != nil {
+			logrus.Debug("received cancel: canceling task context")
+			errOut = context.Canceled
+		} else if c := msg.GetCommitted(); c != nil {
+			logrus.Debug("received comitted")
+			errOut = nil
+		} else {
+			errOut = errUnknownMessage
```
ryancragun: [nit] might be nice to used a typed error that allows you to wrap the error message that was received so that that we don't lose context of what actually happened

lib/cereal/grpc/grpc.go
jaym: all the paths don't explicitly call close send. defering makes sure we always do

lib/cereal/grpc/grpc.go
jaym: there was no error in this branch. we received a proto message we didn't know how to interpret

lib/cereal/grpc/grpc.go
ryancragun: I suppose if that's the worry we could close it before the first err return and wouldn't require the defer to an already closed stream on the happy path.

Issue #1181, opened 2019-08-06
Remove the data-lifecycle-service


ryancragun:
Rather than having an entire service that's sole purpose is making scheduled gRPC calls to various services, we'll remove the data-lifecycle-service and move all scheduled `Purge` jobs to cereal workflows in the services themselves.

Currently there are three services that implement the data lifecycle management interface and rely on the data-lifecycle-service to purge data: 
* ingest-service
* compliance-service
* event-feed

All will need to be migrated to cereal workflows in order to remove the data lifecycle service

### Tasks
- [x] Migrate ingest-service time series purging to cereal workflows.
- [x] Delete data-lifecycle-service
- [x] Update deployment-service to no longer start data-lifecycle-service
- [x] Deprecate data lifecycle config for data-lifecycle-service, ingest-service, event-feed-service and compliance-service
- [x] Migrate the compliance data-lifecycle manageable interface to a cereal workflow.
- [x] Migrate event-feed manageable interface to a cereal workflow.
- [x] Delete data-lifecycle interservice protobufs

Issue #1208, opened 2019-08-10
Add more features to _Data Lifecycle_


susanev:
## Overview
Thanks to the 🥣 work by the deployment team we are able to add the rest of the settings into the _Data Lifecycle_ page without causing any service restarts. Note, this page is currently called _Node Lifecycle_ but will be adjusting to _Data Settings_. If any questions come up with how to connect to the backend please reach out to @danielsdeleo and @ryancragun.

If possible, connect with @susanev for a walkthrough before starting work on this issue, please and thank you.

## Design Details
[UX Designs in Abstract](https://share.goabstract.com/47da49f4-55df-4be0-be66-018952857ae3?mode=design&sha=84902a48f104fe6bb5f4cefa3754cc83a72ccbe0)

### Settings Left-Nav Updates
- [x] add new _General Settings_ heading, please use text-transform for the casing
- [x] move _Node Lifecycle_ up to that section and rename it to _Data Lifecycle_
- [x] move _Notifications_ up below _Data Lifecycle_

### Page Content
- [x] heading: Data Lifecycle
- [x] subheading: Manage the retention of events, service groups, Chef Client runs, compliance reports and scans in Chef Automate.
- [x] other headings on the page are `font-size: 16px; font-weight: 600`; and should be h2s semantically

#### Interaction Patterns
- On _check_ the inputs should become enabled and the defaults should be filled in
- On _uncheck_ the inputs should become disabled with whatever last value they held; `background-color: #E6EBEE;` and `opacity: 0.5;`
- Primary, _Save Changes_ button should be disabled until a change is made to a checkbox or an input it should then become enabled; on action its disabled with the label \<loading spinner\>_Saving..._; after saved the button returns to disabled and _Save Changes_ 
- All number inputs should be restricted to only accept integer values greater than zero
- Checking or unchecking a parent checkbox should have no effect on its visual child

#### Event Feed
- _Remove event feed data after_
    - default: enabled, 30 days
- _Remove Chef Server actions after_
    - default: enabled, 30 days

#### Service Groups
- [x] This section should be completely hidden unless the `EAS Applications` ui feature flag has been turned on
- _When no health check reports have been received for a service in \<n time\>, label the service as disconnected_ 
    - This checkbox is a special case and should be presented as disabled to prevent the user from being able to turn it off, but note that it is possible to turn it off from the api, if that happens it should remain disabled but be shown as an unchecked checkbox with disabled number and time type inputs
    - default: 5 minutes
[...]

PR #1219, opened 2019-08-12
[data-lifecycle] Remove the data-lifecycle-service

ryancragun:
* Removes the `data-lifecycle-service`
* Replace `PurgeManageable` gRPC server with shared `data_lifecycle.PurgeServer` gRPC server and corresponding purge cereal workflows.
* Removes `data_lifecycle` from automate config
* Deprecates retention values in `ingest-service`, `event-feed-service`, and `compliance-service` that are now saved as policies in cereal workflows.

https://github.com/chef/automate/issues/1181

### :+1: Definition of Done
`data-lifecycle-service` has been deleted and all functionality has been moved to cereal workflows.

* Start all services
* `chef-automate config patch` with soon-to-be-removed-or-deprecated config
```toml
[compliance.v1.sys.retention]
compliance_report_days = 33

[event_feed_service.v1.sys.service]
purge_event_feed_after_days = 33

[ingest.v1.sys.service]
purge_converge_history_after_days = 33
purge_actions_after_days = 33

[data_lifecycle.v1.sys.service]
daily_run_at = "12:34:56"
```
* `chef-automate dev stop-converge`
* Rebuild deployment, ingest, event-feed, and compliance
* `chef-automate dev start-converge`

After converge data-lifecycle should be gone. If you attempt to repatch the config as you did before it should fail with deprecation warnings.

You can also check the purge schedules with:
`chef-automate dev grpcurl service-name -- chef.automate.infra.data_lifecycle.api.Purge.Show` replacing the service name with any of: compliance-service, ingest-service, and event-feed-service.

- [x] Migrate ingest-service time series purging to cereal workflows.
- [x] Delete data-lifecycle-service
- [x] Update deployment-service to no longer start data-lifecycle-service
- [x] Deprecate data lifecycle config for data-lifecycle-service, ingest-service, event-feed-service and compliance-service
- [x] Migrate the compliance data-lifecycle manageable interface to a cereal workflow.
- [x] Migrate event-feed manageable interface to a cereal workflow.
- [x] Delete data-lifecycle interservice protobufs
- [x] Tests added/updated?
- [x] Docs added/updated (release note upgrade docs)?


lib/cereal/README.md
```diff
@@ -0,0 +1,23 @@
+# Cereal: a workflow schedule and execution library
```
ryancragun: Whoa, this was a wip that got accidentally checked in

lib/datalifecycle/purge/policy.go
```diff
@@ -0,0 +1,62 @@
...
+	"context"
+	"time"
+
+	es "github.com/chef/automate/api/interservice/es_sidecar"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+// Policy represents a purge workflow purge policy
+type Policy struct {
```
ryancragun: Maybe make this `ElasticsearchPolicy` so we can more easily add `PostgresPolicy`'s for the application-service.

api/config/compliance/config_request.proto
```diff
@@ -27,7 +27,12 @@ message ConfigRequest {
 			Profiles profiles = 4;
 			Storage storage = 5;
 			Logger logger = 6;
-			Retention retention = 7;
+			// NOTE: retention is no longer configurable via config and is now
+			// managed with gRPC endpoints. The field is not reserved so we can
+			// migrate the initial values from config to to the purge cereal
+			// workflows.
+			// Setting retention values is prevented in the Validate() callback.
+			Retention retention = 7; // deprecated
```
jaym: `Retention retention = 7 [deprecated=true];`

components/ingest-service/server/job_scheduler_server.go
```diff
@@ -65,15 +65,68 @@ func (server *JobSchedulerServer) GetStatusJobScheduler(ctx context.Context,
...
-		Jobs:    jobs,
-		Running: true,
-	}, nil
+	jobStatus.Running = true
+	jobStatus.Jobs = jobs
+
+	return jobStatus, nil
 }
 
+// FIXME: runJobNow hardcodes shit that it should not
```
stevendanna: Yikes! This seems like a pretty major bug that we probably want to fix in a separate PR.

stevendanna: Submitted this as a fix here: https://github.com/chef/automate/pull/1223

jaym:
so at a high level, i worry about that config being deprecated. I think in most cases, it wouldn't matter. But i imagine for things like a1 migration, it very much would. Also, I don't think we can ever remove it because the backwards compatibility thing.

jaym:
also, i think until we can set the values in the UI, we should accept them from the config

@susanev made an issue: https://github.com/chef/automate/issues/1208

api/config/compliance/config_request.proto
ryancragun: Cool, as far as I know the field tag has no meaning in the go protoc compiler, but the field option seems like a good option.

moutons:
@jaym what about reading the deprecated config and migrating it to the new config as well as warning about deprecation (would it be possible to require user input confirming understanding of the deprecated config options used)? The settings are already going to be migrated as part of an upgrade, right?

jaym:
@moutons i think this pr will correctly migrate whatever already exists. The change is going to be that it will no longer be settable in the config. There also wont be a UI for it right away. The problem I see with it not being settable in the config is people wont be able to choose a value for things like A1 to A2 migration, and we might end up removing stuff they wanted

lib/cereal/README.md.bak
```diff
@@ -0,0 +1,23 @@
+# Cereal: a workflow schedule and execution library
+
+## Overview
```
ryancragun: remove this wip

.gitignore
```diff
@@ -9,6 +9,10 @@ results/
 # just a particular file called 'src', *not* src directories that exist in some projects!
 src
 !src/
+automate-credentials.toml
```
ryancragun: Maybe don't delete these?

api/config/ingest/config_request.proto
```diff
@@ -28,8 +28,13 @@ message ConfigRequest {
 			message Service {
 				google.protobuf.StringValue host = 1;
 				google.protobuf.Int32Value port = 2 [ (chef.automate.api.port) = { name: "service", default: 10122, protocol: "grpc" } ];
-				google.protobuf.Int32Value purge_converge_history_after_days = 3;
-				google.protobuf.Int32Value purge_actions_after_days = 4;
+				// NOTE: purge actions are no longer configurable via config
+			    // and are now managed with gRPC endpoints. They are not
```
ryancragun: formatting is off

api/interservice/data_lifecycle/purge.proto
```diff
@@ -0,0 +1,62 @@
...
+	rpc Show (ShowRequest) returns (ShowResponse);
+}
+
+message RunRequest { }
+message RunResponse { }
+
+message ConfigureRequest {
+	bool enabled = 1;
+	string recurrence = 2;
+	PolicySettings settings = 3;
```
ryancragun: maybe `PollicyUpdate` instead of setting?

components/automate-deployment/pkg/deployment/deployment.go
```diff
@@ -344,7 +366,7 @@ func (d *Deployment) UpdateExpectedServicesFromManifest() error {
 		}
 	}
 
-	// Remove any removed services
+	// Update the expected services with our new ExpectedServices
```
stevendanna: Seeing this loop, I'm a bit confused about why the above deletion loop is needed.  Because the range here is over the serviceIDs, the new expectedServices array should be free from any removed services without the deletion loop we added above.

lib/product/types.go
```diff
@@ -185,6 +226,7 @@ type Package struct {
 // Metadata is the top level metadata the describes the automate
 // collections and its packages.
 type Metadata struct {
-	Packages    []*Package    `json:"packages"`
-	Collections []*Collection `json:"collections"`
+	Packages        []*Package       `json:"packages"`
+	DeletedPackages []DeletedPackage `json:"deleted_packages"`
```
stevendanna: I'm not seeing where these deleted packages are actually used? Mind pointing me in the right direction?

components/event-feed-service/pkg/server/jobs.go
```diff
@@ -0,0 +1,110 @@
...
+	"context"
+	"time"
+
+	"github.com/chef/automate/components/event-feed-service/pkg/config"
+	"github.com/chef/automate/components/event-feed-service/pkg/persistence"
+	"github.com/chef/automate/lib/cereal"
+	"github.com/chef/automate/lib/datalifecycle/purge"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/teambition/rrule-go"
```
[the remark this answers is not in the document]
ryancragun: Lolz, `goimports` does everything for me.

components/compliance-service/compliance.go
```diff
@@ -333,37 +334,133 @@ func getManagerConnection(connectionFactory *secureconn.Factory,
...
+		currentPolicies := purge.Policies{
+			Es: map[string]purge.EsPolicy{},
+		}
+		if err = sched.GetParameters(&currentPolicies); err != nil {
+			return nil, errors.Wrap(err, "failed to get purge policies from job manager")
+		}
+
+		for name, dp := range defaultPurgePolicies.Es {
+			p, ok := currentPolicies.Es[name]
+			if ok {
```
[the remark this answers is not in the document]
ryancragun: Yes. The current implementation only includes policies that are desired, so if a user doesn't want the policy configured it's removed from the workflow. It's impossible to tell if they've removed it on purpose or not. We could add an `enabled` field to each policy and manage it and always include every policy in the workflow parameters, but that's not how it works right now.

components/ingest-service/server/job_scheduler_server.go
```diff
@@ -65,13 +65,65 @@ func (server *JobSchedulerServer) GetStatusJobScheduler(ctx context.Context,
 			j.LastElapsed = sched.LastEnd.Sub(*sched.LastStart).String()
 		}
 
-		jobs = append(jobs, &j)
+		return &j, nil
+	}
+
+	var (
+		jobs      = []*ingest.Job{}
```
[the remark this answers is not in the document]
ryancragun: premature optimization on a routine that gets called once :joy_cat:

components/event-feed-service/pkg/server/jobs.go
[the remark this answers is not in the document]
stevendanna: You can configure `goimports` to do this with the `-local github.com/chef` option.

components/ingest-service/server/job_scheduler_server.go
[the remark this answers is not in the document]
stevendanna: I'm not sure I'd call more specifically allocating a slice that we know the exact length of an optimization really.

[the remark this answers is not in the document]
ryancragun: It's the `length of the schedules - 1 + number of purge jobs`, which isn't worth figuring out at initialization time.

components/ingest-service/server/jobs.go
```diff
@@ -43,6 +47,13 @@ func InitializeJobManager(c *cereal.Manager, client backend.Client) error {
 			return errors.Wrapf(err, "failed to register workflow for %q", jobName)
 		}
 	}
+	err = purge.ConfigureManager(
+		PurgeJobName, PurgeScheduleName, c, esSidecarClient,
+	)
+	if err != nil {
+		return errors.Wrap(err, "Failed to configure purge workflow")
```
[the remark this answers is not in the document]
ryancragun: Is it just me or is the suggestion the exact same? :thinking:

components/ingest-service/server/job_scheduler_server.go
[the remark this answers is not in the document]
stevendanna: fine.jpg

components/ingest-service/server/jobs.go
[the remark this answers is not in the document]
stevendanna: strange...how does it look now?

lib/product/types.go
ryancragun: We're not actually using them anywhere in the go code right now, but I wanted  them to be in the product.meta, and the parser for that sets `DisallowUnknownFields()`, so we have to valid stuff for it here.

stevendanna: 👍 Just wanted to make sure I wasn't missing something here.

components/automate-deployment/pkg/deployment/deployment.go
ryancragun: Yeah, you're right. Not sure what I was doing here.

components/ingest-service/server/jobs.go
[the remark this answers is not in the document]
ryancragun: Looks good!

components/event-feed-service/pkg/server/jobs.go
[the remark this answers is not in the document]
ryancragun: Oh my, now I have to figure out how my editor plugin does this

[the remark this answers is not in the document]
ryancragun: `let g:go_fmt_command = "goimports -local github.com/chef"` :sake:

components/automate-deployment/pkg/deployment/deployment.go
```diff
@@ -320,10 +320,31 @@ func (d *Deployment) UpdateExpectedServicesFromManifest() error {
...
+				break
+			}
+		}
+
+		if !found {
+			toRemove = append(toRemove, currentSvc)
+		}
+	}
+
+	for _, svc := range toRemove {
```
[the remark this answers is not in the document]
ryancragun: Nice, for some reason modifying in range is forbidden in my brain.

components/compliance-service/compliance.go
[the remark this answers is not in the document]
ryancragun: Thinking about this some more I'm going to write some words to make clear my thinking on it:

* The "default policies" represent both the policies that can be updated or enabled, and the default values of those policies.
* The "current policies" represent policies that are enabled.
* We do a one time migration from configuration to "current policies", after which "current policies" are managed through the servers `Configure()` endpoint.
* We always migrate the "current policies" index at startup in case there has been a migration.

This means that any new additional default policies will never be added to the "current policies" unless it's added through the `Configure()` endpoint.

Even if we always included all policies in the "current policies" we'd have no idea if was enabled by the user or was enabled by default.

So, if we want to support "adding new current policies that are enabled" and we want to do it safely then each policy will need to have a "disabled" field that we'll set explicitly during migration or configuration and all policies will be set in current policies. Then the actions would have to respect that field.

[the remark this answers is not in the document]
ryancragun: Okay, it's been rewritten to properly migrate previous config, handle new policies, and update existing policies if they have an index or custom field change.

components/ingest-service/integration_test/job_scheduler_test.go
```diff
@@ -46,3 +156,19 @@ func TestJobSchedulerBasic(t *testing.T) {
...
+func verifyPurgeJobs(t *testing.T, status *ingest.JobSchedulerStatus) {
+	// Purge jobs
+	for _, pp := range server.DefaultPurgePolicies.Es {
+		for _, job := range status.Jobs {
+			name := strings.ReplaceAll(pp.Name, "-", "_")
+			name = fmt.Sprintf("%s_%s", server.PurgeJobName, name)
+
+			if name == job.Name {
+				assert.NotEqual(t, pp.Disabled, job.Running, "job is disabled and should be running")
+				// we override the default 30d with 1d in suite setup
```
ryancragun: comment is wrong and should be deleted

lib/datalifecycle/purge/policies.go
```diff
@@ -9,6 +9,21 @@ import (
...
+	Es map[string]EsPolicy `json:"es"`
+	Pg map[string]PgPolicy `json:"pg"`
+}
+
+func NewPolicies() *Policies {
+	return &Policies{
+		Es: map[string]EsPolicy{},
+		Pg: map[string]PgPolicy{},
+	}
+
```
ryancragun: extra line

PR #1220, opened 2019-08-12
Dan/denormalize

danielsdeleo:

Starts moving the applications database to a denormalized schema. This makes the data model more closely match the domain operations in the application. Currently data is only created via health check message with full metadata, and entities such as deployments, service groups, and supervisors exist by having at least one service that includes them in the metadata. Specific benefits of this change are:

* health check messages can be ingested with one Upsert query, which is much simpler than what we do now.
* delete (for disconnected services) can just delete the relevant service rows.
* query to power the applications UI will become simpler. Currently it uses a variable number of joins based on filter criteria, we can get rid of all the joins.

Some of the typical downsides of denormalizing are not present:
* we don't expose any way to update any of the attributes that apply to aggregations of services (e.g., rename an application or environment) as a group; instead users would reconfigure all of the supervisors running those services. So we don't have any fan-out edits we need to worry about.

It's almost certain that this change will increase disk usage a little bit, but disk usage isn't the limiting factor of our scalability, so the reduction in complexity is worth it.

As written currently, this change is essentially disabled by dev-level feature flag. I'm trying to do this in little chunks to prevent merge conflicts with other backend work.

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/applications-service/pkg/storage/postgres/hab_event.go
```diff
@@ -193,7 +211,105 @@ DELETE FROM service_group WHERE service_group.id = $1 AND (NOT EXISTS (SELECT 1
...
+		channel = svcMetadata.GetUpdateConfig().Channel
+		updateStrategy = svcMetadata.GetUpdateConfig().Strategy.String()
+	}
+	_, err := db.DbMap.Exec(upsertServiceAndMetadata,
+		pkgIdent.Origin,                 // origin,
+		pkgIdent.Name,                   // name,
+		pkgIdent.Version,                // version,
+		pkgIdent.Release,                // release,
+		pkgIdent.FullPackageIdent(),     // package_ident,
+		"",                              // status,
```
danielsdeleo: need to fix this for the status messages

components/applications-service/pkg/storage/postgres/hab_event.go
```diff
@@ -193,7 +211,105 @@ DELETE FROM service_group WHERE service_group.id = $1 AND (NOT EXISTS (SELECT 1
 	maybeCleanupDeployment = `
 DELETE FROM deployment WHERE deployment.id = $1 AND (NOT EXISTS (SELECT 1 FROM service WHERE service.deployment_id = deployment.id ))
 `
+	// TODO: add last_event_occurred_at
```
danielsdeleo: will add this before merge

components/applications-service/pkg/ingester/v1/ingest.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 const (
-	numWorkers       = 50
+	numWorkers       = 25
```
danielsdeleo: 🤔 this might be correct? I noticed that running the integration tests while all of A2 was running, I ran out of database connections... probably need to make it configurable eventually.

components/applications-service/pkg/storage/postgres/schema/sql/11_add_denormalized_services_table.up.sql
```diff
@@ -0,0 +1,76 @@
...
+
+  service_group_name        TEXT      NOT NULL,
+  service_group_name_suffix TEXT      NOT NULL,
+
+  -- deployment data
+  application               TEXT      NOT NULL DEFAULT '',
+  environment               TEXT      NOT NULL DEFAULT '',
+
+  -- "time wizard" -- what the previous status was, and when it changed
+  -- TODO: figure out how to use a trigger to fill this out when updating (?)
```
danielsdeleo: TODONE

components/applications-service/pkg/storage/postgres/services.go
```diff
@@ -114,7 +114,7 @@ LEFT JOIN deployment AS d
 LEFT JOIN supervisor AS sup
   ON s.sup_id = sup.id
  %s
-ORDER BY %s %s
+ORDER BY %s %s, sup_member_id ASC
```
kmacgugan: Why do we add this?

components/applications-service/pkg/storage/postgres/hab_event.go
danielsdeleo: had doesn't send this yet, will have to wait.

components/applications-service/pkg/storage/postgres/services.go
danielsdeleo: In case anyone else is wondering:

The primary sort here (or before this patch, the only sort) is on health status, which has a lot of elements that are equal. The order that results are returned in this case is defined by Postgres implementation details (probably whatever is most efficient to read from disk or something like that); I found that adding an insert _to a different table_ was changing the order that results got returned in a few of the integration tests. Adding this secondary sort (on a column that will be unique for each element in this set) makes the ordering of the results completely deterministic.

If the order was changing between requests, then that would be a pagination bug, but I don't think that was happening. In any case, this change means there is one correct order for the results which I think is better than not having a correct ordering if we're already doing a sort anyway.

PR #1221, opened 2019-08-13
update dev docs with info on running a specific build

susanev:

## Overview
Added some information to the dev docs on how to run a specific build of Automate locally.

### DIFF
--- a/dev-docs/DEV_ENVIRONMENT.md
+++ b/dev-docs/DEV_ENVIRONMENT.md
@@ -237,6 +237,11 @@ a different terminal, run
 If you are using vagrant, ssh into the VM again in a separate tab in
 run `sl`.
 
+### Running a Specific Build of Automate
+* Save the manifest `.json` of the particular build you need (e.g., https://packages.chef.io/manifests/current/automate/20190806202030.json is for build 20190806202030)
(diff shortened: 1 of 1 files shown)


dev-docs/DEV_ENVIRONMENT.md
```diff
@@ -237,6 +237,11 @@ a different terminal, run
 If you are using vagrant, ssh into the VM again in a separate tab in
 run `sl`.
 
+### Running a Specific Build of Automate
+* Save the manifest `.json` of the particular build you need (e.g., https://packages.chef.io/manifests/current/automate/20190806202030.json is for build 20190806202030)
+* Run `chef-automate dev configure-deployment --config dev/config.toml --manifest-dir json-file-name`
+* `rebuild components/automate-deployment`
```
srenatus: 🤔 Why is that? I think this last step shouldn't be necessary...?

Also, it's probably worth mentioning that you can grab the `results.tar.gz` from a build (the `verify_private` pipeline's "build" job's artifacts) and use that, but I don't know the concrete steps 💭

srenatus: 💡the bind information for a given manifest is built into deployment-service -- that's why this is needed. (Thanks @stevendanna)

PR #1226, opened 2019-08-13
ci: show *why* a component is rebuilt

srenatus:

This is a small quality-of-life improvement in out build annotations -- I hope!

#### Before:
image

#### After:
image

None
### :+1: Definition of Done
NA
NA

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/scripts/changed_components.rb
+++ b/scripts/changed_components.rb
@@ -39,6 +39,7 @@ build_all = (ENV["BUILDKITE_BRANCH"] || "").include?("verify-rebuild-all") || EN
 # on us, we could end up missing a required dependency.
 #
 changed_packages = Hash.new
+changed_package_files = Hash.new
 hab_deps = Hash.new
 
 config.each do |pkg_name, metadata|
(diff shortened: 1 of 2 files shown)


srenatus:
This could become messy when many files change -- but let's try. We could introduce some kind of ellipsis later.

msorens:
I'm puzzled by the sample output above. Can you help me understand why a change exclusively in compliance requires rebuilding other services? I understand the other service interacts with compliance but how does that translate into the other service needing rebuilding?

stevendanna:
@msorens This is exactly the kind of information that I think this change will help socialize.

Without going into the details of the *specific* files here, here is why these kind of unexpected changes crop up:

0) When we do a build, we look at all changed files and compare that to a set of globs for each project. The globs are stored in `.bldr.toml`. That file is generated by looking at all of the go-level dependencies for a project and then consolidating the list, you can see the generation code in `tools/bldr-config-gen`.

1) The compliance-service and services that started life in compliance-service still have their protobuf APIs inside their component folders rather than inside the `api/` folder.

2) Some services depend on packages in other services because they wanted a quick way to share useful code.

3) The "consolidation" mentioned in (0) and the inter-relationships in (1) and (2) often mean that a change to compliance, nodemanager-service, or ingest-service ends up rebuilding many unexpected things.

PR #1235, opened 2019-08-13
Preventing empty error message suggestions

lancewf:
Preventing empty error message suggestions like the one shown below. 
image
The problem was that the `errorMessage` field was filled with an empty string if there was no error. Therefore the empty string suggestion was correct. The bug was prevented in three places ingestion, request to the API from the UI, and the search bar component. 

https://github.com/chef/automate/pull/1211#issuecomment-520996870

### :+1: Definition of Done

The blank suggestion does not appear in the search bar

1. `build components/ingest-service && build components/automate-ui && start_all_services`
1. Add some nodes with empty errors `chef_load_nodes 100`
1. Open https://a2-dev.test/infrastructure/client-runs
1. Type one letter in the search bar for catagory "Error"
1. Ensure the empty suggestion is not displayed.  

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/entities/client-runs/client-runs.requests.ts
+++ b/components/automate-ui/src/app/entities/client-runs/client-runs.requests.ts
@@ -20,6 +20,10 @@ const CONFIG_MGMT_URL = environment.config_mgmt_url;
 const INGEST_URL = environment.ingest_url;
 const DEPLOYMENT_URL = environment.deployment_url;
 
+interface RespSuggestion {
+  text: string;
+}
+
 @Injectable()
 export class ClientRunsRequests {
   constructor(
@@ -79,7 +83,7 @@ export class ClientRunsRequests {
       const params = this.formatFilters(filters).set('type', type).set('text', text);
       const url = `${CONFIG_MGMT_URL}/suggestions`;
 
-      return this.httpClient.get<any[]>(url, {params}).pipe(map(
+      return this.httpClient.get<RespSuggestion[]>(url, {params}).pipe(map(
         (suggestions) => suggestions.filter(s => s && s.text && s.text.length !== 0)));
     } else {
       return observableOf([]);
--- a/components/automate-ui/src/app/entities/client-runs/client-runs.requests.ts
+++ b/components/automate-ui/src/app/entities/client-runs/client-runs.requests.ts
@@ -79,7 +79,8 @@ export class ClientRunsRequests {
       const params = this.formatFilters(filters).set('type', type).set('text', text);
       const url = `${CONFIG_MGMT_URL}/suggestions`;
 
-      return this.httpClient.get<Chicklet[]>(url, {params});
+      return this.httpClient.get<any[]>(url, {params}).pipe(map(
+        (suggestions) => suggestions.filter(s => s && s.text && s.text.length !== 0)));
     } else {
       return observableOf([]);
     }
--- a/components/automate-ui/src/app/page-components/search-bar/search-bar.component.spec.ts
+++ b/components/automate-ui/src/app/page-components/search-bar/search-bar.component.spec.ts
@@ -288,7 +288,7 @@ describe('SearchBarComponent', () => {
         });
         fixture.detectChanges();
 
-        expect(component.suggestions.size).toEqual(3);
+        expect(component.suggestions.sort()).toEqual(['1', '2', '3']);
       });
     });
 
--- a/components/automate-ui/src/app/page-components/search-bar/search-bar.component.spec.ts
+++ b/components/automate-ui/src/app/page-components/search-bar/search-bar.component.spec.ts
@@ -288,7 +288,7 @@ describe('SearchBarComponent', () => {
         });
         fixture.detectChanges();
 
-        expect(component.suggestions.sort()).toEqual(['1', '2', '3']);
+        expect(component.suggestions.map(s => s.name).toArray().sort()).toEqual(['1', '2', '3']);
       });
     });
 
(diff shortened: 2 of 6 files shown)


components/ingest-service/backend/converge.go
```diff
@@ -31,7 +31,7 @@ type Node struct {
 	HasDeprecations   bool      `json:"has_deprecations"`
 	DeprecationsCount int       `json:"deprecations_count"`
 	Projects          []string  `json:"projects"`
-	ErrorMessage      string    `json:"error_message"`
+	ErrorMessage      string    `json:"error_message,omitempty"`
```
lancewf: This prevents and empty string from being stored in elasticsearch.

components/automate-ui/src/app/entities/client-runs/client-runs.requests.ts
```diff
@@ -79,7 +79,8 @@ export class ClientRunsRequests {
       const params = this.formatFilters(filters).set('type', type).set('text', text);
       const url = `${CONFIG_MGMT_URL}/suggestions`;
 
-      return this.httpClient.get<Chicklet[]>(url, {params});
+      return this.httpClient.get<any[]>(url, {params}).pipe(map(
+        (suggestions) => suggestions.filter(s => s && s.text && s.text.length !== 0)));
```
lancewf: Removing empty suggestion from the API response

components/automate-ui/src/app/page-components/search-bar/search-bar.component.ts
```diff
@@ -71,7 +71,8 @@ export class SearchBarComponent implements OnChanges {
   ngOnChanges(changes: SimpleChanges) {
     if (changes.dynamicSuggestions) {
       this.suggestions = List<SuggestionItem>(
-        compact(changes.dynamicSuggestions.currentValue.map((suggestion) => {
+        compact(changes.dynamicSuggestions.currentValue.filter(
+          (suggestion) => suggestion && suggestion.length !== 0).map((suggestion) => {
```
lancewf: Removing empty suggestion passed to the search bar component

components/config-mgmt-service/integration_test/suggestions_test.go
```diff
@@ -873,13 +925,13 @@ func TestSuggestionsWithTableDriven(t *testing.T) {
 		// Suggestions for Environments
 		{"should return all environment suggestions",
 			request.Suggestion{Type: "environment"},
-			[]string{"dev", "prod", "games", "new-york"}},
+			[]string{"dev", "prod", "games", "new-york", ""}},
```
lancewf: The empty string was added because I changed the `extractTextFromSuggestionsResponse` function in this test to return suggestions that are empty.

components/config-mgmt-service/integration_test/suggestions_test.go
```diff
@@ -1323,16 +1375,14 @@ func TestSuggestionsProjectFilter(t *testing.T) {
 // TODO: (@afiune) Is this the normal behavior? If not lets fix it.
 // for now the fixt in the tests will be to check if there is a "string"
 // value or not.
-func extractTextFromSuggestionsResponse(list *gp.ListValue, t *testing.T) []string {
```
lancewf: The ` t *testing.T` was not being used, so it was removed.

components/automate-ui/src/app/entities/client-runs/client-runs.requests.ts
```diff
@@ -79,7 +79,8 @@ export class ClientRunsRequests {
       const params = this.formatFilters(filters).set('type', type).set('text', text);
       const url = `${CONFIG_MGMT_URL}/suggestions`;
 
-      return this.httpClient.get<Chicklet[]>(url, {params});
+      return this.httpClient.get<any[]>(url, {params}).pipe(map(
```
phiggins: Did you mean to loosen the type of the argument from `Chicklet` to `any` here?

components/automate-ui/src/app/page-components/search-bar/search-bar.component.spec.ts
```diff
@@ -279,6 +279,17 @@ describe('SearchBarComponent', () => {
...
+
+      it('remove empty string returned suggestions', () => {
+        component.selectedCategoryType = component.categories[0];
+        component.dynamicSuggestions = ['1', '2', '3', '', undefined, null];
+        component.ngOnChanges({
+          dynamicSuggestions: new SimpleChange(null, component.dynamicSuggestions, false)
+        });
+        fixture.detectChanges();
+
+        expect(component.suggestions.size).toEqual(3);
```
phiggins: This assertion is not very specific. Could this be replaced with something like this?

```
         expect(component.suggestions.sort()).toEqual(['1', '2', '3']);
```

components/automate-ui/src/app/page-components/search-bar/search-bar.component.ts
phiggins: If you used lodash, you might be able to simplify this a bit like `.filter(isEmpty)`.

components/automate-ui/src/app/entities/client-runs/client-runs.requests.ts
lancewf: Yes, I did. The API is not returning a `Chicklet`. As I just noticed today a `Chicklet` is `{text:string, type:string}` and `{"score":1, "text": "some suggestion"}` is returned. We only want the text part. 
But I will create a new interface with text only.

components/automate-ui/src/app/page-components/search-bar/search-bar.component.spec.ts
lancewf: I agree that would be better. Thanks.

lancewf:
@phiggins Thank you for the thorough review. I update the PR with all your suggestions.

PR #1238, opened 2019-08-14
ci: use <details> for files causing rebuilds

srenatus:
In #1226, we've added outputting why components are rebuild. It output can easily go out of hand. An example for this is #1237, where we see a few screenfuls of this:

image

With the change introduced here, these will be collapsed. What you'll see now is
image

...and when you're wondering why, say, `automate-cli` is going to be rebuilt, you can open the details thingy:

image


ryancragun:
Thank you

PR #1243, opened 2019-08-14
automate-dex: bump core/dex (2.17.0 -> 2.18.0), adapt code

srenatus:
### :nut_and_bolt: Description: What code changed, and why?

⚠️Needs https://github.com/habitat-sh/core-plans/pull/2972  to be merged and promoted first

Pulls in dex 2.18.0, and gets rid of some cruft code we've had because we had been lacking that VerifyPassword method that's in the latest release.

### :chains: Related Resources

None.

### :+1: Definition of Done

Stuff still works ✔️Where "stuff" means changing a local user's password.

### :athletic_shoe: How to Build and Test the Change

- `start_all_services`
- `rebuild components/local-user-service`, `rebuild components/automate-dex`
- use the UI to change some passwords

### :white_check_mark: Checklist

- [X] Tests added/updated?
- [ ] Docs added/updated?

PR #1246, opened 2019-08-14
session-service: send the user to dex with their connector selected already

srenatus:

When a user's session cannot be refreshed anymore, they'll be sent to the _Sign In_ page. There, they'll have to select a method to use for Sign In, although we know what they had used before.

With this change, we'll make use of that knowledge, and send them to the right method immediately.

❕*Note*: The `Back` link is a bit wonky: It'll send them back to Automate, but they'll lack an ID token, so they'll be sent back to dex, but since there's no ID token, they'll see the selection. So, it **works**, but in a very odd way. Unfortunately, a proper fix is a bit more involved.

❕ *Note*: For SAML, there's no back button. So, to abort, and get back to the selection once you've used SAML, you'll need to browse to Automate again **manually**.

☑️ Depends on #1243 to get in first, this needs dex v2.18.0. ✔️ 

☑️ *TODO*
- [x] *Verify signout* for SAML and non-SAML: if you're signing out, you should be sent back to the sign in method chooser .

None.

### :+1: Definition of Done

A user that's logged out no longer needs to select their login method. But if they want to, they can.

- `rebuild components/session-service`
- get the UI to be rebuilt (devproxy etc, or `rebuild components/automate-ui`
- login as a local user
- kill all sessions: `chef-automate dev psql chef_session_service -- -c 'delete from sessions'`
- wait for a minute, see that you're sent to the username/password form directly

#### SAML

To see how this behaves with SAML, `rebuild components/automate-dex` after changing the hardcoded token expiry for using SAML like this:
```diff
diff --git a/components/automate-dex/habitat/templates/config.yml b/components/automate-dex/habitat/templates/config.yml
index f29edd636..4e1510ee0 100644
--- a/components/automate-dex/habitat/templates/config.yml
+++ b/components/automate-dex/habitat/templates/config.yml
@@ -227,7 +227,7 @@ expiry:
   signingKeys: {{cfg.expiry.signing_keys}}
   {{- if cfg.connectors.saml}}
   # SAML is being used, so we cannot tweak ID token expiry
-  idTokens: 24h
+  idTokens: 1m
   {{- else}}
   idTokens: {{cfg.expiry.id_tokens}}
   {{- end}}
```
Then login via SAML and wait for a minute. You'll see a series of flickers while your browser goes to Okta, finds the session still intact, and redirects you to Automate.

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/services/chef-session/chef-session.service.spec.ts
+++ b/components/automate-ui/src/app/services/chef-session/chef-session.service.spec.ts
@@ -143,7 +143,7 @@ describe('ChefSessionService', () => {
 
       it('does not ingest any token, but calls logout', () => {
         service.refreshSessionCallback(event);
-        expect(service.logout).toHaveBeenCalledWith('/some/path');
+        expect(service.logout).toHaveBeenCalledWith('/some/path', true);
         expect(service.ingestIDToken).not.toHaveBeenCalled();
       });
     });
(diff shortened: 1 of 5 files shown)


srenatus:
☑️ Added a test.

Issue #959, opened 2019-07-18
cypress: deflake user mgmt tests


susanev:
## Overview

* we need to deflake the [commented out tests](https://github.com/chef/automate/blob/master/e2e/cypress/integration/settings/01_user_mgmt.spec.js#L53-L102) in {‌{user_mgmt_spec}‌}
* tests currently pass locally but proved to be too flaky when run against acceptance environments.
* will require some digging to determine the exact causes of flakiness

PR #1247, opened 2019-08-14
Cypress: restore user management tests

bcmdarroch:

We had the user management tests commented out because they were flaky. I made a few tweaks and wanted to see if they'll be more stable now.

Edit: local runs are stable but pipeline runs still need some debugging

-  https://github.com/chef/automate/issues/959

### :+1: Definition of Done
- buildkite tests pass reliably (gonna retry them a few times)

- see buildkite
- [run locally](https://github.com/chef/automate/blob/master/e2e/README.md#running-cypress-locally)

- [x] Tests added/updated?
- [x] Docs added/updated?

### DIFF
--- a/e2e/cypress/integration/ui/compliance/01_ssh_scan_job.spec.ts
+++ b/e2e/cypress/integration/ui/compliance/01_ssh_scan_job.spec.ts
@@ -2,7 +2,7 @@ import { Base64 } from 'js-base64';
 
 describe('create a manual node ssh scan job and cleanup after', () => {
   before(() => {
-    cy.adminLogin('/');
+    cy.adminLogin('/settings');
   });
   beforeEach(() => {
     // cypress clears local storage between tests
@@ -18,10 +18,6 @@ describe('create a manual node ssh scan job and cleanup after', () => {
   const decoded = Base64.decode(Cypress.env('AUTOMATE_ACCEPTANCE_TARGET_KEY'));
 
   it('can create a credential with a username and key', () => {
-    // navigate to credentials create page:
-    // click on settings
-    cy.get('.nav-link').contains('Settings').click();
-
     // we save the route that will be called when we navigate to the page
     // in order to be able to wait for it later
     cy.route('POST', '/api/v0/secrets/search').as('getSecrets');
@@ -41,8 +37,9 @@ describe('create a manual node ssh scan job and cleanup after', () => {
       cy.get('form input[formcontrolname="name"]').first().type(credName);
       cy.get('form input[formcontrolname="username"]').first()
         .type(Cypress.env('AUTOMATE_ACCEPTANCE_TARGET_USER'));
-      cy.get('form textarea[formcontrolname="key"]')
-        .type(decoded);
+
+      // we save some time by setting this long string as the value rather than typing it out
+      cy.get('form textarea[formcontrolname="key"]').clear().invoke('val', decoded).trigger('input')
 
       // we save the route that will be called when we navigate to the page
       // in order to be able to wait for it later
--- a/e2e/cypress/integration/ui/compliance/01_ssh_scan_job.spec.ts
+++ b/e2e/cypress/integration/ui/compliance/01_ssh_scan_job.spec.ts
@@ -2,7 +2,7 @@ import { Base64 } from 'js-base64';
 
 describe('create a manual node ssh scan job and cleanup after', () => {
   before(() => {
-    cy.adminLogin('/settings');
+    cy.adminLogin('/settings/node-credentials');
   });
   beforeEach(() => {
     // cypress clears local storage between tests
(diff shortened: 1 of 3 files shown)


srenatus:
🔍 from the screenshots, it looks like you're losing keypresses somewhere -- the created users have names like `cs test user` or `cess test user`, so some of the first chars get lost. 😕

e2e/cypress/integration/ui/settings/01_user_mgmt.spec.ts
```diff
@@ -26,81 +27,77 @@ describe('user management', () => {
     cy.get('app-user-table chef-button').contains('Create User').click();
     cy.get('app-user-management chef-modal').should('exist');
 
+    // adding a delay to more closely approximate an average human's type speed (200 char/min)
+    // the `should` will wait until the actual input value matches the expected value
+    // ref: https://github.com/cypress-io/cypress/issues/534
     cy.get('[formcontrolname=fullname]')
-      .type(name);
+      .type(name, { delay: 50 }).should('have.value', name);
```
msorens: I would suggest either this...
```suggestion
      .type(name, TYPE_DELAY).should('have.value', name);
```
or at least this...
```suggestion
      .type(name, { delay: TYPE_DELAY }).should('have.value', name);
```
where you create a global constant of appropriate format rather than propagate that magic number in many places.

(Or, if it is not easy to do global constants with Cypress, at a minimum a file level constant.)

e2e/cypress/integration/ui/settings/01_user_mgmt.spec.ts
```diff
@@ -26,81 +27,77 @@ describe('user management', () => {
...
+    cy.get('[data-cy=save-user]').click();
+    cy.get('app-user-management chef-modal').should('not.be.visible');
+    cy.get('chef-notification.info').should('be.visible');
+
+    cy.get('chef-td').contains(username).should('exist');
+    cy.get('chef-td').contains(name).should('exist');
+  });
+
+  it('can view and edit user details', () => {
+    cy.route('PUT', '**/users/**').as('updateUser');
```
msorens: Seems a very vague specification; can we not be more specific in the path...? If not, probably should have an in-line comment explaining why not. (Same thing for DELETE lower in this file.)

e2e/cypress/integration/ui/settings/01_user_mgmt.spec.ts
```diff
@@ -26,81 +27,77 @@ describe('user management', () => {
...
-  //         cy.get('chef-tbody chef-td').contains(username).should('not.exist')
-  //       })
-  //     })
-  //   })
-  // })
+  it('can delete user', () => {
+    cy.route('DELETE', '**/users/**').as('deleteUser');
+
+      cy.get('chef-td').contains(name).parent().parent()
+        .find('chef-control-menu').as('controlMenu');
```
tylercloke: Ah cool!

e2e/cypress/integration/ui/settings/01_user_mgmt.spec.ts
bcmdarroch: I think the verb is specific enough in these cases.

e2e/cypress/integration/ui/settings/01_user_mgmt.spec.ts
bcmdarroch: Good point! I'm planning to turn this into a helper which should resolve the issue

e2e/cypress/integration/ui/compliance/01_ssh_scan_job.spec.ts
```diff
@@ -2,7 +2,7 @@ import { Base64 } from 'js-base64';
 
 describe('create a manual node ssh scan job and cleanup after', () => {
   before(() => {
-    cy.adminLogin('/');
+    cy.adminLogin('/settings/node-credentials');
```
bcmdarroch: Took off some extra page loading time by directly navigating to the first tested page rather than opening the event feed, then the settings page, then node credentials

e2e/cypress/integration/ui/compliance/01_ssh_scan_job.spec.ts
```diff
@@ -18,100 +18,81 @@ describe('create a manual node ssh scan job and cleanup after', () => {
...
-    cy.get('chef-sidebar-entry').contains('Node Credentials').click();
-    cy.url().should('include', '/settings/node-credentials');
-
-    // wait for data to return
-    cy.wait('@getSecrets');
-
     // click on add credential button to open create form
-    cy.contains('Add Credential').click().then(() => {
-      cy.url().should('include', '/settings/node-credentials/add');
+    cy.contains('Add Credential').click();
```
bcmdarroch: using an empty `then()` is an anti-pattern that shouldn't be necessary since all Cypress commands are run in sequence. `then()` is useful for when we need to do something with the element returned by `get` (per [this doc](https://docs.cypress.io/api/commands/then.html#Syntax))

e2e/cypress/integration/ui/compliance/01_ssh_scan_job.spec.ts
```diff
@@ -18,100 +18,81 @@ describe('create a manual node ssh scan job and cleanup after', () => {
...
+    // fill in name for credential, username, and key
+    cy.get('form input[formcontrolname="name"]').first().type(credName);
+    cy.get('form input[formcontrolname="username"]').first()
+      .type(Cypress.env('AUTOMATE_ACCEPTANCE_TARGET_USER'));
 
-      // we save the route that will be called when we navigate to the page
-      // in order to be able to wait for it later
-      cy.route('POST', '/api/v0/secrets').as('createSecret');
+    // we save some time by setting this long string as the value rather than typing it out
+    cy.get('form textarea[formcontrolname="key"]').clear().invoke('val', decoded).trigger('input');
```
bcmdarroch: this cert takes a while to type, and most people would just copy-paste it anyway, so I decided to programmatically set the value instead of using `type` (see [this issue](https://github.com/cypress-io/cypress/issues/1123#issuecomment-485329565))

e2e/cypress/integration/ui/compliance/01_ssh_scan_job.spec.ts
```diff
@@ -18,100 +18,81 @@ describe('create a manual node ssh scan job and cleanup after', () => {
...
   it('can create a node and attach a credential', () => {
     // navigate to node create page:
     // click on scan jobs
     cy.get('.nav-link').contains('Compliance').click();
-    cy.url().should('include', '/compliance/reports/overview');
 
-    // click into scan jobs
-    cy.get('chef-sidebar-entry').contains('Scan Jobs').click();
+    // add force to avoid waiting on the whole page to load
+    cy.get('chef-sidebar-entry').contains('Scan Jobs').click({ force: true });
```
bcmdarroch: Slow page loading has been the main reason this test flakes out. I decided to add `force: true` to parts of the test where we're clicking buttons to go to other pages. Opening pages takes the longest load time, so I'd rather fudge those clicks so we can get to testing the actual functionality of the page we're trying to get to.

e2e/cypress/support/commands.ts
```diff
@@ -36,7 +36,7 @@ Cypress.Commands.add('applyProjectsFilter', (projectsToFilterOn: string[]) => {
     .find('chef-checkbox').each(child => {
     // deselect every checkbox
     if (child.attr('aria-checked') === 'true') {
-      child.click();
+      child.trigger('click');
```
bcmdarroch: got a lint warning about this: the `child` is an HTML element, so we use JQuery instead of Cypress to click on it. JQuery has deprecated `click()` in favor of `trigger("click")`

PR #1248, opened 2019-08-14
Cypress: tests for projects management in the UI

bcmdarroch:
Follow up to #1156, adding Cypress test coverage

#1156 

### :+1: Definition of Done
- Cypress tests pass locally and on buildkite

- see buildkite
- [run locally](https://github.com/chef/automate/blob/master/e2e/README.md#running-cypress-locally)

- [x] Tests added/updated?


e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+    cy.restoreStorage();
+  });
+
+  afterEach(() => {
+    cy.saveStorage();
+  });
+
+
+  it('list projects', () => {
+    let projectIDs: string[];
```
srenatus: I might be missing something, but do we need this array in this scope?

In [line 51](https://github.com/chef/automate/pull/1248/files#diff-08c99676671eeb48b64efed39ef8b002R51), couldn't we do
```typescript
const projectIDs = response.body.projects.map((project: Project) => project.id);
projectIDs.forEach((id: string) => { // ...
```

or even
```typescript
response.body.projects.map((project: Project) => project.id)
  .forEach((id: string) => { // ...
```
?

e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
```diff
@@ -0,0 +1,132 @@
+import forEach = require("cypress/types/lodash/forEach");
```
srenatus: ❓ What would this be needed for? AFAICT the `.forEach`s below are just [standard JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach#Syntax) 🤔

e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
bcmdarroch: oo good point

e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
bcmdarroch: yeah that seemed to auto-import itself for some reason 🤷‍♀

srenatus:
Looks like this is plagued with the same type-too-fast issue as the other one?
https://buildkite.com/chef/chef-automate-master-verify-private/builds/4523#7fc5714e-415e-48c3-9c45-f84fba7afd37/518-883

e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
```diff
@@ -0,0 +1,206 @@
+interface Project {
+  id: string;
+  name: string;
+  type: string;
+}
+
+// assume 2.0 if not in CI. if you wish something different start cypress with
+// IAM_VERSION set to what you are testing.
+if (Cypress.env('IAM_VERSION') === undefined) {
```
srenatus: Can we put this logic, and the `describeIAMV2P1` method, into some helpers?

e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+  beforeEach(() => {
+    cy.restoreStorage();
+  });
+
+  afterEach(() => {
+    cy.saveStorage();
+  });
+
+
+  it('list projects', () => {
```
msorens: For consistency, need more words, e.g.:
```suggestion
  it('displays a list of projects', () => {
```

e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+    cy.get('[data-cy=rule-value] input').type('Chef Policy Group Foo');
+
+    cy.get('#right-buttons button').contains('Save Rule').click();
+    cy.get('chef-page').should('not.be.visible');
+
+    cy.url().should('include', `/settings/projects/${projectID}`);
+    cy.get('chef-td').contains(ruleID);
+  });
+
+  it('can see a list of rules for a project', () => {
```
msorens:
```suggestion
  it('displays a list of rules for a project', () => {
```

e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+    cy.get('@controlMenu').should('be.visible')
+      .click();
+    cy.get('@controlMenu').find('[data-cy=delete-rule]').click({force: true});
+
+    cy.get('chef-button').contains('Delete Rule').click();
+
+    cy.get('chef-tbody chef-td').contains(ruleID).should('not.exist');
+  });
+
+  it('delete a project', () => {
```
msorens:
```suggestion
  it('can delete a project', () => {
```

e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+  it('list projects', () => {
+    cy.request({
+      auth: { bearer: adminToken },
+      method: 'GET',
+      url: '/apis/iam/v2beta/projects'
+    }).then((response) => {
+      expect(response.status).to.equal(200);
+      response.body.projects.map((project: Project) => project.id)
+        .forEach((id: string) => {
+        cy.get('chef-table chef-td').contains(id);
```
msorens:
```suggestion
          cy.get('chef-table chef-td').contains(id);
```

e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+    // Node
+    cy.get('#create-type-dropdown').select('Node');
+    const nodeAttributes = ['Chef Organization', 'Chef Server', 'Environment',
+        'Chef Role', 'Chef Tag', 'Chef Policy Name', 'Chef Policy Group'];
+
+    nodeAttributes.forEach((att: string) => {
+        cy.get('[data-cy=attribute-dropdown]').select(att);
+    });
+
+    cy.get('[data-cy=operator-dropdown]').select('member of');
```
msorens: This line is not doing anything.

e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+    cy.get('[data-cy=attribute-dropdown]:last')
+      .select('Environment');
+
+    cy.get('[data-cy=operator-dropdown]:last')
+      .select('member of');
+
+    cy.get('[data-cy=rule-value]:last input')
+      .type('Dev, Prod, Test');
+
+    cy.get('#right-buttons button').contains('Save Rule').click();
```
msorens: Consider testing that the Save Rule button is, in fact, enabled.

e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+    cy.get('chef-td').contains('2 conditions');
+  });
+
+  it('can update a project name', () => {
+    const updatedProjectName = `updated ${projectName}`;
+
+    cy.get('[data-cy=details-tab]').click();
+    cy.get('#update-name').find('input').clear().type(updatedProjectName, { delay: 50 })
+      .should('have.value', updatedProjectName);
+    cy.get('chef-button').contains('Save').click();
```
msorens: Again should check for Save being enabled.

e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
bcmdarroch: Unfortunately it's not possible to run custom Cypress commands outside running tests. It would require adding an extra `describe` and `it` block to the tests that use this helper, which seems to me like it would defeat the purpose of the helper. I'll investigate how to add `utility` helpers that don't require the Cypress test runner to be up

e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+    });
+  });
+
+  it('can create a project', () => {
+    cy.get('#create-button').contains('Create Project').click();
+    cy.get('app-create-object-modal chef-modal').should('have.class', 'visible');
+
+    // we increase the default delay to mimic the average human's typing speed
+    // only need this for input values upon which later test assertions depend
+    // ref: https://github.com/cypress-io/cypress/issues/534
```
tylercloke: So where do we actually need the typing helper? Are we just doing it in places its failed without it before? If we need it in a bunch of places, we should make a helper.

e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
bcmdarroch: We wanted to have the test select each available option to confirm they populate as expected.

e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
bcmdarroch: Cypress takes care of that with [actionability checks](https://docs.cypress.io/guides/core-concepts/interacting-with-elements.html#Actionability) on every `click`

e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
bcmdarroch: It's for any input that must be entered correctly for later assertions to pass. So if I create a user named `test` but it gets typed `tst`, when the test tries to click on the name link, it'll be expecting `test`, not `tst`.

Helper seems like a good idea for consistency

bcmdarroch: Might make that a follow-up though because I'm a bit Cypress'd out 😅

e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
msorens: Oh, of course. That makes sense. 👍

e2e/cypress/integration/ui/settings/04_project_mgmt.spec.ts
msorens: Nice. 👍

Issue #854, opened 2019-07-10
IAM v2.1 Ingest mapping rules on projects


susanev:
## Overview

Users are able to define ingest rules and conditions for individual projects. Note, that [Enforce IAM project filtering for nodes list, search agg data, node details](https://chefio.atlassian.net/browse/A2-320) needs to be completed before users will be able to apply the ingest rules to their ingesting nodes.

[Video](https://drive.google.com/file/d/16h8gcw-UOiKRqRNvbf29sMv-EL33DOxo/view?usp=sharing)

## Subtasks
- [x] Full-page design stencil component @tarablack01 #560 
- [x] Projects list page ui @scottopherson #511 
- [x] Projects detail page ui @tarablack01 #736
- [x] Projects create/update ingest rule ui @tarablack01 #736
- [x] Fix bug with subheading wrapping and pushing the iframe down too far @tarablack01 #560 
- [x] Follow-on pr with some clean up related to #736 @msorens #1020 
- [x] Add URL fragments for projects (#rules and #details), similar to teams (#users and #details) and policy (#definition and #members). @msorens #1194
- [x] [Connect APIs to UI components on project detail pages](https://app.zeplin.io/project/5ba808703eb54372d860c737/dashboard?tag=list-view&tag=projects) (NOTE: all requests are already wired in ngrx store)
    - [x] Get Project Rules / Ingest Rules table @blakestier @bcmdarroch #1156 
    - [x] Delete Rule / Delete rule button and modal @blakestier @bcmdarroch #1156 
- [x] Connect APIs to UI components for [create ingest rules](https://zpl.io/anydmek) and [update ingest rules](https://zpl.io/aMWKm1N)
    - [x] Create Rule / Create Rule button and Create Rule modal @blakestier @bcmdarroch #1156 
    - [x] Update Rule / Update Rule button and Update Rule modal @blakestier @bcmdarroch #1156
- [x] [Connect APIs to UI components for project list page](https://app.zeplin.io/project/5ba808703eb54372d860c737/dashboard?tag=list-view&tag=projects)
    - [x] Update All Projects / Update Projects button and modal
    - [x] Project Update Percentage / Update Projects button
    - [x] Project Rule Edits Status / Ingest Rules column @msorens #1250
    - [x] Project Update Status / Project Update Status button @msorens #1309
    - [x] Delete Project / Delete Project button and modal
    - [x] Connect APIs to UI components for stop project update @msorens #1309 
[...]

PR #1250, opened 2019-08-15
[automate-854] Update rules status on Project List page

msorens:

Prior to this PR, the `Ingest Rules` and `Project Update Status` columns on the Projects List page just showed hard-coded, static values ("Applied" and "OK").
This PR brings live data to the `Ingest Rules` column as shown below.

Remaining work:
- [x] refresh Projects List page when rules are applied

I have elected to save this next bit for a separate PR. This current PR is fairly-self contained and thus gives a smaller PR.

- [ ] Activate `Project Update Status` column with live data

The interesting bits here are two:

 - Revising the `query_projects` SQL function to return an additional, computed `status` field, obviating the need for the Go code to do various machinations to determine the value.
 - Updating the property-based tests to handle multiple projects with a mix of no rules, staged rules, and applied rules, so I can then cover ListProjects without having to write all the permutations manually. The new tests and test helpers for this are in rules_property_test.go. Note that these are integration tests, in that they exercise the authz-service API calls to `CreateProject`, `CreateRule`, `ListProjects`, and `ApplyRulesStart`, and talk to the real DB! I decided with those it would not really add any value to add tests in project_test.go.
(You will, though, also find unit tests in postgres_test.go that were instrumental during development.)

Here's a sample set of runs below: Each run begins with `LIST n PROJECTS`. Then, within that run, it shows one or more projects listed plus the random point at which rules are applied. Any given run, therefore, may have projects with staged rules, with applied rules, or with both. 

Note that one test case **not** covered is a single project having both staged and applied rules, as that would take a lot more customization to get at that level of refinement. 

```
LIST 1 PROJECTS ===>
*** applying rules now ***
===> project '__0-0' has 1 rules
LIST 2 PROJECTS ===>
===> project '_---0' has 3 rules
*** applying rules now ***
===> project '_-1' has 2 rules
LIST 1 PROJECTS ===>
*** applying rules now ***
===> project 'c-0' has 4 rules
LIST 5 PROJECTS ===>
===> project 'b-0' has 2 rules
===> project '_-1' has 4 rules
*** applying rules now ***
===> project '5-2' has 3 rules
===> project 'j__7-3' has 0 rules
===> project '_-4' has 0 rules
LIST 5 PROJECTS ===>
===> project '8-0' has 2 rules
===> project '_0764-1' has 5 rules
===> project '_--2' has 4 rules
===> project '_p-3' has 6 rules
*** applying rules now ***
===> project '--4' has 1 rules
LIST 6 PROJECTS ===>
===> project 'o-0' has 7 rules
*** applying rules now ***
===> project '_-1' has 1 rules
===> project '1-2' has 7 rules
===> project 't-3' has 0 rules
===> project 'u-4' has 0 rules
===> project '7-5' has 0 rules
LIST 1 PROJECTS ===>
*** applying rules now ***
===> project '38c9-0' has 2 rules
LIST 2 PROJECTS ===>
*** applying rules now ***
===> project '73--0' has 6 rules
===> project 'h-1' has 0 rules
```

Issue - #854 Ingest mapping rules on projects

### :+1: Definition of Done

The `Ingest Rules` column now reflects real project status:

image

```
# rebuild components/automate-gateway
# rebuild components/authz-service
```
Also rebuild automate-ui based on your own environment.


components/automate-ui/src/app/entities/rules/rule.reducer.ts
```diff
@@ -76,7 +78,8 @@ export const RuleEntityInitialState: RuleEntityState = ruleEntityAdapter.getInit
   createStatus: EntityStatus.notLoaded,
   createError: null,
   deleteStatus: EntityStatus.notLoaded,
-  updateStatus: EntityStatus.notLoaded
+  updateStatus: EntityStatus.notLoaded,
+  projectRulesStatus: <ProjectStatus>'PROJECT_RULES_STATUS_UNSET'
```
srenatus: ❓ Is this needed? `RuleEntityState` has that type, too, so I'd suppose this would be sufficient to guarantee you don't just set anything here....?

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -194,4 +194,11 @@ export class ProjectListComponent implements OnInit {
   cancelApplyStop() {
     this.closeConfirmApplyStopModal();
   }
+
+  getRulesStatus(project: Project): string {
+    return project.status === 'NO_RULES' ? 'No rules'
```
srenatus: Let's use a switch here, please 😉

components/automate-ui/src/app/entities/rules/rule.reducer.ts
msorens: It actually is needed; that surprised me, too. There's a compiler error without it.

msorens: Ah, thanks for noticing that. It actually is needed there because the system does not know this is a `RuleEntityState`--it is just an object we are passing to `ruleEntityAdapter.getInitialState` and that has an untyped argument.
That said, there is a better way to write this so it is not needed--see new commit for that.

components/automate-ui/src/app/pages/project/list/project-list.component.ts
msorens: My preference is for functional-style code, which requires a ternary as given rather than a switch. However, sounds like your preference is strong, so I am happy to oblige.

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -460,7 +463,108 @@ func TestDeleteRuleProperties(t *testing.T) {
...
 				return markedforDeletion && fullyDeleted
 			},
-			createProjectAndRuleGen,
+			createProjectAndRulesGen,
+		))
+
+	properties.TestingRun(t)
+}
+
+func TestListProjectProperties(t *testing.T) {
```
msorens: This is a "base" test; in case of a failure it would be simpler to trace here rather than in the next test function. Also, this test is checking all fields of a project, whereas the next test function is only checking rules status. That said, every property-based test like this takes from a few seconds to 10 or 20 seconds due to its multiple iterations, so there is a case for deleting this one.

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -460,7 +463,108 @@ func TestDeleteRuleProperties(t *testing.T) {
...
+					respProject.Type.String() == "CUSTOM" &&
+					respProject.Status == storage.EditsPending.String()
+			},
+			createProjectAndRulesGen,
+		))
+
+	properties.TestingRun(t)
+}
+
+func TestListProjectPropertiesMixingStagedAndApplied(t *testing.T) {
```
msorens: This is the main work horse test that makes some staged, some applied, and some no-rule projects and checks those rule statuses from `ListProjects`. Sample output of this function was shown in the PR preamble.

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -460,7 +463,108 @@ func TestDeleteRuleProperties(t *testing.T) {
...
+	ctx := context.Background()
+	cl, testDB, _, _, seed := testhelpers.SetupProjectsAndRulesWithDB(t)
+	properties := getGopterParams(seed)
+
+	properties.Property("projects with staged, applied, or no rules are reported",
+		prop.ForAll(
+			func(reqs []projectAndRuleReq) bool {
```
msorens: How to read a property-based test. Step 1: look for the `func` line like this. Gopter feeds a set of test cases, one at a time, in the shape of the argument to this function. The shape was designed by me to specifically to test `ListProjects`. One test case, then, is a list of objects, each of which contains a single project and some rules (or no rules) to go with it.

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -460,7 +463,108 @@ func TestDeleteRuleProperties(t *testing.T) {
...
+						} else {
+							if p.Status != storage.EditsPending.String() {
+								return false
+							}
+						}
+					}
+				}
+				return true
+			},
+			createProjectsAndRulesGen,
```
msorens: How to read a property-based test. Step 2: And how does it know how to generate something of the right shape? The second argument to the `func`, which is way down here. 😁

components/authz-service/storage/postgres/migration/sql/68_add_status_to_projects.up.sql
```diff
@@ -0,0 +1,40 @@
...
+)
+SELECT json_build_object(
+  'id', p.id,
+  'name', p.name,
+  'type', p.type,
+  'status', 
+  CASE WHEN staged.count IS NULL AND applied.count IS NULL THEN 'no-rules'
+       WHEN staged.count IS NULL THEN 'applied'
+       ELSE 'edits-pending'
+  END
```
msorens: The changes to `query_project` are all centered around adding this computed `status` field to the output object. All the work is done here in the DB, saving a lot of Go-based machinations.

components/authz-service/storage/postgres/migration/sql/68_add_status_to_projects.up.sql
```diff
@@ -0,0 +1,40 @@
...
+)
+FROM staged
+FULL JOIN applied
+ON staged.project_id = applied.project_id
```
msorens: Full join because we want to find staged or applied or both.

components/authz-service/storage/postgres/migration/sql/68_add_status_to_projects.up.sql
```diff
@@ -0,0 +1,40 @@
...
+RIGHT JOIN iam_projects p
+ON p.db_id = COALESCE(staged.project_id, applied.project_id)
```
msorens: Then we need to correlate that back to the projects table itself to be able to glue the project fields together with the new computed field.

components/authz-service/storage/v2/memstore/memstore.go
```diff
@@ -10,7 +10,6 @@ import (
 	constants_v2 "github.com/chef/automate/components/authz-service/constants/v2"
 	storage_errors "github.com/chef/automate/components/authz-service/storage"
 	storage "github.com/chef/automate/components/authz-service/storage/v2"
-	v2 "github.com/chef/automate/components/authz-service/storage/v2"
```
msorens: PR Bonus: Remove duplicate.

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -4801,159 +4801,80 @@ func TestListProjects(t *testing.T) {
 			require.NoError(t, err)
 			assert.Empty(t, ps)
 		}},
-		{"when two projects (custom and chef-managed) exist, returns them", func(t *testing.T) {
+		{"when multiple projects exist, filter based on projects lists", func(t *testing.T) {
 			ctx := context.Background()
-			insertTestProject(t, db, "foo", "my foo project", storage.ChefManaged)
-			insertTestProject(t, db, "bar", "my bar project", storage.Custom)
+			p1 := insertTestProject(t, db, "foo", "my foo project", storage.ChefManaged)
+			p2 := insertTestProject(t, db, "bar", "my bar project", storage.Custom)
```
msorens: PR Bonus: use the output of the helper function here to eliminate a lot of code in declaring the expectations just below.

components/authz-service/storage/v2/postgres/postgres_test.go
```diff
@@ -4801,159 +4801,80 @@ func TestListProjects(t *testing.T) {
...
-					Name: "my bar project",
-					Type: storage.Custom,
-				},
-			}
+			expectedProjects := []*storage.Project{&p1, &p2, &p3}
 
 			assert.ElementsMatch(t, expectedProjects, ps)
 		}},
-		{"when multiple projects exist, returns everything when no project filter is specified (v2.0 case)", func(t *testing.T) {
+		{"returns rule status for projects for none, staged, and applied individually", func(t *testing.T) {
```
msorens: Two new DB unit tests start here.

components/automate-ui/src/app/entities/rules/rule.reducer.ts
```diff
@@ -88,8 +92,10 @@ export function ruleEntityReducer(
       return set(GET_ALL_STATUS, EntityStatus.loading, state);
 
     case RuleActionTypes.GET_ALL_SUCCESS:
-      return set(GET_ALL_STATUS, EntityStatus.loadingSuccess,
-        ruleEntityAdapter.addAll(action.payload.rules, state));
+      return pipe(
+        set(GET_ALL_STATUS, EntityStatus.loadingSuccess),
+        set(PROJECT_RULES_STATUS, action.payload.status))
```
msorens: All the back-end work was really to just be able to add the project rules status for each project to the ngrx/store right here.

components/automate-ui/src/app/entities/rules/rule.requests.ts
```diff
@@ -8,10 +8,6 @@ import {
   RulesSuccessPayload, RuleSuccessPayload
 } from './rule.actions';
 
-export interface RulesResponse {
-  rules: Rule[];
-}
```
msorens: PR Bonus: remove unused interface.

components/automate-ui/src/app/pages/project/list/project-list.component.html
```diff
@@ -45,7 +45,7 @@
                   <a [routerLink]="['/settings/projects', project.id]">{{ project.name }}</a>
                 </chef-td>
                 <chef-td>{{ project.id }}</chef-td>
-                <chef-td>Applied</chef-td>
+                <chef-td>{{ getRulesStatus(project) }}</chef-td>
```
msorens: And, finally, use live code instead of a dummy value! 🎉

components/automate-ui/src/app/entities/rules/rule.reducer.ts
```diff
@@ -69,15 +71,17 @@ export const ruleAttributes: RuleTypeMappedObject = {
...
-  getAttributes: ruleAttributes,
-  getAllStatus: EntityStatus.notLoaded,
-  getStatus: EntityStatus.notLoaded,
-  createStatus: EntityStatus.notLoaded,
-  createError: null,
-  deleteStatus: EntityStatus.notLoaded,
-  updateStatus: EntityStatus.notLoaded
-});
+export const RuleEntityInitialState: RuleEntityState =
+  ruleEntityAdapter.getInitialState(<RuleEntityState>{
```
msorens: Now here is a TypeScript anomaly, which I unearthed thanks to @srenatus' question. Originally L83 read
```
projectRulesStatus: <ProjectStatus>'PROJECT_RULES_STATUS_UNSET'
```
And he wondered why. Answer: Because it gave me a compiler error without it, because `getInitialState` takes an untyped object. So I added `<RuleEntityState>` above as a more general solution...
But later I checked further. All of our other reducers using `getInitialState` do **not** add the type assertion, but if I were to, for instance, misspell `getAllStatus` it would still flag as an error in VSCode, thus recognizing it was in fact an _xyz_`EntityState`. What's different here? The `'PROJECT_RULES_STATUS_UNSET'` string value that is both (a) a supertype of the allowed `ProjectStatus` type and (b) not null. Not satisfying those seems to be why I need the type assertion.

components/authz-service/server/v2/project.go
```diff
@@ -113,7 +113,7 @@ func (s *ProjectState) GetProject(ctx context.Context,
 
 func (s *ProjectState) CreateProject(ctx context.Context,
 	req *api.CreateProjectReq) (*api.CreateProjectResp, error) {
-	p, err := storage.NewProject(req.Id, req.Name, storage.Custom)
+	p, err := storage.NewProject(req.Id, req.Name, storage.Custom, 0)
```
srenatus: :x: Can we use one of [the enum constant](https://github.com/chef/automate/blob/76a491ab2d2a7b7d8829e06a00d1b0593a7df803/components/authz-service/storage/v2/project_rules_status.go#L7-L20) here?
```suggestion
	p, err := storage.NewProject(req.Id, req.Name, storage.Custom, storage.RulesStatusError)
```
`0` is `RulesStatusError`. Makes me wonder if `NoRules` wouldn't me more appropriate...

components/authz-service/storage/v2/project_test.go
```diff
@@ -36,7 +36,8 @@ func TestNewProject(t *testing.T) {
 		},
 	} {
 		t.Run(name, func(t *testing.T) {
-			if _, err := storage.NewProject(tc.id, tc.name, tc.typeVal); tc.expectErr != (err != nil) {
+			if _, err := storage.NewProject(
+				tc.id, tc.name, tc.typeVal, 0); tc.expectErr != (err != nil) {
```
srenatus: Funky indentation. There's no strong reason to worry about the scope here (what you'd usually prefer this constract for), so let's keep it simple
```go
_, err :=  storage.NewProject(tc.id, tc.name, tc.typeVal, 0); // falso: fix me! What's 0?
if tc.expectErr != (err != nil) {
	t.Fail()
}
```

components/automate-ui/src/app/entities/rules/rule.reducer.ts
srenatus: Interesting. And weird. Shall we contact some upstream?

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -194,4 +210,13 @@ export class ProjectListComponent implements OnInit {
...
     this.closeConfirmApplyStopModal();
   }
+
+  getRulesStatus(project: Project): string {
+    switch (project.status) {
+      case 'NO_RULES': return 'No rules';
+      case 'EDITS_PENDING': return 'Edits pending';
+      case 'RULES_APPLIED': return 'Applied';
+      default: return '';
+    }
```
srenatus: > My preference is for functional-style code, which requires a ternary as given rather than a switch. However, sounds like your preference is strong, so I am happy to oblige.

Oh I prefer functional code, too. But let's not overdo it, nesting ternaries is something I'd rather avoid.  This is more straightforward, thanks for obliging 😉

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -43,6 +43,9 @@ export class ProjectListComponent implements OnInit {
   public applyRulesButtonText$: Observable<string>;
   public ApplyRulesStatusState = ApplyRulesStatusState;
 
+  private POLLING_INTERVAL_IN_SECONDS = 10;
```
srenatus: For the projects filter's actions, we do the polling in effects, `components/automate-ui/src/app/entities/projects/project.effects.ts` -- I'd prefer we sticked to some convention 😃Any reason this approach is blocked from us here?

💡 We only want to poll while on this page. That's probably the difference. OK then.

msorens: Asked and answered. Exactly so. 😁 😁

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -82,8 +85,21 @@ export class ProjectListComponent implements OnInit {
   }
 
   ngOnInit(): void {
-    this.store.dispatch(new GetProjects());
+    // Get status now; periodic checks are already being done globally so no need to poll further
     this.projects.getApplyRulesStatus();
+
+    // Get projects status now and periodically while on this page
+    this.store.dispatch(new GetProjects());
+    observableInterval(1000 * this.POLLING_INTERVAL_IN_SECONDS).pipe(
```
bcmdarroch: Not sure about the UX of this polling. I see the project list blip every 10 seconds, but it doesn't react at the same time that the `Update Button` becomes re-enabled. Instead of polling why don't we call `GetProjects()` as soon as the update is completed? 

This might require rethinking our current `apply-rules` polling. It'd be great if we only started polling when a project update was in progress.

msorens: A few things here. 
1. In a typical scenario, the projects list will go from "edits pending" to "applied" on the first poll, because that strictly reflects the DB update that we do almost immediately. The updated button, on the other hand, will be otherwise occupied for hours. So your observation that they don't occur simultaneously will almost always be true. Also, it would be very late, then, to wait until the update is complete before calling `GetProjects`.
2. When the update is complete will have a direct affect on the Project Update Status, not the Ingest Rules column, and that will be in the next PR.
3. Regarding polling, it is necessary not for the actions of this user but because of potential actions of other users. Whenever we are on the projects list page, it behooves us to know if someone else has queued some project changes and we will then see those posted every 10 seconds. That is also quite independent of whether we have started an update or not, so the polling needs to be active when simply on this page, unconditionally.

components/authz-service/storage/postgres/migration/sql/68_add_status_to_projects.up.sql
msorens: Not necessarily faster, but I believe it is more suitable here for two reasons:
1. We're already doing part of the work here in determining the result of `ListProjects`; it makes sense to do the rest of it here, too.
2. Go is rather verbose at times due to its lack of simple things common to many high-level languages (e.g., an `array-contains()` function). I believe the number of lines of code to generate this output in Go would be greater.

PR #1251, opened 2019-08-15
ci: show globs, not files

srenatus:
:bulb: @stevendanna's idea. Hoping to strike a middleground between information overload and usefulness.


stevendanna:
I've added a commit to this PR to also fix a small issue in which we weren't printing out things that were built only because one of their hab dependencies were also rebuilt.
Screen Shot 2019-08-15 at 5 43 53 PM

srenatus:
@stevendanna my bad! thank you 🎊

Issue #1245, opened 2019-08-14
HTTP/2 Security Vulnerabilities


skylerto:
Security vulnerabilities were reported with HTTP/2 implementations, gRPC being one of them. I know that this software makes extensive use of gRPC for internal service communications and wanted to bring it up as it could effect some of our clients. Looking for feedback on how the automate team might be addressing these vulnerabilities.

https://www.kb.cert.org/vuls/id/605641/

## Describe the bug

Exploiting gRPC connections can be exploited with denial-of-service attacks causing unexpected resource consumption and/or service failures. 

## To Reproduce

Implement any of the issues brought up in this document: https://www.kb.cert.org/vuls/id/605641/

## Expected behavior

Services should mitigate DOS attacks.

## Screenshots

N/A

## Versions (please complete the following information):

N/A

## Additional context

Following reports:

- https://github.com/Netflix/security-bulletins/blob/master/advisories/third-party/2019-002.md
- https://www.kb.cert.org/vuls/id/605641/

srenatus:
Thanks for opening this issue! I believe we've addressed these with #1237. 🤔 Haven't we? 😃

srenatus:
That last comment from me was re: the internal gRPC APIs. The other HTTP2 APIs, both internal and external, are being worked on:

- [X] stdlib `net/http` https://github.com/habitat-sh/core-plans/pull/2971 (needs promotion)
- [ ] nginx: https://github.com/habitat-sh/core-plans/pull/2970
- [ ] dex: https://github.com/habitat-sh/core-plans/pull/2972

srenatus:
☝️ sorry for that, we've crossed wires when I wanted to fix the typo 😄

skylerto:
@srenatus thanks for the quick reply && typo fix! 

Looks like 3 of the 8 reported CVEs were taken care of as part of the [1.23.0 grpc-go release](https://github.com/grpc/grpc-go/releases/tag/v1.23.0). Those 3 being:

- CVE-2019-9512 (Ping Flood)
- CVE-2019-9514 (Reset Flood)
- CVE-2019-9515 (Settings Flood)

These are covered at the proxy/load balancer (nginx):

- CVE-2019-9511 (Data Dribble)
- CVE-2019-9513 (Resource Loop)
- CVE-2019-9516 (0-Length Headers Leak) 

I'm not sure if the other CVEs not covered are incarnations of the 3 that were fixed these are still not discussed anywhere:

[...]

PR #1254, opened 2019-08-15
trigger rebuilds for core/nginx/1.17.3

srenatus:
For #1245.


srenatus:
@stevendanna what do you think? to pin or not to pin? 🤔 I'm not happy about this, but it's an easy rebuild 😄

stevendanna:
I suppose we could merge this and then immediately merge something that unpins us :D

Since automate-workflow-web is building again, another option would be to trigger a full rebuild in buildkite.  Between the grpc update, the golang update, and I'm sure some others, my guess is we are gonna be close to a full rebuild for the next release anyway.

srenatus:
@stevendanna I agree, it's a bit silly to avoid the full rebuild at this point. Can kick it off tomorrow morning to have the US night for dealing with unexpected issues...

srenatus:
✔️ we've rebuilt all hartifacts using the pipeline.

PR #1255, opened 2019-08-15
lib/cereal: Use in-process notifications to avoid delays

stevendanna:
This change allows us to wake up relevant go-routines in the local
process when certain events happens:

- When we enqueue tasks, we wake up the task polling goroutine
- When we finish a task, we wake up the workflow polling goroutine

For each registered task executor we have a "Poller" which wakes up at an interval
or on notification and starts a task worker.  An additional worker will be spawned as 
long as there is more work to do.  When the worker finds that there is no more work to
do it exits.

Note to reviewers: We probably want to think about this worker management strategy.
We could make it more responsive by growing the queue

### DIFF
--- a/lib/cereal/cereal.go
+++ b/lib/cereal/cereal.go
@@ -30,6 +30,11 @@ var (
 	ErrTaskLost                 = errors.New("task lost before reporting its status")
 )
 
+const (
+	defaultTaskPollInterval     = 2 * time.Second
+	defaultWorkflowPollInterval = 2 * time.Second
+)
+
 // Schedule represents a recurring workflow.
 // TODO(jaym): we should wrap this in the workflow package and provide a getter
 // for the parameters
@@ -424,21 +429,20 @@ func (w waywardWorkflowList) Filter(workflowNames []string) []string {
 	return out
 }
 
-const defaultTaskPollInterval = 2 * time.Second
-
 // Manager is responsible for for calling WorkflowExecutors and
 // TaskExecutors when they need to be processed, along with managing
 // the scheduling of workflows.
 type Manager struct {
 	workflowExecutors map[string]WorkflowExecutor
-	taskExecutors     map[string]registeredExecutor
+	taskExecutors     map[string]*registeredExecutor
 	waywardWorkflows  waywardWorkflowList
 	workflowScheduler *WorkflowScheduler
 	backend           backend.Driver
 	cancel            context.CancelFunc
 	wg                sync.WaitGroup
 
-	taskPollInterval time.Duration
+	workflowWakeupChan chan struct{}
+	taskPollInterval   time.Duration
 }
 
 // ManagerOpt is an option that can be passed to NewManager.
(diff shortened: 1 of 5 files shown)


lib/cereal/cereal.go
```diff
@@ -497,19 +503,188 @@ type TaskExecutorOpts struct {
...
+	logctx := logrus.WithField("task_name", r.name)
+	logctx.Infof("Starting task poller")
+	for {
+		select {
+		case <-ctx.Done():
+			logctx.WithField("active_workers", r.activeWorkers).Info("Waiting for all task workers to exit")
+			r.wg.Wait()
+			logctx.Info("Exiting task poller")
+			return
+		case <-time.After(taskPollInterval):
```
jaym: i think this leak might matter depending on how much wakeupChan gets a message

lib/cereal/cereal.go
```diff
@@ -497,19 +503,188 @@ type TaskExecutorOpts struct {
...
+	if !r.IncActiveWorkers() {
+		logctx.WithFields(logrus.Fields{
+			"max_workers":    r.maxWorkers,
+			"active_workers": r.activeWorkers,
+		}).Warn("Maximum task workers already started")
+		return
+	}
+
+	go func() {
+		r.wg.Add(1)
```
jaym: i think this needs to happen before creating to goroutine

stevendanna: Good catch. this could race with a concurrently running Wait()

lib/cereal/cereal.go
stevendanna: Correct, to avoid the leak we'll need to use a `time.Timer`, which is a bit of a pain to use correctly, but I can wire it up.

lib/cereal/cereal.go
```diff
@@ -497,19 +503,195 @@ type TaskExecutorOpts struct {
...
+				}
+				logctx.Debug("Worker shutting down")
+				r.DecActiveWorkers()
+				r.wg.Done()
+				return
+			}
+			logctx.Debugf("Dequeued task %s", t.Name)
+
+			if !startedNext {
+				r.startTaskWorker(ctx, b, workflowWakeupFun)
```
jaym: i think i've convinced myself doing this is ok. I initially thought concurrently adding to wg would be problematic, but this goroutine already holds a count which would prevent the wg from going to 0 as a new task worker is starting

lib/cereal/cereal.go
```diff
@@ -666,84 +848,53 @@ func (m *Manager) EnqueueWorkflow(ctx context.Context, workflowName string,
...
+}
 
-		t, taskCompleter, err := m.backend.DequeueTask(ctx, taskName)
-		if err != nil {
-			if err != ErrNoTasks {
-				logctx.WithError(err).Error("failed to dequeue task")
-			}
-			time.Sleep(m.taskPollInterval)
-			continue
+func (m *Manager) WakeupTaskPollersByRequests(taskRequests []enqueueTaskRequest) {
```
jaym: this probably shouldn't be public?

lib/cereal/cereal.go
```diff
@@ -872,6 +1029,7 @@ func (m *Manager) processWorkflow(ctx context.Context, workflowNames []string) b
 				return true
 			}
 		}
+		m.WakeupTaskPollersByRequests(decision.tasks)
```
jaym: doesn't this need to happen after the commit?

PR #1257, opened 2019-08-15
Dan/convert to denormalized schema

danielsdeleo:

Converts applications service to run from denormalized schema.

### :+1: Definition of Done

- [ ] decide if/how to handle migrations
- [ ] do we like computing the artificial service group IDs in the application or would we rather do it in the database?
- [x] verify the UI works
- [ ] drop the old tables

build applications-service and automate-gateway, run `applications_populate_database`, things should work as normal.

To test the migration, start a clean studio, remove any packages from `results/`, and `start_all_services`. Then `applications_populate_database` to make some data, verify it's there in the UI. Then build automate-gateway and applications-service, wait a minute for deployment service to deploy them. Then visit the UI and see that the services that were loaded in the previous version are still there. Then run `applications_populate_database` and see that the new services it generates correctly combine with the service groups from before the upgrade.


components/applications-service/pkg/storage/postgres/db.go
```diff
@@ -129,6 +129,7 @@ func (s *supervisor) NeedUpdate() bool {
 // serviceGroup struct is the representation of the service_group table inside the db
 type serviceGroup struct {
 	ID           int32     `db:"id"`
+	HashId       string    `db:"hash_id"`
```
danielsdeleo: this is leftover from something else, but should look at if I can blow away any of these struct types

components/applications-service/pkg/storage/postgres/hab_event.go
```diff
@@ -288,8 +224,13 @@ func (db *Postgres) upsertServiceAndMetadata(
 		updateStrategy = storage.NoneStrategy.String()
 	} else {
 		channel = svcMetadata.GetUpdateConfig().Channel
-		updateStrategy = svcMetadata.GetUpdateConfig().Strategy.String()
+		updateStrategy = storage.HabitatUpdateStrategyToStorageFormat(svcMetadata.UpdateConfig.GetStrategy()).String()
 	}
+
+	serviceGroupUniques := fmt.Sprintf("%s:%s:%s", svcMetadata.GetServiceGroup(), eventMetadata.GetApplication(), eventMetadata.GetEnvironment())
```
danielsdeleo: 🤔  Any ideas on how we might do better here? we could delimit with a different char that is more likely to be unused (like `;`)? Also, do we need to actually hash this value?

api/external/applications/applications.proto
```diff
@@ -94,7 +94,7 @@ message ServicesReq {
 }
 
 message ServicesBySGReq {
-  int32 service_group_id = 1;
+  string service_group_id = 1;
```
danielsdeleo: this is usually very very bad but we're beta so I'd rather do this than carry the cruft forever

components/applications-service/pkg/storage/postgres/schema/sql/12_artificial_ids_for_service_groups.up.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE service_full ADD COLUMN service_group_id text;
```
danielsdeleo: To explain what's going on here a bit:

When we switch to the denormalized table, we no longer have artificial integer keys for service group rows since we don't have a service group table. This is a problem because we expose the service group IDs to the UI, which uses them as IDs for the rows in the data layer, and also puts them in requests back to the API when it fetches the services for the selected service group. So we want to re-introduce the concept of an artificial service group ID which serves the same purposes. I put some questions for discussion in the `hab_event.go` diff.

components/applications-service/pkg/storage/postgres/hab_event.go
kmacgugan: Unicode! No I kid. Let's not do that. I am not sure if the colon will need to be escaped when it is in a url which is kind of gross, maybe actual hash would be better.

PR #1258, opened 2019-08-15
WIP add the generator report samples template files  for chef-load

rickmarry:
Add the sample generator chef-load template files back in.  
These sample reports are still needed in order for chef-load to function properly.  The files have been moved under ./components/compliance-service/test_data/generator.. 

### :+1: Definition of Done
from hab, running:
`
chef_load_compliance_scans
`
causes inspec scans to be ingested and therefore accessible thru compliance reporting apis

- [ ] Tests added/updated?
- [ ] Docs added/updated?


phiggins:
If this is the same data as in `chef-load`, why does it need to be checked in here also? https://github.com/chef/chef-load/tree/master/sample-data/inspec-reports

rickmarry:
> If this is the same data as in `chef-load`, why does it need to be checked in here also? https://github.com/chef/chef-load/tree/master/sample-data/inspec-reports
Actually, I think it would be better to let these files live only in the chef-load repo.  We just need to set it so that when chef-load is installed vi hab in studio for automate, sample files are also installed so that they may be referenced and used instead of these.

phiggins:
I don't have strong feelings about whether the files live in this repo or chef-load, it seems bad for them to be in both places though.

vjeffrey:
in slack we discussed using the `prepare_chef_load` command to pull the needed files down from the repo and then adjusting the toml to point to wherever those are on disk

vjeffrey:
i'm going to close this for now

Issue #809, opened 2019-07-09
Teams: Add projects form input to create and update


susanev:
_Note: This story is dependent on #794 happening first._

## Overview

This task includes adding the projects form input to the create team modal and to the Details tab on team detail pages. This all needs to be feature flagged to iam beta 2.1.

## Important Notes

* On create the projects to choose from should be filtered by any filters the user has applied in the global projects filter (we will revisit this and improve the ux in the future)
* We will revisit the case when a user has access to zero projects in the future
* If the user tries to update a resource's projects and that resource has projects the user does not have access to, then those projects will still be shown and the user wont get an error until trying to update the team
* Maggie has developed a component that we can use for this story - components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.html (see [a2 pr #5485](https://github.com/chef/a2/pull/5485))

## Design Details

[Zeplin](https://app.zeplin.io/project/5ba808703eb54372d860c737?seid=5ca78c9bfe244b013f56f0f6)

* Add a Projects column to the Teams list view
    * When a team is not assigned to any projects display (unassigned)
    * When a team is assigned to exactly one project display that project's name
    * When a team is assigned to two or more projects display <n> projects
* On the Create Team modal add the Projects multi-select form input below the ID input
    * This input is optional, so once a name is provided the Create Team button should be enabled
* On the Details tab on team detail pages...
    * Add the Projects label, and content in the top header, this value should match whatever is shown in the list view
    * Add the Projects multi-select form input below the name input
    * There should be 24px of space between the projects input and the Save button

_There is more past discussion over on Jira._

## Sub-Tasks
- [x] Teams List View and Create Team Modal #1182 @tylercloke 
- [x] Team Details View #1260 @tylercloke

Issue #1199, opened 2019-08-08
NGRX: using addOne is not properly replacing out of date objects


tylercloke:
## TLDR

Completely replace our usage of `addOne` in ngrx/store with `upsertOne`.

## Describe the bug

We are currently using `<entityAdapter>.addOne` in the create reducer functions of our NGRX store. Normally, if a user creates an object with an ID that already exists in the store, the API will return an HTTP error code because the object already exists. However, if outside of that specific browser session, that object gets deleted and the user creates the object via the UI, `addOne` will have the undesired behavior of "appending" to the previous object with the same ID in the store instead of overwriting it. Furthermore, when you re-fetch an object, if you don't completely replace the instance of the object, `addOne` will not trigger ngrx/store change detection.

This results in the projects list getting concatenated between the stale object in the store and the newly created, updated, or refetched one. All of our APIs are last write wins so we can trust the response from the POST if it did not result in an error. `<entityAdapter>.upsertOne` appears to have the behavior we want. Let's update all of our create object reducers to use that.

Also should investigate updates as a similar situation might be occurring.

Additional reading:

https://blog.angular-university.io/ngrx-entity/

## To Reproduce

Load the browser with a team with a single project.
Delete that team via the API or a separate web browser / session.
The team should still be in the first browsers store. Create a team with the same ID with different projects. The resulting team should have the original team's projects as well as the new ones (incorrect behavior).

## Expected behavior

Only the projects from the newly created team should be in the store.

srenatus:
Took the liberty of rewording the title. It's not NGRX's fault, it's ours 😉

susanev:
this work has been deprioritized, so closing.

PR #1260, opened 2019-08-15
[automate-809] Update projects for a team

tylercloke:

Added project update to teams. In the dropdown, all projects a team is a member of will start out pre-checked and any additional projects in the project filter will also be avaliable.

Simplified the state that was contained in the projects-dropdown and updated the object-create-modal to reflect those changes.

Cleaned up the cypress tests.

Start the UI in the standard way, create some projects from the UI, create a team, modify its projects.

### Resources
- https://github.com/chef/automate/issues/809
- [Zeplin Designs](https://app.zeplin.io/project/5ba808703eb54372d860c737?seid=5ca78c9bfe244b013f56f0f6)

- [x] Tests added/updated?
- [ ] Docs added/updated?

Screen Shot 2019-08-15 at 2 42 10 PM

Screen Shot 2019-08-15 at 2 42 19 PM

### DIFF
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
@@ -7,7 +7,8 @@ import { hasIn } from 'lodash/fp';
 import { IdMapper } from 'app/helpers/auth/id-mapper';
 import { Project } from 'app/entities/projects/project.model';
 import {
-  ProjectChecked
+  ProjectChecked,
+  ProjectCheckedMap
 } from 'app/components/projects-dropdown/projects-dropdown.component';
 
 @Component({
@@ -27,7 +28,7 @@ export class CreateObjectModalComponent implements OnInit, OnChanges {
   @Output() close = new EventEmitter();
   @Output() createClicked = new EventEmitter<Project[]>();
 
-  public projects: { [id: string]: ProjectChecked } = {};
+  public projects: ProjectCheckedMap = {};
 
   // Whether the edit ID form is open or not.
   public modifyID = false;
@@ -45,13 +46,14 @@ export class CreateObjectModalComponent implements OnInit, OnChanges {
     // if a new list of projects to populate dropdown with is passed in we update the dropdown
     const checked = false;
     if (hasIn('assignableProjects.currentValue', changes)) {
+      this.projects = {};
       changes.assignableProjects.currentValue.forEach((proj: Project) =>
         this.projects[proj.id] = { ...proj, checked });
     }
   }
 
   onProjectChecked(project: ProjectChecked): void {
-    this.projects[project.id] = project;
+    this.projects[project.id].checked = project.checked;
   }
 
   dropdownDisabled(): boolean {
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
@@ -41,7 +41,7 @@ export class CreateObjectModalComponent implements OnInit, OnChanges {
     });
   }
 
-  ngOnChanges(changes: SimpleChanges) {
+  ngOnChanges(changes: SimpleChanges): void {
     // if a new list of projects to populate dropdown with is passed in we update the dropdown
     const checked = false;
     if (hasIn('assignableProjects.currentValue', changes)) {
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
@@ -1,15 +1,21 @@
-import { Component, EventEmitter, Input, Output, OnInit } from '@angular/core';
+import {
+  Component, EventEmitter, Input, Output, OnInit, OnChanges, SimpleChanges
+} from '@angular/core';
 import { FormGroup } from '@angular/forms';
+import { hasIn } from 'lodash/fp';
 
 import { IdMapper } from 'app/helpers/auth/id-mapper';
 import { Project } from 'app/entities/projects/project.model';
+import {
+  ProjectChecked, projectCheckedFromProject
+} from 'app/components/projects-dropdown/projects-dropdown.component';
 
 @Component({
   selector: 'app-create-object-modal',
   templateUrl: './create-object-modal.component.html',
   styleUrls: ['./create-object-modal.component.scss']
 })
-export class CreateObjectModalComponent implements OnInit {
+export class CreateObjectModalComponent implements OnInit, OnChanges {
   @Input() visible = false;
   @Input() creating = false;
   @Input() objectNoun: string;
@@ -21,10 +27,11 @@ export class CreateObjectModalComponent implements OnInit {
   @Output() close = new EventEmitter();
   @Output() createClicked = new EventEmitter<Project[]>();
 
+  public projects: { [id: string]: ProjectChecked } = {};
+
   // Whether the edit ID form is open or not.
   public modifyID = false;
   public conflictError = false;
-  private projectsSelected: Project[] = [];
 
   ngOnInit(): void {
     this.conflictErrorEvent.subscribe((isConflict: boolean) => {
@@ -32,6 +39,22 @@ export class CreateObjectModalComponent implements OnInit {
       // Open the ID input on conflict so user can resolve it.
       this.modifyID = true;
     });
+  }
+
+  ngOnChanges(changes: SimpleChanges) {
+    // if a new list of projects to populate dropdown with is passed in we update the dropdown
+    if (hasIn('assignableProjects.currentValue', changes)) {
+      changes.assignableProjects.currentValue.forEach(proj =>
+        this.projects[proj.id] = projectCheckedFromProject(proj, false));
+    }
+  }
+
+  onProjectChecked(project: ProjectChecked): void {
+    this.projects[project.id] = project;
+  }
+
+  dropdownDisabled(): boolean {
+    return Object.values(this.projects).length === 0;
   }
 
   public handleNameInput(event: KeyboardEvent): void {
@@ -55,14 +78,10 @@ export class CreateObjectModalComponent implements OnInit {
   }
 
   createObject(): void {
-    this.createClicked.emit(this.projectsSelected);
+    this.createClicked.emit(Object.values(this.projects).filter(p => p.checked));
   }
 
   private isNavigationKey(event: KeyboardEvent): boolean {
     return event.key === 'Shift' || event.key === 'Tab';
-  }
-
-  projectsDropdownAction(projects: Project[]): void {
-    this.projectsSelected = projects;
   }
 }
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
@@ -45,7 +45,7 @@ export class CreateObjectModalComponent implements OnInit, OnChanges {
     // if a new list of projects to populate dropdown with is passed in we update the dropdown
     const checked = false;
     if (hasIn('assignableProjects.currentValue', changes)) {
-      changes.assignableProjects.currentValue.forEach(proj =>
+      changes.assignableProjects.currentValue.forEach((proj: Project) =>
         this.projects[proj.id] = { ...proj, checked });
     }
   }
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
@@ -1,5 +1,6 @@
 import { Component, EventEmitter, Input, Output, OnInit } from '@angular/core';
 import { FormGroup } from '@angular/forms';
+
 import { IdMapper } from 'app/helpers/auth/id-mapper';
 import { Project } from 'app/entities/projects/project.model';
 
@@ -61,7 +62,7 @@ export class CreateObjectModalComponent implements OnInit {
     return event.key === 'Shift' || event.key === 'Tab';
   }
 
-  projectsDropdownAction(projects) {
+  projectsDropdownAction(projects: Project[]): void {
     this.projectsSelected = projects;
   }
 }
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
@@ -7,7 +7,7 @@ import { hasIn } from 'lodash/fp';
 import { IdMapper } from 'app/helpers/auth/id-mapper';
 import { Project } from 'app/entities/projects/project.model';
 import {
-  ProjectChecked, projectCheckedFromProject
+  ProjectChecked
 } from 'app/components/projects-dropdown/projects-dropdown.component';
 
 @Component({
@@ -43,9 +43,10 @@ export class CreateObjectModalComponent implements OnInit, OnChanges {
 
   ngOnChanges(changes: SimpleChanges) {
     // if a new list of projects to populate dropdown with is passed in we update the dropdown
+    const checked = false;
     if (hasIn('assignableProjects.currentValue', changes)) {
       changes.assignableProjects.currentValue.forEach(proj =>
-        this.projects[proj.id] = projectCheckedFromProject(proj, false));
+        this.projects[proj.id] = { ...proj, checked });
     }
   }
 
--- a/components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
+++ b/components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
@@ -1,10 +1,8 @@
 import {
-  Component, EventEmitter, OnInit, Input, Output
+  Component, EventEmitter, Input, Output
 } from '@angular/core';
 
-import { orderBy } from 'lodash/fp';
-
-import { Project, ProjectConstants } from 'app/entities/projects/project.model';
+import { ProjectConstants, Project } from 'app/entities/projects/project.model';
 
 const { UNASSIGNED_PROJECT_ID } = ProjectConstants;
 
@@ -13,11 +11,10 @@ const { UNASSIGNED_PROJECT_ID } = ProjectConstants;
   templateUrl: './projects-dropdown.component.html',
   styleUrls: ['./projects-dropdown.component.scss']
 })
-export class ProjectsDropdownComponent implements OnInit {
+export class ProjectsDropdownComponent {
 
-  // The array of projects that you pass into the component. This array gets updated when a user
-  // makes changes and is emitted to the parent component.
-  @Input() projects:  Array<Project> = [];
+  // The array of projects that you pass into the component.
+  @Input() projects: Array<Project> = [];
 
   // Setting required to true means the dropdown will show an error if a user tries to close the
   // dropdown but hasn't selected any projects.
@@ -34,13 +31,28 @@ export class ProjectsDropdownComponent implements OnInit {
   label = UNASSIGNED_PROJECT_ID;
 
   // Map of projects currently selected
-  selectedProjects = {};
+  selectedProjects: { [id: string]: Project } = {};
 
-  constructor() {
-    this.projects = this.sortProjectsByName(this.projects);
+  constructor() { }
+
+  // If you want to mark some projects as pre-selected, use @ViewChild.
+  // See team-details.component.ts for an example.
+  public updateSelectedProjects(project: Project, enabled: boolean): void {
+    console.log('update');
+    console.log(project);
+    console.log(enabled);
+    console.log(this.selectedProjects);
+    if (enabled) {
+      this.selectedProjects[project.id] = project;
+    } else {
+      delete this.selectedProjects[project.id];
+    }
+    this.updateLabel();
   }
 
-  ngOnInit() {}
+  projectSelected(project: Project): boolean {
+    return this.selectedProjects[project.id] !== undefined;
+  }
 
   toggleDropdown(event: MouseEvent): void {
     event.stopPropagation();
@@ -64,7 +76,13 @@ export class ProjectsDropdownComponent implements OnInit {
     }
 
     if (this.active) {
-      this.onSelection.emit(Object.values(this.selectedProjects));
+      const result = <Project[]>Object.values(this.selectedProjects);
+      const resultProjects = result.map(p => <Project>{
+        id: p.id,
+        name: p.name,
+        type: p.type
+      });
+      this.onSelection.emit(resultProjects);
       this.active = false;
     }
   }
@@ -91,16 +109,7 @@ export class ProjectsDropdownComponent implements OnInit {
     }
   }
 
-  sortProjectsByName(projects: Array<Project>): Array<Project> {
-    return orderBy(['name'], ['asc'], projects);
-  }
-
-  private updateSelectedProjects(project: Project, enabled: boolean): void {
-    if (enabled) {
-      this.selectedProjects[project.id] = project;
-    } else {
-      delete this.selectedProjects[project.id];
-    }
+  private updateLabel(): void {
     const selectedLen = Object.keys(this.selectedProjects).length;
     switch (selectedLen) {
       case 1: {
--- a/components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
+++ b/components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
@@ -6,6 +6,9 @@ import { ProjectConstants, Project } from 'app/entities/projects/project.model';
 
 const { UNASSIGNED_PROJECT_ID } = ProjectConstants;
 
+// Extend the project model with the checked field.
+// This represents whether the project's checkbox is unchecked or not
+// in this component's UI
 export interface ProjectChecked extends Project {
   checked: boolean;
 }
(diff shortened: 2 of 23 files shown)


susanev:
@tylercloke ill take a closer look monday 08/19 but there should be 38px between the bottom of the help text and the save button.. looks like less than that in ur screenshot.

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -182,15 +212,40 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
             }
           });
       });
+
+    combineLatest([
+      this.store.select(getProjectStatus),
+      this.store.select(projectEntities)])
+      .pipe(
+        map(([status, projectMap]) => {
+          if (status === EntityStatus.loadingSuccess) {
```
srenatus: There's no else here -- can we move this into a `filter`?

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -211,11 +266,16 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
     this.saveSuccessful = false;
   }
 
-  public saveNameChange(): void {
+  public updateTeam(): void {
     this.saveSuccessful = false;
     this.saving = true;
-    const name: string = this.updateNameForm.controls.name.value.trim();
-    this.store.dispatch(new UpdateTeam({ ...this.team, name }));
+    const newName: string = this.updateNameForm.controls.name.value.trim();
```
srenatus: [style] I'd have let this be `name` and put `name` instead of `name: newName` below 🤷‍♂

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -211,11 +266,16 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
+    this.store.dispatch(new UpdateTeam({
+        id: this.team.id,
+        name: newName,
+        guid: this.team.guid, // to be deprecated after GA
```
srenatus: What does this mean? This is not part of an API or anything, it's our frontend TS code wrapping some backend thing. If we want the code here to look different, we should just do it 😉

e2e/cypress/integration/ui/settings/04_team_details.spec.ts
```diff
@@ -0,0 +1,231 @@
...
+
+  let iamVersion = <string><Object>Cypress.env('IAM_VERSION');
+  // assume 2.0 if not in CI. if you wish something different start cypress with
+  // IAM_VERSION set to what you are testing.
+  if (iamVersion === undefined) {
+    iamVersion = 'v2.0';
+  }
+
+  const describeIAMV2 = iamVersion.match(/v2/) ? describe : describe.skip;
+  const describeProjectsEnabled = iamVersion === 'v2.1' ? describe : describe.skip;
```
srenatus: I think I've seen this before? If I'm not wrong, can we promote into a shared helper function?

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
```diff
@@ -1,78 +1,69 @@
...
   label = UNASSIGNED_PROJECT_ID;
 
-  // Map of projects currently selected
-  selectedProjects = {};
+  constructor() { }
 
-  constructor() {
-    this.projects = this.sortProjectsByName(this.projects);
+  projectsArray(): ProjectChecked[] {
+    return Object.values(this.projects) ? Object.values(this.projects) : [];
```
srenatus: I don't get it. What is `Object.values(this.projects)` when it's falsy? Did you mean `return this.projects ? Object.values(this.projects) : [];`?

Since this is the constructor, and the default is `{}`, can we just do
```suggestion
    return Object.values(this.projects);
```
?

components/automate-ui/src/app/pages/team-details/team-details.component.ts
bcmdarroch: I think I added this comment a while back. We can drop it, it's just to say we can drop that field, and all the other "versionize" logic, once we've force upgraded everyone to v2

srenatus: 💭It would be a nicer experience to drop it if it was centralized to one place, instead of sprinkled all over it 😉

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -107,6 +120,30 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
+      .pipe(takeUntil(this.isDestroyed))
+      .subscribe((version) => {
+        if (version === null) { return; }
+        this.isMinorV1 = version === 'v1';
+      });
+
+    this.store.select(assignableProjects)
+      .subscribe((assignable: ProjectsFilterOption[]) => {
+        assignable.forEach(p => {
+          const proj = <Project>{
```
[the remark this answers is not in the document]
tylercloke: I was casting because we do need to cast from `string` to `IAMType` but I can make that a bit more explicitly with:


```
          const proj: Project = {
            id: p.value,
            name: p.label,
            type: <IAMType>p.type
          };
```

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -107,6 +120,30 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
+            id: p.value,
+            name: p.label,
+            type: p.type
+          };
+          // we don't want to override values that we fetched
+          // that were part of the team already
+          if (!this.projects.hasOwnProperty(proj.id)) {
```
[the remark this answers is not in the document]
tylercloke: `hasOwnProperty` will actually return a boolean but in practice the `undefined` we'd get in the negative case of `this.projects[proj.id]` should be fine. Don't see the problem with using `hasOwnProperty` but guess it's maybe not as clear?

components/automate-ui/src/app/pages/team-details/team-details.component.ts
tylercloke: Agreed, don't think that's in the scope of this PR though. A lot we could clean up about v1 vs v2 teams in the frontend code.

components/automate-ui/src/app/pages/team-details/team-details.component.ts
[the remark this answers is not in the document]
srenatus: We don't have to cast that, we could also do what we'd do in golang -- _type guards_. see here for docs: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

in terms of effort, it's bigger than the case here, though.

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
```diff
@@ -43,9 +43,10 @@ export class CreateObjectModalComponent implements OnInit, OnChanges {
 
   ngOnChanges(changes: SimpleChanges) {
     // if a new list of projects to populate dropdown with is passed in we update the dropdown
+    const checked = false;
     if (hasIn('assignableProjects.currentValue', changes)) {
       changes.assignableProjects.currentValue.forEach(proj =>
-        this.projects[proj.id] = projectCheckedFromProject(proj, false));
+        this.projects[proj.id] = { ...proj, checked });
```
srenatus: [nit] you could also do `{ ...proj, checked: false }`, you don't _need_ to introduce a constant for this.

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
```diff
@@ -1,78 +1,63 @@
...
+  constructor() { }
 
-  constructor() {
-    this.projects = this.sortProjectsByName(this.projects);
+  projectsArray(): ProjectChecked[] {
+    return Object.values(this.projects);
   }
 
-  ngOnInit() {}
+  ngOnChanges() {
```
msorens:
```suggestion
  ngOnChanges(): void {
```

Actually, is there a reason you are not using the standard pattern of `ngOnChanges(changes: SimpleChanges): void` ? No other reasons to change, I suppose, so probably OK.

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
```diff
@@ -1,78 +1,63 @@
...
+  // Emits a project that changed as a result of a check or uncheck.
+  @Output() onProjectChecked = new EventEmitter<ProjectChecked>();
 
   active = false;
-  showError = false;
   label = UNASSIGNED_PROJECT_ID;
 
-  // Map of projects currently selected
-  selectedProjects = {};
+  constructor() { }
```
msorens: Constructor not needed (but OK if you like to be explicit with it).

components/automate-ui/src/app/pages/component-library/demos/demo-projects-dropdown/demo-projects-dropdown.component.html
```diff
@@ -72,21 +78,19 @@ <h3>Typescript Code Example</h3>
     <h3>API</h3>
 
     <app-cl-api>
-      <api-prop name="projects" type="Array" required>
-        Array of projects to be passed to the dropdown
+      <api-prop name="projects" type="[id: string]: ProjectChecked" required>
+        The map of ProjectChecked by id. Any checked changes propogated via
```
msorens:
```suggestion
        The map of ProjectChecked by id. Any checked changes propagated via
```

tylercloke:
@srenatus Appreciate it! Always learn a bunch :)

components/automate-ui/src/app/pages/team-details/team-details.component.spec.ts
```diff
@@ -84,7 +90,9 @@ describe('TeamDetailsComponent', () => {
           router: routerReducer,
           teams: teamEntityReducer,
           users: userEntityReducer,
-          policies: policyEntityReducer
+          policies: policyEntityReducer,
+          projects: projectEntityReducer,
+          projectsFilter: projectsFilterReducer
```
msorens: The changes you made here also need to be mirrored in `initialState` on L38.

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
```diff
@@ -33,6 +41,23 @@ export class CreateObjectModalComponent implements OnInit {
...
+    }
+  }
+
+  onProjectChecked(project: ProjectChecked): void {
+    this.projects[project.id] = project;
+  }
+
+  dropdownDisabled(): boolean {
+    return Object.values(this.projects).length === 0;
```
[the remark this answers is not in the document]
tylercloke: `this.projects` isn't an array. it's a map.

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
tylercloke: Did not know that! themoreyouknow.gif

components/automate-ui/src/app/entities/teams/team.reducer.ts
```diff
@@ -80,7 +80,7 @@ export function teamEntityReducer(state: TeamEntityState = TeamEntityInitialStat
       return set(
         'getStatus',
         EntityStatus.loadingSuccess,
-        teamEntityAdapter.addOne(action.payload, state)
+        teamEntityAdapter.upsertOne(action.payload, state)
```
bcmdarroch: do we need upsert in the other success actions?

components/automate-ui/src/app/pages/team-details/team-details.component.ts
bcmdarroch: I think the point is more that we don't need that comment here. This should work:
```
    const name: string = this.updateNameForm.controls.name.value.trim();
    const projects: string[] = Object.values(this.projects).filter(p => p.checked).map(p => p.id);
    this.store.dispatch(new UpdateTeam({ ...this.team, name, projects }));
```

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -182,15 +215,39 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
             }
           });
       });
+
+    combineLatest([
```
bcmdarroch: I think this whole block could use an explanatory comment, especially regarding how we're using `teamProjectsLeftToFetch`

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -237,4 +299,17 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
     // Drop the previous fragment and add the incoming fragment.
     this.router.navigate([this.url.split('#')[0]], { fragment: event.target.value });
   }
+
+  onProjectChecked(project: ProjectChecked): void {
+    this.projects[project.id] = project;
```
bcmdarroch: maybe a quick clarifying comment: `updates whether the project was checked or unchecked`

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -237,4 +299,17 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
+  }
+
+  noProjectsUpdated(): boolean {
+    return xor(this.team.projects,
```
bcmdarroch: how does `xor` work?

e2e/cypress/integration/ui/settings/02_team_mgmt.spec.ts
```diff
@@ -21,22 +21,48 @@ describe('team management', () => {
...
+  beforeEach(() => {
+    cy.restoreStorage();
+  });
+
   afterEach(() => {
     cy.saveStorage();
     cy.cleanupTeamsByDescriptionPrefix(adminToken, cypressPrefix);
   });
 
   context('no custom initial page state', () => {
```
bcmdarroch: I've been meaning to ask, what is `context` and how is it being used in these tests? How is it different from `describe`?

e2e/cypress/integration/ui/settings/04_team_details.spec.ts
```diff
@@ -0,0 +1,237 @@
+describe('team management', () => {
+  let adminToken = '';
+  const now = Cypress.moment().format('MMDDYYhhmm');
+  const cypressPrefix = 'cypress-test';
+  const teamName = `${cypressPrefix} team ${now}`;
```
bcmdarroch: this feels like a huge block of variables. can we move some down closer to the tests where they're actually used?

components/automate-ui/src/app/entities/teams/team.reducer.ts
tylercloke: Yeah I didn't wanna make such a sweeping change without tests though. Just did this to make the cypress tests I was writing work properly.

tylercloke: Pretty sure I made a card to address this in general.

tylercloke: https://github.com/chef/automate/issues/1199

components/automate-ui/src/app/pages/team-details/team-details.component.ts
tylercloke: In this case, [xor](https://en.wikipedia.org/wiki/Exclusive_or) finds all projects that were in only one or the other of the two lists. So if either list contains a team that the other doesn't we'll know that there are projects to save.

e2e/cypress/integration/ui/settings/02_team_mgmt.spec.ts
tylercloke: They are functionally exactly the same but they are used in different situations:

>>> The purpose of “describe” is to wrap a set of tests against one functionality while “context” is to wrap a set of tests against one functionality under the same state.

This article goes into good detail:

http://lmws.net/describe-vs-context-in-rspec

tylercloke: If I have used it incorrectly somewhere feel free to call it out! I'm sure I have at some point.

e2e/cypress/integration/ui/settings/04_team_details.spec.ts
tylercloke: I try to do that in general. Unfortunately I think all the ones listed above are used in both the beforeEach block and across different tests.

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -107,31 +121,50 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
+        filter(identity)
+      )
+      .subscribe((version) => {
+        this.isMinorV1 = version === 'v1';
+      });
+
+    this.store.select(assignableProjects)
+      .subscribe((assignable: ProjectsFilterOption[]) => {
+        assignable.forEach(({ value: id, label: name, type: stringType }) => {
+          const type = <IAMType>stringType;
```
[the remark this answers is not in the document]
tylercloke: That results in:

```
Type 'string' is not assignable to type 'IAMType'.
```

I'm pretty sure we actually need to cast it to a string here.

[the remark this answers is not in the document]
msorens: When the types are compatible @bcmdarroch's suggestion is much preferred. But it does not work here, as @tylercloke noted. You would have to do it like this, which does not add any value over the original (sort of worse off, since it is redundant)
```
const type: IAMType = <IAMType>stringType;
```
So I would suggest go back to the original. 🤷‍♂

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -30,21 +30,37 @@ import {
...
+import { ProjectsFilterOption } from 'app/services/projects-filter/projects-filter.reducer';
+import { IAMType } from 'app/entities/policies/policy.model';
 
 const TEAM_DETAILS_ROUTE = /^\/settings\/teams/;
 
 export type TeamTabName = 'users' | 'details';
 
+export interface ProjectMap {
+  [id: string]: ProjectChecked;
+}
```
msorens: Ah, looks like should move this `ProjectMap` definition to projects-dropdown.component.ts so that it can be shared. Should also be used in the definition of `projects` in create-object-modal.component.ts (L30).

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
```diff
@@ -33,6 +41,23 @@ export class CreateObjectModalComponent implements OnInit {
     });
   }
 
+  ngOnChanges(changes: SimpleChanges): void {
+    // if a new list of projects to populate dropdown with is passed in we update the dropdown
+    const checked = false;
+    if (hasIn('assignableProjects.currentValue', changes)) {
+      changes.assignableProjects.currentValue.forEach((proj: Project) =>
+        this.projects[proj.id] = { ...proj, checked });
```
msorens: I think there is a bug here: this assumes the project list will only grow. But someone might delete a project, too. So need to empty out `this.projects` first before the loop.

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
```diff
@@ -91,20 +77,11 @@ export class ProjectsDropdownComponent implements OnInit {
...
-
-  private updateSelectedProjects(project: Project, enabled: boolean): void {
-    if (enabled) {
-      this.selectedProjects[project.id] = project;
-    } else {
-      delete this.selectedProjects[project.id];
-    }
-    const selectedLen = Object.keys(this.selectedProjects).length;
-    switch (selectedLen) {
+  updateLabel(): void {
```
msorens:
```suggestion
  private updateLabel(): void {
```

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
```diff
@@ -33,6 +41,23 @@ export class CreateObjectModalComponent implements OnInit {
...
+    }
+  }
+
+  onProjectChecked(project: ProjectChecked): void {
+    this.projects[project.id] = project;
+  }
```
msorens: Hmm... wondering if you identified some edge case where you need to replace the whole project? Seems better to do what only what you are claiming:
```suggestion
    this.projects[project.id].checked = project.checked;
```

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
```diff
@@ -33,6 +41,23 @@ export class CreateObjectModalComponent implements OnInit {
...
+    // if a new list of projects to populate dropdown with is passed in we update the dropdown
+    const checked = false;
+    if (hasIn('assignableProjects.currentValue', changes)) {
+      changes.assignableProjects.currentValue.forEach((proj: Project) =>
+        this.projects[proj.id] = { ...proj, checked });
+    }
+  }
+
+  onProjectChecked(project: ProjectChecked): void {
+    this.projects[project.id] = project;
```
msorens: Hmm... wondering if you identified some edge case where you need to replace the whole project? Seems better to do what only what you are claiming:
```suggestion
    this.projects[project.id].checked = project.checked;
```

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -237,4 +311,18 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
     // Drop the previous fragment and add the incoming fragment.
     this.router.navigate([this.url.split('#')[0]], { fragment: event.target.value });
   }
+
+  // updates whether the project was checked or unchecked
+  onProjectChecked(project: ProjectChecked): void {
+    this.projects[project.id] = project;
```
msorens: Same thing mentioned on create-object-modal: Should do only what you are claiming:
```suggestion
    this.projects[project.id].checked = project.checked;
```
... and then you do not need the clarifying comment @bcmdarroch suggested. 😁

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
```diff
@@ -20,10 +27,11 @@ export class CreateObjectModalComponent implements OnInit {
   @Output() close = new EventEmitter();
   @Output() createClicked = new EventEmitter<Project[]>();
 
+  public projects: { [id: string]: ProjectChecked } = {};
```
msorens: Another spot to use the `ProjectMap` type, so this would become `public projects: projectMap = {};`

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
```diff
@@ -1,78 +1,64 @@
...
-  // makes changes and is emitted to the parent component.
-  @Input() projects:  Array<Project> = [];
-
-  // Setting required to true means the dropdown will show an error if a user tries to close the
-  // dropdown but hasn't selected any projects.
-  @Input() required = false;
+export class ProjectsDropdownComponent implements OnChanges {
+  // The map of ProjectChecked by id. Any checked changes propagated via
+  // onProjectChecked. Updates should be applied to parent component state.
+  @Input() projects: { [id: string]: ProjectChecked } = {};
```
msorens: Another spot to use the `ProjectMap` type, so this would become `public projects: ProjectMap = {};`

components/automate-ui/src/app/pages/component-library/demos/demo-projects-dropdown/demo-projects-dropdown.component.html
```diff
@@ -72,21 +78,19 @@ <h3>Typescript Code Example</h3>
     <h3>API</h3>
 
     <app-cl-api>
-      <api-prop name="projects" type="Array" required>
-        Array of projects to be passed to the dropdown
+      <api-prop name="projects" type="[id: string]: ProjectChecked" required>
```
msorens:
```suggestion
      <api-prop name="projects" type="ProjectMap" required>
```

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -62,6 +78,13 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
   public removeText = 'Remove User';
 
   public unassigned = ProjectConstants.UNASSIGNED_PROJECT_ID;
+  public projects: ProjectMap = {};
+  // keep a list of projects that were in the teams list that are
+  // left to fetch so we can know if we are fully loaded or not.
+  // don't want to enable the projects-dropdown until we have
+  // checked the proper project checkboxes corresponding to projects
+  // already in the team.
+  public teamProjectsLeftToFetch: string[] = [];
```
msorens: 🤔 	Wondering if you could avoid the complications of this mechanism--coupled with _n_ calls to GetProject--by instead just doing a single `ListProjects`? I believe that would also address the "project not found" error I reported in our team channel, because it just naturally gets the valid projects.

components/automate-ui/src/app/pages/team-details/team-details.component.ts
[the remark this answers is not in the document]
srenatus: Where does `stringType` come from? Can we type it at its source?

Also, let's introduce a type guard for `IAMType` -- this way, we'll get an error that can be dealt with instead of somewhat undefined behaviour at runtime. Right now, this is saying _"whatever `stringType` is, dear compiler, assume it's `IAMType`"_. This throws out type safety 😉

components/automate-ui/src/app/pages/team-details/team-details.component.ts
msorens: In fact, the prior art for this is in the same file. We fire GetUsers and GetTeamUsers, then do a `combineLatest` on `users, uStatus, tUsers, tStatus` to mesh the two together. I suspect we could do something similar here.

PR #1262, opened 2019-08-16
Add airgap backup/restore/upgrade test

jaym:


lib/platform/pg/exporter.go
```diff
@@ -206,33 +216,143 @@ func (db DatabaseExporter) restoreSQLFile(exitOnError bool) error {
...
+		defer close(errChan)
+		scanner := bufio.NewScanner(reader)
+		for scanner.Scan() {
+			txt := scanner.Text()
+			if !strings.Contains(txt, "2615 2200") {
+				_, err := fmt.Fprintln(pgListFile, txt)
+				if err != nil {
+					errChan <- err
+				}
+				logrus.Debugf("Preserving TOC Entry: %s", txt) // REMOVE ME
```
stevendanna: Are we sure we want this?

lib/platform/pg/exporter.go
```diff
@@ -24,6 +28,11 @@ var PGRestoreCmd = []string{"hab", "pkg", "exec", "core/postgresql11-client", "p
 // PSQLCmd is the command we will run for psql.
 var PSQLCmd = []string{"hab", "pkg", "exec", "core/postgresql11-client", "psql"}
 
+// TODO (jaym): fix the actual problem
```
stevendanna: I can do this in a follow up as I have most of it already locally on a branch.

PR #1263, opened 2019-08-16
Ensure license-control-database is included in backup spec

stevendanna:

### DIFF
--- a/.expeditor/verify_private.pipeline.yml
+++ b/.expeditor/verify_private.pipeline.yml
@@ -714,6 +714,10 @@ steps:
       - integration/run_test integration/tests/backup.sh
     timeout_in_minutes: 20
     expeditor:
+      secrets:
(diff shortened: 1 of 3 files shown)


stevendanna:
The test modified here was run without the fix to prove that it fails:

https://buildkite.com/chef/chef-automate-master-verify-private/builds/4552#970ea221-f762-4b25-b3f2-ce91571c13d9

Screen Shot 2019-08-16 at 2 46 43 PM

stevendanna:
@srenatus It turns out we _had_ a test for this but it wasn't effectively being run because those tests weren't configured to use a license.  This tests adds a license to the backup tests so now we test that the license still exists post restore.

stevendanna:
We are discussing adding an additional test that makes sure we can't forget such things in the future, but I think that is going to be a bit more work.

PR #1264, opened 2019-08-16
Apply license in backup tests

stevendanna:
We expect this test to fail currently.


stevendanna:
This was incorporated into the PR that fixes the problem

PR #1265, opened 2019-08-16
diagnostics: fix errant error message and functions names

stevendanna:
This appeared to be testing the OptOut configuration but really tests
whether the license ID has changed. I've updated the comments,
function name, and API endpoint used to be a bit more honest.


stevendanna:
hrm, looks like the endpoint I used isn't right...

stevendanna:
With fresh eyes I see that the issue is that license status endpoint is a bit newer than the telemetry endpoint. OK.

Issue #774, opened 2019-07-03
Add Search/Filter Bar to Applications UI


danielsdeleo:
# Problem

Given the limitation of getting only a services page for MVP, users could not complete basic tasks such as checking the status of one deployment (an application deployed in an environment), or one service across all the environment and applications. 

To fulfill the needs, we would like to provide a search bar with filters that allow users to filter the services table to only the things they care about for a specific task.

# Solutions 
## Limitations
Because of how our database is designed currently, filters of which content require a combination of two columns in the database takes an extra amount of time to implement and is outside the scope for this issue. The combo filters and their corresponding substitutions are:
- `package`(origin/package or service name): substituted by `origin` AND `service name`.
- `service group` (service name/group name): substituted by `service name` AND `group name`.
- `release` (version/build timestamp): substituted by `version` AND `build timestamp`.

After we delivered the feature to users, we will gather feedback and see if the combo filters are needed.

Solution mockup preview:
![image.png](https://images.zenhubusercontent.com/5a721a104b5806bc2bc51768/f652eca5-78b1-4671-a957-3b5b57b0aa07)
![image.png](https://images.zenhubusercontent.com/5a721a104b5806bc2bc51768/2bd3cfac-8067-43f9-b8c2-f900c9475226)
## Solutions by Use Case
1. Find a specific deployment (e.g. by name) in A2, to see the rollout status of that deployment including desired vs. actual package version.
**Search by `application` AND `environment`**

2. When an app/service has been promoted, update the desired state in A2 for all the affected services listening to that depot/package/channel and flag them as "deploying" (to be done in another epic). This is often done in the e.g. Jenkins pipeline automatically via Hab Depot API and sometimes in the Hab Builder GUI.
**Search by `origin` AND `service` AND `channel`**

3. When an app/service is about to be deployed - show the how many systems, environments, deployments will be updated, so customer's deploy team can understand the scope of promoting an app/service. 
**Search by `origin` AND `service` AND `channel`**

4. Monitor the versions and status for a specific service that the development team cares about so that they don't have to go site by site to see the result. e.g. which version of POS service is hosted across the board?
**Search by `service`**

5. When there is a problem with services in a specific site, the ops team would want to see all the services status in that site.
**Search by `site`**

6. When a critical health check report is returned, users want to know the scope of the problem, how many systems or services are affected by the issue.
**Search by `health check message`**

7. Save filters as bookmarks, and share the filters to others via URLs. Sometimes directly change the URL to change the applied filters.
[...]

PR #1276, opened 2019-08-18
Automate-774 Service groups search/filter bar

tarablack01:
Given the limitation of getting only a services page for MVP, users could not complete basic tasks such as checking the status of one deployment (an application deployed in an environment), or one service across all the environment and applications.

API TODO: service_groups_health_counts api needs to update counts on search bar filters, do we need to apply filter params as well?:
Screen Shot 2019-08-18 at 2 49 51 PM
Screen Shot 2019-08-18 at 2 49 58 PM

To fulfill the needs, we would like to provide a search bar with filters that allow users to filter the services table to only the things they care about for a specific task.

https://github.com/chef/automate/issues/774
https://chef.invisionapp.com/share/D3T4FUT26B7#/screens/375625354

### :+1: Definition of Done
Reuse the implementation of the search bar/filters from the Client Runs page. This is an explicit requirement because we want a unified search/filter experience across all of Automate based on a single codebase.

We expect to have the following filters in order in the dropdown list:

- Origin, e.g., "core"
- Service Name, e.g., "nginx"
- Version, e.g., "1.2.3"
- Channel
- Application (note that Deployment == Environment + Application)
- Environment
- Site
- Health Check Message
- Build Timestamp, e.g., "20190318112388"
- Group Name (service group name suffix after the '.')

Saving & Sharing filters: append the applied filters to the URL like what Chef Infra search bar does, add the "share" functionality which allows copying the URL with the applied filters saved. In reverse, when updating the filters and values in the URL and refreshing the page, the applied filters update and match with what's in the URL.

Development tooling such as the applications_populate_database function in the studio should be updated as necessary to support the full demonstration of this feature.

build components/automate-ui && start_all_services && applications_populate_database
Go to https://a2-dev.test/applications and inspect

- [x ] Tests added/updated?
- [x ] Docs added/updated?

search-bar-service-groups
search-bar-event-feed
search-bar-client-runs


components/automate-ui/src/app/entities/service-groups/service-groups.requests.ts
```diff
@@ -79,4 +95,33 @@ export class ServiceGroupsRequests {
 
     return this.httpClient.get<HealthSummary>(url);
   }
+
+  public getSuggestions(type: string, text: string, filters: NodeFilter): Observable<any[]> {
+    if (text && text.length > 0) {
+      const params = this.formatFilters(filters).set('type', type).set('text', text);
+      const url = `${CONFIG_MGMT_URL}/suggestions`;
```
afiune: I might totally be missing something here but, why do we get the suggestions from the `config-mgmt-service`?

components/automate-ui/src/app/entities/service-groups/service-groups.actions.ts
```diff
@@ -21,7 +21,11 @@ export enum ServiceGroupsActionTypes {
 
   GET_SERVICES_BY_SERVICE_GROUP = 'SERVICE_GROUPS::SERVICES::GET',
   GET_SERVICES_BY_SERVICE_GROUP_SUCCESS = 'SERVICE_GROUPS::SERVICES::GET::SUCCESS',
-  GET_SERVICES_BY_SERVICE_GROUP_FAILURE = 'SERVICE_GROUPS::SERVICES::GET::FAILURE'
+  GET_SERVICES_BY_SERVICE_GROUP_FAILURE = 'SERVICE_GROUPS::SERVICES::GET::FAILURE',
+
+  GET_NODE_SUGGESTIONS = 'GET_NODE_SUGGESTIONS::SERVICES::GET',
+  GET_NODE_SUGGESTIONS_SUCCESS = 'GET_NODE_SUGGESTIONS::SERVICES::SUCCESS',
+  GET_NODE_SUGGESTIONS_FAILURE = 'GET_NODE_SUGGESTIONS:SERVICES::FAILURE'
```
afiune: Should we rename these actions? or why do we have the word "node" in the applications tab? 🤔

tarablack01: @afiune Updated name `GET_NODE_SUGGESTIONS` to be specific to service groups: `GET_SERVICE_GROUP_SUGGESTIONS`

components/automate-ui/src/app/entities/service-groups/service-groups.actions.ts
```diff
@@ -87,6 +91,23 @@ export class GetServicesBySGFailure implements Action {
   constructor(public payload: HttpErrorResponse) { }
 }
 
+export class GetNodeSuggestions implements Action {
```
afiune: all the classes are still named `GetNode*`

components/automate-ui/src/app/entities/service-groups/service-groups.requests.ts
tarablack01: @afiune thanks for pointing that out! I was able to confirm the correct api: applications_url: '/apis/beta/applications', `${APPLICATIONS_URL}/services-distinct-values`

tarablack01:
> I think I am totally out of sync from the plans for the search bar for the applications service, I don't understand why are we adding node actions and client-run requests, also I don't know why are we getting the suggestions from the `config-mgmt-service`. **If this is the plan then ignore me.** 😅 Otherwise, I think the main objective of this Issue/Task was to re-use the search bar components but implement new actions, requests, etc (redux stuff) for the applications service.

@afiune You're correct. I didn't have all the details for the api yet. Now I do. It's all implemented and ready to test 👍🏻

components/automate-ui/src/app/entities/service-groups/service-groups.actions.ts
```diff
@@ -21,7 +21,11 @@ export enum ServiceGroupsActionTypes {
 
   GET_SERVICES_BY_SERVICE_GROUP = 'SERVICE_GROUPS::SERVICES::GET',
   GET_SERVICES_BY_SERVICE_GROUP_SUCCESS = 'SERVICE_GROUPS::SERVICES::GET::SUCCESS',
-  GET_SERVICES_BY_SERVICE_GROUP_FAILURE = 'SERVICE_GROUPS::SERVICES::GET::FAILURE'
+  GET_SERVICES_BY_SERVICE_GROUP_FAILURE = 'SERVICE_GROUPS::SERVICES::GET::FAILURE',
+
+  GET_SERVICE_GROUPS_SUGGESTIONS = 'GET_SERVICE_GROUPS_SUGGESTIONS::SERVICES::GET',
```
afiune: To follow the name standard above I would suggest calling it like this:
```suggestion
  GET_SERVICE_GROUPS_SUGGESTIONS = 'SERVICE_GROUPS::SUGGESTIONS::GET',
```
Same for the two below 👇

components/automate-ui/src/app/entities/service-groups/service-groups.requests.ts
```diff
@@ -1,15 +1,23 @@
...
+  ServiceGroupFilters,
   HealthSummary,
-  ServicesPayload, ServicesFilters
+  ServicesPayload,
+  ServicesFilters
 } from './service-groups.model';
 import { environment } from '../../../environments/environment';
 const APPLICATIONS_URL = environment.applications_url;
 
+interface RespSuggestion {
```
afiune: can we move this definition to the `service-groups.model.ts` file?

tarablack01: @afiune updated 👍🏻

components/automate-ui/src/app/entities/service-groups/service-groups.actions.ts
tarablack01: @afiune updated 👍🏻

components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
```diff
@@ -167,13 +261,27 @@ export class ServiceGroupsComponent implements OnInit, OnDestroy {
 
     this.currentPage$.pipe(takeUntil(this.isDestroyed))
       .subscribe(currentPage => this.currentPage = currentPage);
+
+    this.nodeSuggestions$ = this.store.select(createSelector(serviceGroupState,
+      (state) => state.nodeSuggestions)).pipe(map((nodeSuggestions: any[]) =>
```
bcmdarroch: could we use a stronger type than `any` here?

components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
```diff
@@ -200,16 +308,82 @@ export class ServiceGroupsComponent implements OnInit, OnDestroy {
...
+
+  toggleShareDropdown() {
+    this.shareDropdownVisible = !this.shareDropdownVisible;
+  }
+
+  hideShareDropdown() {
+    this.shareDropdownVisible = false;
+  }
+
+  onSuggestValues(event) {
```
bcmdarroch: what types of events are we expecting in these functions?

components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
```diff
@@ -75,7 +75,7 @@ export class ServiceGroupsComponent implements OnInit, OnDestroy {
   filtersVisible = true;
 
   // autocomplete suggestions
-  nodeSuggestions$: Observable<any[]>;
+  serviceGroupsSuggestions$: Observable<any[]>;
```
bcmdarroch: should these all observable string arrays?

bcmdarroch:
Oh and also wanted to mention, a rebase should hopefully take care of that failing Cypress test 🤞 Ping auth team if it doesn't though

components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
msorens: Yes, I would second the suggestion to get rid of `any` for a specific type. (If you search through your code changes, wherever you reference suggestions it uses `any`, so when changing here, please check for the rest, too.) Thx!

components/automate-ui/src/app/entities/service-groups/service-groups.requests.ts
```diff
@@ -1,11 +1,16 @@
...
-  ServiceGroupsPayload, ServiceGroupFilters,
+  ServiceGroupsPayload,
+  ServiceGroupFilters,
+  ServiceGroupSuggestion,
   HealthSummary,
-  ServicesPayload, ServicesFilters
+  ServicesPayload,
+  ServicesFilters
 } from './service-groups.model';
 import { environment } from '../../../environments/environment';
```
msorens: Another instance of that path convention I mentioned above. 😁 
```suggestion
import { environment } from 'environments/environment';
```

components/automate-ui/src/app/entities/service-groups/service-groups.requests.ts
```diff
@@ -79,4 +90,34 @@ export class ServiceGroupsRequests {
 
     return this.httpClient.get<HealthSummary>(url);
   }
+
+  public getSuggestions(
+    field_name: string,
+    query_fragment: string,
```
msorens: For consistency should use camel case instead of snake case, so `fieldName` and `queryFragment`. (I think it is OK that they are not identical to the URL params.)

susanev:
@tarablack01 havent had time to run it locally yet, but from ur screenshot, the space between the subheading and the search bar doesn't look like it matches the rest of the pages. also, @apriofrost  should there be that gap between the search bar and the save button?

components/automate-ui/src/app/entities/service-groups/service-groups.requests.ts
```diff
@@ -79,4 +90,34 @@ export class ServiceGroupsRequests {
...
+
+    if (filters.searchBar) {
+      searchParam = this.flattenSearchBar(filters.searchBar, searchParam);
+    }
+
+    return searchParam;
+  }
+
+  private flattenSearchBar(filters: object[], searchParam: HttpParams): HttpParams {
+    return reduce((params: HttpParams, filter: { type: string, text: string }) => {
```
msorens: This filter type is used in about 8 places in the code base now. Would like to see you promote that to an exported interface so these couple lines could be a bit more clearly written:
```
  private flattenSearchBar(filters: FilterType[], searchParam: HttpParams): HttpParams {
    return reduce((params: HttpParams, filter: FilterType) => {
```

components/automate-ui/src/app/page-components/search-bar/search-bar.component.ts
```diff
@@ -310,6 +310,8 @@ export class SearchBarComponent implements OnChanges {
       this.suggestions = List<SuggestionItem>(type.providedValues);
       this.suggestionsVisible = true;
       this.isLoadingSuggestions = false;
+    } else {
+      this.suggestValues.emit({ detail: { text: '', type: type.type } });
```
msorens: With that `FilterType` I mentioned above, you could then use that to enforce strong typing on this line. Right now, if you make a typo--e.g. change `text` to `txet`--the compiler will not complain. But with strong typing it can.

components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
```diff
@@ -1,14 +1,19 @@
...
 import { NgrxStateAtom } from 'app/ngrx.reducers';
-import { ActivatedRoute, Router } from '@angular/router';
-import { Chicklet, RollupServiceStatus, SortDirection } from '../../types/types';
+import { ActivatedRoute, Router, ParamMap } from '@angular/router';
+import {
+  Chicklet,
+  RollupServiceStatus,
+  SearchBarCategoryItem,
+  SortDirection
+} from '../../types/types';
```
msorens: OK, I won't say anything about `../..` anymore. 😁

components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
```diff
@@ -63,6 +68,85 @@ export class ServiceGroupsComponent implements OnInit, OnDestroy {
   // Sort field by default
   readonly defaultSortField = 'percent_ok';
 
+  // Should the URL share dropdown be displayed
+  shareDropdownVisible = false;
+
+  // Should the search bar filter bar be displayed
```
msorens: Is that correct--two uses of the word "bar" there...?

components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
```diff
@@ -63,6 +68,85 @@ export class ServiceGroupsComponent implements OnInit, OnDestroy {
...
+  filtersVisible = true;
+
+  // autocomplete suggestions
+  serviceGroupsSuggestions$: Observable<any[]>;
+
+  // The catagories allowed for searching
```
msorens:
```suggestion
  // The categories allowed for searching
```

components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
```diff
@@ -200,16 +304,82 @@ export class ServiceGroupsComponent implements OnInit, OnDestroy {
...
       sortDirection: sortDirection,
       page: pageField,
-      pageSize: this.pageSize
+      pageSize: this.pageSize,
+      searchBar: searchBarFilters
     };
     this.store.dispatch(new UpdateServiceGroupFilters({filters: serviceGroupFilters}));
   }
 
+  get shareUrl() {
```
msorens:
```suggestion
  get shareUrl(): string {
```

And while you are in this file, if you don't mind, most of the functions in this file are in need of a `: void` showing no return type. Just a few functions here have it.

tarablack01:
@bcmdarroch @msorens Dang you both went to town on this PR, thanks!! Good job 👍🏻 I'll implemented the requested changes.

tarablack01:
> Interaction-wise, it looks good! A few feedback:
> 
> 1. Spacing change:
>    image
> 2. After a filter is selected from the filter list, could we add a loading state before the suggestion list is shown? In case of loading the suggestion list takes some time.
>    image
> 3. This is a nice to have, doesn't have to be included in this ticket. Wondering if we can make the list align with the focus area of the input field:
>    image

@apriofrost 1. has been updated 2. already exists 3. We won't do now but keep in mind the left position will always change based on the width of the first selected category; makes it a bit more tricky.

tarablack01:
@apriofrost @bcmdarroch @susanev @msorens 
Implemented requested PR changes, some of the requested changes for this PR were changes that I was making in my next story for refactoring the service groups files and adding a service facade. Since I had already started on that story during this PR I incorporated some of those changes in this PR because the same files overlapped and. The facade and service-groups.component will be refactored in that PR.

tarablack01:
@apriofrost 
I added 32px margin between subheading and search input field.
Screen Shot 2019-09-09 at 10 57 42 AM

apriofrost:
Looking great! @tarablack01

PR #1277, opened 2019-08-19
Move platform config into lib/platform/config

stevendanna:
This moves the platform configuration functions to a new package
platform/config. The goal of this change is to make is so that changes
to unrelated parts of the platform package don't rebuild the entire
world.


stevendanna:
To dis-entangle this more, we'll need to change how we pass around the various command.Opts

components/automate-platform-tools/cmd/pg-helper/main.go
```diff
@@ -270,12 +264,6 @@ func newCmd() *cobra.Command {
 		Args:  cobra.ExactArgs(2),
 	}
 
-	sqitchDeployCmd.Flags().StringVarP(
-		&opts.Sqitch.User,
-		"user", "i", "automate",
-		"User to connect with when running sqitch",
-	)
```
stevendanna: As far as I can tell, this feature would not have actually worked but we didn't notice it because we always call this with the default argument.  To avoid confusion, I've removed this option and made it an error on the API side to pass anything but the default user "automate".

lib/platform/pg/conninfo.go
```diff
@@ -55,6 +56,49 @@ var A2SuperuserCerts = TLSCertPaths{
 	RootCert: "/hab/svc/automate-postgresql/config/root.crt",
 }
 
+// NOTE(ssd) 2019-08-19: This is a bit of duplication with code in
+// platform/config to ensure that you can depend on the platform
+// config without pulling in the command class.
+type PlatformConnInfo struct {
```
jaym: niiice

PR #1298, opened 2019-08-19
bugfix: correct navigation path for run history items

vjeffrey:
We got a report in success slack of a bug: "Looks like seeing run history is broken in Automate. (I can see run history, but clicking on the specific run takes you back to the main "Infrastructure" (formerly client-runs) page."
Upon investigation, I noticed we missed a navigation rename when going from client-runs to infrastructure/client-runs.
This corrects that.

N/A

### :+1: Definition of Done
Selecting a past item in the run history for a node correctly loads the node details page with that run's information.

chef_load_nodes at least 2 times
rebuild the ui
select an item in the run history, expect to see that run loaded on the node details page

- [ ] Tests added/updated?
- [ ] Docs added/updated?

Screen Shot 2019-08-19 at 13 14 45
Screen Shot 2019-08-19 at 13 14 54

### DIFF
--- a/components/automate-ui/src/app/pages/node-details/node-details.component.html
+++ b/components/automate-ui/src/app/pages/node-details/node-details.component.html
@@ -4,7 +4,7 @@
     <div class="node-details-container">
       <div class="run-container">
         <chef-breadcrumbs>
-          <chef-breadcrumb [link]="['/client-runs']">Client runs</chef-breadcrumb>
+          <chef-breadcrumb [link]="['/infrastructure/client-runs']">Client runs</chef-breadcrumb>
           Node {{nodeRun.nodeName}}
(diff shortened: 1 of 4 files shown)


phiggins:
This line might have the same problem: https://github.com/chef/automate/blob/master/components/automate-ui/src/app/pages/node-details/node-details.component.html#L7

phiggins:
This might also have the same problem: https://github.com/chef/automate/blob/master/components/automate-ui/src/app/services/node-details/node-details-resolver.service.ts

phiggins:
I see a few more potential problems in test/demo code but probably not stuff that users can see.

vjeffrey:
:+1: i'll push some updates for those spots - thank you for checking!

PR #1299, opened 2019-08-19
Fix more append mistakes

stevendanna:
This fixes append mistakes with respect to the command slices in the
platform/pg package.


lib/platform/pg/exporter_integration_test.go
```diff
@@ -64,9 +64,9 @@ func testSetup(t *testing.T) (string, func()) {
 
 		switch requiredTool {
 		case "psql":
-			pg.PSQLCmd = []string{path}
+			pg.DefaultPSQLCmd = []string{path}
```
stevendanna: This isn't thread-safe, but this also only happens on startup in our tests.

PR #1300, opened 2019-08-20
 lib/cereal: bump default polling interval to 10 seconds

stevendanna:
Now that we have in-process notifications, we think we can push the
default time-based polling interval a bit higher to cut down on
unnecessary database queries.

While I was here, I did some minor cleanups on our logging.

### DIFF
--- a/lib/cereal/grpc/grpc.go
+++ b/lib/cereal/grpc/grpc.go
@@ -37,6 +37,15 @@ func NewGrpcBackendFromConn(domain string, conn *grpc.ClientConn) *GrpcBackend {
 		client: grpccereal.NewCerealClient(conn),
 	}
 }
+
+func (g *GrpcBackend) DefaultTaskPollInterval() time.Duration {
+	return 10 * time.Second
+}
+
+func (g *GrpcBackend) DefaultWorkflowPollInterval() time.Duration {
+	return 2 * time.Second
+}
+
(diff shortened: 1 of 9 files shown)


lib/cereal/grpc/grpc.go
```diff
@@ -41,6 +37,15 @@ func NewGrpcBackendFromConn(domain string, conn *grpc.ClientConn) *GrpcBackend {
...
 	}
 }
+
+func (g *GrpcBackend) DefaultTaskPollInterval() time.Duration {
+	return 10 * time.Second
+}
+
+func (g *GrpcBackend) DefaultWorkflowPollInterval() time.Duration {
+	return 2 * time.Second
+}
```
stevendanna: I added this in response to a test failure that makes sure we are starting workflows within some bound of their expected start time.

We could alternatively simply set different polling intervals in our tests, but this seemed like an easy way to allow backends to bring their own default values if they need something different.

stevendanna:
Note that I picked capitalized log messages just because of the following:

```
> rg '(Debug|Info|Warn|Error)f*\("[A-Z]' | wc -l
      31
> rg '(Debug|Info|Warn|Error)f*\("[a-z]' | wc -l
      54
```

If we would prefer to go the other direction, that is fine with me too.  I just prefer them to be consistent.  Not a big deal though, I can drop those commits from this PR if people feel strongly about it.

PR #1301, opened 2019-08-20
Remove polling on projects page due to flicker

msorens:

Due to flicker, remove polling on projects list page.

### :+1: Definition of Done
No flicker

Rebuild automate-ui.
Set to IAM v2.1.
Go to Settings > Projects


stevendanna:
I'm not opposed to this, but do we know (and, if so, could we include an explanation of) why this polling introduces flickering?

bcmdarroch:
We were polling for a new list of projects. For the instant while the GetProjects is in progress, ngrx store (the frontend state) clears out its current list of projects then reloads it with the latest list, causing the flicker.

PR #1304, opened 2019-08-20
Add single task workflow pattern

jaym:
This workflow executor makes it so you don't have to write a workflow executor if your workflow just needs to run 1 task. You can decide if you want to ignore cancel.

### DIFF
--- a/lib/cereal/patterns/singletaskworkflow.go
+++ b/lib/cereal/patterns/singletaskworkflow.go
@@ -1,0 +1,53 @@
+package patterns
+
+import (
+	"encoding/json"
+
+	"github.com/chef/automate/lib/cereal"
+	"github.com/sirupsen/logrus"
+)
+
+type SingleTaskWorkflowExecutor struct {
+	taskName    string
+	allowCancel bool
+}
+
+func NewSingleTaskWorkflowExecutor(taskName string, allowCancel bool) *SingleTaskWorkflowExecutor {
+	return &SingleTaskWorkflowExecutor{
+		taskName:    taskName,
+		allowCancel: allowCancel,
+	}
+}
+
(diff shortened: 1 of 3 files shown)


lib/cereal/patterns/singletaskworkflow.go
```diff
@@ -0,0 +1,54 @@
+package patterns
```
stevendanna: I like the name

lib/cereal/patterns/singletaskworkflow.go
```diff
@@ -0,0 +1,54 @@
...
+		allowCancel: allowCancel,
+	}
+}
+
+func (s *SingleTaskWorkflowExecutor) OnStart(w cereal.WorkflowInstance, ev cereal.StartEvent) cereal.Decision {
+	var params json.RawMessage
+	err := w.GetParameters(&params)
+	if err != nil {
+		logrus.WithError(err).Error("failed to get parameters")
+		w.Complete()
```
stevendanna: I think you mean to return this.  We probably also want to fail the workflow:
```suggestion
		return w.Fail(err)
```

lib/cereal/patterns/singletaskworkflow.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/chef/automate/lib/cereal"
+	"github.com/sirupsen/logrus"
+)
+
+type SingleTaskWorkflowExecutor struct {
+	taskName    string
+	allowCancel bool
+}
+
+func NewSingleTaskWorkflowExecutor(taskName string, allowCancel bool) *SingleTaskWorkflowExecutor {
```
stevendanna: A doc comment on this or the struct above might be nice.

PR #1305, opened 2019-08-20
Add cypress prefix to projects and clean them up before and after the projects API tests

tylercloke:

This should help make them less flaky since you can only have 6 projects in a single environment at any time. Also prevented the project variables from being global in scope by moving them into the describe block and fixed the tabbing in before.

### DIFF
--- a/e2e/cypress/integration/api/iam/01_projects_api.spec.ts
+++ b/e2e/cypress/integration/api/iam/01_projects_api.spec.ts
@@ -1,41 +1,3 @@
-const avengersProject = {
-  id: `avengers-project-${Cypress.moment().format('MMDDYYhhmm')}`,
-  name: 'Test Avengers Project'
-};
-
-const xmenProject = {
-  id: `xmen-project-${Cypress.moment().format('MMDDYYhhmm')}`,
-  name: 'Test X-men Project'
-};
-
-const avengersRule = {
-  id: 'avengers-rule-1',
-  name: 'first rule of avengers project',
-  type: 'NODE',
-  project_id: avengersProject.id,
-  conditions: [
-    {
-      attribute: 'CHEF_ORGANIZATION',
-      operator: 'EQUALS',
-      values: ['avengers']
-    }
-  ]
-};
-
-const xmenRule = {
-  id: 'xmen-rule-1',
-  name: 'first rule of xmen project',
-  type: 'NODE',
-  project_id: xmenProject.id,
-  conditions: [
-    {
-      attribute: 'CHEF_ORGANIZATION',
-      operator: 'EQUALS',
-      values: ['xmen']
-    }
-  ]
-};
-
 if (Cypress.env('IAM_VERSION') === undefined) {
   // assume local env is on IAM v2.0 since that's the default setting
   Cypress.env('IAM_VERSION', 'v2.0');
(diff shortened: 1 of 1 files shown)


e2e/cypress/integration/api/iam/01_projects_api.spec.ts
```diff
@@ -1,157 +1,164 @@
...
 
 describeIfIAMV2p1('projects API: applying project', () => {
+  const cypressPrefix = 'cypress-test';
+  const avengersProject = {
+    id: `${cypressPrefix}-avengers-project-${Cypress.moment().format('MMDDYYhhmm')}`,
+    name: 'Test Avengers Project'
+  };
+
+  const xmenProject = {
+    id: `${cypressPrefix}-xmen-project-${Cypress.moment().format('MMDDYYhhmm')}`,
```
tylercloke: Actual change 1: added cypressPrefix to the project IDs.

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
```diff
@@ -1,157 +1,164 @@
...
-    // it gets run every time:
-    // tslint:disable-next-line:max-line-length
-    // https://docs.cypress.io/guides/references/best-practices.html#Using-after-or-afterEach-hooks
-    cleanupTestProjects(admin.id_token);
+      // Cypress recommends state cleanup in the before block to ensure
+      // it gets run every time:
+      // tslint:disable-next-line:max-line-length
+      // https://docs.cypress.io/guides/references/best-practices.html#Using-after-or-afterEach-hooks
+      cleanupTestProjects(adminIdToken);
+      cy.cleanupProjectsByIDPrefix(adminIdToken, cypressPrefix);
```
tylercloke: Actual change 2: cleanup all the projects.

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
```diff
@@ -1,157 +1,160 @@
...
-        failOnStatusCode: false,
-        body: project
+        auth: { bearer: adminIdToken},
+        headers: {
+          projects: '(unassigned)'
+        },
+        method: 'GET',
+        url: '/api/v0/cfgmgmt/nodes?pagination.size=10'
       }).then((response) => {
-        expect([200, 409]).to.include(response.status);
```
tylercloke: Actual change 3: removed `failOnStatusCode: false`. If we get a non-2xx code back from creating the projects we should fail because 1) all cypress projects should be cleaned up at this point and 2) we are using timestamps in the IDs so we should never conflict anyway. If we conflict there is a problem that we shouldn't ignore.

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
```diff
@@ -1,157 +1,160 @@
...
+        url: '/api/v0/cfgmgmt/nodes?pagination.size=10'
+      }).then((response) => {
+        expect(response.body).to.have.length(0);
+      });
     });
   });
-});
+
+  after(() => {
+    cy.cleanupProjectsByIDPrefix(adminIdToken, cypressPrefix);
```
tylercloke: Actual change 4: Cleanup all cypress projects after the tests finish.

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
bcmdarroch: Will we ever need different ID prefixes? We could make `cypress-test` available a global config value in cypress.json and simplify some of these cleanup functions

PR #1308, opened 2019-08-20
Fix race condition in cypress test by preventing typing in input before team loaded

tylercloke:

If you have a team in the store already, it will populate the team save input before the team is fetched via the API. Since we didn't prevent typing in the input before the team was fetched via the API, cypress would find the input, find that the team name was correct (from the store cache), and then type. However, this would happen so fast the team wasn't fetched yet. When the fetch completed, it overwrote the updated team name cypress typed before save was clicked.

I've added functionality to disable the team name input before the team is fetched and while a save is in progress. The cypress tests should now pass since they are asserting that the input is enabled, preventing the race condition.

- [x] Tests added/updated?
- [ ] Docs added/updated?

Also added base styling for the chefInput directive when disabled. It looks like this:

Screen Shot 2019-08-20 at 3 39 07 PM

### DIFF
--- a/components/automate-ui/src/app/pages/team-details/team-details.component.ts
+++ b/components/automate-ui/src/app/pages/team-details/team-details.component.ts
@@ -1,5 +1,5 @@
 import { Component, OnInit, OnDestroy } from '@angular/core';
-import { FormBuilder, FormGroup, Validators } from '@angular/forms';
+import { FormBuilder, FormGroup, FormControl, Validators } from '@angular/forms';
 import { Router } from '@angular/router';
 import { Store, select } from '@ngrx/store';
 import { identity, keyBy, at, xor } from 'lodash/fp';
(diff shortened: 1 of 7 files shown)


e2e/cypress/integration/ui/settings/04_team_details.spec.ts
```diff
@@ -1,4 +1,4 @@
-describe('team management', () => {
```
tylercloke: O L D C O P Y P A S T A (bad Tyler)

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -95,7 +95,8 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
     };
     this.updateNameForm = fb.group({
       // Must stay in sync with error checks in team-details.component.html
-      name: ['Loading...', [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]]
+      name: new FormControl({value: 'Loading...', disabled: true},
```
bcmdarroch: maybe a comment here about why we disable it to start

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -279,6 +284,7 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
   public saveTeam(): void {
     this.saveSuccessful = false;
     this.saving = true;
+    this.updateNameForm.controls['name'].disable();
```
msorens: Couldn't you add `[disabled]="saving"` in the template so the disable/enable will be done automatically? (Notice that you are setting `this.saving` exactly where you are calling `disable()` and `enable()` in this function.)

tylercloke: Nope, if you try to use `[disabled]` with an input that is using formbuilder the compiler yells at you to not do that.

tylercloke: Really wish you could. I wanted to do it the way you described. Maybe we can. Need to look deeper but it appears formbuilder gets mad.

PR #1309, opened 2019-08-21
[automate-854] Project update status on Projects List page

msorens:

This first timing diagram shows nominal behavior for adding/editing rules for a project. The user at the top initiates various actions. The two sections below, "Ingest Rules" and "Project Update Status" represent the column names on the Projects List page.

This PR deals mostly with populating the "Project Update Status"  column with live data; "Ingest Rules" was done in an earlier PR, though it is fine-tuned here.
This PR includes front-end support for the "Cancel" operation; however, it is not yet supported by the back-end. Jay's imminent work should close the loop on that.

nominal

Similarly, this shows the sequence around a user cancelling an update:
cancelling

In order to populate those columns on the Projects List page, this adaptive sequence of polling the rules status is applied:

adaptive timing

There are three stage of polling beginning at the moment the user confirms "Update Projects":

1. Initial Burst -- this fires 4 times a second for just 3 seconds. It provides the user some visual feedback in the event that the update process is very quick. Otherwise, they might not even see that the update started and finished.
2. Active Monitoring -- after the initial burst this fires every 5 seconds as long as the update is in process.
3. Dormant Monitoring -- as soon as the update completes, we revert to dormant monitoring, just once per minute, like leaving the pilot light on. As soon as this user--or any other user begins another update, dormant monitoring transitions again to (1) if the current user initiated the update or (2) if some other user did.

PR - #1250 (Update rules status on Project List page)

### :+1: Definition of Done

Both "Ingest Rules" and "Project Update Status" columns on the Projects List page are live.

1. rebuild automate-ui
2. In hab studio, create lots of nodes, e.g. `chef_load_nodes 10000`. That will take minutes. (You could usually see something with just 5000, to make the wait less onerous.)
3. Create project and rules.
4. Note that where you have added rules you will see "needs updating".
5. Select "Update Projects" and confirm.
➡️ Observe the progress in the Update button and soon enough the status changes to OK.
Be sure to look at the unit tests for other than happy path scenarios and for details on how the mechanism works.


components/automate-ui/src/app/pages/project/list/project-list.component.html
```diff
@@ -17,15 +17,16 @@
       <ng-container *ngIf="(sortedProjects$ | async)?.length > 0">
         <chef-toolbar>
           <app-authorized [allOf]="['/iam/v2beta/projects', 'post']">
```
msorens: Cleanup: Just adding some line breaks in this section.

components/automate-ui/src/app/pages/project/list/project-list.component.html
```diff
@@ -46,7 +47,7 @@
                 </chef-td>
                 <chef-td>{{ project.id }}</chef-td>
                 <chef-td>{{ getRulesStatus(project) }}</chef-td>
-                <chef-td>OK</chef-td>
+                <chef-td>{{ getProjectStatus(project) }}</chef-td>
```
msorens: Here is where we introduce live data to the Project Update Status column.

components/automate-ui/src/app/pages/project/list/project-list.component.spec.ts
```diff
@@ -9,18 +9,41 @@ import { ChefPipesModule } from 'app/pipes/chef-pipes.module';
 import { NgrxStateAtom } from 'app/ngrx.reducers';
 import { customMatchers } from 'app/testing/custom-matchers';
 import { FeatureFlagsService } from 'app/services/feature-flags/feature-flags.service';
```
msorens: A lot of noise in this diff due to cleanup. New tests at the end of the file.

components/automate-ui/src/app/pages/project/list/project-list.component.spec.ts
```diff
@@ -346,4 +347,148 @@ describe('ProjectListComponent', () => {
       expect(projectService.applyRulesStop).toHaveBeenCalled();
     });
   });
+
```
msorens: New tests start here.

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -168,46 +201,71 @@ export class ProjectListComponent implements OnInit, OnDestroy {
...
+  public getRulesStatus(project: Project): string {
     switch (project.status) {
       case 'NO_RULES': return 'No rules';
       case 'EDITS_PENDING': return 'Edits pending';
       case 'RULES_APPLIED': return 'Applied';
       default: return '';
     }
   }
+
+  public getProjectStatus(project: Project): string {
```
msorens: This is the workhorse for populating the "Project Update Status" column and what all the new unit tests exercise.

dev-docs/diagrams/rule-updates-status-check.iuml
```diff
@@ -0,0 +1,125 @@
+@startuml
```
msorens: Timing diagram "code". Much more interesting to look at the output (png) files.

components/automate-ui/src/app/entities/projects/project.effects.ts
```diff
@@ -202,7 +204,7 @@ export class ProjectEffects {
     withLatestFrom(this.store.select(iamMajorVersion).pipe(filter(identity))),
     withLatestFrom(this.store.select(iamMinorVersion).pipe(filter(identity))),
     filter(([[_, major], minor]) => {
-      return major === 'v2' && minor === 'v1';
+      return major === 'v2' && minor === 'v1' && major < minor;
```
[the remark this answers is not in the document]
msorens: Oops... that's a debug line. Thanks for catching it--needs to be removed.

components/automate-ui/src/app/entities/projects/project.effects.ts
```diff
@@ -193,23 +196,33 @@ export class ProjectEffects {
...
-    switchMap(() =>
-      this.requests.getApplyRulesStatus().pipe(
-        map((resp) => new GetApplyRulesStatusSuccess(resp)),
-        catchError((error: HttpErrorResponse) =>
-          observableOf(new GetApplyRulesStatusFailure(error))))));
-}
+    switchMap(this.getRulesStatus$()));
 
+  @Effect()
+  getActiveApplyRulesStatus$ = observableInterval(1000 * ACTIVE_RULE_STATUS_INTERVAL).pipe(
```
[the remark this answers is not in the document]
msorens: That's true, but the bigger complication is introducing a variable interval. It can be done, but it is more complicated code--this is so simple comparatively, which is why I went with two emitters.

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -36,10 +38,19 @@ export class ProjectListComponent implements OnInit, OnDestroy {
...
-  public MAX_PROJECTS = 6;
   public confirmApplyStartModalVisible = false;
   public confirmApplyStopModalVisible = false;
 
+  // During an update the Project Update Status needs to show "updating..."
+  // for any project whose status is "edits pending" at the moment the update starts.
+  // Very shortly after the update starts, though, the project's live status changes
+  // from "edits pending" to "applied" so we have lost that initial status.
+  // This statusCache, then, remembers those initial status values during the update.
+  private statusCache: {  [id: string]: ProjectStatus; } = {};
```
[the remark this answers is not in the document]
msorens: Not a bad idea... I'll see how much work is involved.

components/automate-ui/src/app/entities/projects/project.effects.ts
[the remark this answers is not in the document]
msorens: Wait, scratch that last comment. It was a debug line, but it was removed several commits ago... I guess you saw that when you were walking the commits.

msorens:
@tylercloke The polling is only doing what you said--showing that a project update is in progress or not. That information will be on the banner eventually, yes, but it is also directly reflected on the update status of each project.

components/automate-ui/src/app/pages/project/list/project-list.component.ts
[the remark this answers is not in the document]
msorens: I am going to leave this as is for now; I think that this will need to change anyway once Jay finishes changing the way updates are handled.

PR #1317, opened 2019-08-21
Compliance shouldn't fail hard if notifications is down

jaym:


stevendanna:
Might be useful to include some words about why we don't need to block startup for this service in particular.

lancewf:
What happens when the notification service is not there and a failed InSpec report is ingested?

stevendanna:
> What happens when the notification service is not there and a failed InSpec report is ingested?

I don't think blocking here prevents you from having to handle that case though. notifications-service could go away at any time.

components/compliance-service/compliance.go
```diff
@@ -262,7 +262,7 @@ func getNotificationsConnection(connectionFactory *secureconn.Factory,
 	timeoutCtx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
 	defer cancel()
 	conn, err := connectionFactory.DialContext(timeoutCtx, "notifications-service",
```
lancewf: If the notification-service is not running, will this wait until the 60-second timeout and return a connection `conn` without an error?

jaym: i don't think so. I removed the WithBlock so it doesn't wait for anything

PR #1329, opened 2019-08-22
lib/cereal: jitter the task poller wakeup

stevendanna:
This is one of a couple of changes to reduce the churn in database
connections used by cereal. By default, the Go sql library only keeps
2 idle connections. Thus, when we have a burst of task poller
activity, we start and subsequently close many new connections,
causing log spam from automate-pg-gateway.
    
One thing we can and will do is bump the number of idle connections.
    
This PR, however, reduces the "thundering hurd" of task pollers, by
introducing a small jitter into their wakeup time.

### DIFF
--- a/lib/cereal/cereal.go
+++ b/lib/cereal/cereal.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math/rand"
 	"sort"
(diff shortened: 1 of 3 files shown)


stevendanna:
Related: https://github.com/chef/automate/pull/1330

PR #1330, opened 2019-08-22
lib/cereal: Set MaxIdleConnections in postgresql backend to 4

stevendanna:
In an ideal world, we would not be managing this database object
ourselves. However, since we are, this sets the MaxIdleConnections to
a value that we think is a bit more appropriate for our current users
of cereal.

The end goal of this is to cut down on the log spam caused by
perpetually closing connections.

### DIFF
--- a/lib/cereal/postgres/postgres.go
+++ b/lib/cereal/postgres/postgres.go
@@ -17,6 +17,18 @@ import (
 )
 
 const (
+	// defaultMaxIdleConnections is will be used to configure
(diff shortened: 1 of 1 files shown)


lib/cereal/postgres/postgres.go
```diff
@@ -17,6 +17,18 @@ import (
 )
 
 const (
+	// defaultMaxIdleConnections is will be used to configure
```
stevendanna:
```suggestion
	// defaultMaxIdleConnections is used to configure
```

PR #1332, opened 2019-08-22
multi region support for aws-api integration

vjeffrey:
The content team has a new, certified AWS profile that supports multi-region scanning.
This code change modifies the "aws-api" integration and scan job path to support (and encourage) this updated profile/behaviour.

Previously, we were creating one node per region for every "aws-api" integration. Now that the profile itself supports multi-region scanning, we are moving to creating one node per account. We are removing the regions filter on scan job create. The inspec-agent logic has been updated to support both the old code path (one node per region) and the new one (one node per account).
We will be recommending that users delete their old "aws-api" integration and all associated nodes and create a new one to be used with the profile.

https://blog.chef.io/2019/03/13/chef-software-is-the-first-cis-partner-certified-on-aws-azure-and-gcp/
https://blog.chef.io/2019/03/21/hit-the-cloud-running-with-inspec-profiles/
https://github.com/chef/automate/issues/75

### :+1: Definition of Done
- only one node created per account for "aws-api" integration
- region filters on scan job create is removed, just one node available per aws-api integration
- scans for the new style and old style both work

- start up the services
- add an aws api integration (use okta aws functionality to get aws creds):
```
curl -s --insecure -H "api-token: $token" $url/api/v0/nodemanagers -d '{
    "name": "my aws api integration with session token",
    "type": "aws-api",
    "instance_credentials": [],
		"credential_data": [
			{"key": "AWS_ACCESS_KEY_ID", "value": "value" },
			{"key": "AWS_SECRET_ACCESS_KEY", "value": "value" },
			{"key": "AWS_SESSION_TOKEN", "value": "value" }
		]
  }'
```
- see 16 nodes added
- start a recurring job against that account (every 10 min or something)
- rebuild compliance, nodemanager, and ui
- get new creds and create another integration
- expect to see one node for this new integration
- install the new aws profile (i can't link it here b/c it's not public but if anyone needs it for testing pls dm me)
- run a scan job with new integration and new profile
- ensure scan job is created, look at the report
- go back and check on the recurring scan job running on the old integration, make sure it is still doing the right thing/executing scans and creating reports

- [x] Tests added/updated?
- [x] Docs added/updated?

Screen Shot 2019-08-21 at 17 50 04
Screen Shot 2019-08-21 at 17 50 44
Screen Shot 2019-08-21 at 18 08 19


vjeffrey:
still working on test updates and docs for this change.
also still haven't run through the full test case described in the pr yet.

vjeffrey:
ok, i've updated docs and tests, and run through the entire test case described above. taking this out of draft.

components/compliance-service/inspec-agent/resolver/resolver.go
```diff
@@ -259,6 +259,70 @@ func handleRegionFilters(filters []*common.Filter, regions []string) ([]string,
 }
 
 func (r *Resolver) handleAwsApiNodes(ctx context.Context, m *manager.NodeManager, filters []*common.Filter, job *jobs.Job) ([]*types.InspecJob, error) {
+	// we have upgraded our logic to be able to work with mult-region scanning, meaning any new integrations
+	// created will only create one node per account (instead of one per region)
+	// the following logic checks to see if more than one node is associated with the account in our db
+	// if it is, we follow the "old path"
+	// if it is not, we follow a "new path" where we just prepare the one node for scanning
```
phiggins: I'm a little confused about how this works. The old method creates entries in the nodes table if they don't exist, but the new method assumes the one node is already there. Where does that node get created?

phiggins: Does this need to handle the case when 0 nodes exist?

components/compliance-service/inspec-agent/resolver/resolver.go
```diff
@@ -259,6 +259,70 @@ func handleRegionFilters(filters []*common.Filter, regions []string) ([]string,
...
+	if len(m.CredentialId) > 0 {
+		secret, err = r.secretsClient.Read(ctx, &secrets.Id{Id: m.CredentialId})
+		if err != nil {
+			return nil, fmt.Errorf("unable to fetch credential id:%s %s", m.CredentialId, err.Error())
+		}
+		accessKeyID, secretKey, _, sessionToken = managers.GetAWSCreds(secret)
+	}
+
+	jobArray := []*types.InspecJob{}
+	jobArray = append(jobArray, &types.InspecJob{
```
phiggins: nitpick: Instead of creating an empty array and immediately appending one element, you could create the array with the one element in it and return it in one statement.

components/compliance-service/inspec-agent/resolver/resolver.go
phiggins: Ok, I think I just needed to read further. It looks like the nodes get created in `/nodemanager-service/api/manager/server/server.go`. That means the "old path" could probably be simplified but it's probably worth keeping it as is because its in a known working state.

phiggins: Also we expect the things to always be in the database, so 0 results here should probably be an error.

components/nodemanager-service/pgdb/node-managers-cloud.go
```diff
@@ -98,51 +98,50 @@ func (db *DB) AddManagerSubscriptionsToDB(subs []*manager.ManagerNode, managerId
...
-			continue
+			return errors.Wrapf(err, "AddManagerNodeToDB unable to insert node_managers_nodes manager_id: %s node_id: %s", managerId, uuid)
 		}
-		nodeIds = append(nodeIds, uuid)
+		return nil
+	})
+
+	if err != nil {
+		logrus.Errorf("AddManagerRegionsToDB unable to insert node for %s %v", acctAlias, err)
+		return nodeIds, errors.Wrapf(err, "AddManagerRegionsToDB unable to insert node for %s", acctAlias)
```
phiggins: nitpick: If there's an error this probably should return nil for the first argument instead of a potentially half-baked result.

components/compliance-service/inspec-agent/resolver/resolver.go
vjeffrey: good point - i'll push an update to return an error if there are 0 results here

vjeffrey:
currently discussing node name for this:
AWS Account (aws acct id or acct alias)
vs
aws-account-acct-id-or-alias

vjeffrey:
from @jonong1972 :

> More and more I am still leaning towards this format for reasons:
1.) other data values are like this across Automate
2.) if I were to filter on this name, the ease of typing a Hyphen vs. Parenthesis for the auto suggest

vjeffrey:
rebasing on master

jonong1972:
After discussing with the UX team and doing a little research online about conventional naming standards.
We decide to stick with the hyphenated name:

`aws-account-acct-id-or-alias`

As mentioned above, to keep consistent across Automate, we utilize the hyphens, looking through AWS and other documentation, they name their ids with hyphens.  So just to keep an industry standard for naming nodes.

PR #1336, opened 2019-08-23
ci: don't commit to master in update_docker_image script

stevendanna:
We failed to `git add` the nightly.pipeline.yml file. As a result,
part of the update this script performs was being committed directly
to master rather than as part of the PR.


stevendanna:
This explains the following commit: https://github.com/chef/automate/commit/8882385be35978c4d87f4f8240d6000d21753cf9

PR #1346, opened 2019-08-24
Project Update Workflow and patterns to support it

jaym:
The chain workflow pattern is usued to be able to combine two workflow
executors to run one after the other. This runs both in the same
workflow instance. Using this pattern is a good way to take a large
workflow and break in down into smaller workflows.

You may not nest chains as the level of magic needed to make that work is
quite high and this PR is already getting big.

This PR also adds a unit testing framework for cereal workflows. It's
not perfect. There is a bit of weirdness around Decisions. We currently
test the last decision the workflow instance returned instead of the
decision the workflow executor's callback returned. We need to do this
until Decision can be more easily unit tested. It currently holds a lot
of private bits (the tasks that will be enqueued).


lib/cereal/patterns/chain.go
```diff
@@ -0,0 +1,333 @@
...
+import (
+	"context"
+	"encoding/json"
+	"errors"
+
+	"github.com/chef/automate/lib/cereal"
+)
+
+var ErrIncorrectParameters = errors.New("Incorrect number of parameters given")
+var ErrTaskWorkflowInvalid = errors.New("Could not determine workflow for task")
```
stevendanna: potential follow up: move these to error types where we can add more information to them

lib/cereal/patterns/chain.go
```diff
@@ -0,0 +1,333 @@
...
+	executors []cereal.WorkflowExecutor
+}
+
+type ChainWorkflowTaskParam struct {
+	XXX_ChainWorkflowIdx int64 `json:"__idx"`
+}
+
+func NewChainWorkflowExecutor(executors ...cereal.WorkflowExecutor) (*ChainWorkflowExecutor, error) {
+	if len(executors) <= 0 {
+		return nil, errors.New("Can't have zero")
```
stevendanna:
```suggestion
		return nil, errors.New("must have 1 or more chained executors")
```

lib/cereal/patterns/chain.go
```diff
@@ -0,0 +1,333 @@
...
+
+	params := ChainWorkflowParams{}
+	if err := instance.GetParameters(&params); err != nil {
+		return nil, err
+	}
+	chainInstance.params = &params
+	return &chainInstance, nil
+}
+
+func GetChainWorkflowInstance(ctx context.Context, m *cereal.Manager, workflowName string, instanceName string) (*ChainWorkflowInstance, error) {
```
stevendanna: I think this might be unused? Which then would make ToChainWOrkflowInstance unused as well unless it is used in one of the other PRs

lib/cereal/patterns/common.go
```diff
@@ -0,0 +1,269 @@
...
+		if t.Elem().Kind() == reflect.Ptr {
+			return false
+		}
+		return isEmbeddable(t.Elem())
+	default:
+		return false
+	}
+}
+
+func merge(a interface{}, b interface{}) (interface{}, error) {
```
stevendanna: This functions goes away in one of the other PRs in this series.

PR #1347, opened 2019-08-24
[lib/cereal] Add Parallel Workflow pattern

jaym:
The parallel workflow pattern can be used to combine multiple workflow
executors to run in parallel. These subworkflows run under one workflow
instance.

You may not nest parallel workflow executors.

This is based on top of #1346


lib/cereal/patterns/parallel.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+func (instance *immutableWorkflowInstanceImpl) Err() error {
+	if instance.state.Err != "" {
+		return errors.New(instance.state.Err)
+	}
+	return nil
+}
+
+func (instance *ParallelWorkflowInstance) GetSubWorkflow(workflowName string) (cereal.ImmutableWorkflowInstance, error) {
```
stevendanna: I'm not seeing where this gets used.

PR #1349, opened 2019-08-25
Add workflow for project rules update

jaym:
This PR replaces the current project update system with one driven by
workflows in cereal. The workflow is driven by the authz-service and
consists of 2 steps: First, authz commits the staged rules. After that
is done, the domain services update is started and pinged for status
every 10 seconds. These are written as 2 different workflows and tied
together using a combination of the Chain workflow pattern and Parallel
workflow pattern.

This is based on top of #1347


components/compliance-service/habitat/default.toml
```diff
@@ -22,4 +22,4 @@ manual_polling_interval = 1440
 [agent]
 workers = 10
 buffer_size = 1000
-remote_inspec_version = "REPLACE-FROM-PLAN.SH"
+remote_inspec_version = "4.7.24"
```
srenatus: :x: I don't think this should be committed. I do think I have commented this somewhere else already 😉 Probably the PR this is based on... 🤔

components/compliance-service/ingest/server/compliance.go
```diff
@@ -35,78 +29,33 @@ type ComplianceIngestServer struct {
...
-}
-
-func getProjectUpdateID(event *automate_event.EventMsg) (string, error) {
-	if event.Data != nil && event.Data.Fields != nil && event.Data.Fields["ProjectUpdateID"] != nil &&
-		event.Data.Fields["ProjectUpdateID"].GetStringValue() != "" {
-		return event.Data.Fields["ProjectUpdateID"].GetStringValue(), nil
-	}
-
-	return "", fmt.Errorf("Project Rule Update sent without a ProjectUpdateID eventID: %q", event.EventID)
+	return nil, status.Error(codes.Unimplemented, "Endpoint no longer used")
```
srenatus: Uhm... I can imagine situations where we'd care. However, don't we still deploy all services in one go, etc, and can thus just remove the method?

If we in fact **cannot**, we could also embed [`UnimplementedComplianceIngesterServer`](https://github.com/chef/automate/blob/d4e9c336fa73c8ebabf53043e13fdf4bcb91766e/components/compliance-service/ingest/ingest/compliance.pb.go#L218-L230) in `ComplianceIngestServer`, a nice little addition brought by the lastest go-grpc bump.

lib/authz/project_rules_update_workflow.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+func (m *ESStartProjectTagUpdaterTask) Run(
+	ctx context.Context, task cereal.Task) (interface{}, error) {
+	jobIDs, err := m.startProjectTagUpdater(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return StartResult{
+		IsComplete: false,
```
srenatus: [ignoreme] Superfluous but fine for being explicit.

components/compliance-service/habitat/default.toml
jaym: this one is annoying
`git checkout origin/master -- components/compliance-service/habitat/default.toml` is in my history, don't know when i rebuilt the thing

components/compliance-service/ingest/server/compliance.go
jaym: neat

components/authz-service/testhelpers/testhelpers.go
```diff
@@ -21,12 +19,12 @@ import (
 
 	api "github.com/chef/automate/api/interservice/authz/v2"
 	automate_event "github.com/chef/automate/api/interservice/event"
-	"github.com/chef/automate/components/authz-service/config"
 	"github.com/chef/automate/components/authz-service/engine"
 	"github.com/chef/automate/components/authz-service/engine/opa"
 	"github.com/chef/automate/components/authz-service/prng"
 	grpc_server "github.com/chef/automate/components/authz-service/server"
 	server "github.com/chef/automate/components/authz-service/server/v2"
+	v2 "github.com/chef/automate/components/authz-service/server/v2"
```
srenatus: This is already imported, under a different name, one line above.

components/authz-service/server/v2/project.go
```diff
@@ -191,9 +192,53 @@ func (s *ProjectState) ApplyRulesStart(
...
+			}
+		}
+		sleepTime := time.Duration((5 * (1 << uint(tries)))) * time.Millisecond
+		if sleepTime > time.Second {
+			sleepTime = time.Second
+		}
+		select {
+		case <-ctx.Done():
+			logctx.Error("Timed out waiting for ApplyStagedRules")
+			return status.Errorf(codes.DeadlineExceeded, "Timed out waiting for ApplyStagedRules")
```
srenatus: There's no format string here, let's use `status.Error()`

components/authz-service/server/v2/project_update_manager.go
```diff
@@ -20,6 +20,8 @@ import (
 	project_update_tags "github.com/chef/automate/lib/authz"
 )
 
+type ProjectUpdateStage string
```
srenatus: :+1: Why don't we do the same for `...State` and `...Name` below? 🤔

jaym:
@ryancragun the tests are getting ripped apart and rewritten by the auth team.

PR #1350, opened 2019-08-26
Streamline projects on team details page

msorens:

Initially I started to replace the sequence of GetProject (singular) calls
with just one GetProjects (plural) call, to reduce overhead and simplify the code.
Part way through that, though, I realized we do not even need GetProjects;
we already have enough project details from the global projects filter.

So this streamlines the team details code based on the above,
and adds some tests to show it is working.

image

Also addresses one UX issue:
If someone else deleted a project but the user has not refreshed the browser
lately, the global projects filter will still update periodically, but opening
team details looks like this, and the spinner remains indefinitely. Those projects are reporting errors because they are not amongst those in the current list (see above image).

image

Now, those deleted projects are automatically pruned from the team upon opening the details page.

PR - #1260 Update projects for a team

### :+1: Definition of Done

Everything still works, but faster.

Rebuild automate-ui
Go to Settings >> Teams >> <team> >> Details
Observe that the projects dropdown correctly populates and retains memory
of the projects belonging to the team.

### DIFF
--- a/components/automate-ui/src/app/pages/team-details/team-details.component.spec.ts
+++ b/components/automate-ui/src/app/pages/team-details/team-details.component.spec.ts
@@ -24,8 +24,6 @@ import {
   projectEntityReducer,
   ProjectEntityInitialState
 } from 'app/entities/projects/project.reducer';
-import { GetProjects, GetProjectsSuccess } from 'app/entities/projects/project.actions';
-import { Project } from 'app/entities/projects/project.model';
 import {
   userEntityReducer,
   UserEntityInitialState
@@ -179,7 +177,6 @@ describe('TeamDetailsComponent', () => {
       const version: IamVersionResponse = { version: { major: major, minor: 'v0' } };
       store.dispatch(new GetIamVersionSuccess(version));
 
-      expect(store.dispatch).toHaveBeenCalledWith(new GetProjects());
       expect(store.dispatch).toHaveBeenCalledWith(new GetUsers());
       expect(store.dispatch).toHaveBeenCalledWith(new GetTeamUsers({ id: targetId }));
     });
@@ -197,7 +194,7 @@ describe('TeamDetailsComponent', () => {
       expect(component.projects[p.value]).toBeTruthy();
       expect(component.projects[p.value].checked).toEqual(false);
     });
-    // But team's projects are still empty at this point!
+    // But team's projects are still empty at this point
     expect(component.team.projects.length).toEqual(0);
   });
 
@@ -218,23 +215,6 @@ describe('TeamDetailsComponent', () => {
     ];
     store.dispatch(new LoadOptionsSuccess({ fetched: projectOptionList, restored: [] }));
 
-    // That initializes the list for the team projects dropdown, all unchecked
-    expect(Object.keys(component.projects))
-      .toEqual(jasmine.arrayWithExactContents(projectOptionList.map(p => p.value)));
-    Object.keys(component.projects).forEach(id => {
-      expect(component.projects[id].checked).toEqual(false);
-    });
-
-    // Team details initiates GetProjects to fetch a hydrated list of projects
-    const projectList = [
-      genProject('a-proj'),
-      genProject('b-proj'),
-      genProject('c-proj'),
-      genProject('d-proj')
-    ];
-    store.dispatch(new GetProjectsSuccess({ projects: projectList }));
-
-    // And that is used to set checked true for any projects that the team already includes
     projectOptionList.forEach(p => {
       expect(component.projects[p.value].checked).toEqual(teamProjects.includes(p.value));
     });
@@ -341,15 +321,5 @@ describe('TeamDetailsComponent', () => {
       value: value ? value : label
     };
   }
-
-function genProject(id: string): Project {
-  return {
-    id,
-    status: 'NO_RULES', // unused
-    name: id, // unused
-    type: 'CUSTOM' // unused
-  };
-}
-
 
 });
--- a/components/automate-ui/src/app/pages/team-details/team-details.component.spec.ts
+++ b/components/automate-ui/src/app/pages/team-details/team-details.component.spec.ts
@@ -182,22 +182,6 @@ describe('TeamDetailsComponent', () => {
     });
   });
 
-  it('initializes dropdown with all assignable projects, each unchecked', () => {
-    const projectOptionList = [
-      genProjectOption('a-proj'),
-      genProjectOption('b-proj'),
-      genProjectOption('c-proj'),
-      genProjectOption('d-proj')
-    ];
-    store.dispatch(new LoadOptionsSuccess({ fetched: projectOptionList, restored: [] }));
-    projectOptionList.forEach(p => {
-      expect(component.projects[p.value]).toBeTruthy();
-      expect(component.projects[p.value].checked).toEqual(false);
-    });
-    // But team's projects are still empty at this point
-    expect(component.team.projects.length).toEqual(0);
-  });
-
   it('initializes dropdown with those included on the team checked', () => {
     const teamProjects = ['b-proj', 'd-proj'];
     const team: Team = { id: targetId, guid: 'any', name: 'any', projects: teamProjects };
--- a/components/automate-ui/src/app/pages/team-details/team-details.component.spec.ts
+++ b/components/automate-ui/src/app/pages/team-details/team-details.component.spec.ts
@@ -7,43 +7,54 @@ import { MockComponent } from 'ng2-mock-component';
 import { StoreModule, Store } from '@ngrx/store';
 
 import { NgrxStateAtom } from 'app/ngrx.reducers';
+import { using } from 'app/testing/spec-helpers';
+import {
+  projectsFilterReducer,
+  projectsFilterInitialState,
+  ProjectsFilterOption
+} from 'app/services/projects-filter/projects-filter.reducer';
+import { LoadOptionsSuccess } from 'app/services/projects-filter/projects-filter.actions';
 import {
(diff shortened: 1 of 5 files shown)


components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -233,29 +218,6 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
           });
       });
-
-    // we keep a list of all projects that were left in the team
-    // to know if we are fully loaded yet or not.
-    combineLatest([
-      this.store.select(getProjectStatus),
-      this.store.select(projectEntities)])
-      .pipe(filter(([status, _]: [EntityStatus, ProjectCheckedMap]) =>
-          status === EntityStatus.loadingSuccess),
```
msorens: This was actually a race condition, of sorts, too. We sent off requests for several projects--but they all use the same slot in the store for setting their returned status. Thus, we got someone's success, but don't know for whom!

components/automate-ui/src/app/pages/team-details/team-details.component.spec.ts
```diff
@@ -254,5 +312,14 @@ describe('TeamDetailsComponent', () => {
       });
     });
   });
-});
 
+  function genProjectOption(
+    label: string, checked?: boolean, value?: string): ProjectsFilterOption {
+    return <ProjectsFilterOption>{
+      label: label,
+      checked: checked === undefined ? false : checked,
```
srenatus: [style] same as `!!checked`, isn't it? But I guess that's more a ruby idiom than it is a JS/TS one...? 🤔

components/automate-ui/src/app/pages/team-details/team-details.component.spec.ts
```diff
@@ -254,5 +312,14 @@ describe('TeamDetailsComponent', () => {
       });
     });
   });
-});
 
+  function genProjectOption(
+    label: string, checked?: boolean, value?: string): ProjectsFilterOption {
+    return <ProjectsFilterOption>{
+      label: label,
```
srenatus: [nit] 
```suggestion
      label,
```

components/automate-ui/src/app/pages/team-details/team-details.component.spec.ts
```diff
@@ -254,5 +312,14 @@ describe('TeamDetailsComponent', () => {
       });
     });
   });
-});
 
+  function genProjectOption(
+    label: string, checked?: boolean, value?: string): ProjectsFilterOption {
+    return <ProjectsFilterOption>{
```
srenatus: ❓ Do we need the cast? Doesn't TS check that we return something that's valid, as declared in line 317?

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -148,19 +137,15 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
-          }
-        });
+        // compiles a list of all available projects
+        assignable
+          // don't override projects that were already part of the team
+          .filter(p => !this.projects[p.value])
+          .forEach(({ value: id, label: name, type: stringType }) => {
+            const type = <IAMType>stringType;
+            // type and status are unused
+            this.projects[id] = { id, name, type, status: 'NO_RULES', checked: false };
```
srenatus: [nit] The extra const seems superfluous, we could do
```suggestion
            this.projects[id] = { id, name, type: <IAMType>type, status: 'NO_RULES', checked: false };
```
...but it's also not hurting 😉

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -265,9 +227,13 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
     this.team = team;
-    this.team.projects.forEach(pID => {
-      this.teamProjectsLeftToFetch.push(pID);
-      this.store.dispatch(new GetProject({ id: pID }));
+    this.team.projects.forEach(id => {
+      if (this.projects[id]) {
+        this.projects[id].checked = true;
+      } else {
+        // type and status are unused
+        this.projects[id] = { id, name, type: 'CUSTOM', status: 'NO_RULES', checked: true };
```
srenatus: 🔍 where does `name` get defined here? I really don't see it. Weird.

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -333,8 +299,6 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
   }
 
   dropdownDisabled(): boolean {
-    return Object.values(this.projects).length === 0 ||
-      this.teamProjectsLeftToFetch.length !== 0 ||
-      this.saving;
+    return Object.values(this.projects).length === 0 || this.saving;
```
srenatus: ❓ I am aware that this was copied, but -- isn't there a more obvious way to check if `this.projects` is `{}`? That, to me, seems like the only wayt that `Object.values(this.projects).length` could be `0`. 🤔

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -135,38 +138,39 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
+      takeUntil(this.isDestroyed),
+      map(([assignable, pStatus, v1Team, v2Team, major]:
+        [ProjectsFilterOption[], EntityStatus, Team, Team, IAMMajorVersion]) => {
+        this.isMajorV1 = major === 'v1';
+        return [assignable, pStatus, (this.isMajorV1 ? v1Team : v2Team)];
+      }),
+      filter(([_assignable, pStatus, team]:
+        [ProjectsFilterOption[], EntityStatus, Team]) =>
+        pStatus !== EntityStatus.loading && !!team
+      ),
```
srenatus: [nit] I find this a little hard to read. I guess I'd put the types on the same line -- just a matter of preference, though 😄

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -135,38 +138,39 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
+      map(([assignable, _pStatus, team]) => {
```
srenatus: ❓ Why do we not need a type here?

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -135,38 +138,39 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
+      this.store.select(optionsLoadingStatus),
+      this.store.select(v1TeamFromRoute),
+      this.store.select(v2TeamFromRoute),
+      this.store.select(iamMajorVersion)
+    ]).pipe(
+      takeUntil(this.isDestroyed),
+      map(([assignable, pStatus, v1Team, v2Team, major]:
+        [ProjectsFilterOption[], EntityStatus, Team, Team, IAMMajorVersion]) => {
+        this.isMajorV1 = major === 'v1';
+        return [assignable, pStatus, (this.isMajorV1 ? v1Team : v2Team)];
```
srenatus: 🤔 If we used an object (and an interface) here, we could reduce the duplications a bit, I think...like this:
```typescript
map(([assignable, pStatus, v1Team, v2Team, major]: [ProjectsFilterOption[], EntityStatus, Team, Team, IAMMajorVersion]) => {
    this.isMajorV1 = major === 'v1';
    return { assignable, pStatus, team: (this.isMajorV1 ? v1Team : v2Team) };
}),
filter({ pStatus } => pStatus !== EntityStatus.loading), // ignore while loading
filter({ team }) => !!team), // ignore while undefined
map({ assignable, team) => {
  this.team = team;
  assignable.forEach(...)
})
```
(I've omitted the interface, I think we'd want a type there somewhere...) -- Not sure if this is better, just pointing out an alternative (which I haven't tested ⚠️)

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -135,38 +138,39 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
-          this.store.select(v2TeamFromRoute)
-            .pipe(filter(identity), takeUntil(this.isDestroyed))
-            .subscribe(this.getTeamDependentData.bind(this));
-        }
-      });
+        this.updateNameForm.controls.name.setValue(this.team.name);
+        this.store.dispatch(new GetTeamUsers({ id: this.teamId }));
+        this.store.dispatch(new GetUsers());
+      }))
+      .subscribe();
```
srenatus: Uhm... couldn't we do the last `map()`'s contents in the `subscribe()`? This looks a bit odd 😉 but it might just be me.

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -146,42 +138,39 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
+      filter(([_assignable, pStatus, team]:
+        [ProjectsFilterOption[], EntityStatus, Team]) =>
+        pStatus !== EntityStatus.loading && !!team
+      ),
+      map(([assignable, _pStatus, team]) => {
+        this.team = team;
+
+        assignable
+          .forEach(({ value: id, label: name, type: stringType }) => {
+            const type = <IAMType>stringType;
```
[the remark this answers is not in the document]
srenatus: because that's not the type of `stringType`, it's string 😉 We'll have to do some kind of casting, or, maybe better yet, create a type guard for the string => IAMType conversion 🤔

Issue #112, opened 2019-04-18
Add Confirmation of Removal to Node Credentials


vjeffrey:
# User Story

original issue: https://chefio.atlassian.net/secure/RapidBoard.jspa?rapidView=261&modal=detail&selectedIssue=UI-164

# Overview
Currently, there is no confirmation modal when removing a node credential; this story adds one.

# Design Details
[Zeplin](https://zpl.io/bldRgXD)
- Leverage the delete confirmation on the other settings pages (Users, Tokens, etc)
- Heading: Remove Node Credential?
- Message: Removing node credential: \<credential name\> cannot be undone.
- Primary destructive button: Remove Credential
- Tertiary button: Cancel
- Close button: close icon

==> Search for <app-delete-object-modal> in the code base and you will find 6 examples (at last count) of how this reusable component may be used for this task.

# Accessibility Details
Heading should use an h2

## Tab Order
1. close button
2. Remove Credential button
3. Cancel button
4. go to 1.

PR #1351, opened 2019-08-26
added changes to show confirmation box to remove node cred

vinay033:

Currently, there is no confirmation modal when removing a node credential, I have added changes for the confirmation modal, if the user clicks to delete the node credential then confirmation modal will be displayed same like user and API token.

https://github.com/chef/automate/issues/112

### :camera: Screenshots
Screenshot from 2019-08-26 14-46-54


components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.html
```diff
@@ -1,6 +1,7 @@
 <chef-modal [visible]="visible" (close)="closeEvent()">
-  <h2 slot="title">Delete {{objectNoun | titlecase}}?</h2>
-  <div>
+  <h2 *ngIf="!nodeCredObject" slot="title">Delete {{objectNoun | titlecase}}?</h2>
+  <h2 *ngIf="nodeCredObject" slot="title">Remove Node {{objectNoun | titlecase}}?</h2>
```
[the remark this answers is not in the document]
vinay033: I have added changes, please have a look.

components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.html
```diff
@@ -1,6 +1,6 @@
 <chef-modal [visible]="visible" (close)="closeEvent()">
-  <h2 slot="title">Delete {{objectNoun | titlecase}}?</h2>
-  <div>
+  <h2 slot="title">{{ObjectAction}} {{objectNoun | titlecase}}?</h2>
+  <div *ngIf="!nodeCredObject">
```
bcmdarroch: Thanks for adding `ObjectAction`! I've got one more ask: let's drop `nodeCredObject` from this component and use a pipe [like this](https://github.com/chef/automate/blob/master/components/automate-ui/src/app/pipes/pluralize.pipe.ts) instead. I think this would work (though I haven't tested it):
```
export class ContinuousPipe implements PipeTransform {

  transform(value: string): string {
    const lastIndex = value.length - 1
    if (value) {
      if (value.charAt(lastIndex) == 'e') {
        return value.slice(0, lastChar - 1) + 'ing';
      } else {
        return value + 'ing';
      }
    }
    return value;
  }
}
```

Then we use `{{ ObjectAction | continuous }}` in the template

vinay033: I have implemented pipe the same as you suggested, please have a look.

components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.ts
```diff
@@ -1,16 +1,18 @@
...
   templateUrl: './delete-object-modal.component.html',
   styleUrls: ['./delete-object-modal.component.scss']
 })
-export class DeleteObjectModalComponent {
+export class DeleteObjectModalComponent extends ContinuousPipe {
   @Input() visible = false;
   @Input() objectNoun: string;
   @Input() objectName: string;
   @Input() moreDetails: string; // additional details after "action cannot be undone"
+  @Input() ObjectAction: string;
```
bcmdarroch: last small nitpick: 
```suggestion
  @Input() objectAction: string;
```

components/automate-ui/src/app/components/chef-components.module.ts
```diff
@@ -37,13 +37,15 @@ import { SidebarNoShadeComponent } from './sidebar-no-shade/sidebar-no-shade.com
 import { SidebarEntryComponent } from './sidebar-entry/sidebar-entry.component';
 import { TabComponent } from './tab/tab.component';
 import { TabsComponent } from './tabs/tabs.component';
+import { ChefPipesModule } from '../pipes/chef-pipes.module';
```
bcmdarroch: I'm not sure why that needs to be imported there 🤔

vinay033: because if do not import this then getting error like below.
```
ERROR in : The pipe 'continuous' could not be found ("
  <div>
    <span *ngIf="moreDetails">
      {{ [ERROR ->]ObjectAction | continuous }} {{objectNoun}} {{objectName}} <strong>cannot be undone</strong><br>
    ")
: The pipe 'continuous' could not be found ("
    </span>
    <span *ngIf="!moreDetails">
      {{ [ERROR ->]ObjectAction | continuous }} {{objectNoun}} {{objectName}} <strong>cannot be undone</strong>.
    </s")

```

PR #1352, opened 2019-08-26
[#1318] proper logout for intercepted 401s

### LINKED ISSUE #1318: no auto-login for SAML users on IAMv2.1 when session expires
_Problem: Users do not want to have re-login after their SAML session has expired when using iamv2._

## User Story 

As an Automate User, I will use this work when I am logged in via SAML.

In [this PR](https://github.com/chef/automate/pull/1246), we updated session-service to keep track of the user's chosen login connector (SAML, LDAP, local, etc). While on IAM v1 and IAM v2, the following expected behavior occurs:
- I log in via SAML
- after some time, the id token expires
- on expiration, I am automatically re-logged in to Automate. I do not have to interact with login screen at all.

While on IAM v2.1, this doesn't happen. On token expiry, the user is logged out. They must click SAML and enter their credentials to log in again.

## Definition of Done
- the expected session behavior works regardless of IAM version

## Demo Script / Repro Steps
- update `components/automate-dex/habitat/templates/config.yml` with `idTokens: 1m`
- `rebuild components/automate-dex`
- log in as a SAML user
- wait a minute for the token expire
- get logged back in automatically

srenatus:

When a 401 was intercepted by the auth interceptor, and not the periodic call in chef-session, the interceptor would call `logout()` without arguments. This used to have two effects:
1. the user is sent back to `/` (or rather, `/dashboards/event-feed`) after re-signin, regardless of where they were before
2. the user has to select their signin method.

This became apparent most obviously when using IAMv2.1, because it has another peridoc call: checking project rules' apply status. So those calls, when failing, would cause no automatic re-login for SAML users, as desired in #1246.

**Now**, the logic is inverted: If nothing is passed, the signin happens as if it was a re-signin: including the proper state for going back to the page, and skipping signin method selection if possible.

⚠️ The way `chef-session` figures out "which page we're on" is hacky IMHO; it should rather subscribe to something in ngrx store, instead of reading a `window` global. (I might be wrong.)

- #1246
- Fixes #1318

- `start_all_services`, setup automate-ui in your preferred way
- `chef-automate iam upgrade-to-v2 --beta2.1`
- change the dex config template as outlined in #1246
- sign in locally using SAML, go to https://a2-dev.test/settings
  - you will see nothing, but it's the test to verify we end up on the same page after re-signin
- wait for a minute
- notice that there's a few requests, and a roundtrip through dex and okta, and you'll end up on the same page again

- [X] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/services/chef-session/chef-session.service.ts
+++ b/components/automate-ui/src/app/services/chef-session/chef-session.service.ts
@@ -126,7 +126,7 @@ export class ChefSessionService implements CanActivate {
   // available for setSessionOrRedirectToLogin() (part of ChefSession's
   // constructor)
   setSession(uuid, fullname, username, id_token: string, groups: Array<string>): void {
-    this.user = <ChefSessionUser>{
+    this.user = {
       uuid,
       fullname,
       username,
@@ -146,11 +146,14 @@ export class ChefSessionService implements CanActivate {
     return !isNull(localStorage.getItem(sessionKey));
   }
 
-  logout(url: string = '/', refresh: boolean = false): void {
+  // url: UI route to go back to when the (next) signin process has succeeded
+  // noHint: for the sign in, don't try to skip the method selection
+  logout(url?: string, noHint?: boolean): void {
     this.deleteSession();
+    url = url || this.currentPath();
     // note: url will end up url-encoded in this string (magic)
     let signinURL: string;
-    if (refresh && this.user && this.user.id_token) {
+    if (!noHint && this.user && this.user.id_token) {
       signinURL = `/session/new?state=${url}&id_token_hint=${this.user.id_token}`;
     } else {
       signinURL = `/session/new?state=${url}`;
@@ -166,6 +169,9 @@ export class ChefSessionService implements CanActivate {
     }
   }
 
+  // TODO(sr) 2019/08/26: I don't think we should use these global variables.
+  // Instead, we should take the information about the currently-viewed page
+  // from the ngrx store.
   currentPath(): string {
     return window.location.pathname;
   }
--- a/components/automate-ui/src/app/services/chef-session/chef-session.service.ts
+++ b/components/automate-ui/src/app/services/chef-session/chef-session.service.ts
@@ -75,7 +75,7 @@ export class ChefSessionService implements CanActivate {
       if (xhr.status === HTTP_STATUS_OK) {
         this.ingestIDToken(xhr.response.id_token);
       } else if (xhr.status === HTTP_STATUS_UNAUTHORIZED) {
-        this.logout(this.currentPath(), true);
+        this.logout();
       } else {
         // TODO 2017/12/15 (sr): is there anything we could do that's better than
         // this?
--- a/components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
+++ b/components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
@@ -44,12 +44,7 @@ export class HttpClientAuthInterceptor implements HttpInterceptor {
       .handle(request.clone({ headers })).pipe(
         catchError((error: HttpErrorResponse) => {
           if (error.status === 401) {
-            // TODO(sr) When refactoring the session handling in automate-ui and ngrx,
-            // this should be dealt with: right now, if any API request gets a 401,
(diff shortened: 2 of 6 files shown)


components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
```diff
@@ -29,20 +34,19 @@ export class HttpClientAuthInterceptor implements HttpInterceptor {
...
     // const params = request.params.delete('unfiltered');
     if (this.projects && filtered) {
       headers = headers.set('projects', this.projects);
     }
     return next
-      .handle(request.clone({
-        headers
-      })).pipe(
-        catchError((response: HttpEvent<any>) => {
-          if (get('status', response) === 401) {
```
srenatus: This `get('status', response)` was a wonky way to side-step type checks, I believe. `HttpEvent<any>` doesn't have a status propery, but it's a type union that contains `HttpErrorResponse`. By using the more specific type (the latter) here, we can call `response.status` below.

components/automate-ui/src/app/services/chef-session/chef-session.service.ts
```diff
@@ -146,11 +146,14 @@ export class ChefSessionService implements CanActivate {
     return !isNull(localStorage.getItem(sessionKey));
   }
 
-  logout(url: string = '/', refresh: boolean = false): void {
+  // url: UI route to go back to when the (next) signin process has succeeded
+  // noHint: for the sign in, don't try to skip the method selection
+  logout(url?: string, noHint?: boolean): void {
```
srenatus: 👉 The crucial bit is that the http auth interceptor **hasn't changed**: [we're still calling `logout()` there](https://github.com/chef/automate/blob/8a5d470bcb929cf6d6ca88fdc2f120cb1e35dd5b/components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts#L43), but it means something slightly different now.

srenatus:
@msorens thanks for playing with this. I'm afraid I don't follow here:
> (different URL params and on screen param, of course)

Re: the videos of your attempts, 

#### Attempt 4 

The network tab shows
image
which is what we'd expect with this PR -- the failing request is the periodic one (`GET /apply-rules`: 401), and the browser then goes to `/session/new?state=/settings&id_token_hint=YOURIDTOKEN`. As far as `automate-ui` and this PR is concerned, that's what we want to see happen.

The question then is: why did you see the local user login? I can't tell from the video alone: Did you use a SAML user? What's the `id_token`? Anything in the logs of session-service? What are the query parameters of the request to `/dex/auth`?

The fact that the request to dex's `/auth` endpoint redirects you to `/auth/local`
image

 means one of two things:
1. there was a `connector_id=local` query parameter to `/dex/auth` -- which would mean your `id_token` wasn't the one of a SAML user, but a local one
2. there was only one dex connector set up -- so it would automatically send to you the one and only. This is unlikely, because then, we shouldn't see the `Back` link in the top left.

I'd conclude that your `id_token` must not be the one of a SAML user. I suppose it wasn't, since you've been testing this PR 😄

👉 **Can you reproduce this?**

#### Attempt 1 and 7

There's no data to attempt 1, making it hard to assess; however, this error has -- unfortunately! -- nothing to do with the change in this PR: it's related to `session-service` and its cookies. (Those cookies are http-only, which means that our UI code in `automate-ui` can never see them. 🙈)

However, there are multiple tabs involved -- one logging out, which starts **one** session dance, and a fresh one, which starts a **second** session dance. I would have thought that this set of issues had been resolved in #716. This is something to investigate more throughly. We'd need the database contents to see what had happened here; I'll try to reproduce this on my end.

srenatus:
So, letting this one some more, I've noticed that it's racy 😞 

There's two requests getting a 401 when the token is expired, and they cause a call to `logout()`:
image

However, only one includes the `id_token_hint`, and the second one of them is canceling the first. So depending on their order, the right thing happens, or doesn't. (In the screenshot, the order was the lucky one, but it seems just as likely that the order is reversed.)

I'm looking into how to fix this. We might have to push this down the road for the chef-session refactor after all... 🤔

srenatus:
ℹ️ the last commit resolves the last comment I've posted above ☝️ 

@msorens I'm still unsure about your _relay state_ issues, and I haven't been able to reproduce these...

PR #1363, opened 2019-08-27
[lib/cereal] Delete task result when done with it

jaym:
A task result is only seen by the workflow once. It is safe to delete
once we've processed the workflow event that references it. This should
keep the table small for cases where there are long running workflows that
poll

### DIFF
--- a/lib/cereal/postgres/postgres.go
+++ b/lib/cereal/postgres/postgres.go
@@ -29,11 +29,10 @@ const (
 	//
 	defaultMaxIdleConnections = 4
 
-	enqueueTaskQuery      = `SELECT cereal_enqueue_task($1, $2, $3, $4)`
-	dequeueTaskQuery      = `SELECT * FROM cereal_dequeue_task($1)`
-	completeTaskQuery     = `SELECT cereal_complete_task($1, $2, $3, $4)`
-	getTaskResultQuery    = `SELECT task_name, parameters, status, error, result FROM cereal_task_results WHERE id = $1`
-	deleteTaskResultQuery = `DELETE FROM cereal_task_results WHERE id = $1`
+	enqueueTaskQuery       = `SELECT cereal_enqueue_task($1, $2, $3, $4)`
+	dequeueTaskQuery       = `SELECT * FROM cereal_dequeue_task($1)`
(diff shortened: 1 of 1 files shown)


lib/cereal/postgres/postgres.go
```diff
@@ -29,10 +29,11 @@ const (
...
 
-	enqueueTaskQuery   = `SELECT cereal_enqueue_task($1, $2, $3, $4)`
-	dequeueTaskQuery   = `SELECT * FROM cereal_dequeue_task($1)`
-	completeTaskQuery  = `SELECT cereal_complete_task($1, $2, $3, $4)`
-	getTaskResultQuery = `SELECT task_name, parameters, status, error, result FROM cereal_task_results WHERE id = $1`
+	enqueueTaskQuery      = `SELECT cereal_enqueue_task($1, $2, $3, $4)`
+	dequeueTaskQuery      = `SELECT * FROM cereal_dequeue_task($1)`
+	completeTaskQuery     = `SELECT cereal_complete_task($1, $2, $3, $4)`
+	getTaskResultQuery    = `SELECT task_name, parameters, status, error, result FROM cereal_task_results WHERE id = $1`
+	deleteTaskResultQuery = `DELETE FROM cereal_task_results WHERE id = $1`
```
srenatus: [style] There's that weird habbit of always using backticks for SQL.... It's not getting us anything if there's neither newlines nor quotes involved. 🤷‍♂ Fine.

lib/cereal/postgres/postgres.go
```diff
@@ -722,6 +723,10 @@ func (pg *PostgresBackend) DequeueWorkflow(ctx context.Context, workflowNames []
 			return nil, nil, errors.Wrap(err, "failed to retrieve task result for task complete event")
 		}
 		event.TaskResult = tr
+		if err := deleteTaskResult(ctx, tx, taskResultID); err != nil {
+			cancel()
+			return nil, nil, errors.Wrap(err, "failed to remove task result from the database")
```
srenatus: I guess I'm in the camp of _less is more_, if it's an error, that something has failed is implicit:
```suggestion
			return nil, nil, errors.Wrap(err, "remove task result from the database")
```
**but** you folks seem to be in the other camp. That's cool.

PR #1374, opened 2019-08-28
Enable control_tag filtering for compliance API calls

alexpop:
This PR enables `control_tag` filters for the compliance APIs.
I updated test data as well to support integration testing of this new filter.

This is an example where we filter for compliance reports that have at least one control with tagged `scope:nginx` or `scope:apache`, case insensitive:

```
Reporting::ListFilter.new(type: 'control_tag:scope', values: ['nginx', 'apache'])
```

### DIFF
--- a/components/compliance-service/reporting/relaxting/reports.go
+++ b/components/compliance-service/reporting/relaxting/reports.go
@@ -657,6 +657,15 @@ func (backend ES2Backend) getFiltersQuery(filters map[string][]string, latestOnl
 		boolQuery = boolQuery.Must(termQuery)
 	}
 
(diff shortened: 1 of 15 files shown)


components/compliance-service/reporting/relaxting/reports.go
```diff
@@ -126,36 +126,6 @@ func (backend ES2Backend) getNodeReportIdsFromTimeseries(esIndex string,
 		}
 	}
 
-	//When filtering by controls, we are reducing the array of report ids based on
```
alexpop: I removed this block as it's no longer needed. We are using only detailed indices with all filters being handled by `getFiltersQuery`

PR #1375, opened 2019-08-28
deps: bump grpc-gateway to 1.10.0

srenatus:
[Release notes](https://github.com/grpc-ecosystem/grpc-gateway/releases/tag/v1.10.0)

The new feature, skipping the trip through localhost, is worth exploring
in a follow-up PR.


stevendanna:
Very interesting indeed https://github.com/grpc-ecosystem/grpc-gateway/pull/947

srenatus:
I don't suppose we can use this as-is, though, since we're prefixing the paths using the `http.ServeMux` stuff...but there might be ways around that... 🤔

Issue #796, opened 2019-07-08
Tokens: Add projects form input to update and create


susanev:
_Note: This story is dependent on #795 happening first._

## Overview

This task includes adding the projects form input to the create token modal and to the Details tab on token detail pages.

## Important Notes

* On create the projects to choose from should be filtered by any filters the user has applied in the global projects filter (we will revisit this and improve the ux in the future)
* We will revisit the case when a user has access to zero projects in the future
* If the user tries to update a resource's projects and that resource has projects the user does not have access to
* We should use the project picker designed by Maggie for this - components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.html (see [https://github.com/chef/a2/pull/5485](a2 PR #5485))

## Design Details

[Zeplin](https://app.zeplin.io/project/5ba808703eb54372d860c737?seid=5caa8eeb855cd9353ea65446)

### Add a Projects column to the Tokens list view
* It should come directly after the ID column
* When a token is not assigned to any projects display Unassigned
* When a token is assigned to exactly one project display that project's name
* When a token is assigned to two or more projects display <n> projects

### On the Create Token modal add the Projects multi-select form input below the ID input

* This input is optional, so once a name is provided the Create Token button should be enabled

### On the Details tab on team detail pages...

* Add the Projects label, and content in the top header, this value should match whatever is shown in the list view
* Add the Projects multi-select form input below the name input
* There should be 24px of space between the projects input and the Save button

PR #1380, opened 2019-08-28
[automate-796] projects in the tokens UI

bcmdarroch:
Now you can create a token with projects and update their projects from the UI. :)

https://github.com/chef/automate/issues/796

### :+1: Definition of Done
- [x] can create a token with projects
- [x] can update a token's projects

- upgrade to v2.1, start the UI
- create some projects at https://a2-dev.test/settings/projects
- navigate to https://a2-dev.test/settings/tokens
- create a token with projects
- update that token's projects

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
@@ -54,8 +54,8 @@ export class CreateObjectModalComponent implements OnInit, OnChanges {
 
   onProjectChecked(project: ProjectChecked): void {
     this.projects[project.id].checked = project.checked;
-    const projectsSelected = Object.values(this.projects).filter(p => p.checked);
-    this.createForm.controls.projects.setValue(projectsSelected.map(p => p.id));
+    const projectsSelected = Object.keys(this.projects).filter(id => this.projects[id].checked);
+    this.createForm.controls.projects.setValue(projectsSelected);
   }
 
   dropdownDisabled(): boolean {
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
@@ -54,7 +54,7 @@ export class CreateObjectModalComponent implements OnInit, OnChanges {
 
   onProjectChecked(project: ProjectChecked): void {
     this.projects[project.id].checked = project.checked;
-    let projectsSelected = Object.values(this.projects).filter(p => p.checked);
+    const projectsSelected = Object.values(this.projects).filter(p => p.checked);
     this.createForm.controls.projects.setValue(projectsSelected.map(p => p.id));
   }
 
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
@@ -54,6 +54,8 @@ export class CreateObjectModalComponent implements OnInit, OnChanges {
 
   onProjectChecked(project: ProjectChecked): void {
     this.projects[project.id].checked = project.checked;
+    let projectsSelected = Object.values(this.projects).filter(p => p.checked);
+    this.createForm.controls.projects.setValue(projectsSelected.map(p => p.id));
   }
 
   dropdownDisabled(): boolean {
@@ -63,8 +65,8 @@ export class CreateObjectModalComponent implements OnInit, OnChanges {
   public handleNameInput(event: KeyboardEvent): void {
     if (!this.modifyID && !this.isNavigationKey(event)) {
       this.conflictError = false;
-      this.createForm.controls['id'].setValue(
-        IdMapper.transform(this.createForm.controls['name'].value.trim()));
+      this.createForm.controls.id.setValue(
+        IdMapper.transform(this.createForm.controls.name.value.trim()));
     }
   }
 
@@ -81,7 +83,7 @@ export class CreateObjectModalComponent implements OnInit, OnChanges {
   }
 
   createObject(): void {
-    this.createClicked.emit(Object.values(this.projects).filter(p => p.checked));
+    this.createClicked.emit();
   }
 
   private isNavigationKey(event: KeyboardEvent): boolean {
--- a/components/automate-ui/src/app/entities/api-tokens/api-token.requests.ts
+++ b/components/automate-ui/src/app/entities/api-tokens/api-token.requests.ts
@@ -81,21 +81,17 @@ export class ApiTokenRequests {
     }
   }
 
-  public create(tokenData: CreateTokenPayload, version: IAMMajorVersion = 'v1'):
+  public create({ id, name, projects }: CreateTokenPayload, version: IAMMajorVersion = 'v1'):
     Observable<ApiTokenV1 | TokenPayloadResponse> {
     switch (version) {
       case 'v2': {
         return this.http.post<TokenPayloadResponse>(`${env.auth_v2_url}/tokens`,
-          {
-            id: tokenData.id,
-            name: tokenData.name,
-            active: true,
-            projects: tokenData.projects
-          });
+          { id, name, projects, active: true }
+        );
       }
       default: {
         return this.http.post<ApiTokenV1>(`${env.auth_url}/tokens`,
-          { id: tokenData.id, description: tokenData.name, active: true });
+          { id, description: name, active: true });
       }
     }
   }
--- a/components/automate-ui/src/app/entities/api-tokens/api-token.requests.ts
+++ b/components/automate-ui/src/app/entities/api-tokens/api-token.requests.ts
@@ -86,7 +86,7 @@ export class ApiTokenRequests {
     switch (version) {
       case 'v2': {
         return this.http.post<TokenPayloadResponse>(`${env.auth_v2_url}/tokens`,
-          { 
+          {
             id: tokenData.id,
             name: tokenData.name,
             active: true,
--- a/components/automate-ui/src/app/entities/api-tokens/api-token.requests.ts
+++ b/components/automate-ui/src/app/entities/api-tokens/api-token.requests.ts
@@ -5,7 +5,7 @@ import { Observable } from 'rxjs';
 import { environment as env } from 'environments/environment';
 import { IAMMajorVersion } from 'app/entities/policies/policy.model';
 import { ApiToken, ApiTokenV1, isApiTokenV1, ensureApiTokenV2 } from './api-token.model';
-import { GetAllTokensSuccessPayload } from './api-token.actions';
+import { CreateTokenPayload, GetAllTokensSuccessPayload } from './api-token.actions';
 
 export interface TokenPayloadResponse {
   token: ApiToken;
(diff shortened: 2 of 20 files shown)


components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
```diff
@@ -54,6 +54,8 @@ export class CreateObjectModalComponent implements OnInit, OnChanges {
 
   onProjectChecked(project: ProjectChecked): void {
     this.projects[project.id].checked = project.checked;
+    const projectsSelected = Object.values(this.projects).filter(p => p.checked);
+    this.createForm.controls.projects.setValue(projectsSelected.map(p => p.id));
```
srenatus: This is a bit indirect.... assuming that the _keys_ are the project IDs, we could do
```typescript
const projectsSelected = Object.keys(this.projects).filter(id => this.projects[id].checked);
this.createForm.controls.projects.setValue(projectsSelected);
```
or am I missing something?

components/automate-ui/src/app/entities/api-tokens/api-token.requests.ts
```diff
@@ -81,16 +81,21 @@ export class ApiTokenRequests {
     }
   }
 
-  public create(id: string, name: string, version: IAMMajorVersion = 'v1'):
+  public create(tokenData: CreateTokenPayload, version: IAMMajorVersion = 'v1'):
```
srenatus: [nit] if we destructure this here,
```suggestion
  public create({ id, name , projects }: CreateTokenPayload, version: IAMMajorVersion = 'v1'):
```
we can pass `{ id, name, projects, active: true }` below.

components/automate-ui/src/app/pages/api-token/details/api-token-details.component.ts
```diff
@@ -62,6 +81,31 @@ export class ApiTokenDetailsComponent implements OnInit, OnDestroy {
...
         this.store.dispatch(new GetToken({ id }));
       });
+
+    combineLatest([
+      this.store.select(iamMajorVersion),
+      this.store.select(iamMinorVersion)]).pipe(takeUntil(this.isDestroyed))
+        .subscribe(([major, minor]: [IAMMajorVersion, IAMMinorVersion]) => {
+          this.isMajorV1 = major === 'v1';
+          this.isMinorV1 = minor === 'v1';
+        });
```
srenatus: ❓UI-related question: Why are these instance variables and not `Observable<boolean>`, create from a `filter()` that's part of the `pipe()` args above?

components/automate-ui/src/app/pages/api-token/details/api-token-details.component.ts
```diff
@@ -62,6 +81,31 @@ export class ApiTokenDetailsComponent implements OnInit, OnDestroy {
...
+      this.store.select(allProjects),
+      this.store.select(getAllProjectStatus)
+    ]).pipe(
+      takeUntil(this.isDestroyed),
+      filter(([_, pStatus]: [Project[], EntityStatus]) => pStatus !== EntityStatus.loading),
+      filter(() => !!this.token),
+      map(([allowedProjects, _]) => {
+        allowedProjects
+          .forEach(p => {
+            this.projects[p.id] = {
```
srenatus: 💭 Would projects ever be removed from this? Looks like we're only adding/overwriting keys here -- I might be missing something.

components/automate-ui/src/app/pages/api-token/details/api-token-details.component.ts
```diff
@@ -74,7 +118,8 @@ export class ApiTokenDetailsComponent implements OnInit, OnDestroy {
     this.saveInProgress = true;
     const name: string = this.updateForm.controls.name.value.trim();
     const active = <TokenStatus>this.updateForm.controls.status.value === 'active';
```
srenatus: ❓I don't see how this `<TokenStatus>` cast matters here. We're comparing to another string, which happens to be `'active'`, one of the possible `TokenStatus` values (👉 string literal types); but we're building a boolean.
```suggestion
    const active: boolean = this.updateForm.controls.status.value === 'active';
```

components/automate-ui/src/app/pages/api-token/details/api-token-details.component.ts
```diff
@@ -98,4 +143,24 @@ export class ApiTokenDetailsComponent implements OnInit, OnDestroy {
...
+
+    // since the app-projects-dropdown is not a true form input (select)
+    // we have to manage the form reactions
+    this.updateForm.controls.projects.setValue(projectsSelected.map(p => p.id));
+    this.updateForm.controls.projects.markAsDirty();
+  }
+
+  noProjectsUpdated(): boolean {
+    return xor(this.token.projects,
+      Object.values(this.projects).filter(p => p.checked).map(p => p.id)).length === 0;
```
srenatus: [nit] Same as above; I might be missing something 🤔 
```suggestion
      Object.keys(this.projects).filter(id => this.projects[id].checked).length === 0;
```

components/automate-ui/src/app/pages/api-token/details/api-token-details.component.ts
```diff
@@ -98,4 +143,24 @@ export class ApiTokenDetailsComponent implements OnInit, OnDestroy {
...
+  }
+
+  dropdownDisabled(): boolean {
+    return Object.values(this.projects).length === 0 || this.saveInProgress;
```
srenatus: 💭 I've commented on this before. How would `this.projects` ever satisfy `Object.values(this.projects).length === 0` unless it was empty, i.e. `{}`? Unless I'm mistaken, lodash has `isEmpty`...
```suggestion
    return isEmpty(this.projects) || this.saveInProgress;
```

components/automate-ui/src/app/pages/api-token/list/api-token-list.component.html
```diff
@@ -14,6 +14,8 @@
       [conflictErrorEvent]="conflictErrorEvent"
       objectNoun="token"
       [createForm]="createTokenForm"
+      [showProjectsDropdown]="!isMajorV1 && isMinorV1"
```
srenatus: We're using `!isMajorV1 && isMinorV1` in thre places. We're doing this to exclude 1.0 and 2.0, and keeping stuff intact in case there's a 2.2 some day? hrm OK...

components/automate-ui/src/app/pages/api-token/list/api-token-list.component.ts
```diff
@@ -33,14 +36,22 @@ export class ApiTokenListComponent implements OnInit {
...
   public deleteModalVisible = false;
   public tokenToDelete: ApiToken;
   public createModalVisible = false;
   public createTokenForm: FormGroup;
   public creatingToken = false;
   public conflictErrorEvent = new EventEmitter<boolean>();
 
+  public iamMajorVersion$: Observable<IAMMajorVersion>;
+  public iamMinorVersion$: Observable<IAMMinorVersion>;
+  public isMajorV1 = true;
```
srenatus: Same here, I'd expect these to be observables, too. Might be my fault 😄

components/automate-ui/src/app/pages/api-token/list/api-token-list.component.ts
```diff
@@ -65,13 +76,40 @@ export class ApiTokenListComponent implements OnInit {
...
+    this.iamMinorVersion$ = store.pipe(select(iamMinorVersion));
   }
 
   ngOnInit() {
     this.store.dispatch(new GetAllTokens());
+
+    this.iamMajorVersion$
+      .pipe(takeUntil(this.isDestroyed))
+      .subscribe((majorVersion) => {
+        if (majorVersion === null) { return; }
```
srenatus: let's add `filter(identity)` to the pipe instead, it's the thing commonly done, I believe.

components/automate-ui/src/app/pages/api-token/details/api-token-details.component.ts
msorens: That was my first instinct, too, especially since they are not used in the ts file. But it makes the template simpler, so we can say
```
<td class="header-column" *ngIf="!isMajorV1 && isMinorV1">
```
instead of
```
<td class="header-column" *ngIf="!(isMajorV1$ | async) && (isMinorV1$ | async)">
```

components/automate-ui/src/app/pages/api-token/details/api-token-details.component.ts
msorens: Good catch; fixed.

components/automate-ui/src/app/pages/api-token/details/api-token-details.component.ts
msorens: It does add some compile-time safety, as it turns out. With that present, the compiler disallows misspellings of 'active' !

components/automate-ui/src/app/pages/api-token/details/api-token-details.component.ts
msorens: A team could come in with no projects, in which case it would be empty. And, with bug fix above, you can also now edit the project list to be empty, too.

components/automate-ui/src/app/pages/api-token/details/api-token-details.component.ts
srenatus: That's true for these two, but if we only ever use this precise combination, we could also make it a `atLeast2p1$` and use it as `*ngIf="atLeast2p1$ | async"`.

components/automate-ui/src/app/pages/api-token/details/api-token-details.component.ts
srenatus: Ah, right. Well, I guess that's fair, a microscopic, but actual, advantage 😉

components/automate-ui/src/app/pages/api-token/details/api-token-details.component.ts
srenatus: I see. thanks!

components/automate-ui/src/app/pages/api-token/details/api-token-details.component.ts
msorens: Yes, I like that. 👍

components/automate-ui/src/app/entities/policies/policy.selectors.ts
```diff
@@ -21,6 +21,12 @@ export const iamMinorVersion = createSelector(
   (state) => state.iamMinorVersion
 );
 
+export const atLeastV2p1 = createSelector(
```
msorens: Based on @srenatus 's idea, made this a generally available selector.

components/automate-ui/src/app/pages/api-token/details/api-token-details.component.html
```diff
@@ -46,7 +65,7 @@
         </chef-form-field>
       </form>
       <div id="save">
-        <chef-button primary [disabled]="!updateForm.valid || !updateForm.dirty || saveInProgress" (click)="saveToken()">
+        <chef-button primary [disabled]="saveInProgress || !updateForm.valid || !this.updateForm.dirty" (click)="saveToken()">
```
msorens: Nothing magical; just consistent between resources

Issue #1183, opened 2019-08-06
Automated Purge of Disconnected Services in the Applications View


danielsdeleo:
## Description

Give automate the capability to automatically delete services that have been disconnected for a given period of time. Automate administrators should be able to configure the amount of time a service must be disconnected before deleting or disable this behavior entirely. 

### Implementation Notes

* The job scheduler in the ingest-service may have some useful logic for this feature, but much of the implementation is specific to the data for that domain.
* This feature should be multi-node safe. The job-scheduler is being modified to be multi-node safe, so the changes implemented there may be of interest.

## Definition of Done

* Default setting: `disconnected services` threshold by default is 5 minutes. `remove disconnected services` threshold by default is 7 days. Both settings are enabled by default.
* TBD: determine allowable variance in when services are deleted (e.g., if threshold is set to 1 hour, how often should the deletion routine run?)
* disconnected services are deleted on or after the disconnected threshold is exceed when the feature is enabled
* nothing is deleted if the feature is not enabled.

Issue #1371, opened 2019-08-27
Update IAM 1/2 for applications lifecycle settings


danielsdeleo:
## User Story

Ingest uses policies like these to control access to configuring missing nodes settings, we need something similar for applications:

```
    option (chef.automate.api.policy).resource = "cfgmgmt:scheduler:missing-nodes";
    option (chef.automate.api.policy).action = "configure";
    option (chef.automate.api.iam.policy).resource = "retention:nodes";
    option (chef.automate.api.iam.policy).action = "retention:nodes:update";
```

PR #1385, opened 2019-08-28
Dan/periodic mark disconnected

danielsdeleo:

The applications service makes use of cereal service to run a periodic job to mark services as "disconnected." This builds upon some stopgap features we implemented in the CLI to list and delete disconnected services.

* We decided to have a periodic job to mark services as disconnected because this design is most easily extended to include notifications functionality.
* To support the in-progress effort to make Automate cluster able, we make use of the cereal service to ensure the marking job runs only once per installation.
* I attempted to make the API _similar_ to the existing API for missing nodes, by passing durations around in a string format as used by Go's `time.ParseDuration` function.
* I didn't make the recurrence of the job settable at all, except for testing.

https://github.com/chef/automate/issues/1183

build automate-gateway and applications service, then you can:

`T=$(get_admin_token)`

`curl -k -H "api-token: $T" https://a2-dev.test/apis/beta/retention/service_groups/disconnected_services/config`

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/applications-service/pkg/server/periodic.go
```diff
@@ -0,0 +1,252 @@
...
+func ConnectToJobsManager(jobCfg *config.Jobs, connFactory *secureconn.Factory) (*cereal.Manager, error) {
+	svcURL := fmt.Sprintf("%s:%d", jobCfg.Host, jobCfg.Port)
+	conn, err := connFactory.Dial(cerealServiceMutualTLSName, svcURL)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to connect to upstream job service %q at %q", cerealServiceMutualTLSName, svcURL)
+	}
+
+	be := cerealRPC.NewGrpcBackendFromConn("applications", conn)
+	//cerealSvc, err := cereal.NewManager(be)
+	cerealSvc, err := cereal.NewManager(be, cereal.WithTaskPollInterval(1*time.Second), cereal.WithTaskPollIntervalMaxJitter(100*time.Millisecond))
```
jaym: curious why we need to run this more often

components/applications-service/integration_test/disconnected_services_job_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// CONFIGURE PERIODIC DELETION OF DISCONNECTED SERVICES:
+// * this config lives with the cereal service in the job config or something like that
+// The way it's updated is:
+// (m *Manager) UpdateWorkflowScheduleByName(ctx context.Context,
+// 	instanceName string, workflowName string, opts ...WorkflowScheduleUpdateOpts)
+// need to ensure it exists on startup I guess...
```
danielsdeleo: delete this

components/applications-service/pkg/storage/postgres/services.go
```diff
@@ -100,14 +101,40 @@ SELECT s.id
...
+  , s.application AS application
+  , s.environment AS environment
+  , s.supervisor_id AS sup_member_id
+  , s.fqdn AS fqdn
+  , s.channel as channel
+  , s.site as site
+  , s.last_event_occurred_at as last_event_occurred_at
+  , s.previous_health as previous_health
+  , s.health_updated_at as health_updated_at
+	, s.disconnected as disconnected
```
danielsdeleo: TABS 😡

components/applications-service/pkg/server/periodic.go
```diff
@@ -0,0 +1,252 @@
...
+		return errors.Wrap(err, "failed to register as task exector to mark disconnected services")
+	}
+
+	wfX := patterns.NewSingleTaskWorkflowExecutor(DisconnectedServicesJobName, false)
+	err = cerealSvc.RegisterWorkflowExecutor(DisconnectedServicesJobName, wfX)
+	if err != nil {
+		return errors.Wrap(err, "failed to register as workflow exector to mark disconnected services")
+	}
+
+	// TODO: set a timeout
```
danielsdeleo: is that necessary? what's the timeout for?

components/applications-service/pkg/server/periodic.go
```diff
@@ -0,0 +1,252 @@
...
+	atomic.AddInt64(&m.totalRuns, 1)
+	if err != nil {
+		atomic.AddInt64(&m.totalRunsFailed, 1)
+	} else {
+		atomic.AddInt64(&m.totalRunsSuccessful, 1)
+	}
+	return nil, err
+}
+
+func (m *MarkDisconnectedServicesExecutor) run(t cereal.Task) error {
```
danielsdeleo: `runWithoutStats`

components/applications-service/pkg/server/periodic.go
danielsdeleo: In the tests I set the job to run once per second so I can verify that the system actually runs the task I expected it to run. In the studio on my machine this worked fine without this configuration, but in Ci the test always failed until I changed this. Since my job interval is 60s, the 1s poll interval doesn't seem too crazy.

danielsdeleo: In the tests I set the job to run once per second so I can verify that the system actually runs the task I expected it to run. In the studio on my machine this worked fine without this configuration, but in Ci the test always failed until I changed this. Since my job interval is 60s, the 1s poll interval doesn't seem too crazy.

api/external/applications/applications.proto
```diff
@@ -87,6 +87,27 @@ service ApplicationsService {
     option (chef.automate.api.iam.policy).action = "applications:serviceGroups:delete";
   };
 
+  rpc GetDisconnectedServicesConfig(GetDisconnectedServicesConfigReq) returns (GetDisconnectedServicesConfigRes) {
+    option (google.api.http).get = "/beta/retention/service_groups/disconnected_services/config";
+    // TODO: need new resources/policies for this
+    option (chef.automate.api.policy).resource = "service_groups";
```
kmacgugan: Is it service groups or is there like an config resource?

components/applications-service/pkg/server/periodic.go
kmacgugan: Is this polling the cereal service if there is a task to run? Maybe it could be configurable too and run more frequently in CI.

components/automate-deployment/pkg/server/server.go
```diff
@@ -781,7 +781,7 @@ func (s *server) Status(ctx context.Context, d *api.StatusRequest) (*api.StatusR
 	nonSkippedServices = append(nonSkippedServices, deploymentServiceName)
 	for _, svcString := range s.deployment.ServiceNames() {
 		if svc, found := s.deployment.ServiceByName(svcString); found {
-			if svc.DeploymentState != deployment.Skip {
```
danielsdeleo: Excluding `deployment.Skip` means that you include `Running`, `Removed`, and `Installed` states. It's not clear why the code was written to just exclude `Skip` here, but it seems to me that `Running` services are the only ones that should be expected to have a passing health check (?)

api/external/applications/applications.proto
danielsdeleo: This is what ingest does:
```
    option (chef.automate.api.policy).resource = "cfgmgmt:scheduler:missing-nodes";
    option (chef.automate.api.policy).action = "configure";
    option (chef.automate.api.iam.policy).resource = "retention:nodes";
    option (chef.automate.api.iam.policy).action = "retention:nodes:update";
```

I think we'd do something similar. Tracking this in #1371

components/automate-deployment/pkg/server/server.go
stevendanna: The change you have makes sense to me as well. 👍

PR #1390, opened 2019-08-29
tools/cereal-scaffold: separate enqueue and starting

stevendanna:
This separates the enqueue operation into a new command. This allows
you to more easily start multiple cereal managers and then enqueue a
single task. One downside of this is that since these operations live
in different processes, we don't benefit from the in-process
signaling.

I've also changed a bit about how the simple-workflow total time
tracking works.

Alternatively, maybe it is time to nuke this command? I still find it
useful for manually testing changes, but we now have a more complete
test suite and we should probably use that instead.


tools/cereal-scaffold/main.go
```diff
@@ -199,20 +233,211 @@ func runResetDB(_ *cobra.Command, args []string) error {
...
+		if err != nil {
+			panic(err)
+		}
+		grpcBackend := grpccereal.NewGrpcBackendFromConn("test", conn)
+		return grpcBackend
+	}
+	return postgres.NewPostgresBackend(defaultConnURIForDatabase())
+}
+
+var simpleWorkflowStartTime *time.Time
```
stevendanna: This is a bit of a hack.  It could be removed if we passed more data to our OnWorkflowComplete callback.

tools/cereal-scaffold/main.go
```diff
@@ -224,147 +449,96 @@ func (t *SimpleTask) Run(ctx context.Context, task cereal.Task) (interface{}, er
...
+		if err != nil {
+			return w.Fail(errors.Wrap(err, "could not enqueue task"))
+		}
 	}
 
-	initialVal := 0
-	return w.Continue(&initialVal)
+	return w.Continue(&SimpleWorkflowPayload{
+		Counter:   0,
+		StartTime: time.Now(),
```
stevendanna: Putting this here means we don't measure the OnStart time, but for what we typically care about when actually looking at this time, that doesn't matter.

jaym:
I like the command. it's easy to change as we need to when developing. It could be better, but it's fine for what it is

PR #1400, opened 2019-08-29
lib/cereal: log task execution duration

stevendanna:
It might be useful to see how long certain tasks took.

The duration here is the time spend in "user code".

### DIFF
--- a/lib/cereal/cereal.go
+++ b/lib/cereal/cereal.go
@@ -721,7 +721,6 @@ func (r *registeredExecutor) StartPoller(ctx context.Context, d backend.TaskDequ
 		"poll_interval": p.String(),
 	})
 	logctx.Info("Starting task poller")
-
 	timer := time.NewTimer(p.Next())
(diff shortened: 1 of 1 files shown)


lib/cereal/cereal.go
```diff
@@ -794,9 +793,15 @@ func (r *registeredExecutor) startTaskWorker(ctx context.Context, d backend.Task
...
 	result, err := r.executor.Run(ctx, &task{backendTask: t})
+	endTime := time.Now()
+	logctx := logrus.WithFields(logrus.Fields{
+		"task_name":  r.name,
+		"start_time": startTime,
+		"end_time":   endTime,
+		"duration":   endTime.Sub(startTime),
+	})
 	if err != nil {
 		err := taskCompleter.Fail(err.Error())
```
ryancragun: Should we also do a log `WithError` here so we can get an error message with all the context before we return?

PR #1401, opened 2019-08-29
ci: purge fastly path during manifest creation

stevendanna:
The Habitat API has at least 2 layers of caching for the /latest
endpoint in use in create-manifest.rb. 1 layer of caching is in the
service itself. This cache should be invalidated on write.

The second layer of cache is via Fastly.

We have created incorrect manifests on multiple occasions and we
currently believe it is the result of this caching.

In the long-run, we would like to use the expeditor workload directly;
however, a bug in expeditor currently prevents that.

This PR purges fastly for expeditor-based builds in the hopes that it
is the fastly cache that is the cause of our inconsistent manifests.

### DIFF
--- a/.expeditor/create-manifest.rb
+++ b/.expeditor/create-manifest.rb
@@ -8,6 +8,14 @@ require 'json'
 require 'openssl'
 require 'open3'
 
+module Net
+  class HTTP::Purge < HTTPRequest
+    METHOD='PURGE'
+    REQUEST_HAS_BODY = false
+    RESPONSE_HAS_BODY = true
+  end
+end
+
 BLDR_API_HOST="bldr.habitat.sh"
 BLDR_API_USER_AGENT="Chef Expeditor"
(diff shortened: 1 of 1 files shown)


chefsalim:
Fastly should not be caching any of the latest responses as we are setting cache-control to include private.  Is there a trace showing that Fastly is indeed serving the response from a cache hit?

.expeditor/create-manifest.rb
```diff
@@ -88,6 +103,15 @@ def get_latest(channel, origin, name)
   end
 end
 
+def purge_fastly_path(host, url)
```
yzl: Do we know for sure that Instant Purging is enabled for our fastly token? (I assume it is, but I've never had occasion to try so I don't know for sure.)

stevendanna:
Gonna close for now as we don't think fastly should be caching this at all.

PR #1402, opened 2019-08-30
ci: use expeditor environment to get package idents

stevendanna:
In expeditor, we have access to the exact package identifiers that
were included in the build group.  We can use these identifiers to
avoid eventual consistency issues we've seen with the Habitat Depot.

We are currently unsure on the exact cause of the inconsistent manifest
creation.  While we think it is the result of caching in the Depot API,
we have been unable to confirm this.

To facilitate hunting it down, I've also added logging for cases where the expeditor
environment doesn't match what is found in the depot:

```
2019-08-30 11:18:37 +0100 WARN: Found mismatch between sources:
2019-08-30 11:18:37 +0100 WARN:    chef/automate-gateway/0.1.0/20190829202630 from expeditor environment
2019-08-30 11:18:37 +0100 WARN:    chef/automate-gateway/0.1.0/20190829202631 from Habitat Depot
2019-08-30 11:18:37 +0100 WARN: Using chef/automate-gateway/0.1.0/20190829202630
```

This script was growing rather large, so I've broken it up a bit:

- Added a number of `PackageQuerier` classes each of which knows how to
  look up packages from a single source.  The MustExistQuerier is maybe a
  bridge too far, but oh well, I think the rest are fine.

- Added a `ManifestGenerator` class just to provide a line between "setup" logic and
  generation logic since there were a LOT of state variables in the main script.

- Added a logger so that we get timestamps in the logs

- Add some more log statements so we can see what configuration was used to generate
  a given manifest from looking at the logs.

### DIFF
--- a/.expeditor/create-manifest.rb
+++ b/.expeditor/create-manifest.rb
@@ -56,7 +56,9 @@ class PackageQuerier
     def get_latest(_channel, origin, name)
       ret = get_latest_from_disk(origin, origin, name)
       if ret == nil
-        return get_latest_from_disk(@fallback_origin, origin, name)
+        get_latest_from_disk(@fallback_origin, origin, name)
+      else
+        ret
       end
     end
 
--- a/.expeditor/create-manifest.rb
+++ b/.expeditor/create-manifest.rb
@@ -5,7 +5,7 @@
 # package from one of the following sources:
 #
 # - A set of pins maintained in this file
-# - (optionally) Environement variables set by expeditor
+# - (optionally) Environment variables set by expeditor
 # - The Habitat Depot's dev channel
 # - (optionally) Local files on disk
 #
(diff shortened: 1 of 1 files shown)


stevendanna:
It is probably better to just read the file anew than trying to parse this diff.

stevendanna:
We could also avoid Habitat completely by always starting with the most recent release manifest.

.expeditor/create-manifest.rb
```diff
@@ -1,227 +1,385 @@
...
-
-# By default, this script creates a manifest.json file that contains all the packages in the dev channel
-
+#
+# This script creates the Chef Automate release manifest. For each
+# package in products.meta, we lookup the most recent version of the
+# package from one of the following sources:
+#
+# - A set of pins maintained in this file
+# - (optionally) Environement variables set by expeditor
```
yzl: [nit] Environment

stevendanna:
```suggestion
# - (optionally) Environment variables set by expeditor
```

PR #1403, opened 2019-08-30
ci: always git-clean when we expect expeditor to open PRs

stevendanna:
Expeditor by default commits any modified files directly to
master. We currently prefer that expeditor open PRs so that we can
evaluate whether the verify tests have passed and whether we want to
make the given change.

Here, we run `git clean -fxd` to avoid anything getting unexpectedly
committed to master.

### DIFF
--- a/.expeditor/generate-automate-api-docs.sh
+++ b/.expeditor/generate-automate-api-docs.sh
@@ -19,5 +19,5 @@ git commit --message "Sync swagger files for Automate docs." --message "This pul
 open_pull_request
 
 git checkout -
-gi clean -fxd
+git clean -fxd
 git branch -D "$branch"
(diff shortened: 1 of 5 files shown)


.expeditor/generate-automate-api-docs.sh
```diff
@@ -19,4 +19,5 @@ git commit --message "Sync swagger files for Automate docs." --message "This pul
 open_pull_request
 
 git checkout -
+gi clean -fxd
```
phiggins:
```suggestion
git clean -fxd
```

phiggins: It is also important to keep you gi clean if you're practicing judo. 🥋

PR #1405, opened 2019-08-30
ingest-service: make tests pass before 4PM UTC

stevendanna:
The es-sidecar-service's purge API purges based on entire indexes. A
side-effect of this is that a request with OlderThanDays = 1, will
delete any document in the previous day's index. This includes many
documents that are less than a day old.

In the case of this test, we were injecting 4 events and expecting 2
of them to be deleted because only 2 of the RecordedAt timestamps were
less than 24 hours old. However, the two documents we expected not to
be deleted were 8 and 16 hours in the past.  In both cases, for large
portions of the day, these will be in yesterday's index instead of
today's index and thus they will get deleted.

This fixes the test by ensuring that we only expect actions that are
definitely in today's index to still exist.

These tests would have started working again at 4PM UTC.

One might argue that we should change the es-sidecar-service's API
instead of changing this test. I can be convinced.

### DIFF
--- a/components/ingest-service/integration_test/purge_test.go
+++ b/components/ingest-service/integration_test/purge_test.go
@@ -1,15 +1,15 @@
 package integration_test
 
 import (
+	"context"
 	"testing"
 	"time"
 
-	"context"
+	"github.com/stretchr/testify/require"
+	"github.com/teambition/rrule-go"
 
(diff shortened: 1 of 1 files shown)


stevendanna:
Alternatively: https://github.com/chef/automate/pull/1408

jaym:
merged https://github.com/chef/automate/pull/1408

PR #1413, opened 2019-09-02
Add cereal-service to NOPROXY

stevendanna:
The nightlies failed because the application-service had not added the
cereal-service to their NOPROXY environment variable. Other services
had also forgotten to do this, but already had the relevant IP added
to NOPROXY because of other binds.

Signed-off-by: Steven Danna <steve@chef.io>

PR #1414, opened 2019-09-02
scaffolding: Always add sys.ip to NOPROXY

stevendanna:
As can be seen in #1413 and similar previous fixes, we almost always
forget to call the addNoProxy function when adding a new bind.

At the moment, the ceremony around this is completely unnecessary
since we know that we are running all relevant components on the local
machine.  However, the current approach of each bind separately was
put in place when we thought that different services might be living
on different machines.

Adding sys.ip to NORPOXY in the scaffolding will mean that any service
using the scaffoldings can likely omit adding `addNoProxy` lines for
every service bind.

### DIFF
--- a/lib/scaffolding/shared.sh
+++ b/lib/scaffolding/shared.sh
@@ -46,6 +46,8 @@ export A2_SVC_PATH="{{pkg.svc_path}}"
 
 . ${pkg_prefix}/libexec/shell/*.sh
 
+addNoProxy "{{sys.ip}"
+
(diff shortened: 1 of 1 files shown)


stevendanna:
Note that this is not going to rebuild services by default, but services will get this change as they are otherwise modified.

stevendanna:
> Does that mean we could, as a follow-up, remove all those previously-sprinkled-on addNoProxy calls?

Yes, I think so. Doing that would imply that these services won't be on other machines in the near term, but I think that is a very safe assumption.

PR #1415, opened 2019-09-02
Add some more logging in single task executor

jaym:


components/cereal-service/pkg/server/server.go
```diff
@@ -522,6 +522,7 @@ func (s *CerealService) CreateWorkflowSchedule(ctx context.Context, req *cereal.
 		}
 		return nil, err
 	}
+	s.workflowScheduler.Trigger()
```
stevendanna: Should we also do this in cereal.go?

jaym: we should

PR #1421, opened 2019-09-03
[#1342] Rule Save button not enabled when removing a condition

### LINKED ISSUE #1342: Ingest rule save button not enabled when removing a condition
_Problem: Users have no way to remove a condition from an ingest rule unless they make other edits._

## Overview
When updating an ingest rule by removing a condition, the save rule button remains disabled and it should become enabled on any change.

## Repro Steps
- `chef-automate iam upgrade-to-v2 --beta2.1` if not already on 2.1
- navigate to _Settings_ > _Projects_
- create a project, give it 1 ingest rule, with 2 conditions, save and close
- go back to edit the ingest rule, delete one of the conditions and notice that the save button is still disabled

## Desired behaviour
- On any change at all the save button should become enabled

srenatus:

Fixes #1342.

Added a bit of side show, type-related nitpicks, etc. I'm no UI expert, I might have been missing context or nuance when doing that, please apply your scrutiny. 😉 

#1342 

### :+1: Definition of Done

Remove a condition, note that save button is enabled.

Rebuild automate-ui in some way, `start_all_services`, use the UI to:
1. create a project
2. edit its rules, create two rules
3. open the rules dialogue again, remove one of the rules
4. notice that the _Save_ button was enabled

- [X] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/pages/project/rules/project-rules.component.html
+++ b/components/automate-ui/src/app/pages/project/rules/project-rules.component.html
@@ -37,7 +37,7 @@
               Only lowercase letters, numbers, hyphens, and underscores are allowed.
             </chef-error>
             <chef-error *ngIf="conflictError">
-              Rule IDs must be unique--Rule with ID "{{ ruleForm.get('id').value }}" already exists.
+              Rule ID "{{ ruleForm.get('id').value }}" already exists.
             </chef-error>
           </chef-form-field>
           <small class="help">Rule IDs are unique, permanent, and cannot be changed later.</small>
--- a/components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
+++ b/components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
@@ -229,16 +229,14 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
   }
 
   convertToRule(): Rule {
-    // This constant ensures type safety
-    const equals_op: ConditionOperator = 'EQUALS';
     const conditions: Condition[] = this.ruleForm.controls.conditions.value.map(
       (c: {attribute: string, values: string, operator: string}) => ({
         attribute: c.attribute,
         operator: c.operator,
           // Convert values string to storage format
-        values: c.operator === equals_op
+        values: <ConditionOperator>c.operator === 'EQUALS'
           ? [c.values.trim()]
-          : c.values.split(',').map((v: string) => v.trim())
+          : c.values.split(',').map(v => v.trim())
       }));
     return {
       project_id: this.project.id,
--- a/components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
+++ b/components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
@@ -3,7 +3,7 @@ import { FormBuilder, FormGroup, FormArray, Validators } from '@angular/forms';
 import { Router, ActivatedRoute } from '@angular/router';
 import { Store } from '@ngrx/store';
 import { Subject, combineLatest } from 'rxjs';
-import { map, takeUntil, pluck, filter } from 'rxjs/operators';
+import { takeUntil, pluck, filter } from 'rxjs/operators';
 import { identity } from 'lodash/fp';
 
 import { NgrxStateAtom } from 'app/ngrx.reducers';
@@ -80,48 +80,46 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
         this.store.select(updateStatus),
         this.store.select(getProjectStatus)
       ]).pipe(
-        takeUntil(this.isDestroyed),
-        map(([gStatus, uStatus, gpStatus]) => {
-          const routeId = this.route.snapshot.paramMap.get('ruleid');
-          this.isLoading =
-            routeId
-            ? (gStatus !== EntityStatus.loadingSuccess) ||
-              (uStatus === EntityStatus.loading) ||
-              (gpStatus !== EntityStatus.loadingSuccess)
-            : false;
-        })).subscribe();
+        takeUntil(this.isDestroyed)
+      ).subscribe(([gStatus, uStatus, gpStatus]) => {
+        const routeId = this.route.snapshot.paramMap.get('ruleid');
+        this.isLoading =
+          routeId
+          ? (gStatus !== EntityStatus.loadingSuccess) ||
+            (uStatus === EntityStatus.loading) ||
+            (gpStatus !== EntityStatus.loadingSuccess)
+          : false;
+        });
 
       combineLatest([
         this.store.select(routeParams).pipe(pluck('id'), filter(identity)),
         this.store.select(routeParams).pipe(pluck('ruleid'), filter(identity))
       ]).pipe(
-        takeUntil(this.isDestroyed),
-        map(([project_id, rule_id]: string[]) => {
-          this.store.dispatch(new GetProject({ id: project_id }));
-          this.store.dispatch(new GetRule({
-            id: rule_id,
-            project_id
-          }));
-        })).subscribe();
+        takeUntil(this.isDestroyed)
+      ).subscribe(([project_id, rule_id]: string[]) => {
+        this.store.dispatch(new GetProject({ id: project_id }));
+        this.store.dispatch(new GetRule({
+          id: rule_id,
+          project_id
+        }));
+      });
 
       this.store.select(projectFromRoute).pipe(
         filter(identity),
-        takeUntil(this.isDestroyed),
-        map((state) => {
-          this.project = { ...state };
-        })
-      ).subscribe();
+        takeUntil(this.isDestroyed)
+      ).subscribe(project => {
+          this.project = project;
+      });
 
       this.store.select(ruleFromRoute).pipe(
         filter(identity),
-        takeUntil(this.isDestroyed),
-        map((state) => {
-          this.rule = { ...state };
+        takeUntil(this.isDestroyed)
+      ).subscribe(rule => {
+          this.rule = rule;
           this.editingRule = true;
-        })
-        ).subscribe();
+      });
 
-      this.store.select(getRuleAttributes).subscribe((attributes) => {
+      this.store.select(getRuleAttributes).subscribe(attributes => {
         this.attributes = attributes;
       });
   }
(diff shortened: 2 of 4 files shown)


components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -37,7 +37,7 @@
               Only lowercase letters, numbers, hyphens, and underscores are allowed.
             </chef-error>
             <chef-error *ngIf="conflictError">
-              Rule IDs must be unique--Rule with ID "{{ ruleForm.get('id').value }}" already exists.
+              Rule ID "{{ ruleForm.get('id').value }}" already exists.
```
srenatus: sideshow: unify error messages; same as done with teams and tokens.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -39,11 +39,13 @@ interface KVCondition {
...
 })
 export class ProjectRulesComponent implements OnInit, OnDestroy {
-  public project: Project = <Project>{};
   public ruleId: string;
   public ruleForm: FormGroup;
+
+  // FIXME: either make properties optional in interface, or provide them on initialization:
+  public project: Project = <Project>{};
   public rule: Rule = <Rule>{};
-  public conditions: FormGroup[];
```
srenatus: sideshow: not needed as a public variable, using a local one below.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -39,11 +39,13 @@ interface KVCondition {
   styleUrls: ['./project-rules.component.scss']
 })
 export class ProjectRulesComponent implements OnInit, OnDestroy {
-  public project: Project = <Project>{};
```
srenatus: clustered with the other offending instance variable below.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -78,48 +80,44 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
         this.store.select(updateStatus),
         this.store.select(getProjectStatus)
       ]).pipe(
-        takeUntil(this.isDestroyed),
-        map(([gStatus, uStatus, gpStatus]) => {
```
srenatus: ❓I might be missing something, but the `pipe(x, map(y)).subscribe()` logic _seems_ to be equivalent to `pipe(x).subscribe(y)`. Isn't it?

There are a few more below 🔽

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -78,48 +80,44 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
...
-          const routeId = this.route.snapshot.paramMap.get('ruleid');
-          this.isLoading =
-            routeId
```
srenatus: cleanup, if routeId is null, this is false. Still the case below, without the ternary operator.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -78,48 +80,44 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
...
+          id: rule_id,
+          project_id
+        }));
+      });
 
       this.store.select(projectFromRoute).pipe(
         filter(identity),
-        takeUntil(this.isDestroyed),
-        map((state) => {
-          this.project = { ...state };
```
srenatus: ❓ Did we **need** the copy? I didn't think so. Same with `this.rule` below -- I might be missing something, though.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -155,8 +153,6 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
     this.isDestroyed.complete();
   }
 
-  get getConditions() { return this.ruleForm.get('conditions'); }
```
srenatus: cleanup, didn't appear to be used

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -192,6 +188,7 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
   deleteCondition(index: number): void {
     const conditions = this.ruleForm.get('conditions') as FormArray;
     conditions.removeAt(index);
+    this.ruleForm.markAsDirty();
```
srenatus: 💡 this is the only non-sideshow thing here (with its test)

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -234,19 +229,17 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
...
   convertToRule(): Rule {
-    const conditions: Condition[] = [];
     // This constant ensures type safety
     const equals_op: ConditionOperator = 'EQUALS';
-    this.ruleForm.controls.conditions.value.forEach(c => {
-      conditions.push(<Condition>{
+    const conditions: Condition[] = this.ruleForm.controls.conditions.value.map(
+      (c: {attribute: string, values: string, operator: string}) => ({
         attribute: c.attribute,
         operator: c.operator,
```
srenatus: ⚠️ This still is lacking a bit of discipline: `c.operator` is a `string`, `operator` is a string literal type -- we should be checking this properly -- but I also have no idea how to make a mistake appear here at run time...

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -234,19 +229,17 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
...
           // Convert values string to storage format
         values: c.operator === equals_op
           ? [c.values.trim()]
-          : c.values.split(/,/).map((v: string) => v.trim())
-      });
-    });
+          : c.values.split(',').map((v: string) => v.trim())
```
srenatus: cleanup: no regexp needed here

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -275,20 +268,20 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
...
-                pendingCreateError.next(true);
-                pendingCreateError.complete();
-                if (error.status === HttpStatus.CONFLICT) {
-                  this.conflictErrorEvent.emit(true);
-                } else {
-                  // Close on any error other than conflict and display in banner.
-                  this.closePage();
-                }
-            });
+              const pendingCreateError = new Subject<boolean>();
```
srenatus: cleanup: indentation only

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
msorens: Type is correctly inferred, too.
```suggestion
          : c.values.split(',').map(v => v.trim())
```

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
msorens: 🤔  I'm going to think about that a bit...

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -234,19 +229,17 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
...
     // This constant ensures type safety
     const equals_op: ConditionOperator = 'EQUALS';
-    this.ruleForm.controls.conditions.value.forEach(c => {
-      conditions.push(<Condition>{
+    const conditions: Condition[] = this.ruleForm.controls.conditions.value.map(
+      (c: {attribute: string, values: string, operator: string}) => ({
         attribute: c.attribute,
         operator: c.operator,
           // Convert values string to storage format
         values: c.operator === equals_op
```
msorens: Not great, but maybe slightly better than what we have now...?
```suggestion
        values: <ConditionOperator>c.operator === 'EQUALS'
```

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -39,11 +39,13 @@ interface KVCondition {
   styleUrls: ['./project-rules.component.scss']
 })
 export class ProjectRulesComponent implements OnInit, OnDestroy {
-  public project: Project = <Project>{};
   public ruleId: string;
   public ruleForm: FormGroup;
+
+  // FIXME: either make properties optional in interface, or provide them on initialization:
```
bcmdarroch: what's the issue here?

srenatus: The interfaces defined for `Rule` and `Project` don't allow for `{}`; because `{}` is lacking non-optional properties. So, we use force to make the compiler accept that, via `<Project>{}`. That's a footgun. 😄

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
srenatus: Random google result makes me think that the reducer is where you should do the copy: https://medium.com/@tomastrajan/object-assign-vs-object-spread-in-angular-ngrx-reducers-3d62ecb4a4b0

So.... let's eventually do that and leave this as-is? :wink:

msorens: 1. No worry about reducers; they were implemented from the ground up to be immutable in our code. Reducers always need to be pure functions so the ngrx store should always be immutable.
2. My concern is are we subverting that by grabbing something from the store and then manipulating it locally? Not sure, but I think that might change the ngrx store, and that would be bad. 

Even if the above is not messing with the ngrx store, though, one should always strive for immutability. Here's an interesting read on it (though not suggesting we go this full-blown route at the moment 😁 ): https://medium.com/@ezequiel/immutability-and-builders-with-typescript-b69a51c94e8c

PR #1340, opened 2019-08-23
[Auth-1256] Added ValidateProjectAssignment and project assignment enforcement on roles

blakestier:
### :nut_and_bolt: Description: What code changed, and why?

Added `ValidateProjectAssignment` GRPC function for interservice authorization of project diff changes and updated `CreateRole` and `UpdateRole` to enforce project diff authorization.

I also made the postgres struct a singleton and fixed up some of the server bootstrapping so I didn't have to copy-pasta that antipattern one more time. Still cleanup to do there but should be a bit better.

### :athletic_shoe: How to Build and Test the Change

Set up an admin token and some projects:

```
start_all_services
rebuild components/automate-gateway
rebuild components/authz-service
chef-automate iam upgrade-to-v2 --beta2.1
export ADMINTOK=`chef-automate iam token create admin --admin`
curl -H "api-token: $ADMINTOK" -XPOST -d '{"name":"project1", "id":"project1"}' https://a2-dev.test/apis/iam/v2beta/projects --insecure
curl -H "api-token: $ADMINTOK" -XPOST -d '{"name":"project2", "id":"project2"}' https://a2-dev.test/apis/iam/v2beta/projects --insecure
```

Make a non-admin token and grant it access to a single project:

```
export PLEBTOK=`chef-automate iam token create plebtoken`
```

policy.json:

```
{
  "name": "token-access-project-1",
  "id": "token-access-project-1",
  "members": [
    "token:plebtoken"
  ],
  "statements": [
    {
      "effect": "ALLOW",
      "actions": [
        "*"
      ],
      "projects": [
        "project1"
      ]
    }
  ],
  "projects": []
}
```

```
curl -H "api-token: $ADMINTOK" -XPOST -d @policy.json https://a2-dev.test/apis/iam/v2beta/policies --insecure
```

Admin token can create roles with any project:

```
curl -H "api-token: $ADMINTOK" -XPOST -d '{"name":"test1", "id":"test1", "actions": ["iam:users:get"], "projects":["project1", "project2"]}' https://a2-dev.test/apis/iam/v2beta/roles --insecure
{"role":{"name":"test5","id":"test5","type":"CUSTOM","actions":["iam:users:get"],"projects":["project1","project2"]}}
```

Unless it doesn't exist (404):

```
[...]

PR #1431, opened 2019-09-04
[AUTH-1256] Project assignment enforcement for teams

tylercloke:

Follow up PR 2 of 3 to up to #1340, which implemented the framework and enforcement for role creation and update.

* Implemented project assignment enforcement for CreateTeam and UpdateTeam.
* Updated ValidateProjectAssignment to return nil instead of error if not on v2.1 with an explanatory comment.
* De-duped a ton of our cypress test code instead of furthering the pasta spread.
  * Removed a bunch of our delete by prefix commands in favor of a general one for IAM V2 objects.
  * Moved iamVersion and the describeIf constants to a constants.ts file and imported them where needed (deleting dups).

Setup admin token and some projects:

```
start_all_services
rebuild components/teams-service
export ADMINTOK=`chef-automate iam token create admin --admin`
curl -H "api-token: $ADMINTOK" -XPOST -d '{"name":"project1", "id":"project1"}' https://a2-dev.test/apis/iam/v2beta/projects --insecure
curl -H "api-token: $ADMINTOK" -XPOST -d '{"name":"project2", "id":"project2"}' https://a2-dev.test/apis/iam/v2beta/projects --insecure
```

Make a non-admin token and grant it access to a single project assignment and all team actions:

```
export PLEBTOK=`chef-automate iam token create plebtoken`
```

policy.json:

```
{
  "name": "token-access-project-1",
  "id": "token-access-project-1",
  "members": [
    "token:plebtoken"
  ],
  "statements": [
              {
                "effect": "ALLOW",
                "actions": [ "iam:teams:*" ],
                "projects": [ "*" ]
              },
              {
                "effect": "ALLOW",
                "actions": [ "iam:projects:assign" ],
                "projects": [ "project1" ]
              }
  ],
  "projects": []
}
```

Grant PLEBTOKEN universal access team operations and project1 assignment:

```
curl -H "api-token: $ADMINTOK" -XPOST -d @policy.json https://a2-dev.test/apis/iam/v2beta/policies --insecure
```

Admin can create teams in any project:

```
curl -H "api-token: $ADMINTOK" -XPOST -d '{"name":"team1", "id":"team1", "projects":["project1", "project2"]}' https://a2-dev.test/apis/iam/v2beta/teams --insecure
{"team":{"id":"team1","name":"team1","projects":["project1","project2"]}}
```

Nonadmin can only create teams in project1:

```
curl -H "api-token: $PLEBTOK" -XPOST -d '{"name":"team2", "id":"team2", "projects":["project1", "project2"]}' https://a2-dev.test/apis/iam/v2beta/teams --insecure
{"error":"You cannot modify a project for this object because you are missing iam:projects:assign on this project: [\"project2\"]","message":"You cannot modify a project for this object because you are missing iam:projects:assign on this project: [\"project2\"]","code":7,"details":[]}
curl -H "api-token: $PLEBTOK" -XPOST -d '{"name":"team2", "id":"team2", "projects":["project1"]}' https://a2-dev.test/apis/iam/v2beta/teams --insecure
{"team":{"id":"team2","name":"team2","projects":["project1"]}}
```

And can only add or remove project1 on team update:

```
curl -H "api-token: $PLEBTOK" -XPUT -d '{"name":"team1", "id":"team1", "projects":["project1"]}' https://a2-dev.test/apis/iam/v2beta/teams/team1 --insecure
{"error":"You cannot modify a project for this object because you are missing iam:projects:assign on this project: [\"project2\"]","message":"You cannot modify a project for this object because you are missing iam:projects:assign on this project: [\"project2\"]","code":7,"details":[]}

# have admin remove all projects

curl -H "api-token: $ADMINTOK" -XPUT -d '{"name":"team1", "id":"team1", "projects":["project1"]}' https://a2-dev.test/apis/iam/v2beta/teams/team1 --insecure

# then can add project1 and remove it

curl -H "api-token: $PLEBTOK" -XPUT -d '{"name":"team1", "id":"team1", "projects":["project1"]}' https://a2-dev.test/apis/iam/v2beta/teams/team1 --insecure
{"team":{"id":"team1","name":"team1","projects":["project1"]}}
curl -H "api-token: $ADMINTOK" -XPUT -d '{"name":"team1", "id":"team1", "projects":[]}' https://a2-dev.test/apis/iam/v2beta/teams/team1 --insecure
{"team":{"id":"team1","name":"team1","projects":[]}}[85][default:/src:0]
```

- [x] Tests added/updated?
- [ ] Docs added/updated?


components/authz-service/server/v2/authz.go
```diff
@@ -204,8 +204,11 @@ func (s *authzServer) ValidateProjectAssignment(
 	ctx context.Context,
 	req *api.ValidateProjectAssignmentReq) (*api.ValidateProjectAssignmentResp, error) {
 
+	// If we aren't on v2.1, just return success. Other services shouldn't
+	// have to know and we shouldn't return failure for v2.0. We need to reset
+	// all of their projects on upgrade to v2.1 to be safe anyway.
 	if !s.isBeta2p1() {
-		return nil, errors.New("must be on v2.1 to assign projects")
+		return &api.ValidateProjectAssignmentResp{}, nil
```
blakestier: 👍

components/automate-gateway/handler/iam/v2beta/teams/teams.go
```diff
@@ -9,12 +9,18 @@ import (
...
 	gwres "github.com/chef/automate/components/automate-gateway/api/iam/v2beta/response"
+	"github.com/chef/automate/components/automate-gateway/gateway/middleware"
 )
 
 type Server struct {
 	client teams.TeamsV2Client
 }
 
+// asserts that we satisfy the correct interface here -- it's a safeguard
+var _ middleware.AuthContextReader = (*Server)(nil)
```
blakestier: 🤔

tylercloke: That is a true 🤔 for sure.

https://github.com/chef/automate/blob/918c76098729134b39e2126e6d036e25f2ee7973/components/automate-gateway/gateway/middleware/auth_interceptor.go#L154-L166

All requests get the [projects headers atm](https://github.com/chef/automate/blob/918c76098729134b39e2126e6d036e25f2ee7973/components/automate-gateway/gateway/middleware/auth_interceptor.go#L164). However, only handlers that implements the empty interface `AuthContextReader` get all the [other auth headers](https://github.com/chef/automate/blob/918c76098729134b39e2126e6d036e25f2ee7973/components/automate-gateway/gateway/middleware/auth_interceptor.go#L162) (including the subjects header which we now need for project diff calculation).

My TODO there says we should consider just dropping this and alway send the auth headers with every request since we are going to mostly need them in every service eventually, but that has performance impact so I wanted to discuss it more widely before making that change.

So for now, you'll need to add that line to the tokens handler so that the subjects header shows up.

tylercloke: Should probably replace most of the places I say headers with context. Still in http think.

blakestier: That totally helps clarify, thanks!

e2e/cypress/index.d.ts
```diff
@@ -8,10 +8,7 @@ declare namespace Cypress {
     restoreStorage(): void
     applyProjectsFilter(projectsToFilterOn: string[]): void
     generateAdminToken(idToken: string): void
-    cleanupPoliciesByIDPrefix(idToken: string, idPrefix: string): void
-    cleanupProjectsByIDPrefix(idToken: string, idPrefix: string): void
-    cleanupRolesByIDPrefix(idToken: string, idPrefix: string): void
-    cleanupTokensByIDPrefix(idToken: string, idPrefix: string): void
+    cleanupV2IAMObjectsByIDPrefixes(idToken: string, idPrefix: string, objectPlurals: string[]): void
```
tylercloke: Didn't wanna add n+1 cleanup helper so I squashed them down for all IAM V2 objects 😅

components/authz-service/server/v2/authz.go
blakestier: Or is this where we put a message like "um don't assign projects on this version pls"? i guess we can just wait until they're all meaningful . . . just wondering if it will be hard to track down these places 🤔

components/teams-service/storage/postgres/postgres.go
```diff
@@ -179,29 +193,65 @@ func (p *postgres) EditTeam(ctx context.Context, team storage.Team) (storage.Tea
...
+		return storage.Team{}, p.processError(err)
 	}
 
-	err = p.db.QueryRowContext(ctx,
+	projectDiff := projectassignment.CalculateProjectDiff(oldProjects, newProjects)
+	if len(projectDiff) != 0 {
+		// will only return an error if authz is in v2.1 mode
+		_, err := p.authzClient.ValidateProjectAssignment(ctx, &authz_v2.ValidateProjectAssignmentReq{
+			Subjects:   auth_context.FromContext(auth_context.FromIncomingMetadata(ctx)).Subjects,
+			ProjectIds: projectDiff,
```
blakestier: Since we are passing the `projectDiff`, which includes deleted projects, we are potentially returning a `not found` in the event that that project is already deleted. Do we care?

components/teams-service/storage/postgres/postgres.go
```diff
@@ -179,29 +193,65 @@ func (p *postgres) EditTeam(ctx context.Context, team storage.Team) (storage.Tea
 
 // EditTeamByName edits the team's description and projects by name
 func (p *postgres) EditTeamByName(ctx context.Context,
-	teamName string, teamDescription string, teamProjects []string) (storage.Team, error) {
+	teamName string, teamDescription string, newProjects []string) (storage.Team, error) {
```
blakestier: this is not really `newProjects`, necessarily, since we don't have patchy updates. Maybe `updatedProjects`? or just `projects` (since we don't qualify the others such as `teamName`)

e2e/cypress/integration/api/iam/04_teams_api.spec.ts
```diff
@@ -0,0 +1,329 @@
...
+            failOnStatusCode: false,
+            body: teamWithProjects(teamID,
+              [project1.id, project2.id, 'notfound'])
+        }).then((response) => {
+            expect(response.status).to.equal(404);
+        });
+      });
+
+      it('non-admin with project1 assignment access cannot create ' +
+      'a new team with other projects (403)', () => {
```
blakestier: `other` projects or `unauthorized` projects? From the test name I can't tell if it's bc the `nonAdmin` doesn't have perms to create teams or doesn't have perms _on those projects_

e2e/cypress/integration/api/iam/04_teams_api.spec.ts
```diff
@@ -0,0 +1,329 @@
...
+        cy.request({ ...defaultNonAdminReq,
+            method: 'PUT',
+            url: `/apis/iam/v2beta/teams/${teamID}`,
+            body: teamWithProjects(teamID, [project1.id])
+        }).then((response) => {
+            expect(response.body.team.projects).to.have.length(1);
+        });
+      });
+
+      it('non-admin with project1 assignment access can update ' +
```
blakestier: Is it possible to pull these tests into two context blocks, one with the non-admin and one with the admin user? All the tests are duped so if we could just loop through those scenarios, that would be 💯

e2e/cypress/integration/api/iam/02_policies_api.spec.ts
```diff
@@ -23,15 +17,13 @@ describeIfIAMV2('policies API', () => {
...
-  const describeIfIAMV2p1 = Cypress.env('IAM_VERSION') === 'v2.1' ? describe : describe.skip;
 
   before(() => {
     cy.adminLogin('/').then(() => {
       const admin = JSON.parse(<string>localStorage.getItem('chef-automate-user'));
       adminToken = admin.id_token;
       defaultAdminReq.auth.bearer = adminToken;
-      cy.cleanupPoliciesByIDPrefix(adminToken, cypressPrefix);
-      cy.cleanupProjectsByIDPrefix(adminToken, cypressPrefix);
+      cy.cleanupV2IAMObjectsByIDPrefixes(adminToken, cypressPrefix, ['policies', 'projects']);
```
blakestier: so hard to read, but so wonderfully specific 🙃

components/teams-service/storage/postgres/postgres.go
tylercloke: That's one of the reasons we have our new cereal ticket! The case where a team contains a deleted project shouldn't happen if we are doing things properly. Even with cereal though, there is still a race condition between the project getting tombstoned and it getting deleted from the teams domain. In that case, I'd be fine with just aborting the request here. It should be a very small race condition. Would need to think harder about what the right thing to do for that edge case is.

e2e/cypress/integration/api/iam/04_teams_api.spec.ts
tylercloke: The non-admin token only has `iam:project:assign` on `project1`. I guess this whole thing could be updated to read:

```
requestor with iam:project:assign on project1 only cannot create a team with other project
```

blakestier: ohhhh i see. `requestor with project1 assignment permissions but not project2 permissions cannot create a team with both projects` is what this is saying. Lol it's awkward in all the ways. `requestor with assignment permission on project A but not B cannot create a team with both` . . . `requestor cannot create a team containing one project authorized for assignment and other project not authorized for assignment` 😂

components/teams-service/storage/postgres/postgres.go
blakestier: we originally discussed not checking perms for deletion (but I'm personally on board to do it)

.bldr.toml
```diff
@@ -1552,6 +1552,7 @@ paths = [
   "api/external/common/*",
   "api/interservice/authz/*",
   "api/interservice/teams/*",
+  "components/authz-service/*",
```
srenatus: This is a bit unfortunate. I wonder if it's either code we could put into `lib/` or proto files we could put into `api/`. 🤔

components/automate-gateway/handler/iam/v2beta/teams/teams.go
srenatus: > My TODO there says we should consider just dropping this and alway send the auth headers with every request since we are going to mostly need them in every service eventually

If we need them, we need them. The performance impact can't hinder us -- but it may necessitate optimizations, or different approaches. For now, I'd lean towards either dropping this, or, if need be, making it an **opt-out** instead.

components/teams-service/server/v2/server.go
```diff
@@ -62,6 +62,10 @@ func (s *Server) CreateTeam(ctx context.Context,
 	var team storage.Team
 	var err error
 	if team, err = s.service.Storage.StoreTeamWithProjects(ctx, req.Id, req.Name, req.Projects); err != nil {
+		// if the error is already a GRPC status code, return that directly.
+		if _, ok := status.FromError(err); ok {
+			return nil, err
+		}
```
srenatus: It's ok, but this does mean that a response from authz-service will look like it came from teams-service. I wonder if this could be confusing... 🤔

components/teams-service/storage/postgres/postgres.go
```diff
@@ -179,29 +193,65 @@ func (p *postgres) EditTeam(ctx context.Context, team storage.Team) (storage.Tea
...
 	if err != nil {
 		return storage.Team{}, p.processError(err)
 	}
 
-	var t storage.Team
-	if teamProjects == nil {
-		teamProjects = []string{}
+	var oldProjects []string
+	err = tx.QueryRowContext(ctx,
+		`SELECT t.projects FROM teams t
```
srenatus: [nit] There's only one table involved, so we could drop the `t` from `teams t` and  `t.` from the columns.

e2e/cypress/integration/api/iam/04_teams_api.spec.ts
```diff
@@ -0,0 +1,324 @@
...
+              expect(response.status).to.equal(403);
+          });
+        });
+      });
+    });
+  });
+});
+
+function teamWithProjects(id: string, projects: string[]): any {
+  return {
```
srenatus: [nit] this could be
```typescript
return {
  id,
  projects,
  name: 'fun name'
};
```

e2e/cypress/integration/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const iamVersion = (<string><Object>Cypress.env('IAM_VERSION') ?
```
srenatus: 😕 I don't understand this `<string><Object>` cast. What's behind this?

e2e/cypress/integration/constants.ts
```diff
@@ -0,0 +1,5 @@
...
+    <string><Object>Cypress.env('IAM_VERSION') :
+    'v2.0');
```
srenatus: [nit] isn't this ternary the same as `<string><Object>Cypress.env('IAM_VERSION') || 'v2.0'`?

e2e/cypress/integration/api/iam/04_teams_api.spec.ts
```diff
@@ -0,0 +1,324 @@
...
+    beforeEach(() => {
+      cy.cleanupV2IAMObjectsByIDPrefixes(adminToken, cypressPrefix, ['teams']);
+    });
+
+    afterEach(() => {
+      cy.cleanupV2IAMObjectsByIDPrefixes(adminToken, cypressPrefix, ['teams']);
+    });
+
+    describe('POST /apis/iam/v2beta/teams', () => {
+      context('both superadmin and project1 admin can', () => {
```
msorens: For all the test names where you mention "project1", keep in mind that someone just looking at the test results will only see test names where "project1" will not be meaningful, so would be slightly better to just talk about "project" generically.

.bldr.toml
tylercloke: I think this is getting picked up because we are using projectdiff actually. I'll move that into lib.

e2e/cypress/support/commands.ts
```diff
@@ -139,105 +139,48 @@ Cypress.Commands.add('generateAdminToken', (idToken: string) => {
...
-          method: 'DELETE',
-          url: `/apis/iam/v2beta/policies/${policy.id}`,
-          failOnStatusCode: false
-        });
-      }
-    }
-  });
-});
+Cypress.Commands.add('cleanupV2IAMObjectsByIDPrefixes',
+  (idToken: string, idPrefix: string, objectPlurals: string[]) => {
```
msorens: Consider calling it `iamObjects` rather than `objectPlurals`; the latter does not really provide any info to the reader.

e2e/cypress/integration/constants.ts
tylercloke: I'm not sure tbh. I was just factoring it out of the common places we have it.

e2e/cypress/integration/constants.ts
tylercloke: Ooo if so I was not aware you could use ternaries like that. Will try!

components/teams-service/storage/postgres/postgres.go
tylercloke: We need to check permissions for removing projects but maybe shouldn't fail if they don't exist.

components/teams-service/server/v2/server.go
tylercloke: I think we didn't want them to think about the proper response code, but I'm starting to think we should just return the two specific errors like we do internally in authz-service and let them return what they want per domain. We can think more on this.

e2e/cypress/integration/api/iam/04_teams_api.spec.ts
tylercloke: I thought about that but it got super wordy in practice. We've already done a lot of iteration here so let's circle back if we think things are still unclear.

e2e/cypress/support/commands.ts
tylercloke: Disagree that mentioning you want the plural version doesn't provide any value. Don't have a strong opinion though so can change.

e2e/cypress/index.d.ts
bcmdarroch: nice refactor! someday it'd be cool if we got all of the test objects created with the same prefix. then we could have something even more generic like `deleteCypressTestObjects` (and pass whatever paths)

components/teams-service/server/v2/server.go
blakestier: I like the idea of keeping "authz" decisions in authz as much as possible . . .will help with consistency. But whether that's confusing to the user is a different q

e2e/cypress/integration/api/iam/04_teams_api.spec.ts
```diff
@@ -0,0 +1,318 @@
...
+    beforeEach(() => {
+      cy.cleanupV2IAMObjectsByIDPrefixes(adminToken, cypressPrefix, ['teams']);
+    });
+
+    afterEach(() => {
+      cy.cleanupV2IAMObjectsByIDPrefixes(adminToken, cypressPrefix, ['teams']);
+    });
+
+    describe('POST /apis/iam/v2beta/teams', () => {
+      context('both superadmin and project1 admin', () => {
```
bcmdarroch: Cypress doesn't seem to differentiate between `describe` and `context`, so could we just use `describe`? 

I'd like if we had less layering in general in our test structure, since the test names tend to get long and hard to parse. Or if the layering seems necessary, let's pay attention to what the test name comes out to and change the content of each layer so it reads clearly. 

for example, here's what we have now:
```
POST /apis/iam/v2beta/teams
     both superadmin and project1 admin
         project-1-admin can create a new team with no projects
         superadmin can create a new team with no projects
```

I would prefer to drop the `both superadmin and project1 admin` context because it doesn't add anything to what's below. Or move the scenario loop inside each it block, resulting in output like this:

```
POST /apis/iam/v2beta/teams
     both superadmin and project1 admin
           can create a new team with no projects
           can create a team with project1
           cannot create a team with a project that doesn't exist
```

blakestier: I like the context myself in general (as a distinction from describe), but I'm def on board with the tests reading more like ^^.

PR #1443, opened 2019-09-04
Dan/periodic remove disconnected

danielsdeleo:

Adds periodic removal of disconnected services. The feature is enabled by default with a threshold of 7 days disconnected before removal.

#1183

Build applications-service and automate-gateway, then start all services in the studio.

Along with the mark disconnected services feature from a prior PR, the config can be accessed and modified like so:

```
export T=$(get_admin_token)
### mark services disconnected config ###
curl -k -H "api-token: $T" https://a2-dev.test/apis/beta/retention/service_groups/disconnected_services/config}'
# => {"running":true,"threshold":"5m"}
curl -k -H "api-token: $T" -X POST https://a2-dev.test/apis/beta/retention/service_groups/disconnected_services/config -d '{"running":true,"threshold":"4m"}'
curl -k -H "api-token: $T" https://a2-dev.test/apis/beta/retention/service_groups/disconnected_services/config}'
# => {"running":true,"threshold":"4m"}
### delete disconnected services config ###
curl -k -H "api-token: $T" https://a2-dev.test/apis/beta/retention/service_groups/delete_disconnected_services/config
# => {"running":true,"threshold":"7d"}
curl -k -H "api-token: $T" -X POST https://a2-dev.test/apis/beta/retention/service_groups/delete_disconnected_services/config -d '{"running":true,"threshold":"6d"}'
dev.test/apis/beta/retention/service_groups/delete_disconnected_services/config
# => {"running":true,"threshold":"6d"}
```

The flags for disconnected services are not yet exposed in the UI, but you can set the threshold for deleting disconnected services to something short like `"2m"`, then run `applications_populate_database`; your services should show up in the UI and then be removed 2-3 minutes later.

- [x] Tests added/updated?


components/applications-service/integration_test/disconnected_services_job_test.go
```diff
@@ -109,8 +112,96 @@ func TestPeriodicDisconnectedServices(t *testing.T) {
...
+
+		// Patch event timestamp to mock an old service message which we'll delete for being disconnected
+		event.EventMetadata.OccurredAt, err = ptypes.TimestampProto(time.Now().Add(-time.Hour * 24 * 8))
+		require.NoError(t, err)
+		suite.IngestService(event)
+
+		logrus.Info("Starting check for job runner")
+		runsThusFar := runners.DeleteDisconnectedServicesExecutor.TotalRuns()
+		detectedJobRun := false
+		for i := 0; i <= 100; i++ {
```
ryancragun: Interesting way to solve the eventual consistency issue. I like it, though perhaps we should be looking for `TotalRuns() == runsThusFar +1`

components/applications-service/pkg/server/periodic.go
```diff
@@ -61,28 +67,81 @@ func NewJobScheduler(cerealSvc *cereal.Manager) *JobScheduler {
 func (j *JobScheduler) Setup() error {
 	r, err := rrule.NewRRule(rrule.ROption{
 		Freq:     rrule.SECONDLY,
-		Interval: DisconnectedServicesJobIntervalSeconds,
+		Interval: DefaultJobIntervalSeconds,
+		Dtstart:  time.Now(),
```
ryancragun: [nit] might want to add `.UTC()` to that

components/applications-service/pkg/server/periodic.go
```diff
@@ -61,28 +67,81 @@ func NewJobScheduler(cerealSvc *cereal.Manager) *JobScheduler {
...
+	if err != nil {
+		return err
+	}
+
+	// FIXME: this is probably ok for a default but then we need to figure out
+	// how to update this when the threshold is set lower.
+	r, err = rrule.NewRRule(rrule.ROption{
+		Freq:     rrule.SECONDLY,
+		Interval: DefaultJobIntervalSeconds,
 		Dtstart:  time.Now(),
```
ryancragun: Same `UTC` nit here

api/external/applications/applications.proto
```diff
@@ -181,8 +196,12 @@ message UpdateDisconnectedServicesConfigReq {
   string threshold = 2;
 }
 
+message GetDisconnectedServicesConfigReq{}
 message UpdateDisconnectedServicesConfigRes {}
 
+message GetDeleteDisconnectedServicesConfigReq{}
```
blakestier: I know I'm missing context but this message name is ambiguous-- am I getting deleted services? am I deleting disconnected services?

components/applications-service/pkg/server/periodic.go
```diff
@@ -249,3 +392,50 @@ func (m *MarkDisconnectedServicesExecutor) runWithoutStats(t cereal.Task) error
...
+	}
+	logCtx = logCtx.WithFields(log.Fields{"threshold_duration": threshold})
+
+	req := applications.DisconnectedServicesReq{
+		ThresholdSeconds: int32(threshold.Seconds()),
+	}
+
+	svcRemovedRes, err := d.ApplicationsServer.DeleteDisconnectedServices(ctx, &req)
+	if err != nil {
+		logCtx.Error("periodic task failed")
```
ryancragun: might want a more expressive error message here, maybe `WithError` as well.

components/applications-service/pkg/server/periodic.go
```diff
@@ -249,3 +392,50 @@ func (m *MarkDisconnectedServicesExecutor) runWithoutStats(t cereal.Task) error
...
+	atomic.AddInt64(&d.totalRuns, 1)
+	if err != nil {
+		atomic.AddInt64(&d.totalRunsFailed, 1)
+	} else {
+		atomic.AddInt64(&d.totalRunsSuccessful, 1)
+	}
+	return nil, err
+}
+
+func (d *DeleteDisconnectedServicesExecutor) runWithoutStats(ctx context.Context, t cereal.Task) error {
```
ryancragun: I'm not sure I understand what `withoutStats` implies here

api/external/applications/applications.proto
blakestier: OK I figured it out after a few more files-- fwiw, I think the addition of the word `Job` might help, but I know it's already an ungainly thing

components/applications-service/pkg/server/periodic.go
tylercloke: Looks like it just doesn't increment runs succeeded and failed.

danielsdeleo: The one caller this method has is `Run` that just calls this and manages the stats/counters... I did it that way so there's only one place that has to update the counter for failed runs instead of repeating that in each err-early-return clause.

api/external/applications/applications.proto
```diff
@@ -108,6 +108,27 @@ service ApplicationsService {
...
+  rpc UpdateDeleteDisconnectedServicesConfig(PeriodicJobConfig) returns (UpdateDeleteDisconnectedServicesConfigRes) {
+    option (google.api.http) = {
+      post: "/beta/retention/service_groups/delete_disconnected_services/config"
+      body: "*"
+    };
+    // TODO: need new resources/policies for this
+    option (chef.automate.api.policy).resource = "service_groups";
+    option (chef.automate.api.policy).action = "delete";
+    option (chef.automate.api.iam.policy).resource = "applications:serviceGroups";
+    option (chef.automate.api.iam.policy).action = "applications:serviceGroups:delete";
```
msorens: Assuming I am parsing correctly--this is an *update* operation--then L127 and L129 should likely say `update` instead of `delete`. I realize you just copy/pasted this from L90-L109, so my comment applies to those as well--should be changed there, too.

components/applications-service/pkg/server/periodic.go
```diff
@@ -201,6 +328,22 @@ func (j *JobRunnerSet) Start(cerealSvc *cereal.Manager) error {
...
 	}
 
+	err = cerealSvc.RegisterTaskExecutor(
+		DeleteDisconnectedServicesJobName,
+		j.DeleteDisconnectedServicesExecutor,
+		cereal.TaskExecutorOpts{},
+	)
+
+	if err != nil {
+		return errors.Wrap(err, "failed to register as task exector to delete disconnected services")
```
msorens:
```suggestion
		return errors.Wrap(err, "failed to register as task executor to delete disconnected services")
```
I saw the same typo in at least one other place...

api/external/applications/applications.proto
danielsdeleo: Thanks for reviewing! I think I'm ok to let these be horribly wrong in one additional way for now. We have #1371 at the top of the queue to fix the mess :)

PR #1463, opened 2019-09-05
Prevent introspect_projects call unless IAM V2p1 or greater

msorens:

There is a periodic `introspect_projects` call to populate the global projects filter--irrespective of the IAM version. This should only be happening for IAM V2p1. This PR adds a guard function based on IAM version to accomplish that.

NA

### :+1: Definition of Done
No `introspect_projects` calls on V1 or V2.

Rebuild automate-ui.
Login.
Try each IAM version successively:
```
# chef-automate iam reset-to-v1
# chef-automate iam upgrade-to-v2
# chef-automate iam upgrade-to-v2 --beta2.1
```

Open dev tools >> network in your browser and notice that the introspect project calls only occur when on v2p1.

image

### DIFF
--- a/components/automate-ui/src/app/services/projects-filter/projects-filter.effects.ts
+++ b/components/automate-ui/src/app/services/projects-filter/projects-filter.effects.ts
@@ -1,9 +1,12 @@
 import { Injectable } from '@angular/core';
 import { HttpErrorResponse } from '@angular/common/http';
+import { Store } from '@ngrx/store';
 import { Actions, Effect, ofType } from '@ngrx/effects';
+import { NgrxStateAtom } from 'app/ngrx.reducers';
 import { interval as observableInterval, of as observableOf, Observable } from 'rxjs';
-import { catchError, mergeMap, map, tap } from 'rxjs/operators';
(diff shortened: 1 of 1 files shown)


components/automate-ui/src/app/services/projects-filter/projects-filter.effects.ts
```diff
@@ -21,15 +24,18 @@ import {
...
   ) { }
 
   private POLLING_INTERVAL_IN_SECONDS = 120; // 2 minutes
 
   @Effect()
-  latestOptions$ = observableInterval(1000 * this.POLLING_INTERVAL_IN_SECONDS)
-    .pipe(mergeMap(this.loadOptionsAction$()));
+  latestOptions$ = observableInterval(1000 * this.POLLING_INTERVAL_IN_SECONDS).pipe(
+    withLatestFrom(this.store.select(atLeastV2p1)),
+    filter(([_, projectsEnabled]) => projectsEnabled),
```
msorens: Note that I deliberately named the selector itself with an explicit version (`atLeastV2p1`), because it does convey some useful info to the reader, and named its value as `projectsEnabled`, showing what it is used for without having to add any comments.

PR #1465, opened 2019-09-05
nightlies: Use EC2-based apt mirror, add logging

stevendanna:
The nightlies failed because we hit the 30 minute timeout set in our
buildkite configuration. The logs showed that various apt-get
operations were suddenly taking much longer to download packages:

```
Fetched 65.1 MB in 6min 54s (157 kB/s)
Fetched 26.3 MB in 9min 48s (44.7 kB/s)
```

compared to previous runs:

```
Fetched 65.1 MB in 1min 44s (620 kB/s)
Fetched 26.3 MB in 23s (1108 kB/s)
```

I didn't do an investigating into the cause of the slow-down, but it
seems prudent to use a repository mirror closer to our CI system.

This PR changes our apt-sources list in the docker build step of these
tests to use the us-west-2 repository mirrors.

Further, it adds logging to make it easier to spot which section of
the job is taking a long time.

We should consider uploading these images to dockerhub so we don't
need to build them on every run.

### DIFF
--- a/components/automate-deployment/a1-migration/Dockerfile
+++ b/components/automate-deployment/a1-migration/Dockerfile
@@ -3,6 +3,8 @@ MAINTAINER Chef Software <ryan@chef.io>
 
 ENV DEBIAN_FRONTEND noninteractive
 
+RUN sed -i 's/archive.ubuntu.com/us-west-2.ec2.archive.ubuntu.com/' /etc/apt/sources.list
+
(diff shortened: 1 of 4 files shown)


stevendanna:
Running here: https://buildkite.com/chef/chef-automate-master-nightly/builds/172#b3c2f5c8-a3af-4f0f-b0a5-0b2878c4fce4

stevendanna:
This passed here: https://buildkite.com/chef/chef-automate-master-nightly/builds/173#_

PR #1474, opened 2019-09-05
Add script that can correctly setup automate cluster

jaym:
```
HOST_PWD=`pwd` TIMEOUT=120 HAB_ORIGIN=jaym ./integration/run_test ./integration/tests/cluster.sh
```


integration/helpers/loadbalancer/main.go
```diff
@@ -0,0 +1,110 @@
+package main
```
stevendanna: Maybe we can remove: https://github.com/chef/automate/blob/master/tools/proxy/main.go

PR #1475, opened 2019-09-05
[event-services] normalize error handling and logging

ryancragun:

While doing verification for HA compatibility I read through all the event-* services to ensure that they're compatible. While doing that read through I made some minor cleanups, mostly regarding error wrapping and logging, with a few minor changes to fix broken RPC's.


components/event-feed-service/pkg/config/service.go
```diff
@@ -38,9 +38,9 @@ type Cereal struct {
 	Address string `mapstructure:"address"`
 }
 
-// Jobs contains the default settings for the jobs scheduler. In many cases
-// these are only applied when the job workflows are created, after which they
-// are managed through gRPC requests and not config.
+// Jobs contains the default settings for the jobs scheduler. This is only applied
+// when the job workflows are created, after which they are managed through the
+// Purge gRPC Configure request, not the config.
```
yzl: side question for later: would it make sense to have a chef-automate command for Purge?

components/event-feed-service/pkg/persistence/feed-elastic.go
```diff
@@ -167,43 +161,39 @@ func (efs ElasticFeedStore) GetFeed(query *util.FeedQuery) ([]*util.FeedEntry, i
...
 
-	entries := make([]*util.FeedEntry, 0, len(searchResult.Hits.Hits))
+	entries := make([]*feed.FeedEntry, 0, len(searchResult.Hits.Hits))
 	for _, hit := range searchResult.Hits.Hits {
-		entry := new(util.FeedEntry)
+		entry := new(feed.FeedEntry)
 		if err := json.Unmarshal(*hit.Source, entry); err != nil {
-			logrus.WithFields(logrus.Fields{"error": err, "object": hit.Source}).Error("Error unmarshalling the feed entry object(s)")
-			return nil, 0, err
+			return nil, 0, errors.Wrapf(err, "failed to unmarshall feed entry for object %s", hit.Source)
```
yzl: unmarshal

components/event-feed-service/pkg/server/server.go
```diff
@@ -164,51 +173,41 @@ func FromInternalFormatToList(entries []*util.FeedEntry) ([]*event_feed.FeedEntr
...
 	// Validate Interval
 	if req.Interval <= 0 || req.Interval > 24 {
-		return errors.GrpcError(codes.InvalidArgument,
+		return e.GrpcError(codes.InvalidArgument,
 			"Time interval must be greater than 0 and less than or equal to 24 hours")
 	}
 
 	if 24%req.Interval != 0 {
-		return errors.GrpcError(codes.InvalidArgument, "24 must be divisible by time interval")
+		return e.GrpcError(codes.InvalidArgument, "24 must be divisible by time interval")
```
yzl: I think "time interval must divide 24" is a little easier to read, but anyone who gets this error is going to have to stare at it a bit anyway, so either is fine.

ryancragun: :laughing: I didn't write :hocho:

components/event-feed-service/cmd/event-feed-service/commands/serve.go
```diff
@@ -1,6 +1,7 @@
 package commands
 
 import (
```
srenatus: wow. we're stdlib-free here.

components/event-feed-service/pkg/persistence/feed-elastic.go
```diff
@@ -136,14 +131,13 @@ func (efs ElasticFeedStore) CreateFeedEntry(entry *util.FeedEntry) (bool, error)
...
 // GetFeed - get event feed entries with the provided query constraints
-func (efs ElasticFeedStore) GetFeed(query *util.FeedQuery) ([]*util.FeedEntry, int64, error) {
+func (efs ElasticFeedStore) GetFeed(query *feed.FeedQuery) ([]*feed.FeedEntry, int64, error) {
 	exists, err := efs.indexExists(IndexNameFeeds)
 	if err != nil {
-		logrus.Warnf("Could not get feed; error: %+v", err)
 		return nil, 0, err
 	}
 	if !exists {
-		panic(fmt.Sprintf("index %q is missing in Elasticsearch", IndexNameFeeds))
```
srenatus: yay less panic

components/event-feed-service/pkg/server/server.go
```diff
@@ -134,20 +143,20 @@ func (eventFeedServer *EventFeedServer) GetFeedTimeline(ctx context.Context,
 
 func (eventFeedServer *EventFeedServer) HandleEvent(ctx context.Context,
 	request *automate_event.EventMsg) (*automate_event.EventResponse, error) {
-	log.WithFields(log.Fields{
+	logctx := log.WithFields(log.Fields{
 		"request": request.String(),
-		"func":    "HandleEvent",
-	}).Debug("rpc call")
```
srenatus: While I agree with this removal, I vaguely remember superteam being appreciative of these log lines.

components/event-feed-service/pkg/server/server.go
```diff
@@ -164,51 +173,41 @@ func FromInternalFormatToList(entries []*util.FeedEntry) ([]*event_feed.FeedEntr
 func validateFeedTimelineRequest(req *event_feed.FeedTimelineRequest) error {
 	// Validate Timezone
 	if req.Timezone == "" {
-		return errors.GrpcError(codes.InvalidArgument, "A timezone must be provided")
+		return e.GrpcError(codes.InvalidArgument, "A timezone must be provided")
```
srenatus: Whatever it stands for, this should be `GRPCError`, I think. Fine to leave as-is, though. For consistency, don't we want `s/A timezone/a timezone/`?

components/event-gateway/cmd/event-gateway/commands/health-check.go
```diff
@@ -29,15 +28,15 @@ func newHealthCheckCommand() *cobra.Command {
 
 			// Always return healthy if not enabled
 			if !cfg.Service.Enabled {
-				os.Exit(0)
```
srenatus: 👏

components/event-feed-service/pkg/server/server.go
yzl: _carefully does not run `git blame`_

components/event-feed-service/pkg/config/service.go
ryancragun: I think it's totally reasonable if we want to do that, but I sorta hope that we could add stuff via the UI to do these things.

lib/stringutils/slice.go
```diff
@@ -34,3 +34,17 @@ func IndexOf(haystack []string, needle string) (int, error) {
...
+	res := []string{}
+
+	for _, hay := range haystack {
+		if hay != needle {
+			res = append(res, hay)
+		}
+	}
+
+	return res
+}
```
phiggins: This is a special case of `SliceFilter`, so you could use that instead of adding this function.

lib/stringutils/slice.go
```diff
@@ -34,3 +34,17 @@ func IndexOf(haystack []string, needle string) (int, error) {
 	}
 	return -1, ErrNotFound
 }
+
+// SliceReject takes a slice of strings and a string to removed a returns a slice
+// with any matching reject strings removed.
```
phiggins: This reads kinda funny. Maybe something like this is better?

> SliceReject takes a slice of strings and a string to remove and returns a slice of strings with the matching strings removed.

PR #1487, opened 2019-09-06
Remove DENY walkthrough examples from IAM v2 Beta documentation

### LINKED ISSUE #1121: Remove DENY walkthrough examples from IAM v2 Beta documentation
_Problem: Users may get into trouble if they use deny in custom policies without considering the global impact._

## Overview
Since we are not expecting _DENY_ in policies to be a common use case, we would like to change up the IAMv2 walkthrough to not center on a _DENY_ example.

### [Overview > New Policy Definition](https://automate.chef.io/docs/iam-v2-overview/#new-policy-definition)

- Remove or modify the walkthrough examples that refer to DENY
- Also maybe consider adding content to this section to make it more clear the global effects of deny, work with @sdelano 

### [User's Guide](https://automate.chef.io/docs/iam-v2-guide)

- Re-write custom policy walkthrough to not use a DENY example

susanev:

#### Problem
Users may get into trouble if they use deny in custom policies without considering the global impact.

#### Overview
Since we are not expecting DENY in policies to be a common use case, we would like to change up the IAMv2 walkthrough to not center on a DENY example.

### :+1: Definition of Done
Fixes #1121 

- [ ] Tests added/updated?
- [x] Docs added/updated?

### DIFF
--- a/components/automate-chef-io/content/docs/iam-v2-guide.md
+++ b/components/automate-chef-io/content/docs/iam-v2-guide.md
@@ -114,8 +114,8 @@ Let's further assume you want user `Bob` as well as anyone on team `gamma` to be
   ],
 ```
 
-Next, you'll specify the permissions themselves, which in IAM v2 are the `statements`, declared as an array.
-Write a statement that **allows** access to the _get_, _list_, and _update_ actions for _users_ and _teams_:
+Next, you'll specify the permissions themselves--which in IAM v2 are the `statements`-- declared as an array.
+We only need a single statement in this case, though, providing access to the _get_, _list_, and _update_ actions for _users_ and _teams_:
 
 ```json
     {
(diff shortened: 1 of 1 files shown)


components/automate-chef-io/content/docs/iam-v2-guide.md
```diff
@@ -131,20 +131,6 @@ First, write a statement that **allows** access to the _get_, _list_, and _updat
     },
 ```
```
msorens: Need to revise L117-118 accordingly:
```text
Next, you'll specify the permissions themselves--which in IAM v2 are the `statements`-- declared as an array.
We will only have a single statement in this case, though, providing access to the _get_, _list_, and _update_ actions for _users_ and _teams_:
```

Issue #1184, opened 2019-08-06
Identify Disconnected Services in the Applications UI


danielsdeleo:
## Description

Enable users to identify disconnected services in the Applications UI by adding badges to the service groups table rows and individual services.

Out of scope: filtering (that's a different task).

## Definition of Done
![image.png](https://images.zenhubusercontent.com/5a721a104b5806bc2bc51768/22471dfb-354f-4e39-83af-3507b98503ee)

1. [View mockup in InVision] (https://chef.invisionapp.com/share/C3T08NGNKEQ#/380205482_step1-_Adding_Disconnected_Badge). _Please reach out to @apriofrost for access to the InVision project for access to detailed design spec._

2. When a service group has 1 or more disconnected services, it displays a badge in the `service group` column in the table, indicating the presence of disconnected services in the service group row.

    Content for the badge: `disconnected_icon` `disconnected_count`. 

    `disconnected_icon` is downloadable from the InVision project asset once you are in the project. 

    `disconnected_count` can be accessed in the [API] (https://app.zenhub.com/workspaces/ui-team-5cba23a3b14fdc05970c8f87/issues/chef/automate/1493).

3. Services shown in the right-side panel that are in a disconnected state have notifications indicating the disconnected status and information about how long the service has been in that state.

    Content for notification: "Disconnected service: Last health check received `last_event_since` ago."

    `last_event_since` can be accessed in the [API] (https://app.zenhub.com/workspaces/ui-team-5cba23a3b14fdc05970c8f87/issues/chef/automate/1493).

    When hovering on the "`last_event_since` ago" in the UI, display a tooltip with the UTC timestamp for the time the last event was received, using the `last_event_occurred_at` in the same API.

4. Sorting: currently by default, the primary sorting factor is `percentage of OK services` from the lowest to the highest, secondary `latest health check status` from "critical", "warning", "unknown" to "OK". With the new "disconnected" status and service instance counts, add a tertiary sorting order of `connectivity` from disconnected to connected, and a quaternary sorting order of `disconnected instance count` from the highest to the lowest.

PR #1493, opened 2019-09-06
Dan/expose disconnected status in svc groups

danielsdeleo:

Includes information about disconnected services in the APIs for services and service groups:
* Service groups now includes a `disconnected_count` field to show the number of services that have their disconnected flag set to true.
* Services have a boolean field for `disconnected`, a `last_event_occurred_at` field with a UTC timestamp for the time the last event was received, and `last_event_since` which gives an English "time since" description of the elapsed time since the last event (e.g., `2 minutes 57 seconds`).

#1184

Build applications-service and automate-gateway; start all services.

To make services appear as disconnected more quickly, you will likely want to:

```
export T=$(get_admin_token)
curl -k -H "api-token: $T" -X POST https://a2-dev.test/apis/beta/retention/service_groups/disconnected_services/config -d '{"running":true,"threshold":"2m"}'
```

You can run `applications_populate_database` and wait 2-3 minutes for the services to become disconnected.

The new fields described above will then be seen in the API responses at URL paths like `/apis/beta/applications/service-groups` and `/apis/beta/applications/service-groups/:sgId`

- [x] Tests added/updated?


components/applications-service/pkg/storage/postgres/service_groups.go
```diff
@@ -76,6 +76,7 @@ const (
       ,s.application as app_name
       ,s.environment as environment
       ,s.service_group_name_suffix as name_suffix
+			,COUNT(s.disconnected) FILTER (WHERE s.disconnected = 'true') AS disconnected_count
```
srenatus: Is `disconnected` a `text`? If it's bool, we should be able to do `FILTER (WHERE s.disconnected)`

components/applications-service/integration_test/services_test.go
```diff
@@ -162,6 +173,49 @@ func TestGetServicesMultiService(t *testing.T) {
...
+
+	assert.Equal(t, 6, len(res.Services))
+	for _, s := range res.Services {
+		assert.True(t, s.Disconnected, "expected the service to be marked as disconnected")
+		expectedTime, err := ptypes.TimestampProto(eventTime)
+		require.NoError(t, err)
+		assert.Equal(t, expectedTime, s.LastEventOccurredAt)
+		// Ideally we could do:
+		//   assert.Equal(t, "30 minutes", s.LastEventSince)
+		// but we _usually_ get something like "29 minutes, 59 seconds" because we
```
srenatus: Would [`WithinDuration`](https://godoc.org/github.com/stretchr/testify/assert#WithinDuration) help?

danielsdeleo: Not quite, we have code within the API that makes a string (kinda like a similar helper function in ruby on rails) with the relative time in words, I'm checking the content of that string.

components/applications-service/pkg/storage/postgres/service_groups.go
danielsdeleo: I'll give that a go, thanks!

components/applications-service/integration_test/services_test.go
srenatus: Ah now I get it. I guess you could parse the duration string in the test... But probably not worth it. ;)

PR #1495, opened 2019-09-06
Re-enable non-flaky cypress tests in verify_private

tylercloke:
# :nut_and_bolt: Description: What code changed, and why?

Added a conditional itFlaky that will only run tests if CYPRESS_RUN_FLAKY=true is passed when running the cypress tests. Added a set of cypress tasks to run in verify_private that will run the flaky tests in soft_fail mode, while the non-flaky tests are no longer soft_fail.

Will enable in deploy after we are successful in verify_private for a bit.

```
cd e2e
CYPRESS_RUN_FLAKY=true npm run cypress:open
npm run cypress:open
```

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/e2e/cypress/integration/constants.ts
+++ b/e2e/cypress/integration/constants.ts
@@ -1,5 +1,5 @@
 export const iamVersion: string = Cypress.env('IAM_VERSION') || 'v2.0';
 export const describeIfIAMV2 = iamVersion.match(/v2/) ? describe : describe.skip;
 export const describeIfIAMV2p1 = iamVersion === 'v2.1' ? describe : describe.skip;
-export const runFlaky = <boolean><Object>Cypress.env('RUN_FLAKY') || false;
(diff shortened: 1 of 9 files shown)


e2e/cypress/integration/constants.ts
```diff
@@ -1,3 +1,5 @@
 export const iamVersion = <string><Object>Cypress.env('IAM_VERSION') || 'v2.0';
 export const describeIfIAMV2 = iamVersion.match(/v2/) ? describe : describe.skip;
 export const describeIfIAMV2p1 = iamVersion === 'v2.1' ? describe : describe.skip;
+export const runFlaky = <boolean><Object>Cypress.env('RUN_FLAKY') || false;
```
srenatus: Can you explain why we need `<boolean><Object>`? 🤔

tylercloke: Looks like we don't!

Issue #1368, opened 2019-08-27
Correct project update status for other users


msorens:
## User Story

As an Automate User in a multi-user environment, I want the Project Update Status column on the projects list page to reflect accurate information.

With this pull request (https://github.com/chef/automate/pull/1309) the project update status column on the projects list page started reflecting live data. It accurately provides information to the user who instigated the project update, but it will not for other users who view the projects list page from the time the update has started until it has completed.

Let's call the user that pressed "Update Projects" user Artemis. Elsewhere, user Apollo is also looking at the projects list page. Adaptive polling will automatically kick-in for both users, so checking the apply-rules status will occur more frequently when an update in progress. That's all good.

Say we have project P1 edited by Artemis so Artemis saw "Needs update" for the project update status. Artemis then kicks off the update. The database change from staged to applied happens right away, so P1 goes from "edits pending" to "applied" right away in the Ingest Rules column. That change will also be seen by Apollo, so that is all good, too.

Because Artemis pressed "Update Projects", her Automate remembered that this project was dirty, so even though it has now changed to "applied" Automate uses the cached "edits pending" to know that "Project Update Status" should display "updating..." until, sometime later it eventually sets it to "OK". 

The problem is in the "Project Update Status" column for Apollo. His Automate instance does not know P1 was dirty and is now being updated, so it will be immediately reported as "OK" rather than the correct "updating...". 

Second problem:
For all users, if an update is cancelled or an update fails, then for each project with rules, "Project Update Status" is supposed to revert back to "needs updating". But that is true only for the projects that previously had edits pending. There is no way to distinguish projects that had edits pending and those that did not once the former change to "applied" early in the update process; both types now say "applied".

## Definition of Done

1. The "Project Update Status" column shows "updating..." while an update is in progress for all users, not just the one who initiated the update.
2. The "Project Update Status" column shows "needs updating" for all users if the update failed or was cancelled.

## Demo Script / Repro Steps

Load lots of data (e.g. `chef_load_nodes 10000` in hab studio). That helps the update progress slow down enough to be able to watch it (one hopes!). 
Edit a project for user A. 
-> Observe user A shows "edits pending" and "needs updating".

If necessary refresh the project list for user B, who should see the same two values.

Start an update by user A.

-> Observe user A shows "applied" and "updating..."

Refresh the project list for user B who should also see "applied" and "updating..." rather than "applied" and "OK".

bcmdarroch:
[...]

PR #1502, opened 2019-09-06
Surface cancelled project update on the projects page

msorens:

This surfaces the cancelled status of a project update (a result of the user selecting "Stop Updating Projects").

### :+1: Definition of Done

Once an update is cancelled, the Project Update Status column on the projects page will revert to "needs updating" rather than "OK".

Note that the modal becomes non-cancellable, and all its buttons disabled, during the time the update is being stopped (that's up to 10 seconds at present). That means you cannot use the UI for any other operations during that time.

image

```
start_all_services
```
Also rebuild automate-ui in your environment.

Create some projects with rules
Start a project update (Settings >> Projects >> Update Projects).
Quickly select "Stop Updating Projects" (which appears only once you have started an update).
Observe that the project update status has gone from "needs updating" to "updating" and then back to "needs updating".
For a project with no rules, the project update status will remain a constant "OK".
For projects with rules but no edits pending--this will not quite work until https://github.com/chef/automate/issues/1368 is addressed.

### Thorough demo script

1. Do no update
```
Edit a rule
"Update Projects" to open modal
Cancel
=> Nothing happens
```

2. Do an update
```
Edit a rule
"Update Projects" to open modal
Confirm
=> Progress shown on the main page
=> Goes from Needs Updating to Updating to OK
```

3. Think about cancelling the update but waffle
```
Edit a rule
"Update Projects" to open modal
Confirm
"Stop Updating Projects" to open modal
Neither confirm nor deny (so update is not cancelled)
=> Progress shown in the modal; modal closes when operation completes; 
=> Goes from Needs Updating to Updating to OK
```

4. Think about cancelling the update but decide not to
```
Edit a rule
"Update Projects" to open modal
Confirm
"Stop Updating Projects" to open modal
Close modal with "X" or "Cancel" (so update is not cancelled)
=> Progress shown on the main page
=> Goes from Needs Updating to Updating to OK
```

5. Cancel an update
```
Edit a rule
"Update Projects" to open modal
Confirm
"Stop Updating Projects" to open modal
Confirm
=> Progress shown in the modal; modal closes when operation completes; 
=> Goes from Needs Updating to Updating to Needs Updating
```

image


susanev:
@msorens after you stop, since stopping seems to take some time.. could we leave the modal open, disable the primary button and change the label to `<loading spinner> Stopping Project Update...` until the stopping has finished.

Or leverage the full-content loading spinner. 

Or let me know if you have other ideas.

msorens:
Yes, that would definitely improve the UX. Your first suggestion there should be straightforward to do.

lib/cereal/patterns/chain.go
```diff
@@ -178,6 +178,16 @@ func (instance *ChainWorkflowInstance) IsRunning() bool {
 	return instance.isRunning
 }
 
+func (instance *ChainWorkflowInstance) IsCanceled() bool {
+	if instance.payload != nil {
+		return instance.payload.Canceled
+	} else if instance.result != nil {
+		return instance.result.Canceled
+	} else {
```
srenatus: Can we drop the last `} else {`?

api/interservice/authz/v2/project.proto
```diff
@@ -91,6 +91,7 @@ message ApplyRulesStatusResp {
     float percentage_complete = 3;
     bool failed = 4;
     string failure_message = 5;
+    bool cancelled = 6;
```
blakestier: Hopefully this is the kind of nitpick that you'll appreciate-- but I'd prefer we *try* to be consistent with the spelling of _cancelled_ or _canceled_. Given that below we use one `l` on https://github.com/chef/automate/pull/1502/files#diff-ab846629da975cfcfdd9a22c5cbafa28R234, could we go with that?

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -262,7 +262,9 @@ export class ProjectListComponent implements OnInit, OnDestroy {
   public getProjectStatus(project: Project): string {
     const cachedStatus = this.statusCache[project.id];
     let result: string;
-    if (this.applyRulesInProgress) {
+    if (project.status === 'NO_RULES') {
+      result = 'OK';
```
blakestier: Unrelated to this work, but is it weird for this to say `OK` if there are no rules-- _all_ projects are updated during an update, even if they don't have rules, correct?

msorens: It is actually rules that are being applied during a project update so if no rules, nothing changes on the project.

blakestier: That's what it sort of seems like to the user, but from our angle, every project will be updated, bc it's _never_ the project being updated during this call, it's the nodes 🤷‍♂ . I think it's confusing.

components/authz-service/server/v2/project_update_manager.go
```diff
@@ -7,12 +7,13 @@ import (
...
+	"google.golang.org/grpc/status"
+
 	storage "github.com/chef/automate/components/authz-service/storage/v2"
 	"github.com/chef/automate/lib/cereal"
 	"github.com/chef/automate/lib/cereal/patterns"
 	"github.com/chef/automate/lib/logger"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 
 	"github.com/gofrs/uuid"
```
srenatus: Sideshow comments:

1. Since you're already cleaning up the imports blocks (thanks!), let's please move these into the "third party deps" block, too.
2. What's this `uuid` thing here? If we only want v4, we should be using our `lib/uuid4` package instead.

components/automate-ui/src/app/pages/project/confirm-apply-stop-modal/confirm-apply-stop-modal.component.ts
```diff
@@ -42,4 +47,10 @@ export class ConfirmApplyStopModalComponent implements OnChanges {
     this.progressPrefixText = `${Math.floor(percentageComplete * 100)}% complete`;
     this.progressSuffixText = `${durCountdown} until finished`;
   }
+
+  public updateButtonText(inProgress: boolean): void {
+    this.stopRulesButtonText = inProgress
+      ? 'Stopping Project Update...'
+      : 'Stop Updating Projects';
```
srenatus: Just double-checking -- this is the casing we want? Not `Stop updating projects` or anything? 😃 (Asking because _The Right Thing_ is never clear to me in this question)

susanev: Button labels are title case, or that's the standard we are trying to achieve.

srenatus: Cool. Thanks.

I suppose another way to do this, which I think I'd slighly prefer, is to have
```html
<button>{{ inProgress ? 'Stopping Project Update...' : 'Stop Updating Projects' }}</button>
```
in the template. So, looking at the template, you'd **know** what could appear in that position. With `{{ stopRulesButtonText }}` in the template, you've got to go check. @msorens what do you think? 😃

msorens: I am definitely in favor--will change.

msorens:
@susanev In one sense, a bit unexpected, yes. But I think our target audience understands the nature of asynchronous events (stuff in the pipeline still continues for a bit) so would not be confused by it. Also, I would expect that with any sizable installation, the number would not jump that dramatically.

bcmdarroch:
I agree with @susanev, if I hit stop, I would expect the progress bar to stop (even if it's technically still going for a second on the backend)

PR #1503, opened 2019-09-06
[nodemanager-service] cereal-ize the node manager

### LINKED ISSUE #1189: Cerealize nodemanager service.

yzl:
Use the cereal service to coordinate polling of nodemanagers to avoid doing unnecessary work when nodemanager-services are deployed on more than one node. 

Fixes #1189 

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/nodemanager-service/nodemanager.go
```diff
@@ -20,11 +22,15 @@ import (
...
 
+	"github.com/chef/automate/lib/cereal"
+	grpccereal "github.com/chef/automate/lib/cereal/grpc"
+	"github.com/chef/automate/lib/cereal/patterns"
 	"github.com/chef/automate/lib/grpc/health"
 	"github.com/chef/automate/lib/grpc/secureconn"
 	"github.com/chef/automate/lib/version"
 
 	"github.com/chef/automate/lib/tracing"
+	"github.com/sirupsen/logrus"
```
ryancragun: [nit] I think we're trying to have imports grouped by stdlib, external, local, so this and some below should go up with `rrule-go`

components/nodemanager-service/nodemanager.go
```diff
@@ -57,6 +63,32 @@ func Serve(conf config.Nodemanager, grpcBinding string) error {
 	return serve(ctx, &conf, connFactory)
 }
 
+const (
+	Awsec2PollingJobName        = "awsec2_polling"
```
ryancragun: [nit] these constants... look funny? I'm not sure I like the alternatives better though. Perhaps `AWSEC2PollingJobName` or `AwsEc2PollingJobName`?

components/nodemanager-service/nodemanager.go
```diff
@@ -57,6 +63,32 @@ func Serve(conf config.Nodemanager, grpcBinding string) error {
...
+	Awsec2PollingScheduleName   = "awsec2_polling_schedule"
+	AzurevmPollingJobName       = "azurevm_polling"
+	AzurevmPollingScheduleName  = "azurevm_polling_schedule"
```
ryancragun: `AzureVMPollling*`?

components/nodemanager-service/nodemanager.go
```diff
@@ -103,14 +135,165 @@ func serve(ctx context.Context, config *config.Nodemanager, connFactory *securec
...
+		err = errors.New("nodemanager setup, error grpc dialing to cereal-service aborting...")
+		return err
+	}
+	// LATER: look at whatever Dan did for picking the domain
+	cerealBackend := grpccereal.NewGrpcBackendFromConn("nodemanager", cerealConn)
+
+	// Set up a cereal (workflow) manager for managing nodemanager workflows.
+	cerealManager, err := cereal.NewManager(cerealBackend)
+	if err != nil {
+		log.Errorf("could not create cereal manager: %s", err)
```
ryancragun: I think we ought to return this error instead of log it.

components/nodemanager-service/nodemanager.go
```diff
@@ -103,14 +135,165 @@ func serve(ctx context.Context, config *config.Nodemanager, connFactory *securec
...
+	}
+
+	azurePollInterval := config.Manager.AzureVMPollIntervalMinutes
+	azureRule, err := rrule.NewRRule(rrule.ROption{
+		Freq:     rrule.MINUTELY,
+		Interval: azurePollInterval,
+		Dtstart:  time.Now(),
+	})
+	if err != nil {
+		return errors.Wrapf(err, "Could not create recurrence rule for nodemanager workflow %s", AzurevmPollingJobName)
```
ryancragun: [nit] upper case error messages

components/nodemanager-service/nodemanager.go
```diff
@@ -103,14 +135,165 @@ func serve(ctx context.Context, config *config.Nodemanager, connFactory *securec
...
+	}
+
+	// managersPollInterval := 120
+	// poll faster for debugging
+	managersPollInterval := 1
```
ryancragun: We should get the poll interval from config

components/nodemanager-service/nodemanager.go
```diff
@@ -103,14 +135,165 @@ func serve(ctx context.Context, config *config.Nodemanager, connFactory *securec
...
+			}
+		} else {
+			return errors.Wrapf(err, "Could not continue creating workflow schedule %s", Awsec2PollingScheduleName)
+		}
+	}
+
+	err = cerealManager.CreateWorkflowSchedule(AzurevmPollingScheduleName, AzurevmPollingJobName, nil, true, azureRule)
+	if err != nil {
+		if err == cereal.ErrWorkflowScheduleExists {
+			log.Infof("nodemanager workflow schedule %s already exists, not migrating", AzurevmPollingScheduleName)
```
ryancragun: I'm not sure you're migrating anything

components/nodemanager-service/nodemanager.go
```diff
@@ -103,14 +135,165 @@ func serve(ctx context.Context, config *config.Nodemanager, connFactory *securec
...
+			if err != nil {
+				return errors.Wrapf(err, "unable to get rrule for scheduled workflow %s", AzurevmPollingScheduleName)
+			}
+			if scheduledRule != rule {
+				err = cerealManager.UpdateWorkflowScheduleByName(context.Background(), AzurevmPollingScheduleName, AzurevmPollingJobName, cereal.UpdateRecurrence(rule))
+				if err != nil {
+					return errors.Wrapf(err, "unable to update recurrence rule for scheduled workflow %s", AzurevmPollingScheduleName)
+				}
+			}
+		} else {
```
ryancragun: It'd be nice if we could refactor these three block into a function and return here instead of needing more branching

components/nodemanager-service/polling/polling.go
```diff
@@ -15,84 +14,66 @@ import (
...
-
-func PollManagersStatus(ctx context.Context, db *pgdb.DB, secretsClient secrets.SecretsServiceClient) {
-	for {
-		time.Sleep(120 * time.Minute)
-		go checkManagersStatuses(ctx, db, secretsClient)
-	}
-}
-
-func checkManagersStatuses(ctx context.Context, db *pgdb.DB, secretsClient secrets.SecretsServiceClient) {
+func CheckManagersStatuses(ctx context.Context, db *pgdb.DB, secretsClient secrets.SecretsServiceClient) error {
```
ryancragun: :heart_eyes:

components/nodemanager-service/nodemanager.go
yzl: For consistency, you mean? The current non-cerealized code does not: 
https://github.com/chef/automate/blob/83e0e071be9f124f2fba5a4915e7109c5d7da5cd/components/nodemanager-service/polling/polling.go#L31

components/nodemanager-service/nodemanager.go
yzl: Yes, I need to rewrite this error message. We're still using the config to set the various polling intervals, just changing the mechanism that consume them.

components/nodemanager-service/nodemanager.go
yzl: I removed a bunch of reptition/branching just now, so hopefully this is less bad.

components/nodemanager-service/nodemanager.go
```diff
@@ -133,3 +262,33 @@ func newGRPCServer(db *pgdb.DB, connFactory *secureconn.Factory, config *config.
 
 	return s
 }
+
+func createOrUpdateWorkflowSchedule(cerealManager *cereal.Manager, scheduleName string, jobName string, rule *rrule.RRule) error {
+	err := cerealManager.CreateWorkflowSchedule(context.Background(), scheduleName, jobName, nil, true, rule)
+
+	if err == nil {
```
yzl: I have funny feelings about this early return, but I like less branching/indentation...

components/nodemanager-service/nodemanager.go
```diff
@@ -133,3 +262,33 @@ func newGRPCServer(db *pgdb.DB, connFactory *secureconn.Factory, config *config.
...
+		return nil
+	}
+
+	if err == cereal.ErrWorkflowScheduleExists {
+		log.Infof("nodemanager workflow schedule %s already exists, not creating", scheduleName)
```
yzl: Would it make sense to add a function to cereal itself to update a workflow schedule if the recurrence rule is updated? This feels like code people could write several times.

components/nodemanager-service/nodemanager.go
```diff
@@ -103,23 +134,121 @@ func serve(ctx context.Context, config *config.Nodemanager, connFactory *securec
...
+		case "azure":
+			pollInterval = config.AzureVMPollIntervalMinutes
+			jobName = AzureVMPollingJobName
+			scheduleName = AzureVMPollingScheduleName
+		case "manager":
+			// The default for the manager check is *not* set in the config. 120 was the default in the pre-cereal code.
+			pollInterval = 120
+			jobName = ManagersPollingJobName
+			scheduleName = ManagersPollingScheduleName
+		default:
```
jaym: should this case panic?

components/nodemanager-service/nodemanager.go
```diff
@@ -103,23 +134,121 @@ func serve(ctx context.Context, config *config.Nodemanager, connFactory *securec
...
+	// Prelude to initializing the cereal manager: register workflow executors to run tasks
+	for _, jobName := range []string{AwsEc2PollingJobName, AzureVMPollingJobName, ManagersPollingJobName} {
+		err = cerealManager.RegisterWorkflowExecutor(jobName, patterns.NewSingleTaskWorkflowExecutor(jobName, false))
+		if err != nil {
+			return errors.Wrapf(err, "failed to register workflow for %q", jobName)
+		}
+	}
+
+	kindsOfChecks := [3]string{"aws", "azure", "manager"}
+	var pollInterval int
```
jaym: do these need to be declared outside the for loop?

yzl: Nope -- I'll move them.

components/nodemanager-service/cmd/nodemanager-service/cmd/run.go
```diff
@@ -96,4 +100,7 @@ func init() {
 	// Event Service Flags
 	runCmd.Flags().StringVar(&conf.EventConfig.Endpoint, "event-endpoint", conf.EventConfig.Endpoint, "Event Service Endpoint")
 
+	// Cereal Service Flags
+	runCmd.Flags().StringVar(&conf.Cereal.Endpoint, "cereal-endpoint", conf.Cereal.Endpoint, "Cereal Service Endpoint")
+	runCmd.Flags().BoolVar(&conf.Cereal.Skip, "skip-cereal", conf.Cereal.Skip, "Do not deploy Cereal Service (test only)")
```
jaym: deploy is the wrong word here

PR #1505, opened 2019-09-07
Compliance API suggestion for control tags

alexpop:
This PR introduces `control_tag_key` and `control_tag_value` compliance API suggestions.

Because of the level of nesting control tags are at, the suggestions happen in two steps:

1. `engram` ElasticSearch query returns a list with all controls that match the text. We get all tags from these controls with a score we can't associate to a particular tag.

2. Tags are then aggregated to reduce duplicates and sorted by score. The `closestmatch` go lib is used to further `engram` search the aggregated tag key or values.

For `control_tag_value` suggestions a `control_tag` filter is required. This will be used to figure out the control tag key for which to do the value suggestions. Example from the tests in this PR:

```
    actual_data = GRPC reporting, :list_suggestions, Reporting::SuggestionRequest.new(
      type: 'control_tag_value',
      filters: [
        Reporting::ListFilter.new(type: 'end_time', values: ['2018-03-04T23:59:59Z']),
        Reporting::ListFilter.new(type: 'control_tag:satisfies', values: []),
      ]
    )
```

### DIFF
--- a/components/compliance-service/api/reporting/server/server.go
+++ b/components/compliance-service/api/reporting/server/server.go
@@ -139,7 +139,7 @@ func (srv *Server) ListSuggestions(ctx context.Context, in *reporting.Suggestion
 	if err != nil {
 		return nil, errorutils.FormatErrorMsg(err, "")
 	}
-	suggestionsList, err := srv.es.GetSuggestions(in.Type, formattedFilters, in.Text, in.Size)
+	suggestionsList, err := srv.es.GetSuggestions(ctx, in.Type, formattedFilters, in.Text, in.Size)
 	if err != nil {
 		return nil, errorutils.FormatErrorMsg(err, "")
 	}
--- a/components/compliance-service/api/tests/07_suggestions_control_tags_spec.rb
+++ b/components/compliance-service/api/tests/07_suggestions_control_tags_spec.rb
@@ -1,0 +1,136 @@
+##### GRPC SETUP #####
+require 'api/reporting/reporting_pb'
+require 'api/reporting/reporting_services_pb'
+
+describe File.basename(__FILE__) do
+  Reporting = Chef::Automate::Domain::Compliance::Api::Reporting unless defined?(Reporting)
+
+  def reporting
+    Reporting::ReportingService;
+  end
+
+  it "errors" do
+    assert_grpc_error("'control_tag' filter is required for 'control_tag_value' suggestions", 2) do
+      GRPC reporting, :list_suggestions, Reporting::SuggestionRequest.new(
+        type: 'control_tag_value',
+        filters: []
+      )
+    end
+  end
+
+  it "works" do
+    # suggest control tag keys
+    actual_data = GRPC reporting, :list_suggestions, Reporting::SuggestionRequest.new(
+      type: 'control_tag_key',
+      text: 'scope',
+      filters: [
+        Reporting::ListFilter.new(type: 'start_time', values: ['2018-02-01T23:59:59Z']),
+        Reporting::ListFilter.new(type: 'end_time', values: ['2018-03-04T23:59:59Z'])
+      ],
+      size: 10
+    )
+    expected = [
+      "scope",
+      "scoop",
+      "Scoops"
+    ]
+    assert_suggestions_text(expected, actual_data)
+
+    # suggest control tag keys
+    actual_data = GRPC reporting, :list_suggestions, Reporting::SuggestionRequest.new(
+      type: 'control_tag_key',
+      text: 'scope',
+      filters: [
+        Reporting::ListFilter.new(type: 'start_time', values: ['2018-02-01T23:59:59Z']),
+        Reporting::ListFilter.new(type: 'end_time', values: ['2018-03-04T23:59:59Z'])
+      ],
+      size: 2
+    )
+    expected = [
+      "scope",
+      "scoop"
+    ]
+    assert_suggestions_text(expected, actual_data)
+
+    # suggest control tag keys
+    actual_data = GRPC reporting, :list_suggestions, Reporting::SuggestionRequest.new(
+      type: 'control_tag_key',
+      text: 'Scoo',
+      filters: [
+        Reporting::ListFilter.new(type: 'start_time', values: ['2018-02-01T23:59:59Z']),
+        Reporting::ListFilter.new(type: 'end_time', values: ['2018-03-04T23:59:59Z'])
+      ],
+      size: 100
+    )
+    expected = [
+      "scoop",
+      "Scoops",
+      "scope"
+    ]
+    assert_suggestions_text(expected, actual_data)
+
+    # suggest control tag keys without text given
+    actual_data = GRPC reporting, :list_suggestions, Reporting::SuggestionRequest.new(
+      type: 'control_tag_key',
+      filters: [
+        Reporting::ListFilter.new(type: 'start_time', values: ['2018-02-01T23:59:59Z']),
+        Reporting::ListFilter.new(type: 'end_time', values: ['2018-03-04T23:59:59Z'])
+      ],
+      size: 100
+    )
+    expected = [ "cci", "gtitle", "satisfies", "scoop", "Scoops", "scope", "stig_id", "tag1", "web" ]
+    assert_suggestions_text(expected, actual_data)
+
+    # suggest control tag values with a tag key filter without text
+    actual_data = GRPC reporting, :list_suggestions, Reporting::SuggestionRequest.new(
+      type: 'control_tag_value',
+      filters: [
+        Reporting::ListFilter.new(type: 'start_time', values: ['2018-02-01T23:59:59Z']),
+        Reporting::ListFilter.new(type: 'end_time', values: ['2018-03-04T23:59:59Z']),
+        Reporting::ListFilter.new(type: 'control_tag:scope', values: []),
+      ],
+      size: 100
+    )
+    expected = [ "Apache", "apalache", "NginX" ]
+    assert_suggestions_text(expected, actual_data)
+
+    # suggest control tag values with tag key filter without text
+    actual_data = GRPC reporting, :list_suggestions, Reporting::SuggestionRequest.new(
+      type: 'control_tag_value',
+      filters: [
+        Reporting::ListFilter.new(type: 'start_time', values: ['2018-02-01T23:59:59Z']),
+        Reporting::ListFilter.new(type: 'end_time', values: ['2018-03-04T23:59:59Z']),
+        Reporting::ListFilter.new(type: 'control_tag:satisfies', values: []),
+      ],
+      size: 5
+    )
+    expected = [ "apache-1", "apache-2", "NGX-1", "NGX-2", "SRG-00006" ]
+    assert_suggestions_text(expected, actual_data)
+
+    # suggest control tag values with a tag key filter without text
+    actual_data = GRPC reporting, :list_suggestions, Reporting::SuggestionRequest.new(
+      type: 'control_tag_value',
+      text: 'apaCH',
+      filters: [
+        Reporting::ListFilter.new(type: 'start_time', values: ['2018-02-01T23:59:59Z']),
+        Reporting::ListFilter.new(type: 'end_time', values: ['2018-03-04T23:59:59Z']),
+        Reporting::ListFilter.new(type: 'control_tag:scope', values: []),
+      ]
+    )
+    expected = [ "Apache", "apalache" ]
+    assert_suggestions_text(expected, actual_data)
+
+    # suggest control tag values with a missing tag key filter
+    actual_data = GRPC reporting, :list_suggestions, Reporting::SuggestionRequest.new(
+      type: 'control_tag_value',
+      text: 'apaCH',
+      filters: [
+        Reporting::ListFilter.new(type: 'start_time', values: ['2018-02-01T23:59:59Z']),
+        Reporting::ListFilter.new(type: 'end_time', values: ['2018-03-04T23:59:59Z']),
+        Reporting::ListFilter.new(type: 'control_tag:missing', values: []),
+      ]
+    )
+    expected = []
+    assert_suggestions_text(expected, actual_data)
+  end
+end
--- a/components/compliance-service/api/tests/07_suggestions_control_tags_spec.rb
+++ b/components/compliance-service/api/tests/07_suggestions_control_tags_spec.rb
@@ -9,7 +9,7 @@ describe File.basename(__FILE__) do
     Reporting::ReportingService;
   end
 
-  it "errors" do
+  it "errors without control_tag filter" do
     assert_grpc_error("'control_tag' filter is required for 'control_tag_value' suggestions", 2) do
       GRPC reporting, :list_suggestions, Reporting::SuggestionRequest.new(
         type: 'control_tag_value',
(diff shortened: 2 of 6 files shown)


components/compliance-service/api/tests/07_suggestions_control_tags_spec.rb
```diff
@@ -0,0 +1,136 @@
...
+  it "errors" do
+    assert_grpc_error("'control_tag' filter is required for 'control_tag_value' suggestions", 2) do
+      GRPC reporting, :list_suggestions, Reporting::SuggestionRequest.new(
+        type: 'control_tag_value',
+        filters: []
+      )
+    end
+  end
+
+  it "works" do
```
phiggins: I used the `it "works"` pattern because a lot of existing tests couldn't be broken up easily as later tests depended on data created in earlier tests. For a brand new test file it'd be better for readability and debugging to break these into individual tests that could fail independently.

components/compliance-service/reporting/relaxting/suggestions.go
```diff
@@ -502,6 +502,167 @@ func (backend ES2Backend) getControlSuggestions(client *elastic.Client, typePara
...
+	type ControlTagsScore struct {
+		Score      float32
+		StringTags []ESInSpecReportControlStringTags
+	}
+
+	// Using this array to aggregate the unsorted suggestions from the response
+	foundControlTags := make([]ControlTagsScore, 0)
+	if searchResult != nil {
+		for _, hit := range searchResult.Hits.Hits {
+			if hit != nil {
```
phiggins: Instead of wrapping the body of each `for` loop in an `if`, you could probably remove several levels of nesting here by doing something like:

```
if hit == nil {
  continue
}
```

components/compliance-service/reporting/relaxting/suggestions.go
```diff
@@ -67,13 +69,15 @@ func (backend ES2Backend) GetSuggestions(typeParam string, filters map[string][]
...
-			}
+	}
+
+	controlTagFilterKey := ""
+	// Going through all filters to find the ones prefixed with 'control_tag', e.g. 'control_tag:nist'
+	for filterType := range filters {
+		if strings.HasPrefix(filterType, "control_tag:") {
+			_, controlTagFilterKey = leftSplit(filterType, ":")
+			useSummaryIndex = false
+			break
```
phiggins: Instead of embedding the value in the key name, why not use `control_tag` as the key and the string to the right of the colon as the value?

alexpop: Filters like environment, platform, node_id have one or multiple values. Contol tags are different because they have a key and one or multiple values.
We discussed this in one of the standups and decided to follow the pattern already used for the node manager and put the key in the filter type.

components/compliance-service/api/tests/07_suggestions_control_tags_spec.rb
alexpop: Thanks Pete, I wrapped an `it` block around each test in a new commit.

components/compliance-service/reporting/relaxting/suggestions.go
alexpop: I followed your advice Pete and removed 3 layers of nesting in another commit to this PR. Cheers!

components/compliance-service/reporting/relaxting/suggestions.go
```diff
@@ -502,6 +502,170 @@ func (backend ES2Backend) getControlSuggestions(client *elastic.Client, typePara
 	return suggs, nil
 }
 
+func (backend ES2Backend) getControlTagsSuggestions(client *elastic.Client, typeParam string, target string, text string, controlTagFilterKey string, size int, filters map[string][]string, useSummaryIndex bool) ([]*reportingapi.Suggestion, error) {
+	if typeParam == "control_tag_value" && controlTagFilterKey == "" {
+		return nil, errors.New("'control_tag' filter is required for 'control_tag_value' suggestions")
```
vjeffrey: should this be using the "invalid" error option? i.e. https://github.com/chef/automate/blob/master/components/compliance-service/api/reporting/server/server.go#L78 ?
also maybe this should be at the api level since it's a validation path?

components/compliance-service/reporting/relaxting/suggestions.go
phiggins: I don't understand this comment. Why is it better to put a key and value into the key and then not provide a value?

In case I am not making any sense, why is this:

```
Reporting::ListFilter.new(type: 'control_tag:satisfies', values: []),
```

better than this?

```
Reporting::ListFilter.new(type: 'control_tag', values: [satisfies]),
```

phiggins: Do we ever provide a value for `control_tag` where this wouldn't work? Do we need to have a test demonstrating why we'd do that?

vjeffrey: a control tag can have a key and no value, a key and a single string value, or a key an array of string values. so the full tag filter for a control tag with a value will look like: 
```
Reporting::ListFilter.new(type: 'control_tag:scope', values: ['Apache'])
```

vjeffrey: and there's an example of it here: https://github.com/chef/automate/blob/master/components/compliance-service/api/tests/02_nodes_spec.rb#L1414 that was added in the pr that added the control tag filtering to the code

components/compliance-service/reporting/relaxting/suggestions.go
```diff
@@ -502,6 +502,170 @@ func (backend ES2Backend) getControlSuggestions(client *elastic.Client, typePara
...
+	searchResult, err := client.Search().
+		SearchSource(searchSource).
+		Index(esIndex).
+		FilterPath(
+			"took",
+			"hits.total",
+			"hits.hits._id",
+			"hits.hits._score",
+			"hits.hits.inner_hits").
+		Do(context.Background())
```
stevendanna: 💭 Most other functions in this code use context.Background() so it might be best to defer this but --  Shouldn't we be passing the context from the request down into these functions?

components/compliance-service/reporting/relaxting/suggestions.go
```diff
@@ -502,6 +502,170 @@ func (backend ES2Backend) getControlSuggestions(client *elastic.Client, typePara
...
+			for _, inner_hit := range hit.InnerHits {
+				if inner_hit == nil {
+					continue
+				}
+				for _, hit2 := range inner_hit.Hits.Hits {
+					if hit2.Source == nil {
+						continue
+					}
+					var c ControlSourceTags
+					err := json.Unmarshal(*hit2.Source, &c)
```
stevendanna: Should we log this error in the non-nil case?

components/compliance-service/reporting/relaxting/suggestions.go
```diff
@@ -502,6 +502,170 @@ func (backend ES2Backend) getControlSuggestions(client *elastic.Client, typePara
...
+		// was the closest match from all the tag keys the control might have
+		if typeParam == "control_tag_key" {
+			for _, controlTag := range item.StringTags {
+				matches = append(matches, controlTag.Key)
+			}
+		}
+
+		// If the suggestion is for control tag values, we need to find the tag that matches the
+		// tag key and then find the closest match from all the values of that tag
+		if typeParam == "control_tag_value" {
```
stevendanna: Any value in doing this append in the pass we are doing in 612-614 above? I think the conditions are the same, so we could move 621-623 into the same loop unless I've misread.

components/compliance-service/reporting/relaxting/suggestions.go
```diff
@@ -502,6 +502,170 @@ func (backend ES2Backend) getControlSuggestions(client *elastic.Client, typePara
...
+				}
+			}
+		}
+	}
+	sort.Slice(foundControlTags, func(i, j int) bool {
+		return foundControlTags[i].Score > foundControlTags[j].Score
+	})
+
+	// Map the found tag suggestions to remove duplications
+	scoredTagSuggs := make(map[string]*float32)
```
stevendanna: ❓ I'm not clear on why we need *float32 here.  Since our pointers are gonna be 64bit anyway, we aren't really saving any space as far as I can tell and at the call site we dereference the pointer to get the value back anyway.

components/compliance-service/reporting/relaxting/suggestions.go
```diff
@@ -502,6 +502,170 @@ func (backend ES2Backend) getControlSuggestions(client *elastic.Client, typePara
...
+		bestMatch := findBestArrayMatch(text, matches)
+		for _, match := range bestMatch {
+			bestMatchScore := item.Score
+			if scoredTagSuggs[match] == nil || *scoredTagSuggs[match] < item.Score {
+				scoredTagSuggs[match] = &bestMatchScore
+			}
+		}
+	}
+
+	suggs := make([]*reportingapi.Suggestion, 0)
```
stevendanna: [nit] We can pre-allocate this

```suggestion
	suggs := make([]*reportingapi.Suggestion, 0, len(scoredTagSuggs))
```

It doesn't matter, it is just a habit I try to keep up.

components/compliance-service/reporting/relaxting/suggestions.go
```diff
@@ -502,6 +502,170 @@ func (backend ES2Backend) getControlSuggestions(client *elastic.Client, typePara
...
+		// tag key and then find the closest match from all the values of that tag
+		if typeParam == "control_tag_value" {
+			for _, controlTag := range item.StringTags {
+				if controlTag.Key == controlTagFilterKey {
+					matches = append(matches, controlTag.Values...)
+				}
+			}
+		}
+
+		// Find the best engram match from the array of texts
```
stevendanna: I think there is some subtly here that might be good to put into a comment.  Namely, `firstBestArrayMatch` either returns an array containing only the best match OR (in the case of len(text) < 2) it just returns us the original array.  

Since item.Score is invariant over the duration of 629-634 I _think_ that loop is equivalent to just looking at the first item if it exists.

My guess is if we were to look at the assembly the compiler has recognized all of this, but...I'm thinking there might be a way to restructure the code here a _bit_ to make some of this implicit behavior a bit more explicit.

components/compliance-service/reporting/relaxting/suggestions.go
alexpop: That is correct, thank you Victoria.
Pete brings a good point regarding a full (key & value) control_tag filter in the suggestions tests. I added one with commit https://github.com/chef/automate/pull/1505/commits/f28bf948aa9d10eaa706cf5b7f6f955fc7a82f30

components/compliance-service/reporting/relaxting/suggestions.go
```diff
@@ -502,6 +504,170 @@ func (backend ES2Backend) getControlSuggestions(client *elastic.Client, typePara
...
+		FilterPath(
+			"took",
+			"hits.total",
+			"hits.hits._id",
+			"hits.hits._score",
+			"hits.hits.inner_hits").
+		Do(context.Background())
+
+	if err != nil {
+		logrus.Error("getControlTagsSuggestions search failed")
```
stevendanna: [nit] Other parts of this function wrap and return rather than log errors. If we are going to log, let's use WithError to record the error text as well.

components/compliance-service/reporting/relaxting/suggestions.go
alexpop: I switched to invalid error, thank you.
We could validate the control_tag in the API, but will loop twice through the filters to extract the control tag key.

components/compliance-service/reporting/relaxting/suggestions.go
alexpop: Thank you Steven.
Just to make sure I get it right, you suggest refactoring this code:
```go
		// If the suggestion is for control tag key, we need to find the one that
		// was the closest match from all the tag keys the control might have
		if typeParam == "control_tag_key" {
			for _, controlTag := range item.StringTags {
				matches = append(matches, controlTag.Key)
			}
		}

		// If the suggestion is for control tag values, we need to find the tag that matches the
		// tag key and then find the closest match from all the values of that tag
		if typeParam == "control_tag_value" {
			for _, controlTag := range item.StringTags {
				if controlTag.Key == controlTagFilterKey {
					matches = append(matches, controlTag.Values...)
				}
			}
		}
```

to this?

```go
		for _, controlTag := range item.StringTags {
			// If the suggestion is for control tag key, we need to find the one that
			// was the closest match from all the tag keys the control might have
			// If the suggestion is for control tag values, we need to find the tag that matches the
			// tag key and then find the closest match from all the values of that tag
			if typeParam == "control_tag_key" {
				matches = append(matches, controlTag.Key)
			} else if typeParam == "control_tag_value" && controlTag.Key == controlTagFilterKey {
				matches = append(matches, controlTag.Values...)
			}
		}
```

components/compliance-service/reporting/relaxting/suggestions.go
alexpop: Great suggestion, I've done it across all suggestion functions in this commit: https://github.com/chef/automate/pull/1505/commits/d8aae9a4bb6a27ff2890919fa2de4327123c2698

components/compliance-service/reporting/relaxting/suggestions.go
stevendanna: Ha! I misread this when reviewing.  For some reason I didn't see that the typeParam comparison here was different than above.  Looks good as-is.

components/compliance-service/reporting/relaxting/suggestions.go
alexpop: Good idea, I pushed this commit for it: https://github.com/chef/automate/pull/1505/commits/0d988dd666f77ed5f09835669bcc8bf9462df1c3

components/compliance-service/reporting/relaxting/suggestions.go
alexpop: Makes sense, I wrapped similar errors in the file and stopped logging them out: https://github.com/chef/automate/pull/1505/commits/563b2d6ea24aff6df570b0fe4941487792daf32e

components/compliance-service/reporting/relaxting/suggestions.go
phiggins: I guess in this case I would expect all of the values to be in the values (eg, `type: 'control_tag', values: ['scope', 'Apache']`. If this is the convention the code is using this is fine then.

components/compliance-service/reporting/relaxting/suggestions.go
alexpop: Thank you, I split the code like that to make it easier to read.

components/compliance-service/reporting/relaxting/suggestions.go
alexpop: Done, cheers!

components/compliance-service/reporting/relaxting/suggestions.go
alexpop: I see what you mean. I looked into the code and the go `float32` trickled down from the initial proto type of `float` used for the suggestion score:

https://github.com/chef/automate/blob/ap/api-filter-by-control-tags/components/compliance-service/api/reporting/reporting.proto#L178

Is it worth updating the proto for this?

components/compliance-service/api/reporting/server/server.go
```diff
@@ -139,7 +139,7 @@ func (srv *Server) ListSuggestions(ctx context.Context, in *reporting.Suggestion
 	if err != nil {
 		return nil, errorutils.FormatErrorMsg(err, "")
 	}
-	suggestionsList, err := srv.es.GetSuggestions(in.Type, formattedFilters, in.Text, in.Size)
+	suggestionsList, err := srv.es.GetSuggestions(ctx, in.Type, formattedFilters, in.Text, in.Size)
```
vjeffrey: yay for passing ctx through!!!!

PR #1525, opened 2019-09-09
data-feed-service: naively apply cereal library

stevendanna:
This uses the cereal library in the data-feed-service in the most
naive way. It now uses the cereal library to manage the periodic task
management. The entire polling operation is a single task and we are
not currently saving the last polled interval.


components/data-feed-service/service/data-feed-service.go
```diff
@@ -59,95 +55,166 @@ type NotificationSender interface {
...
+	dataFeedPollTask, err := NewDataFeedPollTask(dataFeedConfig, connFactory)
+	if err != nil {
+		return err
+	}
+
+	err = manager.RegisterWorkflowExecutor("data-feed-workflow", patterns.NewSingleTaskWorkflowExecutor("data-feed-poll", false))
+	if err != nil {
+		return err
+	}
+	err = manager.RegisterTaskExecutor("data-feed-poll", dataFeedPollTask, cereal.TaskExecutorOpts{
```
jaym: [nit] maybe pull `"data-feed-poll"` out into a constant

components/data-feed-service/service/data-feed-service.go
```diff
@@ -59,95 +55,166 @@ type NotificationSender interface {
...
+		return err
+	}
+
+	dataFeedTaskParams := DataFeedPollTaskParams{
+		FeedInterval:    dataFeedConfig.ServiceConfig.FeedInterval,
+		AssetPageSize:   dataFeedConfig.ServiceConfig.AssetPageSize,
+		ReportsPageSize: dataFeedConfig.ServiceConfig.ReportsPageSize,
+	}
+
+	err = manager.CreateWorkflowSchedule("periodic-data-feed-workflow", "data-feed-workflow", dataFeedTaskParams, true, r)
```
jaym: [nit] pull `"data-feed-workflow"` out into a constant

components/data-feed-service/service/data-feed-service.go
```diff
@@ -59,95 +55,166 @@ type NotificationSender interface {
...
+	if err == cereal.ErrWorkflowScheduleExists {
+		// TODO(ssd) 2019-09-09: Right now this resets Dtstart on the recurrence, do we want that?
+		err = manager.UpdateWorkflowScheduleByName(context.Background(), "periodic-data-feed-workflow", "data-feed-workflow",
+			cereal.UpdateParameters(dataFeedTaskParams),
+			cereal.UpdateEnabled(true),
```
jaym: is it necessary to enable it if it is disabled?

components/data-feed-service/habitat/default.toml
```diff
@@ -1,7 +1,7 @@
 [service]
 host = "localhost"
 port = 14001
-feed_interval = "30m"
+feed_interval = "1m"
```
satellite15: Any reason for the 1m change? 

I have been running with 3m for development and test purposes but. 30m seems more suitable for a production setting.

stevendanna: Yeah, this just snuck in from development

stevendanna:
I've marked this ready-fore-review.  We still aren't handling the persistence of the last uploaded interval.  We can do that now if we think a future migration will be a pain, but since this is still a POC feature, I think we can do that as a follow up.

jaym:
:shipit: I imagine migration won't be horrible

PR #1537, opened 2019-09-09
removed run initiator and run type

susanev:

These were hard-coded as _Not Available_, so I removed them.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

#### Before
node-detail-metadata

#### After
new-node-detail-metadata


phiggins:
Should we move some things from the right column over to the left one to balance them out?

susanev:
@phiggins yes but not in this pr, im going to to some design work to get some more data added there and we can clean up the visuals then.

PR #1541, opened 2019-09-09
Use custom admin-only IAM resources/actions for disconnected svcs config

danielsdeleo:

Changes the authorization annotations for APIs concerning periodic disconnected services jobs so they are similar to the ones applied to config management/ingest missing nodes jobs.

#1371 

I followed the example of the job scheduler APIs for ingest, which are here: https://github.com/chef/automate/blob/8774493f9f467af54208981fcb79ef1a878efbaf/api/external/ingest/job_scheduler.proto#L16-L49

For example, the API that sets the config for the missing nodes task is:

```
    option (chef.automate.api.policy).resource = "cfgmgmt:scheduler:missing-nodes";
    option (chef.automate.api.policy).action = "configure";
    option (chef.automate.api.iam.policy).resource = "retention:nodes";
    option (chef.automate.api.iam.policy).action = "retention:nodes:update";
```

Based on reading the code, I believe that admin-only entities such as these don't need to be added to any policies because they will be covered by the catch-all policy for admins and we don't have any built-in concept of a not-quite-super-admin. LMK if this is incorrect.

- [ ] Tests added/updated?
- [ ] Docs added/updated?


api/external/applications/applications.proto
```diff
@@ -89,44 +89,44 @@ service ApplicationsService {
...
     option (google.api.http).get = "/beta/retention/service_groups/delete_disconnected_services/config";
-    // TODO: need new resources/policies for this
-    option (chef.automate.api.policy).resource = "service_groups";
-    option (chef.automate.api.policy).action = "list";
-    option (chef.automate.api.iam.policy).resource = "applications:serviceGroups";
-    option (chef.automate.api.iam.policy).action = "applications:serviceGroups:list";
+
+    option (chef.automate.api.policy).resource = "service_groups:scheduler:delete_disconnected_services";
+    option (chef.automate.api.policy).action = "read";
+    option (chef.automate.api.iam.policy).resource = "retention:serviceGroups";
```
bcmdarroch: Have we gotten product input on the namespace `retention`? To me `applications` has more meaning. From [this old spreadsheet](https://docs.google.com/spreadsheets/d/1ccaYDJdMnHBfFgmNC1n2_S1YOnMJ-OgkYd8ySbb-mS0/edit#gid=363200100), here are the other namespaces:

```

bcmdarroch: i did just see your note in slack about job scheduler already using `retention`. so I guess it's fine 🤷‍♀

PR #1542, opened 2019-09-09
Update unknown URL to reroute to event feed page

lancewf:
If a user types in the wrong URL they will be rerouted to the event feed page instead of the client runs page. 

1. Type in an incorrect URL like "https://a2-dev.test/incorrect"
1. Ensure the event feed page opens.

### DIFF
--- a/components/automate-ui/src/app/app-routing.module.ts
+++ b/components/automate-ui/src/app/app-routing.module.ts
@@ -305,14 +305,14 @@ const routes: Routes = [
     pathMatch: 'prefix',
     redirectTo: 'settings/node-credentials'
   },
-  { // ued by projects-filter.service.ts
+  { // used by projects-filter.service.ts
     path: 'reload',
     children: []
   },
   // END Deprecated routes.
   { // everything unknown goes to client runs
     path: '**',
-    redirectTo: 'infrastructure/client-runs'
+    redirectTo: 'dashboards/event-feed'
   }
 ];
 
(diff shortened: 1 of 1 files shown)


components/automate-ui/src/app/app-routing.module.ts
```diff
@@ -305,14 +305,14 @@ const routes: Routes = [
     pathMatch: 'prefix',
     redirectTo: 'settings/node-credentials'
   },
-  { // ued by projects-filter.service.ts
+  { // used by projects-filter.service.ts
     path: 'reload',
     children: []
   },
   // END Deprecated routes.
   { // everything unknown goes to client runs
```
srenatus:
```suggestion
  { // everything unknown goes to event feed
```
or, even better,
```suggestion
```

lancewf:
There is talk about changing this to not redirect but to return a 404. 
https://chefio.slack.com/archives/C6JED4R8W/p1568073655082600

lancewf:
@susanev I do not know an easy way to return a 404. If I remove the redirecting lines we see a blank page with errors in the console. 
Do you want to close this PR and just wait on UX to decided what they want to do?

srenatus:
💭 Doesn't "return a 404" mean _create a new component that displays a giant 404_ and route to that?

susanev:
@lancewf you can merge this, i think going to the event feed is better than client runs for now. ill make a separate issue to revisit having a 404 page.

lancewf:
Yeah, I was thinking that angular would have a default 404 page, but I can not trigger it. I don't think we would want to use the default one anyway.

PR #1546, opened 2019-09-10
Bump Go version to 1.13

stevendanna:
The Go version used by the build is controlled by the dependencies of
core/scaffolding-go, chef/scaffolding-go, and
chef/automate-scaffolding-go. Those packages have all been updated to
use Go 1.13. This GOLANG_VERSION file controls the version of Go that
gets installed in some build jobs and used in some local development
cases.

Merging this is going to force some developers to update their local
version of Go.

Bumping the Go version forces us to upgrade golangci-lint since we need
a version compiled with the same version of Go as the one we are using.
The new version of golangci-lint found a few more violations and added two linters.
I've disabled the new linters for now so those can be fixed in follow-up PRs.


.golangci.toml
```diff
@@ -8,7 +8,9 @@ exclude-use-default = false
 [linters]
 enable-all = true
 disable = [
+  "bodyclose",
```
stevendanna: We probably want this linter but since the fixes involve real code changes in some places, we can sort it out as a separate PR.

.golangci.toml
```diff
@@ -8,7 +8,9 @@ exclude-use-default = false
...
   "dupl",
+  "funlen",
```
stevendanna: good luck applying this.

components/automate-deployment/pkg/backup/runner.go
```diff
@@ -347,7 +347,7 @@ func (r *Runner) ListBackups(ctx context.Context) ([]*api.BackupTask, error) {
 		// to keep track of backups
 		if prefixStr != ".incomplete" && prefixStr != es.RepoBaseName && prefixStr != ".tmp" {
 			t := r.backupTaskFromSharedPrefix(prefixStr)
-			state := api.BackupTask_IN_PROGRESS
+			state := api.BackupTask_IN_PROGRESS // nolint:ineffassign
```
srenatus: `ineffassign` is the worst 😄 
```suggestion
			var state api.BackupTask_BackupState
```

PR #1551, opened 2019-09-10
Introduce authz and validate projects

blakestier:

We've introduced authz-service as a dependency of authn-service in order to validate project assignment on tokens.

### :+1: Definition of Done

- [x] you cannot assign a project that doesn't exist to a token
- [x] you cannot assign a project you don't have permission to assign to a token
- [x] Cypress integration tests that test the above behavior for all of our APIs with projects

Prep
```
export TOK=`chef-automate iam token create ADM --admin`
export TOK2=`chef-automate iam token create regular-token`
```

create a token with a project that doesn't exist
```
curl -kH "api-token: $TOK" -X POST -d '{"name": "new token", "id": "new-tok", "active": true, "projects": ["p1"] }' https://localhost/apis/iam/v2beta/tokens | jq

"error": "rpc error: code = NotFound desc = You cannot modify a project for this object because this project did not exist: [\"p1\"]",
```

create a token with no projects (unassigned)
```
curl -kH "api-token: $TOK" -X POST -d '{"name": "new token", "id": "new-tok", "active": true, "projects": [] }' https://localhost/apis/iam/v2beta/tokens | jq

success
```

try to update that token with a project that doesn't exist
```
curl -kH "api-token: $TOK" -X PUT -d '{"name": "brand new token", "active": true, "projects": ["p1"] }' https://localhost/apis/iam/v2beta/tokens/new-tok | jq

"error": "rpc error: code = NotFound desc = You cannot modify a project for this object because this project did not exist: [\"p1\"]",
```

create 2 projects
```
curl -kH "api-token: $TOK" -X POST -d '{"name": "avengers", "id": "avengers" }' https://localhost/apis/iam/v2beta/projects | jq

curl -kH "api-token: $TOK" -X POST -d '{"name": "xmen", "id": "xmen" }' https://localhost/apis/iam/v2beta/projects | jq
```

create a policy authorizing project `avengers`
```
echo '{
   "name": "avengers token admin",
   "id": "avengers-token-admin",
   "members": [
     "token:regular-token"
   ],
   "statements": [
     {
       "effect": "ALLOW",
       "actions": [
            "iam:tokens:*",
            "iam:projects:assign"
        ],
       "projects": [
         "avengers"
       ]
     }
   ],
   "projects": []
 }' >> pol.json

curl -kH "api-token: $TOK" -X POST -d @pol.json https://localhost/apis/iam/v2beta/policies | jq

success
```

create a token with an allowed project
```
curl -kH "api-token: $TOK2" -X POST -d '{"name": "new token", "id": "new-tok-2", "active": true, "projects": ["avengers"] }' https://localhost/apis/iam/v2beta/tokens | jq

success
```

create a token with an unauthorized project
```
curl -kH "api-token: $TOK2" -X POST -d '{"name": "new token", "id": "new-tok-3", "active": true, "projects": ["xmen"] }' https://localhost/apis/iam/v2beta/tokens | jq

"error": "rpc error: code = PermissionDenied desc = You cannot modify a project for this object because you are missing iam:projects:assign on this project: [\"avengers\"]",
```

try to update that token with an unauthorized project
```
curl -kH "api-token: $TOK2" -X PUT -d '{"name": "brand new token", "active": true, "projects": ["avengers", "xmen"] }' https://localhost/apis/iam/v2beta/tokens/new-tok-2 | jq

error
```

create a new token, update the policy, and update the new token with the newly allowed project
```
curl -kH "api-token: $TOK2" -X POST -d '{"name": "new token", "id": "new-tok-3", "active": true, "projects": ["avengers"] }' https://localhost/apis/iam/v2beta/tokens | jq

echo '{
   "name": "avengers token admin",
   "id": "avengers-token-admin",
   "members": [
     "token:regular-token"
   ],
   "statements": [
     {
       "effect": "ALLOW",
       "actions": [
            "iam:tokens:*",
            "iam:projects:assign"
        ],
       "projects": [
         "avengers",
         "xmen"
       ]
     }
   ],
   "projects": []
 }' >> pol2.json

curl -kH "api-token: $TOK" -X PUT -d @pol2.json https://localhost/apis/iam/v2beta/policies/avengers-token-admin | jq

curl -kH "api-token: $TOK2" -X PUT -d '{"name": "shiny new token", "active": true, "projects": ["avengers", "xmen"] }' https://localhost/apis/iam/v2beta/tokens/new-tok-3 | jq

success
```

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/authn-service/server/grpc.go
```diff
@@ -14,19 +14,21 @@ import (
...
-	authzClient authz.SubjectPurgeClient
+	ts                 tokens.Storage
+	authzSubjectClient authz.SubjectPurgeClient
+	authzV2Client      authz_v2.AuthorizationClient
 }
 
 // NewGRPCServer returns a server that provides our services: token
 // and authentication requests.
-func (s *Server) NewGRPCServer(authzClient authz.SubjectPurgeClient) *grpc.Server {
+func (s *Server) NewGRPCServer(authzSubjectClient authz.SubjectPurgeClient, authzV2Client authz_v2.AuthorizationClient) *grpc.Server {
```
srenatus: 💭 [`authz_v2.AuthorizationClient`](https://github.com/chef/automate/blob/6c7875669ee458ccd58142a3e015f54baf7411b0/api/interservice/authz/v2/authz.pb.go#L596-L602) is quite a big interface. In the interest of making it simpler to mock this for tests, and of being more explicit about **what we need here**, we could introduce a smaller interface, 
```go
type ProjectChecker /* naming is hard */ interface {
    ValidateProjectAssignment(context.Context, *ValidateProjectAssignmentReq, ...grpc.CallOption) (*ValidateProjectAssignmentResp, error)
}
```
and use that instead. Any proper `authz_v2.AuthorizationClient` satisfies that, so we don't need to change a thing in the non-test code. Just a thought, though.

components/authn-service/tokens/pg/adapter.go
```diff
@@ -74,38 +87,76 @@ func (a *adapter) insertToken(ctx context.Context,
...
+			ProjectIds: projectDiff,
+		})
+		if err != nil {
+			// TODO verify/decide if this is what we want
+			// return error unaltered because it's already a GRPC status code
+			return nil, err
+		}
+	}
+
+	// TODO why are we doing this? What if they want to delete the description?
```
srenatus: Good point. Let's 💣it.

components/authn-service/tokens/pg/adapter.go
```diff
@@ -74,38 +87,76 @@ func (a *adapter) insertToken(ctx context.Context,
...
 	if description != "" {
 		row = a.db.QueryRowContext(ctx,
```
srenatus: Since we're opening a tx above, we should probably use it here. Otherwise, this request will be _outside_ of the transaction.

components/authn-service/tokens/pg/pg.go
```diff
@@ -8,11 +8,13 @@ import (
...
 
 	// adapter for database/sql
-	_ "github.com/lib/pq"
-
+	authz_v2 "github.com/chef/automate/api/interservice/authz/v2"
 	tokens "github.com/chef/automate/components/authn-service/tokens/types"
 	platform_config "github.com/chef/automate/lib/platform/config"
 	"github.com/chef/automate/lib/tls/certs"
+
+	// TODO do we need this?
```
srenatus: Does it work without? 😉 
See also [its docs](https://godoc.org/github.com/lib/pq); the reason for importing this with an underscore is that
- we need the postgres code to be loaded
- we don't use the package in any way, we only use `database/sql`

Hence the underscore import. 😃 

For the curious, the registration stuff happens here: https://github.com/lib/pq/blob/3427c32cb71afc948325f299f040e53c1dd78979/conn.go#L51-L53 -- so when the package is imported like this, its `init` function will be run and the `database/sql` package suddenly knows now to speak PG.

components/authn-service/tokens/pg/adapter.go
blakestier: I think the problem is that description is required on v1 . . . but still, we *should* just be throwing an error at the proto level . . . not this sneaky sneak stuff

bcmdarroch: I think this is so we can sort of "patch" tokens when we just want to toggle active. But yeah I don't think it's worth keeping that distinction

components/authn-service/tokens/integration/token_integration_test.go
```diff
@@ -134,7 +135,7 @@ func TestChefClientAuthn(t *testing.T) {
 
 			// set up a token to authenticate with
 			{
-				resp, err := tokenClient.CreateToken(ctx, &authn.CreateTokenReq{Active: true, Description: "mytoken", Projects: []string{"default"}})
+				resp, err := tokenClient.CreateToken(ctx, &authn.CreateTokenReq{Active: true, Description: "mytoken", Projects: []string{}})
```
blakestier: quick fix

components/authn-service/tokens/types/types.go
```diff
@@ -42,9 +44,14 @@ type Resetter interface {
 	Reset(context.Context) error
 }
 
+// ProjectValidator is the more limited interface for authz-service
```
msorens:
```suggestion
// ProjectValidator is a more limited interface for authz-service
```

components/authz-service/server/v2/system_integration_test.go
```diff
@@ -57,6 +59,103 @@ func TestIntegrationSystemPolicies(t *testing.T) {
 	}
 }
 
+// bug: this test passes contingent on the above test being run
```
msorens: Help the reader parse the multi-line comment. 😁 
```suggestion
// bug: this test passes contingent on the above test being run.
```

components/authz-service/server/v2/system_integration_test.go
```diff
@@ -57,6 +59,103 @@ func TestIntegrationSystemPolicies(t *testing.T) {
...
+	req := api_v2.CreatePolicyReq{
+		Id:         "policy-2",
+		Name:       "my favorite policy",
+		Members:    []string{user},
+		Statements: []*api_v2.Statement{&statement},
+	}
+	_, err = ts.Policy.CreatePolicy(ctx, &req)
+	require.NoError(t, err)
+
+	// force sync refresh
```
msorens: Could we be more specific? What are we refreshing? Right now the comment does not really add any value.

components/authz-service/server/v2/system_integration_test.go
```diff
@@ -57,6 +59,103 @@ func TestIntegrationSystemPolicies(t *testing.T) {
...
+
+	cases := map[string]func(*testing.T){
+		"when passed one unauthorized project": func(t *testing.T) {
+			_, err := cl.ValidateProjectAssignment(ctx, &api_v2.ValidateProjectAssignmentReq{
+				Subjects:   []string{user},
+				ProjectIds: []string{unauthorizedProjectId},
+			})
+			grpctest.AssertCode(t, codes.PermissionDenied, err)
+		},
+		"when passed one not found project": func(t *testing.T) {
```
msorens: Slight grammar improvement:
```
"when passed one non-existent project": func(t *testing.T) {
```
(Also applies to several occurrences below)

components/authz-service/server/v2/system_integration_test.go
```diff
@@ -57,6 +59,103 @@ func TestIntegrationSystemPolicies(t *testing.T) {
...
+				ProjectIds: []string{authorizedProjectId, "project-not-found"},
+			})
+			grpctest.AssertCode(t, codes.NotFound, err)
+		},
+		"when passed one unauthorized project and one not found project": func(t *testing.T) {
+			_, err := cl.ValidateProjectAssignment(ctx, &api_v2.ValidateProjectAssignmentReq{
+				Subjects:   []string{user},
+				ProjectIds: []string{unauthorizedProjectId, "project-not-found"},
+			})
+			grpctest.AssertCode(t, codes.NotFound, err)
```
msorens: Would we really consider it a failure if it returned codes.PermissionDenied ? 🤔 
Maybe. But if not, perhaps check for one or the other.

components/authz-service/server/v2/system_integration_test.go
```diff
@@ -57,6 +59,103 @@ func TestIntegrationSystemPolicies(t *testing.T) {
...
+	err = ts.PolicyRefresher.Refresh(ctx)
+	require.NoError(t, err)
+
+	cases := map[string]func(*testing.T){
+		"when passed one unauthorized project": func(t *testing.T) {
+			_, err := cl.ValidateProjectAssignment(ctx, &api_v2.ValidateProjectAssignmentReq{
+				Subjects:   []string{user},
+				ProjectIds: []string{unauthorizedProjectId},
+			})
+			grpctest.AssertCode(t, codes.PermissionDenied, err)
```
msorens: Would like to see also checking for the project name in the error text.
(Similar for other tests below.)

components/authz-service/server/v2/system_integration_test.go
```diff
@@ -57,6 +59,103 @@ func TestIntegrationSystemPolicies(t *testing.T) {
...
+		},
+		"when passed one authorized project": func(t *testing.T) {
+			_, err := cl.ValidateProjectAssignment(ctx, &api_v2.ValidateProjectAssignmentReq{
+				Subjects:   []string{user},
+				ProjectIds: []string{authorizedProjectId},
+			})
+			assert.NoError(t, err)
+		},
```
msorens: Consider adding a test with two unauthorized projects and another with two non-existent projects. Then, when you check the error text, too (as mentioned above).
Those will answer this question: does the error message mention all such projects or just the first one?

e2e/cypress/integration/api/iam/05_iam_integration.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+          auth: { bearer: adminIdToken },
+          method: 'POST',
+          url: '/apis/iam/v2beta/projects',
+          failOnStatusCode: false,
+          body: {
+            id: id,
+            name: id
+          }
+        }).then((resp) => {
+          expect([200, 409]).to.include(resp.status);
```
msorens: This has been in the back of my mind for some time, so I finally looked up the right function.
We have been using `expect(x)` incorrectly. `expect(x)` is supposed to mean `expect(<actual-result>)` not `expect(<expected-result>)`. Let's fix this:
```suggestion
          expect(resp.status).to.be.oneOf([200, 409]);
```

e2e/cypress/integration/api/iam/05_iam_integration.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+      }).then((resp) => {
+        assert.equal(resp.status, 200);
+        assert.isUndefined(resp.body.error);
+        assert.deepEqual(
+            resp.body[`${iamResource}`]['projects'],
+            [authorizedProject1, authorizedProject2]);
+      });
+    });
+
+    it(`when creating ${iamResources}, fails to assign a non-existing project`, () => {
```
msorens: s/non-existing/non-existent/

blakestier: lol

e2e/cypress/integration/api/iam/05_iam_integration.spec.ts
blakestier: ah didn't know `oneOf` 👍

Issue #1341, opened 2019-08-24
global project filter apply should reset pagination


susanev:
_Originally reported by @msorens._

## Overview
When applying a new global filter in the top-right any page that is on a paginated page greater than 1 should be reset back to page 1. This seems to be an issue wherever we have pagination including _Applications_, _Client Runs_, and _Compliance_

## Repro Steps

- `chef-automate iam upgrade-to-v2 --beta2.1` if not already on 2.1
- `chef_load_nodes`, youll need more than 100 to get some pagination
- create a few projects in _Settings_ > _Projects_, in those projects add an ingest rule so you get some nodes in different projects
- Navigate to _Infrastructure_ > _Client Runs_ the navigate to a page greater than 1
- Apply a new global projects filter
- Watch the page refresh and filter the nodes, but notice that you are not sent back to page 1

![pagination-fun](https://user-images.githubusercontent.com/5489125/63640134-966eb880-c651-11e9-939d-c169cbe84e03.png)

PR #1553, opened 2019-09-10
Client Runs page, detecting bad page selection

lancewf:
This change allows the client runs page to revert back to the first page of nodes if the page selected does not have nodes. 

This problem was found with the IAM v2 global project filtering feature. When the project filter is updated it refreshes the page. Say the user is on page 2 with 100 nodes per page. If the user updates the project filter to have only 50 nodes then page 2 would be empty. Since the page 2 selection is in the URL after the refresh the user would be left with zero nodes. 

#1341 
### :+1: Definition of Done

The repro steps in issue #1341 are resolved. 

1. Build the changed code and start Automate: `build components/automate-ui && start_all_services`
1. Turn on IAM v2: `chef-automate iam upgrade-to-v2 --beta2.1`
1. Add some nodes to full at least two pages `chef_load_nodes 120`
1. Find the environment of one of less than 100 nodes. I found this from the search bar. 
1. Create and apply a project with the rule of the environment found above. 
1. Go to the client runs page https://a2-dev.test/infrastructure/client-runs.
1. Go to the second page of nodes. 
1. Apply the newly created project filter. 
1. Ensure that the client runs page goes back to the first page. 

- [x] Tests added/updated?
- [ ] Docs added/updated?

project_filter_client_runs

### DIFF
--- a/components/automate-ui/src/app/pages/client-runs/client-runs.component.ts
+++ b/components/automate-ui/src/app/pages/client-runs/client-runs.component.ts
@@ -580,7 +580,7 @@ export class ClientRunsComponent implements OnInit, OnDestroy {
 
   // If there is a page parameter in the URL the first page is not selected.
   isNotFirstPage(allUrlParameters: Chicklet[]): boolean {
-    return allUrlParameters.findIndex((parm: Chicklet) => parm.type === 'page') >= 0;
+    return allUrlParameters.some((parm: Chicklet) => parm.type === 'page');
   }
 
   private getSelectedPageNumber(allUrlParameters: Chicklet[]): number {
--- a/components/automate-ui/src/app/pages/client-runs/client-runs.component.ts
+++ b/components/automate-ui/src/app/pages/client-runs/client-runs.component.ts
@@ -331,6 +331,26 @@ export class ClientRunsComponent implements OnInit, OnDestroy {
 
     this.loadedStatus$ = this.store.select(clientRunsLoading);
 
+
+    // When zero nodes are loaded successfully and we are not on the 1 page
+    // automatically change to the first page.
+    combineLatest([
+      this.nodes$, this.loadedStatus$, allUrlParameters$])
+      .pipe(
+        map(([nodes, loadedStatus, allUrlParameters]:
+          [Node[], EntityStatus, Chicklet[]]) =>
+            this.isCurrentPageEmpty(nodes, loadedStatus) &&
+            this.isNotFirstPage(allUrlParameters)
+        )).subscribe((noNodesLoadedWithPageSet: boolean) => {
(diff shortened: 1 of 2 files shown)


phiggins:
Would it make sense to always go back to the first page if I change the project filter? Maybe I don't have the right model in my head, but I feel like if I change the project filter it's like I'm doing a new search and not a continuation of my existing one.

lancewf:
@phiggins Yeah, I think that does make sense. That is what we do for all the other filter selections on the client runs page. For the project filter, we do not have an observer to when those changes are made. The way the project filter select changes are currently working is it refreshes the page.

components/automate-ui/src/app/pages/client-runs/client-runs.component.ts
```diff
@@ -555,6 +574,15 @@ export class ClientRunsComponent implements OnInit, OnDestroy {
     this.store.dispatch(new UpdateNodeFilters({filters: nodeFilters}));
   }
 
+  isCurrentPageEmpty(nodes: Node[], loadedStatus: EntityStatus): boolean {
+    return loadedStatus === EntityStatus.loadingSuccess && nodes.length === 0;
+  }
+
+  // If there is a page parameter in the URL the first page is not selected.
+  isNotFirstPage(allUrlParameters: Chicklet[]): boolean {
+    return allUrlParameters.findIndex((parm: Chicklet) => parm.type === 'page') >= 0;
```
srenatus: [nit] How about [`some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)? It's pretty much the same except that it's already a boolean:
```suggestion
    return allUrlParameters.some((parm: Chicklet) => parm.type === 'page');
```

lancewf: Thank you! 
I always forget about 'some'. I wish there was an 'exists'.

components/automate-ui/src/app/pages/client-runs/client-runs.component.ts
```diff
@@ -580,7 +580,7 @@ export class ClientRunsComponent implements OnInit, OnDestroy {
 
   // If there is a page parameter in the URL the first page is not selected.
   isNotFirstPage(allUrlParameters: Chicklet[]): boolean {
-    return allUrlParameters.findIndex((parm: Chicklet) => parm.type === 'page') >= 0;
+    return allUrlParameters.some((parm: Chicklet) => parm.type === 'page');
```
srenatus: [nit] `parm`? interesting choice. But it's OK, of course.

lancewf: Oh, I meant to have 'param', but I changed it to have the full 'parameter'.

PR #1574, opened 2019-09-11
[AUTH-1477] If there are no projects with pending rule edits, disable the update button with proper text

tylercloke:

The rest of the button's behavior should be exactly the same as before.

```
start_all_services
# start webserver locally
```

1. Button is disabled with text `Projects Up-to-date` when no projects
1. Create a project (or multiple)
1. Button still disabled because no pending edits
1. Add a rule to 1 or more projects
1. Button behaves as before
1. Apply projects
1. Button goes back to `Projects Up-to-date` disabled state

- [x] Tests added/updated?
- [ ] Docs added/updated?

Screen Shot 2019-09-11 at 1 33 52 PM

### DIFF
--- a/components/automate-ui/src/app/pages/project/list/project-list.component.ts
+++ b/components/automate-ui/src/app/pages/project/list/project-list.component.ts
@@ -112,7 +112,7 @@ export class ProjectListComponent implements OnInit, OnDestroy {
       takeUntil(this.isDestroyed)
     ).subscribe((projectList: Project[]) => {
       this.projectsHaveStagedChanges =
-        projectList.filter(p => p.status === 'EDITS_PENDING').length > 0;
+        projectList.some(p => p.status === 'EDITS_PENDING');
     });
 
     this.createProjectForm = fb.group({
--- a/components/automate-ui/src/app/pages/project/list/project-list.component.ts
+++ b/components/automate-ui/src/app/pages/project/list/project-list.component.ts
@@ -54,6 +54,11 @@ export class ProjectListComponent implements OnInit, OnDestroy {
   // is not available soon enough--we need to know the instant the user starts the update.
   private applyRulesInProgress = false;
 
+  // True if there are any rules that have a status of 'EDITS_PENDING'.
+  private projectsHaveStagedChanges = false;
+
+  private percentageComplete = 0;
+
   private updateProjectsFailed = false;
   private updateProjectsCancelled = false;
   public cancelRulesInProgress = false;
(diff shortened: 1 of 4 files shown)


blakestier:
Maybe this is how we indicate that a project with rules has been deleted? 🤔

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -145,6 +147,16 @@ export class ProjectListComponent implements OnInit, OnDestroy {
...
   }
 
+  updateButtonLabel(): string {
+    if (this.applyRulesInProgress) {
+      return `Updating Projects ${Math.round(this.percentageComplete * 100)}%...`;
+    }
+    if (this.projectsHaveStagedChanges) {
+      return 'Update Projects';
+    }
+    return 'Projects Up-to-date';
```
blakestier: This looks like it should be capitalized all the way through to me 🙃 `Projects Up-to-Date`

susanev: i wasnt sure... sounds good to me though, thanks @blakestier!

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -80,23 +85,13 @@ export class ProjectListComponent implements OnInit, OnDestroy {
...
-        }
-      })
-    );
-
     this.projects.applyRulesStatus$
-      .subscribe(({ state, failed, cancelled }: ApplyRulesStatus) => {
+      .subscribe(({ state, failed, cancelled, percentageComplete }: ApplyRulesStatus) => {
         if (state === ApplyRulesStatusState.NotRunning) {
           this.applyRulesInProgress = false;
           this.updateProjectsFailed = failed;
```
msorens: I just noticed there's a bug that I created here. L92-96 should be after the condition block on L90 closes. Otherwise your percentage won't change while the update is running. 😁 
Because the rebase-with-master adds changes, too, I'm just going to paste in what I think the net result should be--see if it makes sense. This includes my next comment on your L94, too:
```
        if (state === ApplyRulesStatusState.NotRunning) {
          if (this.applyRulesInProgress) {
            this.cancelRulesInProgress = false;
          }
          this.applyRulesInProgress = false;
        }
        this.updateProjectsFailed = failed;
        this.updateProjectsCancelled = cancelled;
        if (!this.cancelRulesInProgress) {
          this.percentageComplete = percentageComplete;
        }
```

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -145,6 +147,16 @@ export class ProjectListComponent implements OnInit, OnDestroy {
     return '';
   }
 
+  updateButtonLabel(): string {
```
msorens: Consider renaming this to `getButtonText` to be more consistent with other functions in this file (`getRulesStatus`, `getProjectStatus`).

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -151,6 +148,16 @@ export class ProjectListComponent implements OnInit, OnDestroy {
...
+  getButtonText(): string {
+    if (this.applyRulesInProgress) {
+      return `Updating Projects ${Math.round(this.percentageComplete * 100)}%...`;
+    }
+    if (this.projectsHaveStagedChanges) {
+      return 'Update Projects';
+    }
+    return 'Projects Up-To-Date';
+  }
+
```
srenatus: side note -- as mentioned in one of @msorens' PRs recently -- I think it's preferrable to keep all text-y things in the template. Looking at the `component.html`, all we know is that the button is `getButtonText()`, we then have to lookup what this can be in the `component.ts`. here, however, we'd have to nest ternary operators in the `component.html`, and I suppose that's worse 😄

Issue #1565, opened 2019-09-11
Automate does not export tags with client runs


tommybonobo:
## Automate does not export tags with client runs
Automate can show tags while on the web gui's `Client Runs` page but doing the following:
    - Select a node that has been tagged
    - Navigate to `Attributes`
    - Find `tags` array and expand it
    - Tag will show in array as expected

However if the `Client Runs` data is exported to `json` by  clicking the `Download JSON` link, the same node will now not have any elements in the `tags` array within the json. The `tags` array still shows but is always `null` and will look like this.

```
"tags": null
```

## To Reproduce
    - Navigate to the web gui's `Client Runs` page
    - Click the download button
    - Click `Download JSON`
    - Open the json and locate the `tags` for a node which has a tag defined as found in the web gui the method outlined above.
    - Despite having tags in the web gui, the `tags` array will be `null` like so: `"tags": null`
    - If the tag is searched for. It will not be found any where else in the exported json.

## Expected behavior
Tags which are on the gui should show somewhere inside the exported json data, preferably in the `tags` array.

## Versions (please complete the following information):
- OS: macOS
- Browser: chrome
- Automate Build Number 20190711110747

## Additional context
I have also exported the client run data via a curl using an API as documented. I found that this downloads the exact same json (no deltas) so it doesn't seem to be an issue with the way the web gui or browser downloads the file.

lancewf:
Thank you for submitting this bug!

I will try to reproduce it in the dev environment and get back to you.  I hope it is an easy fix.

lancewf:
I am seeing the same problem locally. I will see if it is an easy fix.

lancewf:
@tommybonobo turns out there was an easy fix for the problem! https://github.com/chef/automate/pull/1578. 

Thank you again for submitting the bug.

PR #1578, opened 2019-09-11
Changing to return chef tags for tags in the node export JSON

lancewf:
Changing to return chef tags for tags in the JSON node export.

https://github.com/chef/automate/issues/1565

### :+1: Definition of Done
The client runs JSON node export returns chef tags. 

1. `build components/config-mgmt-service && start_all_services`
1. Load a node with tags `send_chef_run_example`
1. Go to https://a2-dev.test/infrastructure/client-runs
1. Confirm this is tags with the search bar by looking at the "Chef Tag" suggestions. 
1. Export the one node's JSON with the Screen Shot 2019-09-11 at 4 39 57 PM button. 
1. Ensure the "slave", "supermarket", and "builder" tags are present. 

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/config-mgmt-service/integration_test/node_export_test.go
+++ b/components/config-mgmt-service/integration_test/node_export_test.go
@@ -25,6 +25,81 @@ import (
 	"time"
 )
 
+func TestNodeExportMissingTags(t *testing.T) {
+	esClient := elastic.New(elasticsearchUrl)
+	c := config.New(esClient)
+	server := grpcserver.NewCfgMgmtServer(c)
+
+	lis := bufconn.Listen(1024 * 1024)
+	s := grpc.NewServer()
+	api.RegisterCfgMgmtServer(s, server)
+
+	go func() {
+		if err := s.Serve(lis); err != nil {
+			t.Fatalf("Server exited with error: %v", err)
+		}
+	}()
+
+	dialer := func(string, time.Duration) (net.Conn, error) { return lis.Dial() }
+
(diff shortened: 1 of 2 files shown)


components/config-mgmt-service/integration_test/node_export_test.go
```diff
@@ -25,6 +25,81 @@ import (
 	"time"
 )
 
+func TestNodeExportMissingTags(t *testing.T) {
```
phiggins: nitpick: "Node export missing tags" describes the bug you're fixing, but not the behavior you're testing. Something like "Node export includes chef tags" describes the desired behavior.

components/config-mgmt-service/integration_test/node_export_test.go
```diff
@@ -25,6 +25,81 @@ import (
...
+		require.NoError(t, err)
+		tags := []string{}
+		jsonparser.ArrayEach(node, func(tag []byte, _ jsonparser.ValueType, _ int, err error) {
+			tags = append(tags, string(tag))
+		}, "tags")
+		nodeData = append(nodeData, tags)
+	})
+
+	assert.Equal(t, 1, len(nodeData))
+	assert.ElementsMatch(t, []string{"a", "b", "c"}, nodeData[0])
```
phiggins: If the assert above fails because the length is 0, this one will be a null pointer exception that will crash the test run. The one above should probably be `require.Equal`.

PR #1591, opened 2019-09-12
[integration] test purge diagnostics during upgrades

ryancragun:


integration/tests/iam_v2_upgrade_diagnostics.sh
```diff
@@ -4,7 +4,7 @@
 test_name="iam v2 (upgrade, diagnostics)"
 # Note: we can't run diagnostics AND inspec, so no inspec here
 #shellcheck disable=SC2034
-test_diagnostics_filters="~iam-v1 ~purge" # remove purge until 1219 makes it to current
+test_diagnostics_filters="~iam-v1"
```
srenatus: Ooops. I guess we could have used the `Skip` from #1585 to do this. OTOH, the IAMv1 endpoints should all work (still) when using v2, so this is not causing harm. 🤔

Issue #703, opened 2019-06-27
Project Update estimated complete time returns incorrect dates


lancewf:
# User Story

As an Automate API User, I will want the `estimated_time_complete` for a project update to return "0001-01-01T00:00:00Z" until it has a real estimate. 

Currently, the `estimated_time_complete` is returning dates from 1970 and a few minutes before the current time. This was the case when there were no resources to update. 

Below is how to request the time estimate. 
```
$ curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/apply-rules | jq
{
  "state": "not_running",
  "estimated_time_complete": "0001-01-01T00:00:00Z",
  "percentage_complete": 1,
  "failed": false,
  "failure_message": ""
}
```
# Definition of Done

When starting an iamv2 project rules update only "0001-01-01T00:00:00Z" or times in the future are returned. 

# Demo Script / Repro Steps
1. Add some projects with ingest rules for nodes
1. Add 1,000 nodes `chef_load_nodes 1000`
1. Watch the status endpoint https://localhost/apis/iam/v2beta/apply-rules
1. Start the projects update
1. Ensure only "0001-01-01T00:00:00Z" and dates in the future are returned.

PR #1595, opened 2019-09-12
Truncating non future estimates

lancewf:
Forcing all project update estimates to be in the future or "0001-01-01T00:00:00Z" (no estimate). During a project update, estimates of 1970 and a few minutes in the past are provided when the process is started. This change removes these bad estimates. 

#703

### :+1: Definition of Done
No estimates in the past are returned. 

1. Add compliance reports with `chef_load_compliance_scans -N1000 -D11 -M5 -T22000`
1. Turn on projects `chef-automate iam upgrade-to-v2 --beta2.1`
1. Create a project and then update all the projects. To watch the percentage and estimated time complete run `chef-automate dev grpcurl authz-service -- chef.automate.domain.authz.v2.Projects.ApplyRulesStatus`
I like to use `watch "chef-automate dev grpcurl authz-service -- chef.automate.domain.authz.v2.Projects.ApplyRulesStatus"` But you have to install watch first with `install_if_missing core/procps-ng watch`

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/authz-service/server/v2/project_update_manager.go
+++ b/components/authz-service/server/v2/project_update_manager.go
@@ -407,9 +407,10 @@ func (m *CerealProjectUpdateManager) Status() (ProjectUpdateStatus, error) {
 
 // Find the longest estimate that is in the future. If no estimate is found return zero time.Time{}
 func findLongestEstimatedTime(estimatedTimes []time.Time) time.Time {
+	now := time.Now()
 	longestEstimatedTimeComplete := time.Time{}
 	for _, estimatedTime := range estimatedTimes {
-		if estimatedTime.After(time.Now()) && estimatedTime.After(longestEstimatedTimeComplete) {
+		if estimatedTime.After(now) && estimatedTime.After(longestEstimatedTimeComplete) {
 			longestEstimatedTimeComplete = estimatedTime
 		}
 	}
@@ -423,7 +424,7 @@ func (w *workflowInstance) collectAllDomainServiceEstimates() []time.Time {
 		return estimates
 	}
 	domainServicesUpdateInstance, err := w.GetUpdateDomainServicesInstance()
-	if err == cereal.ErrWorkflowInstanceNotFound {
+	if err != nil {
 		return estimates
 	}
 
@@ -439,8 +440,10 @@ func (w *workflowInstance) collectAllDomainServiceEstimates() []time.Time {
 				logrus.WithError(err).Errorf("failed to get payload for %q", d)
 				continue
 			}
-			estimatedTime := time.Unix(payload.MergedJobStatus.EstimatedEndTimeInSec, 0)
-			estimates = append(estimates, estimatedTime)
+			if payload.MergedJobStatus.EstimatedEndTimeInSec != 0 {
+				estimatedTime := time.Unix(payload.MergedJobStatus.EstimatedEndTimeInSec, 0)
+				estimates = append(estimates, estimatedTime)
+			}
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


components/authz-service/server/v2/project_update_manager.go
```diff
@@ -432,3 +404,45 @@ func (m *CerealProjectUpdateManager) Status() (ProjectUpdateStatus, error) {
 	}
 	return projectUpdateInstance, nil
 }
+
+// Find the longest estimate that is in the future. If no estimate is found return zero time.Time{}
+func findLongestEstimatedTime(estimatedTimes []time.Time) time.Time {
+	longestEstimatedTimeComplete := time.Time{}
+	for _, estimatedTime := range estimatedTimes {
+		if estimatedTime.After(time.Now()) && estimatedTime.After(longestEstimatedTimeComplete) {
```
srenatus: It's probably inconsequential, but `time.Now()` will be different for every iteration through this loop. Maybe we should define `noew := time.Now()` before that loop and compare to this here? 💭

components/authz-service/server/v2/project_update_manager.go
```diff
@@ -432,3 +404,45 @@ func (m *CerealProjectUpdateManager) Status() (ProjectUpdateStatus, error) {
...
+
+func (w *workflowInstance) collectAllDomainServiceEstimates() []time.Time {
+	estimates := make([]time.Time, 0)
+	if !w.IsRunning() {
+		return estimates
+	}
+	domainServicesUpdateInstance, err := w.GetUpdateDomainServicesInstance()
+	if err == cereal.ErrWorkflowInstanceNotFound {
+		return estimates
+	}
```
srenatus: ❓ Can there be other errors that we might care about? -- Can `err` be not `nil` and not `cereal.ErrWorkflowInstanceNotFound`?

components/authz-service/server/v2/project_update_manager.go
```diff
@@ -432,3 +404,45 @@ func (m *CerealProjectUpdateManager) Status() (ProjectUpdateStatus, error) {
...
+			logrus.WithError(err).Errorf("failed to get subworkflow for %q", d)
+			continue
+		}
+		payload := project_update_tags.DomainProjectUpdateWorkflowPayload{}
+		if subWorkflow.IsRunning() {
+			if err := subWorkflow.GetPayload(&payload); err != nil {
+				logrus.WithError(err).Errorf("failed to get payload for %q", d)
+				continue
+			}
+			estimatedTime := time.Unix(payload.MergedJobStatus.EstimatedEndTimeInSec, 0)
```
srenatus: For some reason I don't know about, we know that `payload.MergedJobStatus` is not a thing that could be `nil` and `EstimatedEndTimeInSec` is always available?

components/authz-service/server/v2/project_update_manager.go
lancewf: I am not sure. Maybe @jaym might know this question.

jaym: i think not. but it doesn't hurt to defend against it

components/authz-service/server/v2/project_update_manager.go
lancewf: That makes sense. I will add a check to see if it is nil.

lancewf: Wait,`payload.MergedJobStatus` can not be nil, but `payload.MergedJobStatus.EstimatedEndTimeInSec` could be zero.

PR #1600, opened 2019-09-13
automate-gateway: Move license usage log message to debug

stevendanna:
This log message can be _very_ large on a machine that reports a lot
of nodes. I don't work in this code often, so I've just moved it to
Debug, but we might consider just logging the count rather then the
full report.


phiggins:
Logging the total count or even removing this log line would probably be ok.

Issue #1339, opened 2019-08-23
backend functionality demo: control tag search/filter + controls list


vjeffrey:
## User Story
In an effort to demonstrate progress and ensure we're building the right things, we need to present a demo of the backend functionality for control tag search and filter and controls list.

This issue seeks to detail the demo we will show to ensure
1) this is an appropriate demo for the folks interests
2) we ensure we do all the things

## Definition of Done
Recorded video showing the following:

- ingest some reports with control tags, at least 3 different nodes, one of the reports should have no control tags, at least one control on each of the other two reports should have matching control tags

- call the controls list api
        - show json output of that api

- call the suggestions api with a control tag filter that exists on two node reports
        - show the json list results of that api call

- call the compliance/reporting/nodes api with a control tag filter that exists on two node reports
        - show the json results of that api call

- call the controls list api with same control tag filter
        - show the json results of that api call

vjeffrey:
<img width="1243" alt="Screen Shot 2019-09-11 at 10 34 30" src="https://user-images.githubusercontent.com/10341541/64724013-6fecb200-d486-11e9-8997-fb0656e45210.png">

vjeffrey:
```
# nodes
curl --silent --insecure -H "api-token: token" https://a2-dev.test/api/v0/compliance/reporting/nodes/search -d '{}'

# controls
curl --silent --insecure -H "api-token: token" https://a2-dev.test/api/v0/compliance/reporting/controls -d '{}'

# suggestions on control tag key
curl --silent --insecure -H "api-token: token" https://a2-dev.test/api/v0/compliance/reporting/suggestions -d '{"type": "control_tag_key","filters":[] }'

# suggestions on control tag value
curl --silent --insecure -H "api-token: token" https://a2-dev.test/api/v0/compliance/reporting/suggestions -d '{"type": "control_tag_value","filters":[{"type": "control_tag:ident", "values": []}] }'

# nodes with control tag filter
curl --silent --insecure -H "api-token: token" https://a2-dev.test/api/v0/compliance/reporting/nodes/search -d '{"type": "control_tag_value","filters":[{"type": "control_tag:ident", "values": ["CCI*"]}] }'

# controls with control tag filter
curl --silent --insecure -H "api-token: token" https://a2-dev.test/api/v0/compliance/reporting/controls -d '{"type": "control_tag_value","filters":[{"type": "control_tag:ident", "values": ["CCI-002884"]}] }'
[...]

Issue #521, opened 2019-06-06
compliance reporting: controls list view


vjeffrey:
# User Story

As an Automate user, I want to be able to see a list of all the controls being used in my infrastructure scans, the same way I can currently see a list of all the nodes and profiles (https://a2-local-inplace-upgrade-acceptance.cd.chef.co/compliance/reports/overview).

Providing users with a controls list view would allow them to 
- easily see all the controls that were executed as part of the scans, as an extension of the nodes and profiles list views
- easily view all controls that have the same tag (across multiple profiles) in combination with https://github.com/chef/automate/issues/519
- easily apply a control + profile filter to enable deep filtering

## Considerations
- backend work includes creating an endpoint under compliance reporting for a controls list (like `/compliance/reporting/controls`). this should return a list of all controls executed on all nodes scanned in the given timeframe. controls list should include at least control metadata (id, title, impact, profile_sha). other fields may be needed to satisfy designs.
- frontend work includes creating a controls list view similar to the current compliance reporting nodes and profiles list views, including status filters. 

*need ux designs for this work* (i know @jonong1972 has a wip one laying around) 

# Definition of Done
List view of controls in Automate

vjeffrey:
the endpoint should be as (or more) performant as the nodes and profiles list ones

the endpoint should return:

control metadata (title/id/impact)
the profile the control belongs to (title, version, sha)
the endtime of the latest report it's attached to
be able to filter by control status

optional/extra info, can be done as later step:

tags associated with the control
count of nodes that executed the control
count of failed/passed/skipped nodes that executed the control

PR #1601, opened 2019-09-13
control-list api

rickmarry:
This change adds a new API (control-list) to the compliance reporting component. The control-list API fetches stats and other useful data about filtered controls so that problematic controls may be easily identified.  The UI will make good use of this API in order to render the list that is retrieved so that any individual element may then be selected for inclusion to the filter set for subsequent, different API calls.  Using suggestions also allows us to select and add controls to filters but using suggestions is tedious and offers no insight aimed to help form the basis of deciding on a particular control. This API was designed to offer this much-needed insight.

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/automate-gateway/api/compliance/reporting/reporting.proto
```diff
@@ -48,6 +48,22 @@ service ReportingService {
    			action: "compliance:reportids:list"
    		};
    	};
+	// should cover /control-items
+	// This api is useful for getting a limited list of control items for latest runs.  It also honors all reporting filters
+	rpc ListControlItems(Query) returns (ControlItems) {
+		option (google.api.http) = {
+   			post: "/compliance/reporting/control-items"
```
vjeffrey: i'm curious what drove the decision to name the api control-items (vs. controls)

components/compliance-service/reporting/relaxting/reports.go
```diff
@@ -530,6 +530,212 @@ func (backend *ES2Backend) GetReport(esIndex string, reportId string,
...
+	filteredControls.SubAggregation("control", controlTermsAgg)
+
+	controlsAgg := elastic.NewNestedAggregation().Path("profiles.controls")
+	controlsAgg.SubAggregation("filtered_controls", filteredControls)
+
+	searchSource.Aggregation("controls", controlsAgg)
+
+	source, err := searchSource.Source()
+	if err != nil {
+		return nil, errors.Wrap(err, fmt.Sprintf("%s unable to get Source", myName))
```
stevendanna:
```suggestion
		return nil, errors.Wrapf(err, "%s unable to get Source", myName)
```

components/compliance-service/reporting/relaxting/reports.go
```diff
@@ -530,6 +530,212 @@ func (backend *ES2Backend) GetReport(esIndex string, reportId string,
...
+	//for now, we don't search non-latest reports so don't do it.. it's slower for obvious reasons.
+	filtQuery := backend.getFiltersQuery(filters, true)
+
+	searchSource := elastic.NewSearchSource().
+		Query(filtQuery).
+		FetchSource(false).
+		Size(1)
+
+	controlTermsAgg := elastic.NewTermsAggregation().Field("profiles.controls.id").
+		Size(100).
```
stevendanna: Why do we think `100` makes sense here?

components/compliance-service/reporting/relaxting/reports.go
```diff
@@ -530,6 +530,212 @@ func (backend *ES2Backend) GetReport(esIndex string, reportId string,
...
+	source, err := searchSource.Source()
+	if err != nil {
+		return nil, errors.Wrap(err, fmt.Sprintf("%s unable to get Source", myName))
+	}
+	LogQueryPartMin(esIndex, source, fmt.Sprintf("%s query", myName))
+
+	searchResult, err := client.Search().
+		SearchSource(searchSource).
+		Index(esIndex).
+		Do(context.Background())
```
stevendanna: Should we thread the context from the request through to here?

components/compliance-service/reporting/relaxting/reports.go
```diff
@@ -530,6 +530,212 @@ func (backend *ES2Backend) GetReport(esIndex string, reportId string,
...
+	}
+
+	logrus.Debugf("Search query took %d milliseconds\n", searchResult.TookInMillis)
+
+	if outerControlsAggResult, found := searchResult.Aggregations.Nested("controls"); found {
+		if filteredControls, found := outerControlsAggResult.Aggregations.Filter("filtered_controls"); found {
+			if controlBuckets, found := filteredControls.Aggregations.Terms("control"); found && len(controlBuckets.Buckets) > 0 {
+				for _, controlBucket := range controlBuckets.Buckets {
+					contListItem := reportingapi.ControlItem{}
+					contListItem.Id = controlBucket.Key.(string)
```
stevendanna: Should we use the form `contListItem.Id, ok = controlBucket.Key.(string)` here like you do below to ensure this cast doesn't panic.

components/compliance-service/reporting/relaxting/reports.go
```diff
@@ -530,6 +530,212 @@ func (backend *ES2Backend) GetReport(esIndex string, reportId string,
...
+						impactBucket := impactAggResult.Buckets[0]
+						impactAsNumber, ok := impactBucket.Key.(float64)
+						if !ok {
+							logrus.Errorf("could not convert the value of impact: %v, to a float!", impactBucket)
+						}
+						contListItem.Impact = float32(impactAsNumber)
+						controlSummary.Total = int32(impactBucket.DocCount)
+					}
+
+					if profileResult, found := controlBucket.Aggregations.ReverseNested("profile"); found {
```
stevendanna: [minor] I'm wondering if there is a way to make some of this inner loop more clear.  One small improvement might be to move the declaration of `profileMin` right before this loop that uses it, and then assign it to contListItem.Profile right after this section rather than below at 705.

components/automate-gateway/api/compliance/reporting/reporting.proto
rickmarry: nothing.. I can change to controls

components/compliance-service/reporting/relaxting/reports.go
rickmarry: cool.. thanks @stevendanna

components/compliance-service/reporting/relaxting/reports.go
rickmarry: We don't.. this is WIP that number by completion will probably default to whatever may be retrieved while remaining performant. We will probably need to cap this at some value though, as this API does not support pagination.

components/compliance-service/reporting/relaxting/reports.go
rickmarry: Good point.  I see a lot of places where we need to thread the context through.  We should do a pr just for that in all necessary places.

components/compliance-service/reporting/relaxting/reports.go
rickmarry: fixing

components/compliance-service/reporting/relaxting/reports.go
rickmarry: done.. thanks!

components/automate-gateway/api/compliance/reporting/reporting.proto
vjeffrey: > nothing.. I can change to controls

would prob be a good post-standup convo for the group (including ux). maybe control-items makes more sense than controls, or vice-versa?  we have a few diff patterns in our api.. 🤷‍♀

components/compliance-service/reporting/relaxting/reports.go
rickmarry: done

components/compliance-service/reporting/relaxting/reports.go
rickmarry: done

components/compliance-service/reporting/relaxting/reports.go
rickmarry: done

vjeffrey:
@rickmarry could you please give us a checklist or something of the sort enumerating what all remains to be done before this can get out of WIP mode?  do you need any help? happy to throw some more ppl efforts on this to get it done. we're in a bit of a time crunch here.

rickmarry:
> @rickmarry could you please give us a checklist or something of the sort enumerating what all remains to be done before this can get out of WIP mode? do you need any help? happy to throw some more ppl efforts on this to get it done. we're in a bit of a time crunch here.

The only things that I have remaining are:
1- control tags
2- control-status filtration
3- tests

Working to get 2 and 3 done today. Once these are done, I'm thinking, I'm thinking that this may come out of WIP and we can go after control tags immediately thereafter.  It's also possible I get all three done today, in which case issue #521 will be complete

I think adding more resources to this right now may actually get in the way.  
This took a bit longer than expected because of all the reverse nested aggregations and diving necessary to perform calculations.. while keeping things as performant as possible.  This one was trickier than the issue suggested. My bad for not seeing that and voicing fair warning earlier

vjeffrey:
👍  thanks for the update/response!

so #2 on your list will allow us to filter by control status? is that correct?
and what does "control tags" mean..?  do you mean filtering the controls list by control tags? or something else?

rickmarry:
> so #2 on your list will allow us to filter by control status? is that correct?

yes

rickmarry:
> what does "control tags" mean..? do you mean filtering the controls list by control tags? or something else?
from the description in issue #521, I interpret the ask as simply including in the response, any tags associated with respective control. 
do we also need to filter on tags too?

rickmarry:
btw.. this thing is coming out great.. just saying

rickmarry:
This feature is going to be great and we'll have it soon.  I promise..  nanometers away from done

vjeffrey:
> do we also need to filter on tags?

the api should respect all filters, and we have filter by control tag, so yes. that's actually one of the most important bits, because we need to be able to widdle this list down by the control tag filter applied so we can see all controls that include that tag

rickmarry:
> the api should respect all filters

ok

components/automate-gateway/api/compliance/reporting/reporting.proto
```diff
@@ -146,6 +162,31 @@ service ReportingService {
...
+	string title=2;
+	ProfileMin profile = 3;
+	float impact = 4;
+	google.protobuf.Timestamp end_time = 5;
+	ControlSummary control_summary = 6;
+	repeated StringTag string_tags = 7;
+}
+
+message ControlItems {
+	repeated ControlItem control_item = 1;
```
vjeffrey: could we maybe call this controls?  as in `repeated ControlItem controls = 1;` ?
or discuss with ux?  i feel like it sounds weird to have a singular "item" for something that is going to be an array.
```
{
  "control_item": [
    {
      "id": "tmp-3.0",
      "title": "Create /tmp directory",
      "profile": {
        "name": "",
        "title": "InSpec Tag Test Profile",
        "id": "1e93c77aca38eeffa495637011e74bf615628cf6c46fad05e067bbc16b895c6a",
        "version": "0.2.0",
        "status": ""
      },
...
```

components/compliance-service/api/tests/010_controls_list_spec.rb
```diff
@@ -0,0 +1,373 @@
...
+
+describe File.basename(__FILE__) do
+  Reporting = Chef::Automate::Domain::Compliance::Api::Reporting unless defined?(Reporting)
+
+  def reporting
+    Reporting::ReportingService;
+  end
+
+  # it "no filters gets them all filter" do
+  #   assert_grpc_error("rpc error: code = InvalidArgument desc = 'control_tag' filter is required for 'control_tag_value' suggestions", 2) do
```
vjeffrey: assuming you didn't intend to include the commented block here?

vjeffrey:
endpoint is working well from my testing: https://github.com/chef/automate/issues/1339#issuecomment-533926496 
⭐️ 

haven't really reviewed most of the code yet tho. hoping @alexpop can do some review on this in the morning.

components/compliance-service/api/tests/010_controls_list_spec.rb
```diff
@@ -0,0 +1,373 @@
...
+            }
+        ]
+    }
+
+    control_item_array = actual_data['control_item']
+    assert_equal(2, control_item_array.size)
+    assert_equal_json_content(expected_data, actual_data)
+  end
+
+  it "control list items with a size of 2 but only care about failed" do
```
alexpop: Needs change to reflect the test

alexpop:
I'm going to push a few more integration tests in.

components/automate-gateway/api/compliance/reporting/reporting.proto
```diff
@@ -146,6 +162,31 @@ service ReportingService {
...
+}
+
+message ControlItem {
+	string id =1;
+	string title=2;
+	ProfileMin profile = 3;
+	float impact = 4;
+	google.protobuf.Timestamp end_time = 5;
+	ControlSummary control_summary = 6;
+	repeated StringTag string_tags = 7;
```
alexpop: Are sting_tags required for this API? I'm asking because this is the internal representation that we have of the control tags that we support for filtering(`null` value or single string or array of strings value). Any other values will not be stored as string_tags in elastic search.

components/automate-gateway/api/compliance/reporting/reporting.proto
rickmarry: I don't really view these as simply controls.. they are more than that. Maybe ControlStats?
But also, they are also bit specific.  For the given filter and the latest run in our timeseries data:
1.  They provide stats.
2.  They provde the tags specific.

With them being so specific, I guess I just figured that calling them "control" was not enough.

components/compliance-service/api/tests/010_controls_list_spec.rb
rickmarry: good catch.. i'll clean that up

components/automate-gateway/api/compliance/reporting/reporting.proto
vjeffrey: let's ask ux how they feel?  i don't necessarily care about the naming, it's the lack of pluralization that gets me

vjeffrey:
we talked about two items in standup:
- remove the string tags proto bits, we will address this in the future (open an issue)
- open an issue to enumerate additional tests to add if needed

vjeffrey:
i'm re-testing locally with latest changes

PR #1605, opened 2019-09-13
[AUTH-1501] Updated project details page to display pending edits message when rule deletes are staged

tylercloke:

Also added a message for the empty rules UI in the case that there are deletes pending (needs Susan's review)

1)Create two rules in a new projects
2)Apply both rules
3) Go to project and delete one rule, see `Edits are pending...`
4) Delete second rule, see empty UI case and see `Deleted rules are pending...`

- [x] Tests added/updated?
- [ ] Docs added/updated?

@susanev Added a case into the no rules UI. If there are no rules but there are staged deleted rules, it will have the extra message below:

Screen Shot 2019-09-13 at 10 53 00 AM

### DIFF
--- a/components/automate-ui/src/app/entities/projects/project.reducer.ts
+++ b/components/automate-ui/src/app/entities/projects/project.reducer.ts
@@ -79,7 +79,7 @@ export function projectEntityReducer(
 
     case ProjectActionTypes.GET_SUCCESS:
       return set(GET_STATUS, EntityStatus.loadingSuccess,
-        projectEntityAdapter.addOne(action.payload.project, state));
+        projectEntityAdapter.upsertOne(action.payload.project, state));
 
     case ProjectActionTypes.GET_FAILURE:
       return set(GET_STATUS, EntityStatus.loadingFailure, state);
--- a/components/automate-ui/src/app/pages/project/details/project-details.component.html
+++ b/components/automate-ui/src/app/pages/project/details/project-details.component.html
@@ -61,7 +61,7 @@
           <app-authorized [allOf]="['/iam/v2beta/projects', 'post']">
             <chef-button [routerLink]="['/settings', 'projects', project?.id, 'rules']" id="create-button" primary>Create Rule</chef-button>
           </app-authorized>
-          <small *ngIf="project.status === 'EDITS_PENDING'">Edits are pending: update <a [routerLink]="['/settings/projects']">projects</a> to apply edits.</small>
+          <small *ngIf="project?.status === 'EDITS_PENDING'">Edits are pending: update <a [routerLink]="['/settings/projects']">projects</a> to apply edits.</small>
         </chef-toolbar>
         <app-authorized [allOf]="['/iam/v2beta/projects', 'get']">
           <chef-table id="project-rules-table">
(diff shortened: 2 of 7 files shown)


components/automate-ui/src/app/entities/projects/project.reducer.ts
```diff
@@ -79,7 +79,7 @@ export function projectEntityReducer(
 
     case ProjectActionTypes.GET_SUCCESS:
       return set(GET_STATUS, EntityStatus.loadingSuccess,
-        projectEntityAdapter.addOne(action.payload.project, state));
+        projectEntityAdapter.upsertOne(action.payload.project, state));
```
tylercloke: If you use addOne, it doesn't create a new object and thus doesn't trigger ngrx store change detection. That means that any select that is focused on this object (like `projectFromRoute`) won't actually get updates from the store. TL;DR I think we should be using `upsertOne` everywhere we are using `addOne` (not just on create like my tech debt card proposes https://github.com/chef/automate/issues/1199).

components/automate-ui/src/app/pages/project/details/project-details.component.html
```diff
@@ -99,10 +99,15 @@
       <ng-container *ngIf="showFirstRuleMessage()">
         <app-authorized [allOf]="['/iam/v2beta/projects', 'post']">
           <div class="empty-case-container">
-            <p>Create the first ingest rule to get started!</p>
-          </div>
-          <div class="empty-case-container">
+            <div class="empty-case-entry">
+              <small *ngIf="project.status === 'EDITS_PENDING'">Deleted rules are pending: update <a [routerLink]="['/settings/projects']">projects</a> to apply edits.</small>
```
susanev: @tylercloke whoa good catch! could you just make this the same message from the top of the table though?

components/automate-ui/src/app/pages/project/details/project-details.component.ts
```diff
@@ -98,8 +103,19 @@ export class ProjectDetailsComponent implements OnInit, OnDestroy {
...
+
+    // if a rule gets deleted, we need to refresh the project status
+    this.store.select(deleteRuleStatus).pipe(
+      filter(() => this.id !== undefined),
+      takeUntil(this.isDestroyed))
+      .subscribe((status) => {
+        if (status === EntityStatus.loadingSuccess) {
+          this.store.dispatch(new GetProject({ id: this.id }));
+        }
+      });
```
msorens: Suggest let the filter do the work:
```
    this.store.select(deleteRuleStatus).pipe(
      filter(status => this.id !== undefined && status === EntityStatus.loadingSuccess),
      takeUntil(this.isDestroyed))
      .subscribe(() => {
        this.store.dispatch(new GetProject({ id: this.id }));
      });
```

components/automate-ui/src/app/pages/project/details/project-details.component.html
```diff
@@ -99,10 +99,15 @@
       <ng-container *ngIf="showFirstRuleMessage()">
```
msorens: Just a thought: might this be more clear?
```suggestion
      <ng-container *ngIf="showNoRulesMessage()">
```

components/automate-ui/src/app/entities/projects/project.reducer.ts
bcmdarroch: good catch! we should probably update that tech debt card

PR #1609, opened 2019-09-13
added note about being on the latest automate version before upgrading iam

susanev:

Recently, in internal slack we learned about a customer who was on an older version of Automate when they upgraded to iam v2 this caused some bugs we had fixed in later versions. This pr adds a note to the iam v2 docs encouraging users to be on the latest Chef Automate version before upgrading iam.

- [ ] Tests added/updated?
- [x] Docs added/updated?

docs-upgrade-automate-info


susanev:
@mjingle since we are in beta we are pretty much-fixing bugs almost every release... they should really be on latest.

Issue #1469, opened 2019-09-05
Buggy behavior with the Compliance tab calendar popup for the date filter


jonong1972:
On the compliance tab if you attempt to change the date using the Calendar popup it takes you to the wrong date when clicking on one of the grey'd out days for the previous or next month.  What appears to be happening is that it's trying to change to the selected day in the current month instead of the previous/next months day.  If you happen to click on a day that doesn't exist for the current month then it jumps to a day in the next month. 

For example, if you are looking at February and click the January 31st grey'd out date it will jump the calendar to March 3rd. See the screen recording for a demo

## Repro Steps
1. Go to Compliance tab
2. Expand the Report Metadata
2. Click on calendar within Filter bar
3. Select "September 5, 2019"
4. Notice the "Sept 5, 2019" report date in the metadata section.
5. Now click on on calendar again
6. Click on the greyed out "3" (oct3)
7. Notice the metadata section. I returns "September 3, 2019"

## Definition of Done
Expected report date should return October 3, 2019

This happens  also when there are greyed dates at the beginning of the month.

PR #1610, opened 2019-09-13
Fix selecting a date outside current month

lancewf:
This change fixes the problem on the compliance page if you attempt to change the date using the Calendar popup it takes you to the wrong date when clicking on one of the grey'd out days for the previous or next month. 

#1469
### :+1: Definition of Done
The calendar selects the correct date when days for the previous and next month are selected. 

1. Go to https://a2-dev.test/compliance/reports/overview and select dates in the search bar. 

- [ ] Tests added/updated?
- [ ] Docs added/updated?

Before
before2

After
As you can see below what was also add is the highting and curser change. 
after2


components/automate-ui/src/app/components/calendar/calendar.component.ts
```diff
@@ -25,7 +25,7 @@ export class CalendarComponent {
```
lancewf: Storing a moment for each day on the calendar. When clicked the year, month, and day are returned as '2019-10-23'.

PR #1611, opened 2019-09-14
[Automate-793] Add caching to introspect calls

msorens:

There are a large number of introspection calls on each page and this continues to grow. That incurs a small--but not negligible--performance penalty. Also, it makes viewing a network trace very cumbersome as there is a lot of "noise" added to the list of network calls. This PR adds some simple caching with time-based degradation to eliminate the vast majority of network traffic but still keep the introspection data in the browser fresh.

N/A

### :+1: Definition of Done
There should be no discernible difference in the exposed UI. However, if you open the network tab of dev tools you can see the effect. In this screen shot I opened Automate on the Dashboards tab. On the very first load, there will be a handful of network calls. (Note that I have filtered the network list to just show introspection calls.)

I then selected one call in the network tab--the arrow is pointing this out. Within a 30-second period, I then navigated around to different tabs across the top and different entries in the sidebar--_all with no introspection traffic at all!_ 

This PR causes all the `IntrospectSome` calls to be cached. Notice, though that there are a couple calls beyond the entry that the arrow points to. Those are parameterized `IntrospectOne` calls needed on the projects page, and those are **not** cached.

Once you go beyond the 30-second cache expiry time, the next time you go to a different tab or page, a single `IntrospectAll` will occur. (It's possible that a couple `IntrospectSome`s may sneak in there, too.)

image

With the latest commit I added even further noise reduction. The "After" scenario shows the introspection calls from me navigating around Automate for more than 2 minutes. The "Before" scenario is many, many screenfuls long!

image

There is some noise reduction in the Redux Dev Tools pane also, but not quite as much: you will still see the introspection _requests_ but, when cached, no introspection _result_--so it cuts introspection entries down by about half.

Rebuild automate-ui

### DIFF
--- a/components/automate-ui/src/app/entities/userperms/userperms.effects.ts
+++ b/components/automate-ui/src/app/entities/userperms/userperms.effects.ts
@@ -3,7 +3,7 @@ import { HttpErrorResponse } from '@angular/common/http';
 import { Store } from '@ngrx/store';
 import { Actions, Effect, ofType } from '@ngrx/effects';
 import {
-  debounceTime,
+  throttleTime,
   mergeMap,
   map,
   catchError,
@@ -38,12 +38,11 @@ export class UserPermEffects {
     private store: Store<NgrxStateAtom>
   ) { }
 
+  private freshLimitInMilliseconds = 30000;
   @Effect()
   fetchAllPerms$ = this.actions$.pipe(
     ofType(UserPermsTypes.GET_ALL),
-    debounceTime(20), // each request of this type is a true duplicate so debounce them
-    withLatestFrom(this.store.select(getlastFetchTime)),
-    filter(([_action, lastTime]) => this.stale(lastTime)),
+    throttleTime(this.freshLimitInMilliseconds),
     switchMap(() => {
       return this.requests.fetchAll().pipe(
         map((resp: UserPermsResponsePayload) => new GetAllUserPermsSuccess(resp.endpoints)),
--- a/components/automate-ui/src/app/entities/userperms/userperms.effects.ts
+++ b/components/automate-ui/src/app/entities/userperms/userperms.effects.ts
@@ -1,9 +1,20 @@
-import { debounceTime, mergeMap, map, catchError } from 'rxjs/operators';
 import { Injectable } from '@angular/core';
 import { HttpErrorResponse } from '@angular/common/http';
+import { Store } from '@ngrx/store';
 import { Actions, Effect, ofType } from '@ngrx/effects';
+import {
+  debounceTime,
+  mergeMap,
+  map,
+  catchError,
+  withLatestFrom,
+  switchMap,
+  filter
+} from 'rxjs/operators';
 import { of } from 'rxjs';
+import * as moment from 'moment';
 
+import { NgrxStateAtom } from 'app/ngrx.reducers';
 import {
   UserPermsTypes,
   GetAllUserPermsSuccess,
@@ -17,27 +28,35 @@ import {
   GetUserParamPermsFailure
 } from './userperms.actions';
 import { UserPermsRequests } from './userperms.requests';
+import { getlastFetchTime } from './userperms.selectors';
 
 @Injectable()
 export class UserPermEffects {
   constructor(
     private actions$: Actions,
-    private requests: UserPermsRequests
+    private requests: UserPermsRequests,
+    private store: Store<NgrxStateAtom>
   ) { }
 
-  // include debounce here because each request of this type is a true duplicate
   @Effect()
   fetchAllPerms$ = this.actions$.pipe(
     ofType(UserPermsTypes.GET_ALL),
-    debounceTime(20),
-    mergeMap(() => this.requests.fetchAll().pipe(
-      map((resp: UserPermsResponsePayload) => new GetAllUserPermsSuccess(resp.endpoints)),
-      catchError((error: HttpErrorResponse) => of(new GetAllUserPermsFailure(error))))));
+    debounceTime(20), // each request of this type is a true duplicate so debounce them
+    withLatestFrom(this.store.select(getlastFetchTime)),
+    filter(([_action, lastTime]) => this.stale(lastTime)),
+    switchMap(() => {
+      return this.requests.fetchAll().pipe(
+        map((resp: UserPermsResponsePayload) => new GetAllUserPermsSuccess(resp.endpoints)),
+        catchError((error: HttpErrorResponse) => of(new GetAllUserPermsFailure(error))));
+    }));
 
   @Effect()
   fetchSomePerms$ = this.actions$.pipe(
     ofType(UserPermsTypes.GET_SOME),
-    mergeMap((action: GetSomeUserPerms) => this.requests.fetchSome(action.payload).pipe(
+    withLatestFrom(this.store.select(getlastFetchTime)),
+    filter(([_action, lastTime]) => this.stale(lastTime)),
+    mergeMap(
+      ([action, _]: [GetSomeUserPerms, Date]) => this.requests.fetchSome(action.payload).pipe(
       map((resp: UserPermsResponsePayload) => new GetSomeUserPermsSuccess(resp.endpoints)),
       catchError((error: HttpErrorResponse) => of(new GetSomeUserPermsFailure(error))))));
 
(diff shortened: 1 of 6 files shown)


components/automate-ui/src/app/ui.component.ts
```diff
@@ -61,7 +62,14 @@ export class UIComponent implements OnInit {
...
         filter(event => event instanceof ActivationStart)
-    ).subscribe((event: any) => this.renderNavbar = !event.snapshot.data.hideNavBar);
+    ).subscribe((event: ActivationStart) => this.renderNavbar = !event.snapshot.data.hideNavBar);
+
+    this.router.events.pipe(
+        filter(event => event instanceof NavigationEnd)
+    ).subscribe(() => this.store.dispatch(new GetAllUserPerms()));
+
+    // Initial call
+    this.store.dispatch(new GetAllUserPerms());
```
msorens: ➡️ Start here:
(1) This loads all introspection data at time of browser refresh.

components/automate-ui/src/app/ui.component.ts
```diff
@@ -61,7 +62,14 @@ export class UIComponent implements OnInit {
   ngOnInit(): void {
     this.router.events.pipe(
         filter(event => event instanceof ActivationStart)
-    ).subscribe((event: any) => this.renderNavbar = !event.snapshot.data.hideNavBar);
+    ).subscribe((event: ActivationStart) => this.renderNavbar = !event.snapshot.data.hideNavBar);
+
+    this.router.events.pipe(
+        filter(event => event instanceof NavigationEnd)
+    ).subscribe(() => this.store.dispatch(new GetAllUserPerms()));
```
msorens: (2) Every time you navigate to a new Angular route (generally equivalent to when you go to a new view) it will re-fetch the introspection data _but only if it has gone stale_ (been around more than 30 seconds). If it is still fresh, no data is fetched.

There are a variety of lifecycle hooks in routing; this particular one (`NavigationEnd`) typically occurs about once per route (minimum 1, maximum 2, from what I have observed). If it is triggered more than once, that is OK--see point (3).

components/automate-ui/src/app/entities/userperms/userperms.effects.ts
```diff
@@ -17,27 +28,35 @@ import {
...
 
-  // include debounce here because each request of this type is a true duplicate
   @Effect()
   fetchAllPerms$ = this.actions$.pipe(
     ofType(UserPermsTypes.GET_ALL),
-    debounceTime(20),
-    mergeMap(() => this.requests.fetchAll().pipe(
-      map((resp: UserPermsResponsePayload) => new GetAllUserPermsSuccess(resp.endpoints)),
-      catchError((error: HttpErrorResponse) => of(new GetAllUserPermsFailure(error))))));
+    debounceTime(20), // each request of this type is a true duplicate so debounce them
```
msorens: (3) Even if the dispatch fires a few times rapidly, this debounce ensures it only sends one occurrence through.

components/automate-ui/src/app/entities/userperms/userperms.effects.ts
```diff
@@ -17,27 +28,35 @@ import {
...
+    withLatestFrom(this.store.select(getlastFetchTime)),
+    filter(([_action, lastTime]) => this.stale(lastTime)),
```
msorens: (4) The cache expiration check--stop if the data is not yet stale; otherwise re-fetch the data.

components/automate-ui/src/app/entities/userperms/userperms.effects.ts
```diff
@@ -17,27 +28,35 @@ import {
...
+      return this.requests.fetchAll().pipe(
+        map((resp: UserPermsResponsePayload) => new GetAllUserPermsSuccess(resp.endpoints)),
+        catchError((error: HttpErrorResponse) => of(new GetAllUserPermsFailure(error))));
+    }));
 
   @Effect()
   fetchSomePerms$ = this.actions$.pipe(
     ofType(UserPermsTypes.GET_SOME),
-    mergeMap((action: GetSomeUserPerms) => this.requests.fetchSome(action.payload).pipe(
+    withLatestFrom(this.store.select(getlastFetchTime)),
```
msorens: (5) And finally, this line is what reduces network noise. `<app-authorized>` triggers this effect (mostly). Since all `IntrospectSome` calls are guaranteed to be included in the result set of `IntrospectAll`, we can use the same staleness check!

components/automate-ui/src/app/entities/userperms/userperms.reducer.ts
```diff
@@ -49,9 +51,12 @@ export function permEntityReducer(
       // toPairs converts each endpoint in the payload from
       // { <path>: <perms_map> } to [<path>, <perms_map> ].
       const allPairs = toPairs(action.payload);
-      const allPerms = permIndexer(allPairs);
+      // combine with what we already have so as not to erase parameterized endpoints
+      // that are not included here.
+      const allPerms = defaults(state.byId, permIndexer(allPairs));
```
msorens: (6) This is a little nuance that might have caused some intermittent and infrequent errors--just the type that one never wants to see! When `IntrospectAll` was first implemented, "All" meant "All" so it was OK to completely replace the existing slice of the store with this new one. But then along came parameterized introspection, which is not included in that result set. This patch changes `IntrospectAll` so it only updates rather than replacing the slice of store. That way, if there were any parameterized introspection calls, their results would not be eliminated!

components/automate-ui/src/app/entities/userperms/userperms.effects.ts
```diff
@@ -1,9 +1,20 @@
-import { debounceTime, mergeMap, map, catchError } from 'rxjs/operators';
```
msorens: Dear Code Reviewer: Please start at the very **bottom**--with point (1)--and follow the numbers to easily make sense of the changes herein. 🎃 🐰

components/automate-ui/src/app/entities/userperms/userperms.effects.ts
srenatus: `debounceTime` in effect causes a 20ms delay here -- since the first request `a` is kept in waiting for 20ms, before it's fired (to let any subsequent requests `b`, `c`, ... within those 20ms be ignored); and each subsequent requests *resets the 20ms*. This doesn't seem optimal for our use case here.

See https://rxjs-dev.firebaseapp.com/api/operators/debounceTime#description:

img

Do you know an alternative that fires `a` right away but ignores subsequent requests in a 20ms window?

msorens: I originally put that in knowing it fires at the end of the interval, as you correctly state, as I felt it was short enough to be insignificant. However, since you raised the question I did come up with an improvement: https://rxjs-dev.firebaseapp.com/api/operators/throttleTime

srenatus: Cool, thanks. If we're using `throttleTime`, I suppose we could up the `20ms` _a lot_ and save some more requests...? 😃

components/automate-ui/src/app/entities/userperms/userperms.effects.ts
```diff
@@ -17,27 +28,41 @@ import {
...
+  private freshLimitPlusANudge = 33000;
+
   @Effect()
   fetchAllPerms$ = this.actions$.pipe(
     ofType(UserPermsTypes.GET_ALL),
-    debounceTime(20),
-    mergeMap(() => this.requests.fetchAll().pipe(
-      map((resp: UserPermsResponsePayload) => new GetAllUserPermsSuccess(resp.endpoints)),
-      catchError((error: HttpErrorResponse) => of(new GetAllUserPermsFailure(error))))));
+    throttleTime(this.freshLimitInMilliseconds),
```
msorens: (3) The throttling here serves as the cache expiry control for IntrospectAll itself.

components/automate-ui/src/app/entities/userperms/userperms.effects.ts
msorens: Already done, and turned it even further to an advantage using the throttle itself as the cache expiry for `IntrospectAll` itself.

components/automate-ui/src/app/ui.component.ts
bcmdarroch: @tarablack01 or @scottopherson should correct me if I'm wrong, but I'd heard one of ui-team's tech debt goals is to remove ui.component.ts (in favor of `app.component.ts`). So might be a good idea to put this in `app.component` now to save some work down the line

msorens: I checked with them; does not seem to be on the horizon at the moment. But thanks for raising the question.

Issue #285, opened 2019-05-08
connectivity error for node is hard to read when long


vjeffrey:
# User Story

As shown in this screenshot, if an error is long, we do not do any word wrapping, leaving the user with the burden of sliding the bar all the way across to read. This is not very convenient, let's see if we can make this better.

<img width="1278" alt="Screen Shot 2019-05-08 at 12 07 28" src="https://user-images.githubusercontent.com/10341541/57397484-06802e00-718a-11e9-8e66-a4a435f8c0a6.png">

We're going to need some UX opinions here, @jonong1972 

# Definition of Done
Node connectivity error is easier to read

jonong1972:
It looks that currently we are utilizing the side drawer only for the Error message on this Nodes view. Since that is the only thing we are doing on this page, we may be able to just word wrap the whole message, and we can over the whole side drawer with the message.

Looking at the code, for this particular context, we would have to put a CSS style on the pre tag.

white-space: pre-wrap;

Which would then show the error like this:

![Screen Shot 2019-07-29 at 2.16.41 PM.png](https://images.zenhubusercontent.com/5c58b40f704a793fa6c8d7b8/aba9ce28-4f7b-404f-806d-f707f571d1b3)

PR #1615, opened 2019-09-16
word wrap the connectivity error message for node

vinay033:

It looks that currently, we are utilizing the side drawer only for the Error message on this Nodes view. Since that is the only thing we are doing on this page, we may be able to just word wrap the whole message, and we can cover the whole side drawer with the message.
I have added CSS changes for word warp the error message. 

### Repro Steps

1. `rebuild components/automate-ui-devproxy`
1. inside studio `CHEF_USERNAME=username scripts/get_secrets.sh` if you haven't in a while
1. outside of studio `source dev/secrets-env.sh`
1. outside of studio `A2_URL='https://a2-dev.test' A2_TOKEN='token_val' components/compliance-service/scripts/create-pg-data.sh`

https://github.com/chef/automate/issues/285

### :+1: Definition of Done
Node connectivity error is easier to read

### :camera: Screenshots
Screenshot from 2019-09-16 17-27-00

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/+scanner/containers/nodes-list/nodes-list.component.scss
+++ b/components/automate-ui/src/app/pages/+compliance/+scanner/containers/nodes-list/nodes-list.component.scss
@@ -144,6 +144,7 @@
       border-radius: $global-radius;
       background: $chef-lightest-grey;
       font-size: 14px;
+      white-space: pre-wrap;
     }
   }
 }


susanev:
@jonong1972 @vinay033 not sure if we care but im still getting things like...

scan-job-errors

jonong1972:
@susanev Yeah, that is a bigger problem of how we use the side panel. Right now, I would say leave that as is for now, because the across the site, we are treating the PRE tag differently in some cases, and other places using the error remediation/handling slightly different. Or at least the use case I found while trying to duplicate your error.

susanev:
ah i see this issue was specifically about a code block, thx for clarifying.

PR #1616, opened 2019-09-16
lib/cereal: Add StartGuard to replace WaitGroup-based start checks

stevendanna:
We were using WaitGroup to help ensure that we weren't calling Start()
wasn't called twice on various structs that manage long-lived go
routines.

This replaces those uses with a new type, StartGuard, that allows us
to customize the message that users (developers) will get when they
make this mistake.

We could have done this with recover, but this felt a big better.


stevendanna:
@danielsdeleo Hit this during development and recommended we improve the error message a bit since the default wait group panic message doesn't convey much information.

PR #1617, opened 2019-09-16
Use 1 << x rather than math.Pow(2.0, float64(x))

stevendanna:
I noticed this when reading some code. Just a minor style thing.


stevendanna:
Hrm. This build failure implies we are still building on Go 1.12

stevendanna:
I've added the casts back until I can sort out all the places we might be using 1.12 still.  Doesn't seem worth it to fix now.

PR #1628, opened 2019-09-16
automate-cs: reduce number of sql connections

stevendanna:
These are in line with what we are using in other chef-server
deployments.

### DIFF
--- a/api/config/bifrost/config_request.go
+++ b/api/config/bifrost/config_request.go
@@ -32,7 +32,7 @@ func DefaultConfigRequest() *ConfigRequest {
 	c.V1.Sys.Log.MaxErrorLogsPerSecond = w.Int32(1000)
 
 	c.V1.Sys.Sql.Timeout = w.Int32(5000)
-	c.V1.Sys.Sql.PoolInitSize = w.Int32(20)
+	c.V1.Sys.Sql.PoolInitSize = w.Int32(10)
 	c.V1.Sys.Sql.PoolMaxSize = w.Int32(20)
 	c.V1.Sys.Sql.PoolQueueMax = w.Int32(50)
 	c.V1.Sys.Sql.PoolQueueTimeout = w.Int32(2000)
--- a/api/config/erchef/config_request.go
+++ b/api/config/erchef/config_request.go
@@ -71,9 +71,9 @@ func DefaultConfigRequest() *ConfigRequest {
 	c.V1.Sys.Authz.PoolQueueTimeout = w.Int32(2000)
 
 	c.V1.Sys.Sql.Timeout = w.Int32(5000)
-	c.V1.Sys.Sql.PoolInitSize = w.Int32(20)
-	c.V1.Sys.Sql.PoolMaxSize = w.Int32(50)
-	c.V1.Sys.Sql.PoolQueueMax = w.Int32(20)
+	c.V1.Sys.Sql.PoolInitSize = w.Int32(10)
+	c.V1.Sys.Sql.PoolMaxSize = w.Int32(20)
+	c.V1.Sys.Sql.PoolQueueMax = w.Int32(50)
 	c.V1.Sys.Sql.PoolQueueTimeout = w.Int32(2000)
 
 	c.V1.Sys.DataCollector.Timeout = w.Int32(5000)
--- a/components/automate-workflow-server/habitat/default.toml
+++ b/components/automate-workflow-server/habitat/default.toml
@@ -8,8 +8,8 @@
 api_port = 9611
 audit_max_events = 100
 ca_cert_chain_depth = 2
-db_pool_init_count = 20
-db_pool_max_count = 100
+db_pool_init_count = 10
+db_pool_max_count = 20
 default_search="""(recipes:delivery_builder OR recipes:delivery_builder\\\\:\\\\:default
                   OR recipes:delivery_build OR recipes:delivery_build\\\\:\\\\:default
                   OR tags:delivery-build-node)"""
--- a/api/config/bookshelf/config_request.go
+++ b/api/config/bookshelf/config_request.go
@@ -33,7 +33,7 @@ func DefaultConfigRequest() *ConfigRequest {
 	c.V1.Sys.Log.MaxErrorLogsPerSecond = w.Int32(1000)
 
 	c.V1.Sys.Sql.Timeout = w.Int32(5000)
-	c.V1.Sys.Sql.PoolInitSize = w.Int32(20)
+	c.V1.Sys.Sql.PoolInitSize = w.Int32(10)
 	c.V1.Sys.Sql.PoolMaxSize = w.Int32(20)
 	c.V1.Sys.Sql.PoolQueueMax = w.Int32(50)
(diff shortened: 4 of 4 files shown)


api/config/bifrost/config_request.go
```diff
@@ -32,7 +32,7 @@ func DefaultConfigRequest() *ConfigRequest {
 	c.V1.Sys.Log.MaxErrorLogsPerSecond = w.Int32(1000)
 
 	c.V1.Sys.Sql.Timeout = w.Int32(5000)
-	c.V1.Sys.Sql.PoolInitSize = w.Int32(20)
+	c.V1.Sys.Sql.PoolInitSize = w.Int32(10)
```
stevendanna: We may want to leave InitSize == MaxSize. We've definitely hit issues in the past with the pool scaling behavior.

api/config/erchef/config_request.go
```diff
@@ -53,8 +53,8 @@ func DefaultConfigRequest() *ConfigRequest {
 	c.V1.Sys.Keygen.Timeout = w.Int32(20000)
 
 	c.V1.Sys.Index.Timeout = w.Int32(30000)
-	c.V1.Sys.Index.PoolInitSize = w.Int32(25)
-	c.V1.Sys.Index.PoolMaxSize = w.Int32(100)
```
stevendanna: 100 is WAY too high as our max_conns is 100.

components/automate-workflow-server/habitat/default.toml
```diff
@@ -8,8 +8,8 @@
 api_port = 9611
 audit_max_events = 100
 ca_cert_chain_depth = 2
-db_pool_init_count = 20
-db_pool_max_count = 100
+db_pool_init_count = 10
+db_pool_max_count = 20
```
stevendanna: I guess we never made these ones actually configurable, which is fine because no on uses this component right now.

components/automate-workflow-server/habitat/default.toml
```diff
@@ -8,8 +8,8 @@
 api_port = 9611
 audit_max_events = 100
 ca_cert_chain_depth = 2
-db_pool_init_count = 20
-db_pool_max_count = 100
```
irvingpop: do we know why max was so high here? was there no queue?

stevendanna: I think we've been copy/pasta'ing the defaults rather than thinking about it.

https://github.com/chef/chef-server/blob/master/src/oc_erchef/habitat/default.toml#L61

api/config/bifrost/config_request.go
irvingpop: hey @stevendanna do you have any pointers to past discussions or test results that discuss these issues?

irvingpop:
what would be the downsides to the average customer if set init = 1 on all the services and made init/max/queue configurable?

api/config/bifrost/config_request.go
stevendanna: Looking for them now.  I'm suspicious that these issues were never fully investigated but nonetheless became part of the "common wisdom."

Because the queue immediately rejects connections when it is out of connections and _then_ creates a new connection, I think the init=max behavior provided more comprehensible semantics before we had the request queue.

stevendanna:
> what would be the downsides to the average customer if set init = 1 on all the services and made init/max/queue configurable?

init/max/queue should all be configurable for the Chef Server services.  If we set init = 1, I think the main consequence is that on a heavily loaded server, the user might see more failures _from erchef_, as the pool spins up. But, I am not clear if this would actually result in them seeing more errors since if we bump the init size all that means is that they will be getting errors from the load balancer until we are started instead of from erchef.

I wonder if @markan has any opinions here.

irvingpop:
@stevendanna I just want to make sure we're optimizing for the most common case, which *I think* here may be combo servers like OpsWorks or "mixed use" Automate clusters.  

If the biggest concern is dealing with chef-client stampedes (outside of hosted Chef) I'm inclined to even set this too low and gather data about the failure modes before determining the right default value

api/config/bifrost/config_request.go
irvingpop: one more thing I thought of -  we should also adjust erchef's `c.V1.Sys.Authz.Pool*` down to match a bit closer to this.

stevendanna: I looked a bit at this in the morning and really, _all_ of the queues can probably use a second look. I feel like we don't *actually* want to start 100 ES connections, for example.

markan:
A few drive-by thoughts. 
1) I don't feel like we have particularly great instrumentation around this stuff; do we have a clean way of gathering data about how many errors we generate due to connection pool shortage? IIRC, we mostly throw a somewhat opaque erlang stack trace when this happens. That inability to collect data makes me lean towards a precautionary larger number for init.
2) The init size choice stems from our historical understanding of the burstyness of CCRs. People love their top of the hour runs, and pre-allocating the connections prevented a burst of failures. However I'm not sure we have good data outside hosted chef for what tuning choices make sense.
3) My intuition is that making the init size > 1 keeps us from seeing a lot of noise in the logs as bursts fail on their first connect, then succeed on a retry after the somewhat slow process of initing the connections warms the pool. This is built on *very* old data (from the first days of erchef) around how long new connections take to create; there we saw that the queue was useful if we were just waiting for an existing connection to be free, but not as much if we were waiting for a new connection. If we can spin up a new connection fast enough so that we're not timing out sitting in the queue, that would change my opinion.

irvingpop:
confession, I'm now getting cold feet about changing these values without making then configurable.  thinking about about what would happen if someone starts hitting limits and we can't help them.

stevendanna:
@irvingpop They are configurable for everything but workflow and no one uses workflow.

irvingpop:
> @irvingpop They are configurable for everything but workflow and no one uses workflow.

ah Okay @stevendanna I wasn't aware of that.  If they are runtime configurable then I'm 👍.  All I ask is that we document the tunables when we decide to release docs on automate-cs

stevendanna:
@irvingpop @markan Thanks for the important context you've shared.

I'm merging this for now with (init=10, max=20, queue=50), but I think we need to find a way to make more reasonable decisions on these defaults.  Perhaps this should be added to whatever list we have for finalizing the chef-server work.

I have a few concerns:

1) We have a belief that the connections are slow to start up.  They _ought_ to be slower than sending a request to an existing connection, but my expectation is that should still be on the order of 10s of ms on any production system.  I'm wondering if some of this intuition stems from before we did on-demand preparations of queries: 

https://github.com/chef/sqerl/commit/bce94d76357395986e3d96b0815424b514600bc0

Before that work, starting a connection would have included the time to prepare the dozes of queries we use.

2) We have "spooky ghost stories" from Ops about strange behavior when growing and shrinking the queue.  We should probably either (1) investigate them or (2) (if we don't have the time/resources) consider ripping out some of the complexity and sticking with a statically sized pool with a queue in front of it.

3) We have a lot of other pool configurables that aren't touched here that don't make a lot of sense to me and I assume most are the results of us never applying sqerl's queueing ability to various HTTP connections.

Anyway, these are probably best left for follow up work.

PR #1629, opened 2019-09-16
[EAS-1266] Deep filter suggestions for applications search bar

kmacgugan:
Filter autocomplete suggestions on the search bar by any existing filters.
This will miminize searches resulting in nothing.
Status filter is ignored because we want to allow for a more interactive experience
wherein a user could switch between statuses without having autocomplete options trimmed.

The UI changes to support this are already in master, so the filters will be passed. 
-Bring up the automate UI in your dev environment. 
-Rebuild the applications-service
-Populate some hab event data with the command `applications_populate_database`
-Go to the applications tab
-Add a filter and then once one filter is added see that the suggestions are filtered by the existing one. An easy way to do this is using the same filter twice and see that you just get the one suggestion. Try out wildcards too.

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/applications-service/pkg/storage/postgres/services.go
```diff
@@ -242,15 +248,18 @@ func (db *Postgres) GetServicesDistinctValues(fieldName, queryFragment string) (
 	}
 
 	columnName := columnNameForField(fieldName)
-	query := fmt.Sprintf("SELECT DISTINCT %s from service_full AS t WHERE t.%s ILIKE $1 ORDER BY %s ASC LIMIT 100;",
+	queryFirst := fmt.Sprintf("SELECT DISTINCT %s from service_full AS t WHERE t.%s ILIKE $1 ",
 		columnName,
```
stevendanna: I know we pre-validate the values of fieldName elsewhere in the code, but might it be worth also quoting some of these identifiers as a second line of defense? We can quote them with a function like: https://github.com/chef/automate/blob/a4ba9f1f2f9b860fefc40c1b85e05849f61b52bd/lib/platform/pg/queries.go#L117-L129 that might be appropriate to move somewhere shared. #

components/applications-service/pkg/storage/postgres/services.go
```diff
@@ -242,15 +248,18 @@ func (db *Postgres) GetServicesDistinctValues(fieldName, queryFragment string) (
...
 		columnName,
+	)
+	querySecond := fmt.Sprintf(" ORDER BY %s ASC LIMIT 100;",
 		columnName,
 	)
+	query := queryFirst + whereConstraints + querySecond
 	matcher := fmt.Sprintf("%s%%", queryFragment)
```
afiune: Do we need to do any verification of this field? or it doesn't matter what the front end sends to the backend?

afiune: Do we need to do any verification of this field? or it doesn't matter what the front end sends to the backend?

components/applications-service/pkg/storage/postgres/services.go
afiune: Also, one trick you can do to avoid passing multiple times the same parameter and reuse:
```go
queryFirst := fmt.Sprintf("SELECT DISTINCT %[1]s from service_full AS t WHERE t.%[1]s ILIKE $1 ", columnName)
```

_Make sure to use the indicator `%[1]s`_

components/applications-service/pkg/storage/postgres/services.go
```diff
@@ -364,14 +373,16 @@ func convertComposedServiceToStorage(svc *composedService) *storage.Service {
 //      AND health = 'CRITICAL'     OR health = 'UNKNOWN'
 //      AND environment = 'production'
 //
-func buildWhereConstraintsFromFilters(filters map[string][]string) (string, error) {
+func buildWhereConstraintsFromFilters(filters map[string][]string, first bool) (string, error) {
```
afiune: My $0.02 here is to avoid modifying this function since it is well isolated and it has full coverage for the construction of `WHERE` statements, I have to viable suggestions on how you could approach this:

1) If you really want a function that appends constraints from an already generated `WHERE` statement, instead of passing a `bool` to the build function, I would suggest creating a new function that wraps this functions by accepting the filters and the previous `WHERE foo=bar` string. Example:
```go
myWhereStatement := "WHERE foo=bar"
err := appendWhereConstraintsFromFilters(filters, myWhereStatement)
if err != nil {
    return err
}
// Here myWhereStatement should have already both filters
```
2) Continue using this function in the same way but organizing the caller function in such way that you can append your extra WHERE statements to the already generated ones. Example: 
```go
func (db *Postgres) GetServicesDistinctValues(fieldName, queryFragment string, filters map[string][]string) ([]string, error) {
	var (
		columnName          = columnNameForField(fieldName)
		fieldNameConstraint = fmt.Sprintf("t.%s ILIKE $1", columnName)
	)

	// maybe create a function that checks any provided filedName?
	if !isValidFilter(fieldName) {
		return nil, errors.Errorf("field name %q is not valid for filtering, valid values are %v", fieldName, validFilterFields)
	}

	whereConstraints, err := buildWhereConstraintsFromFilters(filters)
	if err != nil {
		return nil, err
	}

	// Append to the generated (or not) our fieldName WHERE Constraints
	if len(whereConstraints) == 0 {
		whereConstraints = fmt.Sprintf("WHERE %s", fieldNameConstraint)
	} else {
		whereConstraints = fmt.Sprintf("%s AND %s", whereConstraints, fieldNameConstraint)
	}

	query := fmt.Sprintf("SELECT DISTINCT %[1]s from service_full AS t %[2]s ORDER BY %[1]s ASC LIMIT 100",
		columnNameForField(fieldName),
		whereConstraints,
	)

	matcher := fmt.Sprintf("%s%%", queryFragment)

	var matches []string
	_, err = db.DbMap.Select(&matches, query, matcher)
	if err != nil {
		return nil, err
	}
	return matches, nil
}
```

lancewf:
@kmacgugan is there a way I can test this locally?

components/applications-service/pkg/storage/postgres/services.go
kmacgugan: Good idea, I will add pgutils.EscapeLiteralForPG to the queryFragment.

kmacgugan:
@lancewf I added some details to the PR on how to test the change. Hit me up if you run into any trouble.

components/applications-service/pkg/storage/postgres/services.go
kmacgugan: Good suggestion! Thanks.

lancewf:
Should you also remove filters for the same type of filter the suggestions are for? This is what we are doing for the client runs and compliance suggestions. Where when you are providing suggestion for type environment all environment type filters should be removed. If these filters are not removed then a user can not add two filters for the same type. For example, say there are three environments 'demo', 'acc', and 'prod'.  A user wants to filter on 'demo' and 'acc'. Once they add the 'acc' filter there will not be a suggestion to add the 'demo' filter.

kmacgugan:
@lancewf That's a good point, and we want the apps search bar to behave in a similar manner. I will work on updating this to not filter on the same field.

components/applications-service/pkg/storage/postgres/services.go
```diff
@@ -177,7 +178,7 @@ var validFilterFields = []string{
 func (db *Postgres) GetServicesHealthCounts(filters map[string][]string) (*storage.HealthCounts, error) {
 	var (
 		sHealthCounts         storage.HealthCounts
-		WhereConstraints, err = buildWhereConstraintsFromFilters(filters)
+		WhereConstraints, err = buildWhereConstraintsFromFilters(filters, "WHERE")
```
lancewf: The 'W' should be lowercase for the variable `WhereConstraints`.

PR #1635, opened 2019-09-17
Use a cypress bind to make user mgmt test less flaky

tylercloke:

1) Attempting to use a cy bind to make the test less flaky. Theory is other instances of 'Create User' on the page might be causing flakiness. Also added a comment about paying attention to how you use data-cy in re-useable components. We should never be using the component in question in the same page twice so we should be fine here but a good thing to be aware of.

2) Also added a check to make sure both global and parameterized introspection calls return as part of login helper.

3) Marked the remainder of the ssh scan job tests as flaky since they've been flaking :(

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/page-components/user-table/user-table.component.html
+++ b/components/automate-ui/src/app/page-components/user-table/user-table.component.html
@@ -7,8 +7,6 @@
   <chef-toolbar>
     <app-authorized [allOf]="createPermissionsPath" [overrideVisible]="overridePermissionsCheck">
       <div *ngIf="showEmptyMessage || showTable" [ngClass]="showEmptyMessage ? 'empty-state' : ''">
-        <!-- TC: be careful when writing cypress tests that use data-cy binds in a re-useable component.
-        if it's used twice in the same page it can lead to unexpected behavior. -->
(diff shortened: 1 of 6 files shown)


components/automate-ui/src/app/page-components/user-table/user-table.component.html
```diff
@@ -7,7 +7,9 @@
   <chef-toolbar>
     <app-authorized [allOf]="createPermissionsPath" [overrideVisible]="overridePermissionsCheck">
       <div *ngIf="showEmptyMessage || showTable" [ngClass]="showEmptyMessage ? 'empty-state' : ''">
-        <chef-button primary [disabled]="!addButtonEnabled" (click)="addClicked.emit($event)"> {{ addButtonText }}
+        <!-- TC: be careful when writing cypress tests that use data-cy binds in a re-useable component.
```
blakestier: I wonder if there's a more global place to put this warning. 🤔

tylercloke: I'll add an entry to the readme.

tylercloke: Have to step out. If it doesn't happen here I'll make a follow up PR.

tylercloke: Done!

PR #1644, opened 2019-09-18
[Automate-1634] Re-enable "Update Projects" button upon cancel…

msorens:

The user needs to be able to restart a project update if (a) the prior run failed or (b) the prior run was cancelled by a user. As soon as an update starts the pending rules are moved from "staging" to "applied", which updates the "Ingest Rules" column on screen. That means there are no longer any staged rules, which was what previously dictated the enabled/disabled state of the <kbd>Update Projects</kbd> button. This PR adds cancel/fail awareness to that button enable/disable condition.

These two additional issues fell out of that seemingly innocuous change, and have been included in this PR:
- [x] Update the button label with the same cancel/fail awareness, so it changes from <kbd>Projects Up-to-Date</kbd> to <kbd>Update Projects</kbd> appropriately.
- [x] Adjust when the check for staged projects occurs so that the button label can change from <kbd>Updating Projects...</kbd> to <kbd>Projects Up-to-Date</kbd> upon successful completion of an update.

### :+1: Definition of Done
In the following [phenakistiscope](https://en.wikipedia.org/wiki/Phenakistiscope) pay close attention to the project update button near the top, as well as the value in the "Ingest Rules" and "Project Update Status" columns.

(1) Starting with a project having no rules...
image

(2) We add a rule so that the project now has edits pending. The button label changes and it becomes enabled:
image

(3) We select that button and things begin to happen!
image

(4) When the update finishes, things settle down again to this:
image

(5) Now let's add in a couple more projects in different initial states:
image

(6) And start an update. Notice that as soon as we do, p1 and p2 are now indistinguishable under "Ingest Rules" with "Applied".
image

(7) And now to throw a twist in, let's select "Stop Updating Projects":
image

(8) By cancelling, we put all project updates in question. But more than that, as we have discussed in team meetings, **Issue 1** is that we currently cannot even distinguish those projects that did not need updates at all (p1 in this case), so now both p1 and p2 look the same with "Needs updating"--but note that <kbd>Update Projects</kbd> is now re-enabled, which was where this PR started!
image

(9) A further side effect of cancelling the update is **Issue 2**: now when we start a fresh update, the "Project Update Status" column now immediately goes to "OK" rather than "Updating..." (because those projects did not start in an "edits pending" state).

Both issues 1 and 2 could be ameliorated with #1368 (Correct project update status for other users) unless we change the UX experience as some have suggested.

image

(10) Of course, once the update finishes, all is again well with the world:
image

Rebuild automate-ui
Go to Settings >> Projects
Create projects and rules as you will, or follow the phenakistiscope above.

### DIFF
--- a/components/automate-ui/src/app/pages/project/list/project-list.component.html
+++ b/components/automate-ui/src/app/pages/project/list/project-list.component.html
@@ -21,7 +21,7 @@
               data-cy="create-project" primary (click)="openCreateModal()">Create Project</chef-button>
           </app-authorized>
           <app-authorized [allOf]="['/iam/v2beta/apply-rules', 'post']">
-            <chef-button secondary [disabled]="!projectsHaveStagedChanges || (projects.applyRulesStatus$ | async)?.state === ApplyRulesStatusState.Running"
+            <chef-button secondary [disabled]="isDisabled()"
               id="update-start-button" (click)="openConfirmUpdateStartModal()">{{ getButtonText() }}</chef-button>
           </app-authorized>
           <app-authorized [allOf]="['/iam/v2beta/apply-rules', 'delete']">
--- a/components/automate-ui/src/app/pages/project/list/project-list.component.spec.ts
+++ b/components/automate-ui/src/app/pages/project/list/project-list.component.spec.ts
@@ -559,14 +559,14 @@ describe('ProjectListComponent', () => {
   });
 
   describe('getButtonText', () => {
-    it('labels the button "Projects Up-To-Date" when no projects are edited', () => {
+    it('labels the button "Projects Up-to-Date" when no projects are edited', () => {
       const uneditedProject1 = genProject('uuid-111', 'RULES_APPLIED');
       const uneditedProject2 = genProject('uuid-112', 'RULES_APPLIED');
       store.dispatch(new GetProjectsSuccess({
         projects: [uneditedProject1, uneditedProject2]
       }));
 
-      expect(component.getButtonText()).toEqual('Projects Up-To-Date');
+      expect(component.getButtonText()).toEqual('Projects Up-to-Date');
     });
 
     it('labels the button "Update Projects" when at least one project is edited', () => {
--- a/components/automate-ui/src/app/pages/project/list/project-list.component.spec.ts
+++ b/components/automate-ui/src/app/pages/project/list/project-list.component.spec.ts
@@ -598,6 +598,77 @@ describe('ProjectListComponent', () => {
       expect(component.getButtonText()).not.toEqual('Updating Projects 50%...');
     });
   });
+
+  describe('update projects button', () => {
+    it('is disabled if no project has changes', () => {
+      const uneditedProject1 = genProject('uuid-111', 'RULES_APPLIED');
+      const uneditedProject2 = genProject('uuid-112', 'RULES_APPLIED');
+      store.dispatch(new GetProjectsSuccess({ projects: [uneditedProject1, uneditedProject2] }));
+
+      expect(component.isDisabled()).toEqual(true);
+    });
+
+    it('is enabled if some project has changes', () => {
+      const editedProject = genProject('uuid-99', 'EDITS_PENDING');
+      const uneditedProject = genProject('uuid-111', 'RULES_APPLIED');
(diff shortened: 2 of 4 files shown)


components/automate-ui/src/app/pages/project/list/project-list.component.html
```diff
@@ -21,7 +21,7 @@
               data-cy="create-project" primary (click)="openCreateModal()">Create Project</chef-button>
           </app-authorized>
           <app-authorized [allOf]="['/iam/v2beta/apply-rules', 'post']">
-            <chef-button secondary [disabled]="!projectsHaveStagedChanges || (projects.applyRulesStatus$ | async)?.state === ApplyRulesStatusState.Running"
+            <chef-button secondary [disabled]="isDisabled()"
```
msorens: Move to a function due to increasing complexity, plus ease of testability.

components/automate-ui/src/app/pages/project/list/project-list.component.spec.ts
```diff
@@ -559,43 +559,118 @@ describe('ProjectListComponent', () => {
   });
 
   describe('getButtonText', () => {
```
msorens: Re-organized this test section (and added a couple new tests) to more closely parallel the new section below it.

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -103,11 +103,12 @@ export class ProjectListComponent implements OnInit, OnDestroy {
...
-      // do not update this cache while an update is in progress
-      filter(() => !this.applyRulesInProgress)
+      takeUntil(this.isDestroyed)
     ).subscribe((projectList: Project[]) => {
-      this.statusCache = projectList.reduce((m, p) => ({ ...m, [p.id]: p.status }), {});
+      if (!this.applyRulesInProgress) {
+        // do not update this cache while an update is in progress
+        this.statusCache = projectList.reduce((m, p) => ({ ...m, [p.id]: p.status }), {});
+      }
       this.projectsHaveStagedChanges = projectList.some(p => p.status === 'EDITS_PENDING');
```
msorens: This check needs to happen all the time, so made the filter more localized.

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -148,16 +149,6 @@ export class ProjectListComponent implements OnInit, OnDestroy {
     return '';
   }
 
-  getButtonText(): string {
```
msorens: Just moved this function lower in the file to be with similar functions.

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -286,4 +277,24 @@ export class ProjectListComponent implements OnInit, OnDestroy {
...
+    }
+    if (this.projectsHaveStagedChanges
+      || this.updateProjectsCancelled
+      || this.updateProjectsFailed) {
+      return 'Update Projects';
+    }
+    return 'Projects Up-to-Date';
+  }
+
+  public isDisabled(): boolean {
```
msorens: The new guy on the block.

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -286,4 +277,24 @@ export class ProjectListComponent implements OnInit, OnDestroy {
...
+  public getButtonText(): string {
+    if (this.applyRulesInProgress) {
+      return `Updating Projects ${Math.round(this.percentageComplete * 100)}%...`;
+    }
+    if (this.projectsHaveStagedChanges
+      || this.updateProjectsCancelled
+      || this.updateProjectsFailed) {
+      return 'Update Projects';
+    }
+    return 'Projects Up-to-Date';
```
tylercloke: `Up-To-Date` was intentional from @susanev. Should change back or chat with her.

susanev: there is was just a bunch of miscommunication, originally it was designed as 'Up-to-date`, but then @blakestier suggested `Up-to-Date` which i think is the way to go. but it ened up getting changed to `Up-To-Date`.. so all of that is to say i support the change here by @msorens

Issue #730, opened 2019-07-01
Style select menus in Create ingest rule flow


susanev:
As part of https://chefio.atlassian.net/browse/A2-826 we decided to make the styling of the select menus (attributes, operator) a separate task, this is that task.

PR #1646, opened 2019-09-18
added new style changes for ingest rule select menu

vinay033:

As part of https://chefio.atlassian.net/browse/A2-826 we make the styling of the select menus (attributes, operator) a separate task, this is that task.
I have added changes for this.
https://github.com/chef/automate/issues/730

#### Before
ingest rule before

#### After
ingest rule after

### DIFF
--- a/components/automate-ui/src/app/pages/project/rules/project-rules.component.html
+++ b/components/automate-ui/src/app/pages/project/rules/project-rules.component.html
@@ -79,8 +79,7 @@
                   <span class="label">{{ getAttributeLabel() }}</span>
                   <chef-select ngDefaultControl formControlName="attribute" data-cy="attribute-dropdown" (change)="patchAttributeValue(ruleForm, $event, i)">
                     <chef-option *ngIf="selectAttributeChefOption && currentRuleIndex !== i">Select...</chef-option>
-                    <chef-option *ngFor="let option of attributes[ruleForm.get('type').value.toLowerCase()]; let i = index"
-                    value="option.key">{{ option.value }}</chef-option>
+                    <chef-option *ngFor="let option of attributeList" value={{option.key}}>{{ option.value }}</chef-option>
                   </chef-select>
                 </label>
                 <chef-error
@@ -92,8 +91,8 @@
                 <label>
                   <span class="label">Operator</span>
                   <chef-select ngDefaultControl formControlName="operator" data-cy="operator-dropdown" (change)="patchOperatorValue(ruleForm, $event, i)">
-                      <chef-option *ngIf="selectOperatorChefOption && currentRuleIndex !== i">Select...</chef-option>
-                    <chef-option *ngFor="let operator of operators" value="operator.key">{{ operator.value }}</chef-option>
+                    <chef-option *ngIf="selectOperatorChefOption && currentRuleIndex !== i">Select...</chef-option>
+                    <chef-option *ngFor="let operator of operators" value={{operator.key}}>{{ operator.value }}</chef-option>
                   </chef-select>
                 </label>
                 <chef-error
--- a/components/automate-ui/src/app/pages/project/rules/project-rules.component.html
+++ b/components/automate-ui/src/app/pages/project/rules/project-rules.component.html
@@ -10,7 +10,6 @@
   <div *ngIf="!isLoading">
     <ng-container>
       <form id="ruleForm" [formGroup]="ruleForm" (ngSubmit)="saveRule()">
-
         <chef-form-field id="create-name">
           <label>
             <span class="label">Rule Name <span aria-hidden="true">*</span></span>
@@ -52,10 +51,11 @@
         <chef-form-field id="create-type">
           <label>
             <span class="label">Resource Type <span aria-hidden="true">*</span></span>
-            <select id="create-type-dropdown" ngDefaultControl formControlName="type">
-              <option value='NODE'>Node</option>
-              <option value='EVENT'>Event</option>
-            </select>
+            <chef-select id="create-type-dropdown" ngDefaultControl formControlName="type" (change)="patchResourceValue(ruleForm, $event)">
+              <chef-option>Select Resource Type</chef-option>
+              <chef-option value='NODE'>Node</chef-option>
+              <chef-option value='EVENT'>Event</chef-option>
+            </chef-select>
           </label>
           <chef-error
             *ngIf="ruleForm.get('type').hasError('required') && ruleForm.get('type').touched"
@@ -77,26 +77,24 @@
               <chef-form-field class="attribute-field">
                 <label>
                   <span class="label">{{ getAttributeLabel() }}</span>
-                  <select ngDefaultControl formControlName="attribute" data-cy="attribute-dropdown">
-                    <option
-                      *ngFor="let option of attributes[ruleForm.get('type').value.toLowerCase()]; let i = index"
-                      [value]="option.key"
-                    >{{ option.value }}</option>
-                  </select>
+                  <chef-select ngDefaultControl formControlName="attribute" data-cy="attribute-dropdown" (change)="patchAttributeValue(ruleForm, $event, i)">
+                    <chef-option>Select {{ getAttributeLabel() }}</chef-option>
+                    <chef-option *ngFor="let option of attributes[ruleForm.get('type').value.toLowerCase()]; let i = index"
+                    [value]="option.key">{{ option.value }}</chef-option>
+                  </chef-select>
                 </label>
                 <chef-error
                   *ngIf="ruleForm.get('conditions').controls[i].get('attribute').hasError('required') && ruleForm.get('conditions').controls[i].get('attribute').touched"
                 >Attribute is required.</chef-error>
               </chef-form-field>
 
-              <chef-form-field class="attribute-field">
+              <chef-form-field class="operator-field">
                 <label>
                   <span class="label">Operator</span>
-                  <select ngDefaultControl formControlName="operator" data-cy="operator-dropdown">
-                    <option *ngFor="let operator of operators"
-                      [value]="operator.key"
-                    >{{ operator.value }}</option>
-                  </select>
+                  <chef-select ngDefaultControl formControlName="operator" data-cy="operator-dropdown" (change)="patchOperatorValue(ruleForm, $event, i)">
+                    <chef-option>Select Operator</chef-option>
+                    <chef-option *ngFor="let operator of operators" [value]="operator.key">{{ operator.value }}</chef-option>
+                  </chef-select>
                 </label>
                 <chef-error
                   *ngIf="ruleForm.get('conditions').controls[i].get('operator').hasError('required') && ruleForm.get('conditions').controls[i].get('operator').touched"
@@ -131,7 +129,6 @@
                   <chef-icon>delete</chef-icon>
                 </chef-button>
               </div>
-
             </div>
           </div>
         </div>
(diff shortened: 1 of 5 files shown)


susanev:
@vinay033 a few things...
- items in the menus should have a font-weight of 400, right now they seem to be bold
- is it possible to do this without the added elements of "Select..."?, meaning it just starts as blank and they have to make a selection, but they cant go back to the blank state.. similar to how it was previously

vinay033:
> @vinay033 a few things...
> 
> * items in the menus should have a font-weight of 400, right now they seem to be bold
> * is it possible to do this without the added elements of "Select ..."?, meaning it just starts as blank and they have to make a selection, but they cant go back to the blank state.. similar to how it was previously. Or could we at a minimum change the text to "Select..." with no object stated and not allow them to go back to that state after they make their first selection.

sure, I will.

vinay033:
> 1. Not really clear what is changed visually. I assume your screenshot is "after". Please add a "before" screenshot as well, and point out what has changed.
> 2. The title suggests this is strictly style changes, but your changes are more than just CSS. Could you explain a bit about why the structure and code-behind needed to change. Thanks!

I have added the screenshot, please have a look and regarding second option I used the common select component for this.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -77,26 +77,24 @@
...
-                  <select ngDefaultControl formControlName="attribute" data-cy="attribute-dropdown">
-                    <option
-                      *ngFor="let option of attributes[ruleForm.get('type').value.toLowerCase()]; let i = index"
-                      [value]="option.key"
-                    >{{ option.value }}</option>
-                  </select>
+                  <chef-select ngDefaultControl formControlName="attribute" data-cy="attribute-dropdown" (change)="patchAttributeValue(ruleForm, $event, i)">
+                    <chef-option *ngIf="selectAttributeChefOption && currentRuleIndex !== i">Select...</chef-option>
+                    <chef-option *ngFor="let option of attributes[ruleForm.get('type').value.toLowerCase()]; let i = index"
+                    value="option.key">{{ option.value }}</chef-option>
```
msorens: Would prefer to see complex logic kept out of the template, so suggest replacing L82-83 with:
```suggestion
<chef-option *ngFor="let option of attributes" value="option.key">{{ option.value }}</chef-option>
```
where you define the `attributes` property in the backing code.
That has the added benefit of this line then mirroring L96 below.
(And I do not believe the `let i = index` portion there is used at all.)

vinay033: @msorens I have updated the changes please have a look.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -52,10 +51,11 @@
         <chef-form-field id="create-type">
           <label>
             <span class="label">Resource Type <span aria-hidden="true">*</span></span>
-            <select id="create-type-dropdown" ngDefaultControl formControlName="type">
-              <option value='NODE'>Node</option>
-              <option value='EVENT'>Event</option>
-            </select>
+            <chef-select id="create-type-dropdown" ngDefaultControl formControlName="type" (change)="patchResourceValue(ruleForm, $event)">
+              <chef-option *ngIf="selectResourceChefOption">Select...</chef-option>
```
msorens: Can do this without the extra variable:
```
              <chef-select id="create-type-dropdown" #ruleType ngDefaultControl formControlName="type" (change)="patchResourceValue(ruleForm, $event)">
              <chef-option *ngIf="!ruleType.value">Select...</chef-option>
```

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -77,26 +77,23 @@
...
                 <label>
                   <span class="label">Operator</span>
-                  <select ngDefaultControl formControlName="operator" data-cy="operator-dropdown">
-                    <option *ngFor="let operator of operators"
-                      [value]="operator.key"
-                    >{{ operator.value }}</option>
-                  </select>
+                  <chef-select ngDefaultControl formControlName="operator" data-cy="operator-dropdown" (change)="patchOperatorValue(ruleForm, $event, i)">
+                    <chef-option *ngIf="selectOperatorChefOption && currentRuleIndex !== i">Select...</chef-option>
+                    <chef-option *ngFor="let operator of operators" value={{operator.key}}>{{ operator.value }}</chef-option>
```
msorens:
```suggestion
                    <chef-option *ngFor="let operator of operators" value="{{operator.key}}">{{ operator.value }}</chef-option>
```

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -77,26 +77,23 @@
...
                 <label>
                   <span class="label">{{ getAttributeLabel() }}</span>
-                  <select ngDefaultControl formControlName="attribute" data-cy="attribute-dropdown">
-                    <option
-                      *ngFor="let option of attributes[ruleForm.get('type').value.toLowerCase()]; let i = index"
-                      [value]="option.key"
-                    >{{ option.value }}</option>
-                  </select>
+                  <chef-select ngDefaultControl formControlName="attribute" data-cy="attribute-dropdown" (change)="patchAttributeValue(ruleForm, $event, i)">
+                    <chef-option *ngIf="selectAttributeChefOption && currentRuleIndex !== i">Select...</chef-option>
```
msorens: This (1) gets rid of two extra variables, similar to above, but also (2) fixes a bug. The currentRuleIndex could be updated frequently if there are multiple new conditions added. So that would make `Select` reappear as an option when it should not.
```
                  <chef-select ngDefaultControl #ruleAttribute formControlName="attribute" data-cy="attribute-dropdown" (change)="patchAttributeValue(ruleForm, $event, i)">
                    <chef-option *ngIf="!ruleAttribute.value">Select...</chef-option>
```

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -77,26 +77,23 @@
...
+              <chef-form-field class="operator-field">
                 <label>
                   <span class="label">Operator</span>
-                  <select ngDefaultControl formControlName="operator" data-cy="operator-dropdown">
-                    <option *ngFor="let operator of operators"
-                      [value]="operator.key"
-                    >{{ operator.value }}</option>
-                  </select>
+                  <chef-select ngDefaultControl formControlName="operator" data-cy="operator-dropdown" (change)="patchOperatorValue(ruleForm, $event, i)">
+                    <chef-option *ngIf="selectOperatorChefOption && currentRuleIndex !== i">Select...</chef-option>
```
msorens: As above, gets rid of two extra variables and addresses the same bug.
```
                  <chef-select ngDefaultControl #ruleOperator formControlName="operator" data-cy="operator-dropdown" (change)="patchOperatorValue(ruleForm, $event, i)">
                    <chef-option *ngIf="!ruleOperator.value">Select...</chef-option>
```

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -41,6 +41,11 @@ interface KVCondition {
 export class ProjectRulesComponent implements OnInit, OnDestroy {
   public ruleId: string;
   public ruleForm: FormGroup;
+  public selectResourceChefOption = true;
+  public selectAttributeChefOption = true;
+  public selectOperatorChefOption = true;
+  public currentRuleIndex: number;
```
msorens: 4 variables no longer needed with template changes above.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -153,6 +158,30 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
     this.isDestroyed.complete();
   }
 
+  public patchResourceValue(form, event) {
```
msorens: Suggest `patchTypeValue` instead.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -153,6 +158,30 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
...
+    this.selectResourceChefOption = false;
+  }
+
+  public patchAttributeValue(form, event, i) {
+    this.currentRuleIndex = i;
+    form.controls.conditions.controls[i].controls['attribute'].setValue(event.target.value);
+    this.selectAttributeChefOption = false;
+    if (this.editingRule) {
+      form.controls.conditions.controls[i].controls['attribute'].markAsDirty();
+    }
```
msorens: This causes a bug.
Previously, for an existing rule, if you start with the attribute at value "A" and change to "B", that would enable the "Save Rule" button. But further, if you change it back to "A", that would disable the "Save Rule" button. That dirty/clean check is broken here due to this block, but might also be tied to switching from `<select>` to `<chef-select>`. 

So first see if you can determine if the bug is due to using `<chef-select>` (by changing it back to `<select>` and deleting L171-173). If that is the case, then try to fix `<chef-select>`--we do not want custom code added in _this_ file to do it, because then everytime we use `<chef-select>` we have to duplicate that kludge.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -153,6 +158,30 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
...
+  }
+
+  public patchOperatorValue(form, event, i) {
+    this.currentRuleIndex = i;
+    form.controls.conditions.controls[i].controls['operator'].setValue(event.target.value);
+    this.selectOperatorChefOption = false;
+    if (this.editingRule) {
+      form.controls.conditions.controls[i].controls['operator'].markAsDirty();
+    }
```
msorens: Same goes for this block.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
vinay033: @msorens thanks for review, I have verified this bug is due to using ` <chef-select> ` so fixing it in ` chef-ui-library ` it self so we can avoid code duplication

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -48,6 +48,7 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
 
   public isLoading = true;
   public saving = false;
+  public attributeList: any;
```
msorens: Please use strong typing:
```suggestion
  public attributeList: KVPair[];
```

components/automate-ui/src/app/pages/project/rules/project-rules.component.scss
```diff
@@ -29,6 +29,71 @@ chef-page {
...
+    border: solid 1px $chef-grey;
+    text-indent: 0.01px;
+    text-overflow: '';
+    font-weight: 400;
+    font-stretch: normal;
+    font-style: normal;
+    line-height: 1.5;
+    letter-spacing: 0.2px;
+    outline: none;
+    /*Hiding the select arrow for firefox*/
```
msorens: For each of your added comments like this please put use a space on either end, so:
```suggestion
    /* Hiding the select arrow for firefox */
```

components/automate-ui/src/app/pages/project/rules/project-rules.component.scss
```diff
@@ -29,6 +29,71 @@ chef-page {
     max-width: 340px;
   }
 
+  /* Custom dropdown */
+  .custom-dropdown {
+    position: relative;
+    vertical-align: middle;
+  }
+
+  .custom-dropdown select {
```
tarablack01: @vinay033 If you could nest all the styling for .custom-dropdown
`.custom-dropdown {`
 `   position: relative;`
 `   vertical-align: middle;`
 ` }`

 ` & select {`

vinay033: Thanks, @tarablack01  I glad to see this kind of appreciation from you, this credit goes to @sachin-msys, he has put his efforts for the new styling.

Issue #1594, opened 2019-09-12
integration tests for Compliance node and event filtering


bcmdarroch:
## User Story

As a Developer, I will use this work when I want to verify through automated testing that ingest rules are applied correctly to Compliance nodes and Events (aka Actions).

We already have similar testing with infra nodes in this Cypress spec [01_projects_api.spec.ts](https://github.com/chef/automate/blob/master/e2e/cypress/integration/api/iam/01_projects_api.spec.ts).

[Compliance node fixture example](https://github.com/chef/automate/blob/master/components/compliance-service/ingest/examples/compliance-success-tiny-report.json)
[Event (Action) fixture example](https://github.com/chef/automate/tree/master/components/ingest-service/examples/actions)

Map of Rule Attributes to Corresponding Node/Action Fields
<img width="543" alt="Screen Shot 2019-09-19 at 8 39 51 AM" src="https://user-images.githubusercontent.com/21015366/65259185-1c4b1b80-dab9-11e9-8578-414e147c522e.png">

## Definition of Done
Test scenarios should run roughly like this:

1. Using a Compliance node fixture:
- we can create ingest rules for each of the following attributes: Chef Organization, Chef Server, Environment, Chef Role, Chef Tag, Chef Policy Group, Chef Policy Name
- when rules are applied, the Compliance node is tagged with the correct project

2. Using an Event fixture
- we can create ingest rules for each of the following attributes: Chef Organization, Chef Server
- when rules are applied, the Event is tagged with the correct project

## Demo Script / Repro Steps
- run integration tests

bcmdarroch:
I looked into starting these tests last week, and the main hurdle I ran into was figuring out how to load Compliance nodes via curl. [This example](https://github.com/chef/automate/blob/master/.studio/automate-gateway#L283) looked promising. Haven't found any examples for actions yet, so we should reach out to superteam about that

sdelano:
Sprint planning notes:
- @lancewf says we also have the nodemanager API that's implemented with project filtering as well
- current example test:
  - sends client runs

PR #1659, opened 2019-09-19
[automate-1618] expand ingest rule integration tests + refactor tokens in API-only tests

bcmdarroch:

Added some extra due diligence tests to ensure ingest project rules get correctly applied to infra nodes for each available attribute.

EDIT: starting with [this commit](https://github.com/chef/automate/pull/1659/commits/33ac8baa0e09daf1daa613b0f29cd72a8c7d83a0), I also refactored our API-only tests so they don't have to open the browser to log in. That should save some test run time!

follow-up issue adds Compliance node and Event test coverage: https://github.com/chef/automate/issues/1594

### :+1: Definition of Done
- tests pass

[how to run cypress tests locally](https://github.com/chef/automate/blob/master/e2e/README.md#running-cypress-locally)

- [x] Tests added/updated?

### DIFF
--- a/.expeditor/buildkite/cypress.sh
+++ b/.expeditor/buildkite/cypress.sh
@@ -25,7 +25,7 @@ do
           return 1
         fi
       else 
-        if ! token=$(chef-automate iam token create $timestamp-tok --admin); then
+        if ! token=$(chef-automate iam token create "$timestamp-tok" --admin); then
           echo "Non-zero exit code, output:"
           echo "$token"
           return 1
--- a/.expeditor/buildkite/cypress.sh
+++ b/.expeditor/buildkite/cypress.sh
@@ -7,6 +7,7 @@ cd /workdir/e2e
 data=$(curl --silent "https://a2-${CHANNEL}.cd.chef.co/assets/data.json")
 instances_to_test=$(jq -nr --argjson data "$data" '$data[] | select(.tags | any(. == "e2e")) | .fqdn')
 versions=(v1 v2 v2.1)
+timestamp=$(date +"%m-%d-%y-%H-%M")
 
 for instance in ${instances_to_test[*]}
 do
@@ -17,6 +18,21 @@ do
   for version in ${versions[*]}
   do
     if jq -enr --arg version "$version" --arg fqdn "$instance" --argjson data "$data" '$data[] | select(.fqdn == $fqdn) | .tags | any(. == $version)'; then
+      if [[ "$version" == "v1" ]]; then
+        if ! token=$(chef-automate admin-token); then
+          echo "Non-zero exit code, output:"
+          echo "$token"
+          return 1
+        fi
+      else 
+        if ! token=$(chef-automate iam token create $timestamp-tok --admin); then
+          echo "Non-zero exit code, output:"
+          echo "$token"
+          return 1
+        fi
+      fi
+      
+      export CYPRESS_ADMIN_TOKEN=$token
       export CYPRESS_IAM_VERSION=$version
     fi
   done
--- a/.expeditor/verify_private.pipeline.yml
+++ b/.expeditor/verify_private.pipeline.yml
@@ -412,7 +412,7 @@ steps:
   #
   - label: ":cypress: (flaky included) IAMv1"
     command:
-      - IAM=v1.0 integration/run_test integration/tests/cypress_e2e.sh
+      - IAM=v1 integration/run_test integration/tests/cypress_e2e.sh
     timeout_in_minutes: 20
     soft_fail: true
     expeditor:
@@ -437,7 +437,7 @@ steps:
 
   - label: ":cypress: (flaky included) IAMv2"
     command:
-      - IAM=v2.0 integration/run_test integration/tests/cypress_e2e.sh
+      - IAM=v2 integration/run_test integration/tests/cypress_e2e.sh
     timeout_in_minutes: 20
     soft_fail: true
     expeditor:
@@ -463,7 +463,7 @@ steps:
 
   - label: ":cypress: IAMv1"
     command:
-      - IAM=v1.0 integration/run_test integration/tests/cypress_e2e.sh
+      - IAM=v1 integration/run_test integration/tests/cypress_e2e.sh
     timeout_in_minutes: 20
     expeditor:
       secrets: &cypress_secrets
--- a/e2e/cypress.json
+++ b/e2e/cypress.json
@@ -1,8 +1,5 @@
 {
   "projectId": "yvg8zo",
   "defaultCommandTimeout": 25000,
-  "pageLoadTimeout": 80000,
-  "env": {
-    "adminTokenValue": ""
-  }
+  "pageLoadTimeout": 80000
 }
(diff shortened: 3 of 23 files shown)


e2e/cypress/integration/api/iam/01_projects_api.spec.ts
```diff
@@ -481,6 +488,79 @@ describeIfIAMV2p1('projects API: applying project', () => {
...
+      }).then((response) => {
+        expect(response.body).to.have.length(1);
+        expect(response.body[0].id).to.equal(targetNodeId);
+      });
+
+      // delete rule
+      cy.request({
+        headers: { 'api-token': Cypress.env('adminTokenValue') },
+        method: 'DELETE',
+        url: `/apis/iam/v2beta/projects/${avengersRule.project_id}/rules/${rule.id}`
```
blakestier: Should we have expectations for this and the other below?

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
```diff
@@ -159,6 +159,13 @@ describeIfIAMV2p1('projects API: applying project', () => {
     });
   });
 
+  beforeEach(() => {
+    cy.restoreStorage();
+  });
+  afterEach(() => {
+    cy.saveStorage();
+  });
```
msorens: Is this already in use in other tests or is this something new and should be used in every test?

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
```diff
@@ -481,6 +488,79 @@ describeIfIAMV2p1('projects API: applying project', () => {
       expect(response.body).to.have.length(2);
     });
   });
+
+  const attributeValuesMap = new Map([
```
msorens: Please add this comment below.
(Note this identical comment is used elsewhere, so I think it is a good idea to reuse it verbatim. If you want to revise it, then suggest revising it in both places.)
```
must correspond to enum type in automate-gateway/.../common/rules.proto
```

.expeditor/buildkite/cypress.sh
```diff
@@ -17,6 +18,21 @@ do
...
+        fi
+      else 
+        if ! token=$(chef-automate iam token create $timestamp-tok --admin); then
+          echo "Non-zero exit code, output:"
+          echo "$token"
+          return 1
+        fi
+      fi
+      
+      export CYPRESS_ADMIN_TOKEN=$token
```
bcmdarroch: all the changes in this commit are to ensure an admin token is always available to the Cypress test runner.

e2e/cypress/integration/constants.ts
```diff
@@ -3,3 +3,4 @@ export const describeIfIAMV2 = iamVersion.match(/v2/) ? describe : describe.skip
 export const describeIfIAMV2p1 = iamVersion === 'v2.1' ? describe : describe.skip;
 export const runFlaky: boolean = Cypress.env('RUN_FLAKY') || false;
 export const itFlaky = runFlaky ? it : it.skip;
+export const adminToken: string = Cypress.env('ADMIN_TOKEN');
```
bcmdarroch: Cypress will now depend on having the env variable `CYPRESS_ADMIN_TOKEN` always being set. This adds a little extra overhead when running these tests locally, but it'll help our API-only tests run much more quickly by avoiding the time-consuming browser login interaction

e2e/cypress/integration/api/iam/02_policies_api.spec.ts
```diff
@@ -1,9 +1,8 @@
-import { describeIfIAMV2, describeIfIAMV2p1 } from '../../constants';
+import { describeIfIAMV2, describeIfIAMV2p1, adminToken } from '../../constants';
 
 describeIfIAMV2('policies API', () => {
-  let adminToken = '';
   const defaultAdminReq = {
-    auth: { bearer: adminToken },
+    headers: { 'api-token': adminToken },
```
bcmdarroch: most of the changes here are swapping the admin user's bearer token for the admin API token saved as a constant.

e2e/cypress.json
```diff
@@ -1,8 +1,5 @@
 {
   "projectId": "yvg8zo",
   "defaultCommandTimeout": 25000,
-  "pageLoadTimeout": 80000,
-  "env": {
-    "adminTokenValue": ""
```
bcmdarroch: we no longer need to use Cypress to overwrite the env variable set on the OS.

e2e/cypress/index.d.ts
```diff
@@ -7,9 +7,8 @@ declare namespace Cypress {
     saveStorage(): void
     restoreStorage(): void
     applyProjectsFilter(projectsToFilterOn: string[]): void
-    generateAdminToken(idToken: string): void
-    cleanupV2IAMObjectsByIDPrefixes(idToken: string, idPrefix: string, objectPlurals: string[]): void
-    cleanupUsersByNamePrefix(idToken: string, namePrefix: string): void
-    cleanupTeamsByDescriptionPrefix(idToken: string, namePrefix: string): void
+    cleanupV2IAMObjectsByIDPrefixes(idPrefix: string, objectPlurals: string[]): void
```
bcmdarroch: as an added bonus to this change, I dropped the idToken from all of these helpers. Before, we had to pass the admin user's bearer token for all of these API requests. Now we can just use the admin API token available as an environment variable.

e2e/cypress/integration/ui/settings/03_team_add_users.spec.ts
```diff
@@ -8,20 +8,20 @@ describe('team add users', () => {
   const descriptionForTeam = cypressPrefix + ' team ' + now;
   const nameForTeam = 'testing-team-' + now;
   let teamID = '';
-  let adminToken = '';
+  let adminIdToken = '';
```
bcmdarroch: renamed these variables to distinguish the admin user's ID token from the generated admin API token saved in the environment

.expeditor/verify_private.pipeline.yml
```diff
@@ -412,7 +412,7 @@ steps:
   #
   - label: ":cypress: (flaky included) IAMv1"
     command:
-      - IAM=v1.0 integration/run_test integration/tests/cypress_e2e.sh
+      - IAM=v1 integration/run_test integration/tests/cypress_e2e.sh
```
bcmdarroch: must be `v1 | v2 | v2.1` to match the way the acceptance/dev builds are tagged in terraform (you can find the tags on the `.tf` files in the [terraform directory](https://github.com/chef/automate/blob/master/terraform/test-environments/scenarios))

.expeditor/buildkite/cypress.sh
blakestier: yaaaay

e2e/cypress/integration/api/iam/01_projects_api.spec.ts
bcmdarroch: This was needed when we were passing around the admin user's id token around. But after refactoring these tests to use the admin API token from the env, I took it out. 

I think all of the UI tests should be using these helpers - they help prevent the test getting kicked off by a session refresh. We mock the session/refresh response coming back with the same ID token each time - but if that ID token gets dropped btwn the tests because we're not saving storage, that mocked response will still lead to a forced logout.

PR #1661, opened 2019-09-19
Introspection fixes for perm-based routing

msorens:

The recent PR that introduced introspection caching introduced a bug that caused permission-based routing to fail for non-admin users.
This manifested as a mostly blank screen when navigating to settings:
image

There were actually two problems:
First, this revealed a bug in `IntrospectAll`. Some time ago all introspection methods were changed to allegedly return all requested endpoints, whether they had any permissions on them or not. This worked for `Introspect(One)` and `IntrospectSome` but it turns out not for `IntrospectAll`.

The `LandingComponent`, which does permission-based routing to the first page amongst a set of pages for which the user has permission, waits until all of the pages on its list are accounted for in the ngrx store, but because of the bug in `IntrospectAll` that never happened.

Once that was fixed, the next issue was that the `LandingComponent` was relying on a subscription to `allPerms` to kick off the next check in the list of pages. But the way caching was implemented, there were no subsequent fetches, so again it waited in vain. This was fixed by enhancing the caching so that regardless of whether it made a network call or not, it completed the "cycle" of get/get-success, and that allowed the proper triggers to work.

### :+1: Definition of Done

No more blank pages.
The simple test is to assign the non-admin user to the viewers team. But that would land you on the very first page in the set of possible pages. So just to make it a little more interesting, here I did *not* include the user on the viewers team, but rather used this custom policy:
```
{
  "id": "perm-test",
  "name": "testing",
  "projects": [],
  "statements": [
    {
      "effect": "ALLOW",
      "projects": ["*"]
    }
  ]
}
```
You will see, then, that the user correctly lands on the Teams page, which is several items down in the list of possible pages, but is the first allowable page for this user:

image

rebuild automate-gateway
rebuild automate-ui
add a non-admin user
assign to a team or use a custom policy
navigate to settings

### DIFF
--- a/components/automate-gateway/handler/authz.go
+++ b/components/automate-gateway/handler/authz.go
@@ -118,11 +118,9 @@ func (a *AuthzServer) IntrospectAll(
 	mapByResourceAndActionV1 := pairs.InvertMapNonParameterized(methodsInfoV1)
 	mapByResourceAndActionV2 := pairs.InvertMapNonParameterized(methodsInfoV2)
 
-	// Extra work of hydrating not needed for IntrospectAll; already does that.
 	endpointMap, err := a.getAllowedMap(ctx,
 		mapByResourceAndActionV1, mapByResourceAndActionV2,
-		methodsInfoV1, methodsInfoV2,
-		false)
+		methodsInfoV1, methodsInfoV2)
 	if err != nil {
 		return nil, err
 	}
@@ -155,7 +153,7 @@ func (a *AuthzServer) IntrospectSome(
 
 	endpointMap, err := a.getAllowedMap(ctx,
 		mapByResourceAndActionV1, mapByResourceAndActionV2,
-		methodsInfoV1, methodsInfoV2, true)
+		methodsInfoV1, methodsInfoV2)
 	if err != nil {
 		return nil, err
 	}
@@ -195,7 +193,7 @@ func (a *AuthzServer) Introspect(
 
 	endpointMap, err := a.getAllowedMap(ctx,
 		mapByResourceAndActionV1, mapByResourceAndActionV2,
-		methodsInfoV1, methodsInfoV2, true)
+		methodsInfoV1, methodsInfoV2)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


components/automate-gateway/handler/authz.go
```diff
@@ -296,8 +295,7 @@ func reportBadPaths(log *logrus.Entry, lookupHash map[string]bool) {
 func (a *AuthzServer) getAllowedMap(
 	ctx context.Context,
 	mapByResourceAndActionV1, mapByResourceAndActionV2 map[pairs.Pair][]string,
-	methodsInfoV1, methodsInfoV2 map[string]pairs.Info,
-	fullyHydrate bool) (map[string]*gwAuthzRes.MethodsAllowed, error) {
+	methodsInfoV1, methodsInfoV2 map[string]pairs.Info) (map[string]*gwAuthzRes.MethodsAllowed, error) {
```
msorens: Getting rid of `fullyHydrate` parameter--want to always do it!

components/automate-ui/src/app/entities/userperms/userperms.effects.ts
```diff
@@ -60,11 +62,22 @@ export class UserPermEffects {
   fetchSomePerms$ = this.actions$.pipe(
     ofType(UserPermsTypes.GET_SOME),
     withLatestFrom(this.store.select(getlastFetchTime)),
-    filter(([_action, lastTime]) => this.stale(lastTime)),
```
msorens: Instead of filtering for staleness, do the same check in the body of the effect below, then issue a new action--either one that returns immediately or one that fetches data with a network call.

components/automate-gateway/handler/authz.go
```diff
@@ -312,7 +309,7 @@ func (a *AuthzServer) getAllowedMap(
 		return nil, err
 	}
 	endpointMap, err := pairs.GetEndpointMapFromResponse(
-		resp.Pairs, resp.MethodsInfo, resp.MapByResourceAndAction, fullyHydrate)
+		resp.Pairs, resp.MethodsInfo, resp.MapByResourceAndAction, true)
```
bcmdarroch: do we still need the last parameter at all?

msorens: Not really, but I did not see any urgency in stripping that. I prefer to minimize changes when there is time pressure.

msorens: Good eye, by the way, @bcmdarroch !

PR #1662, opened 2019-09-19
temporary fix for chef-server tests

stevendanna:
This pins us to my branch with a fix until that fix is merged and
released upstream.

We can revert when

   https://github.com/chef-cookbooks/audit/pull/388

is merged


stevendanna:
Looks like we got out a fix for this very quickly so we shouldn't need this.

PR #1666, opened 2019-09-19
Jamie/buttons hide profile description

SEAjamieD:
On Profile pages, the description of the profile would sometimes be hidden by the Remove and Download buttons, preventing a user from being able to read the full description.

To resolve, a container was placed around the slot element (the buttons) to more easily target that container's style.  A max width was set on `chef-subheadings` within profile pages, preventing the buttons from ever reaching a point where they can cause overlap.  The buttons container was also shifted down ~10px in order not cover the profile title.

### :+1: Definition of Done
The buttons container no longer covers the profile title or description.

1. Build components/automate-ui-devproxy && start_all_services
2. Navigate to https://a2-dev.test/compliance/compliance-profiles
3. Click on Available profiles tab
4. Click on profiles titles to view various profiles from list
5. Drag the screen in and out - buttons should not overlap content

- [ ] Tests added/updated?
- [ ] Docs added/updated?

buttons-hide-content

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/+profile/+profile-details/profile-details.component.html
+++ b/components/automate-ui/src/app/pages/+compliance/+profile/+profile-details/profile-details.component.html
@@ -35,23 +35,25 @@
       <chef-page-header>
         <chef-heading>{{ profile.title }}</chef-heading>
         <chef-subheading>{{ profile.summary }}</chef-subheading>
-        <div slot="header-buttons">
-          <ng-container *ngIf="!isInstalled">
-            <chef-button secondary caution id="install-btn" (click)="installProfile(profile)">
-              <chef-icon>get_app</chef-icon>
-              <span>Get</span>
+        <div class="header-buttons-container">
+          <div slot="header-buttons">
+            <ng-container *ngIf="!isInstalled">
+              <chef-button secondary caution id="install-btn" (click)="installProfile(profile)">
+                <chef-icon>get_app</chef-icon>
+                <span>Get</span>
+              </chef-button>
+            </ng-container>
+            <ng-container *ngIf="isInstalled">
+              <chef-button secondary caution id="uninstall-btn" (click)="showDeleteModal()">
+                <chef-icon>delete</chef-icon>
+                <span>Remove</span>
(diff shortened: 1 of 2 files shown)


components/automate-ui/src/app/pages/+compliance/+profile/+profile-details/profile-details.component.html
```diff
@@ -35,23 +35,25 @@
       <chef-page-header>
         <chef-heading>{{ profile.title }}</chef-heading>
         <chef-subheading>{{ profile.summary }}</chef-subheading>
-        <div slot="header-buttons">
-          <ng-container *ngIf="!isInstalled">
-            <chef-button secondary caution id="install-btn" (click)="installProfile(profile)">
-              <chef-icon>get_app</chef-icon>
-              <span>Get</span>
+        <div class="header-buttons-container">
```
SEAjamieD: Added a container <div> so that the actual buttons container could be targeted only in profile pages.  The `header-buttons` <div> is rendered as a slot and I'm unable to target locally without this container.

components/automate-ui/src/app/pages/+compliance/+profile/+profile-details/profile-details.component.scss
```diff
@@ -6,6 +6,20 @@ chef-loading-spinner {
   width: 100px;
 }
 
+chef-page-header {
+  chef-subheading {
+    max-width: 65%;
```
SEAjamieD: I feel max-width is better as a percentage so that it's a bit more flexible to screen widths than a hard pixel number.

components/automate-ui/src/app/pages/+compliance/+profile/+profile-details/profile-details.component.scss
```diff
@@ -6,6 +6,20 @@ chef-loading-spinner {
...
+  }
+  
+  .header-buttons-container {
```
SEAjamieD: These are essentially the same styling as the header-buttons slot element, with a higher top pixel, so that it matches what is generally used across the site.

Issue #744, opened 2019-07-02
compliance tab users cannot filter Redhat platforms by version like they can with windows - "Customer escalation:  ZD 18086"


vinay033:
# User Story
If a user is searching for a Redhat platform in the compliance filter dropdown they cannot select a particular version of the platform. When searching for windows systems it will include the entire string for the platform name. For example, when doing a drop down it will show "RedHat" or "Windows Server 2012 R2 Standard".
# Definition of Done
User should be able to select full platform name with a version like they can with Windows systems from the filter
# Demo Script / Repro Steps
go to the compliance tab and select Platform from the dropdown filter list. If you have Windows systems you will see the full platform name with version. See the ZD ticket for screenshots with examples of windows and Linux options.

Issue #633, opened 2019-06-19
implement wildcard search on platform version number and profile version number for compliance


vjeffrey:
# User Story
This is a known issue carried over from https://github.com/chef/automate/issues/325

Let's revisit this when we have some time to give it the special attention it needs.

# Definition of Done
Can filter with wildcard on platform version number and profile version number.

vjeffrey:
let's investigate why this is.
if it's not possible to do the wildcard searching on those fields, then let's go ahead and note that here and close the issue

lancewf:
@vjeffrey The reason that wildcard search does not work with the platform and profile version numbers is that the version numbers are in a separate field. For the client runs page we fixed this problem for the platform field by concatenating the version on to the platform field.

vjeffrey:
changed the issue title to reflect what needs to be done to make this work; sounds like we just need alias fields for both of those where we concatenate the version number onto the text field, as @lancewf mentioned was done for client runs

lancewf:
So we do have a field with the platform name and version combined. 
https://github.com/chef/automate/blob/2e9f37aafb847850da5b5d38a443b9d7c6facfc2/components/compliance-service/ingest/pipeline/processor/report.go#L57

We need to update suggestions to use that field 
https://github.com/chef/automate/blob/72c60c610a100fd8578eb36c419b1e65507a1a10/components/compliance-service/reporting/relaxting/suggestions.go#L32. 

We also need to update filtering to use that field. 
https://github.com/chef/automate/blob/997be2f1f15bb25fea61d315d61a3ac72f83fadf/components/compliance-service/reporting/relaxting/reports.go#L547


I will look to see if profile version will be the same.

lancewf:
We will need to do a migration to switch platform.full field in Elasticsearch to allow wildcards. 
https://github.com/chef/automate/blob/be1cb11c6b66d48896e314f0b46086f9d99bbc4c/components/compliance-service/ingest/ingestic/mappings/comp-rep-date.go#L193

PR #1672, opened 2019-09-19
Platform And Profile version wildcard filtering

lancewf:
### :clipboard: TODO
- [x] elasticsearch migration
- [x] Create a new filtering field 'platform_with_version'
- [x] Update the UI to use 'platform_with_version'. 
- [x] API Tests for the 'platform_with_version' filter type
- [x] Create a new filtering field 'profile_with_version'
- [x] Update the UI to use field 'profile_with_version'
- [x] Add migration for profile.full to fill the field from current data
- [x] Return the profile.full and platform.full field on node requests
- [x] Test for profile.full and platform.full field on node requests
- [x] migration tests

Adding the ability to filter reports based on a platform and profile names and version numbers. 
This changes the UI and the compliance-service. The compliance-service adds two new filter types with the name of `platform_with_version` and `profile_with_version`. These allow suggestions and filtering of reports with the versions combined with the name of the item. 

These changes involved a migration that creates the profile full name (that is the profile title and version combined) and allows suggestions on the platform's full name. This migration works all the back to Automate 1. 

Added the full name to the platform and profile fields for API requests. 

#633 
#744

### :+1: Definition of Done
Can filter and get suggestions of nodes from the platform and profile name and version in the compliance reporting search bar. 

1. `build components/compliance-service && build components/automate-gateway && build components/automate-ui && start_all_services`
1. `chef_load_compliance_nodes 10`
1. Go to https://a2-dev.test/compliance/reports/nodes
1. In the search bar select "Platform" and type one character. 
1. Ensure the suggestions have version numbers in them
1. Type all a platform's name and the first number of the version and then a "\*" and press enter. Example "ubuntu 12*"
1. Ensure only the proper nodes are shown. 
1. Do the same for the profile type in the search bar. 

- [x] Tests added/updated?
- [ ] Docs added/updated?


vinay033:
https://github.com/chef/automate/issues/744

lancewf:
@vinay033 Thank you for pointing this duplicate issue.

.expeditor/verify.pipeline.yml
```diff
@@ -751,3 +751,30 @@ steps:
     plugins:
       - gopath-checkout#v1.0.1:
           import: github.com/chef/automate
+
+  - label: "compliance-service upgrade from Automate v2v4"
```
lancewf: Adding a new buildkite test for the new migration.

.studio/chef-load
```diff
@@ -95,6 +95,25 @@ function chef_load_compliance_scans() {
...
+function chef_load_compliance_nodes() {
+  local nodes=${1:-50}
+
+  local total=$((nodes * 10))
+
+  echo "running: 'chef_load_compliance_scans -N$nodes -D10 -M1 -T$total'"
+
+  chef_load_compliance_scans -N$nodes -D10 -M1 -T$total
+}
+
```
lancewf: This will make it easier to add compliance nodes to Automate.

components/automate-cli/pkg/diagnostics/integration/compliance_report.go
```diff
@@ -228,7 +228,8 @@ func CreateComplianceReportDiagnostic() diagnostics.Diagnostic {
 
 					tries++
 					if tries >= maxTries {
-						assert.Failf(tstCtx, "POST %s failed with status code %d", reportReqPath, resp.StatusCode)
+						assert.Fail(tstCtx, "POST Failed", "POST %s failed with status code %d",
```
lancewf: This was not showing the `reportReqPath` and `StatusCode` before.

components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-overview/reporting-overview.component.spec.ts
```diff
@@ -305,7 +305,7 @@ describe('ReportingOverviewComponent', () => {
       component.onPlatformChanged({name: 'ubuntu'});
 
       expect(router.navigate).toHaveBeenCalledWith(['/compliance', 'reports', 'nodes'],
-        {queryParams: { platform: ['ubuntu'] }});
+        {queryParams: { platform_with_version: ['ubuntu*'] }});
```
lancewf: Created new filter types of `platform_with_version` and `profile_with_version` to not break the existing API requests.

components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-overview/reporting-overview.component.ts
```diff
@@ -118,7 +118,7 @@ export class ReportingOverviewComponent implements OnInit, OnDestroy {
 
   onPlatformChanged(platformItem) {
     if (platformItem && platformItem.name) {
-      this.updateUrlWithFilter(platformItem.name, 'platform');
+      this.updateUrlWithFilter(platformItem.name + '*', 'platform_with_version');
```
lancewf: When the platform bubble is selected filter with the `platform_with_version` type with a wildcard for all versions.

components/compliance-service/Makefile
```diff
@@ -224,6 +224,10 @@ start-es-pg: start-pg create-es-data-dir
 	  printf "\n===> ${yellow}Loading test A2v3 data into ElasticSearch${reset} ($$data_dir)...\n"; \
 	  tar -xzvf test_data/es-pg/es6-a2v3-data.tar.gz -C "$$data_dir"; \
 	fi; \
+	if [ "$(A2V4_DATA)" == "true" ]; then \
```
lancewf: Adding the new migration 'A2V4' to the makefile.

components/compliance-service/api/tests/60_A1_migration_spec.rb
```diff
@@ -31,7 +31,8 @@
           "name": "centos6-node1",
           "platform": {
             "name": "mac_os_x",
-            "release": "10.12.6"
+            "release": "10.12.6",
```
lancewf: All the ruby migration tests need to be updated because they all migrated from their version to the current version.

components/compliance-service/api/tests/64_A2V4_migration_spec.rb
```diff
@@ -0,0 +1,273 @@
+##### GRPC SETUP #####
+require 'api/reporting/reporting_pb'
+require 'api/reporting/reporting_services_pb'
+
+describe File.basename(__FILE__) do
+  Reporting = Chef::Automate::Domain::Compliance::Api::Reporting unless defined?(Reporting)
+
+  it "works" do
```
phiggins: I used the `it "works"` pattern because a lot of existing tests couldn't be broken up easily as later tests depended on data created in earlier tests. For a brand new test file it'd be better for readability and debugging to break these into individual tests that could fail independently.

components/compliance-service/api/tests/64_A2V4_migration_spec.rb
```diff
@@ -0,0 +1,273 @@
...
+    assert_equal('bb93e1b2-36d6-439e-ac70-cccccccccc07', res['id'])
+    assert_equal('9b9f4e51-b049-4b10-9555-10578916e149', res['node_id'])
+    assert_equal('centos(2)-beta-nginx-apache', res['node_name'])
+    assert_equal('DevSec Prod Beta', res['environment'])
+    assert_equal('failed', res['status'])
+    assert_equal(Google::Protobuf::Timestamp.new(seconds: 1520233322, nanos: 0), res['end_time'])
+    assert_equal(Reporting::Platform.new(name: "centos", release: "5.11"), res['platform'])
+    assert_equal(Reporting::Statistics.new(duration: 4.109065055847168), res['statistics'])
+    assert_equal(Google::Protobuf::RepeatedField, res['profiles'].class)
+    if res['profiles'].is_a?(Google::Protobuf::RepeatedField)
```
phiggins: This is probably a vestige from the way these tests were run previously. This conditional is unnecessary because the test on the line above will fail if this isn't true.

vjeffrey:
testing this locally on an "upgrade" system -- in other words, i already had some data loaded before i got on your branch and rebuild/loaded more data.
the platform stuff is all working great for me, but i find this behavior odd for the profile search/filter:
Screen Shot 2019-10-03 at 12 45 03

is that expected for the upgrade scenario?  or odd?

lancewf:
@vjeffrey Yeah, you are right. I am not filtering the profiles. For some reason, I was thinking that only the nodes got filtered. The profiles shown were all the ones from the displayed nodes. This is a released bug for wildcard filtering. 

Below is a profile wildcard search on https://a2-local-inplace-upgrade-acceptance.cd.chef.co
Screen Shot 2019-10-03 at 12 23 11 PM

vjeffrey:
oh interesting! yes, i see what you mean -- filtering by single profile still does the right thing, but filtering by wildcard profile just filters the nodes -- cool, that can be an issue/bug we work on separately then. would you mind opening the issue for that please?

components/compliance-service/api/tests/64_A2V4_migration_spec.rb
lancewf: This line is here to prevent the test from crashing on the lines inside of the 'if'. In golang, I would have a "require.Equals" for the test on line 245.

phiggins: That's exactly what I mean, if the assertion on line 245 fails, line 246 will not be executed. If the assertion on line 245 passes, the condition on line 246 will always be true.

lancewf: Oh, I thought that the test would keep running if an assert fails.
I will remove it then.

components/compliance-service/api/tests/64_A2V4_migration_spec.rb
lancewf: Thank you for pointing that out. I broke it up into many tests.

PR #1674, opened 2019-09-19
Jamie/fix ingest save rule button

SEAjamieD:
When creating an Ingest Rule inside a Project, after the user clicks on Save Button, a small loading icon is shown until complete.  Currently the text remains the same. 

In this branch, a small function was added to determine what the button text should be, depending on the state of the saving operation.  If the rule is in process of being saved, it will read as "Rule Saving..."

### :+1: Definition of Done
The Save Rule button reads as "Save Rule" by default, and changes to "Saving Rule..." when in the process of saving.

Visually
1. Build components/automate-ui-devproxy && start_all_services
1. Change your settings to slow down ur network, in chrome, dev tools > network > change online to slow 3g
1. Navigate to https://a2-dev.test/settings/projects
1. Create a project and click to enter that project
1. Click 'Create Rule'
1. Complete the steps to create the rule
1. Click "Save Rule"

Programmatically
Run the Spec file at Automate/components/automate-ui/src/app/pages/project/rules/project-rules.component.spec.ts

- [x] Tests added/updated?
- [ ] Docs added/updated?

Full Speed
SavingButtonsFull

Slow Speed
SavingRule2

### DIFF
--- a/components/automate-ui/src/app/pages/project/rules/project-rules.component.html
+++ b/components/automate-ui/src/app/pages/project/rules/project-rules.component.html
@@ -1,6 +1,6 @@
 <chef-page
   subheading="Resources that match the type and all conditions are included in the project."
-  [attr.confirm-btn-text]="getConfirmBtnText()"
+  [attr.confirm-btn-text]=" saving ? 'Saving Rule...' : 'Save Rule' "
   [attr.heading]="getHeading()"
   [attr.disable-confirm]="!ruleForm.valid || !ruleForm.dirty"
   [attr.page-loading]="isLoading"
--- a/components/automate-ui/src/app/pages/project/rules/project-rules.component.html
+++ b/components/automate-ui/src/app/pages/project/rules/project-rules.component.html
@@ -1,6 +1,6 @@
 <chef-page
   subheading="Resources that match the type and all conditions are included in the project."
-  confirm-btn-text="Save Rule"
+  [attr.confirm-btn-text]="getConfirmBtnText()"
   [attr.heading]="getHeading()"
   [attr.disable-confirm]="!ruleForm.valid || !ruleForm.dirty"
   [attr.page-loading]="isLoading"
--- a/components/automate-ui/src/app/pages/project/rules/project-rules.component.spec.ts
+++ b/components/automate-ui/src/app/pages/project/rules/project-rules.component.spec.ts
@@ -3,7 +3,6 @@ import { RouterTestingModule } from '@angular/router/testing';
 import { ReactiveFormsModule } from '@angular/forms';
 import { StoreModule } from '@ngrx/store';
 import { MockComponent } from 'ng2-mock-component';
-import { DebugElement } from '@angular/core';
 
 import { using } from 'app/testing/spec-helpers';
 import { ChefPipesModule } from 'app/pipes/chef-pipes.module';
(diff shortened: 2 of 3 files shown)


components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -1,6 +1,6 @@
 <chef-page
   subheading="Resources that match the type and all conditions are included in the project."
-  confirm-btn-text="Save Rule"
+  [attr.confirm-btn-text]="getConfirmBtnText()"
```
SEAjamieD: bind the attribute to the getConfirmBtnText function

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
```diff
@@ -290,6 +290,10 @@ export class ProjectRulesComponent implements OnInit, OnDestroy {
     }
   }
 
+  getConfirmBtnText(): string {
+    return this.saving ? 'Saving Rule...' : 'Save Rule';
+  }
+
```
SEAjamieD: Small ternary returning the desired text.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
msorens: A new pattern we started using just recently: when the function for generating a label is a simple expression (e.g. just 2 text phrases like this) then inline it to make it easier for the code reader. Thus:
```suggestion
  [attr.confirm-btn-text]="saving ? 'Saving Rule...' : 'Save Rule'"
```
Here's another example of its use: https://github.com/chef/automate/blob/f92742d7b5033b306ac2fb86cb6e75ea9bd0f60e/components/automate-ui/src/app/pages/project/confirm-apply-stop-modal/confirm-apply-stop-modal.component.html#L15

Of course, that means a little extra work for your tests: you will have to query the DOM.
But it is usually better to make opt for simpler code-under-test rather than simpler test.

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
msorens: Question for @susanev: For tokens, projects, and teams we use just "Save" and "Saving...". Is this one different for a reason?

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
SEAjamieD: Thank you for the specific examples to reference

components/automate-ui/src/app/pages/project/rules/project-rules.component.spec.ts
```diff
@@ -254,5 +255,17 @@ describe('ProjectRulesComponent', () => {
       component.addCondition();
       expect(component.showDelete()).toBeTruthy();
     });
+
+    fit('should show "Save Rule" Button by default', () => {
+      const componentDebug: DebugElement = fixture.debugElement.nativeElement;
+      const confirmButton = componentDebug.querySelector('#right-buttons button:first-child');
+      expect(confirmButton).toContain('Save Rule');
+    });
+
```
SEAjamieD: @msorens - I can't seem to find the correct way to query the DOM in my tests.  I've searched around and have mainly been following Angular's guide lines [on using debugElement](https://angular.io/guide/testing#debugelement), but I can seem to query inside the root element.  Any tips?

components/automate-ui/src/app/pages/project/rules/project-rules.component.ts
susanev: @msorens yes intentional since we are in this full-page design and trying to orient users into what they are saving.

components/automate-ui/src/app/pages/project/rules/project-rules.component.spec.ts
msorens: Ah, I suspect the wall you are hitting is Stencil; good chance that Stencil components are opaque to such probing. But you actually have what you need without penetrating that component. Try something like this:
```
      const element = fixture.nativeElement;
      expect(element.querySelector('chef-page')
        .getAttribute('confirm-btn-text')).toEqual('Save Rule');
```

SEAjamieD: Thank you @msorens - That works!  I'll look deeper into testing stencil components 👍

PR #1676, opened 2019-09-20
Print docker ps and free in buildkite

jaym:


.buildkite/hooks/pre-command
```diff
@@ -2,6 +2,8 @@
 
 set -eu
 
+docker ps || true
```
stevendanna: Wanna throw a free -m in here too?

PR #1681, opened 2019-09-20
[authz 1667] Use owner role for admin policy

blakestier:

We've added the `owner` role to the `admin` policy instead of using `actions: {*}`. Functionality is unchanged.

`start_all_services`
`rebuild components/authz-service`
`chef-automate dev psql chef_authz_service`
`select * from iam_statements where policy_id=policy_db_id('administrator-access') AND effect='allow';`

This statement should now have a role id attached that is the same as the owner role.

- [ ] Tests added/updated?
- [x] Docs added/updated?


susanev:
theres a screenshot of the admin policy that we should prolly update https://automate.chef.io/docs/iam-v2-overview/#policies-and-roles-in-automate

PR #1687, opened 2019-09-23
Resync query_projects function

msorens:
## :nut_and_bolt: Description: What code changed, and why?
Mirror recent changes to `query_project` (singular) back to `query_projects` (plural) for consistency, (while still fresh).

NA

### :+1: Definition of Done
Everything should still work.

### DIFF
--- a/components/authz-service/storage/postgres/migration/sql/71_sync_query_projects_with_query_project.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/71_sync_query_projects_with_query_project.up.sql
@@ -1,0 +1,39 @@
+BEGIN;
+
+CREATE OR REPLACE FUNCTION
+  query_projects(_projects_filter TEXT[])
+  RETURNS setof json AS $$
+
+WITH
+staged AS (
+  SELECT r.project_id
+  FROM iam_staged_project_rules AS r
+  WHERE projects_match_for_rule(r.project_id, _projects_filter)
+  GROUP BY r.project_id
+),
+applied AS (
+  SELECT r.project_id
+  FROM iam_project_rules AS r
+  WHERE projects_match_for_rule(r.project_id, _projects_filter)
+  GROUP BY r.project_id
+)
+SELECT json_build_object(
+  'id', p.id,
+  'name', p.name,
+  'type', p.type,
+  'status',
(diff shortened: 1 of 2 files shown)


components/authz-service/storage/postgres/migration/sql/71_sync_query_projects_with_query_project.up.sql
```diff
@@ -0,0 +1,39 @@
+BEGIN;
+
+CREATE OR REPLACE FUNCTION
+  query_projects(_projects_filter TEXT[])
+  RETURNS setof json AS $$
+
+WITH
+staged AS (
+  SELECT r.project_id
```
msorens: This line (and L15 similarly) previously read
```
  SELECT r.project_id, count(r.id) AS count
```
The `count` was actually not needed (thanks @blakestier !) so was removed when `query_project` (singular) was created; just mirroring that change here.

components/authz-service/storage/postgres/migration/sql/71_sync_query_projects_with_query_project.up.sql
```diff
@@ -0,0 +1,39 @@
...
+  GROUP BY r.project_id
+)
+SELECT json_build_object(
+  'id', p.id,
+  'name', p.name,
+  'type', p.type,
+  'status',
+  CASE WHEN staged.project_id IS NULL AND applied.project_id IS NULL THEN 'no-rules'
+       WHEN staged.project_id IS NULL THEN 'applied'
+       ELSE 'edits-pending'
```
msorens: The above `case` statement had `count` globally changed to `project_id`.

components/authz-service/storage/postgres/migration/sql/71_sync_query_projects_with_query_project.up.sql
```diff
@@ -0,0 +1,39 @@
...
+  END
+)
+FROM iam_projects AS p
+LEFT JOIN applied
+  ON applied.project_id = p.db_id
+LEFT JOIN staged
+  ON staged.project_id = p.db_id
```
msorens: The block above is a bit cleaner than its predecessor:
```
FROM staged
    FULL JOIN applied
    ON staged.project_id = applied.project_id
RIGHT JOIN iam_projects p
    ON p.db_id = COALESCE(staged.project_id, applied.project_id)
```

Issue #1311, opened 2019-08-21
compliance reporting ui: search by control tags


vjeffrey:
# User Story
As an Automate user, I want to be able to filter my nodes by control *tags*
Why?
InSpec provides a mechanism that can be used for 'grouping' controls across profiles (tags)
Users would like to filter their set of nodes being scanned in Automate by selecting a control tag, so they can narrow their query to a group of controls across several profiles.

## Considerations
backend work is in progress: https://github.com/chef/automate/issues/519 
the work for this card will include extending the search bar to include control tag filter option and properly handling that filter for communication with the api

## Designs
**Filter Functionality**
Follow the same pattern as the Events Page Search component when choosing Event Type, but expanding to 3rd tier drop down menu.

**Step 1:** Default state of search box
![Screen Shot 2019-08-21 at 11 06 32 AM](https://user-images.githubusercontent.com/1316587/63456814-9ded5200-c404-11e9-917b-439726f2d86c.png)

**Step2:** User clicks or tabs to search box, category drop down displays.
![Screen Shot 2019-09-11 at 11 09 20 AM](https://user-images.githubusercontent.com/1316587/64723093-ae816d00-d484-11e9-8e9a-1fbd09db94fa.png)

**Step 3:** User chooses Tag category, system then prompts user to choose Tag Key from dropdown
![Screen Shot 2019-09-11 at 11 13 20 AM](https://user-images.githubusercontent.com/1316587/64723364-41baa280-d485-11e9-973d-fd759e01c94e.png)

**Step 4:** System then shows category|tag Name label in search box, and then prompts user to enter in a Tag Key value.
![Screen Shot 2019-09-11 at 11 16 06 AM](https://user-images.githubusercontent.com/1316587/64723622-b7267300-d485-11e9-921e-3f9350e76727.png)

**Step 5:** user can then type in a value with/without wild card, and system can suggest values.
![Screen Shot 2019-09-11 at 11 22 31 AM](https://user-images.githubusercontent.com/1316587/64724026-7844ed00-d486-11e9-957d-956c8fd19be3.png)

**Step 6:** after user enters in desired value, system displays filter criteria tag below.
![Screen Shot 2019-09-11 at 11 21 42 AM](https://user-images.githubusercontent.com/1316587/64723978-64998680-d486-11e9-9e17-bd19d1d024f6.png)

# Definition of Done
Can search (get suggestions and filter on) control tags in Automate

alexpop:
[...]

PR #1694, opened 2019-09-23
Add control_tag filters to reports search bar 

tarablack01:
InSpec provides a mechanism that can be used for 'grouping' controls across profiles (tags)
Users would like to filter their set of nodes being scanned in Automate by selecting a control tag, so they can narrow their query to a group of controls across several profiles.

https://github.com/chef/automate/issues/1311

### :+1: Definition of Done
Can search (get suggestions and filter on) control tags in Automate

hab studio enter
start_automate_ui_background
start_all_services
chef_load_compliance_scans -N20

- [x] Tests added/updated?
- [x] Docs added/updated?

compliance-reports-control-tag-filter

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.ts
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.ts
@@ -468,10 +468,10 @@ export class ReportingSearchbarComponent implements OnInit {
   // When value is clicked for the control tag key
   onControlTagKeyValueClick(value: any) {
     const reportQuery = this.reportQuery.getReportQuery();
-    const type = { ...this.selectedType,
-      name: `control_tag:${value.text}`,
-      title: `Control Tag | ${value.text}`,
-      placeholder: 'Enter Control Tag Values...'};
+    const type = this.selectedType;
+    type.name = `control_tag:${value.text}`;
+    type.title = `Control Tag | ${value.text}`;
+    type.placeholder = 'Enter Control Tag Values...';
     reportQuery.filters.push({
       type: type,
       value: {text: undefined }
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.ts
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.ts
@@ -200,13 +200,18 @@ export class ReportingSearchbarComponent implements OnInit {
     if (this.highlightedIndex >= 0) {
       const search = this.filterValues[this.highlightedIndex];
       const type = this.selectedType;
-      this.ClearAll();
-      this.filterAdded.emit({
-        detail: {
-          value: search,
-          type: type
-        }
-      });
+      if (type.name === 'control_tag_key') {
+        this.onControlTagKeyValueClick(search);
+      } else {
+        this.ClearAll();
+        this.suggestionsVisible = false;
+        this.filterAdded.emit({
+          detail: {
+            value: search,
+            type: type
+          }
+        });
+      }
     } else {
       if (currentText.indexOf('?') >= 0 || currentText.indexOf('*') >= 0) {
         const type = this.selectedType;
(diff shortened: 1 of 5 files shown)


vjeffrey:
another option for testing is:
```
cd automate/components/compliance-service/test_data/audit_reports/
 ./send_to_data_collector.sh https://my-automate.example.com TOKEN
```

jonong1972:
I am getting different interactions when using keyboard vs. mouse.  

Repro on keboard:
1. click into the input field
2. Arrow down to the Control Tags
3. press `return`
4. Arrow down to a tag
5. press `return`

It initiates the search with the tag without the option to choose a key value.

It works great with the mouse though!

jonong1972:
Oh, also this opens up another functionality issue.  

Sometimes Keys do not have values. And so when a user selects a Key without any value, hopefully there will be logic to detect that and initiate the filter from there.

I will create a separate ticket for that.

jonong1972:
Ticket created for the accommodation of Key Pair Values without any values
https://github.com/chef/automate/issues/1713

tarablack01:
@jonong1972 Updated the enter workflow
control_tag-enter

components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.ts
```diff
@@ -14,6 +14,9 @@ import {
   debounceTime, switchMap, distinctUntilChanged
 } from 'rxjs/operators';
 import { FilterC } from '../types';
+import {
+  ReportQueryService
+} from '../../shared/reporting';
```
msorens: Minor point: our loose convention is to avoid `..` in pathing of imports. Things should either start with `app/...` or `./` (single dot for stuff in the current directory).

components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.ts
```diff
@@ -432,7 +441,10 @@ export class ReportingSearchbarComponent implements OnInit {
   }
 
   requestForSuggestions(c: any): void {
```
msorens: Please consider adding some strong typing to cut down on the use of `any` in this file. For this one, for example: define this type at the top of the file...
```
interface SuggestionEntry {
  text: string;
  type: string;
}
```
Then you can do this...
```suggestion
  requestForSuggestions(c: SuggestionEntry): void {
```
... so we know more clearly that `c.type` is a valid thing. (L444).

components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.ts
```diff
@@ -446,4 +458,30 @@ export class ReportingSearchbarComponent implements OnInit {
...
+    return type.search('control_tag:') === 0 ? 'control_tag_value' : type;
+  }
+
+  // When value is clicked for the control tag key
+  onControlTagKeyValueClick(value: any) {
+    const reportQuery = this.reportQuery.getReportQuery();
+    const type = this.selectedType;
+    type.name = `control_tag:${value.text}`;
+    type.title = `Control Tag | ${value.text}`;
+    type.placeholder = 'Enter Control Tag Values...';
```
msorens: Consider this slightly more encapsulated variation that shows the intent a bit more clearly:
```
const type = { ...this.selectedType,
  name: `control_tag:${value.text}`,
  title: `Control Tag | ${value.text}`,
  placeholder: 'Enter Control Tag Values...'};
```

components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.ts
tarablack01: Agreed, Updated!

components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.ts
tarablack01: Added type

components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.ts
tarablack01: @msorens good suggestion 👍🏻

PR #1638, opened 2019-09-17
Fixing wrong trend graph date with time zone + UTC

lancewf:
### :nut_and_bolt: Description: What code changed, and why?
The problem this PR is trying to fix is, when a user has their local timezone set to any time larger that UTC the trend graph shifts each column one day ahead. This can be seen better in the screenshot below. 

This problem was caused by a timezone shift. The data sent from the API has the dates like this '2019-09-14T23:59:59Z'. When this time was parse into a date object it would be '2019-09-15T00:59:59GMT+01:00'. Therefore the 15th would be displayed and not the 14th. 

### :chains: Related Resources

Similar timezone problems. 
https://github.com/chef/automate/pull/1624

### :+1: Definition of Done
When a user has local timezone UTC +1 the trend graph does not shift the columns one day ahead. 

### :athletic_shoe: How to Build and Test the Change
1. Switch your local computer time to UTC +1 (London)
1. `build components/automate-ui && start_all_services`
1. Add some compliance nodes with `chef_load_compliance_scans -N100 -D10 -M1 -T1000`
1. Open https://a2-dev.test/compliance/reports/overview
1. Ensure the date selected on the calendar matches the last trend graph date.

### :white_check_mark: Checklist

- [x] Tests added/updated?
- [ ] Docs added/updated?

### :camera: Screenshots, if applicable
The below image shows the error with the local time set to UTC+1
trend_graph_bad_date

PR #1696, opened 2019-09-23
Passing date as a string

lancewf:
Fix a bug where the calendar selects the day before. 

The two PR that caused the problem
https://github.com/chef/automate/pull/1638
https://github.com/chef/automate/pull/1610

1. Open https://a2-dev.test/compliance/reports/overview
1. Test that the calendar selects the date that is clicked. 

- [x] Tests added/updated?
- [ ] Docs added/updated?

Below is showing the bug. When the 24th is clicked the 23 will be selected. 
calendar

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.ts
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.ts
@@ -405,9 +405,9 @@ export class ReportingSearchbarComponent implements OnInit {
     this.visibleDate.setFullYear(year);
   }
 
-  onDaySelect(date) {
+  onDaySelect(date: string) {
     this.visibleDate = new Date(date);
-    this.dateChanged.emit({ detail: new Date(this.visibleDate) });
+    this.dateChanged.emit({ detail: date });
   }
 
   handleFocusOut() {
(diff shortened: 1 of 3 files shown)


components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.ts
```diff
@@ -405,9 +405,9 @@ export class ReportingSearchbarComponent implements OnInit {
     this.visibleDate.setFullYear(year);
   }
 
-  onDaySelect(date) {
+  onDaySelect(date: string) {
     this.visibleDate = new Date(date);
-    this.dateChanged.emit({ detail: new Date(this.visibleDate) });
+    this.dateChanged.emit({ detail: date });
```
lancewf: Sending the string '2019-10-23' instead of the date object which will be in local timezone.

components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.ts
```diff
@@ -285,7 +285,7 @@ export class ReportingComponent implements OnInit, OnDestroy {
   }
 
   onEndDateChanged(event) {
-    const endDate = event.detail;
+    const endDate: string = event.detail;
```
lancewf: The `endDate` will be a string like '2019-10-23'

vjeffrey:
oh man, i think i had this pr open and didn't see the comment from michael/didn't refresh the page:
```
Looks like boundary conditions still have a problem (that's why I like that using construct a lot 😁 ). Clicking on almost any date works fine... unless you click on the 1st of a month. It does display the selected date up in the date button (if that's what you call it) but it also switches the calendar to the previous month. (Lacking context I thought it might have been desired behavior, but then I would have expected picking the last day of the month would do a similar action and it does not.)
```
let's address that with a separate issue - sorry @msorens !

lancewf:
@msorens Thank you for reviewing. I will create a PR to fix that problem also.

PR #1701, opened 2019-09-24
Restore with many indices

jaym:
Restores seem to fail if you have too many indices. This is an attempt to find alternatives / skip calls that put indexes in the URL

### DIFF
--- a/components/automate-cli/pkg/diagnostics/option.go
+++ b/components/automate-cli/pkg/diagnostics/option.go
@@ -39,14 +39,14 @@ func OptionFromString(option string) (*Option, error) {
 	}, nil
 }
 
-func (o *Option) GetString(defaultVal string) string {
+func (o *Option) AsString(defaultVal string) string {
 	if o == nil || o.Value == "" {
 		return defaultVal
 	}
 	return o.Value
 }
 
-func (o *Option) GetInt(defaultVal int) int {
+func (o *Option) AsInt(defaultVal int) int {
 	if o == nil || o.Value == "" {
 		return defaultVal
 	}
--- a/components/automate-cli/pkg/diagnostics/option.go
+++ b/components/automate-cli/pkg/diagnostics/option.go
@@ -1,0 +1,75 @@
+package diagnostics
+
+import (
+	"fmt"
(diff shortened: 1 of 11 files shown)


components/automate-cli/pkg/diagnostics/option.go
```diff
@@ -0,0 +1,75 @@
...
+	if len(parts) == 0 {
+		return nil, ErrInvalidOption
+	}
+	return &Option{
+		Key:   parts[0],
+		Value: parts[1],
+	}, nil
+}
+
+func (o *Option) GetString(defaultVal string) string {
```
ryancragun: I don't think I see this used anywhere, but that's probably fine.

components/automate-cli/pkg/diagnostics/option.go
```diff
@@ -0,0 +1,75 @@
...
+	if o == nil || o.Value == "" {
+		return defaultVal
+	}
+	return o.Value
+}
+
+func (o *Option) GetInt(defaultVal int) int {
```
ryancragun: [nit] I don't think the `GetInt()` and `GetString()` method names make it very clear what's happening here, which is that we want the option option value as either a string or int, if it's set, otherwise we want the default that's given. The best I can summarize that with some brevity is `ValAsIntOr` and `ValAsStrOr`, not the best, but maybe a starting point.

components/automate-cli/pkg/diagnostics/option.go
```diff
@@ -0,0 +1,75 @@
...
+	if o == nil || o.Value == "" {
+		return defaultVal
+	}
+	i, err := strconv.Atoi(o.Value)
+	if err != nil {
+		logrus.WithError(err).Warnf("could not parse int (%s)", o.Value)
```
ryancragun: Not terribly in love with a method that can return an error swallowing it instead of returning it.

components/automate-cli/pkg/diagnostics/option.go
ryancragun: Could also see `ValueOr` and `ValueStrOr`

PR #1709, opened 2019-09-25
success message for creating, updating and deleting the scan job

### LINKED ISSUE #237: ui: add a success/failure message after deletion of a scan job.
# User Story

As a Automate user, I want to be informed that my deletion of a scan job was successful.

# Definition of Done
After the user deletes a scan job, show a success/fail message (use existing pattern) at the top of the page.

# Demo Script / Repro Steps
Scan Job Deletion
1. Goto Scan Jobs `/compliance/scanner/jobs`
2. Click the "delete icon" on a scan job
3. confirm the deletion with message. "Deleted a scan job."

Scan Job Creation
1. Goto Scan Jobs `/compliance/scanner/jobs`
2. Create a new scan job
3. confirm the deletion with message. "Created a scan job."

**Successfully Created**
Screen Shot 2019-09-16 at 3 17 04 PM

**Succesfully Deleted**
Screen Shot 2019-09-16 at 3 20 40 PM

**Interaction:**
Same components and interaction as creating/deleting a team:
https://a2-dev.test/settings/teams

Current:
If there is a long list of Scan jobs, and the job deleted was below the fold, it is hard to determine if it was deleted or not. Also, a confirmation message would be nice.

vinay033:
…can job

As an Automate user, I want to be informed that my creation, updation, and deletion of a scan job was successful.

fixes #237

### :+1: Definition of Done
After the user deletes a scan job, show a success/fail message (use existing pattern) at the top of the page.
### :camera: Screenshots
Screenshot from 2019-09-25 19-22-12
Screenshot from 2019-09-25 19-22-50
Screenshot from 2019-09-25 19-22-57


components/automate-ui/src/app/entities/jobs/job.effects.ts
```diff
@@ -37,9 +39,25 @@ export class JobEffects {
     mergeMap((action: JobActions) => this.requests.jobCreate(action.payload)),
     map((payload: {id: string}) => new JobCreateSuccess(payload)));
 
+  @Effect()
+  JobCreateSuccess$ = this.actions$.pipe(
+    ofType<JobCreateSuccess>(JobActionTypes.JOB_CREATE_SUCCESS),
+    map(() => new CreateNotification({
+      type: Type.info,
+      message: 'Created a scan job.'
```
msorens: I believe this wording should be changed here, and similarly for updated and deleted below.
@jonong1972 can you confirm?
```suggestion
      message: 'Scan job created.'
```

components/automate-ui/src/app/entities/jobs/job.effects.ts
```diff
@@ -37,9 +39,25 @@ export class JobEffects {
...
+    })));
+
   @Effect()
   jobUpdate$ = this.actions$.pipe(
     ofType(JobActionTypes.JOB_UPDATE),
     mergeMap((action: JobActions) => this.requests.jobUpdate(action.payload)),
     map(payload => new JobUpdateSuccess(payload)));
+
+  @Effect()
+  JobUpdateSuccess$ = this.actions$.pipe(
```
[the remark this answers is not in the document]
vinay033: sure @msorens I will make these changes.

components/automate-ui/src/app/entities/jobs/job.effects.ts
jonong1972: Sweet, I wasnt sure if we can pull that. But would like this. @vinay033 can update this?

jonong1972:
As @msorens mentioned, is it possible to add the name of the object being created/updated/deleted in the messaging as he suggested?

components/automate-ui/src/app/entities/jobs/job.effects.ts
vinay033: sure, I will update.

vinay033:
> As @msorens mentioned, is it possible to add the name of the object being created/updated/deleted in the messaging as he suggested?

yes, I will change it.

vinay033:
> As @msorens mentioned, is it possible to add the name of the object being created/updated/deleted in the messaging as he suggested?

@msorens  and @jonong1972  
for display the `object` name in the notification message, it is only possible for a create scan job, it is not possible with an update and deletes. 
I have added changes in compliance services to getting a name in scan job API response(create scan job API), previously it was only returning the ID like `{"id":"ec2e890f-f58c-4ca2-95de-bd474f97b86a"}`.

@vj also helped to me for automate-gateway changes because scan job API changes depend on automate gateway.
I also tried with some changes in `update` and `delete` scan job API to get the scan job name because currently both returning the empty response `{}` but if we go with the new changes then it will breaking some other stuff. we can’t change the response message type in Golang functions (compliance-service, automate-gateway).

msorens:
@vinay033 	Your notes sound reasonable; if you can only add it easily for "create", I think that is sufficient for now.

vinay033:
> @vinay033 Your notes sound reasonable; if you can only add it easily for "create", I think that is sufficient for now.

@msorens , I added changes for `create` in previous commits, please take a look :)

vinay033:
> console.log(payload);

@msorens , changes are working fine, before creating the `scan job` first run the below command.
```
go_update_component compliance-service
go_update_component automate-gateway
```
inside the hab studio.

components/compliance-service/dao/pgdb/jobs.go
```diff
@@ -604,8 +604,9 @@ func (db *DB) AddJob(inJob *jobs.Job) (string, error) {
 
 		return nil
 	})
+	logrus.Debugf("printtttt the id and name %s %s", job.ID, job.Name)
```
stevendanna: [nit] Looks like development logging we can likely remove

components/compliance-service/api/jobs/server/server.go
```diff
@@ -106,12 +106,13 @@ func (srv *Server) Create(ctx context.Context, in *jobs.Job) (*jobs.Id, error) {
 		return nil, status.Error(codes.InvalidArgument, "Invalid job: nodes or node selectors required.")
 	}
 
-	sID, err := srv.db.AddJob(in)
+	sID, name, err := srv.db.AddJob(in)
```
stevendanna: I'm a little confused about why we need to change the signature of AddJob.  `in` is a `jobs.Job` and should already have the Name assigned to it. I don't see anything in AddJob that mutates that name.

I can see some value in being defensive here since AddJob mutates `Id` we should indeed maybe assume it mutates everything. But, in that case, I think one of the following signatures for AddJob might be preferable:

1) `AddJob(inJob jobs.Job) (jobs.Job, error)` (i.e take the input by value return the an entire job object)
2) `AddJob(inJob *jobs.Job) error` (i.e. take the input by reference, assume AddJob mutates it appropriately)

If I'm right about not actually needing this `name` return, it might be best to leave AddJob as-is for now to reduce the scope of this change.

vinay033: @stevendanna , I have made changes for this, please have a look.

components/automate-ui/src/app/entities/jobs/job.effects.ts
```diff
@@ -35,11 +37,27 @@ export class JobEffects {
...
     mergeMap((action: JobActions) => this.requests.jobCreate(action.payload)),
-    map((payload: {id: string}) => new JobCreateSuccess(payload)));
+    map((payload: { id: string, name: string }) => new JobCreateSuccess(payload)));
+
+  @Effect()
+  jobCreateSuccess$ = this.actions$.pipe(
+    ofType<JobCreateSuccess>(JobActionTypes.JOB_CREATE_SUCCESS),
+    map((payload) => new CreateNotification({
+      type: Type.info,
+      message: `Created a scan job ${payload.payload.name}.`
```
msorens:
```suggestion
      message: `Created scan job ${payload.payload.name}.`
```

vinay033: I have updated the changes.

PR #1639, opened 2019-09-17
Jamie/natural sort library function

SEAjamieD:
### :nut_and_bolt: Description: What code changed, and why?
It was determined that it might be useful if we created a helper function for Normal Sorting, which is what we use in our projects-dropdown component.  This could then be reused across our codebase where necessary.

A sorter file was created in app/helpers/auth with a ChefSorters class to hold any future sorting functions we might want to build.

the internals of the projectsArray function in the projects-dropdown component were replaced to use the ChefSorters.normalSort function.

Functions updated in:
- [x] project-dropdown.component.ts
- [x] api-token-list.component.js
- [x] policy-add-members.component.ts
- [x] policy-list.component.js
- [x] project-list.component.js
- [x] projects-filter.reducer.ts
- [x] roles-list.component.js
- [x] team-add-users.component.js
- [x] team-details.component.js
- [x] team-management.component.js
- [x] user-management.component.js

### :chains: Related Resources
https://github.com/chef/automate/pull/1620

### :+1: Definition of Done
A more universal version of a Normal Sorting function is available to use across the codebase.

### :athletic_shoe: How to Build and Test the Change

Visually -
1. build components/automate-ui-devproxy && start_all_services
2. Navigate to https://a2-dev.test/settings/tokens
3. Create an API Token
4. Navigate to https://a2-dev.test/settings/projects
5. Create several projects
6. Navigate to https://a2-dev.test/settings/tokens/ your-api-token
7. Click on the projects dropdown to see your projects sorted normally.

Testing -
1. run the specfile at app/helpers/auth/sorter.spec.ts

### :white_check_mark: Checklist

- [x] Tests added/updated?
- [ ] Docs added/updated?

### :camera: Screenshots, if applicable
Screen Shot 2019-09-17 at 3 46 26 PM

PR #1620, opened 2019-09-16
Projects in dropdown should be Normally sorted

SEAjamieD:
### :nut_and_bolt: Description: What code changed, and why?
This change updates the default sorting of the projects dropdown component so that when a user views their projects in a dropdown, they see them sorted Normally.  This matches our tables as well as the sorting we use in some of our other components.

### :chains: Related Resources
[[A2-338] fix admin UI list sorting](https://github.com/chef/a2/pull/4434)

### :+1: Definition of Done
The dropdown is sorted Normally.  First by numbers, and then Alphabetically.

### :athletic_shoe: How to Build and Test the Change
Visually -
1. `build components/automate-ui-devproxy && start_all_services`
2. Navigate to https://a2-dev.test/settings/tokens
3. Create an API Token
4. Navigate to https://a2-dev.test/settings/projects
5. Create several projects
6. Navigate to https://a2-dev.test/settings/tokens/ your-api-token
7. Click on the projects dropdown to see your projects sorted normally.

Using the specfile
1. Run the spec file at src/app/components/projects-dropdown/projects-dropdown.component.spec.ts

### :white_check_mark: Checklist

- [x] Tests added/updated?
- [ ] Docs added/updated?

### :camera: Screenshots, if applicable

After: Dropdown is sorted normally.

Screen Shot 2019-09-16 at 10 37 41 AM

PR #1715, opened 2019-09-25
Jamie/natural sort library function 2

SEAjamieD:
It was determined that it might be useful if we created a helper function for Normal Sorting, which is what we use in our projects-dropdown component.  This could then be reused across our codebase where necessary.

A sorter file was created in app/helpers/auth with a ChefSorters class to hold any future sorting functions we might want to build.

Components refactored to use ChefSorters helper class:
- [x] project-dropdown.component.ts
- [x] api-token-list.component.js
- [x] policy-add-members.component.ts
- [x] policy-list.component.js
- [x] project-list.component.js
- [x] projects-filter.reducer.ts
- [x] roles-list.component.js
- [x] team-add-users.component.js
- [x] team-details.component.js
- [x] team-management.component.js
- [x] user-management.component.js

https://github.com/chef/automate/pull/1620
[First PR where rebase mistake was made](https://github.com/chef/automate/pull/1639)

### :+1: Definition of Done
A more universal version of a Normal Sorting function is available to use across the codebase, and all components that current use natural sorting in this manner are updated.

Visually -
2. Navigate to https://a2-dev.test/settings/tokens/ <your-api-token>
3. Click on the projects dropdown to see your projects sorted normally.
4. Navigate to https://a2-dev.test/settings/tokens --> View tokens sorted by Name
5. Navigate to https://a2-dev.test/settings/policies --> View policy list sorted by Name
6. From policies, click on Editors then Members Tab.  click Add Member --> View Members list sorted by Display Name first, then Name if possible.
7.  Navigate to https://a2-dev.test/settings/projects --> View projects sorted by Name
8. Navigate to https://a2-dev.test/dashboards/event-feed. click on "All Projects" in top right corner --> view Projects sorted by name, with (Unassigned) as last.
9. Navigate to https://a2-dev.test/settings/roles --> view roles list sorted by Name
11. Navigate to https://a2-dev.test/settings/teams/admins --> see list sorted by Name first, then by id if necessary.
10. Click "Add User". --> View list sorted by Name first, then by id if necessary.
11. Navigate to https://a2-dev.test/settings/teams ---> view teams sorted by ID
12. Navigate to https://a2-dev.test/settings/users ---> view users list sorted by Name first, then by id if necessary.

Testing -
1. run the specfile at app/helpers/auth/sorter.spec.ts

- [x] Tests added/updated?
- [ ] Docs added/updated?

Screen Shot 2019-09-17 at 3 46 26 PM


components/automate-ui/src/app/helpers/auth/sorter.spec.ts
```diff
@@ -0,0 +1,125 @@
+import { ChefSorters } from './sorter';
+
+describe('naturalSort', () => {
```
SEAjamieD: I moved a whole slew of tests over from various components that now use ChefSorter so that the same style of test is being run in the sorter.spec file.  I didn't feel comfortable removing the similar tests from the other files, because many of those were using ChefSorter _inside_ of another function and it seemed like a good idea that those functions are still tested in the same manner.

https://github.com/chef/automate/blob/693c7f1be31d9df6bdbe46ba1c2a369e9af20983/components/automate-ui/src/app/pages/api-token/list/api-token-list.component.spec.ts#L96-L119

components/automate-ui/src/app/helpers/auth/sorter.ts
```diff
@@ -0,0 +1,35 @@
...
+                a[propName].localeCompare(b[propName], undefined, { numeric: true });
+        };
+
+        const dualPropComparator = (propOne, propTwo) => {
+            return (a, b) => a[propOne].localeCompare(b[propOne], undefined, opts) ||
+                a[propOne].localeCompare(b[propOne], undefined, { numeric: true }) ||
+                a[propTwo].localeCompare(b[propTwo], undefined, opts);
+        };
+
+        return property instanceof Array
```
SEAjamieD: If the property passed in is an Array, we know that the user expects a fallback comparison, so we'll use the dualPropComparator

components/automate-ui/src/app/helpers/auth/sorter.ts
```diff
@@ -0,0 +1,36 @@
+export class ChefSorters {
+    // ***********
+    // naturalSort takes in an Array of Objects and returns an Array of Objects
+    // sorted 'naturally' by the property chosen.
+    // @input: Array of Objects
+    // @property: string or an Array of strings, max length is 2 properties
+    //            when using an Array, will sort first by string one, then by string 2;
```
msorens:
```suggestion
    //            when using an Array, sort by first property then by second property
```

components/automate-ui/src/app/helpers/auth/sorter.ts
```diff
@@ -0,0 +1,36 @@
...
+        // https://stackoverflow.com/questions
+        const propComparator = (propName) => {
+            return (a, b) => a[propName].localeCompare(b[propName], undefined, opts) ||
+                a[propName].localeCompare(b[propName], undefined, { numeric: true });
+        };
+
+        const dualPropComparator = (propOne, propTwo) => {
+            return (a, b) => a[propOne].localeCompare(b[propOne], undefined, opts) ||
+                a[propOne].localeCompare(b[propOne], undefined, { numeric: true }) ||
+                a[propTwo].localeCompare(b[propTwo], undefined, opts);
```
msorens: For good measure should probably introduce symmetry, so `propTwo` has the additional numeric-only check last.

components/automate-ui/src/app/helpers/auth/sorter.spec.ts
msorens: On your concern about being inside other functions: good observation! But that was simply necessary boilerplate needed in order to get to test the thing we wanted to test. 

I think it would be beneficial to remove all those tests for a few reasons.
1. It was actually a code smell that all of those identical tests were propagated throughout the code base. Now, you've got them consolidated in a single place.
2. Testing this function that is now external to those other classes is somewhat outside the purview of those other classes to test, in the spirit of unit testing.
3. What if down the road we modify sorter.ts? One would then update sorter.spec.ts, but then one would likely need to update tests in all those other classes, too.

SEAjamieD: That makes sense 👍 thanks for the explanation.  I've gone ahead and removed any tests only testing the sorting ability.

components/automate-ui/src/app/helpers/auth/sorter.ts
```diff
@@ -0,0 +1,39 @@
...
+    // @input: Array of Objects
+    // @property: string or an Array of strings, max length is 2 properties
+    //            when using an Array, sort by first property then by second property
+    //
+    // example use with single prop:  ChefSorters.naturalSort(users, 'name');
+    // example use with dual prop:    ChefSorters.naturalSort(users, ['name', 'id']);
+    // ***********
+
+    // tslint:disable-next-line: max-line-length
+    public static naturalSort(input: Array<any>, property: Array<string> | string): Array<any> {
```
msorens: Prefer to use `tslint` decorators only when it is a true burden; here you can just add a carriage return, e.g.:
```suggestion
    public static naturalSort(
      input: Array<any>, property: Array<string> | string): Array<any> {
```

PR #1730, opened 2019-09-27
Fix external automate trusted cert for cs nginx

### LINKED ISSUE #1719: Adding an external automate trusted certificate breaks automate-cs-nginx
## Describe the bug
If you add a trusted ssl certificate for chef-server infra on a2 to use to trust a data-collector endpoint it will put automate-cs-nginx into a bootloop.

The core of the issue is very straightforward, the template for `automate-cs-nginx/config/external_automate_root_ca.crt` contains `{{cfg.external_automate.ssl.root_cert}}`. However the user.toml for this service gets the following key when the patch is applied.

```
[external_automate]
  root_cert = "-----BEGIN CERTIFICATE...
```
Note that ssl is not part of the path that was put in the user.toml, but is in the template. Because of this the template comes up blank and nginx goes into a restart loop with an ssl error on the blank file.

automate-cs-oc-erchef works on the same config, but does have the ssl field and works fine.

## To Reproduce
On an automate server with chef infra installed, create and apply patch.toml like so, with a self signed cert.
```
      [global.v1.external.automate]
        enable = true
        node = "https://automate-fqdn.example"
        [global.v1.external.automate.auth]
          token = "93a49a4f2482c64126f7b6015e6b0f30284287ee4054ff8807fb63d9cbd1c506"
        [global.v1.external.automate.ssl]
          root_cert = """
-----BEGIN CERTIFICATE-----
YOURCERTHERE
-----END CERTIFICATE-----
"""
After applying, check for these lines in your log.
automate-cs-nginx.default(O): 2019/09/26 19:18:17 [emerg] 4870#0: SSL_CTX_load_verify_locations("/hab/svc/automate-cs-nginx/config/external_automate_root_ca.crt") failed (SSL:)

## Expected behavior
External trusted certs for data collector should work


Version: 2
CLI Build: 20190903214656
Server Build: 20190904132002

### PR DESCRIPTION
**Pull request by jaym**
Fixes #1719

### DIFF
--- a/components/automate-cs-nginx/habitat/config/external_automate_root_ca.crt
+++ b/components/automate-cs-nginx/habitat/config/external_automate_root_ca.crt
@@ -1,1 +1,1 @@
-{{cfg.external_automate.ssl.root_cert}}
+{{cfg.external_automate.root_cert}}

### CODE REVIEW

**Comment by stevendanna** [comment]
Looks like we might have a few more to fix:

```
 > rg 'external_automate.*root_cert'
components/automate-cs-oc-erchef/habitat/config/external_automate_root_ca.crt
1:{{cfg.external_automate.ssl.root_cert}}

components/automate-cs-oc-erchef/habitat/config/sys.config
208:                                                  {{#if cfg.external_automate.ssl.root_cert}}

components/automate-cs-nginx/habitat/config/external_automate_root_ca.crt
1:{{cfg.external_automate.ssl.root_cert}}

components/automate-cs-nginx/habitat/config/chef_https.conf
92:      {{~#if cfg.external_automate.root_cert }}
118:      {{~#if cfg.external_automate.root_cert }}
```

jaym:
erchef looks fine. it uses a different config

PR #1752, opened 2019-10-01
datalifecycle: register correct workflow executor

stevendanna:
A recent CI failure revealed we were not correctly purging data. This
was the result of registering our workflow executor under the wrong
name. We were registering the workflow executor under the name of the
"workflow instance" rather than the "workflow".

This fix should result in the purge operation running as expected
again.

As follow up work we need to add a test for purge operations and make
it much harder to make this mistake.

### DIFF
--- a/lib/datalifecycle/purge/workflow.go
+++ b/lib/datalifecycle/purge/workflow.go
@@ -10,8 +10,8 @@ import (
 	"github.com/chef/automate/lib/cereal"
 )
 
-// ConfigureManager registers the purge workflow schedule and job with purge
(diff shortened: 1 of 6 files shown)


lib/datalifecycle/purge/workflow.go
```diff
@@ -12,17 +12,17 @@ import (
 
 // ConfigureManager registers the purge workflow schedule and job with purge
```
ryancragun: We should fix the comment here. We're registering a workflow and task, not a schedule.

PR #1757, opened 2019-10-02
Jamie/member expression dialog improvement

SEAjamieD:
Currently, the member expression builder consists of a single text input, where the user must consult documentation in order to correctly input a member expression.  This branch updates the user experience of this modal so that the user is guided through the building of the member expression one step at a time.

### :+1: Definition of Done
User is stepped through the member expression builder modal.
`Add Expression` button is disabled unless formatting is acceptable
Errors fire when the member expression is already being used
Documentation is updated

Navigate to https://a2-dev.test/settings/policies/editor-access/add-members
Click on `Add Member Expression` in bottom left corner
Build Member expression step by step --> click add `Add Expression` when enabled

- [x] Tests added/updated?

It was determined that the Docs do not need to be updated at this time
- [ ] Docs added/updated?

Before:
MemberExpression-Before

After:
MemberExpressionDialog-After

### DIFF
--- a/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
+++ b/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
@@ -72,7 +72,7 @@
                 <chef-error *ngIf="expressionForm.get('name').hasError('required') && expressionForm.get('name').touched">
                   {{ nameOrId }} is a required field.
                 </chef-error>
-                <chef-error *ngIf="expressionForm.get('name').hasError('pattern') && expressionForm.get('name').touched">
+                <chef-error *ngIf="expressionForm.get('name').hasError('pattern') && expressionForm.get('name').dirty">
                   {{ nameOrId }} must be a text or a wildcard but not both.
                 </chef-error>
               </chef-form-field>
--- a/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
+++ b/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
@@ -73,7 +73,7 @@
                   {{ nameOrId }} is a required field.
                 </chef-error>
                 <chef-error *ngIf="expressionForm.get('name').hasError('pattern') && expressionForm.get('name').dirty">
-                  {{ nameOrId }} must be a text or a wildcard but not both.
+                  {{ nameOrId }} must be text or a wildcard but not both.
                 </chef-error>
               </chef-form-field>
 
--- a/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
+++ b/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
@@ -64,7 +64,7 @@
                   <span class="label">Name<span aria-hidden="true">*</span></span>
                   <chef-input id="expression-name-dropdown" 
                               ngDefaultControl 
-                              (change)="displayExpressionOutput()"
+                              (keyup)="displayExpressionOutput()"
                               formControlName="name"></chef-input>
                 </label>
                 <p class="note">Wildcards (*) are allowed.</p>
--- a/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
+++ b/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
@@ -22,7 +22,7 @@
                   <span class="label">Type<span aria-hidden="true">*</span></span>
                   <select id="expression-type-dropdown"
                           ngDefaultControl 
-                          (change)="updateFormDisplay('identity')"
+                          (change)="updateFormDisplay('type')"
                           formControlName="type">
                     <option value='user'>User</option>
                     <option value='team'>Team</option>
@@ -37,13 +37,13 @@
 
               <chef-form-field 
                 id="create-identity" 
-                *ngIf="showIdentity === true"
+                *ngIf="expressionForm.contains('identity')"
                 [@dropInAnimation]="{ value: triggerValue, params: { height: 74.6875 }}" >
                 <label>
                   <span class="label">Identity Provider<span aria-hidden="true">*</span></span>
                   <select id="expression-identity-dropdown" 
                           ngDefaultControl 
-                          (change)="updateFormDisplay('name')"
+                          (change)="updateFormDisplay('identity')"
                           formControlName="identity">
                     <option value="ldap">LDAP</option>
                     <option value="local">local</option>
@@ -58,13 +58,13 @@
 
               <chef-form-field 
                 id="create-name" 
-                *ngIf="showName === true"
+                *ngIf="expressionForm.contains('name')"
                 [@dropInAnimation]="{ value: triggerValue, params: { height: 119.427 }}">
                 <label>
                   <span class="label">{{ nameOrId }}<span aria-hidden="true">*</span></span>
                   <chef-input id="expression-name-dropdown" 
                               ngDefaultControl 
-                              (keyup)="updateFormDisplay()"
+                              (keyup)="updateFormDisplay('name')"
                               formControlName="name"></chef-input>
                 </label>
                 <p class="note">Wildcards (*) are allowed.</p>
--- a/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
+++ b/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
@@ -39,7 +39,8 @@
               <chef-form-field 
                 id="create-identity" 
                 *ngIf="showInputs('identity')"
-                [@dropInAnimation]="{ value: triggerValue, params: {height: 75}}" >
+                [@dropInAnimation]="{ value: triggerValue, params: { height: 74.6875 }}" 
+                >
                 <label>
                   <span class="label">Identity Provider<span aria-hidden="true">*</span></span>
                   <select id="expression-identity-dropdown" 
@@ -59,7 +60,7 @@
               <chef-form-field 
                 id="create-name" 
                 *ngIf="showInputs('name')"
-                [@dropInAnimation]="{ value: triggerValue, params: {height: 118}}">
+                [@dropInAnimation]="{ value: triggerValue, params: { height: 119.427 }}">
                 <label>
                   <span class="label">Name<span aria-hidden="true">*</span></span>
                   <chef-input id="expression-name-dropdown" 
--- a/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
+++ b/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
@@ -37,7 +37,7 @@
               <chef-form-field 
                 id="create-identity" 
                 *ngIf="showInputs('identity')"
-                [@dropInAnimation]>
+                [@dropInAnimation]="{ value: triggerValue, params: {height: 75}}" >
                 <label>
                   <span class="label">Identity Provider<span aria-hidden="true">*</span></span>
                   <select id="expression-identity-dropdown" 
@@ -55,7 +55,7 @@
               <chef-form-field 
                 id="create-name" 
                 *ngIf="showInputs('name')"
-                [@dropInAnimation]>
+                [@dropInAnimation]="{ value: triggerValue, params: {height: 118}}">
                 <label>
                   <span class="label">Name<span aria-hidden="true">*</span></span>
                   <chef-input id="expression-name-dropdown" 
--- a/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
+++ b/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
@@ -12,8 +12,6 @@
   (confirm)="addMembers()"
   (close)="closePage()">
     <div *ngIf="!loading">
-
-      <!-- MODAL START -->
       <chef-modal label="member-expression-label" class="member-expression-modal" [visible]="modalVisible" (close)="closeModal()">
         <h2>Add Member Expression</h2>
         <div id="modal-content-container">
@@ -93,8 +91,6 @@
           </div>
         </div>
       </chef-modal>
-      <!-- MODAL END -->
-
 
       <div id="no-members-container" *ngIf="sortedMembersAvailable.length === 0">
         <h3>Create more local users or local teams first, or add a member expression.</h3>
--- a/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
+++ b/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
@@ -46,7 +46,7 @@
                     <option value="LDAP">LDAP</option>
                     <option value="LOCAL">local</option>
                     <option value="SAML">SAML</option>
-                    <option value="*">All teams (*)</option>
+                    <option value="*">All {{ allIdentity | titlecase }}s (*)</option>
                   </select>
                 </label>
                 <!-- Note: might not need the errors in this stepper setup  -->
--- a/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
+++ b/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
@@ -31,8 +31,10 @@
                     <option value='*'>All members (*)</option>
                   </select>
                 </label>
-                <!-- Note: might not need the errors in this stepper setup  -->
               </chef-form-field>
+              <chef-error *ngIf="expressionForm.get('type').hasError('required') && expressionForm.get('type').touched">
+                Type is a required field.
+              </chef-error>
 
               <chef-form-field 
                 id="create-identity" 
@@ -49,8 +51,10 @@
                     <option value="*">All {{ allIdentity | titlecase }}s (*)</option>
                   </select>
                 </label>
-                <!-- Note: might not need the errors in this stepper setup  -->
               </chef-form-field>
+              <chef-error *ngIf="expressionForm.get('identity').hasError('required') && expressionForm.get('identity').touched">
+                Identity Provider is a required field.
+              </chef-error>
 
               <chef-form-field 
                 id="create-name" 
@@ -63,9 +67,9 @@
                               formControlName="name"></chef-input>
                 </label>
                 <p class="note">Wildcards (*) are allowed.</p>
-                <!-- Note: probably need errors here  -->
+
                 <chef-error *ngIf="expressionForm.get('name').hasError('required') && expressionForm.get('name').touched">
-                  Some sort of error goes here for name selection
+                  Name is a required field.
                 </chef-error>
               </chef-form-field>
 
@@ -77,7 +81,6 @@
             </chef-form-field>
           </form>
 
-          <!-- expression builder output -->
           <div *ngIf="expressionOutput">
             <p id="expression-output" class="note">Member expression: {{ expressionOutput }}</p>
           </div>
--- a/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
+++ b/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
@@ -22,7 +22,7 @@
                   <span class="label">Type<span aria-hidden="true">*</span></span>
                   <select id="expression-type-dropdown"
                           ngDefaultControl 
-                          (change)="displayExpressionOutput()"
+                          (change)="updateFormDisplay('identity')"
                           formControlName="type">
                     <option value='user'>User</option>
                     <option value='team'>Team</option>
@@ -37,13 +37,13 @@
 
               <chef-form-field 
                 id="create-identity" 
-                *ngIf="showInputs('identity')"
+                *ngIf="showIdentity === true"
                 [@dropInAnimation]="{ value: triggerValue, params: { height: 74.6875 }}" >
                 <label>
                   <span class="label">Identity Provider<span aria-hidden="true">*</span></span>
                   <select id="expression-identity-dropdown" 
                           ngDefaultControl 
-                          (change)="displayExpressionOutput()"
+                          (change)="updateFormDisplay('name')"
                           formControlName="identity">
                     <option value="ldap">LDAP</option>
                     <option value="local">local</option>
@@ -58,19 +58,22 @@
 
               <chef-form-field 
                 id="create-name" 
-                *ngIf="showInputs('name')"
+                *ngIf="showName === true"
                 [@dropInAnimation]="{ value: triggerValue, params: { height: 119.427 }}">
                 <label>
                   <span class="label">{{ nameOrId }}<span aria-hidden="true">*</span></span>
                   <chef-input id="expression-name-dropdown" 
                               ngDefaultControl 
-                              (keyup)="displayExpressionOutput()"
+                              (keyup)="updateFormDisplay()"
                               formControlName="name"></chef-input>
                 </label>
                 <p class="note">Wildcards (*) are allowed.</p>
 
-                <chef-error *ngIf="expressionForm.get('name').hasError('required') && expressionForm.get('name').touched">
+                <chef-error *ngIf="expressionForm.get('name').hasError('required') && expressionForm.get('name').dirty">
                   Name is a required field.
+                </chef-error>
+                <chef-error *ngIf="expressionForm.get('name').hasError('pattern') && expressionForm.get('name').dirty">
+                  This is a pattern error. --> consult for text
                 </chef-error>
               </chef-form-field>
 
--- a/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
+++ b/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
@@ -23,9 +23,9 @@
                   <select id="expression-type-dropdown"
                           ngDefaultControl 
                           formControlName="type">
-                    <option value='USER'>User</option>
-                    <option value='TEAM'>Team</option>
-                    <option value='TOKEN'>Token</option>
+                    <option value='user'>User</option>
+                    <option value='team'>Team</option>
+                    <option value='token'>Token</option>
                     <option value='*'>All members (*)</option>
                   </select>
                 </label>
@@ -43,9 +43,9 @@
                   <select id="expression-identity-dropdown" 
                           ngDefaultControl 
                           formControlName="identity">
-                    <option value="LDAP">LDAP</option>
-                    <option value="LOCAL">local</option>
-                    <option value="SAML">SAML</option>
+                    <option value="ldap">LDAP</option>
+                    <option value="local">local</option>
+                    <option value="saml">SAML</option>
                     <option value="*">All {{ allIdentity | titlecase }}s (*)</option>
                   </select>
                 </label>
--- a/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
+++ b/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
@@ -22,6 +22,7 @@
                   <span class="label">Type<span aria-hidden="true">*</span></span>
                   <select id="expression-type-dropdown"
                           ngDefaultControl 
+                          (change)="displayExpressionOutput()"
                           formControlName="type">
                     <option value='user'>User</option>
                     <option value='team'>Team</option>
@@ -42,6 +43,7 @@
                   <span class="label">Identity Provider<span aria-hidden="true">*</span></span>
                   <select id="expression-identity-dropdown" 
                           ngDefaultControl 
+                          (change)="displayExpressionOutput()"
                           formControlName="identity">
                     <option value="ldap">LDAP</option>
                     <option value="local">local</option>
@@ -62,6 +64,7 @@
                   <span class="label">Name<span aria-hidden="true">*</span></span>
                   <chef-input id="expression-name-dropdown" 
                               ngDefaultControl 
+                              (change)="displayExpressionOutput()"
                               formControlName="name"></chef-input>
                 </label>
                 <p class="note">Wildcards (*) are allowed.</p>
--- a/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
+++ b/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
@@ -70,10 +70,10 @@
                 <p class="note">Wildcards (*) are allowed.</p>
 
                 <chef-error *ngIf="expressionForm.get('name').hasError('required') && expressionForm.get('name').dirty">
-                  Name is a required field.
+                  {{ nameOrId }} is a required field.
                 </chef-error>
                 <chef-error *ngIf="expressionForm.get('name').hasError('pattern') && expressionForm.get('name').dirty">
-                  This is a pattern error. --> consult for text
+                  {{ nameOrId }} must be a text or a wildcard but not both.
                 </chef-error>
               </chef-form-field>
 
--- a/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
+++ b/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
@@ -61,7 +61,7 @@
                 *ngIf="showInputs('name')"
                 [@dropInAnimation]="{ value: triggerValue, params: { height: 119.427 }}">
                 <label>
-                  <span class="label">Name<span aria-hidden="true">*</span></span>
+                  <span class="label">{{ nameOrId }}<span aria-hidden="true">*</span></span>
                   <chef-input id="expression-name-dropdown" 
                               ngDefaultControl 
                               (keyup)="displayExpressionOutput()"
@@ -90,7 +90,7 @@
             <chef-button primary 
                         [disabled]="!expressionForm.valid" 
                         (click)="validateAndAddExpression()">Add Expression</chef-button>
-            <chef-button secondary (click)="closeModal()">Cancel</chef-button>
+            <chef-button tertiary (click)="closeModal()">Cancel</chef-button>
           </div>
         </div>
       </chef-modal>
--- a/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
+++ b/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
@@ -69,10 +69,10 @@
                 </label>
                 <p class="note">Wildcards (*) are allowed.</p>
 
-                <chef-error *ngIf="expressionForm.get('name').hasError('required') && expressionForm.get('name').dirty">
+                <chef-error *ngIf="expressionForm.get('name').hasError('required') && expressionForm.get('name').touched">
                   {{ nameOrId }} is a required field.
                 </chef-error>
-                <chef-error *ngIf="expressionForm.get('name').hasError('pattern') && expressionForm.get('name').dirty">
+                <chef-error *ngIf="expressionForm.get('name').hasError('pattern') && expressionForm.get('name').touched">
                   {{ nameOrId }} must be a text or a wildcard but not both.
                 </chef-error>
               </chef-form-field>
--- a/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
+++ b/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
@@ -31,16 +31,15 @@
                     <option value='*'>All members (*)</option>
                   </select>
                 </label>
+                <chef-error *ngIf="expressionForm.get('type').hasError('required') && expressionForm.get('type').touched">
+                  Type is a required field.
+                </chef-error>
               </chef-form-field>
-              <chef-error *ngIf="expressionForm.get('type').hasError('required') && expressionForm.get('type').touched">
-                Type is a required field.
-              </chef-error>
 
               <chef-form-field 
                 id="create-identity" 
                 *ngIf="showInputs('identity')"
-                [@dropInAnimation]="{ value: triggerValue, params: { height: 74.6875 }}" 
-                >
+                [@dropInAnimation]="{ value: triggerValue, params: { height: 74.6875 }}" >
                 <label>
                   <span class="label">Identity Provider<span aria-hidden="true">*</span></span>
                   <select id="expression-identity-dropdown" 
@@ -52,10 +51,10 @@
                     <option value="*">All {{ allIdentity | titlecase }}s (*)</option>
                   </select>
                 </label>
+                <chef-error *ngIf="expressionForm.get('identity').hasError('required') && expressionForm.get('identity').touched">
+                  Identity Provider is a required field.
+                </chef-error>
               </chef-form-field>
-              <chef-error *ngIf="expressionForm.get('identity').hasError('required') && expressionForm.get('identity').touched">
-                Identity Provider is a required field.
-              </chef-error>
 
               <chef-form-field 
                 id="create-name" 
--- a/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
+++ b/components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
@@ -40,10 +40,10 @@
                 *ngIf="expressionForm.contains('identityProvider')"
                 [@dropInAnimation]="{ value: triggerValue, params: { height: 74.6875 }}" >
                 <label>
-                  <span class="label">Identity Provider<span aria-hidden="true">*</span></span>
+                  <span class="label">Identity<span aria-hidden="true">*</span></span>
                   <select id="expression-identity-dropdown" 
                           ngDefaultControl 
-                          (change)="updateFormDisplay('identity')"
+                          (change)="updateFormDisplay('identityProvider')"
                           formControlName="identityProvider">
                     <option value="ldap">LDAP</option>
                     <option value="local">local</option>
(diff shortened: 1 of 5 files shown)


components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -34,8 +35,24 @@ import { Regex } from 'app/helpers/auth/regex';
 @Component({
   selector: 'app-policy-add-members',
   templateUrl: './policy-add-members.component.html',
-  styleUrls: ['./policy-add-members.component.scss']
+  styleUrls: ['./policy-add-members.component.scss'],
+  animations: [
```
SEAjamieD: Animations used for the fade in/out of form elements as they are needed.  I feel that we're ok animating height here because there are minimal elements involved and they are sitting in a modal.  This helps the modal resize more smoothly without having to adjust the modals actual height separately.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -66,15 +83,20 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
     private store: Store<NgrxStateAtom>,
     private router: Router,
     fb: FormBuilder) {
 
     this.expressionForm = fb.group({
-      // Must stay in sync with error checks in policy-add-members.component.html
-      expression: ['', [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]]
+      type: ['', [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]],
+      identity: [''],
+      name: ['']
```
SEAjamieD: We start the second two form elements off with no validation, so that we can conditionally add/remove validations later on.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -169,14 +192,14 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
     this.store.select(addPolicyMembersStatus).pipe(
       filter(identity),
       takeUntil(pendingAdd))
-      .subscribe((state) => {
-        if (state === EntityStatus.loadingSuccess) {
+      .subscribe((p_state) => {
```
SEAjamieD: Small update to the variable name from `state` to `p_state` only because our new animation is importing a `state` variable and there is no variable shadowing allowed.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -228,12 +251,16 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
   public closeModal(): void {
+    // unsubscribe from Changes to the form
+    this.unsubscribeFromExpressionFormChanges();
     this.resetModal();
     this.modalVisible = false;
   }
 
   public openModal(): void {
     this.resetModal();
+    // subscribe to expressionForm changes
```
SEAjamieD: subscribing to the form changes once the form is actually available.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -228,12 +251,16 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
   }
 
   public closeModal(): void {
+    // unsubscribe from Changes to the form
+    this.unsubscribeFromExpressionFormChanges();
```
SEAjamieD: unsubscribing from the form when we no longer need it.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +332,74 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
     this.closeModal();
   }
+
+  public showInputs(inputName: string): boolean {
+    const typeValue = this.expressionForm.get('type').value;
+    const identityValue = this.expressionForm.get('identity').value;
+
+    switch (inputName) {
+      case 'identity':
+        if (typeValue === 'USER' || typeValue === 'TEAM') {
```
SEAjamieD: if the user has entered either Team or User, then we want to show the identity element.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +332,74 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+            this.updateValidations('identity', true);
+            return true;
+          } else {
+            this.updateValidations('identity', false);
+            this.expressionForm.get('identity').reset();
+            return false;
+          }
+        break;
+      case 'name':
+        if ( typeValue === 'TOKEN' || (identityValue && identityValue !== '*') ) {
```
SEAjamieD: if the user has chosen Toke as their type and the identity value is not a `*`, we want to show the name element.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +332,74 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+
+  private updateValidations(inputName: string, active: boolean): void {
+    const updateInput = this.expressionForm.get(inputName);
+
+    active === true
+      ? updateInput.setValidators([Validators.required,
+        Validators.pattern(Regex.patterns.NON_BLANK)])
+      : updateInput.clearValidators();
+
+      updateInput.updateValueAndValidity();
```
SEAjamieD: When an input is active, we turn the validations on.  When an input is not active, we turn the validations off.  This allows us to properly activate and deactivate the `Add Expression` button.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
```diff
@@ -14,27 +14,84 @@
...
+                </label>
+                <chef-error *ngIf="expressionForm.get('type').hasError('required') && expressionForm.get('type').touched">
+                  Type is a required field.
+                </chef-error>
+              </chef-form-field>
+
+              <chef-form-field 
+                id="create-identity" 
+                *ngIf="showInputs('identity')"
+                [@dropInAnimation]="{ value: triggerValue, params: { height: 74.6875 }}" >
```
SEAjamieD: params let us conditionally pass in a variable for the animation.  Since we do not have to worry too much about screen size changes, I've chose to pass in the exact height the element sits at naturally, this helps to reduce any animation jitters.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
msorens: Please use `pState` though to follow conventions (or `addPolicyState` might help, too).

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
msorens: The comment on L254 is not needed; same on L262. In fact, that falls under the, well, "very evil" category of comments. Please see https://www.red-gate.com/simple-talk/opinion/opinion-pieces/fighting-evil-code-comments-comments/ 😁

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +332,74 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+        break;
+      case 'name':
+        if ( typeValue === 'TOKEN' || (identityValue && identityValue !== '*') ) {
+          this.updateValidations('name', true);
+          return true;
+        } else {
+          this.updateValidations('name', false);
+          this.expressionForm.get('name').reset();
+          return false;
+        }
```
msorens: Hmm... a lot of duplication in the switch above but not sure if there's a DRYer approach. However, I would at least suggest to use `inputName` instead of `'name'` and `'identity'` in the several function calls.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +332,74 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+        break;
+      default:
+        return false;
+    }
+  }
+
+  private updateValidations(inputName: string, active: boolean): void {
+    const updateInput = this.expressionForm.get(inputName);
+
+    active === true
```
msorens: Nothing invalid with using the ternary this way but convention in the code base is to use it only when you are using its return value; otherwise use a regular `if-then-else`. Then it is also more clear when you just say `if (active)...` without the `=== true`.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +332,74 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+
+      updateInput.updateValueAndValidity();
+    }
+
+
+  private displayExpressionOutput(formValues: object): void {
+    const entries = Object.entries(formValues);
+    const output = [];
+
+    entries.forEach( ([key, value]) => {
```
msorens: Consider adding some strong typing either for `entries` or at least for `key` and `value`. And definitely for `output`.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +332,74 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+    // initialize stream
+    const expressionValueChanges = this.expressionForm.valueChanges;
+    // subscribe to form changes
+    this.formSubscription = expressionValueChanges.subscribe(formValues =>
+              this.displayExpressionOutput(formValues));
+  }
+
+  private unsubscribeFromExpressionFormChanges(): void {
+    this.formSubscription.unsubscribe();
+  }
```
msorens: Where these are used are already tiny functions; would suggest to inline the code of these two functions without losing any clarity. (Also, don't think the comments on L395 and L397 are needed.)

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+            identity: [''],
+            name: ['']
+        });
+        fixture.detectChanges();
+    });
+
+    it('should be created', () => {
+        expect(component).toBeTruthy();
+    });
+
```
msorens: Nice tests; let me suggest to see if you could combine most of these into one or perhaps two sets of data-driven tests. Search the code base for `using([` and you will find many examples.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +332,74 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+        return false;
+    }
+  }
+
+  private updateValidations(inputName: string, active: boolean): void {
+    const updateInput = this.expressionForm.get(inputName);
+
+    active === true
+      ? updateInput.setValidators([Validators.required,
+        Validators.pattern(Regex.patterns.NON_BLANK)])
```
msorens: You will need one more validator: If a wildcard is used, it must constitute the _entire_ value. Thus:
Legal values: *, fred, _foo, etc.
Illegal values: abc*, *foo, f*o*o*, **

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +332,74 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+    });
+
+    this.expressionOutput = output.join(':');
+  }
+
+  private subscribeToExpressionFormChanges(): void {
+    // initialize stream
+    const expressionValueChanges = this.expressionForm.valueChanges;
+    // subscribe to form changes
+    this.formSubscription = expressionValueChanges.subscribe(formValues =>
```
msorens: This is firing constantly when the modal is opened, not just on changes. Need to understand why this is happening.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +332,74 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+      ? updateInput.setValidators([Validators.required,
+        Validators.pattern(Regex.patterns.NON_BLANK)])
+      : updateInput.clearValidators();
+
+      updateInput.updateValueAndValidity();
+    }
+
+
+  private displayExpressionOutput(formValues: object): void {
+    const entries = Object.entries(formValues);
```
msorens: Add strong typing for entries
```suggestion
    const entries: string[][] = Object.entries(formValues);
```
... and that will give key and value strong types too.
Also add type for `output`.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +332,74 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+    const output = [];
+
+    entries.forEach( ([key, value]) => {
+      if (key === 'name' && value != null && value.length > 0) {
+        return output.push(value);
+      } else if (value != null && value.length > 0) {
+        return output.push(value.toLowerCase());
+      }
+    });
```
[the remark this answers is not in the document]
SEAjamieD: Thanks!  changing the HTML ended up allowing me to completely pull out the forEach.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
SEAjamieD: This ended up being a result of my validation updates inside an `*ngIf` which then would fire off another formChange event.  I ended up deciding to nix the form subscription and use onChange events instead.

susanev:
@SEAjamieD i took a quick look, can take a better look tomorrow..

(1) can you make the cancel button tertiary?
(2) for token, the next field should be `ID` instead of `Name`, that may have been wrong in my prototype

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.spec.ts
SEAjamieD: Thanks for the suggestion!  `using` is awesome! 🎉

SEAjamieD:
@susanev I was looking at the documentation and I'm not sure it really needs any updating.  The concepts are still the same, only difference is now we're helping them build the expression so they don't have to figure it out as much.  Thoughts?

https://automate.chef.io/docs/iam-v2-guide/#member-expressions

susanev:
@SEAjamieD yah actually we are doing a huge overhaul on those docs right now, so it doesnt make sense to have you do an update now anyway.. so feel free to ignore that content from the issue, sorry about that.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +328,68 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+  }
+
+  private setFormLabels(typeValue: string): void {
+    if (typeValue === 'token') {
+      this.nameOrId = 'ID';
+    } else {
+      this.allIdentity = typeValue;
+      this.nameOrId = 'Name';
+    }
+  }
```
SEAjamieD: I decided to move this piece from showInputs() into its own small function just to separate concerns.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -169,14 +192,14 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
     this.store.select(addPolicyMembersStatus).pipe(
       filter(identity),
       takeUntil(pendingAdd))
-      .subscribe((state) => {
-        if (state === EntityStatus.loadingSuccess) {
+      .subscribe((addPolicyState) => {
```
SEAjamieD: variable `state` was only changed because it became a shadow now that we are importing `state` from the angular animations API

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
```diff
@@ -14,27 +14,87 @@
...
+                  <span class="label">{{ nameOrId }}<span aria-hidden="true">*</span></span>
+                  <chef-input id="expression-name-dropdown" 
+                              ngDefaultControl 
+                              (keyup)="displayExpressionOutput()"
+                              formControlName="name"></chef-input>
+                </label>
+                <p class="note">Wildcards (*) are allowed.</p>
+
+                <chef-error *ngIf="expressionForm.get('name').hasError('required') && expressionForm.get('name').touched">
+                  Name is a required field.
```
msorens: 1. This is not working--an empty field does not display the error pop-up. (Works ok for for Type and Identity Provider.
2. Need an additional `<chef-error>` to check for `hasError('pattern')` with the text `Name must be a user name or a wildcard but not both`

@susanev Need a check on that messaging: Normally we would say "Only lowercase letters, numbers, hyphens are allowed", but this needs to allow any LDAP or SAML user name, too, but LDAP at least allows all sorts of punctuation characters, too. So that's why I think just saying "a user name" conveys sufficient information.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -66,15 +83,20 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
 
   constructor(
     private store: Store<NgrxStateAtom>,
     private router: Router,
     fb: FormBuilder) {
 
     this.expressionForm = fb.group({
-      // Must stay in sync with error checks in policy-add-members.component.html
-      expression: ['', [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]]
+      type: ['', [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]],
```
msorens: Since this is a dropdown and not a typeable field, do not need the pattern validator. (And, you are not even using it in the HTML--that's why you probably should not delete the comment on L76; it is there as a reminder for precisely that. 😁 )

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +328,68 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+      this.allIdentity = typeValue;
+      this.nameOrId = 'Name';
+    }
+  }
+
+  private updateValidations(inputName: string, active: boolean): void {
+    const updateInput = this.expressionForm.get(inputName);
+
+    if (active) {
+      return inputName === 'name'
```
msorens: This does not quite make sense; why are you trying to return something from a function that declares it returns void? Also, by returning early, you are not executing L387 (which the docs say must follow a call to `setValidators`).
I think you just want this to be an `if (input === 'name')`

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +328,68 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+              ? updateInput.setValidators([Validators.required,
+                                      Validators.pattern(Regex.patterns.NON_BLANK),
+                                      Validators.pattern(Regex.patterns.NO_MIXED_WILDCARD)])
+              : updateInput.setValidators([Validators.required,
+                                      Validators.pattern(Regex.patterns.NON_BLANK)]);
```
msorens: Same comment I made for type applies here: identity is a dropdown and not a typeable field, so do not need the pattern validator

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +328,68 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+  }
+
+  private updateValidations(inputName: string, active: boolean): void {
+    const updateInput = this.expressionForm.get(inputName);
+
+    if (active) {
+      return inputName === 'name'
+              ? updateInput.setValidators([Validators.required,
+                                      Validators.pattern(Regex.patterns.NON_BLANK),
+                                      Validators.pattern(Regex.patterns.NO_MIXED_WILDCARD)])
```
msorens: By adding these here that wires up the enable/disable of the save button. But you are missing the other half--notifying the user of the reason. See my earlier comment to add a `<chef-error>` to handle that.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +328,68 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+            this.updateValidations(inputName, true);
+            return true;
+          } else {
+            this.updateValidations(inputName, false);
+            this.expressionForm.get(inputName).reset();
+            return false;
+          }
+        break;
+      case 'name':
+        if ( typeValue === 'token' || (identityValue && identityValue !== '*') ) {
```
msorens: Prefer to use a named constant for clarity:
```suggestion
        if ( typeValue === 'token' || (identityValue && identityValue !== matchAllWildcard) ) {
```

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +328,68 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+    } else {
+      updateInput.clearValidators();
+    }
+
+    updateInput.updateValueAndValidity();
+  }
+
+  public displayExpressionOutput(): void {
+    const values: string[] = Object.values(this.expressionForm.value);
+    const output: string[] = values.filter(value => value != null && value.length > 0);
```
msorens: Type of `output` is properly inferred from `values` so better to omit in this case.
```suggestion
    const output = values.filter(value => value != null && value.length > 0);
```

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
```diff
@@ -14,27 +14,87 @@
...
+                *ngIf="showInputs('identity')"
+                [@dropInAnimation]="{ value: triggerValue, params: { height: 74.6875 }}" >
+                <label>
+                  <span class="label">Identity Provider<span aria-hidden="true">*</span></span>
+                  <select id="expression-identity-dropdown" 
+                          ngDefaultControl 
+                          (change)="displayExpressionOutput()"
+                          formControlName="identity">
+                    <option value="ldap">LDAP</option>
+                    <option value="local">local</option>
```
msorens: The order LDAP, local, SAML seems odd. I would suggest local, LDAP, SAML.
@susanev Did you have a strong reason for this existing order?

susanev: it's alphabetical (natural) which is what we've been trying to standardize on

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
```diff
@@ -14,27 +14,87 @@
...
+                *ngIf="showInputs('name')"
+                [@dropInAnimation]="{ value: triggerValue, params: { height: 119.427 }}">
+                <label>
+                  <span class="label">{{ nameOrId }}<span aria-hidden="true">*</span></span>
+                  <chef-input id="expression-name-dropdown" 
+                              ngDefaultControl 
+                              (keyup)="displayExpressionOutput()"
+                              formControlName="name"></chef-input>
+                </label>
+                <p class="note">Wildcards (*) are allowed.</p>
```
msorens: Actually wildcards are **not** allowed; a single wildcard is allowed.
So perhaps `Use a wildcard to specify all {{user_or_team_label}}`
@susanev ??

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+    describe('showInputs', () => {
+
+        it('should return false when inputName is not identity or name', () => {
```
msorens: Consider `returns` instead of `should return` for brevity (here and below).

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+        });
+
+        using([
+            ['user', true, 'identity and when typeValue equals user', 'identity'],
+            ['team', true, 'identity and when typeValue equals team', 'identity'],
+            ['*', false, 'identity and typeValue does not equal team or user', 'identity'],
+            ['token', true, 'name and when typeValue equals token', 'name'],
+            ['team', false, 'name + typeValue is not token + identityValue does not exist', 'name']
+        ], function (type: string, outcome: boolean, description: string, input: string) {
+            it(`should return ${outcome} when inputName is ${description}`, () => {
```
msorens:
```suggestion
            it(`returns ${outcome} when inputName is ${input} ${description}`, () => {
```
... with supporting changes to description.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+            expect(component.showInputs('blank')).toBe(false);
+        });
+
+        using([
+            ['user', true, 'identity and when typeValue equals user', 'identity'],
+            ['team', true, 'identity and when typeValue equals team', 'identity'],
+            ['*', false, 'identity and typeValue does not equal team or user', 'identity'],
+            ['token', true, 'name and when typeValue equals token', 'name'],
+            ['team', false, 'name + typeValue is not token + identityValue does not exist', 'name']
+        ], function (type: string, outcome: boolean, description: string, input: string) {
```
msorens: Consider:
`description` should be first or last rather than in the middle since it is different than the other args.
Also, might want `input` before `output` since that is a more "natural" order.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
msorens: "Alphabetical with special circumstances", seems preferable. 🤷‍♂

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
SEAjamieD: we DO need it to be non-blank though, correct?

SEAjamieD: Ah ok - thank you!

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
SEAjamieD: @susanev / @msorens - Is there a final consensus here?

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
```diff
@@ -14,27 +14,90 @@
...
+                  </select>
+                </label>
+                <chef-error *ngIf="expressionForm.get('identity').hasError('required') && expressionForm.get('identity').touched">
+                  Identity Provider is a required field.
+                </chef-error>
+              </chef-form-field>
+
+              <chef-form-field 
+                id="create-name" 
+                *ngIf="expressionForm.contains('name')"
```
SEAjamieD: All inputs are not being displayed based on whether or not the FormGroup contains its specific control.  This was suggested by Angular Core team as a preferred check: https://github.com/angular/angular/issues/7970#issuecomment-228624899

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
```diff
@@ -14,27 +14,90 @@
...
+                <label>
+                  <span class="label">{{ nameOrId }}<span aria-hidden="true">*</span></span>
+                  <chef-input id="expression-name-dropdown" 
+                              ngDefaultControl 
+                              (keyup)="updateFormDisplay('name')"
+                              formControlName="name"></chef-input>
+                </label>
+                <p class="note">Wildcards (*) are allowed.</p>
+
+                <chef-error *ngIf="expressionForm.get('name').hasError('required') && expressionForm.get('name').touched">
```
SEAjamieD: Anyone have suggestions for rendering this error?  `chef-input` is not reacting to touched/untouched.  My thought was to instead build one ourselves using an blur event that set a flag in `.component`, but it seems like `chef-input` isn't firing onBlur events either 🤷‍♂

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +327,79 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+          this.addNameControl();
+        } else {
+          this.expressionForm.removeControl('name');
+        }
+        break;
+      default:
+        break;
+    }
+
+  }
```
SEAjamieD: MAJOR overhaul in showInputs().  I'm now following Angular Core team suggestion and adding/removing form controls as we go.  This has provided a much simpler way of making sure the inputs are showing and the correct validators are applied.  Although small, I broke reused patterns out into their own functions to make showInputs() switch statement easier to read and follow.  showInputs() is now only firing on form input changes.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -137,6 +158,7 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
 
     this.store.dispatch(new GetTeams());
     this.store.dispatch(new GetUsers());
+
```
bcmdarroch: [nit] new line

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +327,79 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
     this.addOrRemoveQueuedMember(true, member);
     this.closeModal();
   }
+
+  private showInputs(inputName: string): void {
```
msorens: Consider changing `inputName` to `fieldName`; I think that makes the intent more clear.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +327,79 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+
+  private addIdentityControl(): void {
+    this.expressionForm.addControl('identity', new FormControl('', Validators.required));
+  }
+
+  private addNameControl(): void {
+    this.expressionForm.addControl('name', new FormControl('',
+      [
+        Validators.required,
+        Validators.pattern(Regex.patterns.NON_BLANK),
```
msorens: In this case, `NON_BLANK` is redundant so this line can be removed; it is covered by `NO_MIXED_WILDCARD`

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +327,79 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+        Validators.pattern(Regex.patterns.NO_MIXED_WILDCARD)
+      ]
+    )
+    );
+  }
+
+  private setExpressionOutput(): void {
+    const values: string[] = Object.values(this.expressionForm.value);
+    const output = values.filter(value => value != null && value.length > 0);
+    this.expressionOutput = output.join(':');
```
msorens: This is really personal preference so feel free to ignore this note; functional style is a good pattern to be aware of, though.
You will see this a lot in our *.effects.ts files, or example.
Here, you could write this block as:
```
    this.expressionOutput =
      (Object.values(this.expressionForm.value) as string[])
      .filter(value => value != null && value.length > 0)
      .join(':');
 ```

File: components/automate-ui/src/app/pages/policy/add-members/policy-add-members.spec.ts [hunk 40]
```diff
@@ -0,0 +1,87 @@
...
+
+    it('should be created', () => {
+        expect(component).toBeTruthy();
+    });
+
+    describe('updateFormDisplay', () => {
+
+        using([
+            [{ type: 'team', identity: null, name: null }, 'type', 'team'],
+            [{ type: '*', identity: null, name: null}, 'team', '*'],
```
msorens: I believe 'team' in that parameter is an invalid input.
```suggestion
            [{ type: '*', identity: null, name: null}, 'type', '*'],
```

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+        fixture.detectChanges();
+    });
+
+    it('should be created', () => {
+        expect(component).toBeTruthy();
+    });
+
+    describe('updateFormDisplay', () => {
+
+        using([
```
msorens: Please add tests for `user:*`, `team:*`, and `token:*`.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
SEAjamieD: thanks for showing, I think i'll update to use this style

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
```diff
@@ -14,27 +14,90 @@
...
+              <chef-form-field 
+                id="create-identity" 
+                *ngIf="expressionForm.contains('identity')"
+                [@dropInAnimation]="{ value: triggerValue, params: { height: 74.6875 }}" >
+                <label>
+                  <span class="label">Identity Provider<span aria-hidden="true">*</span></span>
+                  <select id="expression-identity-dropdown" 
+                          ngDefaultControl 
+                          (change)="updateFormDisplay('identity')"
+                          formControlName="identity">
```
msorens: I finally put my finger on something that had been bothering me here: every time I read this I had to map in my head identity to mean identity provider so this makes sense in context. So it is longer, but I think `identityProvider` would be much more helpful than `identity`.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +325,79 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
     this.addOrRemoveQueuedMember(true, member);
     this.closeModal();
   }
+
+  private showInputs(fieldName: string): void {
```
msorens:
```suggestion
  private showInputs(fieldName: FieldName): void {
```
then define this literal string type
```
export type FieldName = 'type' | 'identity' | 'name';
```
...just above the class definition in the file.

Types in TypeScript are... "strong-ish." This change does not actually add as much benefit as you might think. For instance, that correction I suggested to the unit test (using "team" instead of "type") would still not be noticed.
But it does provide two small but important benefits.
1. A reader of this code will know precisely what inputs are supposed to be fed to showInputs; with type "string" as in L329 one has no idea of the constraints.
2. This will generate compile-time errors if a typo is introduced on L337 or L345.

Please see _TypeScript Type Traps_ at https://docs.google.com/presentation/d/1ir7P7V0tzO4cd4vCtyUb-a5fHFnBsRYTUoCWLq6xksU/edit#slide=id.g20011f5b3f_0_0

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -305,4 +325,79 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+        }
+        break;
+      case 'identity':
+        if (formValues.identity !== matchAllWildCard) {
+          this.addNameControl();
+        } else {
+          this.expressionForm.removeControl('name');
+        }
+        break;
+      default:
```
msorens: Right now the reader may wonder whether or not `name` was left out deliberately or not. Suggest add `case 'name'` just above the default so it is clear that for `name` we need do nothing.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    beforeEach(() => {
+        fixture = TestBed.createComponent(PolicyAddMembersComponent);
+        component = fixture.componentInstance;
+        component.expressionForm = new FormBuilder().group({
+            type: ['', Validators.required],
+            identity: '',
+            name: ''
+        });
+        fixture.detectChanges();
```
msorens: I do not think this line is required since you are testing a pure function.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
msorens: Would probably be helpful to add some in-code comments as to what this whole animation block is doing.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
SEAjamieD: Thanks for this explanation!

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.spec.ts
```diff
@@ -9,7 +9,7 @@ import { runtimeChecks } from 'app/ngrx.reducers';
 import { PolicyAddMembersComponent } from './policy-add-members.component';
 import { using } from 'app/testing/spec-helpers';
 
-
+type FieldName = 'type' | 'identity' | 'name';
```
msorens: Oooh! Please don't duplicate this; just import it from the main file.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
```diff
@@ -14,27 +14,90 @@
...
+
+              <chef-form-field 
+                id="create-identity" 
+                *ngIf="expressionForm.contains('identityProvider')"
+                [@dropInAnimation]="{ value: triggerValue, params: { height: 74.6875 }}" >
+                <label>
+                  <span class="label">Identity Provider<span aria-hidden="true">*</span></span>
+                  <select id="expression-identity-dropdown" 
+                          ngDefaultControl 
+                          (change)="updateFormDisplay('identity')"
```
msorens: Ah, my previous comment was to update the name on this line, too, to `identityProvider`.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
SEAjamieD: As discussed - this will be rolled into https://github.com/chef/automate/issues/1906

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
susanev: @SEAjamieD keep the order as defined in the issue for now, we can iterate later if we receive user feedback

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.html
SEAjamieD: @susanev bumping to top

susanev:
@SEAjamieD a few more things to address if possible...

- that cancel button looks weird, it should be re-sizing to have 16px margin on either side instead of stretching to match the size of the other button
- if you use the close-x it sends me back to the list page, but it should stay on the add members page
- can we get a space between the label and the asterisks
- lets change that wildcard text on the name/id fields to `A single wildcard (*) is allowed`
- itd be cool to get the select styles that we added for ingest rules into this, but that can be a separate pr if you want

SEAjamieD:
> * if you use the close-x it sends me back to the list page, but it should stay on the add members page

@susanev - I'm going to suggest we roll this into its own ticket.  Current functionality now is closing the entire page out.  I haven't touched this piece of the logic and when investigating it looks like this is being passed down via `chef-page` stencil component and may take a little longer to diagnose.

susanev:
new issue for close-x issue #1934

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -238,7 +261,7 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
   }
 
   resetModal(): void {
-    this.expressionForm.reset();
+    this.resetForm();
     this.unparsableMember = false;
     this.duplicateMember = false;
```
msorens: Replace these two lines with
```
this.resetErrors();
```
Also, please make `resetErrors` private; it is not used outside the file.

PR #1758, opened 2019-10-02
compliance-service: remove most nakedret findings

stevendanna:
This removes most of the findings of the nakedret linter which was the
last custom lint configuration for the compliance-service. The one
case that I ignored requires someone more familiar with the compliance
service to decide what to do with some of the error cases.

### DIFF
--- a/components/compliance-service/Makefile
+++ b/components/compliance-service/Makefile
@@ -1,6 +1,3 @@
-# nakedret: naked returns in long functions
-LINTERARGS=-v -D nakedret ./...
-# govet in 1.16 is very confused by some of the code in this component
 include ../../Makefile.common_go
 
 .PHONY: test
--- a/components/compliance-service/api/reporting/server/server.go
+++ b/components/compliance-service/api/reporting/server/server.go
@@ -19,7 +19,6 @@ import (
 	"github.com/chef/automate/lib/errorutils"
 	"github.com/chef/automate/lib/grpc/auth_context"
 	"github.com/chef/automate/lib/io/chunks"
-	"github.com/chef/automate/lib/stringutils"
 )
 
 // Chosen somewhat arbitrarily to be a "good enough" value.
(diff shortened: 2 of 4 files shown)


components/compliance-service/reporting/relaxting/depth.go
```diff
@@ -160,7 +161,7 @@ func getControlLevelControlSums(hit *elastic.SearchHit) (nodeControlSummary repo
 			}
 		}
 	}
-	return
+	return // nolint:nakeret
```
stevendanna: This function and most of the functions in this file appear to drop error information. @rickmarry is looking into it.

components/compliance-service/Makefile
```diff
@@ -1,5 +1,3 @@
-# nakedret: naked returns in long functions
-LINTERARGS=-v -D nakedret ./...
 # govet in 1.16 is very confused by some of the code in this component
```
phiggins: Is this comment relevant anymore?

stevendanna:
```suggestion
```

components/compliance-service/api/reporting/server/server.go
```diff
@@ -383,38 +382,48 @@ func formatFilters(filters []*reporting.ListFilter) map[string][]string {
...
+			}
+			sorter.Strings(validFields)
+			return ret, &errorutils.InvalidError{Msg: fmt.Sprintf("Parameter 'sort' only supports one of the following fields: %v", validFields)}
 		}
 		sort = in.Sort
-	} else {
-		sort = defaultSortField
 	}
-	return
+	ret.sort = sort
```
phiggins: I know it's not really your job to make this squeaky clean (thanks though!), but this could probably be the value of the sort field map rather than the key. Does `ret.sort` get used for anything other than looking up the value in the map?

components/compliance-service/reporting/relaxting/depth.go
phiggins: https://github.com/chef/automate/issues/1763

components/compliance-service/api/reporting/server/server.go
stevendanna: I like this suggestion and since this isn't urgent I went ahead and did it.

Issue #536, opened 2019-06-07
update nodes in postgres during update process for ingested nodes


vjeffrey:
# User Story

We recently modified the ingest service and compliance service to call the nodemanager service with a small amount of metadata on every ingestion. We need to ensure the projects information for those nodes is always up to date to maintain rbac integrity.

This means that when ingest rules are changed and we kick off the event cycle for the ProjectUpdateManager (https://github.com/chef/automate/issues/60), we also need to update the nodes in postgres (i.e. `api/v0/nodes`). 

The projects data is stored for the nodes as json, so something like 
` UPDATE nodes SET projects = $1 WHERE projects_data ->> 'organization_name' ...` should work

# Definition of Done
nodes in nodemanager db are updated as part of the ProjectUpdateManager process

# Demo Script / Repro Steps
add nodes with chef_load_nodes and chef_load_compliance_scans
add a couple nodes via manual node add too
create a project with ingest rules to match a set of nodes
edit the ingest rules
ensure the nodes in the nodemanager db have the correct projects

PR #1772, opened 2019-10-02
Update IAMv2 project tags in nodemanager

lancewf:
Adding the node manager to the IAMv2 project update. 

#536

### :+1: Definition of Done
Project tags for nodes in the nodemanager are updated during a project update event. 

1. `build components/automate-deployment && build components/authz-service && build components/nodemanager-service`
1. Upgrade to beta2.1 and load data `chef-automate iam upgrade-to-v2 --beta2.1`
1. Add some compliance nodes with `chef_load_compliance_nodes 5`
1. Add some client runs nodes with `chef_load_nodes 5`
1. Create a project with rules
1. Check PostgreSQL that the projects on the nodes were updated. To look at postgres use `chef-automate dev psql nodemanager_service`. The query I have been using is `select n.id, COALESCE(array_to_json(array_remove(array_agg(p.project_id), NULL)), '[]') AS projects, n.projects_data from nodes n LEFT JOIN nodes_projects np on np.node_id = n.id LEFT JOIN projects p on np.project_id = p.id group by n.id;`

- [x] Tests added/updated?
- [ ] Docs added/updated?


components/ingest-service/integration_test/project_update_test.go
```diff
@@ -1489,7 +1489,7 @@ func TestProjectUpdatePainlessElasticsearchScript(t *testing.T) {
 					jobStatus, err = suite.ingest.JobStatus(ctx, esJobID)
 					assert.Nil(t, err)
 					if err != nil {
-						assert.FailNow(t, "testing job complete")
```
lancewf: This is the only real change in this file. The rest of the changes are from golang's code simplifier.

components/authz-service/server/v2/project_update_manager.go
```diff
@@ -41,6 +41,7 @@ const (
 var ProjectUpdateDomainServices = []string{
 	"ingest",
 	"compliance",
+	"nodemanager",
```
lancewf: Adding the nodemanager to one of the services that get called for project updates.

components/nodemanager-service/nodemanager.go
```diff
@@ -228,6 +230,33 @@ func serve(ctx context.Context, config *config.Nodemanager, connFactory *securec
...
+		defer authzConn.Close() // nolint: errcheck
+
+		authzProjectsClient := iam_v2.NewProjectsClient(authzConn)
+
+		projectUpdateManager, err := createProjectUpdateCerealManager(connFactory, config.Cereal.Endpoint)
+		if err != nil {
+			return err
+		}
+
+		err = project_update_lib.RegisterTaskExecutors(projectUpdateManager, "nodemanager",
```
lancewf: Registering the node manager with the cereal service for project updates.

components/nodemanager-service/pgdb/db.go
```diff
@@ -12,12 +12,21 @@ import (
...
 	"github.com/chef/automate/lib/logger"
 )
 
+type ProjectUpdate struct {
+	jobStatus      project_update_lib.JobStatus
+	jobStatusError error
+	running        bool
+	ID             string
+}
+
```
lancewf: This object holds the current state of a project update.

components/nodemanager-service/pgdb/project_update.go
```diff
@@ -0,0 +1,281 @@
+package pgdb
```
lancewf: This is where the majority of the work is done for the project update.

components/nodemanager-service/pgdb/project_update.go
```diff
@@ -0,0 +1,281 @@
...
+  n.id,
+  n.projects_data
+FROM nodes n
+`
+
+type NodeProjectData struct {
+	ID                   string `db:"id"`
+	Data                 string `db:"projects_data"`
+	ProjectDataKeyValues []*nodes.ProjectsData
+}
```
lancewf: The "JobCancel", "UpdateProjectTags", and "JobStatus" are the three functions that are called from the cereal project update service.

components/nodemanager-service/pgdb/project_update.go
```diff
@@ -0,0 +1,281 @@
...
+		if err != nil {
+			db.projectUpdateFail(err)
+			return
+		}
+
+		db.projectUpdateStatusUpdate(float32((index + 1)) / numberOfNodes)
+	}
+
+	db.projectUpdateComplete()
+}
```
lancewf: What this is doing is for each node in the node manager update their project IDs with the `projectRules map[string]*iam_v2.ProjectRules` rules. 

I am wondering if all this go code can be replaced with an SQL script.

components/nodemanager-service/tests/project_update_test.go
```diff
@@ -0,0 +1,2231 @@
+package manager
+
+import (
```
lancewf: A lot of tests :sweat_smile:

components/nodemanager-service/pgdb/project_update.go
```diff
@@ -0,0 +1,281 @@
...
+			return project_update_lib.JobStatus{}, db.ProjectUpdate.jobStatusError
+		}
+		return db.ProjectUpdate.jobStatus, nil
+	}
+
+	return project_update_lib.JobStatus{
+		Completed:             true,
+		PercentageComplete:    1,
+		EstimatedEndTimeInSec: 0,
+	}, nil
```
msorens: Consider adding an in-code comment explaining why if the id is no longer present that means the job has completed (just my guess).

components/nodemanager-service/pgdb/project_update.go
```diff
@@ -0,0 +1,281 @@
...
+			return
+		}
+
+		db.projectUpdateStatusUpdate(float32((index + 1)) / numberOfNodes)
+	}
+
+	db.projectUpdateComplete()
+}
+
+func (db *DB) projectUpdateStatusUpdate(percentageComplete float32) {
```
msorens: Maybe `updateProjectStatus` to be less wordy without loss of info.

components/nodemanager-service/pgdb/project_update.go
```diff
@@ -0,0 +1,281 @@
...
+
+func (db *DB) getAllNodes() ([]*NodeProjectData, error) {
+	var nodesDaos []*NodeProjectData
+	_, err := db.Select(&nodesDaos, selectNodesProjectData)
+
+	for _, node := range nodesDaos {
+		dbProjectsData := []*nodes.ProjectsData{}
+		err := json.Unmarshal([]byte(node.Data), &dbProjectsData)
+		if err != nil {
+			return nodesDaos, errors.Wrap(err, "fromDBNode unable to unmarshal projects data")
```
msorens: Should that be
```suggestion
			return nodesDaos, errors.Wrap(err, "getAllNodes unable to unmarshal projects data")
```

components/nodemanager-service/pgdb/project_update.go
```diff
@@ -0,0 +1,281 @@
...
+		switch condition.Attribute {
+		case iam_v2.ProjectRuleConditionAttributes_CHEF_SERVER:
+			values := node.getValues("chef_server")
+			if len(values) == 0 {
+				return false
+			}
+
+			if !stringutils.SliceContains(condition.Values, values[0]) {
+				return false
+			}
```
msorens: Consider more concisely:
```
return len(values) > 0 && stringutils.SliceContains(condition.Values, values[0])
```

components/nodemanager-service/pgdb/project_update.go
```diff
@@ -0,0 +1,281 @@
...
+			for _, projectRole := range condition.Values {
+				if stringutils.SliceContains(values, projectRole) {
+					foundMatch = true
+					break
+				}
+			}
+			if !foundMatch {
+				return false
+			}
```
msorens: Cannot be quite as concise in this case, but could do just
```
return foundMatch
```

components/nodemanager-service/tests/project_update_test.go
```diff
@@ -0,0 +1,2231 @@
...
+									Attribute: iam_v2.ProjectRuleConditionAttributes_ENVIRONMENT,
+									Values:    []string{"env1"},
+								},
+							},
+						},
+					},
+				},
+			},
+			originalProjectIDs: []string{"project9"},
+			expectedProjectIDs: []string{},
```
msorens: 😮 If this expectation is correct (getting no projects rather than retaining the original project), then the description is misleading. It does, in fact, update the project to none because of no match.

components/nodemanager-service/tests/project_update_test.go
```diff
@@ -0,0 +1,2231 @@
...
+			},
+			projectRules: map[string]*iam_v2.ProjectRules{
+				"project3": {
+					Rules: []*iam_v2.ProjectRule{
+						{
+							Type: iam_v2.ProjectRuleTypes_NODE,
+							Conditions: []*iam_v2.Condition{
+								{
+									Attribute: iam_v2.ProjectRuleConditionAttributes_ENVIRONMENT,
+									Values:    []string{"env1", "env2"},
```
msorens: Good! But suggest use env3, env1, env2, i.e., put env1 *not* as first or last. Otherwise, it might accidentally be only looking at the first (or the last) slot, but no one is going to write code to just look at a middle slot. 😁

components/nodemanager-service/tests/project_update_test.go
```diff
@@ -0,0 +1,2231 @@
...
+		expectedProjectIDs []string
+	}{
+		// Environment
+		{
+			description: "Environment - updating project",
+			projectsData: []*nodes.ProjectsData{
+				{Key: "environment", Values: []string{"env1"}},
+			},
+			projectRules: map[string]*iam_v2.ProjectRules{
+				"project3": {
```
msorens: Suggest using a named constant like `targetProject` for this line and L57 to make the intent a bit clearer.

components/nodemanager-service/tests/project_update_test.go
```diff
@@ -0,0 +1,2231 @@
...
+	cases := []struct {
+		description        string
+		projectsData       []*nodes.ProjectsData
+		projectRules       map[string]*iam_v2.ProjectRules
+		originalProjectIDs []string
+		expectedProjectIDs []string
+	}{
+		// Environment
+		{
+			description: "Environment - updating project",
```
msorens: Would suggest make your test names more uniform in the format of
"`<characteristic> <expectation> <when what happens>`"
For this test you give the expectation but not the what, so suggest change to:
 "Environment - project updates with single rule, single matching condition"

Right now, you give the expectation explicitly just for the first 2 tests.
So test 3 should change from:
"Environment - Single rule two matching conditions"
to
"Environment - project updates with single rule, two matching conditions"

And test 4 from:
"Environment - Single rule, one non-matching condition"
to
"Environment - project dropped with single rule, one non-matching condition"

components/nodemanager-service/pgdb/project_update.go
lancewf: That change does not work because it would return after the first condition. All conditions must be true for the rule to be true. The conditions are 'or'ed together. 

I did combine the two if statements like below. 
`if len(values) == 0 || !stringutils.SliceContains(condition.Values, values[0]) {`

components/nodemanager-service/pgdb/project_update.go
lancewf: This has the same problem as above.

components/nodemanager-service/tests/project_update_test.go
lancewf: The description is incorrect. I will remove the original project.

lancewf:
@msorens Thank you for the review. I updated the code with your suggestions.

components/nodemanager-service/pgdb/project_update.go
```diff
@@ -0,0 +1,267 @@
...
+	if db.ProjectUpdate.ID == projectID {
+		db.ProjectUpdate.running = false
+	}
+	return nil
+}
+
+// UpdateProjectTags - start a project update
+func (db *DB) UpdateProjectTags(ctx context.Context,
+	projectRules map[string]*iam_v2.ProjectRules) ([]string, error) {
+	jobUUID, _ := uuid.NewV4()
```
phiggins: nitpick: This only ever uses the `.String()` version and repeats the conversion 3 times, so that could be done once here.

components/nodemanager-service/pgdb/project_update.go
```diff
@@ -0,0 +1,267 @@
...
+
+	// setup for new project update run
+	db.ProjectUpdate.jobStatusError = nil
+	db.ProjectUpdate.running = true
+	db.ProjectUpdate.jobStatus = project_update_lib.JobStatus{
+		Completed:             false,
+		PercentageComplete:    0,
+		EstimatedEndTimeInSec: 0,
+	}
+	db.ProjectUpdate.ID = jobUUID.String()
```
phiggins: This code knows a lot about the ProjectUpdate struct, often specifying the same values for the same fields, etc. It might make sense to move a lot of the ProjectUpdate state manipulation where that struct is defined, for example, here this is resetting the ProjectUpdate to some initial state which could be encapsulated with the struct definition.

phiggins: In particular, code like this block is repeated several times and could be a constant where the struct is defined.

```
project_update_lib.JobStatus{
		Completed:             false,
		PercentageComplete:    0,
		EstimatedEndTimeInSec: 0,
	}
```

components/nodemanager-service/pgdb/project_update.go
```diff
@@ -0,0 +1,267 @@
...
+		db.projectUpdateFail(err)
+		return
+	}
+
+	numberOfNodes := float32(len(nodes))
+	for index, node := range nodes {
+		if !db.ProjectUpdate.running {
+			db.projectUpdateComplete()
+			return
+		}
```
phiggins: This if statement doesn't need to be inside the for loop does it? Is this to check for that status changing while doing the update? Does this mean if I cancel an update it could get partially applied?

components/nodemanager-service/pgdb/project_update.go
```diff
@@ -0,0 +1,267 @@
...
+			if !foundMatch {
+				return false
+			}
+		default:
+			return false
+		}
+	}
+
+	return true
+}
```
phiggins: There is probably a way to translate this to SQL but it'd probably be pretty messy. I think this being in golang is fine unless we run into performance problems or something.

components/nodemanager-service/pgdb/db.go
```diff
@@ -12,12 +12,21 @@ import (
...
 type DB struct {
 	*gorp.DbMap
+	ProjectUpdate *ProjectUpdate
 }
```
phiggins: It seems weird to me that the nodemanager's DB struct always has a ProjectUpdate in it now, but I can't really suggest a better way to do this.

components/nodemanager-service/pgdb/project_update.go
lancewf: This if statement is needed to allow the process to be canceled. That is correct, if the process is cancled it would be partially applied. This is the same for the Elasticsearch updates for the ingest-service and compliance-service.

lancewf:
@phiggins Thank you for the review. I update the code with your suggestions.

PR #1773, opened 2019-10-02
Turn off periodic test during integration test to reduce flakiness

kmacgugan:
We have seen some flakey tests around disconnected services within
the applications service, it is suspected that the periodic job that always
runs is interfereing with the tests every once in a while.

The mark disconnected services periodic job is a mandatory job,
but we can turn it off before and after running integration tests during the
suite setup. Then we turn it on for testing the periodic jobs again.

There is no way to test this, we just keep an eye out for failing tests over the next week or so. 

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/applications-service/pkg/server/periodic.go
+++ b/components/applications-service/pkg/server/periodic.go
@@ -237,6 +237,29 @@ func (j *JobScheduler) DisableDeleteDisconnectedServicesJob(ctx context.Context)
 
 }
 
+func (j *JobScheduler) EnableDisconnectedServicesJob(ctx context.Context) error {
+	err := j.CerealSvc.UpdateWorkflowScheduleByName(
+		ctx,
+		DisconnectedServicesScheduleName, DisconnectedServicesJobName,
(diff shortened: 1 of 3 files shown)


components/applications-service/pkg/server/periodic.go
```diff
@@ -237,6 +237,29 @@ func (j *JobScheduler) DisableDeleteDisconnectedServicesJob(ctx context.Context)
...
+		return errors.Wrap(err, "failed to set disconnected_services job to enabled")
+	}
+	return nil
+}
+
+func (j *JobScheduler) DisableDisconnectedServicesJob(ctx context.Context) error {
+	err := j.CerealSvc.UpdateWorkflowScheduleByName(
+		ctx,
+		DisconnectedServicesScheduleName, DisconnectedServicesJobName,
+		cereal.UpdateEnabled(false))
```
lancewf: I did not know you could do that. Nice!

Issue #1736, opened 2019-09-30
filter report by control tag


vjeffrey:
## User Story

When viewing/download the report for a node, results should respect the applied control tag filter(s).

## Definition of Done
Node report results respect control tag filter(s).

## Demo Script / Repro Steps
use the following profile to run a scan job: 
[my-test-tagging-0.2.0.tar.gz](https://github.com/chef/automate/files/3669530/my-test-tagging-0.2.0.tar.gz)
download the report for the node: see 3 controls
open scan results for the node: see 3 controls
filter by control tag
download the report for the node: see 1 control
open scan results for the node: see 1 control

PR #1776, opened 2019-10-02
only return controls matching control tag filter when fetching single node report

vjeffrey:
This change modifies the get report call to only return controls matching the provided control tag filters.  
In order to be able unit test the functionality, I pulled a portion of the code into a separate function.

https://github.com/chef/automate/issues/1736

### :+1: Definition of Done
A single node report filtered by control tag returns only the controls matching the given control tag.

```
build components/compliance-service
start_all_services
cd components/compliance-service/test_data/audit_reports/
./send_to_data_collector.sh https://a2-dev.test token_value
filter by control tag
open scan results on one of those nodes, see only controls with tags you filtered on
open node report (node details) on one of those nodes, see only controls with tags you filtered on
```

- [x] Tests added/updated?
- [ ] Docs added/updated? (N/A)

Screen Shot 2019-10-02 at 17 46 30


vjeffrey:
opening as a draft pr while i test it out a bit more

components/compliance-service/api/reporting/reporting.proto
```diff
@@ -175,6 +175,10 @@ message Control {
 	repeated Result results = 7;
 	repeated Ref refs = 8;
 	map<string, string> tags = 9;
+	map<string, TagValues> string_tags = 10;
```
vjeffrey: new field here b/c i couldn't squeeze the tags into the existing tags field due to the one:many relationship

components/compliance-service/api/tests/03_reports_spec.rb
```diff
@@ -712,6 +712,19 @@ def reporting
...
 
+    # Get a specific report; ensure only the control with the specified filter tag is returned
+    res = GRPC reporting, :read_report, Reporting::Query.new(id: '44024b50-2e0d-42fa-a57c-dddddddddddd',       
+        filters: [
+            Reporting::ListFilter.new(type: 'control_tag:tag1', values: ['value-1']),
+        ]
+    )
+    assert_equal(Reporting::Report, res.class)
+    assert_equal(Google::Protobuf::RepeatedField, res['profiles'].class)
+    if res['profiles'].is_a?(Google::Protobuf::RepeatedField)
```
vjeffrey: my docker es (non-studio) was being real punky so not sure if this test will pass or not/didn't run it locally

components/compliance-service/inspec/types.go
```diff
@@ -42,6 +42,7 @@ type Control struct {
 	Refs           *json.RawMessage             `json:"refs"`
 	Tags           *json.RawMessage             `json:"tags"`
 	Results        []*reportingapi.Result       `json:"results,omitempty"`
+	StringTags     []*json.RawMessage           `json:"string_tags"`
```
vjeffrey: 🤷‍♀ i kinda went with json rawmsg here cause it covers everything, but happy to change that if ppl want

components/compliance-service/reporting/relaxting/reports.go
```diff
@@ -429,49 +429,19 @@ func (backend *ES2Backend) GetReport(esIndex string, reportId string,
...
-								var jsonRefs []*reportingapi.Ref
-								refs, _ := json.Marshal(profileControl.Refs)
-								json.Unmarshal(refs, &jsonRefs) // nolint: errcheck
-								convertedControl.Refs = jsonRefs
-								// store controls to returned report
-								convertedControls = append(convertedControls, &convertedControl)
-							}
+						// store controls to returned report
+						convertedControl := convertControl(profileControlsMap, reportControlMin, filters)
+						if convertedControl != nil {
```
vjeffrey: don't wanna add nil controls to the array!

components/compliance-service/reporting/relaxting/reports.go
```diff
@@ -530,6 +499,87 @@ func (backend *ES2Backend) GetReport(esIndex string, reportId string,
 	return report, errorutils.ProcessNotFound(nil, reportId)
 }
 
+func convertControl(profileControlsMap map[string]*reportingapi.Control, reportControlMin ESInSpecReportControl, filters map[string][]string) *reportingapi.Control {
```
vjeffrey: i'm kinda tempted to also break the profile conversion code out of that big function so i can unit test it, but i'd rather not bloat this change with that. thinking i'll open an issue for it as cleanup/tech debt

components/compliance-service/api/tests/03_reports_spec.rb
vjeffrey: this test is failing, so i'll get my local makefile dev figured out and see what's going on

components/compliance-service/reporting/relaxting/reports.go
```diff
@@ -530,6 +499,87 @@ func (backend *ES2Backend) GetReport(esIndex string, reportId string,
...
+						for _, val := range filterVals {
+							if contains(tagVal.Values, val) || val == "null" {
+								matchingTagFound = true
+								break
+							}
+						}
+					}
+
+				}
+			}
```
lancewf: This could be moved to its own function like below. 
```
func doesControlTagMatchFilter(filters map[string][]string,
	jsonTags map[string]*reportingapi.TagValues) bool {
	for filterKey, filterVals := range filters {
		if !strings.HasPrefix(filterKey, "control_tag") {
			return true
		}
		trimmed := strings.TrimPrefix(filterKey, "control_tag:")
		if tagVal, ok := jsonTags[trimmed]; ok {
			for _, val := range filterVals {
				if contains(tagVal.Values, val) || val == "null" {
					return true
				}
			}
		}
	}
	return false
}
```
Then in the function you would have the below
```
	if !doesControlTagMatchFilter(filters, jsonTags) {
		return nil
	}
```

components/compliance-service/reporting/relaxting/reports.go
```diff
@@ -530,6 +499,87 @@ func (backend *ES2Backend) GetReport(esIndex string, reportId string,
 	return report, errorutils.ProcessNotFound(nil, reportId)
 }
 
+func convertControl(profileControlsMap map[string]*reportingapi.Control, reportControlMin ESInSpecReportControl, filters map[string][]string) *reportingapi.Control {
+	profileControl := profileControlsMap[reportControlMin.ID]
+
+	if profileControl != nil {
```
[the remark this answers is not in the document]
vjeffrey: 👍 good call!

components/compliance-service/reporting/relaxting/reports.go
vjeffrey: 👍

components/compliance-service/api/tests/03_reports_spec.rb
phiggins: Same comment here, we don't need this if statement: https://github.com/chef/automate/pull/1672#discussion_r330807475

components/compliance-service/reporting/relaxting/reports.go
phiggins: There doesn't appear to be an `else` on this condition, so you could remove a level of nesting by putting a guard in front checking for nil and returning.

PR #1778, opened 2019-10-03
lib/datalifecycle: w.Fail always needs to be returned

stevendanna:


stevendanna:
Note: this isn't a major issue because the only way EnqueueTask can fail here is if options are passed.  So this would have never hit the error case, but let's defend against future changes.

PR #1802, opened 2019-10-04
cereal: remove backend package

stevendanna:
### Description

When working on #1777 we realized extending our new name types to the Driver
interface would be difficult because directly using them would result in circular dependencies.

Before this change, the import pattern in the cereal library looked like this:

```
+-----------+         +-----------+
|           |         |           |
|  cereal   +-------->+  backend  |
|           |         |           |
+-----+-----+         +---+-------+
      ^                   ^
      |                   |
      |                   |
      |                +--+-------------+
      +----------------+                |
                       | grpc/postgres  |
                       |                |
                       +----------------+
```

With the `backend` package holding the Driver interface and an assortment of types for that interface. This led to a few problems:

1) The Driver interface could not refer to any times in the cereal package.
2) Users occasionally had to import yet-another package when they wanted to refer to Driver.

While we can solve (1) by introducing yet another type and converting between them, the `backend` package was really just a hold-over from when the library only had one backend.  Since the package only held a handful of types, here we've opted to remove it entirely, moving all of its types into `cereal` directly.

### Known Issues

I've done a very naive merging of the two packages.  Since we had types that were named the same in both packages, I've renamed them with pretty bad names:

  - backend.WorkflowInstance -> cereal.WorkflowInstanceData
  - backend.TaskResult -> cereal.TaskResultData
  - backend.Task -> cereal.TaskData

I believe we can probably clean this up substantially and remove some of the wrapper types that we have.  However, the goal here was to do mostly mechanical transformations.

### Future Work

- Apply WorkflowName and TaskName types to Driver interface
- Investigate refactoring WorkflowInstanceData, TaskResultData, and TaskData

### Definition of Done

- All tests pass, no behavior changes, one less packages.


stevendanna:
Review Notes:  I'm not sure we really want to do this.  The `backend` package has annoyed me for a bit, but there is a long way to go on top of this to really make it pretty.

stevendanna:
Hrm, it doesn't seem to be running my tests on this non-master PR ....

PR #1803, opened 2019-10-04
event-service: fix minor lint violations

stevendanna:
This service doesn't have any unit/static tests so I haven't
integrated the linter, but I noticed that there were only a couple of
violations using our current linter configuration and they were mostly
in test code so they were pretty low risk to fix.

### DIFF
--- a/components/event-service/cmd/event-service/commands/serve.go
+++ b/components/event-service/cmd/event-service/commands/serve.go
@@ -5,13 +5,14 @@ import (
 
 	"fmt"
 
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


components/event-service/cmd/event-service/commands/serve.go
```diff
@@ -36,7 +37,7 @@ func newServeCmd() *cobra.Command {
 
 			closer, err := tracing.NewGlobalTracer("event-service")
 			if err != nil {
-				errors.Wrap(err, "starting tracer for event-service")
+				return errors.Wrap(err, "starting tracer for event-service")
```
stevendanna: This is the only meaningful change.  Previous we were ignoring this error.

PR #1804, opened 2019-10-04
notifications-client: apply linter

stevendanna:
Most of these violations are in test code.  I've wired the linter up to 
CI since it was easy to slot into the existing unit tests.

### DIFF
--- a/components/notifications-client/builder/ccr.go
+++ b/components/notifications-client/builder/ccr.go
@@ -6,9 +6,10 @@ import (
 	"strings"
 	"time"
 
+	_struct "github.com/golang/protobuf/ptypes/struct"
+
 	chef "github.com/chef/automate/api/external/ingest/request"
 	. "github.com/chef/automate/components/notifications-client/api"
-	_struct "github.com/golang/protobuf/ptypes/struct"
 )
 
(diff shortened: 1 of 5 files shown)


components/notifications-client/builder/ccr.go
```diff
@@ -30,7 +31,7 @@ func ChefClientConverge(url string, run *chef.Run) (*Event, error) {
 		}
 
 		err_desc := err.GetDescription()
-		if err == nil {
+		if err_desc == nil {
```
phiggins: If this value isn't used anywhere, you could do `if err.GetDescription() == nil {` and avoid the variable altogether.

stevendanna: It's used below at line 50 of the new file.  Although the use has me thinking this error check isn't needed?  The whole point of the Get() function is that they work on nil and you eventually get a zero-value.  Is it really an error if we didn't get sent a description?  Perhaps we should remove this check completely?

phiggins: Oops, I think I misunderstood this change. To avoid unintentional changes I'd probably go with what you've got now. 😅

Issue #1781, opened 2019-10-03
add controls count to reporting/summary/stats api call


vjeffrey:
## User Story

The ui uses a reporting stats api call to load the totals for each of the tabs on the compliance reporting page.  This is useful b/c it means we can use that one api call to populate the totals, and not make the call to get the list of nodes/profiles until a user selects that tab.
with the addition of a controls list api/tab, it came to light that we need to add a total for controls to that api call to maintain the pattern.
see: https://github.com/chef/automate/issues/1310#issuecomment-537651075

## Definition of Done
controls total count is added to `/compliance/reporting/stats` api call.

vjeffrey:
![Screen Shot 2019-10-22 at 11.15.20.png](https://images.zenhubusercontent.com/5cfebc28d99ac71fbfbdcb73/820dd296-b35c-48d1-8cee-e39ab4eb7132)

thank you!

PR #1806, opened 2019-10-04
deprecate nodes in favor of nodes_cnt in proto

rickmarry:
This change deprecates the "nodes" field from the compliance Stats proto struct.  This "nodes" attribute is of type int64, which ultimately ends up being serialized in json response as a string (unlike all other stats which are serialized as numbers).  This would be fine if we really needed a number that large to represent node count.  In reality, 2 billion is enough and therefore, int32 is sufficient for node count.

This change introduces nodes_cnt which will be of type int32. 

- [ ] Tests added/updated?
- [ ] Docs added/updated?


vjeffrey:
https://github.com/chef/automate/issues/1781

components/automate-gateway/api/compliance/reporting/stats/stats.proto
```diff
@@ -122,10 +122,14 @@ message NodeSummary {
...
-	int32 environments = 3;
-	int32 profiles = 4;
+	int32 nodes_cnt = 1;
+	//int64 types render into string types when serialized to satisfy all browsers
+	//we don't really need for nodes counts to be int64 as int32 limits us to 2billion nodes which is plenty for now
+	//we are therefore deprecating nodes and favor nodesCnt
+	int64 nodes = 2 [deprecated = true];
+	int32 platforms = 3;
+	int32 environments = 4;
+	int32 profiles = 5;
```
stevendanna: Changing the IDs here is technically a breaking API change.  Because of how we build it probably don't matter, but the "protobufy" way to do this would be to leave all the existing IDs as is and add nodes_cnt to the end.

components/compliance-service/api/stats/stats.proto
```diff
@@ -56,12 +56,18 @@ message NodeSummary {
...
-	int32 environments = 3;
-	int32 profiles = 4;
+	int32 nodes_cnt = 1;
+	//int64 types render into string types when serialized to satisfy all browsers
+	//we don't really need for nodes counts to be int64 as int32 limits us to 2billion nodes which is plenty for now
+	//we are therefore deprecating nodes and favor nodesCnt
+	int64 nodes = 2 [deprecated = true];
+	int32 platforms = 3;
+	int32 environments = 4;
+	int32 profiles = 5;
```
stevendanna: Same issue here re field IDs and compatibility.

components/automate-gateway/api/compliance/reporting/stats/stats.proto
rickmarry: does that really matter?  who uses the ID?

stevendanna: The IDs are used in the serialization of protobuf message on the wire.  So, if you want to be able to receive messages from a client compiled with the previous version of your protobuf, you need to keep the IDs consistent.

rickmarry: ok, changing now. thanks

stevendanna: FWIW, I believe that changing the numeric type of an item (i.e. `int64` -> `int32`) is "compatible" in the wire-format sense so you can sometimes get away with just changing one numeric type to another, but I haven't thought about it hard enough to know if that would be OK here.

rickmarry: > FWIW, I believe that changing the numeric type of an item (i.e. `int64` -> `int32`) is "compatible" in the wire-format s.. 
 
yeah, the problem is that for API users that are deserializing the json and using the nodes attribute that is currently of type string (because that's what it does to int64), and doing whatever subsequent type conversion to turn it to a number, we would break them if we just changed this to int32 because it would change the json representation to a number

PR #1808, opened 2019-10-04
ingest-service: apply linter

stevendanna:
- nakedret: I disabled most of these as I didn't want to touch the
  functions they were in.

- errcheck: I tried to address all the non-defer cases here that
  looked important, please review carefully.

### DIFF
--- a/components/ingest-service/backend/elastic/jobs.go
+++ b/components/ingest-service/backend/elastic/jobs.go
@@ -13,9 +13,10 @@ import (
 	"encoding/json"
 	"time"
 
+	"github.com/olivere/elastic"
+
 	"github.com/chef/automate/components/ingest-service/backend"
 	"github.com/chef/automate/components/ingest-service/backend/elastic/mappings"
-	"github.com/olivere/elastic"
 )
 
 // MarkNodesMissing marks all nodes that have passed the specified threshold as 'missing'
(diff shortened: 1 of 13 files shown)


components/ingest-service/backend/elastic/jobs.go
```diff
@@ -175,9 +176,9 @@ func (es *Backend) DeleteMarkedNodes(ctx context.Context, threshold string) (upd
 	}
 
 	// Clear Scroll
-	scrollService.Clear(ctx)
+	scrollService.Clear(ctx) // nolint: errcheck
 
-	return
+	return // nolint:nakedret
```
stevendanna: This function and some of the others are maybe worth reviewing.  Because we shadow `err` in the loop body, it is pretty hard to understand what error conditions here actually result in an error being returned.

components/ingest-service/backend/elastic/migration.go
```diff
@@ -260,9 +259,7 @@ func (es *Backend) ReindexNodeStateA1(ctx context.Context, a1NodeStateIndexName
 		}
 	}
 
-	es.RefreshIndex(ctx, mappings.NodeState.Index)
-
-	return err
+	return es.RefreshIndex(ctx, mappings.NodeState.Index)
```
stevendanna: Wasn't sure if we wanted to ignore errors from this refresh or not.

components/ingest-service/server/jobs.go
```diff
@@ -112,7 +112,10 @@ func (s *SingleTaskWorkflow) OnStart(w cereal.WorkflowInstance, ev cereal.StartE
 		return w.Fail(err)
 	}
 
-	w.EnqueueTask(s.taskName, params)
+	err = w.EnqueueTask(s.taskName, params)
+	if err != nil {
+		return w.Fail(err)
+	}
```
stevendanna: This code goes away in a different PR.

components/ingest-service/backend/elastic/migration.go
phiggins: So the reasoning here is that `err` would have already been returned if it was set? Makes sense.

PR #1812, opened 2019-10-04
[automate-825] IAM v2 Reference Doc Updates

msorens:

_Note to docs team: Feel free to peruse if you like, but you are not the intended audience for this partial update._ 😁 

Update the reference.md document file for IAM v2.1 beta.
This includes a major rewrite based on reviewing other API references out there.
Each section starts with a table summarizing the HTTP Requests, then an example JSON for the given resource, then notes on each method. I removed much redundancy by cutting out the curl call on every item, making the text tighter and easier to digest.

TODO: 
API calls for users and teams

NA

### :+1: Definition of Done
Reference page describes all API calls.

Launch hugo locally on your machine per the instructions in components/automate-chef-io/README.md, then browse to http://localhost:1313/docs/iam-v2-overview/

### DIFF
--- a/components/automate-chef-io/content/docs/iam-v2-api-reference.md
+++ b/components/automate-chef-io/content/docs/iam-v2-api-reference.md
@@ -10,60 +10,82 @@ toc = true
     weight = 30
 +++
 
-This guide shows you how to manage Chef Automate IAM v2 policies and roles from the command line.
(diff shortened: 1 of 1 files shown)


msorens:
Oops! Was supposed to be based against working branch, not master, so closing.

PR #1813, opened 2019-10-04
[automate-825] IAM v2 Reference Doc Updates

msorens:

_Note to docs team: Feel free to peruse if you like, but you are not the intended audience for this partial update._ 😁 

Update the reference.md document file for IAM v2.1 beta.
This includes a major rewrite based on reviewing other API references out there.
Each section starts with a table summarizing the HTTP Requests, then an example JSON for the given resource, then notes on each method. I removed much redundancy by cutting out the curl call on every item, making the text tighter and easier to digest.

Launch hugo locally on your machine per the instructions in components/automate-chef-io/README.md, then browse to http://localhost:1313/docs/iam-v2-overview/


components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -10,60 +10,145 @@ toc = true
...
-
-1. Upgrade to IAM v2 using the [IAM v2 User's Guide]({{< relref "iam-v2-guide.md#upgrade-to-iam-v2" >}}).
-
-2. _Optional_ Install the [jq](https://stedolan.github.io/jq/) JSON processor to ensure readable output.
-
-3. Generate a new admin token for IAM v2.
-   (The v1 command `chef-automate admin-token` will not work for v2, though any existing v1 admin tokens will still work.) Use:
+<!--The TOC below is auto-generated by the VSCode "Markdown TOC" plugin; do not edit manually.
+With the plugin installed, this TOC will update automatically when you save the file! -->
+<!-- TOC depthFrom:2 depthTo:3 -->
```
msorens: Per this comment, I have included an auto-generated Table-of-Contents. I am not aware of this is good practice in a Hugo document, so if Hugo has an alternate way of doing this, great. But it is vital in this file to help folks quickly find what they want.

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +159,556 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
+Deleting a token is permanent and cannot be undone.
+
+*In the browser:* **Settings**  >> **API Tokens** >> [open control menu for target token] >> Delete Token
+
 ## Restoring Admin Access
 
 While we have [safeguards]({{< relref "iam-v2-overview.md#policy-types" >}}) to prevent it, it is possible to lock yourself out of Chef Automate.
 If you have root access to the node where Chef Automate is installed, use the following commands to restore admin access:
 
 <!-- Wording/Replace password/replace token/v1 commands won't work on c2 -->
```
msorens: No idea what this comment is trying to say. If someone does, say so, otherwise would like to delete it.

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +159,556 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
+
+*In the browser:* **Settings**  >> **API Tokens** >> [select a token] >> Details
+
+### Deleting a Token
+
+Deleting a token is permanent and cannot be undone.
+
+*In the browser:* **Settings**  >> **API Tokens** >> [open control menu for target token] >> Delete Token
+
 ## Restoring Admin Access
```
msorens: This section sticks out as the only section in this document that is not really part of the API reference; suggest it move to the guide.md doc.

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +159,556 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
+
+**Example payload:**
+
+```json
+{
+  "id": "team-1",
+  "user_ids": [
+    "1a5a63d7-0b14-465a-aec1-5eea08a72343",
+    "6c4e3e8a-fade-4a9c-ac9f-99b86677530b"
+  ]
```
msorens: It seems peculiar that for team membership we list users by GUIDs whereas for policy membership users (and teams and tokens, of course) are listed by id. 🤷‍♂

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +159,556 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
       "effect": "ALLOW",
       "role": "editor"
     }
+  ],
+  "projects": [
+    "admin-group-1",
+    "it-support",
   ]
 }
 ```
```
msorens: This API reference, of course, needs to list all the properties for a given resource (like a policy in this instance). As a first pass, I chose to do this with a sample like the one just above. I am not sure that is adequate though--should we also have a table that lists each property and what it means?

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -10,60 +10,145 @@ toc = true
...
-This example policy is made up of two statements, each enforcing a different set of permissions.
-The first policy statement grants users list view access.
-The second statement grants editor access via the `editor` role.
-You can see the actions contained in the `editor` role in the [Roles section]({{< relref "iam-v2-api-reference.md#roles" >}}).
+Here is a summary:
+
+Type               | May be empty ? | Include * ? | May specify (unassigned) ?
+-------------------|----------------|-------------|------------
+top-level projects |    yes         |   no        |  no
+statement projects |    no          |   yes       |  yes
```
msorens: Sure feels like we should have two different terms instead of just "projects". 🤷‍♂

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -10,60 +10,145 @@ toc = true
...
+
+## Policies
+
+Method | HTTP request              | Description
+-------|---------------------------|------------
+list   | GET /policies             | lists both *Chef-managed* and *Custom* policies
+get    | GET /policies/***id***    | gets the specified policy
+create | POST /policies            | creates a *Custom* policy
+update | PUT /policies/***id***    | updates a *Custom* policy
+delete | DELETE /policies/***id*** | deletes a *Custom* policy
```
msorens: Each section provides a table like this. The first column in this policies table has dubious value, because its values are echoed in the third column. And that is true of most such tables, **except** for Policy Membership and for Users. Opinions on deleting the first column?

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -10,60 +10,145 @@ toc = true
...
+  - [Updating a Token](#updating-a-token)
+  - [Deleting a Token](#deleting-a-token)
+- [Restoring Admin Access](#restoring-admin-access)
+
+<!-- /TOC -->
+
+## General Notes
+
+1. If you do not have one already, generate an admin token for IAM v2.
+   (Neither The v1 command `chef-automate admin-token` nor any v1 admin tokens will work for v2.)
```
blakestier: Is this true? I think those v1 tokens _do_ work...

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -10,60 +10,145 @@ toc = true
...
+  https://{{< example_fqdn "automate" >}}/apis/iam/v2beta/policies?pretty
 ```
 
-The output includes both `Chef-managed` and `custom` policies.
-Policies migrated from IAM v1 are included as `custom` policies.
-Policy names that begin with `[Legacy]` are default policies migrated from v1.
-We recommend deleting these legacy policies and setting up new IAM v2 policies in their place.
-
-In the browser: Navigate to the *Policies* page under the **Settings**  tab.
+For those API methods that take JSON data (typically `create` and `update` methods)
```
blakestier: @msorens, I'm curious about the deleted information, here. I don't see it in the issue, but maybe I'm missing something?

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -10,60 +10,145 @@ toc = true
...
 ```
 
-In the browser: Navigate to the *Policies* page under the **Settings**  tab, and then select an individual policy to open its details.
+## Notes on the Projects Property
 
-### Creating a Policy
+Teams, tokens, roles, and policies all contain a top-level `projects` property; you will see that in the example JSON for each resource in the following sections.
+This property associates the particular IAM resource with one or more projects.
+Users must have permissions for those projects to be able to view or modify those resources.
+This `projects` property may be empty, which would give access only to users included on policies specifying the special `(unassigned)` project designation.
```
blakestier: I think
> the special `(unassigned)` project designation 
is bleeding some implementation detail. Maybe 
`If the projects property is left empty, that resource will be categorized as `unassigned`.  ?

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -10,60 +10,145 @@ toc = true
...
+This `projects` property may **not** specify `*` to indicate all projects.
```
blakestier: The passive language is also distracting. We take advantage of saying "you" elsewhere in the docs, and perhaps should here as well. ("You cannot set the projects field to be `*`")

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -10,60 +10,145 @@ toc = true
...
+-------------------|----------------|-------------|------------
+top-level projects |    yes         |   no        |  no
+statement projects |    no          |   yes       |  yes
+
+## Policies
+
+Method | HTTP request              | Description
+-------|---------------------------|------------
+list   | GET /policies             | lists both *Chef-managed* and *Custom* policies
+get    | GET /policies/***id***    | gets the specified policy
```
blakestier: for consistency, should this be `gets either a *Chef-managed* or *Custom* policy`?

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -10,60 +10,145 @@ toc = true
...
+create | POST /policies            | creates a *Custom* policy
+update | PUT /policies/***id***    | updates a *Custom* policy
+delete | DELETE /policies/***id*** | deletes a *Custom* policy
+
+**Example Policy:**
+
+This policy shows two statements, each enforcing a different set of permissions.
+The first policy statement grants list access to view users.
```
blakestier: this sentence could use some formatting (like "LIST" access) or something.

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -10,60 +10,145 @@ toc = true
...
+The second statement grants access to all the actions comprising the `editor` role.
+Note that you may specify actions either inline with **actions** or using a **role** reference.
```
blakestier: Is this example confusing bc it has _both_ role _and_ actions? Perhaps we want two different versions of this policy to highlight the convenience of roles and prevent redundancy where possible.

Or maybe we want to encourage them adding on? 🤔

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +159,556 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
 }
 ```
 
-Assuming you stored the above JSON in the file "policy.json", pass it to `curl` to create the policy:
+### Listing Policies
 
-```bash
-curl -s -H "api-token: $TOKEN" -d @policy.json https://{{< example_fqdn "automate" >}}/apis/iam/v2beta/policies?pretty
-```
+The output includes both *Chef-managed* and *Custom* policies.
```
blakestier: oh here it is 😄

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +159,556 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
+### Getting a Policy
 
-* You can modify the *membership* of both Chef-managed and custom policies; see [Policy Membership]({{< relref "iam-v2-api-reference.md#policy-membership" >}}) for more information.
-* You can modify the *definition* of custom policies but not Chef-managed policies.
-* You can change the policy name, members, or statements, but you cannot change the policy ID.
-* Modifying any part of a policy resets *all* of its properties.
-  Properties that aren't included in the command are reset to empty values.
-  Thus, you must supply all of a policy's properties, not just the ones you wish to update.
+In order to fetch a policy on the command-line, you must know its ID.
+This may be obtained by first fetching the list of all policies and getting the `id` field of the policy of interest.
```
blakestier: We wrote this before we'd built out the UI, so now we have some easier ways to do this 💁

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +159,556 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
-        "iam:users:list"
-      ]
-    }
- ]
-}
-```
+### Updating a Policy
 
-To update your policy, replace the last component of the path in the command with the ID of the target policy.
+- You can modify the *membership* of both Chef-managed and custom policies; see [Policy Membership]({{< relref "iam-v2-api-reference.md#policy-membership" >}}) for more information.
```
blakestier: sometimes we capitalize `custom` and sometimes we don't . . . 😑

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +159,556 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
+- You can modify the *definition* of custom policies but not Chef-managed policies.
+- The policy ID is immutable; it can only be set at creation time.
+- The update operation modifies *all* policy properties, not just the ones you specify.
+  Properties that you do not specify are reset to empty values.
+  Thus, you should supply all of a policy's properties, not just the ones you wish to update.
```
blakestier: I'd lead with this since it's 🚨

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +159,556 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
+Using `POST` lets you add members to (or remove members from) an existing membership list.
 For more information, see [Adding Members to Policies]({{< relref "iam-v2-api-reference.md#adding-members-to-policies" >}}) and [Removing Members from Policies]({{< relref "iam-v2-api-reference.md#removing-members-from-policies" >}}).
 
-#### Setting Policy Membership
+Method | HTTP request                           | Description
+-------|----------------------------------------|------------
+list   | GET /policies/***id***/members         | lists the membership
+update | PUT /policies/***id***/members         | replaces the membership
+add    | POST /policies/***id***/members:add    | add to the membership
+remove | POST /policies/***id***/members:remove | remove from the membership
```
blakestier: Hmm, if we have it here, maybe it's weird to delete the first column above

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +159,556 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
 
-From the command line, specifying a complete list of members is often simpler than keeping track of a policy's membership and adding or deleting specific members (whether users, team, or tokens).
-This command *replaces* the policy's membership.
```
blakestier: I liked the deleted warning: `This command *replaces* the policy's membership.`

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +159,556 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
-```
+### Updating All Members on a Policy
 
-In the browser: Navigate to the *Policies* page under the **Settings** tab. and select an individual policy to open its detail view.
-Select the **Members** to view the current membership.
-Use the **Add Members** button to open a list of candidate members.
-Select any members that should be added to the policy.
-Use the **Add # Members** button to complete the operation.
+Specifying a complete list of members is often simpler than keeping track of a policy's membership and adding or deleting specific members (whether users, team, or tokens).
+Use this HTTP request to replace the policy's membership with a full, new list.
```
blakestier: > Use this HTTP request to replace the policy's membership ~with a full, new list~.

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +159,556 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
 
-In the IAM v2 beta, policies must be deleted from the command line.
-
-See [Listing Policies]({{< relref "iam-v2-api-reference.md#listing-policies" >}}) to fetch a list of all your policies and locating the relevant policy ID.
-Use the policy ID as the final path component to delete a policy:
-
-```bash
-curl -s -H "api-token: $TOKEN" -X DELETE https://{{< example_fqdn "automate" >}}/apis/iam/v2beta/policies/{policy-id}?pretty
-```
+Specify just members to remove from a policy's membership via this HTTP request.
```
blakestier: ~just~

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +159,556 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
 A role encapsulates a list of actions.
-IAM v2 provides several default Chef-managed roles, which are essential to the operation of Chef Automate and cannot be altered.
-Roles you create are Custom roles.
+IAM v2 provides several default *Chef-managed* roles, which are essential to the operation of Chef Automate and cannot be altered.
+Roles you create are *Custom* roles.
+
+Method | HTTP request              | Description
+-------|---------------------------|------------
+list   | GET /roles                | lists both *Chef-managed* and *Custom* roles
+get    | GET /roles/***id***       | gets the specified role
```
blakestier: same note as above

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +159,556 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
+
+### Creating a Role
+
+Create a role by composing JSON with all the necessary role properties (see the beginning of the section above).
+
+### Updating a Role
+
+The update operation modifies *all* role properties, not just the ones you specify.
+Properties that you do not specify are reset to empty values.
+Thus, you should supply all of a role's properties, not just the ones you wish to update.
```
blakestier: Again, I'd flip this based on importance: 

>You should supply all of a role's properties, not just the ones you wish to update. Because the update operation modifies *all* role properties, not just the ones you specify, _properties that you do not specify are reset to empty values_.

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +159,556 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
+get    | GET /projects/***id***    | gets the specified role
+create | POST /projects            | creates a project
+update | PUT /projects/***id***    | updates a project
+delete | DELETE /projects/***id*** | deletes a project
+
+**Example Project:**
+
+```json
+{
+  "name": "Test Project",
```
blakestier: I wonder if we could come up with some example names other than _test role_ or _test project_ that might be more revealing/clarifying

components/automate-chef-io/content/docs/iam-v2-api-reference.md
blakestier: That _is_ odd

blakestier: We use the `ID` to add members, though? That's kind of a problem given our lack of patchy updates . . .

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +159,556 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
+    "east-region",
+    "west-region"
+  ]
+}
+```
 
 ### Listing Tokens
 
-List all tokens with this command:
+The lists admin and non-admin tokens.
```
blakestier: The -> This

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +159,556 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
-* This command, which is also available on IAM v1, resets the local `admin` user's password and ensures that user is a member of the local `admins` team, which is a permanent member of the Chef-managed `Administrator` policy.
+This command, which is also available on IAM v1, resets the local `admin` user's password and ensures that user is a member of the local `admins` team, which is a permanent member of the Chef-managed `Administrator` policy.
 
 ```bash
-  chef-automate iam admin-access restore NEW_PASSWORD
+  chef-automate iam admin-access restore <your new password here>
 ```
 
-* Generate a new token and add that token as a new member of the Chef-managed `Administrator` policy. This is the equivalent of the v1 command `chef-automate admin-token`.
+Generate a new token and add that token as a new member of the Chef-managed `Administrator` policy. This is the equivalent of the v1 command `chef-automate admin-token`.
```
blakestier: This sentence (however old), makes me think I need to generate the token with the command _and then_ add it to the policy (even though that step/functionality is integrated into the command itself)

blakestier: It also feels like it shouldn't be under "Restore" access . . .

components/automate-chef-io/content/docs/iam-v2-api-reference.md
msorens: I had confirmed it; it is true.

components/automate-chef-io/content/docs/iam-v2-api-reference.md
msorens: My earlier question was meant to ask "should we delete it for all tables?" I have updated it to include "for all tables" now. Thx.

components/automate-chef-io/content/docs/iam-v2-api-reference.md
msorens: No, the GUID is used to reference members for input and for output, so it is consistent by itself, just not consistent with policy members.

components/automate-chef-io/content/docs/iam-v2-api-reference.md
msorens: Yep, I see you found it had moved. The diff is a poor helper in this PR.

components/automate-chef-io/content/docs/iam-v2-api-reference.md
blakestier: I just tested locally and I'm still seeing the v1 token work fine (the reverse is not true-- a v2 token will not work when you reset), but all v1 admin tokens are added to the admin team (or policy) on upgrade.

Mind double checking on your end?

msorens: Hmm, OK, let's go with "they work" then. I will go back and check my machine when I have a chance.

components/automate-chef-io/content/docs/iam-v2-api-reference.md
msorens: Seems redundant to me, that's why I did not do that.

components/automate-chef-io/content/docs/iam-v2-api-reference.md
msorens: Right on!

components/automate-chef-io/content/docs/iam-v2-api-reference.md
msorens: Well, depends. Remember that I put in these examples to illustrate everything that could go in the JSON; that's why both need to show up.

components/automate-chef-io/content/docs/iam-v2-api-reference.md
msorens: Can't say I've been completely consistent, but my intent was to use `Custom` when referring to the property value and `custom` when it is just being an adjective.

components/automate-chef-io/content/docs/iam-v2-api-reference.md
msorens: Again the diff is getting in the way: that is retained under `### Updating All Members on a Policy`

components/automate-chef-io/content/docs/iam-v2-api-reference.md
msorens: (Commented on the first one so marking resolved.)

components/automate-chef-io/content/docs/iam-v2-api-reference.md
msorens: Nice wording! Applied to all sections.

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +159,556 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
+}
 ```
 
+### Listing Team Users
+
+The output lists all local users on the specified team.
+
+### Adding Users to a Team
+
+Specify just new users to add to a team's membership via this HTTP request.
```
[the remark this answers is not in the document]
msorens: Agreed! So here's the thing. Each of these `###` sections is here to provide notes on the command. Even if there is nothing notable. So L615 for example provides a wee bit of info; the fact that it only applies to local users. But for L619, there is nothing notable, so this was my attempt to say nothing. Please suggest some better wording. 😁

components/automate-chef-io/content/docs/iam-v2-api-reference.md
blakestier: Yes, but by showing that it all can go in, we probably encourage them to use both, which IMO adds complexity for them-- even though we want the option for more savvy users who have a complex system. I think a lot of folks just look at the examples and glaze over the text  . . . prob a question for @susanev

components/automate-chef-io/content/docs/iam-v2-api-reference.md
blakestier: I agree-- perhaps that's because we don't really need it as part of `list`?

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -10,60 +10,143 @@ toc = true
...
+Teams, tokens, roles, and policies all contain a top-level `projects` property; you will see that in the example JSON for each resource in the following sections.
+This property associates the particular IAM resource with one or more projects.
+Users must have permissions for those projects to be able to view or modify those resources.
+If the `projects` property is left empty, that resource is categorized as *unassigned*.
+
+To create those permissions for users, there is a separate and distinct `projects` property on each policy statement.
+This `projects` property may **not** be empty; you must either specify specific projects or use a wildcard (`*`) to indicate all projects.
 
-Create a policy by composing JSON with all the necessary policy data.
+Here is a summary:
```
susanev: i think you could remove this line...

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -10,60 +10,143 @@ toc = true
...
 
-This example policy is made up of two statements, each enforcing a different set of permissions.
-The first policy statement grants users list view access.
-The second statement grants editor access via the `editor` role.
-You can see the actions contained in the `editor` role in the [Roles section]({{< relref "iam-v2-api-reference.md#roles" >}}).
+Type               | May be empty ? | Allows "*" ? | May specify (unassigned) ?
```
susanev: could we have the data in the columns be specific instead of `yes`, `no` so its consumable without the headers? so `allowed to be empty`, `not allowed to be empty`... and `allows wildcards`, `does not allow wildcards` or something like that...

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -10,60 +10,143 @@ toc = true
...
+update | PUT /policies/***id***    | updates a *Custom* policy
+delete | DELETE /policies/***id*** | deletes a *Custom* policy
+
+**Example Policy:**
+
+This policy shows two statements, each enforcing a different set of permissions.
+The first policy statement grants "list" access to view users.
+The second statement grants access to all the actions comprising the `editor` role.
+Note that you may specify actions either inline with **actions** or using a **role** reference.
+(Within a single statement it is even OK to use both if you like; the set of actions is the union of the inline actions and the actions defined by the role.)
```
susanev: this is kind of a lot to parse when reading, any way to break it up or make it more simple?

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +157,548 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
-Assuming you stored the above JSON in the file "policy.json", pass it to `curl` to create the policy:
+### Listing Policies
 
-```bash
-curl -s -H "api-token: $TOKEN" -d @policy.json https://{{< example_fqdn "automate" >}}/apis/iam/v2beta/policies?pretty
-```
+The output includes both *Chef-managed* and *Custom* policies.
+Policies migrated from IAM v1 are included as *Custom* policies.
+Policy names that begin with `[Legacy]` are default policies migrated from v1.
+We recommend deleting these legacy policies and setting up new IAM v2 policies in their place.
```
susanev: could we add something that says we recommend deleting them after moving associated members to the new chef-managed policies?

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +157,548 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
-In order to update a policy, you must know its ID.
-Use the command from [Listing Policies]({{< relref "iam-v2-api-reference.md#listing-policies" >}}) to locate your policy ID.
+*In the browser:* **Settings**  >> **Policies** >> [select a policy]
 
-The simplest way to pass a policy is by storing it in a file; we used a sample `policy.json` in the previous section.
-To update that policy you still need to pass in the complete policy definition--plus its members--just as you did when creating it; any omitted properties will be erased from the policy!
+### Creating a Policy
 
-As an example, to update the example policy to keep the first policy statement allowing users list-access, while removing `editor` access, edit your `policy.json` file:
+Create a policy by composing JSON with all the necessary policy properties (see the beginning of the section above).
```
susanev: can we do an anchor link?

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +157,548 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
+
+Method | HTTP request              | Description
+-------|---------------------------|------------
+list   | GET /roles                | lists both *Chef-managed* and *Custom* roles
+get    | GET /roles/***id***       | gets the specified role
+create | POST /roles               | creates a *Custom* role
+update | PUT /roles/***id***       | updates a *Custom* role
+delete | DELETE /roles/***id***    | deletes a *Custom* role
+
+**Default Chef-managed Roles:**
```
susanev: why arent these headings?

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +157,548 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
+
+### Getting a Project Rule
+
+This shows the details of a single project, selected by its ID.
+
+*In the browser:* **Settings**  >> **Projects** >> [select a project] >> [select a rule]
+
+### Creating a Project Rule
+
+Create a rule by composing JSON with all the necessary rule properties (see the beginning of the section above).
```
bcmdarroch: link?

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +157,548 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
+The choice of attributes depends on the rule **type**.
+
+Rule Type  | Available Attributes
+-----------|---------------------
+EVENT      | CHEF_ORGANIZATION, CHEF_SERVER
+NODE       | CHEF_ORGANIZATION, CHEF_SERVER, ENVIRONMENT, CHEF_ROLE, CHEF_TAG, CHEF_POLICY_NAME, CHEF_POLICY_GROUP
+
+Each rule condition also specifies an **operator**, either `MEMBER_OF` or `EQUALS`.
+To match a single value, use `EQUALS` and include just a single value in the **values** array.
+To match any one of a set of values, use `MEMBER_OF` and include the choices in the **values** array.
```
bcmdarroch: This might be a good spot for this Important Detail: 

> Emphasize that all rule conditions must be true for the rule to apply. (for example: the conditions chef_org EQUALS project1 and chef_org EQUALS project2 contradict each other. the operator MEMBER_OF should be used instead)

components/automate-chef-io/content/docs/iam-v2-api-reference.md
bcmdarroch: To make this consistent we'd probably need a table in the teams db that correlated v2 user IDs and v1 user GUIDs

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -74,227 +157,548 @@ You can see the actions contained in the `editor` role in the [Roles section]({{
...
+```json
+{
+  "id": "test-rule",
+  "name": "my test rule",
+  "type": "NODE",
+  "project_id": "eastern_region",
+  "conditions": [
+    {
+      "operator": "MEMBER_OF",
+      "attribute": "CHEF_SERVERS",
```
bcmdarroch:
```suggestion
      "attribute": "CHEF_SERVER",
```

components/automate-chef-io/content/docs/iam-v2-api-reference.md
susanev: for the most part id like to  center on using _roles_ in policies, maybe with the only mention of being able to add a list of actions in the api reference?

components/automate-chef-io/content/docs/iam-v2-api-reference.md
susanev: yes please!!

components/automate-chef-io/content/docs/iam-v2-api-reference.md
msorens: That was a deliberate choice. To me, it confuses the hierarchy. Right now, each item that is a heading (and therefore each item in the table of contents) is an endpoint. I thought this would clutter unnecessarily.

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -10,60 +10,143 @@ toc = true
...
-In the browser: Navigate to the *Policies* page under the **Settings**  tab, and then select an individual policy to open its details.
+## Notes on the Projects Property
 
-### Creating a Policy
+Teams, tokens, roles, and policies all contain a top-level `projects` property; you will see that in the example JSON for each resource in the following sections.
+This property associates the particular IAM resource with one or more projects.
+Users must have permissions for those projects to be able to view or modify those resources.
+If the `projects` property is left empty, that resource is categorized as *unassigned*.
+
+To create those permissions for users, there is a separate and distinct `projects` property on each policy statement.
```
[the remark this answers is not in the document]
msorens: 🤔 But that is precisely what this section is already doing. Could you elaborate on what could be added?

components/automate-chef-io/content/docs/iam-v2-api-reference.md
msorens: Not sure if you are advocating for a change here or not @susanev , as this is in the api reference.

susanev: yah i couldnt figure out the right away to say it, i havent had a time to look at everything all together, but as long as its only mentioned here thats cool with me

components/automate-chef-io/content/docs/iam-v2-api-reference.md
msorens: I think it is useful to at least mention once. 🤷‍♂

PR #1814, opened 2019-10-05
Jamie/reporting profile status visuals

SEAjamieD:
When viewing compliance/reports/profiles, and then clicking to view a specific profile ==> then clicking on scan results of each control, a sidebar opens to show each specific node and its status.  After the node viewed, all nodes would show a `failure` icon, even when the node was successful.

It turned out that the way the two filters were being concatenated caused the `profileFilter` to add extras profile IDs to each call, stacking in multiples as it went.  

### :+1: Definition of Done
Status sidebar shows the correct status icon when being viewed

2. Navigate to https://a2-dev.test/compliance/reports/profiles
3. Click on a Profile to view
4. Click on `Scan Results` on the right side of any of the list items
5. View the correct Icon (blue check, pink triangle, grey circle) for related node.

- [x] Tests added/updated?
- [ ] Docs added/updated?

Before:  Icon does not match true status
Before
Before - Sidebar

After:  Icon matches status
After
After - Sidebar

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-profile/reporting-profile.component.ts
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-profile/reporting-profile.component.ts
@@ -124,7 +124,7 @@ export class ReportingProfileComponent implements OnInit, OnDestroy {
   hideScanResults() {
     this.scanResults.opened = false;
     this.scanResults.clearParams();
-    this.scanResults.control = null;
+    this.scanResults.control = {};
     this.openControls = {};
   }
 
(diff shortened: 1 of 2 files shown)


components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-profile/reporting-profile.component.ts
```diff
@@ -129,6 +129,10 @@ export class ReportingProfileComponent implements OnInit, OnDestroy {
   }
 
   getNodes(reportQuery: ReportQuery, params: any): Observable<Array<any>> {
+    // remove profile_id and control_id as filters if they are already applied
```
SEAjamieD: I ended up opting to remove the profile_id and control_id filters just before loading in the new ones in the case that other filters were being applied, it wouldn't remove ALL filters, if we opted to add more.

SEAjamieD: Initially I was clearing out ALL filters at the end of the function, which I still think may be the cleaner way to do it.

PR #1818, opened 2019-10-08
UI and database for data-feed-service

satellite15:

New settings UI page ( feature flagged ) for the data-feed-service. This is backed by a new DB for the data-feed-service which holds destination configurations for the data-feed. 

- [x] Tests added/updated?
- [ ] Docs added/updated?

No docs yet. Doc to be finalised once the feature is complete and ready to remove the feature flag.

Screenshot 2019-10-08 at 16 36 27
Screenshot 2019-10-08 at 16 38 02
Screenshot 2019-10-08 at 16 39 07


jonong1972:
The UX team will review. we are meeting tomorrow and can discuss this. There may be parallel  design patterns we are working on that may help out with this.

api/external/data_feed/data_feed.proto
```diff
@@ -0,0 +1,136 @@
...
+		option (chef.automate.api.policy) = {
+			resource: "datafeed:destination/{id}"
+			action: "update"
+		};
+		option (chef.automate.api.iam.policy) = {
+			resource: "datafeed:destination:{id}"
+			action: "datafeed:destination:update"
+		};
+	};
+    rpc ListDestinations(Empty) returns (ListDestinationResponse){
```
lancewf: Instead of `Empty` for the request parameter, you should create a new parameter for each RPC. This allows for easy additions and changes to the RPC. For example, create a `ListDestinationsRequest` that is empty like below. 
```
message ListDestinationsRequest {}
```

Then later if the RPC needs a filter or any other argument it can be easily added to the `ListDestinationsRequest` object without breaking the API.

api/external/data_feed/data_feed.proto
```diff
@@ -0,0 +1,136 @@
...
+		option (chef.automate.api.policy) = {
+			resource: "datafeed:destination:{id}"
+			action: "read"
+		};
+		option (chef.automate.api.iam.policy) = {
+			resource: "datafeed:destination:{id}"
+			action: "datafeed:destination:get"
+		};
+	};
+    rpc DeleteDestination(DestinationId) returns (DestinationResponse){
```
lancewf: The RPC should not share the same request parameter `DestinationId`. Create a new one for each RPC. This is because changing the request parameter for an RPC will break existing users, but adding a new filed to a request parameter does not effect existing users. If later the "GetDestination" needs a filters field it can be easily added.

components/automate-ui/src/app/pages/data-feed-form/data-feed-form.component.ts
```diff
@@ -0,0 +1,152 @@
...
+        .subscribe( destination => {
+          this.showLoading = false;
+          return this.model =
+            new Model(new Destination(destination.id, destination.name, destination.targetUrl,
+              destination.targetSecretId), '', '');
+        });
+    }
+  }
+
+  ngOnInit() {
```
lancewf: This should be removed along with the `implements OnInit` on line 37.

components/data-feed-service/server/server.go
```diff
@@ -0,0 +1,182 @@
...
+	httpRequest.Header.Add("Accept", "application/json")
+	client := http.Client{}
+	httpResponse, err := client.Do(httpRequest)
+	if err != nil {
+		log.Errorf("Error sending test message %v", err)
+		return response, err
+	} else {
+		log.Infof("Test data posted to %v, Status %v", url, httpResponse.Status)
+	}
+	if httpResponse.StatusCode != http.StatusCreated {
```
lancewf: Why is only a 201 the correct response? When testing a new data feed from the UI I am seeing "We were unable to connect to your destination" because the test server is returning a 200.

components/data-feed-service/service/data-feed-service.go
```diff
@@ -358,6 +326,10 @@ func (client DataClient) sendNotification(notification datafeedNotification) err
 	if response.StatusCode != http.StatusCreated {
```
lancewf: Could this success if the response is 200 also?

components/automate-ui/src/app/pages/data-feed/data-feed.component.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  private sendCountToTelemetry(destinations: Destination[]) {
+    this.telemetryService.track('datafeedDestinationsCount', destinations.length);
+  }
+
+
+  // TODO - this was common in all three uses, but I'm not sure this is the best
+  // way to do it - do we really need to refresh after the server confirms the action
+  // successful? Seem it should be possible to update the local model with the
+  // changes and have that trigger view updates?
```
lancewf: I would get rid of this comment. It just shows this code was copied from the NotificationsComponent 😄.

components/automate-ui/src/app/pages/data-feed/data-feed.component.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  constructor(
+    private service: DatafeedService,
+    public dialog: MatDialog,
+    public snackBar: MatSnackBar,
+    private telemetryService: TelemetryService
+  ) { }
+
+  ngOnInit() {
+    this.destinations$ = this.service.fetchDestinations();
```
lancewf: This will have to be updated to use NGRX, but I would leave that for a later PR.

components/data-feed-service/server/server.go
satellite15: The service now code returns 201 as the import is asynchronous and we have only created the import records, these need processed before the CMDB tables are populated. Since this isn't restricted to service now that's a very valid point. I'll change to a 200 and return 200 from service now in the event of a test message.

PR #1831, opened 2019-10-09
deploy depot in automate

jaym:
This brings up builder and you can see a thing that looks like builder if you navigate to `https://a2-dev.test/bldr`

Needs builder-api from https://github.com/habitat-sh/builder/pull/1182

TODO:
- [x] builder-api needs to bind to minio
- [x] builder-api needs to be able to do its auth things with the automate auth stack
- [x] builder wants to live at `/`, we need it to live at `/bldr`. This means some of the assets are broken
- [x] we need to wire up config for memcached
- [x] we need to wire up config for builder-api
- [x] we need to wire up config for builder-api-proxy

Some useful config:
```
[deployment.v1.svc]
products = ["automate","depot"]
#products = ["automate"]
```


components/automate-builder-api-proxy/habitat/config/mime.types
```diff
@@ -0,0 +1,89 @@
+
+types {
```
stevendanna: I think that this file also lives in the nginx package that we ship, so unless we are doing something special, we could do something like:

```
components/automate-load-balancer/habitat/config/nginx.conf
12:  include        {{ pkgPathFor "core/nginx" }}/config/mime.types;
```

and exclude this file.

components/automate-builder-api-proxy/habitat/config/nginx.conf
```diff
@@ -0,0 +1,143 @@
...
+    #ssl_dhparam {{pkg.svc_data_path}}/dhparams.pem;
+    #ssl_protocols {{cfg.ngx.http.ssl_protocols}};
+    #ssl_ciphers {{cfg.ngx.http.ssl_ciphers}};
+    #ssl_session_timeout 5m;
+    #ssl_prefer_server_ciphers on;
+
+
+    proxy_send_timeout 300;
+    proxy_read_timeout 300;
+    proxy_connect_timeout 300;
```
stevendanna: 5m connect timeout seems high.

components/automate-deployment/pkg/server/server.go
```diff
@@ -423,9 +423,8 @@ func (s *server) configRenderer() (ConfigRenderer, error) {
 
 		preparableCfg, found := s.deployment.Config.PlatformServiceConfigForService(service.Name())
 		if !found {
-			err := errors.Errorf("unable to render configuration for unknown service %q", service.Name())
-			logrus.WithField("service", service.Name).WithError(err).Error("unable to render configuration")
-			return "", err
+			logrus.WithField("service", service.Name).WithError(err).Warnf("unable to render configuration for unknown service %q", service.Name())
```
stevendanna: Why are we OK with this just being a warning?

While we are touching this, `service.Name` here should probably be `service.Name()` (or we could just remove it since we also include the service name in the log message.

jaym: it was annoying having to define all the protobuf config stuff while i was wiring stuff up. i just wanted it to use the default.toml and start the hab package. I can change it back if it is too scary

.bldr.toml
```diff
@@ -1910,6 +1922,30 @@ paths = [
   "vendor/google.golang.org/grpc/*",
   "vendor/gopkg.in/yaml.v2/*"
 ]
+
+[chef/automate-minio]
```
yzl: `automate-builder-minio`? Or are we using the same minio for everything?

jaym: not sure yet. its one of the areas that needs to be thought out

api/config/builder_api_proxy/config_request.go
```diff
@@ -0,0 +1,65 @@
+package bldrapiproxy
+
+import (
+	config "github.com/chef/automate/api/config/shared"
+	w "github.com/chef/automate/api/config/shared/wrappers"
+)
+
+// NewConfigRequest returns a new ConfigRequests instance with zero values.
```
yzl: "new ConfigRequests" -> "new ConfigRequest"

components/automate-builder-api-proxy/habitat/config/nginx.conf
```diff
@@ -0,0 +1,143 @@
...
+  server {
+    index /index.html;
+    server_name  {{sys.hostname}};
+    root         {{pkgPathFor "habitat/builder-api-proxy"}}/app;
+
+    listen {{cfg.service.host}}:{{cfg.service.port}};
+
+    access_log /dev/stdout chef;
+
+    #ssl on;
```
yzl: Are these here for an eventual TODO?

jaym: yea, unclear if we should bother with ssl. i don't think builder-api supports it

components/automate-builder-api-proxy/habitat/plan.sh
```diff
@@ -0,0 +1,55 @@
...
+pkg_version="0.1.0"
+pkg_maintainer="Chef Software Inc. <support@chef.io>"
+pkg_license=("Chef-MLSA")
+pkg_upstream_url="https://www.chef.io/automate"
+pkg_svc_user="root"
+pkg_deps=(
+  core/runit
+  chef/mlsa
+  "${local_platform_tools_origin:-chef}/automate-platform-tools"
+  # We need to pin here to get a build from unstable
```
yzl: Is this just to get the build that has your fixes in it?

components/automate-minio/habitat/hooks/run
```diff
@@ -0,0 +1,50 @@
...
+
+MINIO_ACCESS_KEY="$(secrets-helper show minio.access_key)"
+export MINIO_ACCESS_KEY
+MINIO_SECRET_KEY="$(secrets-helper show minio.secret_key)"
+export MINIO_SECRET_KEY
+export MINIO_BROWSER=off
+export SSL_CERT_FILE={{pkgPathFor "core/cacerts"}}/ssl/cert.pem
+export SSL_CERT_DIR={{pkgPathFor "core/cacerts"}}/ssl/certs
+
+#{{#if cfg.gateway.storage.s3.bucket.name~}}
```
yzl: Are the commented-out bits going to be used eventually?

jaym: ya. we need to figure out how external s3 is going to work. i think it relates to you automate-minio vs automate-builder-minio question as well

components/automate-builder-api-proxy/habitat/plan.sh
jaym: we'll want to auto-bump this

PR #1836, opened 2019-10-10
[datalifecycle] Fix purging for old indices

jaym:
- fixes an issue where document based purging wasn't correctly
happening
- purges old versions of time series indices


jaym:
this is wrong, but it does fix one bug

jaym:
replacing with https://github.com/chef/automate/pull/1853 as it fixes only the bug

PR #1837, opened 2019-10-10
[go] migrate to go modules

ryancragun:
# Migrate from dep to Go modules

Go introduced module mode in v1.11 and it became the default in v1.13.
This change replaces `dep` with `go mod` using a vendor directory.

## What's changed?

* You no longer need to have the Chef Automate git repository inside of a `GOPATH`.
  Instead, all `go` utilities will recognize that they are in module mode and will
  handle dependency loading for you. You're now free to keep the Chef Automate
  source code in any directory you wish. You can even unset `GOPATH`. :tada:

* The `go` command supports using a module vendor directory but it has to be
  explicitly enabled with the `-mod=vendor` go flag. If you prefer not specifying
  it each time you can set the `GOFLAGS=-mod=vendor` environment variable. We
  have done this in the studio so all go commands should utilize the vendored
  dependencies, but if you run go commands outside of the studio you'll need to
  specify the flag or set the environment variable to utilize the vendored
  dependencies. If you forget, commands should still execute but they will download
  the modules into your go installations modules directory instead of using the
  vendored packages.

* Inside the studio the `GOBIN` has been set to `/src/bin`, which means `/bin`
  relative to the Chef Automate source code directory outside of the studio. This
  has the effect that installed binaries that are compiled in the studio are
  also available outside of the studio. This provides the benefit that go
  development tools that have compiled inside the studio don't have to be
  re-compiled each time you start a new studio container. It also means that
  protobuf compilers and extensions are also cached, therefore you might have to
  manually remove them if changes are made. If you notice strange behavior when
  compiling protobufs, remove the compiled binaries in `/bin` and try again.

* `enter_go_workspace`, `setup_go_workspace`, and other various studio functions
  that are required when working with go in `GOPATH` mode are no longer required.
  I tried to update every studio function but it's likely that something slipped
  though the cracks. If you notice that your studio function is behaving incorrectly
  please accept my apologies now.

* Some, but not all, of the protobuf compilers and extensions that we use in
  Chef Automate are not compatible in module mode. All of the `grpc.sh` protobuf
  scripts have been updated to utilize relative source path outputs where supported,
  otherwise measures have been put in place to compile them and move them into
  the correct directory afterwards.

* Vendoring is no longer done with `dep`. Adding a new dependency to Chef Automate
  is now as easy as importing the package into your go application and running `make
  revendor` from the root of the Chef Automate source code repo or in the studio. If you
  require go packages for development that are not included in the Chef Automate
  repository,  add them to the `vendor.go` as ignored dependencies and revendor.

* Many packages were updated as part of this migration. Everything has been
  recompiled with the new vendor directory and the single known regression has
  been fixed. If a dep change has caused you problems we can work to pin it.

* We previously relied on 3 different layers of scaffolding: `core/scaffolding-go`,
  `chef/scaffolding-go` and `chef/automate-scaffolding-go`. That's two layers of
  scaffolding too many. Because we're now in module mode our scaffolding that
  handled much of the `GOPATH` wonkiness is no longer required, thus all packages
  that require go have been migrated to `chef/automate-scaffolding-go`, which has
  been updated for the go mod with vendor directory workflow. If you add  a new go
  component  you should use `chef/automate-scaffolding-go`. If you need to
  opt out of the platform scaffolding for a go package you can set the new
  `scaffolding_no_platform` variable to true in the plan.

* Default version linker flags are now set in the scaffolding. You don't need to
  manually set them in your plans.

* Platform TLS CA and certs are handled by the scaffolding, you don't need to
  specify them in your plans.

## Help, something broke

Sorry, we did our best to make this transition as seamless as possible. Reach
out to #a2-deployment and we'll do our best to help out.

## Changelog

[go] replace dep with go mod vendoring
[automate-scaffolding-go] add go mod support to automate-scaffolding-go
[automate-scaffolding-go] remove dependency on core and chef go scaffolding
[hab-plans] migrate all plans to automate-scaffolding-go
[hab-plans] remove unused platform TLS config
[hab-plans] use scaffolding version linker flags everywhere
[gprc] update all grpc.sh scripts to use relative source paths where possible
[studio] remove GOPATH requirements from all studio functions
[lib/cereal] fix a regression caused by `lib/pq` library breaking changes
[bldr] refactor bldr generator tool for module mode
[repo-health] update repo health checks for module mode


.bldr.toml
```diff
@@ -35,36 +35,7 @@ paths = [
...
-  "vendor/github.com/uber/jaeger-lib/*",
-  "vendor/go.uber.org/atomic/*",
-  "vendor/go.uber.org/multierr/*",
-  "vendor/golang.org/x/net/*",
-  "vendor/golang.org/x/sys/*",
-  "vendor/golang.org/x/text/*",
-  "vendor/google.golang.org/genproto/*",
-  "vendor/google.golang.org/grpc/*",
-  "vendor/gopkg.in/yaml.v2/*"
+  "lib/version/*"
```
ryancragun: The bldr generator is not including vendor. Gotta look into it

vendor/github.com/burntsushi/toml/.travis.yml
```diff
@@ -0,0 +1,15 @@
+language: go
```
stevendanna: It looks to me like `go mod` doesn't do pruning of `.travis`, `.gitignore`, `README` and other non-go files.  Is there any value in adding a post-processing step?  I'm thinking it might be best to just live with it.

.studio/common
```diff
@@ -265,8 +266,7 @@ DOC
 function generate_dev_root_ca() {
     install_if_missing core/certstrap certstrap
     (
```
stevendanna: Since we aren't changing directories anymore. We can also remove this subshell.

vendor/github.com/burntsushi/toml/.travis.yml
ryancragun: I went ahead and added a little hacky delete to the revendor make target. It doesn't catch _everything_ that `dep` did, but it's a start and we can add or remove things from it as necessary going forward.

.studio/common
```diff
@@ -33,16 +33,17 @@ function compile_go_protobuf_component() {
...
-    install_go_tool "${proto_go_tools[@]}"
-    GO_TOOL_VENDOR=true install_go_tool github.com/envoyproxy/protoc-gen-validate
+  local proto_go_tools=(
+    github.com/chef/automate/components/automate-grpc/protoc-gen-policy
+    github.com/chef/automate/components/automate-grpc/protoc-gen-grpc-mock # generate mock grpc server implementations
+    github.com/chef/automate/components/automate-grpc/protoc-gen-a2-config
   )
+  install_go_tool "${proto_go_tools[@]}"
+  install_go_tool github.com/envoyproxy/protoc-gen-validate
+  install_go_tool github.com/ckaznocha/protoc-gen-lint
```
yzl: [nit] Can `protoc-gen-validate` and `protoc-gen-lint` be added to `proto_go_tools`?

api/scripts/grpc.sh
```diff
@@ -1,16 +1,37 @@
 #!/bin/bash
 # set -x
-GOPATH=$(go env GOPATH)
 
-echo "Generate Go, Swagger, and GRPC Gateway"
+echo "Generate Go, Swagger, Validation and GRPC Gateway"
```
yzl: [nit] "Validation, and GRPC Gateway"?

components/automate-gateway/scripts/grpc.sh
```diff
@@ -1,16 +1,38 @@
...
 echo "Generate Go, Swagger, and GRPC Gateway"
 
-IMPORTS=(-I.
-         -I"$GOPATH/src"
-         -Ivendor
-         -Ivendor/github.com/grpc-ecosystem/grpc-gateway
-         -Ivendor/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis
-         -Ivendor/github.com/envoyproxy/protoc-gen-validate
-        )
+# Unlike gen-go, the gateway,and swagger protoc extension don't support the
```
yzl: [nit] "the gateway,and" -> "the gateway and"

components/automate-cs-nginx/habitat/plan.sh
```diff
@@ -47,7 +46,7 @@ pkg_upstream_url="https://www.chef.io/automate"
 # the other nginx instances in A2 at the moment:
 pkg_svc_user="root"
 
-pkg_scaffolding=chef/scaffolding-go
+pkg_scaffolding="${local_scaffolding_origin:-chef}/automate-scaffolding-go"
```
jaym: its unclear why this works. yzl added some stuff to the packages.meta file that you have to set to true to say you have platform scaffolding

jaym: it seems the meaning of not editing package.meta is that the service wont get external postgres info

vendor.go
```diff
@@ -0,0 +1,20 @@
...
+	_ "github.com/ckaznocha/protoc-gen-lint"
+	_ "github.com/envoyproxy/protoc-gen-validate"
+	_ "github.com/go-delve/delve/cmd/dlv"
+	_ "github.com/golang/mock/gomock"
+	_ "github.com/golang/mock/mockgen"
+	_ "github.com/golang/protobuf/protoc-gen-go"
+	_ "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway"
+	_ "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger"
+	_ "golang.org/x/perf/cmd/benchstat"
+)
```
stevendanna: We might want to add go-bindata to this

PR #1850, opened 2019-10-11
Table Overlap w/ Sidebar Bug in Docs

Shadae:
Signed-off-by: Shadae <sholmes@chef.io>

On the default policies page in the docs the left-nav sometimes overlaps the content. Specifically, if you scroll down to a wide table and then scroll left you will notice that the left-nav overlaps the content.

Added a shortcode which is just a simple div, that will wrap the table (as long as you add {{% table %}}{{% /table %}} around the table markdown) and keep the table contained the same visual way other text and code snippets are contained within the rest of the documentation. 

Before: 
https://app.zenhub.com/workspaces/ui-team-5cba23a3b14fdc05970c8f87/issues/chef/automate/1625

After: 
Screen Shot 2019-10-10 at 6 41 42 PM
Screen Shot 2019-10-10 at 6 41 28 PM

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-chef-io/content/docs/default-policies.md
+++ b/components/automate-chef-io/content/docs/default-policies.md
@@ -43,6 +43,7 @@ the specific endpoints they protect.
 
 > Corresponds to "Client Runs" tab or `/nodes`
 
+{{% table %}}
 Service | Method | HTTP Endpoint | HTTP Method | Resource | Action
 ---|---|---|---|---|---
 ConfigMgmt | GetNodes | /cfgmgmt/nodes | GET | cfgmgmt:nodes | read
@@ -61,6 +62,7 @@ JobScheduler | GetStatusJobScheduler | /retention/nodes/status | GET | cfgmgmt:s
 JobScheduler | ConfigureNodesMissingScheduler | /retention/nodes/missing-nodes/config | POST | cfgmgmt:scheduler:missing-node | configure
 JobScheduler | ConfigureDeleteNodesScheduler | /retention/nodes/delete-nodes/config | POST | cfgmgmt:scheduler:delete-node | configure
 JobScheduler | ConfigureMissingNodesForDeletionScheduler | /retention/nodes/missing-nodes-deletion/config | POST | cfgmgmt:scheduler:missing-nodes-deletion | configure
+{{% /table %}}
 
 ## Compliance
 
@@ -83,6 +85,7 @@ JobScheduler | ConfigureMissingNodesForDeletionScheduler | /retention/nodes/miss
 > Corresponds to "Compliance" tab (`/compliance/reporting/overview`), "Scan Jobs" tab
 (`/compliance/scanner/jobs`) & "Profiles" tab (`/profiles`)
 
+{{% table %}}
 Service | Method | HTTP Endpoint | HTTP Method | Resource | Action
 ---|---|---|---|---|---
 ProfilesService | Read | /compliance/profiles/read/{owner}/{name}/version/{version} | GET | compliance:profiles:storage:{owner}:{name}:{version} | read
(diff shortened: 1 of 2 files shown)


components/automate-chef-io/content/docs/default-policies.md
```diff
@@ -61,6 +62,7 @@ JobScheduler | GetStatusJobScheduler | /retention/nodes/status | GET | cfgmgmt:s
 JobScheduler | ConfigureNodesMissingScheduler | /retention/nodes/missing-nodes/config | POST | cfgmgmt:scheduler:missing-node | configure
 JobScheduler | ConfigureDeleteNodesScheduler | /retention/nodes/delete-nodes/config | POST | cfgmgmt:scheduler:delete-node | configure
 JobScheduler | ConfigureMissingNodesForDeletionScheduler | /retention/nodes/missing-nodes-deletion/config | POST | cfgmgmt:scheduler:missing-nodes-deletion | configure
+{{% /table %}}
```
Shadae: {{% table %}} before and {{% /table %}} after the data content of the table will add the table shortcode. If you would rather this be named something such as `responsive-table` or `table-container` let me know.

components/automate-chef-io/layouts/shortcodes/table.html
```diff
@@ -0,0 +1,3 @@
+<div class="table" style="overflow-x:auto;">
+  {{ .Inner }}
+</div>
```
Shadae: @kagarmoe I left the style here so I can chat with you about where this would live.

components/automate-chef-io/content/docs/default-policies.md
kagarmoe: @IanMadd Look at this shortcode!
Thanks @Shadae! I think calling it responsive-table would be helpful--at least as a reminder for ourselves and info for future docs people why we used it.

components/automate-chef-io/layouts/shortcodes/table.html
IanMadd: Maybe: 

```
<div class="table"> 
  {{- .Inner | markdownify -}}
</div>
```

And then somewhere in the scss:
```
.table {
    overflow-x:auto; 
}
```
Once you put markdown text inside a div tag (or any other html tag I believe) blackfriday treats it like html so you have to explicitly tell it to render the text as markdown. Hence the `markdownify` command. With the css, I think it's better to have it one location instead of added to the html templates. My 2 cents.

Also, I'd add the shortcode (however it's named) to the Hugo theme so it can be used on other Hugo sites if it were ever needed.

Shadae: Yes. That is the question. I have no idea where it would go in the hugo theme @IanMadd - I wanted to defer to you two on naming and file organization since you deal with the docs. There isn't css in the themes main files, everything is imported so I'm just making sure I need to create a directory and a file or two to import into that file to keep that convention and make sure I'm not missing anything to y'alls knowledge. 

Also super Interesting it knows it's markdown without that "markdownify" tag added I had read that in the black friday docs as well, but examples on this work around (for this common hugo table problem) all had it without. Wasn't sure if it just works because it's a table?

IanMadd: That's odd. I just tried your PR without `markdownify` and the tables look like raw text. 

Screen Shot 2019-10-15 at 10 20 54 AM

Ok, it took a bit of sleuthing but I think I figured it out. I think we're running [different versions](https://discourse.gohugo.io/t/0-55-what-am-i-doing-wrong/18034) of Hugo. Check which version you have. There was an update to this behavior starting with 0.55. Before that you didn't have to add `markdownify`, afterwards you do. I added ```{{ $_hugo_config := `{ "version": 1 }` }}``` to your shortcode and it gave the behavior that you describe.

Yeah, I'd go with @kagarmoe's suggestion and call it `responsive-table`. If you want I could open a PR against the Chef Hugo theme and add a "responsive-table" shortcode and the css to sass/automate/_homepage.scss.

Shadae:
Haven't forgot about this. Going to make a change on the hugo doc template side first.

components/automate-chef-io/layouts/shortcodes/table.html
Shadae: Perfect!

Shadae: Super catch on the versions thing @IanMadd ! I will put up a PR on the hugo side. Thanks for pointing me in the right direction.

Issue #1451, opened 2019-09-05
Force upgrade from beta to GA 30-days after GA becomes available and decommission beta content & APIs


susanev:
## User Story

As an Automate User, when I upgrade to the latest release with the IAM v2 force upgrade, my system will be permanently upgraded to IAM v2. All my data should be safe and any v1 policies should be migrated. If I was already on IAM v2, all my v2 policies should be intact.

My v1 data will be preserved for a period of time (TODO how long?) after the upgrade. Then it will be cleaned up.

Aha! Link: https://chef.aha.io/epics/A2-E-185

susanev:
closing this, cause its become a release instead.

PR #1861, opened 2019-10-14
Always upgrade to v2.1 on chef-automate iam upgrade-to-v2

tylercloke:

Always turn on projects when upgrading to IAM V2.

Deprecates and always hides the --beta2.1 flag, which will simply be ignored.

```
```

```
chef-automate iam version
chef-automate reset-to-v1
```

- [x] Tests added/updated?
- [ ] Docs added/updated?


components/automate-cli/cmd/chef-automate/iam.go
```diff
@@ -100,15 +100,15 @@ func newIAMUpgradeToV2Cmd() *cobra.Command {
 		"skip-policy-migration",
```
tylercloke: Only non-testing related changes are in this file.

components/automate-cli/cmd/chef-automate/iam.go
```diff
@@ -100,15 +100,15 @@ func newIAMUpgradeToV2Cmd() *cobra.Command {
...
 		false,
 		"Do not migrate policies from IAM v1.")
+
+	// this flag is deprecated and does nothing but we don't wanna error on it
```
bcmdarroch: should we add this flag as a cleanup task in https://github.com/chef/automate/issues/1451?

inspec/a2-iam-v2-integration/controls/iam_v2.rb
```diff
@@ -1036,9 +1092,6 @@
       }
 
       before(:all) do
-        puts "wat"
```
bcmdarroch: 😂 the desperate cry of someone trying to debug these inspec tests..

integration/tests/cypress_e2e.sh
```diff
@@ -25,29 +25,14 @@ do_deploy() {
                 return 1
             fi
             ;;
-        v2)
+        v2.1)
```
bcmdarroch: you'll want to drop v2 from the [deploy pipeline script](https://github.com/chef/automate/blob/master/.expeditor/buildkite/cypress.sh#L9) too

inspec/a2-iam-v2-integration/controls/iam_v2.rb
tylercloke: I think that was me too :sob:

integration/tests/cypress_e2e.sh
tylercloke: Good catch!

components/automate-cli/cmd/chef-automate/iam.go
tylercloke: Will add after this merges so I can have an exact line number!

tylercloke: Actually, I think we are covered already, since we are planning on removing the [upgrade command completely](https://github.com/chef/automate/issues/1879).

components/automate-cli/cmd/chef-automate/iam.go
```diff
@@ -170,22 +170,13 @@ func display(v *policies_common.Version) string {
...
 	}
 
-	isBetaVersion := iamCmdFlags.betaVersion
-	if isBetaVersion {
-		upgradeReq.Flag = policies_common.Flag_VERSION_2_1
-		writer.Title("Enabling IAM v2.1")
-	} else {
-		writer.Title("Upgrading to IAM v2")
-	}
+	writer.Title("Enabling IAM v2.1")
```
msorens: Hmm... I think this needs to reflect the spirit of L187 not L185, so should change to "Upgrading"... not quite sure.

tylercloke: Unintentional. Good catch.

PR #1873, opened 2019-10-15
[cypress] fix cypress tests in deploy pipeline

bcmdarroch:

Mission: get cypress running on the deploy pipeline again. 

These tests stopped running after I added a dependency on the environment variable `CYPRESS_ADMIN_TOKEN` and mistakenly assumed I could use `chef-automate` in the deploy pipeline script to get that token. 😩 

Now I've added a before block that always runs before all of the Cypress tests. It checks to see if the admin token was provided in the env, and if not, generates one.

to run locally without passing an admin token: 
```
cd e2e
CYPRESS_IAM_VERSION=v2.1 CYPRESS_BASE_URL=https://a2-dev.test npm run cypress:open
```

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/e2e/cypress/support/commands.ts
+++ b/e2e/cypress/support/commands.ts
@@ -101,6 +101,8 @@ Cypress.Commands.add('generateAdminToken', (idToken: string) => {
       }
     });
   }
+
+  waitUntilAdminTokenPermissioned(100);
 });
 
 interface MemoryMap {
@@ -249,5 +251,26 @@ function LoginHelper(username: string) {
     cy.saveStorage();
 
     cy.wait(['@getAuthPopulateCache', '@getAuthParameterized']);
+  });
+}
+
+function waitUntilAdminTokenPermissioned(attempts: number): void {
+  if (attempts === -1) {
+    throw new Error('admin token failed to generate');
+  }
+  // admin-only API endpoint
+  cy.request({
+    headers: { 'api-token': Cypress.env('ADMIN_TOKEN') },
+    url: '/apis/iam/v2beta/projects',
+    method: 'GET',
+    failOnStatusCode: false
+  }).then((response) => {
+    if (response.status === 200) {
+      return;
+    } else {
+      cy.log(`${attempts} attempts remaining: waiting for admin token to be permissioned`);
+      cy.wait(1000);
+      waitUntilAdminTokenPermissioned(--attempts);
+    }
   });
 }
--- a/e2e/cypress/support/commands.ts
+++ b/e2e/cypress/support/commands.ts
@@ -66,10 +66,10 @@ Cypress.Commands.add('generateAdminToken', (idToken: string) => {
     url: `/apis/iam/v2beta/tokens/${adminTokenObj.id}`,
     failOnStatusCode: false
   }).then((resp: Cypress.ObjectLike) => {
-    expect([200, 404]).to.include(resp.status);
+    expect(resp.status).to.be.oneOf([200, 404]);
   });
 
-  // create token
+  // create token (OK to run v2 endpoints even if system is v1)
   cy.request({
     auth: { bearer: idToken },
     method: 'POST',
@@ -84,7 +84,7 @@ Cypress.Commands.add('generateAdminToken', (idToken: string) => {
     cy.request({
       auth: { bearer: idToken },
       method: 'POST',
-      url: '/apis/iam/v2beta/policies',
+      url: '/api/v0/auth/policies',
       body: {
         action: '*',
         resource: '*',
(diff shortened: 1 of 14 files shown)


e2e/cypress/support/index.ts
```diff
@@ -14,7 +14,16 @@
...
 // require('./commands')
+
+before(function () {
+  if (Cypress.env('ADMIN_TOKEN') === undefined || Cypress.env('ADMIN_TOKEN') === '') {
+    cy.adminLogin('/').then(() => {
+      const admin = JSON.parse(<string>localStorage.getItem('chef-automate-user'));
+      cy.generateAdminToken(admin.id_token);
+    });
+  }
+});
```
tylercloke: Oh interesting! Can you throw a comment about how this works in our cypress readme? I could see this feeling a bit magical if you weren't aware it existed. I like it though.

e2e/cypress/support/commands.ts
```diff
@@ -53,6 +53,58 @@ Cypress.Commands.add('applyProjectsFilter', (projectsToFilterOn: string[]) => {
...
+  };
+
+  // delete old token if exists
+  cy.request({
+    auth: { bearer: idToken },
+    method: 'DELETE',
+    url: `/apis/iam/v2beta/tokens/${adminTokenObj.id}`,
+    failOnStatusCode: false
+  }).then((resp: Cypress.ObjectLike) => {
+    expect([200, 404]).to.include(resp.status);
```
msorens: Reading cypress tests recently I noticed we have been doing this wrong.
It should really be "expect _actual value_ to satisfy _expected value_", which means:
```suggestion
    expect(resp.status).to.be.oneOf([200, 404]);
```
Not suggesting you fix all of the occurrences in this PR, but better to start using the correct version. 😁

e2e/cypress/support/commands.ts
```diff
@@ -53,6 +53,58 @@ Cypress.Commands.add('applyProjectsFilter', (projectsToFilterOn: string[]) => {
...
+  }).then((resp: Cypress.ObjectLike) => {
+    Cypress.env('ADMIN_TOKEN', resp.body.token.value);
+  });
+
+  // grant permissions based on IAM version
+  if (Cypress.env('IAM_VERSION') === 'v1') {
+    cy.request({
+      auth: { bearer: idToken },
+      method: 'POST',
+      url: '/apis/iam/v2beta/policies',
```
msorens: Need to use a v1 URL here.

e2e/cypress/support/commands.ts
```diff
@@ -201,3 +253,24 @@ function LoginHelper(username: string) {
...
+    url: '/apis/iam/v2beta/projects',
+    method: 'GET',
+    failOnStatusCode: false
+  }).then((response) => {
+    if (response.status === 200) {
+      return;
+    } else {
+      cy.log(`${attempts} attempts remaining: waiting for admin token to be permissioned`);
+      cy.wait(1000);
+      waitUntilAdminTokenPermissioned(--attempts);
```
msorens: You are just doing a simple loop here, so no need for recursive calls (which take a lot more overhead).

e2e/cypress/support/index.ts
```diff
@@ -14,7 +14,16 @@
...
 
 // Import commands.js using ES2015 syntax:
-import './commands'
+import './commands';
 
 // Alternatively you can use CommonJS syntax:
 // require('./commands')
+
+before(function () {
+  if (Cypress.env('ADMIN_TOKEN') === undefined || Cypress.env('ADMIN_TOKEN') === '') {
```
msorens: Consider simplifying to
```suggestion
  if (!Cypress.env('ADMIN_TOKEN')) {
```

e2e/cypress/support/commands.ts
```diff
@@ -53,6 +53,58 @@ Cypress.Commands.add('applyProjectsFilter', (projectsToFilterOn: string[]) => {
...
+  cy.request({
+    auth: { bearer: idToken },
+    method: 'DELETE',
+    url: `/apis/iam/v2beta/tokens/${adminTokenObj.id}`,
+    failOnStatusCode: false
+  }).then((resp: Cypress.ObjectLike) => {
+    expect([200, 404]).to.include(resp.status);
+  });
+
+  // create token
```
msorens: I had to think for a minute whether it is OK to use a v2 endpoint if this might be a v1 system... so perhaps add a comment to that effect:
```
// OK to run v2 endpoints even if system is v1
```

e2e/cypress/support/commands.ts
bcmdarroch: good point! 👍

e2e/cypress/support/commands.ts
bcmdarroch: oops 😅

PR #1874, opened 2019-10-16
Auth 1770/enforce unassigned perms on object create

blakestier:

We were allowing users who did not have permission on unassigned resources to _create_ unassigned resources (that they could not see afterward). We now validate project assignment on create for teams, tokens, policies, roles.

### :+1: Definition of Done

Users without permission on unassigned resources cannot create a resource with empty projects.
They can, however, update a resource to no longer have their project, even if that means that resource becomes unassigned.

```
rebuild components/teams-service
```
As an admin, create a project and token (non-admin).
Create a policy with role `project-owner` for that project with your token/local user as members.
```
curl -k -XPOST -d '{"name":"test", "id":"token-project-owner-policy", "members":["user:local:dave", "token:dave-token"], "statements":[{"effect":"ALLOW", "role":"project-owner", "projects":["foo-project"]}]}' -H "api-token: $TOK" $TARGET_HOST/apis/iam/v2beta/policies
```
Create a policy that permits create/updates on iam resources with the same members:
```
curl -k -XPOST -d '{"name":"test", "id":"dave-create-update-policy", "members":["user:local:dave", "token:dave-token"], "statements":[{"effect":"ALLOW", "actions":["iam:*:create", "iam:*:update"], "projects":["foo-project"]}]}' -H "api-token: $TOK" $TARGET_HOST/apis/iam/v2beta/policies
```

Attempt to create a team or token (UI) or policy or role (curl) without projects with that token or as that user. Should fail.
Attempt to create the above with the allowed project. Should succeed.
Attempt to update the above to remove allowed project. Should succeed.


lib/projectassignment/project_assignment.go
```diff
@@ -22,22 +34,38 @@ func EnsureProjectAssignmentAuthorized(ctx context.Context, authorizer engine.V2
...
+	previouslyUnassigned := len(oldProjects) == 0
+	newlyUnassigned := len(newProjects) == 0
+
+	// because we don't do patchy updates, users will need to submit the projects
+	// field on update even if they are not changing it. While a user is often
+	// not allowed to _create_ an unassigned resource, they may be permitted to
+	// change the name of one, provided the projects are not altered. In those
+	// cases, we do not want to check authorization for moving resources to
+	// being unassigned.
+	if isUpdateRequest && previouslyUnassigned && newlyUnassigned {
```
blakestier: We could also just append `(unassigned)` into both arrays and continue with the work below (as the diff would end up empty), but seemed worth saving us a little work.

components/authz-service/server/v2/authz.go
```diff
@@ -211,19 +211,34 @@ func (s *authzServer) ValidateProjectAssignment(
 		return &api.ValidateProjectAssignmentResp{}, nil
```
tylercloke: I bet the comment above this function will need an update.

A super invalid case feels like they passed a non-empty array and also the IsCreate flag. Can we throw an error in that case so other teams consuming this in the future don't misuse it by accident?

components/authz-service/server/v2/authz.go
```diff
@@ -211,19 +211,34 @@ func (s *authzServer) ValidateProjectAssignment(
...
-		if _, ok := err.(*projectassignment.ProjectsMissingError); ok {
-			return nil, status.Error(codes.NotFound, err.Error())
+	newProjects := req.NewProjects
+	if len(newProjects) == 0 {
+		newProjects = []string{}
+	}
+	oldProjects := req.OldProjects
+	if len(oldProjects) == 0 {
+		oldProjects = []string{}
+	}
```
tylercloke: Are these checking for the nil case? Is that possible to pass in with our validation? I feel like we should prevent this from being possible so that the API doesn't get used incorrectly by accident.

components/authz-service/server/v2/authz.go
```diff
@@ -331,13 +346,14 @@ func (v *VersionSwitch) Interceptor(ctx context.Context,
 
 	// These methods skip the check, though they are in the relevant service
 	// definition:
-	switch info.FullMethod {
-	case "/chef.automate.domain.authz.Authorization/GetVersion":
+	fullMethod := info.FullMethod
+	if strings.HasSuffix(fullMethod, "GetVersion") ||
+		strings.HasSuffix(fullMethod, "ValidateProjectAssignment") {
```
tylercloke: Not sure why we are adding this? I would think that we ValidateProjectAssignment should only ever be called in v2 mode? I might not be following here.

components/authz-service/server/v2/authz.go
tylercloke: Ah I think `InvalidCreateRequest` is covering this?

lib/projectassignment/project_assignment.go
```diff
@@ -6,12 +6,24 @@ import (
...
 )
 
-func EnsureProjectAssignmentAuthorized(ctx context.Context, authorizer engine.V2Authorizer, subjects, projectIDs []string) error {
+const UnassignedProjectID = "(unassigned)"
+
+func AuthorizeProjectAssignment(ctx context.Context, authorizer engine.V2Authorizer,
+	subjects, oldProjects, newProjects []string, isUpdateRequest bool) error {
+
+	if len(oldProjects) != 0 && !isUpdateRequest {
+		return NewInvalidCreateRequestError()
```
tylercloke: 👍

components/authz-service/server/v2/authz.go
blakestier: Yeah we should verify 👍

components/authz-service/server/v2/authz.go
blakestier: This is the token problem we were seeing-- we call `ValidateProjAssignment` in the tokens create code regardless of version. A comment would be helpful (or we make tokens version aware)...

e2e/cypress/integration/api/iam/02_policies_api.spec.ts
```diff
@@ -203,13 +203,14 @@ describeIfIAMV2('policies API', () => {
         });
```
msorens: Not shown but should change L113 from this
```
      headers: { 'api-token': nonAdminToken },
```
to
```
      headers: {  }, // must fill in before use
```
It looks like it is doing something useful but it is deceptive because at that point `nonAdminToken` is an empty string by definition.

e2e/cypress/integration/api/iam/02_policies_api.spec.ts
```diff
@@ -356,7 +358,7 @@ describeIfIAMV2('policies API', () => {
       });
 
       it('non-admin with project1 assignment access cannot update ' +
-      'a policy with no projects to have other projects', () => {
+      'a policy with no projects to have project1', () => {
```
msorens: This test is now a duplicate of the one on L304

e2e/cypress/integration/api/iam/02_policies_api.spec.ts
```diff
@@ -368,12 +370,30 @@ describeIfIAMV2('policies API', () => {
...
+        });
+
+        cy.request({ ...defaultNonAdminReq,
+            method: 'PUT',
+            url: `/apis/iam/v2beta/policies/${policyID}`,
+            body: policyWithProjects(policyID, [], statementProjects)
+        }).then((response) => {
+            expect(response.body.policy.projects).to.have.length(0);
+        });
+      });
```
msorens: 1. This test is now a duplicate of the one on L323.

2. I am puzzled though; I thought that this would be disallowed...?

lib/projectassignment/errors.go
```diff
@@ -10,37 +10,68 @@ type ProjectsMissingError struct {
...
-func NewProjectsMissingErroror(projectsMissing []string) error {
+func NewProjectsMissingError(projectsMissing []string) error {
 	return &ProjectsMissingError{projectsMissing: projectsMissing}
 }
 
 func (e *ProjectsMissingError) Error() string {
 	var errorStr string
 	if len(e.projectsMissing) > 1 {
-		errorStr = fmt.Sprintf("You cannot modify projects for this object because these projects did not exist: %q", e.projectsMissing)
+		errorStr = fmt.Sprintf("Projects for this object cannot be modified because these projects did not exist: %q", e.projectsMissing)
```
msorens: 1. Changing "did" to "do" was apparently missed so that should be fixed for consistency.
2. Consider also deleting "for this object" as that context seems clear.
```suggestion
		errorStr = fmt.Sprintf("Projects cannot be modified because these projects do not exist: %q", e.projectsMissing)
```

lib/projectassignment/errors.go
```diff
@@ -10,37 +10,68 @@ type ProjectsMissingError struct {
...
 	} else {
-		errorStr = fmt.Sprintf("You cannot modify projects for this object because project %q does not exist", e.projectsMissing)
+		errorStr = fmt.Sprintf("Projects for this object cannot be modified because project %q does not exist", e.projectsMissing)
```
msorens:
```suggestion
		errorStr = fmt.Sprintf("Projects cannot be modified because project %q does not exist", e.projectsMissing)
```

lib/projectassignment/errors.go
```diff
@@ -10,37 +10,68 @@ type ProjectsMissingError struct {
...
+		sub = "subjects are"
+	}
+
+	// just don't want the error msg out of hand
+	if len(authedSubs) > 5 {
+		sub = "(truncated) subjects are"
+		authedSubs = authedSubs[0:5]
+	}
+
+	return fmt.Sprintf("Projects for this object cannot be modified because the %q %s missing iam:projects:assign on %s: %q",
```
msorens:
```suggestion
	return fmt.Sprintf("Projects cannot be modified because the %q %s missing iam:projects:assign on %s: %q",
```

lib/projectassignment/project_assignment_test.go
```diff
@@ -7,37 +7,82 @@ import (
...
+	rand.Shuffle(len(cases), func(i, j int) {
+		cases[i], cases[j] = cases[j], cases[i]
+	})
+
+	for _, test := range cases {
+		t.Run(test.desc, test.f)
+	}
+}
+
+func TestCalculateProjectsToAuthorizeForCreate(t *testing.T) {
```
msorens: Nice set of tests!

components/authz-service/storage/v2/postgres/postgres.go
```diff
@@ -85,16 +85,20 @@ func (p *pg) CreatePolicy(ctx context.Context, pol *v2.Policy, checkProjects boo
 		return nil, p.processError(err)
 	}
 
-	if checkProjects {
-		err = p.ensureNoProjectsMissingWithQuerier(ctx, tx, pol.Projects)
+	projects := pol.Projects
```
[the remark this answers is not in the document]
blakestier: So we've discussed this a few times in the past. We have a few options, here:

1) A user can delete any resource they have permission to delete, regardless of projects.
2) A user can delete any resource that is in at least one of their projects.
3) A user can only delete resources if they are in projects they own (no unauthorized).

We previously landed on #1. @susanev do you want to revisit? Does another seem more appropriate now?

e2e/cypress/integration/api/iam/02_policies_api.spec.ts
blakestier: This case is allowed, although it is a tricky edge case. We decided to allow updates to unassigned for the time being.

components/authn-service/tokens/integration/token_integration_test.go
```diff
@@ -226,6 +223,7 @@ func newAuthzMock(t *testing.T) (authz.SubjectPurgeClient, authz_v2.Authorizatio
 	mockCommon := authz.NewSubjectPurgeServerMock()
 	mockCommon.PurgeSubjectFromPoliciesFunc = defaultMockPurgeFunc
 	authz.RegisterSubjectPurgeServer(g, mockCommon)
+	authz_v2.RegisterAuthorizationServer(g, mockV2Authz)
```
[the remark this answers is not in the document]
blakestier: We had to make some modifications to the authz server initialization to mock it properly in other files-- that resulted in some changes, here.

components/authz-service/storage/v2/postgres/postgres.go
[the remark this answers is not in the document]
susanev: im good with staying with #1 until we get user feedback that indicates otherwise.

e2e/cypress/integration/api/iam/02_policies_api.spec.ts
blakestier: Good catch-- I'll modify the other to leave another project behind

lib/projectassignment/project_assignment.go
```diff
@@ -62,5 +89,11 @@ func CalculateProjectDiff(oldProjects []string, newProjects []string) []string {
 		}
 	}
 
+	// if project goes from no projects to some or the reverse on update
+	// or has no projects on create
+	if (previouslyUnassigned && isUpdateRequest) || (newlyUnassigned && !isUpdateRequest) {
```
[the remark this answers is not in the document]
blakestier: We do have unit tests-- we *do* allow you to remove a project you have permissions on; therefore, we need to confirm that it's a create request. Otherwise, we don't permission on it.

https://github.com/chef/automate/pull/1874/files#diff-74f23555795d803f112f9f7a88c3ed2bR286

e2e/cypress/integration/api/iam/04_teams_api.spec.ts
```diff
@@ -152,23 +169,6 @@ describe('teams API', () => {
     describe('PUT /apis/iam/v2beta/teams', () => {
       context('both superadmin and project1 admin', () => {
         scenarios.forEach((scenario) => {
-          it(`${scenario.id} can update a team with no projects to have project1`, () => {
```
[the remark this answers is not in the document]
blakestier: A project1 admin does not have rights to assign their project to an unassigned resource.

components/authz-service/server/v2/authz.go
blakestier: No way to differentiate []string{} :( . We'll have to continue assuming that a nil projects field means (unassigned) (which seems ok to me)

components/authz-service/server/v2/authz.go
```diff
@@ -211,19 +211,34 @@ func (s *authzServer) ValidateProjectAssignment(
...
+	newProjects := req.NewProjects
+	if len(newProjects) == 0 {
+		newProjects = []string{}
+	}
+	oldProjects := req.OldProjects
+	if len(oldProjects) == 0 {
+		oldProjects = []string{}
+	}
+
+	if len(newProjects) != 0 {
```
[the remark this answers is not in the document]
blakestier: I wanted the security in case we called it from elsewhere (given that it's exposed)

PR #1905, opened 2019-10-18
deployment: use config information for dynamic port preflight checks

stevendanna:
The preflight checks were previously a hard-coded list of ports. This
list was out of date, did not include ports from optional collections,
and only tested the default port even if the user had changed their
configuration.

Now, we use the AutomateConfig struct and related interfaces to check
all ports, using any custom user configuration.

As a result of this change, in the default configuration, the
following unused ports will no longer be checked:

2133
2134
2192
10114
10118
10159

And the following new ports will be checked:

 4222 -> event-gateway
10101 -> cereal
10134 -> event-feed-service
10140 -> event-service
10130 -> authz-service
10132 -> event-service
10143 -> backup-gateway
10144 -> es-gateway
10145 -> pg-gateway
10147 -> event-service
10148 -> event-gateway
14001 -> data-feed-service

Further, when the user configures an optional product, the ports for
those products will also be checked.


components/automate-deployment/pkg/preflight/checks.go
```diff
@@ -364,168 +369,59 @@ func ConnectivityCheck(urls []string) Check {
...
+
+	servicesToCheck, _ := deployment.ExpectedServiceIDsForConfig(config.GetDeployment()) // nolint:errcheck
+	servicesToCheck = append(servicesToCheck, habpkg.New("chef", "deployment-service"))
+
+	portsToCheck := make([]int, 0, len(servicesToCheck)+8)
+	// Append habitat ports, currently hard coded nearly everywhere.
+	portsToCheck = append(portsToCheck, 9631, 9638)
+	acValue := reflect.ValueOf(*config)
+	for i := 0; i < acValue.NumField(); i++ {
+		if v, ok := acValue.Field(i).Interface().(a2conf.A2ServiceConfig); ok {
```
stevendanna: I'm thinking that instead of using reflect here, we might want to add some generated code that returns []a2conf.A2ServiceConfig

PR #1910, opened 2019-10-18
Generate fully namespaced entity names in swagger docs

phiggins:
This PR makes a change to our proto compilation to force it to prepend the entire package name onto the entity names generated in the swagger files. There were two things happening that made this necessary:

* The package names defined in our proto files often do not have a unique trailing section, for example a few proto's package names end in `.v1`. A few examples:
  * https://github.com/chef/automate/blob/master/components/automate-gateway/api/compliance/reporting/reporting.proto
   * https://github.com/chef/automate/blob/master/components/automate-gateway/api/compliance/reporting/stats/stats.proto
* `protoc-gen-swagger` attempts to generate unique strings to identify each entity by concatenating the trailing section of the package name and the name of the entity, resulting in strings like `v1Query` being generated for entities in different proto files.

The result of this was conflicts between different proto files causing things to be overwritten when generating the final swagger file used for API documentation. I had been working exclusively with the `compliance/reporting/reporting.proto` for so long I hadn't noticed this problem until I tried generating the docs for every endpoint.

The commit that updates the manually generated swagger file is a good illustration of the changes this PR has.

### DIFF
--- a/components/automate-chef-io/Makefile
+++ b/components/automate-chef-io/Makefile
@@ -3,11 +3,8 @@
 SHELL=bash
 SWAGGER_RESULT_FILE=static/api-docs/all-apis.swagger.json
 SWAGGER_DIR=data/docs/api_chef_automate
-#SWAGGER_FILES=$(shell find $(SWAGGER_DIR) -name '*.swagger.json')
-SWAGGER_FILES=data/docs/api_chef_automate/compliance/reporting/reporting.swagger.json
+SWAGGER_FILES=$(shell find $(SWAGGER_DIR) -name '*.swagger.json')
 STATIC_SWAGGER_FILES=$(shell find data/docs/api-static/ -type f | sort -n)
-
-EXPECTED_CONFLICTS=46
 
 themes/chef:
 	scripts/clone_hugo.sh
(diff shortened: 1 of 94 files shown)


components/automate-chef-io/Makefile
```diff
@@ -3,12 +3,9 @@
 SHELL=bash
 SWAGGER_RESULT_FILE=static/api-docs/all-apis.swagger.json
 SWAGGER_DIR=data/docs/api_chef_automate
-#SWAGGER_FILES=$(shell find $(SWAGGER_DIR) -name '*.swagger.json')
-SWAGGER_FILES=data/docs/api_chef_automate/compliance/reporting/reporting.swagger.json
+SWAGGER_FILES=$(shell find $(SWAGGER_DIR) -name '*.swagger.json')
 STATIC_SWAGGER_FILES=$(shell find data/docs/api-static/ -type f | sort -n)
 
-EXPECTED_CONFLICTS=46
```
stevendanna: 🎉

phiggins: I think this was mostly "solved" by `jq` eating the return value of the `swagger` call, unfortunately:

```
Petes-MacBook-Pro:automate-chef-io pete$ ruby -rjson -e "puts JSON.dump({foo: 'bar'}) ; exit 1" ; echo $?
{"foo":"bar"}
1
Petes-MacBook-Pro:automate-chef-io pete$ ruby -rjson -e "puts JSON.dump({foo: 'bar'}) ; exit 1" | jq . ; echo $?
{
  "foo": "bar"
}
0
```

Issue #1894, opened 2019-10-17
restrict nodes returned from api/v0/nodes/search to assigned projects


vjeffrey:
## User Story
We've implemented project/ingest rule assignment on all ingested nodes in client runs and compliance, and have implemented the logic to restrict access to those nodes based on the projects the user is allowed to view.
At report ingestion time, we send a message to the nodemanager with metadata about the node (name, tags, status, time of run/scan, assigned projects). In https://github.com/chef/automate/issues/536, we did work to ensure the nodemanager node references are always kept up to date with the latest projects information.

This leaves us with the following scenario:
- user is allowed to see nodes a, b -- not x, y, or z
- in client runs/compliance, user only sees nodes a, b as expected
- user calls api/v0/nodes/search, user sees metadata information about nodes a, b, x, y, z

Now we need to add the "filter by allowed projects" logic to the api/v0/nodes/search endpoint to be able to ensure the user can only see information about the nodes they have been assigned.
This means adding something similar to https://github.com/chef/automate/blob/master/components/compliance-service/api/reporting/server/server.go#L430
to the list (search) endpoint.

### Questions
But we haven't implemented project assignment for manually added nodes or nodemanager nodes yet...so how can we do this? how can we tell which is which?
```
manually added nodes: manager == 'automate'
nodemanager nodes: manager == 'aws-ec2', 'azure-api', etc
ingested nodes: manager == ''
```
also, only ingested nodes will have projects associated with them.

## Definition of Done
user that is only allowed to see nodes a, b in compliance reporting nodes or client runs nodes can only see nodes a, b (+manually added & nodemanager nodes) from the api/v0/nodes/search endpoint

## Demo Script / Repro Steps
- ingest 5 nodes, set up ingest rules/projects so a user is only allowed to see a portion of the nodes  (a, b -- not x, y, or z)
- in client runs/compliance, confirm, as that user, that you can only see nodes a, b
- call api/v0/nodes/search, only see info about nodes a, b + manually added + nodemanager nodes)

vjeffrey:
validated in iamv2p1 environment

Issue #55, opened 2019-04-13
Compliance-service endpoints need to guard against requests with no projects data


phiggins:
Related to chef/a2#5116

As discussed in the iamv2 meeting yesterday (4/11), for the initial implementation of projects filtering for `compliance-service` we are ignoring the case where incoming requests have no project data attached. By design all requests should include projects data, so this is guarding at the seams between services that this assumption is being maintained. 

Being lenient to non-compliant requests is useful in the short term to allow existing code that has no knowledge of projects to continue working, but eventually we will want to be strict in how that is handled.

lancewf:
This needs to be done for the config-mgmt-service and the node-manager also.

vjeffrey:
we cannot do this work until everyone is on iamv2. backing out of the sprint/board until then.

PR #1919, opened 2019-10-18
nodemanager project filtering

lancewf:
Adding project filtering to the "api/v0/nodes/search"/List RPC function. This is needed before IAMv2 GA because the nodemanager stores both ingested nodes and manually added nodes. These changes filter the ingested nodes only. The manually added nodes will not be filtered by the project tags in the context object. 

#1894

### :+1: Definition of Done
All ingested client run and compliance nodes are filtered by projects. 

1. `build components/nodemanager-service && start_all_services`
1. Turn on IAMv2.1 `chef-automate iam upgrade-to-v2 --beta2.1`
1. Ingest some client run nodes with `chef_load_nodes 10`
1. Ingest some compliance nodes with `chef_load_compliance_nodes 10`
1. ...

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/nodemanager-service/api/nodes/server/server.go
+++ b/components/nodemanager-service/api/nodes/server/server.go
@@ -266,6 +266,7 @@ func (srv *Server) Delete(ctx context.Context, in *nodes.Id) (*pb.Empty, error) 
 
 // List nodes based on a query
 func (srv *Server) List(ctx context.Context, in *nodes.Query) (*nodes.Nodes, error) {
+	logrus.Debugf("Getting Nodes with query: %+v", in)
 	filters, err := addProjectFilters(ctx, in.Filters)
 	if err != nil {
 		return nil, errorutils.FormatErrorMsg(err, "")
--- a/components/nodemanager-service/api/nodes/server/server.go
+++ b/components/nodemanager-service/api/nodes/server/server.go
@@ -11,11 +11,13 @@ import (
 	"google.golang.org/grpc"
 
 	"github.com/chef/automate/api/external/secrets"
+	"github.com/chef/automate/components/compliance-service/api/common"
 	"github.com/chef/automate/components/compliance-service/secretsint"
 	"github.com/chef/automate/components/nodemanager-service/api/nodes"
 	"github.com/chef/automate/components/nodemanager-service/mgrtypes"
 	"github.com/chef/automate/components/nodemanager-service/pgdb"
 	"github.com/chef/automate/lib/errorutils"
+	"github.com/chef/automate/lib/grpc/auth_context"
 	"github.com/chef/automate/lib/grpc/secureconn"
 )
 
@@ -264,8 +266,17 @@ func (srv *Server) Delete(ctx context.Context, in *nodes.Id) (*pb.Empty, error) 
 
 // List nodes based on a query
 func (srv *Server) List(ctx context.Context, in *nodes.Query) (*nodes.Nodes, error) {
-	logrus.Debugf("Getting Nodes with query: %+v", in)
-	dbnodes, totalCount, err := srv.db.GetNodes(in.Sort, in.Order, in.Page, in.PerPage, in.Filters)
+	logrus.Debugf("1Getting Nodes with query: %+v", in)
+	projectFilters, err := filterByProjects(ctx)
+	if err != nil {
+		return nil, errorutils.FormatErrorMsg(err, "")
+	}
+
+	filtersWithProjects := append(in.Filters, &common.Filter{
+		Key:    "project",
+		Values: projectFilters,
+	})
+	dbnodes, totalCount, err := srv.db.GetNodes(in.Sort, in.Order, in.Page, in.PerPage, filtersWithProjects)
 	if err != nil {
 		return nil, errorutils.FormatErrorMsg(err, "")
 	}
@@ -275,6 +286,18 @@ func (srv *Server) List(ctx context.Context, in *nodes.Query) (*nodes.Nodes, err
 		TotalUnreachable: totalCount.Unreachable,
 		TotalReachable:   totalCount.Reachable,
 		TotalUnknown:     totalCount.Unknown}, nil
+}
+
+func filterByProjects(ctx context.Context) ([]string, error) {
+	projectsFilter, err := auth_context.ProjectsFromIncomingContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if auth_context.AllProjectsRequested(projectsFilter) {
+		return []string{}, nil
+	}
+
+	return projectsFilter, nil
 }
 
 // Delete nodes based on a query
--- a/components/nodemanager-service/api/nodes/server/server.go
+++ b/components/nodemanager-service/api/nodes/server/server.go
@@ -266,17 +266,12 @@ func (srv *Server) Delete(ctx context.Context, in *nodes.Id) (*pb.Empty, error) 
 
 // List nodes based on a query
 func (srv *Server) List(ctx context.Context, in *nodes.Query) (*nodes.Nodes, error) {
-	logrus.Debugf("1Getting Nodes with query: %+v", in)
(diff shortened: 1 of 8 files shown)


components/nodemanager-service/api/nodes/server/server.go
```diff
@@ -264,8 +266,12 @@ func (srv *Server) Delete(ctx context.Context, in *nodes.Id) (*pb.Empty, error)
 
 // List nodes based on a query
 func (srv *Server) List(ctx context.Context, in *nodes.Query) (*nodes.Nodes, error) {
-	logrus.Debugf("Getting Nodes with query: %+v", in)
-	dbnodes, totalCount, err := srv.db.GetNodes(in.Sort, in.Order, in.Page, in.PerPage, in.Filters)
+	filters, err := addProjectFilters(ctx, in.Filters)
```
lancewf: Getting the project filters from the context and adding them to the query's filters.

components/nodemanager-service/pgdb/filtering.go
```diff
@@ -244,15 +246,29 @@ func whereNodeManagerNodeExists(field string, arr []string, tableAbbrev string)
 func whereProjectsMatch(_ string, arr []string, tableAbbrev string) (string, error) {
```
lancewf: Nice surprise to find the project filter was already in the code.

components/nodemanager-service/pgdb/filtering.go
```diff
@@ -244,15 +246,29 @@ func whereNodeManagerNodeExists(field string, arr []string, tableAbbrev string)
...
 	var condition string
 
-	if len(arr) == 0 {
-		condition = fmt.Sprintf("not exists (select 1 from projects p join nodes_projects np on p.id = np.project_id where np.node_id = %s.id)", tableAbbrev)
-	} else {
-		value, err := pq.Array(arr).Value()
+	refinedValues := stringutils.SliceFilter(arr, func(projectId string) bool {
```
lancewf: Create a list of values without the unassigned project ID included.

components/nodemanager-service/pgdb/filtering.go
```diff
@@ -244,15 +246,29 @@ func whereNodeManagerNodeExists(field string, arr []string, tableAbbrev string)
...
+		return projectId != authzConstants.UnassignedProjectID
+	})
+
+	if len(refinedValues) > 0 {
```
lancewf: Add all the allowed project tags to the query.

components/nodemanager-service/pgdb/filtering.go
```diff
@@ -244,15 +246,29 @@ func whereNodeManagerNodeExists(field string, arr []string, tableAbbrev string)
...
+		value, err := pq.Array(refinedValues).Value()
 		if err != nil {
 			return "", err
 		}
 		condition = fmt.Sprintf("exists (select 1 from projects p join nodes_projects np on p.id = np.project_id where np.node_id = %s.id and p.project_id = ANY('%s'::text[]))", tableAbbrev, value)
 	}
 
+	if stringutils.SliceContains(arr, authzConstants.UnassignedProjectID) {
```
lancewf: If there was an unassigned project ID add if the node has not project tags.

components/nodemanager-service/pgdb/filtering.go
```diff
@@ -244,15 +246,29 @@ func whereNodeManagerNodeExists(field string, arr []string, tableAbbrev string)
...
+		unassignedCondition := fmt.Sprintf("not exists (select 1 from projects p join nodes_projects np on p.id = np.project_id where np.node_id = %s.id)", tableAbbrev)
+		if condition == "" {
+			condition = unassignedCondition
+		} else {
+			condition = condition + " OR " + unassignedCondition
+		}
+	}
+
+	// Or the node is manually added
+	condition = condition + " OR n.manager <> ''"
```
lancewf: Also, include the node if it is a manually added node.

components/nodemanager-service/pgdb/nodes_integration_test.go
```diff
@@ -261,26 +261,6 @@ func (suite *NodesIntegrationSuite) TestGetNodesCanFilterByProjects() {
 	suite.Equal("Taco Node", fetchedNodes[0].Name)
 }
 
-func (suite *NodesIntegrationSuite) TestGetNodesCanFilterByNoProjects() {
```
lancewf: We have not resolved what to do in the case where an empty list of project tags is requested. 
https://github.com/chef/automate/issues/55
Currently, if an empty list of projects is requested, the nodes will not be filtered by projects.

bcmdarroch: That's how the IAM APIs work with project filtering. Is there a possibility this behavior will change?

components/nodemanager-service/pgdb/filtering.go
```diff
@@ -245,14 +251,32 @@ func whereProjectsMatch(_ string, arr []string, tableAbbrev string) (string, err
...
 		}
 		condition = fmt.Sprintf("exists (select 1 from projects p join nodes_projects np on p.id = np.project_id where np.node_id = %s.id and p.project_id = ANY('%s'::text[]))", tableAbbrev, value)
 	}
 
+	if stringutils.SliceContains(arr, authzConstants.UnassignedProjectID) {
+		unassignedCondition := fmt.Sprintf("not exists (select 1 from projects p join nodes_projects np on p.id = np.project_id where np.node_id = %s.id)", tableAbbrev)
+		if condition == "" {
+			condition = unassignedCondition
+		} else {
+			condition = condition + " OR " + unassignedCondition
```
msorens: Consider using a query builder instead of using raw strings.
(Why? See my article [Code Smells: Raw Strings and the Subtle Syntax Catastrophe](https://www.red-gate.com/simple-talk/dotnet/net-development/code-smells-raw-strings-subtle-syntax-catastrophe/) )

I have been wanting to try out the [Squirrel query builder](https://github.com/Masterminds/squirrel) for some time, so hoping to pique your interest. So this line generates your alternation expression.
```
	condition, _, _ =
		sq.Or{
			sq.Expr(condition),
			sq.Expr(unassignedCondition),
		}.ToSql()
```
You can similarly build up more complex expressions like you have on L270.

vjeffrey: while an interesting thing to look into, that seems like the kinda thing worthy of its own pr, since we would want to thoroughly test that all the filtering still works as expected

lancewf:
@bcmdarroch currently none of the node managers have project tags. Only the ingested nodes stored in the nodemanager-service themselves have project tags. These ingested nodes are duplicates of the nodes stored in client runs and compliance. All the other nodes stored in the nodemanager-service are manually added and can have several different node managers. Two of the node managers that I can remember are Automate and aws-ec2. We are not adding projects to these nodes until after IAMv2.1 GA.

components/nodemanager-service/pgdb/nodes_integration_test.go
lancewf: @bcmdarroch We have an issue #55 to look into how domain services should work if requests do not provide any projects. Are you saying that the current behavior that an empty list of project filters means don't filter projects or the "All" filter?

bcmdarroch: Yeah the way we've implemented [ProjectsAuthorized](https://github.com/chef/automate/blob/master/components/authz-service/server/v2/authz.go#L110), when we get a request without projects, we check what projects the subject has permission on out of all of the possible projects and return only those allowed projects, to be passed to the domain.

components/nodemanager-service/pgdb/filtering.go
lancewf: Thank you @msorens I have always used some kind of ORM before. With that library at least the SQL will be partially checked during compile time. 

@vjeffrey is right I would like to look into adding this type of work in a separate PR.

Issue #1862, opened 2019-10-15
ingested reports that include "ref"s do not display the refs


vjeffrey:
## Describe the bug
[original bug report](https://chefio.atlassian.net/secure/RapidBoard.jspa?rapidView=278&modal=detail&selectedIssue=A2-670)
When a report with "refs" on it is ingested, those refs are not viewable as part of the downloaded report.

## To Reproduce
- ingest a report with refs 
- executing this report with inspec on the cli, you will see a section "refs":
```
"refs": [
{"ref": "blabla", "url": "flafla"}
]
```
- ingest the report
- notice the report pulled out of compliance does not have the refs section

## Expected behavior
Refs show up on compliance report.

## Versions (please complete the following information):
- Automate Build Number: 20191007115919

vjeffrey:
wip branch: vj/refs

got confirmation that ref object should have ref:<string_value> and uri or url with a string value: https://github.com/inspec/inspec/blob/master/lib/inspec/schema.rb#L77

PR #1921, opened 2019-10-20
ensure refs with objects of type ref, url/uri are correctly read in and included on report

vjeffrey:
This change modifies the ingestion path to correctly read refs (or supported type, see https://github.com/chef/automate/issues/1862#issuecomment-542972093).
It also modified the report retrieval code to ensure the refs are included on the outputted report.

https://github.com/chef/automate/issues/1862

### :+1: Definition of Done
A report sent into Automate with refs includes those refs in the outputted report from Automate.

run a scan job using a profile with refs: 
[ref-test-0.1.0.tar.gz](https://github.com/chef/automate/files/3748255/ref-test-0.1.0.tar.gz)
ensure the downloaded json report includes the refs

- [ ] Tests added/updated?
- [ ] Docs added/updated? N/A

N/A

### DIFF
--- a/components/compliance-service/api/tests/03_reports_spec.rb
+++ b/components/compliance-service/api/tests/03_reports_spec.rb
@@ -755,6 +755,9 @@ describe File.basename(__FILE__) do
       assert_equal('nginx-01', res['profiles'][0]['controls'][0]['id'])
       assert_equal(1, res['profiles'][0]['controls'][0]['impact'])
       assert_equal(Google::Protobuf::Map, res['profiles'][0]['controls'][0]['tags'].class)
+      assert_equal("testing-ref", res['profiles'][0]['controls'][0]['refs'][0]['ref'])
+      assert_equal("test-url", res['profiles'][0]['controls'][0]['refs'][0]['url'])
+
       assert_equal('Running worker process as non-privileged user', res['profiles'][0]['controls'][0]['title'])
       assert_equal(1, res['profiles'][0]['controls'][0]['results'].length)
 
--- a/components/compliance-service/ingest/events/compliance/helper_methods.go
+++ b/components/compliance-service/ingest/events/compliance/helper_methods.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/golang/protobuf/jsonpb"
 	structpb "github.com/golang/protobuf/ptypes/struct"
+	"github.com/sirupsen/logrus"
 
 	"github.com/chef/automate/components/compliance-service/ingest/events/inspec"
 	inspecTypes "github.com/chef/automate/components/compliance-service/inspec"
@@ -116,6 +117,29 @@ func ReportProfilesFromInSpecProfiles(profiles []*inspec.Profile, profilesSums [
 				}
 			}
 
+			refs := make([]relaxting.ESInSpecReportControlRefs, 0)
+			var refVal, urlVal string
+			for _, ref := range control.Refs {
+				if len(ref.Fields) != 2 {
+					logrus.Warnf("ref object contains more than two fields: %v", ref)
+					continue
+				}
+				for key, val := range ref.Fields {
+					if key == "ref" {
+						refVal = val.GetStringValue()
+					} else {
+						// both url and uri values accepted here.
+						urlVal = val.GetStringValue()
+					}
+
+				}
+
+				refs = append(refs, relaxting.ESInSpecReportControlRefs{
+					Ref: refVal,
+					Url: urlVal,
+				})
+			}
+
 			minControls[i] = relaxting.ESInSpecReportControl{
 				ID:         control.Id,
 				Title:      control.Title,
@@ -123,6 +147,7 @@ func ReportProfilesFromInSpecProfiles(profiles []*inspec.Profile, profilesSums [
 				Status:     control.Status(),
 				Results:    minResults,
 				StringTags: stringTags,
+				Refs:       refs,
 			}
 		}
 
--- a/components/compliance-service/ingest/ingestic/mappings/comp-rep-date.go
+++ b/components/compliance-service/ingest/ingestic/mappings/comp-rep-date.go
@@ -228,6 +228,37 @@ var ComplianceRepDate = Mapping{
                     }
                   }
                 },
+                "refs": {
+                  "type": "nested",
+                  "properties": {
+                    "key": {
+                      "type": "keyword",
+                      "fields": {
+                        "engram": {
+                          "type": "text",
+                          "analyzer": "autocomplete"
+                        },
+                        "lower": {
+                          "type": "keyword",
+                          "normalizer": "case_insensitive"
+                        }
+                      }
+                    },
+                    "values": {
+                      "type": "keyword",
+                      "fields": {
+                        "engram": {
+                          "type": "text",
+                          "analyzer": "autocomplete"
+                        },
+                        "lower": {
+                          "type": "keyword",
+                          "normalizer": "case_insensitive"
+                        }
+                      }
+                    }
+                  }
+                },
                 "string_tags": {
                   "type": "nested",
                   "properties": {
--- a/components/compliance-service/reporting/relaxting/es_types.go
+++ b/components/compliance-service/reporting/relaxting/es_types.go
@@ -103,6 +103,11 @@ type ESInSpecReportControlStringTags struct {
 	Values []string `json:"values"`
 }
 
+type ESInSpecReportControlRefs struct {
+	Ref string `json:"ref"`
+	Url string `json:"url"`
+}
+
 type ESInSpecReportControl struct {
 	ID         string                            `json:"id"`
 	Impact     float32                           `json:"impact"`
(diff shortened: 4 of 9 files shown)


components/compliance-service/reporting/relaxting/reports.go
```diff
@@ -553,8 +553,14 @@ func convertControl(profileControlsMap map[string]*reportingapi.Control, reportC
 
 	convertedControl.StringTags = jsonTags
 	var jsonRefs []*reportingapi.Ref
-	refs, _ := json.Marshal(profileControl.Refs)
-	json.Unmarshal(refs, &jsonRefs) // nolint: errcheck
+	refs, err := json.Marshal(reportControlMin.Refs)
```
stevendanna: [question] Would it help readability if we made this a function with an explicit name?  It is a bit strange since `jsonRefs` is type `[]*reportingapi.Ref`  and `reportControlMin.Refs` is _also_ of type `[]*reportingapi.Ref`, so I'm not clear on why we don't copy this from the reportControlMin to the convertedControl directly.

vjeffrey: reportControlMin.Refs was a different type, ESInSpecReportControlRef - which is why we had to marshal/unmarshal, but after taking a look i figured i can just make it the same type throughout and then just copy as you said 🎉

stevendanna: > reportControlMin.Refs was a different type, ESInSpecReportControlRef 

Interesting! My editor lied to me! I shall forgive it, but just this once.

Thanks for fixing this up nonetheless :D

components/compliance-service/ingest/ingestic/mappings/comp-rep-date.go
```diff
@@ -228,6 +228,37 @@ var ComplianceRepDate = Mapping{
...
                   }
                 },
+                "refs": {
+                  "type": "nested",
+                  "properties": {
+                    "key": {
+                      "type": "keyword",
+                      "fields": {
+                        "engram": {
+                          "type": "text",
```
vjeffrey: is this right??  this is a copy/pasta job. i have no idea what i'm doing when it comes to es mappings

components/compliance-service/reporting/relaxting/es_types.go
```diff
@@ -110,6 +111,7 @@ type ESInSpecReportControl struct {
 	Status     string                            `json:"status"`
 	Results    []*ESInSpecReportControlsResult   `json:"results"`
 	StringTags []ESInSpecReportControlStringTags `json:"string_tags"`
+	Refs       []*reportingapi.Ref               `json:"refs"`
```
vjeffrey: is this ok? i see the other types in here are types defined in this file. is there a reason i should avoid using reportingapi.Ref here as a type? cc @alexpop @rickmarry

components/compliance-service/ingest/ingestic/mappings/comp-rep-date.go
lancewf: You would only need the `engram` and `lower` fields if this value was going to be in the search bar or be used to filter on.

components/compliance-service/reporting/relaxting/es_types.go
rickmarry: I don't see it as a problem. All attributes of Ref are exported so you won't run into any visibility issues in using this.  There are other structs in this files that use type that are defined in other packages too.

lancewf: The `reportingapi.Ref` is defined for the RPC API and the `ESInSpecReportControl` is defined for Elasticsearch. I would say that another `Ref` should be defined in this file. Maybe a "ESRef".

components/compliance-service/ingest/ingestic/mappings/comp-rep-date.go
vjeffrey: ah cool, good to know, thank you! ya, no plans to include this in searchbar atm

components/compliance-service/ingest/ingestic/mappings/comp-rep-date.go
```diff
@@ -228,6 +228,37 @@ var ComplianceRepDate = Mapping{
                     }
                   }
                 },
+                "refs": {
+                  "type": "nested",
+                  "properties": {
+                    "key": {
```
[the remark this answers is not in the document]
vjeffrey: errr...probably? i am so completely unfamiliar with es mappings. i just copy/pasted til it worked -- ill try to change it and see what happens!

components/compliance-service/reporting/relaxting/es_types.go
rickmarry: >  I would say that another `Ref` should be defined in this file. Maybe a "ESRef".
I suppose for consistency you're right @lancewf but it seems like we have a lot of these mirror structs in our code

lancewf: Creating the new ESRef struct would allow flexibility in changing how the data comes in through the API and how we store the data with Elasticsearch.

rickmarry: cool

components/compliance-service/reporting/relaxting/reports.go
vjeffrey: so i ended up reverting the change where i made the types the same; see commit msg: 
```
there was some discussion on the pr about this. making the ref
type consistent means putting the api defined type in the es structs
file, which is bad practice b/c we're then bound to the api definition of
the struct when dealing with elasticsearch. so reverting the change!
```

stevendanna: That's cool.  Note that we could still avoid round-tripping through JSON with something like (warning: code written in comment window):

```golang
convertedControl.Refs = make([]*reportingapi.Ref, len(reportControlMin.Refs))
for i, ref := range reportControlMin.Refs {
	convertedControl.Refs[i] = &reportingapi.Ref{
		Ref: ref.Ref,
		Url: ref.Url,
	}
}
```

My _guess_ is this is a bit faster and has the benefit of not having any errors to handle.  However, I'm also fine with leaving this for some future cleanup PR.

components/compliance-service/ingest/events/compliance/helper_methods.go
```diff
@@ -116,13 +117,37 @@ func ReportProfilesFromInSpecProfiles(profiles []*inspec.Profile, profilesSums [
 				}
 			}
 
+			refs := make([]relaxting.ESInSpecReportControlRefs, 0)
```
stevendanna: If we think the `continue` case is going to be rare, you can do:

```suggestion
			refs := make([]relaxting.ESInSpecReportControlRefs, 0, len(control.Refs))
```

components/compliance-service/reporting/relaxting/reports.go
vjeffrey: that does seem safer than the json round tripping -- ill make the change. thank you!!

components/compliance-service/ingest/events/compliance/helper_methods.go
vjeffrey: i honestly have no idea if the continue case will be rare or not, cause there's no real type checking on the inspec profiles. someone could put completely diff data in that block on the profile, and we would get that on the report :/

components/compliance-service/api/tests/03_reports_spec.rb
```diff
@@ -755,6 +755,9 @@ def reporting
       assert_equal('nginx-01', res['profiles'][0]['controls'][0]['id'])
       assert_equal(1, res['profiles'][0]['controls'][0]['impact'])
       assert_equal(Google::Protobuf::Map, res['profiles'][0]['controls'][0]['tags'].class)
+      assert_equal("testing-ref", res['profiles'][0]['controls'][0]['refs'][0]['ref'])
+      assert_equal("test-url", res['profiles'][0]['controls'][0]['refs'][0]['url'])
```
phiggins: These might be easier to read if you used some local variables like:

```
control = res['profiles'].first['controls'].first
# ...
assert_equal('nginx-01', control.id)
# ...
ref = control['refs'].first
assert_equal("testing-ref", ref.ref)
```

components/compliance-service/ingest/events/compliance/helper_methods.go
```diff
@@ -116,13 +117,37 @@ func ReportProfilesFromInSpecProfiles(profiles []*inspec.Profile, profilesSums [
...
+			for _, ref := range control.Refs {
+				var refVal, urlVal string
+				if len(ref.Fields) != 2 {
+					logrus.Warnf("ref object contains more than two fields: %v", ref)
+					continue
+				}
```
phiggins: Does this need to handle the case where it's not 2, where one or both of these values are not present?

vjeffrey: i think for now it's ok to be strict on this. there's a recommended data format for inspec (ref + uri/url) -- we don't want to reject reports that don't adhere to this b/c there isn't strict typing on the profile format, we just want to support the users that do correctly use the feature

PR #1935, opened 2019-10-21
WIP: Quick POC for how we could clean up non-existent projects in tea…

tylercloke:
…ms and tokens

Signed-off-by: Tyler Cloke <tylercloke@gmail.com>

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/teams-service/cmd/teams-service/main.go
+++ b/components/teams-service/cmd/teams-service/main.go
@@ -134,6 +134,7 @@ Please pass a config file as the only argument to this command.`))
 	authzClient := authz.NewSubjectPurgeClient(authzConn)
 	authzV2PoliciesClient := authz_v2.NewPoliciesClient(authzConn)
 	authzV2AuthorizationClient := authz_v2.NewAuthorizationClient(authzConn)
+	authzV2ProjectsClient := authz_v2.NewProjectsClient(authzConn)
 
 	mustBeADirectory(cfg.MigrationsPath)
 	mustBeADirectory(cfg.DataMigrationsPath)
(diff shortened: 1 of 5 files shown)


sdelano:
@tylercloke thanks for the quick POC on this. Let me check if I understand / can summarize this correctly.

Firstly, there's a regular old SQL migration in here that sets up the database so we can run special data migrations in Golang. :ballot_box_with_check: 

We need to run this migration in Golang here because it needs to be dynamic. We need data from another service (in this case we need to know what projects exist) and based on that data we need to update the records in our service. It's essentially a cleanup task. Our current migration tool, `migrate` doesn't support this.

So now, here's a rough idea of how our service startup goes:
1. run all the sql migrations that haven't been run
2. run all the golang migrations that haven't been run

I have a feeling that this will potentially cause problems at some point down the road. Imagine that we create more sql migrations that change how the code in the golang migrations works. Some customers will have already run the golang migrations. Some won't have because they'll upgrade many versions at once. Do we have a way to ensure that all the golang migrations will always run against the latest version of the sql schema? Is that what happens on a fresh intsall? Do we already have sufficient testing for this with our "upgrade from the may 2018 version (aka deep upgrade)" test?

Thoughts?

sdelano:
@tylercloke is it true that on a fresh install we'll:

1. run all the sql migrations 1-x
2. run all the golang migtations 1-x

If that's the case - it seems that we should catch at least the most basic case where this would fail when we do the fresh install test - the schema of the migration code not matching the new schema of the database.

tylercloke:
@sdelano Summary and second comment are all accurate. All the schema migrations will always run before all the golang migrations. So we'd just have to keep the golang migrations up to date with any schema changes. That is a downside to this approach for sure.

msorens:
Concerns noted above need to be kept in mind for the future, of course, and those would likely involve some further iteration, but seems like this will certainly meet our present needs.

PR #1948, opened 2019-10-23
Attempt to de-flake projects API tests

tylercloke:
There was a slight logic error in how the totalNodes was being calculated. We were using an unassigned filter inconsistently.

I also removed a ton of the unnecessary parts of the fixture JSON to hopefully speed things up.


e2e/cypress/integration/api/iam/01_projects_api.spec.ts
```diff
@@ -163,21 +163,9 @@ describeIfIAMV2p1('projects API', () => {
...
-          'api-token': Cypress.env('ADMIN_TOKEN'),
-          projects: '(unassigned)'
-        },
-        method: 'GET',
-        url: '/api/v0/cfgmgmt/nodes?pagination.size=10'
-      }).then((response) => {
-        expect(response.body).to.have.length(totalNodes + 4);
-      });
+      const maxRetries = 200;
+      waitForUnassignedNodes(totalNodes + 4, maxRetries);
```
sdelano: What's the `4` here for and what's the potential for that to change?

tylercloke: I can make it a constant but it's just how many nodes we are adding.

bcmdarroch: Four for the four node fixtures that get ingested as part of the test. We could make an array variable of [the nodes](https://github.com/chef/automate/pull/1948/files#diff-e2f3f7d6dd26e33ac215cf69fb13ef5dR154) and use the length of that

PR #1949, opened 2019-10-23
error_utils: Don't log time into message

stevendanna:
Our standard Logrus setup already logs the timestamp. Additionally,
systemd will typically also log a timestamp. Including a timestamp
here meant that we were logging 3 timestamps for this message.

I've also done some minor formatting cleanups on our error returns
that I found when looking at callers to this function.

### DIFF
--- a/lib/errorutils/error_utils.go
+++ b/lib/errorutils/error_utils.go
@@ -4,7 +4,6 @@ import (
 	"database/sql"
 	"fmt"
 	"net"
-	"time"
 
 	"github.com/lib/pq"
 	"github.com/olivere/elastic"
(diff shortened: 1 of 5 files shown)


lib/errorutils/error_utils.go
```diff
@@ -85,7 +84,7 @@ func FormatErrorMsg(err error, id string) error {
 	if err == nil {
 		return nil
 	}
-	logrus.Error(fmt.Sprintf("%s:%s", time.Now().UTC(), err.Error()))
```
stevendanna: This is the meat of the change.

PR #1950, opened 2019-10-23
hide global projects filter if zero projects

### LINKED ISSUE #1274: hide global projects filter if a user has access to zero projects
_Problem: A user with zero access is presented with a confusing dropdown that does not open._

## Overview
If a user has access to zero projects, the global projects filter should be hidden from the top-nav. Currently, they are shown an all projects dropdown with a count of zero, that does not open because there are zero projects.

### All zero projects, lol
zero-projects

### Dropdown does not open, cause there are zero projects
zero-projects-dropdown-but-no-functionaility

vinay033:

If a user has access to zero projects, the global project filter should be hidden from the top-nav. Currently, they are shown an all projects dropdown with a count of zero, that does not open because there are zero projects.
project fileter

fixes #1274

### :+1: Definition of Done
If a user logged in and if the project count is `0` then the project dropdown is hidden on the dashboard.
logged in as non-admin user navigate to the dashboard.
### :camera: Screenshots
Chef Automate_hide_project_dropdown

### DIFF
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
@@ -84,6 +84,6 @@ export class ProjectsFilterDropdownComponent implements OnChanges {
   }
 
   filterVisible() {
-    return this.selectionCount === 0 ? false : true;
+    return this.selectionCount > 0;
   }
 }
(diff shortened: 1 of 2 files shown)


components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
```diff
@@ -82,4 +82,8 @@ export class ProjectsFilterDropdownComponent implements OnChanges {
       (element as HTMLElement).focus();
     }
   }
+
+  filterVisible() {
+    return this.selectionCount === 0 ? false : true;
```
msorens:
```suggestion
    return this.selectionCount > 0;
```

msorens:
Also, in your "How to Build and Test the Change", please try to always include the list of components that need to be rebuilt. In this case, automate-ui.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
vinay033: @msorens I have updated the changes.

PR #1963, opened 2019-10-23
Fix bug where duplicate projects getting returned from OPA in 2.1 could result in *

tylercloke:

This is an edge case where the duplicates would be the same length as the total projects in the system.

For example, if you have a single custom project named project1 one and two policies that allow user bob on project1 for a specific resource and action, then` V2ProjectsAuthorized` would return `["project1", "project1"]`. Since the total projects in the system would be `["(unassigned)", "project1"]`, `ProjectsAuthorized` would incorrectly equate the result of` V2ProjectsAuthorized` to `*` due to the fact that their lengths matched. This edge case would happen anytime `NumberOfProjectsAllowingAProjectForASpecificResourceAndAction` = `NumberOfCustomProjects` + 1 (for the unassigned project).

`V2ProjectsAuthorized` has been updated to not return duplicates which was never the intended behavior. The duplicates were introduced in this change:

https://github.com/chef/automate/commit/c21ddbe4cfd751f0b5513211a439caaef6cddabf

Due to the introduction of using a partial document:

https://github.com/chef/automate/commit/c21ddbe4cfd751f0b5513211a439caaef6cddabf#diff-2d317bb33ff7babca7e1290b0fb1dcf2R241

Which unfortunately contain duplicates:

https://github.com/open-policy-agent/opa/issues/429

Added tests that did fail before my change was applied that reproduce the situation described below.

1) To see the bug, start your dev env off `dev`.
2) Create a user named `dave`
3) Create a single custom project named `foo-projects`
4) Create a policy for dave granting him access on `foo-project` to specific resources (in this case everything):

```
curl -k -XPOST -d '{"name":"test", "id":"dave-all-policy", "members":["user:local:dave"], "statements":[{"effect":"ALLOW", "actions":["*"], "projects":["foo-project"]}]}' -H "api-token: $TOK" $TARGET_HOST/apis/iam/v2beta/policies
```

Visit the UI, check the logs etc. He only sees things scoped to `foo-project`. Note the `result` field of the ProjectsAuthorized query:

```
[2019-10-23T18:47:08+00:00] authz-service.default(O): time="2019-10-23T18:47:08Z" level=info msg="Projects Authorized Query" action="iam:tokens:list" projects="[]" resource="iam:tokens" result="[foo-project]" subject="[user:local:dave]"
```

Now, add another policy that give him access to some overlapping resources and actions also in `foo-project`. Now, for those overlapping resources, authz will return `*` because `V2ProjectsAuthorized` will return `["foo-project", "foo-project"]` which will match the length of total projects in the system `["(unassigned)", "foo-project"]`:

```
curl -k -XPOST -d '{"name":"test", "id":"dave-policy", "members":["user:local:dave", "token:dave-token"], "statements":[{"effect":"ALLOW", "role":"project-owner", "projects":["foo-project"]}]}' -H "api-token: $TOK" $TARGET_HOST/apis/iam/v2beta/policies
```

Actions overlapping with `project-owner` just added results in `*`:

```
[2019-10-23T18:48:17+00:00] authz-service.default(O): time="2019-10-23T18:48:17Z" level=info msg="Projects Authorized Query" action="iam:teams:list" projects="[]" resource="iam:teams" result="[*]" subject="[user:local:dave]"
```

Actions that don't overlap will not (because `["foo-project"]` is length 1):

```
[2019-10-23T18:47:08+00:00] authz-service.default(O): time="2019-10-23T18:47:08Z" level=info msg="Projects Authorized Query" action="iam:tokens:list" projects="[]" resource="iam:tokens" result="[foo-project]" subject="[user:local:dave]"
```

NOW, add another custom project, you don't even have to use it. Now dave will only have access to `foo-project` for all resources and actions, including actions that overlap with the `project-owner` role, because `len(["foo-project", "foo-project"]) != len(["my-new-project", "foo-project", "(unassgined)"]`. However, if you added one more overlapping policy, then things would break again: ``len(["foo-project", "foo-project", "foo-project"]) == len(["my-new-project", "foo-project", "(unassgined)"]`.

Feel free to try this all out then `rebuild components/authz-service` and everything will work.

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/authz-service/engine/conformance/conformance_v2_test.go
+++ b/components/authz-service/engine/conformance/conformance_v2_test.go
@@ -99,7 +99,6 @@ func TestV2p1ProjectsAuthorized(t *testing.T) {
 				require.NoError(t, err)
 				assert.Equal(t, []string{}, actual)
 			})
-
 			t.Run("policy with one of the requested projects returns matched project", func(t *testing.T) {
 				pol := map[string]interface{}{
 					"members": engine.Subject(sub),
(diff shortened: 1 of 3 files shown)


components/authz-service/server/v2/system_integration_test.go
```diff
@@ -59,6 +60,68 @@ func TestIntegrationSystemPolicies(t *testing.T) {
 	}
 }
 
+func TestIntegrationValidateProjectAssignmentWithOverlappingProjects(t *testing.T) {
```
blakestier: Here's hoping this doesn't fail with my branch :(

components/authz-service/engine/opa/opa.go
```diff
@@ -286,7 +287,12 @@ func dumpData(ctx context.Context, store storage.Store, l logger.Logger) error {
 		return err
 	}
 
-	l.Infof("data: %#v", data)
+	jsonData, err := json.Marshal(data)
```
bcmdarroch: 🙌

components/authz-service/engine/conformance/conformance_v2_test.go
```diff
@@ -140,6 +139,39 @@ func TestV2p1ProjectsAuthorized(t *testing.T) {
...
+							"projects":  []string{"other-project", proj2},
+						},
+					},
+				}
+				role := map[string]interface{}{
+					"id":      "handyman",
+					"actions": []string{act},
+				}
+				setPoliciesV2p1(t, e, pol1, pol2, role)
+				actual, err := e.V2ProjectsAuthorized(args([]string{proj1, proj2}))
```
msorens: To get deeper coverage, consider expanding this test to take a string[] as input, then feed in these separate test inputs:
```
[]string{proj1} // non-repeated project is returned without repeats
[]string{proj2} // repeated project is returned without repeats
[]string{proj1, proj2} // non-repeated and repeated are returned without repeats
[]string{proj1, proj2, "not-defined policy"} // superfluous policies are skipped while others are returned without repeats
```

tylercloke: Not sure what you mean by the last test but the others should be unit tests of the opa code. I'll add a few unit tests there.

PR #1965, opened 2019-10-24
Automate dev docs updates

phiggins:
I wanted to add a note about compiling all the proto files at once and noticed a few missing things and formatting inconsistencies.

### DIFF
--- a/dev-docs/DEV_ENVIRONMENT.md
+++ b/dev-docs/DEV_ENVIRONMENT.md
@@ -97,7 +97,7 @@ following prerequisites:
 
    1. Determine the supported version of Habitat. It is 0.77.0 at the time of writing this, but the current version can be found in [studio-common's `.hab-version` file](https://github.com/chef/studio-common/blob/master/.hab-version).
    2. To install version `0.77.0` of Habitat:
-  `curl https://raw.githubusercontent.com/habitat-sh/habitat/master/components/hab/install.sh | sudo bash -s -- -v 0.77.0`
+  `curl https://raw.githubusercontent.com/habitat-sh/habitat/master/components/hab/install.sh | sudo bash -s -- -v 0.77.0` or refer to the [Habitat installation guide](https://www.habitat.sh/docs/install-habitat/).
    3. To configure the Habitat cli run and follow the instructions:
   `hab cli setup`
 
--- a/dev-docs/DEV_ENVIRONMENT.md
+++ b/dev-docs/DEV_ENVIRONMENT.md
@@ -225,6 +225,15 @@ Anytime you edit a .proto file you need to recompile it.
 compile_go_protobuf_component <COMPONENT_NAME>
 ```
 
+Some proto files can not be recompiled individually. If you are seeing
+unexpected changes after recompiling a .proto file (such as `git status`
+reporting missing files) or just want to be on the safe side, you can
+recompile all .proto files at once:
+
+```
+compile_all_protobuf_components
+```
+
 ### Accessing the Logs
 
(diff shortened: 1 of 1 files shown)


dev-docs/DEV_ENVIRONMENT.md
```diff
@@ -1,4 +1,4 @@
-# A2 Development Environment
+# Automate Development Environment
```
phiggins: I changed most of the references to A2 in this document to Automate to reflect the current state of the project. There are many references to `a2-dev.test` that I did not change because that is likely a larger effort.

dev-docs/DEV_ENVIRONMENT.md
```diff
@@ -78,27 +78,27 @@ password, respectively.
...
 
-1. Install `direnv` and integrate it into your shell
+2. Install `direnv` and integrate it into your shell
 
-   a. First, install the package. On Homebrew: run `brew install direnv`, on Ubuntu: run `apt-get install direnv`
-   b. Next, [hook direnv into your shell](https://github.com/direnv/direnv#setup)
-   c. Finally, run `direnv allow` from the A2 source directory.
+   1. First, install the package. On Homebrew: run `brew install direnv`, on Ubuntu: run `apt-get install direnv`
+   2. Next, [hook direnv into your shell](https://github.com/direnv/direnv#setup)
+   3. Finally, run `direnv allow` from the Automate source directory.
```
phiggins: Using abc's for ordered lists is not valid markdown syntax, so some of this formatting was being eaten when the document was rendered.

dev-docs/DEV_ENVIRONMENT.md
```diff
@@ -78,27 +78,27 @@ password, respectively.
...
    From a terminal:
 
-   a. To install version `0.59.0` of Habitat:
-  `curl https://raw.githubusercontent.com/habitat-sh/habitat/master/components/hab/install.sh | sudo bash -s -- -v 0.59.0`
-
-   b. To configure the Habitat cli run and follow the instructions:
+   1. Determine the supported version of Habitat. It is 0.77.0 at the time of writing this, but the current version can be found in [studio-common's `.hab-version` file](https://github.com/chef/studio-common/blob/master/.hab-version).
+   2. To install version `0.77.0` of Habitat:
+  `curl https://raw.githubusercontent.com/habitat-sh/habitat/master/components/hab/install.sh | sudo bash -s -- -v 0.77.0` or refer to the [Habitat installation guide](https://www.habitat.sh/docs/install-habitat/).
+   3. To configure the Habitat cli run and follow the instructions:
```
phiggins: If someone followed these instructions exactly they would install an outdated version of Habitat and would see a big scary warning when starting the studio that they are using an unsupported version of Habitat.

dev-docs/DEV_ENVIRONMENT.md
```diff
@@ -225,6 +225,15 @@ Anytime you edit a .proto file you need to recompile it.
...
 ```
 
+Some proto files can not be recompiled individually. If you are seeing
+unexpected changes after recompiling a .proto file (such as `git status`
+reporting missing files) or just want to be on the safe side, you can
+recompile all .proto files at once:
+
+```
+compile_all_protobuf_components
+```
```
phiggins: This only applies to the proto files in `api/`, if I remember correctly.

dev-docs/DEV_ENVIRONMENT.md
stevendanna: 👍 As a follow up, my vote is that we remove the check.  We should be encouraging people to aggressively upgrade hab in the studio so we can find errors.

PR #1967, opened 2019-10-24
preflight: larger disk space check for builder

stevendanna:

### DIFF
--- a/components/automate-deployment/pkg/preflight/checks.go
+++ b/components/automate-deployment/pkg/preflight/checks.go
@@ -17,7 +17,6 @@ import (
 	"github.com/chef/automate/components/automate-deployment/pkg/services"
 	"github.com/chef/automate/components/automate-grpc/protoc-gen-a2-config/api/a2conf"
 	"github.com/chef/automate/lib/proc"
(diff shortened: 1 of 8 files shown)


components/automate-deployment/pkg/preflight/checks.go
```diff
@@ -33,8 +35,19 @@ func RootUserRequiredCheck() Check {
...
 }
 
-func DefaultMinimumDiskCheck() Check {
-	return MinimumDiskCheck(5 * (1 << 30))
+const GB = 1 << 30
+
+func minDiskBytesForConfig(c *dc.AutomateConfig) uint64 {
+	collections := deployment.CollectionsForConfig(c.GetDeployment())
+	if stringutils.SliceContains(collections, services.BuilderCollectionName) {
+		return 15 * GB
```
jaym: [nit] should we sum across the products. So builder alone with be 15. builder + automate would be 20

products.meta
```diff
@@ -120,6 +120,13 @@
         "chef/automate-gateway"
       ]
     },
+    {
+      "name": "builder",
```
jaym: if you merge this:
```
"hidden": true
```

PR #1969, opened 2019-10-24
ci: Add test for deploying with the unstable Habitat build

stevendanna:
We would like to be able to upgrade to the latest version to hab much
more quickly than we currently do. One issue that causes delay is that
we often don't test the latest version of Habitat until after it is
released. If we find an issue that we need to address upstream in
Habitat, this delays our upgrade as we need to submit the fix, wait
for the release, and then re-test.

The goal of this change is to get us continuously testing the dev
builds in unstable so we can catch issues or confirm fixes before the
release of Hab.

### DIFF
--- a/components/automate-deployment/pkg/server/startup_helpers.go
+++ b/components/automate-deployment/pkg/server/startup_helpers.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 
+	"github.com/chef/automate/components/automate-deployment/pkg/habpkg"
 	"github.com/chef/automate/lib/platform/command"
 )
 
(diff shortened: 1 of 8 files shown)


components/automate-deployment/pkg/server/startup_helpers.go
```diff
@@ -37,6 +38,35 @@ func habSupVersion() string {
 	return strings.TrimSpace(output)
 }
 
+// maybeResetHabPath adds the bindir from our manifest's hab package
+// to our PATH. This is a developent-mode only operation to account
+// for test cases where we want to deploy with the latest Habitat.
+func maybeResetHabPath(s *server) {
```
stevendanna: A perhaps better path would be to _always_ use the version from the manifest via `hab pkg exec`, avoiding this problem completely; however, that change will require a bit more testing.

jaym: i'm kind of surprised we're not already automatically calling the hab from the manifest

PR #1976, opened 2019-10-24
compliance: turn off cereal-service based managers in test mode

stevendanna:
Until we integrate cereal-service into compliance's test setup, having
these turned on means we get constant spamming of errors in the logs,
making it harder to debug problems.

### DIFF
--- a/components/compliance-service/compliance.go
+++ b/components/compliance-service/compliance.go
@@ -166,16 +166,20 @@ func serveGrpc(ctx context.Context, db *pgdb.DB, connFactory *secureconn.Factory
 
 	s := connFactory.NewServer(tracing.GlobalServerInterceptor())
 
-	cerealProjectUpdateManager, err := createProjectUpdateCerealManager(connFactory, conf.CerealConfig.Endpoint)
-	if err != nil {
-		logrus.WithError(err).Fatal("could not create cereal manager")
-	}
-	err = project_update_lib.RegisterTaskExecutors(cerealProjectUpdateManager, "compliance", ingesticESClient, authzProjectsClient)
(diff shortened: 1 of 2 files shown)


components/compliance-service/compliance.go
```diff
@@ -195,11 +199,15 @@ func serveGrpc(ctx context.Context, db *pgdb.DB, connFactory *secureconn.Factory
 	version.RegisterVersionServiceServer(s, versionserver.New())
 	status.RegisterComplianceStatusServer(s, statusSrv)
 
-	purgeServer, err := setupDataLifecyclePurgeInterface(ctx, connFactory, conf, cerealManager)
-	if err != nil {
-		logrus.Fatalf("serveGrpc aborting, can't setup purge server: %s", err)
+	if os.Getenv("RUN_MODE") == "test" {
+		logrus.Warn(`Skipping data-lifecycle setup due to RUN_MODE env var being set to "test"`)
+	} else {
+		purgeServer, err := setupDataLifecyclePurgeInterface(ctx, connFactory, conf, cerealManager)
```
ryancragun: I thought we already handled it for data lifecycle in the setup function https://github.com/chef/automate/blob/5354eae026acae85e257cb9e18a70834bf57d67a/components/compliance-service/compliance.go#L443

stevendanna: I don't think that will stop the manager this creates from polling for jobs every few seconds and creating an error trying to talk to cereal-service

ryancragun: Cool cool, I think we should probably combine those. It doesn't make sense to check for test in a function that won't be called.

PR #1980, opened 2019-10-24
No magic numbers :angryface:

tylercloke:


e2e/cypress/integration/api/iam/01_projects_api.spec.ts
```diff
@@ -103,6 +103,7 @@ describeIfIAMV2p1('projects API', () => {
     const noRulesStr = 'NO_RULES';
     const editsPendingStr = 'EDITS_PENDING';
     const rulesAppliedStr = 'RULES_APPLIED';
+    const numberOfNodesAdded = 4;
```
bcmdarroch: maybe a comment that it must correspond to the number of fixtures loaded L151-155?

PR #2002, opened 2019-10-25
[chef-server] update to latest chef server

ryancragun:
Upgrade to the latest version of the chef server


ryancragun:
The erlang TLS/SSL `public_key` module has been modified significantly between OTP 17 and 22. At the time of writing, the latest Chef Infra Server 13.1.13 uses OTP 20, whose hostname check verifier isn't compatible with the internal Chef Automate CA. In OTP 20, if the certificate has SAN keys, the `public_key` module will [verify the DNS and URI fields](http://erlang.org/doc/man/public_key.html#pkix_verify_hostname-3) but the Chef Automate CA uses DNS and IP. In OTP 21/22, you can provide the [`customize_hostname_check` SSL option](http://erlang.org/doc/man/ssl.html#type-customize_hostname_check) which will use the configured fun for verification. We'll need to write our own verifier implementation that verifies the DNS and IP subject alt name keys and configure erchef/bifrost to use it in the erlang config. Updating to OTP/22 will be [merged for Chef Infra Server 13.3 window](https://github.com/chef/chef-server/pull/1931). We could probably begin writing that custom matcher off of the OTP/22 Chef Infra Server branch. There may still be issues with the pedant tests that initially caused us to pin the Chef Infra Server build. Those will need to be sorted out.

stevendanna:
It looks to me like this may have been fixed for us in OTP 20.2: 

https://github.com/erlang/otp/commit/8cee0d079821288d3aa9b6a039b11497d315d390

Chef Server 13.2 should be on OTP 20.2: https://bldr.habitat.sh/#/pkgs/chef/oc_erchef/13.2.0/20200414184413

So, it is possible all we need to do now is update all the pins and run the pins again.

stevendanna:
Well, it looks like in 20.2 we just hit a different hostname verification issue.

Issue #1863, opened 2019-10-15
ensure only ingested nodes get project updates


vjeffrey:
## User Story

As part of https://github.com/chef/automate/pull/1772/files, we started updating nodes stored in postgres with the appropriate projects info via the project update manager.

We need to ensure this process only updates nodes that were ingested, not manually added nodes or nodes added via nodemanager. 

This can be done by adding a small where condition to the postgres query, b/c all ingested nodes will have a value of manager == ''
https://github.com/chef/automate/blob/master/components/nodemanager-service/pgdb/project_update.go#L21

## Definition of Done
Updates to projects on nodes from the project update manager only update ingested nodes (nodes with no manager)

PR #2008, opened 2019-10-26
ensure only nodes with no manager (ingested nodes) are updated by project mgr

vjeffrey:
this is a bit of an "overly cautious" change.
-whenever a node report is ingested, we send a msg to nodemanager about it
-the project update manager ensures that when project updates are made, we update those nodemanager node references to be accurate
-this change ensures we're only grabbing non-managed nodes (ingested nodes). non-ingested nodes shouldn't ever have any "projects_data", so this is probably overly cautious, but it doesn't hurt, and the where clause should make the loop of nodes we go through smaller, which is good!

https://github.com/chef/automate/issues/1863

### :+1: Definition of Done
making extra sure only ingested nodes get project updates

N/A -- you could go through building it all and doing all the steps, but it's kinda overkill in this case.

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/nodemanager-service/pgdb/project_update.go
+++ b/components/nodemanager-service/pgdb/project_update.go
@@ -18,7 +18,8 @@ const selectNodesProjectData = `
 SELECT
   n.id,
   n.projects_data
-FROM nodes n
+FROM nodes n 
+WHERE n.manager == '';
 `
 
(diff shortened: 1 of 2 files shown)


vjeffrey:
rebasing on master

PR #2009, opened 2019-10-26
make scheduler scanner job less flaky

vjeffrey:
Just trying to address flakiness we've recently seen. pete awesomely modified the file so we could see what was failing, and this seems to be one of the common failures

N/A

### :+1: Definition of Done
scanner tests less flaky

### DIFF
--- a/components/compliance-service/api/tests/50_scheduler_spec.rb
+++ b/components/compliance-service/api/tests/50_scheduler_spec.rb
@@ -158,6 +158,7 @@ describe File.basename(__FILE__) do
   it "creates new jobs as children of existing ones" do
     counter = 0
     while counter <= 10
+      puts "sleeping 10sec, counter is: ", counter
       sleep 10
       all_jobs = GRPC jobs, :list, Jobs::Query.new()
       if all_jobs.total >= 7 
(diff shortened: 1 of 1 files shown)


components/compliance-service/api/tests/50_scheduler_spec.rb
```diff
@@ -156,8 +156,15 @@ def cleanup
   end
 
   it "creates new jobs as children of existing ones" do
-    sleep 70
-    all_jobs = GRPC jobs, :list, Jobs::Query.new()
+    counter = 0
+    while counter <= 10
+      sleep 10
```
stevendanna: I'm wondering if it makes sense to print here.  I'm worried we are hiding a bit of a bug/inefficiency so it might be good to keep an eye on how long this takes.

vjeffrey: 👍  good point, i can add that!

components/compliance-service/api/tests/50_scheduler_spec.rb
```diff
@@ -156,8 +156,16 @@ def cleanup
   end
 
   it "creates new jobs as children of existing ones" do
-    sleep 70
-    all_jobs = GRPC jobs, :list, Jobs::Query.new()
+    counter = 0
+    while counter <= 10
```
phiggins: nitpick: you could avoid doing the manual counter iteration by using `Integer#times`: https://ruby-doc.org/core-2.5.0/Integer.html#method-i-times

vjeffrey: oh wow, i did not know that existed! im gonna give it a try, thanks!

PR #2017, opened 2019-10-28
Fix incorrect border around compliance search

scottopherson:
This commit fixes issue where the border around compliance searchbar would render incorrectly in Firefox.

Before
Screen Shot 2019-10-28 at 1 33 11 PM

After
Screen Shot 2019-10-28 at 1 32 42 PM


vjeffrey:
gonna rebase on master to try to get past CI

PR #2021, opened 2019-10-28
Make admin e2e tests less flakey

jaym:
These tests were not waiting for the conditions they expected. Since we
disable the automatic version of that, we need to do it manually.


components/automate-ui/e2e/admin.e2e-spec.ts
```diff
@@ -1,8 +1,9 @@
-import { browser, $, $$, ExpectedConditions as EC, ElementFinder } from 'protractor';
+import { browser, $, $$, ExpectedConditions as EC, ElementFinder, promise } from 'protractor';
 import { fakeServer } from './helpers/fake_server';
+import { expectCountElements, waitForElement, waitForSpinners } from './helpers/browser_expectations'
 import { DatetimePipe } from '../src/app/pipes/datetime.pipe';
 
-describe('Admin pages', () => {
+fdescribe('Admin pages', () => {
```
bcmdarroch: don't forget to drop the `f`

components/automate-ui/e2e/admin.e2e-spec.ts
```diff
@@ -164,57 +165,69 @@ describe('Admin pages', () => {
...
       });
 
       it('displays the returned token info', () => {
-        const name = $('chef-table chef-tbody chef-tr:nth-child(1) chef-td:first-child');
-        expect(name.getText()).toBe('This token was generated by the chef-automate CLI tool. ' +
-          'It has admin level access on the entire Automate API.');
-
-        const id = $('chef-table chef-tbody chef-tr:nth-child(1) chef-td:nth-child(2)');
-        expect(id.getText()).toBe('admin-token');
+        //const name = browser.wait(EC.presenceOf($('chef-table chef-tbody chef-tr:nth-child(1) chef-td:first-child')));
```
bcmdarroch: is this for debugging? i'd drop it from here and use it as an example in the browser expectations file

jaym:
protractor has some magic that knows how to wait for all the things you asked angular to do are done (like make http requests, switch pages, etc)
We disable this everywhere with `browser.waitForAngularEnabled(false)` probably because we do a lot of polling and not doing that messes stuff up.
So I think that means that we have to wait for certain things to happen manually. Its always been like this, no idea what pushed it over the edge. Put you can easily reproduce by modifying `fake_server.js` to sleep:
```javascript
  server = createServer((req, res) => {
    setTimeout(() => {
      hock.handler(req, res)
    }, 1000);
  });
  server.listen(fakeServerPort, done);
};
```

PR #2022, opened 2019-10-28
[deployment] Provide an http endpoint for monitoring Automate and its services

### LINKED ISSUE #1883: Expose service status via HTTP
## User Story

As an Automate user, I want to be able to quickly check the health of Automate without signing into the node and running `chef-automate status`

As an Automate admin, I want to be able to easily monitor the health of Automate (and component services) without signing into the node and running `chef-automate status`

## Definition of Done

This work is considered Done when application health is exposed via a status endpoint similar to Chef Infra Server's `/_status`

## Demo Script / Repro Steps

n/a

yzl:
Expose an http endpoint that provides service status.

Fixes #1883.

- [ ] Tests added/updated?
- [x] Docs added/updated?


components/automate-gateway/api/deployment/deployment.proto
```diff
@@ -27,6 +27,14 @@ service Deployment {
     option (chef.automate.api.iam.policy).resource = "system:service:version";
     option (chef.automate.api.iam.policy).action = "system:serviceVersion:list";
   };
+
+   rpc GetDeploymentStatus (DeploymentStatusRequest) returns (DeploymentStatusResponse) {
+    option (google.api.http).get = "/deployment/status";
```
yzl: Do we want to make this `/status`?

components/automate-gateway/api/deployment/deployment.proto
```diff
@@ -49,3 +57,15 @@ message ServiceVersion {
 	string version = 3;
 	string release = 4;
 }
+
+message DeploymentStatusRequest {}
+
+message DeploymentStatusResponse {
```
yzl: There is a legacy `StatusResponse` and I wasn't sure whether I could repurpose `StatusResponse`, hence `DeploymentStatusResponse`.

components/automate-chef-io/content/docs/monitoring.md
```diff
@@ -0,0 +1,56 @@
++++
+title = "Monitoring Chef Automate"
```
yzl: I didn't add a corresponding entry to the `config.toml` in `components/automate-chef-io`; let me know if I should.

components/automate-gateway/handler/deployment.go
```diff
@@ -48,3 +48,34 @@ func (s *DeploymentServer) ServiceVersions(ctx context.Context, _ *deployment.Se
...
+
+func (s *DeploymentServer) GetDeploymentStatus(ctx context.Context, _ *deployment.DeploymentStatusRequest) (*deployment.DeploymentStatusResponse, error) {
+	status, err := s.deploymentClient.Status(ctx, &api.StatusRequest{})
+	if err != nil {
+		logrus.WithError(err).Error("failed to get deployment status")
+		return nil, err
+	}
+
+	serviceStatus := make([]*deployment.ServiceStatus, len(status.ServiceStatus.Services))
+	oks := 0
```
stevendanna: If you wanted, rather than tracking the count you could track the bool directly. Perhaps something like:

```golang
overallOk = overallOK && (svc.State.String() == "OK")
```

but it also looks correct as-is.

components/automate-gateway/handler/deployment.go
```diff
@@ -48,3 +48,34 @@ func (s *DeploymentServer) ServiceVersions(ctx context.Context, _ *deployment.Se
...
+
+	for index, svc := range status.ServiceStatus.Services {
+		if svc.State.String() == "OK" {
```
stevendanna: I'm wondering if it might be cleaner to compare this to `deployment.ServiceState_OK` rather than the string compare?

components/automate-gateway/api/deployment/deployment.proto
stevendanna: Mind pointing me to where the legacy version is?

components/automate-gateway/handler/deployment.go
```diff
@@ -48,3 +48,34 @@ func (s *DeploymentServer) ServiceVersions(ctx context.Context, _ *deployment.Se
 
 	return &deployment.ServiceVersionsResponse{Services: gatewayServices}, nil
 }
+
+func (s *DeploymentServer) GetDeploymentStatus(ctx context.Context, _ *deployment.DeploymentStatusRequest) (*deployment.DeploymentStatusResponse, error) {
+	status, err := s.deploymentClient.Status(ctx, &api.StatusRequest{})
+	if err != nil {
+		logrus.WithError(err).Error("failed to get deployment status")
+		return nil, err
```
stevendanna: We might consider using `status.Errorf` here so we can return an appropriate status code.

components/automate-gateway/handler/deployment.go
```diff
@@ -48,3 +48,34 @@ func (s *DeploymentServer) ServiceVersions(ctx context.Context, _ *deployment.Se
...
+	}
+
+	var overallState bool
+	if oks == len(status.ServiceStatus.Services) {
+		overallState = true
+	} else {
+		overallState = false
+	}
+
+	return &deployment.DeploymentStatusResponse{Ok: overallState, Services: serviceStatus}, nil
```
stevendanna: 💭 For our HTTP APIs in the past we've returned a non-200 response from the status endpoint if it is not healthy.  HTTP allowed us to still return the JSON body along with the non-200 response, but I'm not sure if the gateway gRPC->HTTP translation stuff supports that.  Might be worth looking at whether we can return a non-200 response in the case of error and still get a response body.

components/automate-gateway/api/deployment/deployment.proto
yzl: @stevendanna It's https://github.com/chef/automate/blob/master/components/automate-gateway/api/legacy/legacy.proto

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -11,7 +11,7 @@ toc = true
 +++
 
 This API reference details Chef Automate IAM v2 features from the command line.
-If you are not already on IAM v2, [upgrade to IAM v2]({{< relref "iam-v2-guide.md#upgrade-to-iam-v2" >}}) first to ensure expected outcomes when using `curl` commands.
+If you are not already on IAM v2, [upgrade to IAM v2]({{< relref "#upgrade-to-iam-v2" >}}) first to ensure expected outcomes when using `curl` commands.
```
yzl: I fixed the relrefs so that I could build the docs but left the commit separate so that it can be pulled out easily if someone needs the change sooner, or has already fixed it in their branch.

components/automate-chef-io/content/docs/iam-v2-guide.md
```diff
@@ -139,7 +139,7 @@ The statement allows us to specify the `actions` a user is permitted to take upo
 The `projects` field on a statement is an array that may contain more than one existing project, a wildcard `*` to indicate permission to resources in _any project_, or `(unassigned)` to indicate permission to resources that have not been assigned to a project.
 
 Note that the `projects` property in statements designates permission for the resources within the statement (here, that is `iam:users` and `iam:teams`), _not_ for the policy itself, and _cannot_ be left empty.
-For more about projects, please see [Projects]({{< relref "iam-v2-guide.md#projects" >}}) or [Projects Property]({{< relref "iam-v2-api-reference/#projects-property" >}}) documentation.
+For more about projects, please see [Projects]({{< relref "iam-v2-guide.md#projects" >}}) or [Projects Property]({{< relref "iam-v2-api-reference.md#projects-property" >}}) documentation.
```
yzl: I fixed the relrefs so that I could build the docs.

components/automate-gateway/gateway/services.go
```diff
@@ -673,6 +684,84 @@ func (s *Server) configMgmtReportExportHandler(w http.ResponseWriter, r *http.Re
...
+	if overallOks {
+		w.WriteHeader(http.StatusOK)
+	} else {
+		w.WriteHeader(http.StatusInternalServerError)
+	}
+
+	// Provide prettyprinted output.
+	var jsonBody []byte
+	if _, ok := r.URL.Query()["pretty"]; ok {
+		w.Header().Set("Content-Type", "application/json+pretty")
```
yzl: Should the content-type be "application/json"? I wasn't sure. ("application/json+pretty" is the accept header for negotiating with grpc-gateway itself.)

stevendanna: My guess is we are find using application/json for both.

components/automate-chef-io/content/docs/monitoring.md
mjingle: You don't necessarily need a corresponding entry. I did, however, move this entry to be under the Reference section.

Issue #1494, opened 2019-09-06
when a compliance or client run node's role, chef tag, or any array field for projects is null the project update fails


bcmdarroch:
## User Story

As an Automate User, I will use this work when I want to apply projects to ingested resources.

During a demo, we ran into the following bug where Elasticsearch got into a bad state when a project update attempted to update a scan job's projects. The projects field does not yet exist on scan jobs. After hitting this error once, project update no longer works for any ingested resource.

## Demo Script / Repro Steps
1. Upgrade to beta2.1 and load data
```
chef-automate iam upgrade-to-v2 --beta2.1
chef_load_nodes 20
chef_load_compliance_scans -N10 -M100 -T10 -D10
chef_load_actions 20
```
2. Confirm project update works when there are no scan jobs 

- log in to the UI as `admin`
- go to infrastructure > client runs
- identify some environments in the table view
- create a new project
- add a node ingest rule for the environment you identified
- run `sl | grep elastic` in your studio
- update projects
- you'll see the following elasticsearch logs in the server logs:
```
[2019-09-06T19:47:58+00:00] automate-elasticsearch.default(O): [2019-09-06T19:47:58,849][WARN ][o.e.m.j.JvmGcMonitorService] [fWalQHA] [gc][2239] overhead, spent [1.1s] collecting in the last [1.7s]
[2019-09-06T19:48:00+00:00] automate-elasticsearch.default(O): [2019-09-06T19:48:00,099][INFO ][o.e.t.LoggingTaskListener] 74999 finished with response BulkByScrollResponse[took=3.5s,timed_out=false,sliceId=null,updated=1030,created=0,deleted=0,batches=2,versionConflicts=0,noops=0,retries=0,throttledUntil=0s,bulk_failures=[],search_failures=[]]
[2019-09-06T19:48:24+00:00] automate-elasticsearch.default(O): [2019-09-06T19:48:24,622][INFO ][o.e.t.LoggingTaskListener] 74977 finished with response BulkByScrollResponse[took=28s,timed_out=false,sliceId=null,updated=1030,created=0,deleted=0,batches=2,versionConflicts=0,noops=0,retries=0,throttledUntil=0s,bulk_failures=[],search_failures=[]]
```
- go back to client runs
- filter your projects view and you should see only nodes for that project you created

3. Add a scan job
```
send_inspec_example
```

4. During the project update operation, elasticsearch fails to apply a project to the scan job
- create a new project called `tiny`
- add a Node ingest rule with the condition 
[...]

PR #2024, opened 2019-10-28
Project update for compliance and client runs elasticsearch bug

lancewf:
When running a project update with a chef tag project rule the compliance-service would fail to update. The problem is that the array field is sometimes stored as "null" and not an empty array. The solution is to check if the field is null before using it. 

Client runs had the same problem which is fixed in the PR also. 

#1494

### :+1: Definition of Done
Able to successfully run a project update with a project with chef tag ingest rules when there are compliance reports with empty chef tags. 

1. `build components/compliance-service && start_all_services`
1. Add some compliance nodes with `chef_load_compliance_nodes 100`
1. Open the compliance page https://a2-dev.test/compliance/reports/nodes
1. Copy one of the node's environment, its displayed in the list view
    - or use the environment filter at the top of the page to identify what a node's enviornment is
1. Create a project with two rules. One matching on a chef tag with any value. The second rule matching the environment found above. 
1. Update the projects
1. Go back to the https://a2-dev.test/compliance/reports/nodes page and filter by the project created to ensure the project update worked correctly. 

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/compliance-service/ingest/ingestic/ingestic.go
+++ b/components/compliance-service/ingest/ingestic/ingestic.go
@@ -500,6 +500,12 @@ func (backend *ESClient) JobStatus(ctx context.Context, jobID string) (project_u
 
 	percentageComplete := getPercentageComplete(tasksGetTaskResponse.Task.Status)
 
+	// If the task is marked complete but the percentage complete is not 1 then the task stopped unexpectedly
+	if tasksGetTaskResponse.Completed && percentageComplete != 1 {
+		return project_update_lib.JobStatus{}, fmt.Errorf("Task %s stopped unexpectedly. "+
+			"For more information go to http://localhost:10141/.tasks/task/%s", jobID, jobID)
+	}
+
 	if percentageComplete != 0 {
 		runningTimeNanos := float64(tasksGetTaskResponse.Task.RunningTimeInNanos)
 		timeLeftSec := int64(runningTimeNanos / percentageComplete / 1000000000.0)
--- a/components/compliance-service/ingest/ingestic/ingestic.go
+++ b/components/compliance-service/ingest/ingestic/ingestic.go
@@ -514,7 +514,15 @@ func (backend *ESClient) JobStatus(ctx context.Context, jobID string) (project_u
 
 	var estimatedEndTimeInSec int64
 
-	percentageComplete := getPercentageComplete(tasksGetTaskResponse.Task.Status)
+	percentageComplete, ok := getPercentageComplete(tasksGetTaskResponse.Task.Status)
+
+	if !ok {
+		return project_update_lib.JobStatus{
+			Completed:             tasksGetTaskResponse.Completed,
+			PercentageComplete:    0,
+			EstimatedEndTimeInSec: estimatedEndTimeInSec,
+		}, nil
+	}
 
 	// If the task is marked complete but the percentage complete is not 1 then the task stopped unexpectedly
 	if tasksGetTaskResponse.Completed && percentageComplete != 1 {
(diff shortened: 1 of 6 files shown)


components/compliance-service/ingest/ingestic/ingestic.go
```diff
@@ -284,6 +284,10 @@ func (backend *ESClient) UpdateReportProjectsTags(ctx context.Context, projectTa
 						}
 					}
 					if (condition.roles.length > 0) {
+						if (ctx._source['roles'] == null) {
```
lancewf: The condition does not match if its type is roles and there are no roles on the report.

components/compliance-service/ingest/ingestic/ingestic.go
```diff
@@ -300,6 +304,10 @@ func (backend *ESClient) UpdateReportProjectsTags(ctx context.Context, projectTa
 						}
 					}
 					if (condition.chefTags.length > 0) {
+						if (ctx._source['chef_tags'] == null) {
+							match = false;
+							break;
+						}
 						def found = false;
 						for (def ruleChefTag : condition.chefTags) {
 							for (def ccrChefTag : ctx._source.chef_tags) {
```
lancewf: This where the error was thrown. When the `ctx._source.chef_tags` == null

components/compliance-service/integration_test/reports_projects_update_test.go
```diff
@@ -410,6 +410,30 @@ func TestProjectUpdate(t *testing.T) {
 			},
 			projectIDs: []string{},
 		},
+		{
```
lancewf: These two tests failed with the same elasticsearch error from the issue before the above changes.

components/compliance-service/ingest/ingestic/ingestic.go
```diff
@@ -416,6 +424,10 @@ func (backend *ESClient) UpdateSummaryProjectsTags(ctx context.Context, projectT
 						}
 					}
 					if (condition.roles.length > 0) {
```
msorens: Since each attribute has an identical code block pattern--and indeed identical code across `UpdateReportProjectsTags`, `UpdateSummaryProjectsTags`, and `UpdateNodeProjectTags`--you could reduce a lot of duplication by introducing two simple code generation functions, so that `script` in this function could be reduced to this, making it easier to maintain:
```
script := `
		ArrayList matchingProjects = new ArrayList();
		for (def project : params.projects) {
			for (def rule : project.rules) {
				def match = rule.conditions.length != 0;
				for (def condition : rule.conditions) {` +
					genSingleLoop("chefServer", "source_fqdn") + 
					genSingleLoop("environment", "environment") + 
					genSingleLoop("organization", "organization_name") + 
					genSingleLoop("policyName", "policy_name") + 
					genSingleLoop("policyGroup", "policy_group") + 
					genDoubleLoop("roles", "roles", "Role") + 
					genDoubleLoop("chefTags", "chef_tags", "ChefTag") + `
				}
				if ( match ) {
					matchingProjects.add(project.name);
					break;
				}
			}
		}
		ctx._source.projects = matchingProjects.toArray();
`
```

components/compliance-service/ingest/ingestic/ingestic.go
```diff
@@ -500,6 +516,12 @@ func (backend *ESClient) JobStatus(ctx context.Context, jobID string) (project_u
 
 	percentageComplete := getPercentageComplete(tasksGetTaskResponse.Task.Status)
 
+	// If the task is marked complete but the percentage complete is not 1 then the task stopped unexpectedly
+	if tasksGetTaskResponse.Completed && percentageComplete != 1 {
+		return project_update_lib.JobStatus{}, fmt.Errorf("Task %s stopped unexpectedly. "+
+			"For more information go to http://localhost:10141/.tasks/task/%s", jobID, jobID)
```
tylercloke: This seems like it might not be the best info in some multi-node world but it's fine for now.

components/compliance-service/ingest/ingestic/ingestic.go
lancewf: How do you create a function in the painless script?

components/compliance-service/ingest/ingestic/ingestic.go
```diff
@@ -513,26 +546,37 @@ func (backend *ESClient) JobStatus(ctx context.Context, jobID string) (project_u
 	}, nil
 }
 
-func getPercentageComplete(status interface{}) float64 {
+func getPercentageComplete(status interface{}) (float64, bool) {
```
lancewf: This function is used for other types of async jobs. Therefore we need to add the create and deletes to the completed amounts.

lancewf: There is also the case that the "created", "deleted", and "updated" are missing from the status. In that case, we return a false to signal that the percentage should not be used, but it is not an error.

components/compliance-service/integration_test/reports_projects_update_test.go
bcmdarroch: Not sure if this should be a refactor or an additional test but it'd be great to see a test struct with all possible combinations of `attribute`, `operator`, and `value` (array or single value), to make sure we're not missing any edge cases. We could use two fully-populated `&relaxting.ESInSpecSummary` structs, one expected to match and one not to match, and iterate over the condition combinations on each test run.

lancewf: I would like to cover all the cases. I don't fully understand what you are proposing, but it sounds interesting. 
The reason you don't see the `operator` in these tests is because it is completely ignored in the code.

PR #2025, opened 2019-10-29
Cereal workflow and tasks for data-feed-service

satellite15:

Removed the basic integration with the cereal service and adopted a new workflow to manage multiple tasks associated with the data-feed. The aggregation of data is now managed completely by this workflow and the associated tasks.

The workflow has 5 associated tasks. OnStart the work flow has 2 tasks to determine which nodes to gather data for - DataFeedPollTask & DataFeedListReportsTask. These collate a map of node ipaddress to node ID(to gather client run and attrribute data), report ID (to gather compliance report data). Once these are complete the workflow then batches the ID's before processing each batch with the remaining tasks DataFeedCLientTask, DataFeedComplianceTask & DataFeedNotifierTask which collate the data and send to the registered endpoints (destinations). Each batch is processed sequentially and the default batch size is 100 nodes and is configurable. Batching manages the memory requirements of the data-feed-service

#### DataFeedPollTask
Checks for nodes that have had a client run in the polled interval
Stores the Node ID's in its the workflow payload in a map with the ipaddress as key to a node ID, report ID tuple

#### DataFeedListReports tasks
Checks for compliance reports in the polled interval
Stores the report ID's in its the workflow payload in a map with the ipaddress as key to a node ID, report ID tuple

#### DataFeedClientTask
Started when the poll task completes, takes the nodes from poll task
For each node builds the ohai and client run data for the node and it's most recent compliance report. removes the ipaddress of the node from the stored map.
Builds array of datafeedMessages to return in it's result

#### DataFeedComplianceTask
Takes the stored map as params. Any entries left in the mpa have not had a client run but have had a report.This taks gets the compliance report and also the most recent ohai and client data The results contain the report set of datafeed messages

#### DataFeedNotifierTask
takes the batched datafeed message set and posts it to the destinations. 
If there are further batches to be processed the DatafeedClient task is scheduled to process the next batch. When all batches are complete the workflow completes.

Configure the poll interval in the config.toml for the data-feed-service to a short period 3m or 5m
Add a destination (service now or requestbin) in the Fata Feed UI page.
Perform client runs and compliance scans on nodes managed by automate.
Data should be periodically posted to the configured destination

- [x] Tests added/updated? No new automated tests, tested using the ServiceNow dev/test environment
- [x] Docs added/updated? Code comments only at this stage


jaym:
This doesn't look like a new problem, but I think this requires us to have all the data (the reports and node attributes?) in memory. Is this correct? Any idea how large memory usage will get for the 3-5 min interval?

jaym:
Some numbers of gathered:
Our HA cluster setup claims it can handle 40k CCRs / hour with an average node size of 3MB

satellite15:
> This doesn't look like a new problem, but I think this requires us to have all the data (the reports and node attributes?) in memory. Is this correct? Any idea how large memory usage will get for the 3-5 min interval?

Currently yes, "This is the first stage of preparing the service to run within the HA framework."

I wanted to get some HA feedback before I pushed ahead with the next stage which will be to break this down and get batches of data, hence the initial tasks that map the ipaddress to client run and compliance report id report. The next stage will be to involve splitting that map up to manage memory better.

satellite15:
> Some numbers of gathered:
> Our HA cluster setup claims it can handle 40k CCRs / hour with an average node size of 3MB

I have now committed the batch processing which processes 100 nodes at a time. Data is gathered for 100 (default 100, it is configurable) nodes at a time and posted to the registered endpoints. For example if 530 nodes have had a CCR or compliance report in the interval (default 30 minutes) the poll and list reports tasks run once and then the sequence of Client, Compliance & notifier tasks will run 6 times processing up to 100 nodes per batch. The configurable batch limit reduces the memory requirements.

satellite15:
Closing and creating a new one with the latest code

PR #2027, opened 2019-10-29
Upload nodemanager logs

jaym:
Trying to track down why compliance tests keep dying


.expeditor/verify.pipeline.yml
```diff
@@ -601,6 +602,7 @@ steps:
       - /tmp/event-service.log
       - /tmp/compliance.log
       - /tmp/nodemanager.log
```
stevendanna: I take it this one doesn't actually exist?

jaym: it does not

PR #2033, opened 2019-10-29
Jamie/chef input touched

SEAjamieD:
The `chef-input` component does not properly register ng-untouched/ng-touched classes.   Rather than build in a workaround, we have opted to use plain html `input` until a new `chef-input` is used across the site.  This also required an update to the input styling in the respective scss file.

https://github.com/chef/ui-team/issues/7 created to update chef-input across Automate

### :+1: Definition of Done
Inputs on the Create Rule page, show errors after they have been touched.  Tabbing through is one way of touching inputs.

2. navigate to https://a2-dev.test/settings/projects and click on any one of your projects or create a new project if you do not yet have any projects.
3. click on `Create Rule`
4. focus in on the typeable inputs but don't type anything.  When you exit the inputs, you should see errors which will then disappear once they've been typed in.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

Tabbingthroughrulecreate

### DIFF
--- a/components/automate-ui/src/app/pages/project/rules/project-rules.component.html
+++ b/components/automate-ui/src/app/pages/project/rules/project-rules.component.html
@@ -14,10 +14,10 @@
         <chef-form-field id="create-name">
           <label>
             <span class="label">Rule Name <span aria-hidden="true">*</span></span>
-            <chef-input ngDefaultControl formControlName="name" (keyup)="handleNameInput($event)"></chef-input>
+            <chef-input ngDefaultControl #test formControlName="name" (keyup)="handleNameInput($event)"></chef-input>
           </label>
           <chef-error
-            *ngIf="(ruleForm.get('name').hasError('required') || ruleForm.get('name').hasError('pattern')) && ruleForm.get('name').dirty"
+            *ngIf="(ruleForm.get('name').hasError('required') || ruleForm.get('name').hasError('pattern')) && test.classList.contains('ng-touched')"
           >Rule Name is required.</chef-error>
         </chef-form-field>
 
--- a/components/automate-ui/src/app/pages/project/rules/project-rules.component.html
+++ b/components/automate-ui/src/app/pages/project/rules/project-rules.component.html
@@ -14,7 +14,7 @@
         <chef-form-field id="create-name">
           <label>
             <span class="label">Rule Name <span aria-hidden="true">*</span></span>
-            <input formControlName="name" (keyup)="handleNameInput($event)"/>
+            <input chefInput formControlName="name" (keyup)="handleNameInput($event)"/>
           </label>
           <chef-error
             *ngIf="(ruleForm.get('name').hasError('required') || ruleForm.get('name').hasError('pattern')) && ruleForm.get('name').touched"
@@ -25,7 +25,7 @@
           <chef-form-field id="create-id">
             <label>
               <span class="label">Rule ID <span aria-hidden="true">*</span> </span>
-              <input formControlName="id" (keyup)="handleIDInput($event)" />
+              <input chefInput formControlName="id" (keyup)="handleIDInput($event)" />
             </label>
             <chef-error *ngIf="ruleForm.get('id').hasError('maxlength') && ruleForm.get('id').dirty">
               Rule ID must be 64 characters or less.
@@ -113,6 +113,7 @@
                 <label>
                   <span class="label">Value</span>
                   <input
+                    chefInput
                     formControlName="values"
                     [value]="condition.controls.values.value"
                     data-cy="rule-value"
--- a/components/automate-ui/src/app/pages/project/rules/project-rules.component.html
+++ b/components/automate-ui/src/app/pages/project/rules/project-rules.component.html
@@ -112,11 +112,8 @@
               <chef-form-field class="attribute-field">
                 <label>
                   <span class="label">Value</span>
-                  <input
-                  chefInput
-                  formControlName="values"
-                  [value]="condition.controls.values.value"
-                  data-cy="rule-value"
+                  <input chefInput formControlName="values" 
+                        [value]="condition.controls.values.value" data-cy="rule-value"
                   />
                 </label>
                 <div>
--- a/components/automate-ui/src/app/pages/project/rules/project-rules.component.html
+++ b/components/automate-ui/src/app/pages/project/rules/project-rules.component.html
@@ -121,7 +121,8 @@
                 </label>
                 <div>
                   <chef-error
-                      *ngIf="(ruleForm.get('conditions').controls[i].get('values').hasError('required') || ruleForm.get('conditions').controls[i].get('values').hasError('pattern')) && ruleForm.get('conditions').controls[i].get('values').touched"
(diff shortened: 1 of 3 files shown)


msorens:
1. If you just tab through the field, the error pop-up appears, yes, so you have addressed the key issue.
But I noticed something else that was working has now stopped working and I am not sure why: type "a" then backspace. Empty field should show "value required" and it does not, so the `required` check is not kicking in. Yet if you now type a space it shows the `pattern` check kicks in properly. This same behavior occurs in other places, too, e.g. Create Team, so it is not caused by changes here.

2. I thought the intent was to create a new, Angular-based `<chef-input>` rather than just use a plain `<input>`. Has that changed? However, if we do want to go with the latter, should always be `<input chefInput...>`, i.e., using the `chef-input` directive, for consistent styling.

susanev:
wait. @msorens are you suggesting we present the user an error while they are editing in the input? if so, id prefer not to do that and instead only present the error once they leave the input. it can be a very frustrating experience to be editing and have it send you errors before you have a chance to finish.

vjeffrey:
> I thought the intent was to create a new, Angular-based <chef-input> rather than just use a plain <input>. Has that changed

@msorens yes -- the decision made in the mtg about this was that @SEAjamieD is going to use a regular input for this task, then he will make a card to replace the stencil component with an angular one, so that we don’t use the broken stencil one anymore

msorens:
@susanev I am not suggesting making any changes to the way those controls have always worked. I am just suggesting to fix something that has changed.

SEAjamieD:
> 1. If you just tab through the field, the error pop-up appears, yes, so you have addressed the key issue.
>    But I noticed something else that was working has now stopped working and I am not sure why: type "a" then backspace. Empty field should show "value required" and it does not, so the `required` check is not kicking in. Yet if you now type a space it shows the `pattern` check kicks in properly. This same behavior occurs in other places, too, e.g. Create Team, so it is not caused by changes here.

Thanks @msorens - now that we have access to ng-touched AND ng-dirty here, the errors needed a little reworking.  You'll see in this updated push that now we get an error when we tab away from the input box.  You'll also see the same error while `dirty` if the pattern is not allowed.. i.e. a space.

SEAjamieD:
Uncovered an extra issue styling issue barely outside scope that should be handled while at it, please don't merge or review again just yet, I'll request re-review when all complete so as not to waste anyones time - thanks!

components/automate-ui/src/app/pages/project/rules/project-rules.component.scss
```diff
@@ -16,9 +16,16 @@ chef-page {
...
+  input {
     height: 45px;
     width: 100%;
+    outline: 0;
+  }
+
+  chef-error {
+    background-color: #DC267F;
+    color: #FFFFFF;
+    border-radius: 0 0 4px 4px;
```
SEAjamieD: `:ng-deep` from the chefInput Directive/styling isn't reaching deep enough when the error is sectioned inside an extra div, so I ported the style over here.  This extra div is needed to guarantee that the error will show above the helper text.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -112,15 +112,19 @@
...
                 </label>
-                <chef-error
-                  *ngIf="(ruleForm.get('conditions').controls[i].get('values').hasError('required') || ruleForm.get('conditions').controls[i].get('values').hasError('pattern')) && ruleForm.get('conditions').controls[i].get('values').dirty"
-                >Value is required.</chef-error>
+                <div>
+                  <chef-error
+                      *ngIf="( ruleForm.get('conditions').controls[i].get('values').hasError('required') && ruleForm.get('conditions').controls[i].get('values').touched )
+                              || ( ruleForm.get('conditions').controls[i].get('values').hasError('pattern') && ruleForm.get('conditions').controls[i].get('values').dirty )"
+                    >Value is required.</chef-error>
+                </div>
```
SEAjamieD: I changes these so that we're checking two different setups now
1.  If the input doesn't have a value AND it has been touched - we need to show the error
OR
2.  If the input has a value (is dirty) AND that value does not meet our pattern criteria, we are showing an error

SEAjamieD:
@msorens / @chef/ui-team - This is ready for review again, thanks!

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -112,15 +112,19 @@
...
-                    formControlName="values"
-                    [value]="condition.controls.values.value"
-                    data-cy="rule-value"
-                  ></chef-input>
+                  <input
+                  chefInput
+                  formControlName="values"
+                  [value]="condition.controls.values.value"
+                  data-cy="rule-value"
+                  />
```
msorens: Consider less line breaks in this element; should fit fine on just two lines.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -112,15 +112,19 @@
...
                 </label>
-                <chef-error
-                  *ngIf="(ruleForm.get('conditions').controls[i].get('values').hasError('required') || ruleForm.get('conditions').controls[i].get('values').hasError('pattern')) && ruleForm.get('conditions').controls[i].get('values').dirty"
-                >Value is required.</chef-error>
+                <div>
+                  <chef-error
+                      *ngIf="( ruleForm.get('conditions').controls[i].get('values').hasError('required') && ruleForm.get('conditions').controls[i].get('values').touched )
+                              || ( ruleForm.get('conditions').controls[i].get('values').hasError('pattern') && ruleForm.get('conditions').controls[i].get('values').dirty )"
```
msorens: Wow, that's quite an expression! 😁 
Suggest moving that to the code-behind; try to keep expressions in a template very simple as a rule; either a single variable or a simple expression.
Couple reasons:
The template should define presentation; the code-behind should define execution, so one should strive to maintain separation of concerns.
Also, it makes both testing and diagnosis of issues much more accessible.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
```diff
@@ -14,18 +14,18 @@
         <chef-form-field id="create-name">
           <label>
             <span class="label">Rule Name <span aria-hidden="true">*</span></span>
-            <chef-input ngDefaultControl formControlName="name" (keyup)="handleNameInput($event)"></chef-input>
```
msorens: Nice that you noticed to remove `ngDefaultControl` since it is no longer being sent to a 3rd party control! However (and technically outside the scope, but since you are in the file), I notice there are several more `ngDefaultControl` in this file that could be removed for the same reason: those were converted from `chef-select` to plain `select`.

components/automate-ui/src/app/pages/project/rules/project-rules.component.html
SEAjamieD: I'm going to go ahead and merge this, and open up an new ticket to explore ways to shorten this, as I don't want this to be held up any longer.  I agree I need to find a way to better destructure this.  So far, i'm stuck on getting past the `controls` piece, but I imagine there is a better way to do this in a lot places across the application.  Thank you for talking this through earlier with me.

PR #1944, opened 2019-10-22
Update Timestamps to use UTC

SEAjamieD:
### :nut_and_bolt: Description: What code changed, and why?

Problem: Time across the site (not just run history drawer) is showing in local time instead of UTC as desired.

For times converted in our HTML, I've added a `.utc()` to our datetime pipe, which uses moment.js.  this will return UTC instead of local time.

For times requested where the conversion is done in component.ts files, i've added `.utc()` to the moment conversion.

### :chains: Related Resources
https://github.com/chef/automate/issues/1678
https://github.com/chef/automate/issues/1679
https://github.com/chef/automate/issues/1705
fixes #1938 

### :+1: Definition of Done
Time is being displayed in UTC where labeled as such
- liscense lockoutcomponents
- node scan results and scan history
- compliance run results and drawer

### :athletic_shoe: How to Build and Test the Change
1. build components/automate-ui-devproxy && start_all_services

License Lockout modal
1. chef-automate dev psql chef_license_control_service -- -c "UPDATE licenses SET active=NULL" to remove the current license
2. Navigate to https://a2-dev.test/
3. Fill out the trial license form and submit
4. Observe the trial expiration modal

Node Details
1. Navigate to https://a2-dev.test/infrastructure/client-runs
2. Click on any of your nodes available in the list
3. View the Failure or Success Tag Timestamp in Bold Blue or Pink Div
4. View the Run Duration Time Stamp on Node Home
5. Click View Error Log to see Timestamp in Modal
6. Click on Run History button in top right to view sidebar -> view updated timestamp.

Node Credentials
1. Navigate to https://a2-dev.test/settings/node-credentials
2. Add a Credential
3. view timestamp in Last Modified column

Node Integrations
1. Navigate to https://a2-dev.test/settings/node-integrations
2. Create an Integration (follow instructions here - GCP was easiest)
3. View timestamp in date created column

API Tokens
1. Navigate to https://a2-dev.test/settings/tokens
2. Create a token
3. View timestamp in Created Date column

### :white_check_mark: Checklist

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### :camera: Screenshots, if applicable

PR #2036, opened 2019-10-29
Remove utc from calculation s its not really necessary, not sure if t…

SEAjamieD:
Recently when updating UTC across client run, I updated the `getDuration` function inside the Client Runs Drawer as well.  While the code works fine locally, when using real data it is spitting out "a few seconds" as the result to all time differences.  Because this is duration and it doesn't actually need to be UTC for the correct answer, i've reverted the `getDuration` function back to the way it was before the UTC change.

https://github.com/chef/automate/pull/1944

### :+1: Definition of Done
Client runs history drawer shows the correct amount of time since the client has been run, not just "a few seconds ago"

navigate to https://a2-dev.test/infrastructure/client-runs
click on any specific node
Click on 'run history' button in top right

- [ ] Tests added/updated?
- [ ] Docs added/updated?

Screen Shot 2019-10-29 at 11 42 59 AM

### DIFF
--- a/components/automate-ui/src/app/page-components/run-history/run-history.component.ts
+++ b/components/automate-ui/src/app/page-components/run-history/run-history.component.ts
@@ -110,7 +110,7 @@ export class RunHistoryComponent implements OnInit, OnDestroy {
   }
 
   getDuration(start_time, end_time) {
-    return moment.duration(moment.utc(end_time).diff(moment.utc(start_time))).humanize();
+    return moment.duration(moment(end_time).diff(moment(start_time))).humanize();
   }
 
   onDownloadRunsReport() {
--- a/components/automate-ui/src/app/page-components/run-history/run-history.component.html
+++ b/components/automate-ui/src/app/page-components/run-history/run-history.component.html
@@ -46,7 +46,7 @@
               <p>
                 <strong>{{ history.startTime | datetime: RFC2822 }}</strong>
               </p>
-              <p>{{ getDuration(history.startTime, history.endTime) }} ago</p>
+              <p><span>Duration: </span>{{ getDuration(history.startTime, history.endTime) }}</p>
             </div>
             <chef-button secondary (click)="onSelect(history)">
               <chef-icon>chevron_right</chef-icon>


components/automate-ui/src/app/page-components/run-history/run-history.component.ts
```diff
@@ -110,7 +110,7 @@ export class RunHistoryComponent implements OnInit, OnDestroy {
   }
 
   getDuration(start_time, end_time) {
-    return moment.duration(moment.utc(end_time).diff(moment.utc(start_time))).humanize();
+    return moment.duration(moment(end_time).diff(moment(start_time))).humanize();
```
vjeffrey: this is interesting.
this seems to be calculating the duration of a run. but i would assume, when looking at that panel, that this was trying to tell me how long ago it happened in human terms, not the duration of run..

SEAjamieD: When i print it out to the console - pre AND post change, it prints the correct numbers as well.

vjeffrey: what makes this extra interesting to me is that...it's always been like this it seems...am i confusing myself or something?

SEAjamieD: nope - it has always been like this ( without `.utc`) - this is essentially just a revert.  Its been like this since the at least the initial open sourcing.

vjeffrey: hmm. ya it def doesn't make sense to show the duration of a run and then put "ago" at the end...i'm asking ux! :)

susanev: @SEAjamieD @vjeffrey we would like to keep it as duration for now
but can you please add a `Duration: ` label before the text of the duration (e.g., Duration: 15 seconds)?

@jonong1972 is going to follow up later with another issue exploring having all three pieces of data (duration, relative time, and absolute time)

SEAjamieD: @susanev - I should also remove `ago` text, correct?

jonong1972: @SEAjamieD Yes please. thank you!

scottopherson: Yea I'd imagine `{{ history.endTime | timeFromNow }}` would be used if we wanted the "ago" text.

PR #2040, opened 2019-10-29
Update GenerateAdminToken to inject auth_context

tylercloke:


components/automate-deployment/pkg/server/api_token.go
```diff
@@ -33,6 +35,7 @@ func generateAdminToken(ctx context.Context,
 	req *api.GenerateAdminTokenRequest, connFactory *secureconn.Factory,
 	authNAddress, authZAddress string) (*api.GenerateAdminTokenResponse, error) {
 
+	ctx = auth_context.NewOutgoingContext(auth_context.NewContext(ctx, []string{"tls:service:deployment-service:internal"}, []string{}, "", "", "v2.1"))
```
blakestier: is this better than adding this call as an exempted function to the interceptor? (also, tokens! :shakes_fist:)

blakestier: nope, i'm all the wrong. carry on.

PR #2048, opened 2019-10-30
Jamie/utc compliance calendar scan job

### LINKED ISSUE #1978: update compliance calendar and scan job schedule format and change to use utc instead of local time
## Overview
Part of the continued work to get everything in the ui using UTC and a consistent format.

- calendar component
- scan job schedule
- compliance reports calendar button

### Compliance calendar filter control
Change format to, day of month, followed by month abbreviation, followed by year, e.g., `25 Oct 2019`. Date should be UTC, not local time.

calendar

### Scan job schedule
- update both start and end controls
- add `:` between two time controls
- change datetime to be using utc instead of local time
- remove am/pm control and add `UTC` label at the end
- continue the pattern of 8px of space between text (at, :, UTC) and controls
- we can remove the stretching thats happening on those controls if we want as well

scan-job-schedule

## Reference

```
./src/app/components/calendar/calendar.component.ts:    this.onDaySelect.emit(day[1].format('YYYY-MM-DD'));
./src/app/pages/+compliance/+scanner/containers/jobs-list/jobs-list.component.ts:      const endDateString = moment(endDate).format('YYYY-MM-DD');
./src/app/pages/+compliance/shared/reporting/stats.service.ts:        {'end_time': moment(dateRange.end).format('YYYY-MM-DDTHH:mm:ssZ')},
./src/app/pages/+compliance/shared/reporting/stats.service.ts:        {'start_time': moment(dateRange.start).format('YYYY-MM-DDTHH:mm:ssZ')}
./src/app/pages/+compliance/+reporting/+reporting-overview/reporting-overview.component.ts:        queryParams['end_time'] = endDateMoment.format('YYYY-MM-DD');
./src/app/pages/+compliance/+reporting/reporting.component.ts:      queryParams['end_time'] = moment(endDate).format('YYYY-MM-DD');
./src/app/pages/+compliance/+reporting/reporting.component.ts:      const endDate = moment(foundFilter.text + 'GMT+00:00', 'YYYY-MM-DDZ');
```

SEAjamieD:
This is an ongoing effort to update Automate to use UTC across the site.  This particular branch updates the calendar component, scan job scheduling, and compliance reports calendar button.

fixes #1978 

### :+1: Definition of Done

- [x] Calendar Component uses of moment altered to use UTC
- [x]  Scan Job Schedule selectors use UTC time
- [x] Compliance reports calendar button shows UTC time and `25 Oct 2019` format

navigate to https://a2-dev.test/compliance/reports/overview --> view calendar button with new formatting in UTC time
navigate to https://a2-dev.test/compliance/scan-jobs/jobs
click `Create New Job` button --> follow steps to create new Job
when at `Add Schedule` Section, click on `Set Schedule` to see new UTC formats
Save your new job
In not already there, navigate back to navigate to https://a2-dev.test/compliance/scan-jobs/jobs
Click on button to edit your job, run through same steps, and change your schedule timing.

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/e2e/job-add.e2e-spec.ts
+++ b/components/automate-ui/e2e/job-add.e2e-spec.ts
@@ -556,7 +556,7 @@ describe('Job Add', () => {
           scheduleToggle.click();
 
           expect(scheduleToggle.isSelected()).toEqual(true);
-          expect(startInputs.count()).toEqual(6);
+          expect(startInputs.count()).toEqual(5);
         });
 
         it('provides optional end datetime input', () => {
@@ -575,7 +575,7 @@ describe('Job Add', () => {
           endFormGroupToggle.click();
 
           expect(endFormGroupToggle.isSelected()).toEqual(true);
-          expect(endInputs.count()).toEqual(6);
+          expect(endInputs.count()).toEqual(5);
         });
 
         it('provides optional repeat interval input', () => {
--- a/components/automate-ui/src/app/components/calendar/calendar.component.ts
+++ b/components/automate-ui/src/app/components/calendar/calendar.component.ts
@@ -17,8 +17,8 @@ import { concat,
 })
 export class CalendarComponent {
 
-  private _date: m.Moment = m();
-  private _selected: m.Moment = m();
+  private _date: m.Moment = m.utc();
+  private _selected: m.Moment = m.utc();
 
   private _month: string;
   private _year: number;
@@ -34,7 +34,7 @@ export class CalendarComponent {
 
   @Input()
   set date(input) {
-    const date = m.isMoment(input) ? input : m(input);
+    const date = m.isMoment(input) ? input : m.utc(input);
     this._date = date;
     this._month = m.months(date.month());
     this._year = date.year();
@@ -45,7 +45,7 @@ export class CalendarComponent {
 
   @Input()
   set selected(input) {
-    const date = m.isMoment(input) ? input : m(input);
+    const date = m.isMoment(input) ? input : m.utc(input);
     this._selected = date;
   }
   get selected() {
@@ -79,7 +79,7 @@ export class CalendarComponent {
     const vy = this.date.year();
     const sm = this.selected.month();
     const sy = this.selected.year();
-    const today = m().month() === this.date.month() ? m().date() : null;
+    const today = m.utc().month() === this.date.month() ? m.utc().date() : null;
     const selected = (vm === sm && vy === sy) ? this.selected.date() : null;
     const tag = (marker) => (d: m.Moment) => [marker, d];
     const tagActive = (d: m.Moment) => {
--- a/components/automate-ui/src/app/page-components/job-schedule-form/job-schedule-form.component.html
+++ b/components/automate-ui/src/app/page-components/job-schedule-form/job-schedule-form.component.html
@@ -38,11 +38,11 @@
         <div class="field">
           <chef-form-field formGroupName="datetime">
             <div class="datetime">
-              <select formControlName="month">
-                <option *ngFor="let n of months" [value]="n">{{ n | number:'2.0-0' }}</option>
-              </select>
               <select formControlName="date">
                 <option *ngFor="let n of dates" [value]="n">{{ n | number:'2.0-0' }}</option>
+              </select>
+              <select formControlName="month">
+                <option *ngFor="let n of months" [value]="n">{{ n }}</option>
               </select>
               <select formControlName="year">
                 <option *ngFor="let n of years" [value]="n">{{ n }}</option>
@@ -53,13 +53,14 @@
               <select formControlName="hour">
                 <option *ngFor="let n of hours" [value]="n">{{ n | number:'2.0-0' }}</option>
               </select>
+              <span>:</span>
               <select formControlName="minute">
                 <option *ngFor="let n of minutes" [value]="n">{{ n | number:'2.0-0' }}</option>
               </select>
-              <select formControlName="meridiem">
+              <!-- <select formControlName="meridiem">
                 <option value="AM">AM</option>
                 <option value="PM">PM</option>
-              </select>
+              </select> -->
             </div>
           </chef-form-field>
         </div>
@@ -73,7 +74,7 @@
           <chef-form-field formGroupName="datetime">
             <div class="datetime">
               <select formControlName="month">
-                <option *ngFor="let n of months" [value]="n">{{ n | number:'2.0-0' }}</option>
+                <option *ngFor="let n of months" [value]="n">{{ n }}</option>
               </select>
               <select formControlName="date">
                 <option *ngFor="let n of dates" [value]="n">{{ n | number:'2.0-0' }}</option>
--- a/components/automate-ui/src/app/page-components/job-schedule-form/job-schedule-form.component.html
+++ b/components/automate-ui/src/app/page-components/job-schedule-form/job-schedule-form.component.html
@@ -78,7 +78,7 @@
                 <option *ngFor="let n of dates" [value]="n">{{ n | number:'2.0-0' }}</option>
               </select>
               <select formControlName="month">
-                <option *ngFor="let n of months" [value]="n"></option>
+                <!-- <option *ngFor="let n of months | keyvalue" [value]="n.key">{{ n.value }}</option> -->
               </select>
               <select formControlName="year">
                 <option *ngFor="let n of years" [value]="n">{{ n }}</option>
--- a/components/automate-ui/src/app/page-components/job-schedule-form/job-schedule-form.component.html
+++ b/components/automate-ui/src/app/page-components/job-schedule-form/job-schedule-form.component.html
@@ -42,7 +42,7 @@
                 <option *ngFor="let n of dates" [value]="n">{{ n | number:'2.0-0' }}</option>
               </select>
               <select formControlName="month">
-                <option *ngFor="let n of months" [value]="n">{{ n }}</option>
+                <option *ngFor="let n of months | keyvalue" [value]="n.key">{{ n.value }}</option>
               </select>
               <select formControlName="year">
                 <option *ngFor="let n of years" [value]="n">{{ n }}</option>
@@ -78,7 +78,7 @@
                 <option *ngFor="let n of dates" [value]="n">{{ n | number:'2.0-0' }}</option>
               </select>
               <select formControlName="month">
-                <option *ngFor="let n of months" [value]="n">{{ n }}</option>
+                <option *ngFor="let n of months" [value]="n"></option>
               </select>
               <select formControlName="year">
                 <option *ngFor="let n of years" [value]="n">{{ n }}</option>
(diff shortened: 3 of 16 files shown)


components/automate-ui/e2e/job-add.e2e-spec.ts
```diff
@@ -556,7 +556,7 @@ describe('Job Add', () => {
           scheduleToggle.click();
 
           expect(scheduleToggle.isSelected()).toEqual(true);
-          expect(startInputs.count()).toEqual(6);
+          expect(startInputs.count()).toEqual(5);
```
SEAjamieD: We've removed the AM/PM input selection, so this test needed to be altered to look for 5 inputs instead of 6

components/automate-ui/e2e/job-add.e2e-spec.ts
```diff
@@ -575,7 +575,7 @@ describe('Job Add', () => {
           endFormGroupToggle.click();
 
           expect(endFormGroupToggle.isSelected()).toEqual(true);
-          expect(endInputs.count()).toEqual(6);
+          expect(endInputs.count()).toEqual(5);
```
SEAjamieD: Same reason as above. Only 5 inputs now.

components/automate-ui/src/app/components/calendar/calendar.component.ts
```diff
@@ -45,7 +45,7 @@ export class CalendarComponent {
 
   @Input()
   set selected(input) {
-    const date = m.isMoment(input) ? input : m(input);
+    const date = m.isMoment(input) ? input : m.utc(input);
```
SEAjamieD: I'm thinking this needs to be altered to check not only if `input` is a moment object, but also if it's using utc?

components/automate-ui/src/app/page-components/job-schedule-form/job-schedule-form.component.html
```diff
@@ -38,12 +38,12 @@ <h4>Set optional scan date and schedule for scans</h4>
...
           <chef-form-field formGroupName="datetime">
             <div class="datetime">
-              <select formControlName="month">
-                <option *ngFor="let n of months" [value]="n">{{ n | number:'2.0-0' }}</option>
-              </select>
               <select formControlName="date">
                 <option *ngFor="let n of dates" [value]="n">{{ n | number:'2.0-0' }}</option>
               </select>
+              <select formControlName="month">
+                <option *ngFor="let n of months" [value]="n">{{ monthsStringArray[n] }}</option>
```
SEAjamieD: I chose to use a simple array to display the months in text form, so that I could essentially leave the rest of the input the same and wouldn't need to do much to change the way job sch are being added

components/automate-ui/src/app/page-components/job-schedule-form/job-schedule-form.component.html
```diff
@@ -72,12 +70,12 @@ <h4>Set optional scan date and schedule for scans</h4>
...
```
SEAjamieD: Same as above - uses an array of months to display the month in text format.

components/automate-ui/src/app/page-components/job-schedule-form/job-schedule-form.component.ts
```diff
@@ -12,13 +13,14 @@ export class JobScheduleFormComponent {
 
   public RRule = RRule;
 
-  public months = Array(12).fill(1).map((_, i) => i + 1);
+  public months = Array(12).fill(1).map((_, i) => i);
+  public monthsStringArray = moment.monthsShort();
```
SEAjamieD: returns the months in short form as an array i.e. [Jan, Feb, Mar, ...]

components/automate-ui/src/app/pages/job-add/job-add.component.ts
```diff
@@ -98,29 +98,27 @@ export class JobAddComponent implements OnDestroy {
...
-          month: defaultStart.month() + 1,
+          month: defaultStart.month(),
           date: defaultStart.date(),
           year: defaultStart.year(),
-          hour: this.hours24to12(defaultStart.hour()),
-          minute: defaultStart.minute(),
-          meridiem: defaultStart.hour() - 12 > 0 ? 'PM' : 'AM'
+          hour: defaultStart.hour(),
+          minute: defaultStart.minute()
         })
```
SEAjamieD: No longer using AM/PM - so we've removed it from the FormGroup.  We also no longer need to convert the hours from 12h to 24h, so we can remove that function because UTC formate is 24h

components/automate-ui/src/app/pages/job-add/job-add.component.ts
```diff
@@ -98,29 +98,27 @@ export class JobAddComponent implements OnDestroy {
...
-          month: defaultEnd.month() + 1,
+          month: defaultEnd.month(),
           date: defaultEnd.date(),
           year: defaultEnd.year(),
-          hour: this.hours24to12(defaultEnd.hour()),
-          minute: defaultEnd.minute(),
-          meridiem: defaultEnd.hour() - 12 > 0 ? 'PM' : 'AM'
+          hour: defaultEnd.hour(),
+          minute: defaultEnd.minute()
         }),
```
SEAjamieD: Same as above.

components/automate-ui/src/app/pages/job-add/job-add.component.ts
```diff
@@ -256,23 +254,24 @@ export class JobAddComponent implements OnDestroy {
     const ruleOpts = {
       dtstart: new Date(
         parseInt(start.datetime.year, 10),
-        parseInt(start.datetime.month, 10) - 1,
+        parseInt(start.datetime.month, 10),
         parseInt(start.datetime.date, 10),
-        this.hours12to24(parseInt(start.datetime.hour, 10), start.datetime.meridiem === 'PM'),
+        parseInt(start.datetime.hour, 10),
         parseInt(start.datetime.minute, 10)
```
SEAjamieD: Similar to above, but for parsing now.  We are no longer using AM/PM and are using 24h format, so function is not needed.

components/automate-ui/src/app/pages/job-edit/job-edit.component.ts
```diff
@@ -111,29 +111,27 @@ export class JobEditComponent implements OnDestroy {
...
-          month: defaultEnd.month() + 1,
+          month: defaultEnd.month(),
           date: defaultEnd.date(),
           year: defaultEnd.year(),
-          hour: this.hours24to12(defaultEnd.hour()),
-          minute: defaultEnd.minute(),
-          meridiem: defaultEnd.hour() - 12 > 0 ? 'PM' : 'AM'
+          hour: defaultEnd.hour(),
+          minute: defaultEnd.minute()
         }),
```
SEAjamieD: Same changes that were made in job-add.component are applied to job-edit.component

components/automate-ui/src/app/components/calendar/calendar.component.ts
SEAjamieD: @msorens do you see any large issue with changing this to be

`const date = m.utc(input)`

so that we can be absolutely sure it's in utc?

per https://github.com/moment/moment/issues/793 , there doesn't seem to be a rock solid way of easily checking if the object is already utc.  The suggestion from moment team mentioned "only for debugging"

SEAjamieD: or possibly check using `isSame()` https://momentjs.com/docs/#/query/is-same/ , however i'm confused by the note at the bottom of the isSame() documenation as to whether they are saying do not use isSame() with nothing passed in, or if they've discovered it is not reliable....
>NOTE: moment().isSame() has undefined behavior and should not be used! If the code runs fast the initial created moment would be the same as the one created in isSame to perform the check, so the result would be true. But if the code runs slower it's possible that the moment created in isSame is measurably after the one created in moment(), so the call would return false.

susanev:
@SEAjamieD if we choose a month without 31 days (or similar) it looks like i can still pick 31, do we care?

also, any chance we can reduce the space between the inputs and the `:` just a bit?
scanjob-scedule

SEAjamieD:
> @SEAjamieD if we choose a month without 31 days (or similar) it looks like i can still pick 31, do we care?
> 
> also, any chance we can reduce the space between the inputs and the `:` just a bit?
> something similar to the screenshot, doesnt need to match exactly
> scanjob-scedule

That I don't know - I haven't changed the actual functionality of how the scan jobs work in terms of how many days are in a month, but I thought that might be an issue.  It seems to me like we SHOULD care - curious what someone else may have to say.

Can definitely shrink of the spacing - no prob!

susanev:
ive created a separate issue #2109 for making the scan job schedule inputs dynamic

components/automate-ui/src/app/pages/job-add/job-add.component.ts
```diff
@@ -98,29 +98,27 @@ export class JobAddComponent implements OnDestroy {
...
-    const defaultEnd = moment(defaultStart).add(1, 'days');
+    const defaultStart = moment.utc();
+    const defaultEnd = moment.utc(defaultStart).add(1, 'days');
     const scheduleGroup = this.fb.group({
       name: ['', Validators.required],
       include: false,
       start: this.fb.group({
         datetime: this.fb.group({
-          month: defaultStart.month() + 1,
+          month: defaultStart.month(),
```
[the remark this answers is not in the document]
SEAjamieD: This is because previously, we were listing the months as a numeric (i.e. Jan = 1) and we were populating that initial array by counting up 12 times, starting from 0. [0, 1, .... 11].  To display properly required that we add a 1 to each month.  

Now that we are displaying the list based on an array of strings ['Jan', 'Feb',.....], we need the Array to start at 0 again, otherwise it will display "Feb" for 1 in the UI.

components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-overview/reporting-overview.component.ts
```diff
@@ -144,9 +144,9 @@ export class ReportingOverviewComponent implements OnInit, OnDestroy {
 
   onDateChanged(endDate) {
     const queryParams = {...this.route.snapshot.queryParams};
-    const endDateMoment = moment(endDate);
+    const endDateMoment = moment.utc(endDate);
     if (endDate && endDateMoment.isValid()) {
-      if (moment().utc().format('YYYY-MM-DD') === endDateMoment.format('YYYY-MM-DD')) {
+      if (moment.utc().format('YYYY-MM-DD') === endDateMoment.format('YYYY-MM-DD')) {
```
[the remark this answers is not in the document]
SEAjamieD: I don't believe it does, but since the documentation provides all of its examples as `moment.utc()` it seemed like something I should update. https://momentjs.com/docs/#/parsing/utc/

components/automate-ui/src/app/components/calendar/calendar.component.ts
SEAjamieD: a date that has been acted on by `moment()` is converted into an object of key value pairs.  one of those keys is an `_isUTC` boolean which by default is set to False.  To answer your question, when an input is already a moment, then you do a `utc()` on it, the utc flag is then set to True.  Nothing else in the object is affected.  The downside being that it runs a function that would be unnecessary if it was already utc.

PR #2054, opened 2019-10-30
document external facing reporting stats proto

vjeffrey:

Documenting the external facing stats proto. 

fixes https://github.com/chef/automate/issues/2018

### :+1: Definition of Done
stats proto is documented

- check out this branch
- run make serve from the automate-chef-io directory
- visit http://localhost:1313/docs/api

### DIFF
--- a/components/automate-chef-io/data/docs/api-static/03-tag-groups.swagger.json
+++ b/components/automate-chef-io/data/docs/api-static/03-tag-groups.swagger.json
@@ -9,7 +9,11 @@
         "Compliance Reporting Profiles",
         "Compliance Reporting Report IDs",
         "Compliance Reporting Reports",
-        "Compliance Reporting Suggestions"
+        "Compliance Reporting Suggestions",
+        "Compliance Stats Summary",
+        "Compliance Stats Trend",
+        "Compliance Stats Profiles",
+        "Compliance Stats Failures"
       ]
     }
   ]
--- a/components/automate-gateway/api/compliance/reporting/stats/stats.proto
+++ b/components/automate-gateway/api/compliance/reporting/stats/stats.proto
@@ -11,7 +11,13 @@ import "components/automate-grpc/protoc-gen-policy/iam/annotations.proto";
 import "components/automate-gateway/api/compliance/reporting/reporting.proto";
 
 service StatsService {
-	// should cover /summary, /summary/nodes, /summary/controls
+	/*
+	Read Summary
+
+	Returns summary statistics for compliance reports. 
+	General report summary information is the default. Adding a type value of nodes or controls will return summary statistics for that object.
+	Supports filtering.
+	*/
 	rpc ReadSummary(Query) returns (Summary) {
 		option (google.api.http) = {
 			post: "/compliance/reporting/stats/summary"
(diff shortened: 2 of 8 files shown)


components/automate-chef-io/data/docs/api-static/03-tag-groups.swagger.json
```diff
@@ -9,7 +9,11 @@
         "Compliance Reporting Profiles",
         "Compliance Reporting Report IDs",
         "Compliance Reporting Reports",
-        "Compliance Reporting Suggestions"
+        "Compliance Reporting Suggestions",
+        "Compliance Stats Summary",
+        "Compliance Stats Trend",
+        "Compliance Stats Profiles",
+        "Compliance Stats Failures"
```
phiggins: This makes me realize there is some cruft / complexity around how we're handling tags and groups due to learning how they worked. I'll add it to the TODO list.

components/automate-chef-io/data/docs/api_chef_automate/iam/v2beta/tokens.swagger.json
```diff
@@ -188,6 +188,9 @@
         }
       }
     },
+    "chef.automate.api.iam.v2beta.ResetAllTokenProjectsResp": {
+      "type": "object"
+    },
     "chef.automate.api.iam.v2beta.Token": {
       "type": "object",
       "properties": {
```
phiggins: Some of these changes shouldn't be in this PR. I just merged #2004 which should clean this up if you rebase.

components/automate-gateway/api/compliance/reporting/stats/stats.proto
```diff
@@ -11,7 +11,13 @@ import "components/automate-grpc/protoc-gen-policy/iam/annotations.proto";
...
 
 service StatsService {
-	// should cover /summary, /summary/nodes, /summary/controls
+	/*
+	Read Summary
+
+	Returns summary statistics for compliance reports. 
+	General report summary information is the default. Adding a type value of nodes or controls will return summary statistics for that object.
+	Supports filtering.
+	*/
```
phiggins: We haven't really established any conventions, but this supports markdown so you can do

> Adding a `type` value of `nodes` or `controls`

etc to show what are keys/values.

components/automate-chef-io/data/docs/api-static/03-tag-groups.swagger.json
vjeffrey: ya i kinda just added those cause i saw there were some for reporting. not actually sure what they do! 😺

components/automate-gateway/api/compliance/reporting/stats/stats.proto
```diff
@@ -77,47 +106,73 @@ service StatsService {
...
 	int32 per_page = 24;
 }
 
 message Summary {
+	// Statistics for the controls executed in the compliance reports.
 	ControlsSummary controls_summary = 1;
+	// Statistics about the nodes scanned in the compliance reports.
 	NodeSummary node_summary = 2;
+	// Statistics on the overall compliance reports.
 	ReportSummary report_summary = 3;
```
phiggins: I haven't checked this locally, but I think these probably won't show up in the browser. The ones commented `// Intentionally blank.` in `reporting.proto` have the same behavior due to a limitation of swagger. The description for a single message reference is taken from where the message is defined, not where its referenced. That is, you'd need to document this line to provide documentation for this field: https://github.com/chef/automate/pull/2054/files#diff-35e6bfeb760490166078b62bdee851c4R196

vjeffrey: oooooh interesting!

vjeffrey: oooh i see. if it's a repeated field, the description written above the field in the message is included. if it's not a repeated field, we end up in this situation. got it.

components/automate-chef-io/data/docs/api-static/03-tag-groups.swagger.json
kagarmoe: I handled it in the 00 file, which looks like it overrides this file.

PR #2055, opened 2019-10-30
compliance: worker count default based on CPU count

stevendanna:
The default of 10 workers can easily overwhelm a system with a small
number of CPUs. This sets the default the numCPUs + 2, clamped to [2,
10].


stevendanna:
We might want to factor memory into this calculation as well, but this is a start.

PR #2071, opened 2019-10-31
Refactor EAS api to be consistent

kmacgugan:
Use filter=status:blah for all calls
Removes health from sidebar call
disconnected is also a status now for the api
Does not support multi select with disconnected status and another status

Both service group and service side bar calls will now look like this for status filters:
filter=status:critical
filter=status:warning
filter=status:ok
filter=status:unknown
filter=status:disconnected

After starting all services, rebuild the applications service
Populate some data with `applications_populate_database`
Use status filters on the service groups table and service groups sidebar and see that they are giving you the correct entities in response. 

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/applications-service/integration_test/service_groups_filters_test.go
```diff
@@ -223,32 +223,34 @@ func TestDisconnectedStatusFilter(t *testing.T) {
...
-		require.NoError(t, err)
-
-		req := &applications.ServiceGroupsReq{
-			Filter: []string{"connectedStatus:disconnected", "status:critical"},
-		}
-		response, err := suite.ApplicationsServer.GetServiceGroups(ctx, req)
-		assert.Nil(t, err)
-		assert.Len(t, response.ServiceGroups, 1)
-	})
+	// TODO: Re-enable this test when we support multiple selected filters
```
vjeffrey: is there a card to follow up on this todo?

components/automate-ui/src/app/entities/service-groups/service-groups.requests.ts
```diff
@@ -68,15 +68,15 @@ export class ServiceGroupsRequests {
     const options = {
       params: this.buildServicesBySGFilterParams(filters)
     };
-
+    console.log("object: %O", options)
```
vjeffrey: ? accidentally included in the change?

components/applications-service/integration_test/service_groups_filters_test.go
kmacgugan: No we don't have one yet, this idea is still in the ideation phase. @apriofrost has talked about making it a cross-automate design change on all tabs.

components/automate-ui/src/app/entities/service-groups/service-groups.requests.ts
kmacgugan: oops! yes it snuck in there.

PR #2083, opened 2019-11-01
update golangci-lint to 1.21.0

stevendanna:
This allows us to set the modules-download-mode option. I've disabled
some newly added linters. I've also converted our configuration file
to YAML since the upstream example documentation is in yaml and this
makes it easier to copy/paste between the documentation and our
configuration.

The newly disabled liters are `whitespace`, `wsl`, `dogsled`, and `gocognit`. `gocognit` looks useful but required non-trivial code changes to fix.  The others look like low-value style related linters.


msorens:
Consider documenting (i.e. just in this PR) why those particular linters need to be disabled.

stevendanna:
> Consider documenting (i.e. just in this PR) why those particular linters need to be disabled.

Well, they don't *need* to be. I could have fixed the violations, but, I've updated the PR comment with an explanation fo why I didn't.

PR #2093, opened 2019-11-01
Auth 1877/2p1 conditional removal

blakestier:

We now only have a v1 and a v2.1; we can remove any conditionals around minor versions.


components/automate-ui/e2e/admin.e2e-spec.ts
```diff
@@ -103,485 +102,6 @@ describe('Admin pages', () => {
     });
   });
 
-
-  describe('IAM v2', () => {
```
msorens: Why delete all the token tests here? Or have they just been moved to the new 06_token_mgmt_spec ? Though, seems like the set of tests for token details has been reduced a lot.

components/automate-ui/src/app/entities/policies/policy.selectors.ts
```diff
@@ -16,15 +16,9 @@ export const iamMajorVersion = createSelector(
   (state) => state.iamMajorVersion
 );
 
-export const iamMinorVersion = createSelector(
+export const isIAMv2 = createSelector(
```
msorens: Most of the UI code uses not-equals-v1 rather than equals-v2 to allow for future expansion. The latter would make things stop working when v3 comes along, which will probably not be what we want.
Thus, would suggest either make this `isIAMv1` or rename to `atLeastV2`, depending on what polarity you want.
Then L21 would compare to 'v1' in either case.

blakestier: Thank you for the reminder to search for `!== 'v1'`! Will do.

I'd like to leave the equality for now. We don't know what v3 looks like, yet, so I'm not confident we won't need to modify this variable anyway.

components/automate-ui/e2e/admin.e2e-spec.ts
blakestier: They've all been moved, short of checking for the heading (API Tokens) and the date. 

Protractor has been on our 💥 list for awhile, so we took this opportunity to port to cypress. You think those extra assertions are worth it? (Would just be some awkwardness around date conversion but 🤷‍♂ )

msorens: Did not look deeply; if you've considered it already, that's sufficent.

components/automate-ui/src/app/components/settings-sidebar/settings-sidebar.component.html
```diff
@@ -1,4 +1,4 @@
-<chef-sidebar [attr.major-version]="iamMajorVersion$ | async">
+<chef-sidebar>
```
msorens: FWIW: That mechanism was for use by Cypress tests, which I had forgotten about. Might be less work and less error-prone to use that (though moved to the top level app file) instead of having to set the cypress version in many places. Just a thought.

components/automate-ui/src/app/components/settings-sidebar/settings-sidebar.component.spec.ts
```diff
@@ -110,10 +105,6 @@ describe('SettingsSidebarComponent', () => {
   });
 
   describe('IAM v1', () => {
-    beforeEach(() => {
-      component.iamMajorVersion$ = observableOf(<IAMMajorVersion>'v1');
```
msorens: Wondering if this doesn't need a corresponding `observableOf(false)` somewhere...?

components/automate-ui/src/app/pages/team-management/team-management.component.ts
```diff
@@ -169,10 +162,12 @@ export class TeamManagementComponent implements OnInit, OnDestroy {
   }
 
   public openCreateModal(): void {
-    if (this.isMajorV1) {
-      this.createV1TeamModalVisible = true;
-    } else {
+    let isV2 = true;
+    this.isIAMv2$.subscribe(latest => isV2 = latest);
```
msorens: This will leak subscriptions a lot. These lines need to be part of the initialization for the file (`ngOnInit`).
Also, because you need the value _both_ in the code-behind and in the template, you can make the template cleaner. Instead of multiple uses of `isIAMv2$ | async` you can just use `isV2`.

blakestier: Oh, great-- I was actually trying to understand this a little more. Wasn't clear on when I could get away with that vs the async call.

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -95,20 +92,15 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
         [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]),
       projects: [[]]
     });
-    this.store.pipe(
-      select(isIAMv2),
-      takeUntil(this.isDestroyed))
-      .subscribe(projectsEnabled => {
-        this.projectsEnabled = projectsEnabled;
-      });
+    store.select(isIAMv2).subscribe(latest => this.isIAMv2 = latest);
```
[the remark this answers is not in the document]
blakestier: I moved it to the init 🎉

e2e/cypress/integration/ui/settings/06_token_mgmt.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    cy.get('chef-notification').should('be.visible');
+    cy.get('chef-notification').should('contain', 'API Token copied to clipboard.');
+  });
+
+  it('can delete token', () => {
+    cy.get('chef-tbody').contains(tokenName1).parent().parent()
+      .find('[data-cy=token-control]').as('controlMenu');
+    cy.get('@controlMenu').should('be.visible').click();
+    cy.get('@controlMenu').find('[data-cy=delete]').click({ force: true });
```
tylercloke: Do we need the force here?

components/automate-ui/src/app/pages/team-management/team-management.component.ts
```diff
@@ -75,13 +73,7 @@ export class TeamManagementComponent implements OnInit, OnDestroy {
...
   ngOnInit() {
     this.store.dispatch(new GetTeams());
-
-    this.store.select(iamMajorVersion)
-      .pipe(takeUntil(this.isDestroyed))
-      .subscribe((majorVersion) => {
-        this.isMajorV1 = majorVersion === 'v1';
-      });
-
+    this.store.select(isIAMv2).subscribe(latest => this.isIAMv2 = latest);
```
msorens: Placement is good. 👍 
But one more thing--just did not think to comment yesterday: you have to have the `takeUntil(this.isDestroyed)` there, otherwise you have the classic leaking subscription we have been talking about. That is, just reuse the pattern on the deleted lines L79-83.

(Applies to team-details as well.)

blakestier: oh @bcmdarroch pointed this out-- didn't realize I need both!

e2e/cypress/integration/ui/settings/06_token_mgmt.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+    // can still use v2beta APIs while on v1
+    cy.cleanupV2IAMObjectsByIDPrefixes(cypressPrefix, ['tokens']);
+  });
+
+  it('displays heading and table', () => {
+    cy.get('chef-heading').contains('API Tokens');
+    cy.get(tokensTable).should('exist');
+  });
+
+  it('displays the returned tokens info', () => {
```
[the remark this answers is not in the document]
blakestier: Fwiw, this test passed. Several below didn't since the `data-cy` attribute isn't on acceptance yet.

e2e/cypress/integration/ui/settings/06_token_mgmt.spec.ts
blakestier: I'll double check.

e2e/cypress/integration/ui/settings/06_token_mgmt.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+        url: '/api/v0/auth/tokens',
+        body: {
+          id: tokenID2,
+          description: tokenName2,
+          active: false
+        }
+      });
+
+      // load new tokens
+      cy.reload(true);
```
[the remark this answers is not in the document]
bcmdarroch: good point, `reload` was a bit of a lazy shortcut anyway

PR #2094, opened 2019-11-01
Fix proto compilation generated file CI task

phiggins:
This CI task was changed in #1837 from being specified inline in the buildkite config:

https://github.com/chef/automate/blob/6f05b7a037f2457852ee540bfc5b9439c3581035/.expeditor/verify.pipeline.yml#L13-L20

to being a bash function:

https://github.com/chef/automate/blob/91329cbb7d6093152eacbc8053bc5cfde163cb2f/.studio/common#L116-L132

My guess is that this if statement does not do the correct thing, the `if !` is two bash commands so one of them must change the exit code before the `return $?` statement.

The first commit changes the code buildkite runs, so it failed as expected because there are proto differences between what's checked in on master and what is generated by compiling the protos.

The second commit adds the proto file changes that are missing, so it should pass.

### DIFF
--- a/.studio/common
+++ b/.studio/common
@@ -119,14 +119,10 @@ DOC
 function verify_all_protobuf_components() {
   install_if_missing core/git git
 
-  if ! compile_all_protobuf_components; then
-    return $?
-  fi
+  compile_all_protobuf_components || return $?
 
   git add .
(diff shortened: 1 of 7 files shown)


phiggins:
@stevendanna Did you mean to push a commit to this branch? I'm not sure what that does...

PR #2096, opened 2019-11-02
Redo API docs tags.

phiggins:
While reviewing #2054, I noticed the setup for tags and tag groups was more complicated than it needed to be and there was some unnecessary cruft present, mostly as a result of figuring out how tags worked across the proto/swagger/redoc interactions.

In this PR:

* Remove unnecessary tag setup that we're not using anymore.
* Move all the tag setup into one file.
* Use the default tags in the proto to remove an unnecessary level of redirection.


components/automate-gateway/api/compliance/reporting/reporting.proto
```diff
@@ -67,9 +64,6 @@ service ReportingService {
 			resource: "compliance:reporting:report-ids"
 			action: "compliance:reportids:list"
 		};
-		option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
-			tags: "Reports";
-		};
```
phiggins: As a result of this unnecessary level of redirection, we forgot to update the tag of this endpoint (`ListReportIds`) so it wasn't being displayed when viewing the API docs in the browser.

phiggins:
> undoing the tags stuff makes it so that all of the proto docs are visible. do we want to hide the ones we haven't yet actually documented?

It's my understanding that only things tagged with a tag in a tag group will show up in the UI, but I'll double check.

phiggins:
I just rebuilt the docs and only see the Reporting and Stats endpoints in the UI, so I think this is doing what's intended. If you're seeing other behavior please let me know!

vjeffrey:
oh interesting. i def saw them all this morning but i’ll pull things down and check again later!

PR #2099, opened 2019-11-04
Account for -mod=vendor in `go list` usages

stevendanna:
We've seen CI failures that indicate that we are still not using
-mod=vendor everywhere.  This PR accounts for uses of `go list`.

At the heart of this change is the fact that doing:

    export GOFLAGS = -mod=vendor

in a Makefile does not affect commands started with `$(shell
CMD)` (see http://savannah.gnu.org/bugs/?10593 for details).

To account for this, I've take the following strategies:

- Remove unnecessary uses of `$(shell go list)`. Most go commands can
  take package list directly.

- Replacing some uses of `go list | grep -v some_package` with build
  tags.

- Adding GOFLAGS='$(GOFLAGS)' to `go list` invocations that appeared
  harder to replace.

In addition to this I've

- Removed explicit `go vet` Make targets since we also run govet as
  part of golangci-lint.

- Replaced a custom check-go-version implementation in
  compliance-service with the shared implementation.

### DIFF
--- a/.studio/event-service
+++ b/.studio/event-service
@@ -13,5 +13,5 @@ document "event_integration" <<DOC
   Run event service integration tests
 DOC
 function event_integration() {
-  go_test -v "github.com/chef/automate/components/event-service/integration_test"
+  go_test -v -tags=integration "github.com/chef/automate/components/event-service/integration_test"
 }
--- a/components/automate-cli/Makefile
+++ b/components/automate-cli/Makefile
@@ -59,4 +59,4 @@ linux darwin windows:
 	@echo Done: ${CROSS_TARGET_DIR}/${BUILD_TARGET}
 
 unit:
-	@go test --count 1 $(shell go list ./cmd/... ./pkg/...)
+	@go test --count 1 ./...
--- a/components/data-feed-service/Makefile
+++ b/components/data-feed-service/Makefile
@@ -1,13 +1,10 @@
-GOFILES=$(shell go list ./... | grep -v '/vendor|asset\.go|proto/' | sed 's/^_//')
-GOPATH=$(shell go env GOPATH)
-
 include ../../Makefile.common_go
 
 default: ci
 
 ci: review build
 
-review: lint vet
+review: lint
 
 # Utilities
 build:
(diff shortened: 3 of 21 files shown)


.studio/event-service
```diff
@@ -13,5 +13,5 @@ document "event_integration" <<DOC
   Run event service integration tests
 DOC
 function event_integration() {
-  go_test -v "github.com/chef/automate/components/event-service/integration_test"
+  go_test -v -tags=integration "github.com/chef/automate/components/event-service/integration_test"
```
msorens: Based solely on the pattern I see of "integration tests should add this tag", I would think the same change should happen for these: 
.studio/secrets-service
.studio/pg-sidecar-service
.studio/ingest-service
.studio/event-feed-service
.studio/es-sidecar-service
.studio/config-mgmt-service
.studio/compliance-servic

components/automate-cli/Makefile
```diff
@@ -59,4 +59,4 @@ linux darwin windows:
 	@echo Done: ${CROSS_TARGET_DIR}/${BUILD_TARGET}
 
 unit:
-	@go test --count 1 $(shell go list ./cmd/... ./pkg/...)
+	@go test --count 1 ./...
```
msorens: Based on the pattern "go test with go list should change to this", also consider:
api/Makefile

components/data-feed-service/Makefile
```diff
@@ -1,13 +1,10 @@
-GOFILES=$(shell go list ./... | grep -v '/vendor|asset\.go|proto/' | sed 's/^_//')
-GOPATH=$(shell go env GOPATH)
```
msorens: Based on the pattern "GOPATH should go away" also consider:
components/cereal-service/Makefile
components/compliance-service/Makefile

These two files are unique in their use of GOPATH without actually defining it; is that a problem?
components/nodemanager-service/Makefile
components/secrets-service/Makefile

components/data-feed-service/Makefile
```diff
@@ -1,13 +1,10 @@
...
-
 include ../../Makefile.common_go
 
 default: ci
 
 ci: review build
 
-review: lint vet
```
msorens: There are a few leftover `vet` targets, all in `.PHONY` lines:
components/authn-service/Makefile
components/data-feed-service/Makefile
components/local-user-service/Makefile
components/session-service/Makefile

stevendanna:
Thanks @msorens I'm going to merge this, but I'll follow up with some PRs over the week each which does a specific cleanup rather than putting them all in this build-fix PR.

Issue #2104, opened 2019-11-04
Audit Angular Subscriptions for Leakage


sdelano:
We have noticed that some UI pages perform multiple of the same requests when switching between the tabbed sections on the page. This task is to find all the pages that have this problem. Then fix them. 

One example page is the policy detail page. When switching between the "Definition" and "Member" tab an extra request to "apis/iam/v2beta/policies/administrator-access" is made. 
https://a2-dev.test/settings/policies/administrator-access#definition

lancewf:
The two pages below are showing this problem:
- [  ] https://a2-dev.test/settings/teams/admins
- [  ] https://a2-dev.test/settings/policies/administrator-access

msorens:
I would add that **all** auth pages need to be audited for subscription leakage. I am reasonably sure that we have missed a couple `unsubscribe` actions here and there, but that they did not seem harmful at the time.

sdelano:
Adding the `emergent` tag here because this work was discovered as part of the investigation in solving #1062. We started this even though it wasn't a required part of solving that problem.

PR #2113, opened 2019-11-05
Removing Duplicate requests for details pages

lancewf:
Updating the policy details and team details pages to not request data when the page's internal tabs are changed. This solution combines two observers that are triggered off URL updates. Added the "distinctUntilChanged" filter to prevent extra data request when the internal tabs are changed. The "distinctUntilChanged" filter prevents multiple events from being sent when the data has not changed. 

https://github.com/chef/automate/issues/2104

### :+1: Definition of Done
The data on the policy details and team details pages is only requested once when the page is first loaded. Changing the page's internal tabs does not re-request data. 

1. Open the Teams detail page for the admin https://a2-dev.test/settings/teams/admins
1. Open the Chrome inspect tool then select the network tab. 
1. On the page move between the "Users" and "Details" tab. 
1. Ensure that there are no calls to the https://a2-dev.test/apis/iam/v2beta/teams/admins API.
1. Do the same for the Policies details page https://a2-dev.test/settings/policies. 

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/pages/policy/details/policy-details.component.ts
+++ b/components/automate-ui/src/app/pages/policy/details/policy-details.component.ts
@@ -7,7 +7,6 @@ import { filter, map, takeUntil, distinctUntilChanged } from 'rxjs/operators';
 
 import { NgrxStateAtom } from 'app/ngrx.reducers';
 import { routeParams } from 'app/route.selectors';
-import { Params } from '@angular/router';
 import { routeURL } from 'app/route.selectors';
 import { GetPolicy } from 'app/entities/policies/policy.actions';
 import { policyFromRoute } from 'app/entities/policies/policy.selectors';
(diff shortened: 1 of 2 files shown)


components/automate-ui/src/app/pages/policy/details/policy-details.component.ts
```diff
@@ -74,23 +75,20 @@ export class PolicyDetailsComponent implements OnInit, OnDestroy {
...
-              this.store.dispatch(new GetPolicy({ id }));
-            }
-        });
-      });
+    combineLatest([
+      this.store.select(routeParams),
+      this.store.select(routeURL)
+    ]).pipe(
+      takeUntil(this.isDestroyed),
+      map(([params, url]: [Params, string]) => [params.id, url]),
```
msorens: Until recently I was diligent in advocating to make sure to add explicit types (like you have done here) for these rxjs operators, because there was a typescript bug: type inference was giving them all the same type even if they were not! But it looks like that has been fixed, so now it is redundant. That is, with just this... 
```suggestion
      map(([params, url]) => [params.id, url]),
```
...if you hover over `params` or `url` you will see the correct types inferred.

On the other hand, sometimes it is helpful to provide explicit types if the types are not "obvious", so it is a personal preference.

But on the third hand, this code block uses that pattern four times, which makes the block more busy; it would be more readable if you eliminate all four occurrences.

And finally, if doing so, this line really needs to become:
```suggestion
      map(([params, url]) => [params.id as string, url]),
```
Because `params.id` is of type `any`.

lancewf: I am not seeing that the types are inferred in VS code.

lancewf:
@msorens Thank you for the review. I will look into fixing the other problems in the next PR.

lancewf:
@bcmdarroch That is a funny story. I was watching the below video to figure out what the rxjs pluck function is. 
https://www.youtube.com/watch?v=IVBJ3-M_YsQ&t=136s

components/automate-ui/src/app/pages/policy/details/policy-details.component.ts
```diff
@@ -74,23 +75,20 @@ export class PolicyDetailsComponent implements OnInit, OnDestroy {
...
+      // Only fetch if we are on the policy details route, otherwise
+      // we'll trigger GetPolicy with the wrong input on any route
+      // away to a page that also uses the :id param.
+      filter(([id, url]: [string, string]) => POLICY_DETAILS_ROUTE.test(url) && id !== undefined),
+      // Remove the url because we only need to check if the id has changed
+      map(([id, _url]: [string, string]) => id),
+      distinctUntilChanged()
```
tylercloke: 🎉

components/automate-ui/src/app/pages/policy/details/policy-details.component.ts
lancewf: I am seeing it now. 🤷‍♂

PR #2115, opened 2019-11-05
bump lib/pq to 1.2.0

stevendanna:
Full set of changes:

https://github.com/lib/pq/compare/83612a56d3dd...3427c32cb71afc948325f299f040e53c1dd78979

Highlights include:

- A fix for a potential deadlock when related to commit and rollback
- QuoteLiteral function that can replace our custom implementation


stevendanna:
This appears to be hitting this:  https://github.com/lib/pq/issues/870

Namely, if HAProxy times out our db connection before at the wrong time, we can get a broken pipe write error that we didn't before. We'll have to sort out what we want to do here.

phiggins:
There's some similar looking quoting code in here: https://github.com/chef/automate/blob/master/lib/pgutils/pgutils.go

vjeffrey:
can we close this, since we did the lib/pq thing?

stevendanna:
Yup! Thanks @danielsdeleo

Issue #2061, opened 2019-10-31
Convert Stencil components to Angular components


SEAjamieD:
## User Story

As a Developer, I will use this work when I want to build out any html template so that styles, effects, and structures are more consistent across Automate.  This will also improve performance on many of the larger, more complex components.

## Definition of Done

Stencil is no longer being used in Automate

## Priorities
Automate team has Identified some components that have taken a higher priority over others.
As of Nov. 5 2019, Priority is as follows:

Chef-option
Chef-contol-menu
chef-table

SEAjamieD:
UI has had a change of plans and this epic is no longer needed

PR #2128, opened 2019-11-05
Jamie/chef input from stencil

SEAjamieD:
As we move away from stencil based components, we will be removing them from the codebase one tag at a time.  This time its `<chef-input>`.  Because `<chef-input>` holds no logic in it, it will be replaced with a native `input` and styled using the `chefInput` directive

https://github.com/chef/automate/issues/2061

### :+1: Definition of Done
`<chef-input>` stencil component is no longer used anywhere on the site, and has been replaced with a native `<input>` tag.  Visual appearance should remain the same.

1.  build components/automate-ui-devproxy && start_all_services
2. Navigate across the site and all modals to see that <chef-input> has been replaced with native inputs.

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/components/input/input.directive.scss
+++ b/components/automate-ui/src/app/components/input/input.directive.scss
@@ -17,10 +17,10 @@
 
 ::ng-deep .chef-input, ::ng-deep [compound-input] {
   width: 100%;
+  height: 45px;
   background-color: $white;
   padding: 1em;
-  // testing
-  // font-size: 0.75em;
+  font-size: 14px;
   border: 1px solid $chef-light-grey;
   border-radius: $global-radius;
   box-sizing: border-box;
--- a/components/automate-ui/src/app/components/input/input.directive.scss
+++ b/components/automate-ui/src/app/components/input/input.directive.scss
@@ -19,7 +19,8 @@
   width: 100%;
   background-color: $white;
   padding: 1em;
-  font-size: 0.75em;
+  // testing
+  // font-size: 0.75em;
   border: 1px solid $chef-light-grey;
   border-radius: $global-radius;
   box-sizing: border-box;
--- a/components/automate-ui/src/app/components/input/input.directive.scss
+++ b/components/automate-ui/src/app/components/input/input.directive.scss
@@ -29,6 +29,7 @@
   &:focus,
   &:active {
     border-color: $chef-primary-bright;
+    outline: none;
   }
 
   &[disabled] {
--- a/components/automate-ui/src/app/components/input/input.directive.scss
+++ b/components/automate-ui/src/app/components/input/input.directive.scss
@@ -12,7 +12,6 @@
   color: $chef-primary-dark;
   font-size: 0.75em;
   font-weight: bold;
-  padding-bottom: 0.5em;
 }
 
 ::ng-deep .chef-input, ::ng-deep [compound-input] {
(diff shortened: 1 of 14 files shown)


components/automate-ui/src/app/components/input/input.directive.scss
```diff
@@ -19,7 +19,8 @@
   width: 100%;
   background-color: $white;
   padding: 1em;
-  font-size: 0.75em;
+  // testing
+  // font-size: 0.75em;
```
SEAjamieD: @susanev - Is there a uniform font size we want to be using across all inputs?  I found that different inputs of the same type (i.e. all type="text" inputs) were in certain cases, using a very slightly different font size.  Removing the font size here set them all to uniform, but I'm thinking it is probably inheriting whatever our 1em is set to.

components/automate-ui/src/app/components/input/input.directive.scss
```diff
@@ -29,6 +30,7 @@
   &:focus,
   &:active {
     border-color: $chef-primary-bright;
+    outline: none;
```
SEAjamieD: we defined our own version of a focused input outline, so this outline needs to be removed - uniform across all inputs.

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.scss
```diff
@@ -10,8 +10,9 @@ chef-td {
   }
 }
 
-chef-input {
+input {
   width: 60px;
+  height: 100%;
```
SEAjamieD: this height was needed in order to keep the inputs on this form all uniform size.

components/automate-ui/src/app/pages/project/rules/project-rules.component.scss
```diff
@@ -16,9 +16,8 @@ chef-page {
     overflow-y: scroll;
   }
 
-  chef-input {
+  input {
     height: 45px;
```
SEAjamieD: @susanev - This seems to be the only place where 45px needs to be explicitly set because it is surrounded by selects that are set to 45px.  Not sure if this has been seen, would be helpful if we could make all the inputs the same uniform height....or maybe they are already supposed to be?

components/automate-ui/src/app/components/input/input.directive.scss
susanev: 14px would be so rad!

components/automate-ui/src/app/pages/project/rules/project-rules.component.scss
susanev: yes, really want all the inputs the same size.
im good with whatever height makes the most sense for everything
heres an example where the difference is kinda bad
create-team-modal

susanev:
@SEAjamieD looks good, i noticed the errors on the user create still have some extra space above them; same issue on the your profile page

create-user



also im wondering if we should have a default length as well? there are some really long inputs in places....

SEAjamieD:
> @SEAjamieD looks good, i noticed the errors on the user create still have some extra space above them; same issue on the your profile page
> also im wondering if we should have a default length as well? there are some really long inputs in places....

Thanks @susanev - I'll tackle those error gaps now.  With regards to input sizing.  I think that by default, the inputs should be 100%, and then in the areas where we need them shorter, we can place explicit widths on them, this way they will always fill their containers, but we can constrain them when needed.

PR #2130, opened 2019-11-06
[automate-2104] reset items in ngrx/store between GET requests

bcmdarroch:

We have some data leakage issues between the auth list and details pages. If you navigate to the list page and then go to one of the details pages, the data still exists in the store and gets loaded on the details page. This is a problem in cases when a user might have List permission but not Get permission.

- [x] tokens
- [x] users
- [x] teams

- follow-up: https://github.com/chef/automate/issues/2138
- in the future we may rethink what data our List and Get APIs should be returning

### :+1: Definition of Done
- [x] the tokens details page always displays fresh data from the API instead of pulling stale data from the store
- [x] the teams details page always displays fresh data from the API instead of pulling stale data from the store
- [x] the users details page always displays fresh data from the API instead of pulling stale data from the store

```
$ export TOK=`chef-automate iam token create adm --admin`

$ echo '{
    "name": "testing",
    "id": "test1",
    "members": [
      "user:local:coolbean"
    ],
    "statements": [
      {
        "effect": "ALLOW",
        "actions": ["iam:*:list"],
        "projects": [
          "*"
        ]
      }
    ],
    "projects": []
  }' >> pol.json

$ curl -kH "api-token: $TOK" -X POST -d @pol.json https://localhost/apis/iam/v2beta/policies | jq

{
  "policy": {
    "name": "testing",
    "id": "test1",
    "type": "CUSTOM",
    "members": [
      "user:local:coolbean"
    ],
    "statements": [
      {
        "effect": "ALLOW",
        "actions": [
          "iam:*:list"
        ],
        "role": "",
        "resources": [
          "*"
        ],
        "projects": [
          "*"
        ]
      }
    ],
    "projects": []
  }
}
```

- Log in to the UI
- Create some tokens, users, teams
- Create user `coolbean` and log in as that user
- Navigate to https://a2-dev.test/settings/tokens
- Make a token and go to its details page
- None of that token's data will load, all the fields will be blank
- Repeat with teams and users

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/entities/api-tokens/api-token.reducer.ts
+++ b/components/automate-ui/src/app/entities/api-tokens/api-token.reducer.ts
@@ -45,7 +45,12 @@ export function apiTokenEntityReducer(
   switch (action.type) {
 
     case ApiTokenActionTypes.GET_ALL: {
-      return set(STATUS, EntityStatus.loading, state) as ApiTokenEntityState;
+      return set(
+        STATUS,
+        EntityStatus.loading,
+        // clear token state to prevent data leakage
+        apiTokenEntityAdapter.removeAll(state)
+      ) as ApiTokenEntityState;
     }
 
     case ApiTokenActionTypes.GET_ALL_SUCCESS: {
@@ -61,7 +66,13 @@ export function apiTokenEntityReducer(
     }
 
     case ApiTokenActionTypes.GET: {
-      return set(GET_STATUS, EntityStatus.loading, state) as ApiTokenEntityState;
+      return set(
+        GET_STATUS,
+        EntityStatus.loading,
+        // clear token state to prevent data leakage
+        // for example, if the user visited the list page first
+        apiTokenEntityAdapter.removeAll(state)
+      ) as ApiTokenEntityState;
     }
 
     case ApiTokenActionTypes.GET_SUCCESS: {
--- a/components/automate-ui/src/app/entities/api-tokens/api-token.reducer.ts
+++ b/components/automate-ui/src/app/entities/api-tokens/api-token.reducer.ts
@@ -48,7 +48,7 @@ export function apiTokenEntityReducer(
       return set(
         STATUS,
         EntityStatus.loading,
-        // clear token state to prevent data leakage
+        // clear token state to avoid leaking potentially unauthorized data
         apiTokenEntityAdapter.removeAll(state)
       ) as ApiTokenEntityState;
     }
@@ -69,8 +69,7 @@ export function apiTokenEntityReducer(
       return set(
         GET_STATUS,
         EntityStatus.loading,
-        // clear token state to prevent data leakage
-        // for example, if the user visited the list page first
+        // clear token state to avoid leaking potentially unauthorized data
         apiTokenEntityAdapter.removeAll(state)
       ) as ApiTokenEntityState;
     }
--- a/components/automate-ui/src/app/entities/api-tokens/api-token.reducer.ts
+++ b/components/automate-ui/src/app/entities/api-tokens/api-token.reducer.ts
@@ -48,7 +48,7 @@ export function apiTokenEntityReducer(
       return set(
         STATUS,
         EntityStatus.loading,
-        // clear token state to avoid leaking potentially unauthorized data
+        // clear token state to ensure we fetch the latest data
         apiTokenEntityAdapter.removeAll(state)
       ) as ApiTokenEntityState;
     }
@@ -69,7 +69,7 @@ export function apiTokenEntityReducer(
       return set(
         GET_STATUS,
         EntityStatus.loading,
-        // clear token state to avoid leaking potentially unauthorized data
+        // clear token state to ensure we fetch the latest data
         apiTokenEntityAdapter.removeAll(state)
       ) as ApiTokenEntityState;
     }
--- a/components/automate-ui/src/app/entities/users/user.effects.ts
+++ b/components/automate-ui/src/app/entities/users/user.effects.ts
@@ -1,9 +1,10 @@
-import { catchError, mergeMap, map } from 'rxjs/operators';
+import { catchError, mergeMap, map, filter } from 'rxjs/operators';
 import { Injectable } from '@angular/core';
 import { HttpErrorResponse } from '@angular/common/http';
 import { Store } from '@ngrx/store';
 import { Actions, Effect, ofType } from '@ngrx/effects';
 import { of, combineLatest } from 'rxjs';
+import { identity } from 'lodash/fp';
 
 import { NgrxStateAtom } from 'app/ngrx.reducers';
 import { CreateNotification } from 'app/entities/notifications/notification.actions';
@@ -43,7 +44,6 @@ export class UserEffects {
     private store$: Store<NgrxStateAtom>
   ) { }
 
-  // 7/6/18: TODO in follow-up PR: fetch teams for all users - bd
   @Effect()
   getUsers$ = combineLatest([
     this.actions$.pipe(ofType<GetUsers>(UserActionTypes.GET_ALL)),
@@ -65,11 +65,10 @@ export class UserEffects {
       });
     }));
 
-  // TODO rename
   @Effect()
   getUser$ = combineLatest([
     this.actions$.pipe(ofType<GetUser>(UserActionTypes.GET)),
-    this.store$.select(iamMajorVersion)])
+    this.store$.select(iamMajorVersion).pipe(filter(identity))])
     .pipe(
       mergeMap(([action, version]: [GetUser, IAMMajorVersion]) =>
         this.requests.getUser(action.payload.id, version).pipe(
--- a/components/automate-ui/src/app/entities/users/user.effects.ts
+++ b/components/automate-ui/src/app/entities/users/user.effects.ts
@@ -47,7 +47,7 @@ export class UserEffects {
   @Effect()
   getUsers$ = combineLatest([
     this.actions$.pipe(ofType<GetUsers>(UserActionTypes.GET_ALL)),
-    this.store$.select(iamMajorVersion)])
+    this.store$.select(iamMajorVersion).pipe(filter(identity))])
     .pipe(
       mergeMap(([_action, version]: [GetUsers, IAMMajorVersion]) =>
         this.requests.getUsers(version).pipe(
@@ -89,7 +89,7 @@ export class UserEffects {
   @Effect()
   updateUser$ = combineLatest([
     this.actions$.pipe(ofType<UpdateUser>(UserActionTypes.UPDATE)),
-    this.store$.select(iamMajorVersion)])
+    this.store$.select(iamMajorVersion).pipe(filter(identity))])
     .pipe(
       mergeMap(([action, version]: [UpdateUser, IAMMajorVersion]) =>
       this.requests.updateUser(action.payload, version).pipe(
@@ -118,7 +118,7 @@ export class UserEffects {
   @Effect()
   updateSelf$ = combineLatest([
     this.actions$.pipe(ofType<UpdateSelf>(UserActionTypes.UPDATE_SELF)),
-    this.store$.select(iamMajorVersion)])
+    this.store$.select(iamMajorVersion).pipe(filter(identity))])
     .pipe(
       mergeMap(([action, version]: [UpdateSelf, IAMMajorVersion]) =>
       this.requests.updateSelf(action.payload, version).pipe(
@@ -136,7 +136,7 @@ export class UserEffects {
   @Effect()
   deleteUser$ = combineLatest([
     this.actions$.pipe(ofType<DeleteUser>(UserActionTypes.DELETE)),
-    this.store$.select(iamMajorVersion)])
+    this.store$.select(iamMajorVersion).pipe(filter(identity))])
     .pipe(
       mergeMap(([action, version]: [DeleteUser, IAMMajorVersion]) =>
       this.requests.deleteUser(action.payload, version).pipe(
@@ -165,7 +165,7 @@ export class UserEffects {
   @Effect()
   createUser$ = combineLatest([
     this.actions$.pipe(ofType<CreateUser>(UserActionTypes.CREATE)),
-    this.store$.select(iamMajorVersion)])
+    this.store$.select(iamMajorVersion).pipe(filter(identity))])
     .pipe(
       mergeMap(([action, version]: [CreateUser, IAMMajorVersion]) =>
       this.requests.createUser(action.payload, version).pipe(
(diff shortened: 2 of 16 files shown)


components/automate-ui/src/app/pages/api-token/details/api-token-details.component.ts
```diff
@@ -56,20 +56,33 @@ export class ApiTokenDetailsComponent implements OnInit, OnDestroy {
...
+      .subscribe((id: string) => {
+        this.store.dispatch(new GetToken({ id }));
+      });
+
+    combineLatest([
+      this.store.select(getStatus),
+      this.store.select(apiTokenFromRoute)
+    ]).pipe(
+      filter(([gStatus, _]) => gStatus === EntityStatus.loadingSuccess),
       filter(identity),
```
tylercloke: This isn't gonna quite be what you want now. `filter(identity)` is shorthand for:

`filter(x => return x)`

Which in this case is:

```
filter([gStatus, apiTokenFromRoute] => return [gStatus, apiTokenFromRoute])
```

That could create a race condition where the `getStatus` has been defined but `apiTokenFromRoute` is still `undefined` (just ran into that on my end!) and `filter(identify)` will return true there.

You can just do:

```
import { isNil } from 'lodash/fp';
filter(([_, tokenState]) => !isNil(tokenState))
```

bcmdarroch: good find!

components/automate-ui/src/app/entities/api-tokens/api-token.reducer.ts
```diff
@@ -45,7 +45,12 @@ export function apiTokenEntityReducer(
   switch (action.type) {
 
     case ApiTokenActionTypes.GET_ALL: {
-      return set(STATUS, EntityStatus.loading, state) as ApiTokenEntityState;
+      return set(
```
bcmdarroch: figured we could start refreshing the state on every GET

components/automate-ui/src/app/entities/users/user.effects.ts
```diff
@@ -65,11 +65,10 @@ export class UserEffects {
       });
     }));
 
-  // TODO rename
   @Effect()
   getUser$ = combineLatest([
     this.actions$.pipe(ofType<GetUser>(UserActionTypes.GET)),
-    this.store$.select(iamMajorVersion)])
+    this.store$.select(iamMajorVersion).pipe(filter(identity))])
```
bcmdarroch: without this, a `nil` was sneaking into the getUser request initially, resulting in a call to the v1 API by default, followed by a v2 call once `iamMajorVersion` was actually populated

components/automate-ui/src/app/entities/users/user.reducer.spec.ts
```diff
@@ -227,28 +227,51 @@ describe('userStatusEntityReducer', () => {
...
     entities[u.id] = u;
     ids.push(u.id);
   }
-  return {
-    updateStatus: isUpdate ? EntityStatus.loadingSuccess : EntityStatus.notLoaded,
-    status: isUpdate ? EntityStatus.notLoaded : EntityStatus.loadingSuccess,
-    entities,
-    ids
-  };
+  switch (action) {
```
bcmdarroch: probably a way to make this a bit shorter 🤔

components/automate-ui/src/app/pages/user-details/user-details.component.ts
```diff
@@ -177,7 +156,20 @@ export class UserDetailsComponent implements OnInit, OnDestroy {
 
   public deleteUser(): void {
     this.store.dispatch(new DeleteUser(this.user));
-    this.closeDeleteConfirmationModal();
+
+    const pendingDelete = new Subject<boolean>();
+    this.store.pipe(
+      select(deleteStatus),
```
bcmdarroch: added the `deleteStatus` to use in place of the `$done` observable

components/automate-ui/src/app/pages/user-details/user-details.component.ts
```diff
@@ -45,56 +43,41 @@ export class UserDetailsComponent implements OnInit, OnDestroy {
...
-    // TODO (tc) This needs to be refactored to resemble our other patterns
-    // for specific object pages.
-    // combineLatest depends on the user object existing already.
-    this.user = {
-      id: '',
-      name: '',
-      membership_id: ''
-    };
-
-    this.done$ = <Observable<boolean>>combineLatest([
```
bcmdarroch: adding the store reset to the GET actions broke this page, so I refactored it to work a little more like team and token details.

components/automate-ui/src/app/pages/user-details/user-details.component.scss
```diff
@@ -4,6 +4,7 @@
   display: flex;
   justify-content: space-between;
   padding: 0 0 25px 0;
+  height: 115px;
```
bcmdarroch: CSS changes to preserve the page header height even when heading values are `nil`

components/automate-ui/src/app/pages/user-details/user-details.component.html
```diff
@@ -29,29 +29,38 @@
...
           <ng-container *ngIf="!editMode">
-            <chef-button secondary class="edit-button" (click)="setEditMode(true)">
-              <chef-icon>mode_edit</chef-icon>
-              <span>Edit</span>
-            </chef-button>
-            <chef-button *ngIf="isAdminView" secondary caution class="delete-button" (click)="openDeleteConfirmationModal()">
-              <chef-icon>delete</chef-icon>
-              <span>Delete user</span>
-            </chef-button>
+            <app-authorized [allOf]="[['/auth/users/{username}', 'put', user?.id], ['/iam/v2beta/users/{username}', 'put', user?.id],
```
bcmdarroch: the editing capabilities of this page really fall apart if you don't have GET permission, so I decided to add an authorization check around the `Edit` and `Delete` buttons

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -103,6 +97,21 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
+      takeUntil(this.isDestroyed),
+      map(([params, url]) => [params.id as string, url]),
+      // Only fetch if we are on the team details route, otherwise
+      // we'll trigger GetTeam with the wrong input on any route
+      // away to a page that also uses the :id param.
+      filter(([id, url]) => TEAM_DETAILS_ROUTE.test(url) && id !== undefined),
+      // Remove the url because we only need to check if the id has changed
+      map(([id, _url]) => id),
+      distinctUntilChanged()
+    ).subscribe(id => this.store.dispatch(new GetTeam({ id })));
```
bcmdarroch: the GET action dispatch must be the first call on init

lancewf:
I am seeing this problem on this branch. When you delete a user on the Users page the row is not removed until the page is refreshed. I did not see where this change was in the code though. 

delete_user

components/automate-ui/src/app/entities/users/user.reducer.ts
```diff
@@ -64,14 +78,14 @@ export function userEntityReducer(state: UserEntityState = UserEntityInitialStat
           changes: action.payload}, state));
 
     case UserActionTypes.DELETE:
-      return set('status', EntityStatus.loading, state);
+      return set('deleteStatus', EntityStatus.loading, state);
```
msorens: I really like that you added `deleteStatus` here. 🎉

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -124,7 +133,7 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
       takeUntil(this.isDestroyed)
     ).subscribe(([gStatus, uStatus, usersStatus]) => {
       this.isLoadingTeam =
-        (gStatus !== EntityStatus.loadingSuccess) ||
+        (gStatus === EntityStatus.loading) ||
         (uStatus === EntityStatus.loading) ||
         (usersStatus === EntityStatus.loading);
```
msorens: Suggest use the new `allLoadedSuccessfully` function we introduced in Lance's PR (see https://github.com/chef/automate/pull/2151).
Also, would be more consistent with most other files to change `isLoadingTeam` to `loading`.

components/automate-ui/src/app/pages/team-details/team-details.component.ts
```diff
@@ -190,26 +199,6 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
...
-      map(([params, url]) => [params.id as string, url]),
-      // Only fetch if we are on the team details route, otherwise
-      // we'll trigger GetTeam with the wrong input on any route
-      // away to a page that also uses the :id param.
-      filter(([id, url]) => TEAM_DETAILS_ROUTE.test(url) && id !== undefined),
-      // Remove the url because we only need to check if the id has changed
-      map(([id, _url]) => id),
-      distinctUntilChanged()
-    ).subscribe(id => this.store.dispatch(new GetTeam({ id })));
  }
```
msorens: (nit) I think it helps readability within `ngOnInit` to have a blank line after each block (a couple of them already do).

components/automate-ui/src/app/pages/user-details/user-details.component.html
msorens: Thank you for adding that!!

components/automate-ui/src/app/pages/team-details/team-details.component.ts
msorens: Moving this block up as a benefit for human readers is fine, but note that because these are all asynchronous setup calls the order does not matter. If you are seeing some difference that may be hinting at a race condition that could need attention.

components/automate-ui/src/app/pages/user-details/user-details.component.html
```diff
@@ -2,25 +2,25 @@
...
-      [objectName]="user.name"
-      (close)="closeDeleteConfirmationModal()"
-      (deleteClicked)="deleteUser()"
-      objectAction="Delete">
+    [visible]="modalVisible"
+    objectNoun="user"
+    [objectName]="user?.name"
+    (close)="closeDeleteConfirmationModal()"
+    (deleteClicked)="deleteUser()"
+    objectAction="Delete">
```
msorens: Probably an auto-format caused this change; would be preferable to revert to keep the indents for readability, though.

components/automate-ui/src/app/pages/user-details/user-details.component.scss
msorens: Nice--and thank you for the comment!!

components/automate-ui/src/app/pages/user-details/user-details.component.ts
msorens: Excellent!

components/automate-ui/src/app/pages/user-details/user-details.component.ts
```diff
@@ -45,56 +43,41 @@ export class UserDetailsComponent implements OnInit, OnDestroy {
...
+      });
+
     this.route.data.subscribe((data: { isNonAdmin: boolean }) => {
       this.isAdminView = !data.isNonAdmin;
       // Update the forms once this info has come in
       this.createForms(this.fb);
       this.resetForms();
     });
+
+    this.store.select(updateStatus).subscribe(this.handleUpdateStatus.bind(this));
```
msorens: Don't even remember the reason for splitting the setup from the handler, but I am pretty sure that reason is no longer relevant.
It would be cleaner to use our more conventional pattern--and also add in the required `takeUntil`:
```
    this.store.select(updateStatus).pipe(
      takeUntil(this.isDestroyed),
      filter(status => status === EntityStatus.loadingSuccess))
      // same status is used for updating password or full name, so we just reset both
      .subscribe(() => this.resetForms());

```
Then you can delete `handleUpdateStatus` entirely.

msorens:
I think I found another error that may be related to what @lancewf reported: `Users >> Create User >> [x]`, i.e., after opening the "create user" modal attempt to close it with the [x] and it does not close. The "cancel" button, however does work correctly.

components/automate-ui/src/app/entities/api-tokens/api-token.reducer.ts
```diff
@@ -61,7 +66,13 @@ export function apiTokenEntityReducer(
     }
 
     case ApiTokenActionTypes.GET: {
-      return set(GET_STATUS, EntityStatus.loading, state) as ApiTokenEntityState;
+      return set(
+        GET_STATUS,
+        EntityStatus.loading,
+        // clear token state to prevent data leakage
```
blakestier: for this comment and the one above, I'm looking for something around authorization leakage-- in a world without the auth concerns, this wouldn't matter as much. That said, I'm not totally sure on the wording 🤔 
> leaking potentially unauthorized data 
? 🤷‍♀ . Non-blocking.

components/automate-ui/src/app/entities/users/user.reducer.ts
bcmdarroch: it's been a long time coming 😅

components/automate-ui/src/app/pages/team-details/team-details.component.ts
bcmdarroch: trying to scope down the PR so i'll make a note of that for a quick follow-up PR

components/automate-ui/src/app/entities/api-tokens/api-token.reducer.ts
```diff
@@ -45,7 +45,12 @@ export function apiTokenEntityReducer(
   switch (action.type) {
 
     case ApiTokenActionTypes.GET_ALL: {
-      return set(STATUS, EntityStatus.loading, state) as ApiTokenEntityState;
+      return set(
+        STATUS,
+        EntityStatus.loading,
+        // clear token state to avoid leaking potentially unauthorized data
```
tylercloke: I think these comments are a bit misleading. We are only showing data the user is authorized to see via the API. We are working around a quark in our API permissions, not leaking any new data.

components/automate-ui/src/app/pages/api-token/details/api-token-details.component.ts
```diff
@@ -79,31 +91,20 @@ export class ApiTokenDetailsComponent implements OnInit, OnDestroy {
...
-          .forEach(p => {
+        allowedProjects.forEach(p => {
             this.projects[p.id] = { ...p, checked: this.token.projects.includes(p.id)
-            };
-          });
-      }))
-      .subscribe();
+          };
+        });
+      });
```
tylercloke: Nice cleanup!

PR #2134, opened 2019-11-06
products.meta: auth collection depends on cereal

stevendanna:
The auth services depend on cereal.


components/automate-deployment/pkg/services/internal/generated/gen.go
```diff
@@ -375,7 +375,7 @@ var ProductMetadataJSON = `
     }
   ],
   "deleted_packages": [
-    "chef/data-lifecycle-service/0.0.1/20190816162731"
+    "chef/data-lifecycle-service/0.0.1/20191101111721"
```
stevendanna: I guess we forgot to run `make generate` at some point.

PR #2139, opened 2019-11-06
Increase min chef-server version to 12.11

teknofire:

Automate requires at least TLSv1.2 for web connections and Chef-Server 12.11.0 is the first version with OpenSSL 1.0.2 that includes TLSv1.2 support needed for the data-collector proxy to work.

Chef Server 12.11.0 is also needed to use proxying Inspec profiles to Automate `profile['root_url']`

You can configure the data-collector with 12.8.0 but the `profile` config option will throw an error.

- [ ] Tests added/updated?
- [x] Docs added/updated?

### DIFF
--- a/components/automate-chef-io/content/docs/data-collection.md
+++ b/components/automate-chef-io/content/docs/data-collection.md
@@ -106,9 +106,9 @@ To apply the changes, run:
 sudo chef-server-ctl reconfigure
 ```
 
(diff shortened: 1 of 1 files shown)


teknofire:
12.11.0 is also needed as that is when support for proxying Inspec profiles to Automate was added (`profile['root_url']` config).

vjeffrey:
rebasing on master to try to make it green!

PR #2142, opened 2019-11-06
support `chef-automate upgrade run --version VERSION`

stevendanna:
This change allows users to run

       chef-automate upgrade run --version RELEASE

to upgrade to a specific version of Chef Automate. It does not allow
downgrades and does not allow specific version upgrades when in airgap
mode.

### DIFF
--- a/components/automate-cli/cmd/chef-automate/upgrade.go
+++ b/components/automate-cli/cmd/chef-automate/upgrade.go
@@ -221,13 +221,11 @@ func init() {
 		"airgap-bundle",
 		"",
 		"Path to an airgap install bundle")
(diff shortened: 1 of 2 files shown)


vjeffrey:
> does not allow specific version upgrades when in airgap mode

just saw this fly by in slack and was curious...why is it not allowed when in airgap mode?

vjeffrey:
also yay! 🎉 i've been hoping we would enable this. seems very useful

stevendanna:
> does not allow specific version upgrades when in airgap mode

When in airgap mode, the user provides the next airgap bundle to promote to, so passing the version here might be confusing as it would be a no-op.

vjeffrey:
any reason this hasn't merged yet?

PR #2146, opened 2019-11-06
update codeowners

vjeffrey:
just updating codeowners on a few things to ensure we get the correct notifications


phiggins:
While you're in here, it'd be easier to scan for missing things if that list was alphabetized.

Issue #1713, opened 2019-09-25
Control Tag Filtering UI: Accomodate for Key Pair Values without values defined


jonong1972:
## User Story

As a Automate User, I will use this work when I want to filter a control tag when it has no values associated to it.

## Overview
We need to accommodate for Control Tags (keys) that do not have any values within the filter component. 

There may be some logic to be added to the filter component to accommodate for this. To detect if a Control Tag has any values. If it does not, it essentially transforms the filter into a Boolean search to see which items have that tag.

## Demo Script / Repro Steps
1. Goto Compliance Reports page
2. in the filter box select `Control Tags`
3. System will provide suggestions for Control Tags
4. Choose a control tag (for this test, a control tag that does not have any values)
5. System will determine if the control tag has a value or not associated with it.
6.a If the control tag does not: System will then initiate filter and add the tag
6.b if the control tag does: System will then suggest the values.

### Badge for Key Pair Value when no Value is defined
![Screen Shot 2019-09-25 at 11 17 40 AM](https://user-images.githubusercontent.com/1316587/65628391-39369180-df86-11e9-8314-19b5dcfbfa0e.png)

### Badge for Key Pair Value when a Value is defined
![Screen Shot 2019-09-25 at 11 17 46 AM](https://user-images.githubusercontent.com/1316587/65628404-3cca1880-df86-11e9-90c0-4c0c823b8b62.png)

alexpop:
@scottopherson we looked into this today in standup and we have two main ways to accomplish this:

1. Update the suggestions API to provide a special value when a control tag key has no value. Based on this the UI can display the right text to the user. I prefer not to do much magic in the backend because it can get confusing and inconsistent over time.

2. Provide some default suggestions options that are not related to the content. These can be instant and not have to wait for the backend to provide the possible values. For example, one option to filter for a key with no value and second to filter for a key with any value. Or the user selects one of the suggested values the backend returns. This is what the AWS EC2 dashboard is doing when filtering by tags:


<img width="432" alt="Screen Shot 2019-10-24 at 8 19 26 PM" src="https://user-images.githubusercontent.com/107378/67518518-d592a780-f69c-11e9-9f6c-1e1729f76960.png">

<img width="553" alt="Screen Shot 2019-10-24 at 8 20 09 PM" src="https://user-images.githubusercontent.com/107378/67518517-d592a780-f69c-11e9-98c8-793b6d41e8a6.png">

scottopherson:
[...]

PR #2148, opened 2019-11-07
load more compliance reports

vjeffrey:
just wanted to add a studio function to make it easier to load the data in audit_reports. this data is what we use in testing and has control tags, roles, environments -- more diverse data. 
this came up in slack when a developer was trying to work on control tags but didn't have data to populate the control tags with the existing commands in the studio.

https://github.com/chef/automate/issues/1713#issuecomment-547622738

### :+1: Definition of Done
can load reports with control tags easily in the studio

### DIFF
--- a/.studio/compliance-service
+++ b/.studio/compliance-service
@@ -92,7 +92,7 @@ document "load_compliance_reports" <<DOC
   Loads compliance report.
 DOC
 function load_compliance_reports() {
-    cd components/compliance-service/test_data/audit_reports
+    pushd components/compliance-service/test_data/audit_reports &> /dev/null
     ./send_to_data_collector.sh https://a2-dev.test "$(get_admin_token)"
-    cd ../../../../
+    popd &> /dev/null
 }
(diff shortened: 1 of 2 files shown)


vjeffrey:
i have an item on my todo list to make loading data for compliance reports/scan jobs/etc in the studio easier overall. i still want to spend some time cleaning things up to do that - this was just a quick fix to ensure devs aren't blocked

.studio/compliance-service
```diff
@@ -87,3 +87,12 @@ function compliance_configure_purge() {
...
   }"
 }
+
+document "load_compliance_reports" <<DOC
+  Loads compliance report.
+DOC
+function load_compliance_reports() {
+    cd components/compliance-service/test_data/audit_reports
+    ./send_to_data_collector.sh https://a2-dev.test "$(get_admin_token)"
+    cd ../../../../
```
phiggins: `pushd` and `popd` are typically used in automate's shell scripts to avoid all the dots.

vjeffrey: ah interesting, ok ill make the change!

PR #2153, opened 2019-11-07
[automate-2105] Replace Stencil table with Angular table in tok…

msorens:

This is the first step of replacing the StencilJS table component -- `<chef-table>`, aka chef-table.tsx -- with an equivalent Angular component. Ultimately this new Angular component will also be called `<chef-table>` but for the interim -- which should be on the order of a couple weeks -- it is `<chef-table-new>`, as introduced in this PR.

Just as that top-level `<chef-table>` element is temporarily renamed--while the StencilJS and the Angular components co-exist--so are the child elements:

StencilJS  | Angular
-----------|--------
chef-table | chef-table-new
chef-thead | chef-table-header
chef-tbody | chef-table-body
chef-tr    | chef-table-row
chef-th    | chef-table-header-cell
chef-td    | chef-table-cell

We (the Auth team) felt a need to get the token table patched quickly because the rendering is unacceptably slow once you have over 500 or so tokens. At 1000 tokens, for example, it takes 20 to 30 seconds, so it looks like the application has frozen. Coordinating with the UI team, they will manage the changeover to this new Angular component for all other places it is used, confirming that there are no idiosyncrasies popping up.

Here is the outline of that process:
1. Remove the StencilJS component and rebuild the chef-ui-lib.
2. Rename the six Angular elements above back to the names used by StencilJS, which means just update the `selector` decorator in each of those 6 component files--neither the file nor class names need change.
3. Revert the tag names in `components/automate-ui/src/app/pages/api-token/list/api-token-list.component.html`. 
4. Revert the corresponding changes to the test file `e2e/cypress/integration/ui/settings/06_token_mgmt.spec.ts`.
4. All other pages in the application should then just work; each should be reviewed in the running app for any idiosyncrasies.

Note that the control menu (the ellipsis at the end of each row) is still using a StencilJS component, so that is still experiencing the dramatic rendering delay. Work is currently underway to replace that with an Angular component so that will be taken care of very soon.

What changed? Timing diagram 1 shows the "before" -- both table and control menus appeared an interminably long time after the spinner went away.
Diagrams 2 and 3 show "after"--there is a slight difference between a page refresh (where the browser cache is empty) vs. just navigating back to the table once seen before in the data/spinner behavior, but both are quite acceptable. Note that long tail showing that the control menu still lags, though.

image

Finally, here is the what the token list looks like with the StencilJS component (before) and the Angular component (after):
image

NA

### :+1: Definition of Done
Token table looks just as it did before.

1. Rebuild angular-ui
2. Load lots of tokens. (The example command here will generate 1000 tokens with indicies in the range 1001 to 2000.)
```
. scripts/auth_collections.sh
export TOK=your-admin-token-here
export TARGET_HOST=https://a2-dev.test
authGen tokens create 1000 1001
```

### DIFF
--- a/components/automate-ui/src/app/components/table/table-body/table-body.component.scss
+++ b/components/automate-ui/src/app/components/table/table-body/table-body.component.scss
@@ -1,0 +1,5 @@
+@import '../table.component';
+
+:host {
+  display: table-row-group;
+}
--- a/components/automate-ui/src/app/components/table/table-body/table-body.component.ts
+++ b/components/automate-ui/src/app/components/table/table-body/table-body.component.ts
@@ -1,15 +1,10 @@
-import { Component, OnInit } from '@angular/core';
+import { Component, HostBinding } from '@angular/core';
 
 @Component({
   selector: 'chef-table-body',
   templateUrl: './table-body.component.html',
   styleUrls: ['./table-body.component.scss']
 })
-export class TableBodyComponent implements OnInit {
-
-  constructor() { }
-
-  ngOnInit() {
-  }
-
+export class TableBodyComponent {
+  @HostBinding('attr.role') role = 'rowgroup';
 }
--- a/components/automate-ui/src/app/components/table/table-body/table-body.component.ts
+++ b/components/automate-ui/src/app/components/table/table-body/table-body.component.ts
@@ -1,0 +1,15 @@
+import { Component, OnInit } from '@angular/core';
+
+@Component({
+  selector: 'chef-table-body',
+  templateUrl: './table-body.component.html',
+  styleUrls: ['./table-body.component.scss']
+})
+export class TableBodyComponent implements OnInit {
+
+  constructor() { }
+
+  ngOnInit() {
+  }
+
+}
--- a/components/automate-ui/src/app/components/chef-components.module.ts
+++ b/components/automate-ui/src/app/components/chef-components.module.ts
@@ -1,4 +1,5 @@
 // Modules
+import { ChefPipesModule } from '../pipes/chef-pipes.module';
 import { CommonModule } from '@angular/common';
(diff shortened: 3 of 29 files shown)


components/automate-ui/src/app/components/table/table-body/table-body.component.scss
```diff
@@ -0,0 +1,5 @@
+@import '../table.component';
+
+:host {
+  display: table-row-group;
```
msorens: The table* values for display provide the semantics for creating a table without using `<table>`.

components/automate-ui/src/app/components/table/table-body/table-body.component.ts
```diff
@@ -0,0 +1,10 @@
+import { Component, HostBinding } from '@angular/core';
+
+@Component({
+  selector: 'chef-table-body',
+  templateUrl: './table-body.component.html',
+  styleUrls: ['./table-body.component.scss']
+})
+export class TableBodyComponent {
+  @HostBinding('attr.role') role = 'rowgroup';
```
msorens: Similarly, the roles provide accessibility semantics for tables (see https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Table_Role)

SEAjamieD: Thanks for thinking of accessibility here 🥇

SEAjamieD:
I'm unable to load up 1000 tokens, hitting an error saying I "Need jo installed", however I'm able to see the new components rendering properly and the way that they should.  Looks great!

msorens:
@SEAjamieD Oops, I guess I missed pointing out a dependency for that script. Easiest way to get `jo` is `brew install jo`.

PR #2151, opened 2019-11-07
Removing duplicate requests

lancewf:
### :nut_and_bolt: Description: What code changed, and why?

Removing duplicate API requests from the policy and team pages. When loading the pages the observable subscribe functions were being called multiple times for with the same data. This caused the page's data to be requested each time.

### :chains: Related Resources

#2104

### :+1: Definition of Done
All of the auth pages do not perform multiple requests for page data. 

### :athletic_shoe: How to Build and Test the Change
1. `build components/automate-ui && start_all_services`
1. Open the Teams detail page for the admin https://a2-dev.test/settings/teams/admins
1. Open the Chrome inspect tool then select the network tab. 
1. Ensure each API request is only called once.
1. Do the same thing for the team add user page, policy details page, and policy add member page. 

### :white_check_mark: Checklist

- [x] Tests added/updated?
- [ ] Docs added/updated?

### :camera: Screenshots, if applicable

PR #2158, opened 2019-11-07
[AUTOMATE-2138] Always refetch latest data in UI for projects, policies, and roles

tylercloke:

When calls to `GET` and `GET_ALL` happen in the store, clear the store of the object in question and refetch so that we always have the latest data.

Also updated all the subscriptions to filter until the fetch is completed to avoid flickering and other undesirable side effects of hitting the subscribe code multiple times.

Spin up UI. Go to a role, policy, and project from both the list view and by refreshing from the details page. Everything should function correctly.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/pages/policy/details/policy-details.component.html
+++ b/components/automate-ui/src/app/pages/policy/details/policy-details.component.html
@@ -36,15 +36,15 @@
       </section>
     </section>
     <section class="page-body" *ngIf="tabValue === 'members'">
-      <div *ngIf="(members$ | async)?.length === 0" class="empty-state">
+      <div *ngIf="members.length === 0" class="empty-state">
         <p>Add some members to get started!</p>
       </div>
       <chef-toolbar>
-        <div [ngClass]="(members$ | async)?.length === 0 ? 'empty-state' : ''">
+        <div [ngClass]="members.length === 0 ? 'empty-state' : ''">
           <chef-button primary [routerLink]="['/settings', 'policies', policy?.id, 'add-members']">Add Members</chef-button>
         </div>
       </chef-toolbar>
-      <chef-table *ngIf="(members$ | async)?.length > 0">
+      <chef-table *ngIf="members.length > 0">
         <chef-thead>
           <chef-tr>
             <chef-th>ID</chef-th>
@@ -53,7 +53,7 @@
           </chef-tr>
         </chef-thead>
         <chef-tbody>
-          <chef-tr *ngFor="let member of members$ | async">
+          <chef-tr *ngFor="let member of members">
             <chef-td>
               <a *ngIf="memberURLs.hasOwnProperty(member.name)" [routerLink]="memberURLs[member.name]">
                 {{ member.displayName }}
--- a/components/automate-ui/src/app/pages/policy/details/policy-details.component.ts
+++ b/components/automate-ui/src/app/pages/policy/details/policy-details.component.ts
@@ -1,14 +1,15 @@
 import { Component, OnInit, OnDestroy } from '@angular/core';
 import { Router } from '@angular/router';
 import { Store } from '@ngrx/store';
-import { identity } from 'lodash/fp';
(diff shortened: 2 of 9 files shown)


components/automate-ui/src/app/pages/project/details/project-details.component.ts
```diff
@@ -85,28 +95,25 @@ export class ProjectDetailsComponent implements OnInit, OnDestroy {
...
-        });
+    combineLatest([
+      this.store.select(getStatus),
+      this.store.select(getAllRulesForProjectStatus),
+      this.store.select(projectFromRoute),
+      this.store.select(allRules)
+    ]).pipe(
+        filter(([getProjectSt, getRulesSt, _projectState, _allRulesState]) =>
+          getProjectSt === EntityStatus.loadingSuccess &&
+          getRulesSt === EntityStatus.loadingSuccess),
```
msorens: Wondering if this could be split into having a `loading$` Observable in its own `combineLatest` as Lance did in his recent PR, then you could leverage the new `allLoadedSuccessfully` function he is introducing. See https://github.com/chef/automate/pull/2151.
I bring that up because I think it would then make this a bit simpler.

components/automate-ui/src/app/pages/policy/details/policy-details.component.ts
```diff
@@ -70,9 +75,7 @@ export class PolicyDetailsComponent implements OnInit, OnDestroy {
           }
         });
         delete this.policy.members;
-        return members;
-      }));
-    this.members$.subscribe();
```
lancewf: Thank you for cleaning this up. I keep seeing this but thought it was out of the scope of my work.

components/automate-ui/src/app/pages/policy/details/policy-details.component.html
```diff
@@ -36,15 +36,15 @@
       </section>
     </section>
     <section class="page-body" *ngIf="tabValue === 'members'">
-      <div *ngIf="(members$ | async)?.length === 0" class="empty-state">
+      <div *ngIf="members.length === 0" class="empty-state">
```
bcmdarroch: I've been wondering if it's better to use `members` or observable of `members` in the template. Why did you change it here?

components/automate-ui/src/app/pages/policy/details/policy-details.component.ts
```diff
@@ -51,16 +52,20 @@ export class PolicyDetailsComponent implements OnInit, OnDestroy {
...
-    this.members$ = this.store.select(policyFromRoute).pipe(
-      filter(identity),
-      takeUntil(this.isDestroyed),
-      map((state) => {
+    combineLatest([
+      this.store.select(getStatus),
+      this.store.select(policyFromRoute)
+    ]).pipe(
+      filter(([status, _]) => status === EntityStatus.loadingSuccess),
+      filter(([_, state]) => !isNil(state)),
```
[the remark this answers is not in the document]
tylercloke: I think the intention is more clear as separate filters but I don't have a strong opinion

components/automate-ui/src/app/pages/policy/details/policy-details.component.html
tylercloke: In all the other places I've seen we just initialize to an empty array and then populate after the subscribe finishes. Doesn't look to me that the complexity of having it be an obserable gives us anything? Switching it to a plain old array just simplified the templates and made the ts code more standardized.

components/automate-ui/src/app/pages/project/details/project-details.component.ts
tylercloke: I added a `allLoaded` function but it wasn't obvious to me how to easily move `isLoading` to an observable since it's being used in typescript functions. Can make a follow up PR if this is easy to fix.

tylercloke:
@bcmdarroch fixed

PR #2160, opened 2019-11-08
automate-builder-{api, memcached}: enable TLS

stevendanna:
This configures TLS support in the automate-builder-api and
automate-builder-memchaced packages, in line with our other Automate
services.

This change:

- Upgrades builder-api to a version that supports TLS listening and TLS memcache
- Reconfigures automate-builder-api-proxy to connect to automate-builder-api via TLS
- Rebuilds automate-builder-memcached from source with TLS support
- Adds an automate-openssl11 package to facilitate memached TLS support

Known Issues:

- Builder will not support FIPS environments because of OpenSSL 1.1 usage. If this is problematic,
   we can do follow-up work to patch memcached to support OpenSSL 1.0.2.

- Minio still does not require client certificates. This will need to be addressed upstream in both 
   minio and builder-api.  I think we are OK with no client-certificates for now since the access key
   is also required to make requests.

Upstream PRs of Interest:

~https://github.com/habitat-sh/builder/pull/1218~ ✅ 
https://github.com/habitat-sh/core-plans/pull/3111
~https://github.com/habitat-sh/core-plans/pull/3112~ ❌ 
https://github.com/stevendanna/rust-memcache/pull/1
~https://github.com/habitat-sh/builder/pull/1220~ ✅ 


stevendanna:
Fix for session-reuse:  https://github.com/habitat-sh/builder/pull/1231

Ideally we will wait to merge until that fix is merged so we can turn session reuse back on at the lb.

components/automate-openssl11/habitat/plan.sh
```diff
@@ -0,0 +1,133 @@
...
+do_end() {
+  do_default_end
+
+  # Clean up binlinked rm if we made it
+  if [[ $BINLINKED_RM == true ]]; then
+    rm -f /bin/rm
+  fi
+}
+
+# ----------------------------------------------------------------------------
```
jaym: [nit] we probably dont need to keep this

Issue #2161, opened 2019-11-08
Use `chef-table` Angular component for all auth tables


sdelano:
PR #2153 introduced an Angular equivalent of the StencilJS `<chef-table>` and applied it to the token list page. Temporarily, this Angular component uses `<chef-table-new>` as its root-level element, which is necessary until the UI team is ready to examine all the call sites of `<chef-table>` and drop the StencilJS version.

In the meantime, auth team decided to swap in the Angular component for all Auth tables (Token list already complete):

- Policy list
- Project list
- Role list
- User list
- Team list
- Policy member list
- Team users list
- Project rule list

To swap in the Angular component, change the tags in each template per this table:

StencilJS  | Angular
-----------|--------
chef-table | chef-table-new
chef-thead | chef-table-header
chef-tbody | chef-table-body
chef-tr    | chef-table-row
chef-th    | chef-table-header-cell
chef-td    | chef-table-cell

Check the corresponding Cypress test file for the same substitutions.

blakestier:
closed by https://github.com/chef/automate/pull/2196

PR #2196, opened 2019-11-13
[automate-2161] Use angular component chef table

tylercloke:

Replace `chef-table` stencil component with our new angular component on the following pages:

- Policy list
- Project list
- Role list
- User list
- Team list
- Policy member list
- Team users list
- Project rule list

Please click through if you have a sec and make sure everything is up to snuff.

I more or less used sed here with a bit of css simplification.

- [x] Tests added/updated?
- [ ] Docs added/updated?

Should look more or less exactly the same as master. Click through with me to confirm ;)


components/automate-ui/src/app/page-components/user-table/user-table.component.html
```diff
@@ -13,23 +13,23 @@
...
             <a [routerLink]="['/settings', 'users', user.id]">{{ user.name }}</a>
-          </chef-td>
-          <chef-td class="username-column">
+          </chef-table-cell>
+          <chef-table-cell class="username-column">
             {{ user.id }}
-          </chef-td>
-          <chef-td class="garbage-column">
+          </chef-table-cell>
+          <chef-table-cell class="garbage-column">
```
msorens: Would like to suggest changing the CSS class to a more consistent name. 

Policy-details, also in this PR, uses "three-dot-entry". 
Policy-list uses "button-column".
Team-management uses "button-column".
Tokens-list uses "controls".
Projects-list uses "controls".

Maybe standardize on one of those?

components/automate-ui/src/app/pages/policy/details/policy-details.component.scss
```diff
@@ -18,41 +18,10 @@ chef-tab-selector {
...
-
-.type-column {
-  width: 66%;
-  padding-top: 0px;
-  padding-left: 0px;
-}
-
-.three-dot-entry {
-  justify-content: flex-end;
+chef-table-new {
```
msorens: [nit] Nothing wrong with this, but could do away with the outer wrapper. Specifying just `.three-dot-entry` has enough specificity. (And minor benefit, that limits future edits on chef-table to only html files.)

(Same comment for each scss file.)

components/automate-ui/src/app/page-components/user-table/user-table.component.html
susanev: yes please. also is it needed? compliance tables with the control menu don't have special styling for that column

components/automate-ui/src/app/pages/project/details/project-details.component.scss
```diff
@@ -67,42 +67,9 @@ form {
   }
 }
 
-thead {
```
lancewf: Where did all this CSS go?

components/automate-ui/src/app/pages/team-management/team-management.component.scss
```diff
@@ -1,20 +1,8 @@
 @import "~styles/variables";
 
-chef-table {
-  .id-column {
```
bcmdarroch: lost these styles 👀

components/automate-ui/src/app/pages/project/details/project-details.component.scss
tylercloke: The base ([display: table])[https://github.com/chef/automate/blob/master/components/automate-ui/src/app/components/table/table.component.scss#L4] that the new component uses displays things in a very sane way. I removed a ton of css and things look more or less identical. I couldn't notice a difference. We can do a side by side comparison with acceptance or master but I was trying to do that as I went and literally couldn't notice the difference. This is what we did for the tokens table fwiw.

tylercloke: In this specific instance, I think some of this CSS was redundant with parent css. Must have been written before it moved to the parent? I double-checked here since I found the CSS strange but I don't think how the output looks changed at all after removing this CSS. I'll triple-check in the morning though.

components/automate-ui/src/app/page-components/user-table/user-table.component.html
tylercloke: We could make it a default of the table CSS but maybe that's cleaner to do in the three-dot menu itself after we pull it back into angular?

components/automate-ui/src/app/pages/project/details/project-details.component.scss
msorens: Glad to hear your experience @tylercloke was what I was expecting you to find! (Does not always work out that well 🤞 )

susanev:
i totally support the consistent table columns, so i dont mind the change from what was in regards to what @lancewf found

components/automate-ui/src/app/page-components/user-table/user-table.component.html
tylercloke: Sounds like we are gonna look at a general approach to this CSS fix once we get the three-dot menu in, but I will standardize the class names for now.

blakestier:
closes https://github.com/chef/automate/issues/2161

PR #2199, opened 2019-11-13
Wire up full backup/restore for builder

jaym:


components/automate-deployment/pkg/backup/builder_operations.go
```diff
@@ -0,0 +1,253 @@
...
+	objects, _, err := backupCtx.builderBucket.List(ctx, "", false)
+	if err != nil {
+		return err
+	}
+	builderArtifacts, err := ioutil.TempFile("", "builderartifacts")
+	if err != nil {
+		return err
+	}
+	defer builderArtifacts.Close()
+	defer os.Remove(builderArtifacts.Name())
```
jaym: i dont think this needs to be defered

stevendanna: I assume we are writing to a file here (as apposed to a byte.Buffer) because we are worried about keeping the full manifest in memory?

components/automate-deployment/pkg/backup/builder_operations.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	return nil
+}
+
+func (d *BuilderMinioDumpOperation) String() string {
+	return d.Name
+}
+
+// applyBatch falls f with sub-slices of the src slice. The function f
```
jaym: falls -> calls

components/automate-deployment/pkg/backup/builder_operations.go
```diff
@@ -0,0 +1,253 @@
...
+	objVerifier := &NoOpObjectVerifier{}
+	for _, obj := range objects {
+		if err := d.copyObjectFromBuilder(backupCtx, backupBucketPrefix, obj, objVerifier); err != nil {
+			return err
+		}
+		if _, err := fmt.Fprintln(builderArtifacts, obj.Name); err != nil {
+			return err
+		}
+	}
+	// Write file with artifact list
```
stevendanna: Instead of this comments, I suppose we could extract this to a function.  I'm fine with it as is, but if we don't extract it.  Perhaps:

```suggestion
	// Write file with artifact list
	//
	// We first seek to the beginning of the file so io.Copy works as expected.
```

components/automate-deployment/pkg/backup/builder_operations.go
stevendanna:
```suggestion
// applyBatch calls f with sub-slices of the src slice. The function f
```

components/automate-deployment/pkg/backup/builder_operations.go
```diff
@@ -0,0 +1,253 @@
...
+// applyBatch falls f with sub-slices of the src slice. The function f
+// should know what it's doing if it modifies the elements of the
+// passed slice.
+func applyBatch(src []string, batchSize int, f func([]string) error) error {
+	srcLen := len(src)
+	if srcLen < batchSize {
+		return f(src)
+	}
+
+	for i := 0; i < srcLen; i = i + batchSize {
```
jaym: we can probably reduce the number of `i + batchSize` if  we use a while loop

components/automate-deployment/pkg/backup/builder_operations.go
jaym: we could probably replace the with a reader of some sort, but i guess its fine for now

jaym:
i think this is fine for now, we should merge it

PR #2204, opened 2019-11-13
Update member expression regex to include special characters

SEAjamieD:
This branch updates the regex validation used in the member expression modal so that users can use whatever characters they use, with the exception of `:` and `;`

### :+1: Definition of Done
Users can now use special characters.  The use of `*` along with other characters is still disallowed.

2. navigate to https://a2-dev.test/settings/policies/administrator-access/add-members
3. click on `Add Member Expression` on the bottom left
4. set Type => Team, Identity Provider => LDAP
5. in Name => type special characters and see them allowed
6. in Name => add a `*` to any text, see that it is disallowed.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

Screen Shot 2019-11-13 at 3 16 53 PM

### DIFF
--- a/components/automate-ui/src/app/helpers/auth/regex.ts
+++ b/components/automate-ui/src/app/helpers/auth/regex.ts
@@ -15,7 +15,7 @@ export class Regex {
     // Only allows wildcard alone or words and numbers, but not combined
     // Legal Values: *, chef, _state, etc.
     // Illegal Values: abc*, *chef, c*h*e*f, **
-    NO_MIXED_WILDCARD: '^(\\*|[-\\w]+)$'
+    NO_MIXED_WILDCARD: '^(\\*|[-!\\[\\]@#$%^&()+=/?><.,~\\w]+)$'
   };
 
 }
--- a/components/automate-ui/src/app/helpers/auth/regex.ts
+++ b/components/automate-ui/src/app/helpers/auth/regex.ts
@@ -15,7 +15,12 @@ export class Regex {
     // Only allows wildcard alone or words and numbers, but not combined
     // Legal Values: *, chef, _state, etc.
     // Illegal Values: abc*, *chef, c*h*e*f, **
-    NO_MIXED_WILDCARD: '^(\\*|[-!\\[\\]@#$%^&()+=/?><.,~\\w]+)$'
+
+    // Allows no special characters except hyphen
+    NO_MIXED_WILDCARD_ALLOW_HYPHEN: '^(\\*|[-\\w]+)$',
+
+    // Allows all special characters except colon :
+    NO_MIXED_WILDCARD_ALLOW_SPECIAL: '^(\\*|[^:*]+)$'
   };
 
 }
--- a/components/automate-ui/src/app/modules/policy/add-members/policy-add-members.component.ts
+++ b/components/automate-ui/src/app/modules/policy/add-members/policy-add-members.component.ts
@@ -334,15 +334,14 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
         if (formValues.type === 'user' || formValues.type === 'team') {
           this.addIdentityControl();
         } else if (formValues.type === 'token') {
-          this.addNameControl();
+          this.addNameControl(formValues.identityProvider);
         }
         break;
       case 'identityProvider':
         this.resetErrors();
+        this.expressionForm.removeControl('name');
         if (formValues.identityProvider !== matchAllWildCard) {
-          this.addNameControl();
-        } else {
-          this.expressionForm.removeControl('name');
+          this.addNameControl(formValues.identityProvider);
         }
         break;
       case 'name': // fallthrough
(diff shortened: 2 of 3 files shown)


susanev:
@SEAjamieD sorry i wasnt clear i dont think...
- token id, user:local name, team:local name all only allow hyphens
- but user:ldap name, team:ldap name, user:saml name, and team:ldap name all should allow more special characters, specifically we know people use email addresses so we need @ and . at a minimum

components/automate-ui/src/app/helpers/auth/regex.ts
```diff
@@ -15,7 +15,7 @@ export class Regex {
     // Only allows wildcard alone or words and numbers, but not combined
     // Legal Values: *, chef, _state, etc.
     // Illegal Values: abc*, *chef, c*h*e*f, **
-    NO_MIXED_WILDCARD: '^(\\*|[-\\w]+)$'
+    NO_MIXED_WILDCARD: '^(\\*|[-!\\[\\]@#$%^&()+=/?><.,~\\w]+)$'
```
msorens: I had a blind spot last go-around--this validation needs to match just what the back-end validation does:
```suggestion
    NO_MIXED_WILDCARD: '^(\\*|[^:*]+)$'
```
Reference: https://github.com/chef/automate/blob/5b88f9da5daeeb9ec58016a6f9c13f52444703cd/api/interservice/authz/v2/authz.proto#L63

Also, this raises a small UX question, @susanev:
The error pops up to say "Name must be text or a wildcard but not both.".
But in light of this change you might want to revise.
"Name may not contain colons or wildcards"
... except that is a bad message if the user types "**", which is also not allowed, so possibly:
"Use either a single wildcard or a name without colons or wildcards."
That is very precise now, but probably not the greatest wording.

SEAjamieD: Thanks for this @msorens!  Looks like it may be easier to use two different Regex's.  One for LDAP/SAML and the other for token ID and locals.  I'll swap this Regex out for the former.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -334,15 +334,14 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
         if (formValues.type === 'user' || formValues.type === 'team') {
           this.addIdentityControl();
         } else if (formValues.type === 'token') {
-          this.addNameControl();
+          this.addNameControl(formValues.identityProvider);
```
SEAjamieD: In order to know which validation type should be added, we pass in what the identity provider is

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -334,15 +334,14 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
         }
         break;
       case 'identityProvider':
         this.resetErrors();
+        this.expressionForm.removeControl('name');
```
SEAjamieD: We now need to remove the name control, so that we can add a new name control incase it has changed.  I think we're better off just removing and then adding instead of checking to see if we should update it if it already exists.  This will remove some unnecessary logic.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -376,16 +375,27 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+      );
+    } else {
+        this.expressionForm.addControl('name', new FormControl('',
+          [
+            Validators.required,
+            Validators.pattern(Regex.patterns.NO_MIXED_WILDCARD_ALLOW_HYPHEN)
+          ]
+        )
+      );
+    }
```
SEAjamieD: Updated the addNameControl logic to accept the identityProvider.  All we are really doing here is checking to see if it is ldap/saml or not.

I could probably simplify this by interpolating the pattern type... think i'll try that right now actually.

SEAjamieD: scratch that, I think this is better left untouched

components/automate-ui/src/app/modules/policy/add-members/policy-add-members.component.ts
```diff
@@ -376,16 +375,27 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+            Validators.required,
+            Validators.pattern(Regex.patterns.NO_MIXED_WILDCARD_ALLOW_SPECIAL)
+          ]
+        )
+      );
+    } else {
+        this.expressionForm.addControl('name', new FormControl('',
+          [
+            Validators.required,
+            Validators.pattern(Regex.patterns.NO_MIXED_WILDCARD_ALLOW_HYPHEN)
```
bcmdarroch: I'm not sure this regex is getting enforced. I can create local team, user, and token member expressions with special characters (like: `,BDFda&&&!!!___`)

bcmdarroch: actually I just realized we also don't have the right validation on our membership API 😬

SEAjamieD: thanks for looking!

SEAjamieD: @bcmdarroch - I wonder if maybe I had a messed up version pushed up?  Will you give this a shot again.  All patterns are working for me as expected.

Screen Shot 2019-11-26 at 10 13 20 AM
Screen Shot 2019-11-26 at 10 13 57 AM
Screen Shot 2019-11-26 at 10 11 08 AM
Screen Shot 2019-11-26 at 10 10 40 AM

blakestier: Just passing by and realized this is a very odd error message given the input (which is text, but does not contain a wildcard).

SEAjamieD: @susanev - do you have an error message wording request or would you like me to use whatever we can come up with?

@msorens suggested the following above:
>"Use either a single wildcard or a name without colons or wildcards."
>That is very precise now, but probably not the greatest wording.

susanev: let me check with docs folks next week, they have better skills with words than i do.

Shadae:
Can we make a new card to change the text on this card? 

Would love to get this in after @msorens & @bcmdarroch take another look.

susanev:
Sorry, @shadae can you clarify, by change the text do you mean the error message or something else?

Shadae:
Yes, error message wording text. While waiting for input from doc folx 😄

susanev:
i dont mind merging now if we want to do that in a separate pr.

SEAjamieD:
@susanev @Shadae - I'll update the text now with @msorens suggestion so that at the very least the wording is more clear than before.  And then we can merge and open a card to update with whatever the docs team thinks is best.

Issue #2162, opened 2019-11-08
Find Nested Subs and Sub Missing `takeUntil(this.isDestroyed)`


sdelano:
We merged a bunch of PRs to get rid of leaky subscriptions. We want to do one final pass of all auth UI pages to look for:

1) Nested Subscriptions:

```
 this.store.select(someStuff).subscribe((state) => {
         ...
        this.store.select(moreStuff).subscribe((otherState) => {
```

Instead use `combineLatest` with `filter` to ensure that everything is fetched before processing data.

2) Every subscription should have `takeUntil(this.isDestroyed)` (or equivalent that completes the Observable) so they get shut down when the component unloads.

msorens:
_This issue is an investigation. The results of the investigation will probably generate more issues for the parent epic._

Though we still have 3 open PRs that may affect the results for the better, I wanted to get a picture of how big the scope of this issue is, so I merged those 3 PRs then did an audit of all auth pages. The results are in the spreadsheet below.

### Legend

> Green -> OK as is
> Red -> needs updating (either add a `takeUntil`, or unnest subscriptions, depending on which column it is in)
> Yellow -> `takeUntil` is applied... sometimes; we probably want it to apply all the time, but a point worth discussing.

### Summary of the Spreadsheet Data

- 6 subscription missing `takeUntil`
- 6 subscriptions nested
- 16 subscriptions have `takeUntil` conditionally

https://docs.google.com/spreadsheets/d/17n8G693ChnCjz9pfie5Gmj6q6hfs-0v-T7R557tYgzU/edit?usp=sharing

PR #2206, opened 2019-11-14
[automate-2180] Fix policy subscriptions

msorens:
Per the items noted in the [spreadsheet for subscription fixes](https://docs.google.com/spreadsheets/d/17n8G693ChnCjz9pfie5Gmj6q6hfs-0v-T7R557tYgzU/edit#gid=0) this PR cleans up subscriptions for policies:
-- removes nested subscription
-- removes conditional completion on an Observable

#2162 (Find Nested Subs and Subs Not Completed)

### :+1: Definition of Done
Adding members successfully or not successfully does the right thing, as indicated below.

Rebuild automate-ui

Navigate to `Settings >> Policies >> [policy] >> Add Members`

Add a member successfully and the view should auto-navigate back to the members tab of the policy details page.

Add a member unsuccessfully and the add-members should remain open with an error (note the formatting of that error is being reviewed by UX).

One way to force the add-member failure is to add this to components/authz-service/server/v2/policy.go::AddPolicyMembers just after the test for not nil (L329).
```
if err == nil {
		return nil, status.Errorf(codes.InvalidArgument, "parse member: %s", "hello")
	}
```
Then `go_update_component authz-service` in hab studio to compile that change.

### DIFF
--- a/components/automate-ui/src/app/modules/policy/add-members/policy-add-members.component.ts
+++ b/components/automate-ui/src/app/modules/policy/add-members/policy-add-members.component.ts
@@ -1,17 +1,16 @@
 import { Component, OnInit, OnDestroy } from '@angular/core';
 import { FormBuilder, FormGroup, Validators, FormControl } from '@angular/forms';
 import { trigger, transition, style, animate, state, keyframes } from '@angular/animations';
-import { HttpErrorResponse } from '@angular/common/http';
 import { Router } from '@angular/router';
 import { Store } from '@ngrx/store';
-import { identity } from 'lodash/fp';
+import { identity, isNil } from 'lodash/fp';
 import { filter, pluck, takeUntil, distinctUntilChanged, map } from 'rxjs/operators';
 import { combineLatest, Subject, Observable } from 'rxjs';
 
 import { ChefSorters } from 'app/helpers/auth/sorter';
 import { NgrxStateAtom } from 'app/ngrx.reducers';
 import { routeParams } from 'app/route.selectors';
-import { EntityStatus, allLoadedSuccessfully } from 'app/entities/entities';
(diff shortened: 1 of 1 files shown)


components/automate-ui/src/app/entities/entities.ts
```diff
@@ -38,6 +38,10 @@ export function pendingState(state: StateWithStatus): boolean {
   );
 }
 
+export function pending(status: EntityStatus): boolean {
```
msorens: Same function from other (unmerged) PR.

components/automate-ui/src/app/pages/policy/add-members/policy-add-members.component.ts
```diff
@@ -152,6 +151,33 @@ export class PolicyAddMembersComponent implements OnInit, OnDestroy {
...
+      takeUntil(this.isDestroyed),
+      filter(() => this.addingMembers),
+      filter(([addState, error]) => addState === EntityStatus.loadingFailure && !isNil(error)))
+      .subscribe(([_, error]) => {
+        this.addingMembers = false;
+        if (error.message === undefined) {
+          this.addMembersFailed = 'An error occurred while attempting ' +
+            'to add members. Please try again.';
+        } else {
+          this.addMembersFailed = `Failed to add members: ${error.message}`;
```
bcmdarroch: `error.message` returns the top level `message` but that's not the API error message. for that we want the `message` nested under the top-level `error` object. so i would change the `error` variable to `resp` or something then have the string use `resp.error.message`
Screen Shot 2019-11-15 at 11 49 04 AM

tylercloke: Fixed!

PR #1957, opened 2019-10-23
Docs for eas apps tab

kmacgugan:
### :white_check_mark: Checklist

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### :camera: Screenshots, if applicable

PR #2207, opened 2019-11-14
Kg/eas editing

kagarmoe:

Branch for editing

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/automate-chef-io/content/docs/applications.md
```diff
@@ -0,0 +1,162 @@
...
+## Chef Automate + Chef Habitat
+
+The Chef Automate EAS Applications feature allows you to organize and display your applications data from Chef Habitat in an intuitive way.
+
+### Glossary
+
+_Application_ is the program or group of related programs that you are delivering to users.
+
+Chef Habitat provides automation capabilities for building and delivering applications to any environment.
+Chef Habitat defines the application and its dependencies,configuration, management, and behavior in a `plan.sh` file.
```
nellshamrell: Might be worth calling out that it is in either a plan.sh file or a plan.ps1 file - to highlight that it works on both Linux and Windows

components/automate-chef-io/content/docs/applications.md
```diff
@@ -0,0 +1,145 @@
...
+draft = true
+bref = ""
+toc = true
+[menu]
+  [menu.docs]
+    parent = "applications"
+    weight = 30
++++
+
+<!-- Task
```
kagarmoe: @mjingle We're almost there! I need your help revising this content. The "task" section that I've added & commented out is an information design/pattern for describing how to get things done.
Could you try it out on documenting the actions that a user can take on the Applications tab?
In particular, I think we need the context of what the Applications Tab does, and the steps, actions ("commands") and results that users can expect.

components/automate-chef-io/content/docs/eas.md
```diff
@@ -0,0 +1,108 @@
...
+Chef EAS application, infrastructure, and security automation helps your organization manage the lifecycle of your application with stable, consistent, and repeatable processes, letting you focus on delivering new features--and new value--to your customers.
+Chef EAS automates your application's build and release stages, the configuration for its underlying environments and infrastructure, audits both your application and its infrastructure for security and compliance, while also providing a uniform user experience and visibility accross your entire system.
+
+1. Chef Habitat automates how you build and deploy your applications because delivering applications is where you derive your business value
+1. Chef Infra automates how you configure your infrastructure, because your applications also needs to run on a stable and scalable system
+1. Chef Inspec automates how you validate the security of both your applications and your infrastructure
+1. Chef Automate provides visibility into your applications, their infrastructure, and the compliance status for both.
+1. Chef Workflow provides a unified command-line user experience for all of Chef EAS
```
susanev: should this be `Chef Workstation`?

components/automate-chef-io/content/docs/eas.md
```diff
@@ -0,0 +1,108 @@
...
+    example(optional, precedes post-req): Illustrates or supports task
+    postreq: What to do next, he next task or related resources.
+
+## The Chef Enterprise Application (EAS) Stack
+
+The Chef Enterprise Application Stack (EAS) provides the tools for defining and managing your application throughout its entire lifecycle.
+Chef EAS application, infrastructure, and security automation helps your organization manage the lifecycle of your application with stable, consistent, and repeatable processes, letting you focus on delivering new features--and new value--to your customers.
+Chef EAS automates your application's build and release stages, the configuration for its underlying environments and infrastructure, audits both your application and its infrastructure for security and compliance, while also providing a uniform user experience and visibility accross your entire system.
+
+1. Chef Habitat automates how you build and deploy your applications because delivering applications is where you derive your business value
```
susanev: so it kinda jumped out of me that most of these say _automates_ and then automate doesn't .. and I know thats a product name issue.. but since this in the automate docs I wonder if a different word may be better here

components/automate-chef-io/content/docs/eas.md
```diff
@@ -0,0 +1,108 @@
++++
+title = "Chef EAS"
```
susanev: i kinda want this to be `Enterprise Application Stack` in the heading and left nav.. or with Chef if it fits... idk acronyms are difficult

components/automate-chef-io/content/docs/eas.md
```diff
@@ -0,0 +1,108 @@
...
+This results in a lightweight and portable artifact that you can deploy to bare metal or virtual machines, as well as export to immutable formats, such as Docker containers or Kubernetes pods.
+Bundling applications with a complete dependency manifest ensures that your application behaves as consistently on a developer's laptop as it does in its on-prem VM farm staging environment, and it also behaves as consistently there as it does in its cloud-based deployment environment.
+
+## The Chef Automate EAS Application Tab
+
+The Chef Automate EAS Applications feature allows you to organize and display your applications data from Chef Habitat in an intuitive way.
+The Chef Automate EAS Application Tab provides visibility into your application artifacts and channels by letting you see which versions of your packages are running on your system.
+Chef Automate organizes data from the application and environment tags provided by the the Chef Habitat supervisor.
+
+## Prerequisites
```
susanev: im not sure the prerequisites parent section is necessary here.. unless yall are planning to add more content to it. and to clarify, def want the glossary, just not sure of this section and sentence saying to go to the glossary

components/automate-chef-io/content/docs/applications-setup.md
```diff
@@ -0,0 +1,218 @@
++++
+title = "Setting up the Applications Tab"
```
susanev: im kind of not a fan of using `Tab` especially in the left nav and heading
i wonder if this could be just `Applications Setup`

components/automate-chef-io/content/docs/applications-setup.md
```diff
@@ -0,0 +1,218 @@
...
+For more information on setting up Chef Habitat, see the Chef Habitat [Service Groups](https://www.habitat.sh/docs/using-habitat/#service-groups) documentation.
+
+## Setting Up the Applications Tab
+
+To get started, you need:
+
+1. Running installations of Chef Automate and Chef Habitat Builder SasS or Builder on-prem
+1. A [Chef Automate API token]({{< ref "api-tokens#creating-a-standard-api-token" >}}) specifically for this integration
+
+## Configure Chef Habitat to Send Event Data to Chef Automate
```
susanev: maybe `Sending Chef Habitat Data to Chef Automate`

components/automate-chef-io/content/docs/applications-setup.md
susanev: or i think this is more config info, right? so maybe `Configuring Applications`

components/automate-chef-io/content/docs/eas.md
susanev: and if so, chef workstation has an app, its not command-line only.. not sure if that matters

components/automate-chef-io/content/docs/applications-setup.md
kagarmoe: I agree that "tab" in the nav is :yuck:. I'll find a better title.

components/automate-chef-io/content/docs/eas.md
kagarmoe: Good catch!!! Thanks!

components/automate-chef-io/content/docs/eas.md
kagarmoe: Agree. I think I'm going to use the word marks & drop the ordered list, too.

components/automate-chef-io/content/docs/applications-setup.md
kagarmoe: Dashboard!

components/automate-chef-io/content/docs/eas.md
```diff
@@ -0,0 +1,108 @@
...
+For more information about automated deployments, see the Chef Habitat [update-strategy](https://www.habitat.sh/docs/using-habitat/#update-strategy) documentation.
+
+_Service_
+A service is any individual Chef Habitat package that you use in your application.
+For example, if you have a simple Node.js web application, it would depend on Chef Habitat's `core/node` package.
+When you upload your application's build artifact to Builder and promote it to the 'stable' channel, the superisors then deploy it into the 'production' environment.
+That single running instance of Node in your production environment is called a service.
+
+_Service Group_
+A service group is the collection of your application's services, that are all part of the same artifact, tagged with the same channel, and deployed to the same environment. In some cases, very complex applications may have more than one service group.
```
christophermaier: A "service group" is probably best thought of as a kind of handle that lets you can deal with multiple instances of a service running in your "environment" as one thing.

Whether you have one instance of, say, Redis running, or five separate replicas, they would all be running in a single "service group". Commands like `hab config apply` use service group names to target which services get the newly-applied configuration values. This lets you reconfigure those five separate Redis replicas in a single logical operation, rather than having to seek out the individual machines they're running on and apply the configuration to each in turn.

A "service group" also allows you to refer to the collection of services in configuration templates. If you need to know the IP addresses of all those Redis replicas to configure your web app, you get that information by going through the service group.

The service group also defines the domain over which leader elections occur (the leader of those Redis services has to _be_ one of those Redis services, and only those Redis services can vote on which one becomes the leader). I wouldn't draw much attention to that, however, since we're talking about marking that functionality as "beta", because there are some kinks that need to be worked out.

I wouldn't say "tagged with the same channel" for the same reasons I argue that we shouldn't necessarily talk about channels as tags to begin with.

I'd also suggest rewording the "part of the same artifact" bit as well. I suspect the intention is to convey that all the services in a service group are running the same release of that service. "Part of the same artifact" implies some kind of containment relationship that's not really true.

(Also note that while it is the desired and ideal case that all the services in a service group are indeed running the same release of a given service, and that they're all being updated in the same way, from the same channel, and pulling from the same Builder instance, there isn't anything directly in Habitat today that guarantees that. Ideally there would be some "definition" of a Service Group that would exist somewhere "in the network", but that is not currently the case. This is only achieved through consistent management of Supervisors and services, and would probably need to be layered on by Automate or something similar.

If applications are collections of multiple services, then by definition they will _all_ have more than one service group.

components/automate-chef-io/content/docs/eas.md
```diff
@@ -0,0 +1,108 @@
...
+## Prerequisites
+
+The Chef Automate EAS application feature introduces several concepts from Chef Habitat, which are introduced in the following [Glossary]({{< relref "applications.md#Glossary" >}})
+
+### Glossary
+
+_Application_
+An application is a program, or a group of related programs, that you are delivering to users.
+Chef Habitat provides automation capabilities for building and delivering applications to any environment regardless of operating system or platform.
+In Chef Habitat, you define your application and its dependencies, configuration, management, and behavior in a `plan.sh` or `plan.ps1` file.
```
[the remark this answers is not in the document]
kagarmoe: I admit to conflating an application and its services. I think we need to grapple with the question "Is an application the sum of its parts?" If an application is an aggregation of services, and Chef Habitat builds those services, then can we say the Chef Habitat builds an application?

I accept that it is more correct to say that each service in the application has a plan and I'll happily make that change.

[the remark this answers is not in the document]
christophermaier: Until Habitat has a concrete way of representing applications in the way we're talking about them here, I would refrain from saying that "Habitat builds applications".

components/automate-chef-io/content/docs/eas.md
kagarmoe: I suspect that people are slinging around "EAS" without much explanation, so I want the acronym to come up #1 in search. Putting it in the title and the nav shoud do the trick. I can't tweak the search results in Automate.

susanev: Ah! That makes total sense, thanks for explaining the why.

components/automate-chef-io/content/docs/eas.md
kagarmoe: That is a great definition!  
Sadly, I've boiled it down to two lines, but I'm holding on to this for future use.

components/automate-chef-io/content/docs/eas.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Prerequisites
+
+The Chef Automate EAS application feature introduces several concepts from Chef Habitat, which are introduced in the following [Glossary]({{< relref "applications.md#Glossary" >}}).
+
+Find Chef Automate EAS system requirements, installation, and configuration instructions in the [Setting up the Applications Dashboard]({{< relref "applications-setup.md" >}}).
+
+### Glossary
+
+_Application_
```
mjingle: Do we want to put a dash or something to separate the term from the definition? The site just puts the two lines together as in a paragraph otherwise.

components/automate-chef-io/content/docs/applications-setup.md
```diff
@@ -0,0 +1,218 @@
...
+  --event-stream-site="us-west-2" \
+  --event-stream-url="automate.company.com:4222" \
+  --event-stream-token="1234567890abcdefgh"
+```
+
+## Authorize Communication from Chef Habitat to Chef Automate
+
+The Transport Layer Security (TLS) protocol provides privacy and data integrity during communication between Chef Automate and Chef Habitat. Provide Chef Habitat with Chef Automate's TLS certificate to authorize communication between the two.
+
+### Retrieve Chef Automate's TLS key
```
kmacgugan: We should call it the "External TLS Certificate"

components/automate-chef-io/content/docs/applications-setup.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Troubleshooting
+
+### Create a new TLS key pair
+
+You can change Chef Automate's automate front-end TLS key pair by following the [load balancing configuration](https://automate.chef.io/docs/configuration/#load-balancer-certificate-and-private-key) documentation.
+
+### Re-Enable TLS on Chef Automate
+
+In some cases, front-end TLS communication may be previously disabled.
```
kmacgugan: disabled for the event gateway service only. We should disambiguate that this does not mean we disabled TLS for automate, just the habitat event stream.

components/automate-chef-io/content/docs/applications-setup.md
```diff
@@ -0,0 +1,218 @@
...
+```output
+Updating deployment configuration
+
+Applying deployment configuration
+  Started event-gateway
+```
+
+#### Disable TLS on Chef Automate
+
+To disable front-end TLS communication and prevent Chef Automate from receiving event data from Chef Habitat, on the Chef Automate host:
```
kmacgugan: It would not prevent the reception of data, it would mean that the data received is not using TLS to secure it.

kagarmoe: Ah, that's an important difference!

components/automate-chef-io/content/docs/applications-setup.md
```diff
@@ -0,0 +1,218 @@
...
+## Troubleshooting
+
+### Create a new TLS key pair
+
+You can change Chef Automate's automate front-end TLS key pair by following the [load balancing configuration](https://automate.chef.io/docs/configuration/#load-balancer-certificate-and-private-key) documentation.
+
+### Re-Enable TLS on Chef Automate
+
+In some cases, front-end TLS communication may be previously disabled.
+To re-enable front-end TLS communication, which allows Chef Automate to receive event data from Chef Habitat, on the Chef Automate host:
```
kmacgugan: which allows Chef Automate to receive **encrypted** event data (it can still receive data without TLS comms, it is just not encrypted)

components/automate-chef-io/content/docs/applications.md
```diff
@@ -0,0 +1,162 @@
...
+
+Disconnected services are defined services that have not reported back with a health check message for a user-defined period of time.
+
+Use the Chef Automate Service Groups API endpoint to manage the activity of periodic jobs, and to assess disconnection status and removal of services.
+
+### Periodic disconnection checks
+
+#### Health Checks
+Chef Automate assesses the status of Chef Habitat services based on the receipt of a health check message.
+If a node does not send a health check message to Chef Automate after a defined period of time, then Chef Automate marks that node as disconnected.
```
kmacgugan: maybe configurable period of time? instead of defined.

kagarmoe: This is the `threshold` setting?

kagarmoe:
Merged onto #1957

PR #2218, opened 2019-11-14
Update empty page message

blakestier:

This message needed an update to indicate that there might be projects, but you can't see them.

Given that this is the list page, updated the mention of "create" to "view" (opting not to draw attention to the absent Create button)


blakestier:
pulled this into https://github.com/chef/automate/pull/2222

PR #2221, opened 2019-11-14
Refactor profile list code

phiggins:
This code is extracted from another branch I had going with experiments to speed up the available profile list page in the app. The logic to decide which SQL query to run was spread across many functions that felt redundant and convoluted, so I consolidated from 8 functions to 2. This also removes a few redundant log messages.

### DIFF
--- a/components/compliance-service/profiles/db/store.go
+++ b/components/compliance-service/profiles/db/store.go
@@ -400,10 +400,6 @@ func validateSortOrder(sort string, order string) error {
 }
 
 func (s *Store) marketProfilesAll(sort string, order string) (*sql.Rows, error) {
-	err := validateSortOrder(sort, order)
-	if err != nil {
-		return nil, err
-	}
 	selectMarketSQL := fmt.Sprintf(`
 		SELECT sha256, info
 		FROM store_profiles WHERE sha256 IN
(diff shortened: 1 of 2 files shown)


components/compliance-service/profiles/db/store.go
```diff
@@ -367,163 +367,82 @@ func (s *Store) parseList(rows *sql.Rows) ([]inspec.Metadata, error) {
...
+			FROM store_profiles WHERE
+			sha256 IN (SELECT sha256 FROM store_namespace WHERE owner=$1)
+			AND info->>'name' = $2
+			ORDER BY store_profiles.info #>> '{%s}' %s, store_profiles.info #>> '{version}' DESC;
+		`, sort, order)
+
+		rows, err = s.DB.Query(query, namespace, name)
 	}
-
 	defer rows.Close() // nolint: errcheck
```
stevendanna: Is `rows` guaranteed to be non-nil on error?  Should this be below the `if err != nil` check?

phiggins: Good question, I assumed it would be which is why I moved it above the potential return. I will check the docs and see what it says.

phiggins: The docs don't say either way, but it is returning a pointer so it could be null. Also I see this in [the example code](https://golang.org/pkg/database/sql/#DB.Query):

```
        rows, err := db.Query(q, age)
	if err != nil {
		log.Fatal(err)
	}
	defer rows.Close()
```

Potentially not calling `.Close` seems like an ok tradeoff when the alternative is crashing with a null pointer exception. I'll change this.

PR #2224, opened 2019-11-14
[AUTOMATE-2143] Throw error if all cypress environment variables are not set

tylercloke:

- Changed CYPRESS_RUN_FLAKY to 'yes', 'no', from a boolean, because there is no way to tell the difference between false being passed and the env var not being passed at all
- Added a test to assert that the IAM version returned via the API matches what you passed in via CYPRESS_IAM_VERSION
- Also fixes bug where the flaky tests weren't actually running in the flaky pipelines 😓 

I did not do `CYPRESS_BASE_URL` because cypress itself will throw a fit if that isn't set to a valid address.

```
CYPRESS_IAM_VERSION="x" CYPRESS_RUN_FLAKY="x" npm run cypress:open
```

with various missing and incorrect values for the env vars.

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/.expeditor/verify_private.pipeline.yml
+++ b/.expeditor/verify_private.pipeline.yml
@@ -412,7 +412,7 @@ steps:
   #
   - label: ":cypress: (flaky included) IAMv1"
     command:
-      - IAM=v1 integration/run_test integration/tests/cypress_e2e.sh
+      - FLAKY=yes IAM=v1 integration/run_test integration/tests/cypress_e2e.sh
     timeout_in_minutes: 20
     soft_fail: true
     expeditor:
@@ -432,12 +432,10 @@ steps:
       executor:
         linux:
           privileged: true
-          environment:
-            - CYPRESS_RUN_FLAKY=true
 
   - label: ":cypress: (flaky included) IAMv2"
     command:
-      - IAM=v2.1 integration/run_test integration/tests/cypress_e2e.sh
+      - FLAKY=yes IAM=v2.1 integration/run_test integration/tests/cypress_e2e.sh
     timeout_in_minutes: 20
     soft_fail: true
     expeditor:
@@ -445,12 +443,10 @@ steps:
       executor:
         linux:
           privileged: true
-          environment:
-            - CYPRESS_RUN_FLAKY=true
 
   - label: ":cypress: IAMv1"
     command:
-      - IAM=v1 integration/run_test integration/tests/cypress_e2e.sh
+      - FLAKY=no IAM=v1 integration/run_test integration/tests/cypress_e2e.sh
     timeout_in_minutes: 20
     expeditor:
       secrets: &cypress_secrets
@@ -472,7 +468,7 @@ steps:
 
   - label: ":cypress: IAMv2"
     command:
-      - IAM=v2.1 integration/run_test integration/tests/cypress_e2e.sh
+      - FLAKY=no IAM=v2.1 integration/run_test integration/tests/cypress_e2e.sh
     timeout_in_minutes: 25
     expeditor:
       secrets: *cypress_secrets
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -46,11 +46,11 @@ dev or acceptance; typically start with your local box with this:
 export CYPRESS_BASE_URL=https://a2-dev.test
 ```
 
-Most Cypress tests also expect the following IAM-related environment variables. Set them
-according to your targeted Automate instance's IAM version.
+You must also set the IAM version, whether or not run run the flaky tests, and optionally the admin token:
 
 ```bash
 export CYPRESS_IAM_VERSION=v1 || v2.1
+export CYPRESS_RUN_FLAKY=yes || no
 export CYPRESS_ADMIN_TOKEN=<admin token value> # optional
 ```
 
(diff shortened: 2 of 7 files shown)


e2e/README.md
```diff
@@ -46,11 +46,11 @@ dev or acceptance; typically start with your local box with this:
 export CYPRESS_BASE_URL=https://a2-dev.test
 ```
 
-Most Cypress tests also expect the following IAM-related environment variables. Set them
-according to your targeted Automate instance's IAM version.
+You must also set the IAM version, whether or not run run the flaky tests, and optionally the admin token:
 
 ```bash
 export CYPRESS_IAM_VERSION=v1 || v2.1
```
tylercloke: Should we just change this to v2? Now seems like as good of time as any to do that. LMK what y'alls thoughts are here.

e2e/README.md
```diff
@@ -46,11 +46,11 @@ dev or acceptance; typically start with your local box with this:
 export CYPRESS_BASE_URL=https://a2-dev.test
 ```
 
-Most Cypress tests also expect the following IAM-related environment variables. Set them
-according to your targeted Automate instance's IAM version.
+You must also set the IAM version, whether or not run run the flaky tests, and optionally the admin token:
```
lancewf: You have to "run"s. I think you meant "whether or not to run the flaky tests"

e2e/README.md
bcmdarroch: sure, we just have to make sure we update it everywhere (deploy scripts, terraform tags, other places i can't remember off the top of my head)

.expeditor/verify_private.pipeline.yml
```diff
@@ -412,7 +412,7 @@ steps:
   #
   - label: ":cypress: (flaky included) IAMv1"
     command:
-      - IAM=v1 integration/run_test integration/tests/cypress_e2e.sh
+      - FLAKY=yes IAM=v1 integration/run_test integration/tests/cypress_e2e.sh
```
bcmdarroch: `true` instead of `yes`?

e2e/README.md
```diff
@@ -46,11 +46,11 @@ dev or acceptance; typically start with your local box with this:
...
 
 ```bash
 export CYPRESS_IAM_VERSION=v1 || v2.1
+export CYPRESS_RUN_FLAKY=yes || no
```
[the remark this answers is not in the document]
bcmdarroch: I guess because `undefined` would be read as false 💭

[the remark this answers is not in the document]
tylercloke: It's because if you set a bash variable to false you won't be able to detect if the variable was passed at all or not

e2e/cypress/support/index.ts
```diff
@@ -15,19 +15,64 @@
...
 // require('./commands')
 
+const IAMV2 = 'v2.1';
+const IAMV1 = 'v1';
+
 before(function () {
-  if (!Cypress.env('ADMIN_TOKEN')) {
+  if (!Cypress.env('IAM_VERSION')) {
+    // tslint:disable-next-line:no-string-throw
+    throw('MISSING ENVIRONMENT VARIABLE: ' +
```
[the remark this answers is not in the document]
tylercloke: @bcmdarroch cypress added the sarcastic smile, not me. If it's trying to communicate what you are suggesting that was completely lost on me LOL. Lemme give that a try.

e2e/cypress/support/index.ts
```diff
@@ -15,19 +15,64 @@
...
+      `You passed: ${Cypress.env('IAM_VERSION')}`);
+  }
+
+  if (!Cypress.env('RUN_FLAKY')) {
+    // tslint:disable-next-line:no-string-throw
+    throw('MISSING ENVIRONMENT VARIABLE: You must pass CYPRESS_RUN_FLAKY.' +
+      'Must be one of: "yes", "no".');
+  }
+
+  if (Cypress.env('RUN_FLAKY') !== 'yes' && Cypress.env('RUN_FLAKY') !== 'no') {
```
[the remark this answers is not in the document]
tylercloke: I can't do any better for that one. If it's not set, cypress will throw an error before it gets to this code. I added it originally but removed it for that reason.

e2e/README.md
tylercloke: Let's make that a separate change in that case.

e2e/README.md
tylercloke: Too much running

e2e/cypress/support/index.ts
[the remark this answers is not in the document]
tylercloke: Looks much better with `throw new Error`. Good catch! Totally missed that that's what that cypress output was trying to communicate lol. The smilie threw me off:

```
Error: MISSING ENVIRONMENT VARIABLE: You must pass CYPRESS_IAM_VERSION. Must be one of: "v2.1", "v1".

Because this error occurred during a 'before all' hook we are skipping all of the remaining tests.
```

e2e/cypress/support/index.ts
```diff
@@ -15,19 +15,64 @@
...
+      cy.request({
+        auth: { bearer: admin.id_token },
+        url: '/apis/iam/v2beta/policy_version'
+      }).then((response) => {
+        if (Cypress.env('IAM_VERSION') === IAMV2) {
+          expect(response.body.version.major).to.equal('V2');
+        } else {
+          expect(response.body.version.major).to.equal('V1');
+        }
+      });
```
tylercloke: `A test asserts that the version returned from the API matches CYPRESS_IAM_VERSION` This is covered here, correct?

e2e/cypress/support/index.ts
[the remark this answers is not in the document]
bcmdarroch: Yeah not super helpful, cypress 😵

e2e/cypress/support/index.ts
bcmdarroch: oops missed that, my b

PR #2233, opened 2019-11-15
Change builder test to test standalone builder

jaym:


jaym:
we're not going to allow this for now because user management features are tightly coupled to automate and there is no CLI to use them atm

PR #2239, opened 2019-11-16
Update team name on save

blakestier:
Signed-off-by: Brenna Hewer-Darroch <brenna@chef.io>

Before:
image

After:
Screen Shot 2019-11-15 at 4 52 34 PM

### DIFF
--- a/components/automate-ui/src/app/modules/team/team-details/team-details.component.ts
+++ b/components/automate-ui/src/app/modules/team/team-details/team-details.component.ts
@@ -152,7 +152,8 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
     combineLatest([this.isIAMv2$, team$]).pipe(
       takeUntil(this.isDestroyed),
       filter(([_isV2, team]) => team !== undefined),
-      distinctUntilChanged(([isV2, team]) => isV2 === isV2 && team === team)
+      distinctUntilChanged(([isV2A, teamA], [isV2B, teamB]) =>
+        isV2A === isV2B && teamA === teamB)
     ).subscribe(([isV2, team]) => {
       if (isV2) {
         this.teamId = team.id;
--- a/components/automate-ui/src/app/modules/team/team-details/team-details.component.ts
+++ b/components/automate-ui/src/app/modules/team/team-details/team-details.component.ts
@@ -152,8 +152,8 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
     combineLatest([this.isIAMv2$, team$]).pipe(
       takeUntil(this.isDestroyed),
       filter(([_isV2, team]) => team !== undefined),
-      distinctUntilChanged(([isV2A, teamA], [isV2B, teamB]) =>
-        isV2A === isV2B && teamA === teamB)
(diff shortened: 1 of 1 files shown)


components/automate-ui/src/app/modules/team/team-details/team-details.component.ts
```diff
@@ -145,8 +145,7 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
 
     combineLatest([this.isIAMv2$, team$]).pipe(
       takeUntil(this.isDestroyed),
-      filter(([_isV2, team]) => team !== undefined),
-      distinctUntilChanged(([isV2, team]) => isV2 === isV2 && team === team)
```
blakestier: The fix also works if I change `distinctUntilChanged` to `distinctUntilChanged()`

I'm not 100% sure what the function offers over the filter 🤔

lancewf: The `distinctUntilChanged` function tests the previously seen value with the current one. The event only moves on to the subscribe block if the current data is different than the previous data. 

I am not sure why removing the line fixed the problem. I am going to test it locally.

lancewf: I found the problem. I was using the `distinctUntilChanged` function wrong. With the `distinctUntilChanged` you have two parameters (`distinctUntilChanged((current, previous) => test if the current the same as previous)`). I only had it with one parameter. The below code is working for me. 
```
distinctUntilChanged(([isV2A, teamA], [isV2B, teamB]) =>
      isV2A === isV2B && teamA === teamB)
```

blakestier: Does this mean we need to revisit other places where we are using `distinctUntilChanged`, @lancewf ?

lancewf: I don't think we have to because this is the only place where we use arguments. The only reason we are using arguments here is that the value is an array. If the value is just a string then you don't need any arguments.

components/automate-ui/src/app/modules/team/team-details/team-details.component.ts
```diff
@@ -152,7 +152,8 @@ export class TeamDetailsComponent implements OnInit, OnDestroy {
     combineLatest([this.isIAMv2$, team$]).pipe(
       takeUntil(this.isDestroyed),
       filter(([_isV2, team]) => team !== undefined),
-      distinctUntilChanged(([isV2, team]) => isV2 === isV2 && team === team)
+      distinctUntilChanged(([isV2A, teamA], [isV2B, teamB]) =>
```
bcmdarroch: `lastTeam` and `currentTeam` maybe?

blakestier: mmm yessss

blakestier: it's almost like we want `previousIAMVersion` and `previousTeamType`

PR #2257, opened 2019-11-18
Document that `chef-automate backup restore` without ID uses the latest backup

yzl:
A couple of users have mentioned that it would be helpful to be able to restore using the latest backup without having to string together *nix commands to find the correct backup ID. It turns out that running `chef-automate backup restore </path/to/backup/bucket>` with no backup ID will restore using the latest backup. I added documentation to that effect.

Reference: idea A2-I-51 in Aha.

- [ ] Tests added/updated?
- [x] Docs added/updated?


kagarmoe:
Can you also make an issue or PR in https://github.com/habitat-sh/habitat?

PR #2271, opened 2019-11-19
csv export: truncate responses with more than 32k characters

vjeffrey:
small quick one! @teknofire let us know that excel chokes on cells with over 32k characters, which can sometimes happen with unruly controls. this change ensures that we check the character count on result messages and code descriptions for controls, returning a truncated message if it is 32k or more characters.

didn't open an issue for this, so nothing to link to here.

### :+1: Definition of Done
items with more than 32k characters are truncated

i did the manual hacky way of testing this, which is basically just inserting a string that is over 32k characters long into the maxCharLimit function. 

- [x] Tests added/updated?
- [n/a ] Docs added/updated?

### DIFF
--- a/components/compliance-service/reporting/util/csv.go
+++ b/components/compliance-service/reporting/util/csv.go
@@ -86,8 +86,9 @@ func ReportToCSV(report *reportingapi.Report) (string, error) {
 
 func maxCharLimit(item string) string {
 	if utf8.RuneCountInString(item) > 32000 {
-		countToReturn := 32000 - 52
-		return fmt.Sprintf("character limit exceeds 32,000. truncating message: %s", string([]rune(item)[0:countToReturn]))
+		msg := "character limit exceeds 32,000. truncating message: "
+		countToReturn := 32000 - utf8.RuneCountInString(msg)
+		return fmt.Sprintf("%s%s", msg, string([]rune(item)[0:countToReturn]))
 	}
 	return item
 }
--- a/components/compliance-service/reporting/util/csv.go
+++ b/components/compliance-service/reporting/util/csv.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strconv"
 	"time"
+	"unicode/utf8"
 
 	"github.com/golang/protobuf/ptypes"
 	"github.com/sirupsen/logrus"
@@ -67,8 +68,8 @@ func ReportToCSV(report *reportingapi.Report) (string, error) {
 					ControlImpact:         strconv.FormatFloat(float64(control.Impact), 'f', 2, 32),
 					ResultStatus:          result.Status,
 					ResultRunTime:         strconv.FormatFloat(float64(result.RunTime), 'f', 3, 32),
-					ResultCodeDescription: result.CodeDesc,
-					ResultMessage:         result.Message,
+					ResultCodeDescription: maxCharLimit(result.CodeDesc),
+					ResultMessage:         maxCharLimit(result.Message),
 					ResultSkipMessage:     result.SkipMessage,
 				})
 			}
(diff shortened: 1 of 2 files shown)


components/compliance-service/reporting/util/csv.go
```diff
@@ -82,3 +83,10 @@ func ReportToCSV(report *reportingapi.Report) (string, error) {
 	}
 	return content, nil
 }
+
+func maxCharLimit(item string) string {
+	if utf8.RuneCountInString(item) >= 32000 {
+		return fmt.Sprintf("character limit exceeds 32,000. truncating message: %s", item[0:31000])
```
phiggins: Is there a reason this is limiting to 31k? I would've expected this to be 31999 or something.

components/compliance-service/reporting/util/csv_test.go
```diff
@@ -14,6 +16,16 @@ const (
...
+	returnedString := maxCharLimit(my5kstring)
+	assert.Equal(t, my5kstring, returnedString)
+
+	returnedString = maxCharLimit(my5kstring + my5kstring + my5kstring + my5kstring + my5kstring + my5kstring + my5kstring)
+	assert.Equal(t, strings.HasPrefix(returnedString, "character limit exceeds 32,000. truncating message"), true)
+	assert.Equal(t, utf8.RuneCountInString(returnedString) < 32000, true)
+}
```
phiggins: I think it'd be valuable to test the limits directly, by having one test that is exactly the amount allowed and I've that is exactly one character too many.

The strings could also be generated in a loop, that might be easier to work with than just a huge string literal.

components/compliance-service/reporting/util/csv.go
stevendanna: Do we care about non-ASCII data here?

The slice indexing here assuming that all of your characters are 1-byte characters (i.e. restricted to the ASCII range of UTF-8).

I think you can do something like:
````golang
string([]rune(item)[0:310000])
````

 but I'm unsure on the runtime cost of the conversion.  Perhaps something like this might be cheaper, we'd have to test:

```golang
func safeTruncate(s string, size int) string {
	truncated := s[0:size+1]
	for {
		r, _ := utf8.DecodeRune([]byte{truncated[len(truncated)-1]})
		if r == utf8.RuneError {
		   truncated = truncated[0:len(truncated) - 1]
		} else {
		   break
		}

	} 
	return truncated
}
```

Here you can see an example of the malformed character you can get by indexing into a string that might have utf8 data:

https://play.golang.org/p/PRkuh4EeJ4R

I'm also wonder if, since this is a very long string, we might want to play with how we build the truncation message.  That is, I wonder if strings.Builder or some other method might be more efficient than Sprintf with a string that we know is 31k.

stevendanna: I think at most it could be `32,000 - 52` since 52 is the length of the truncation warning.

vjeffrey: not really a reason, i just figured if we're truncating it we don't need to send all that back, but i can totally update 32k-52!

components/compliance-service/reporting/util/csv_test.go
vjeffrey: ya, that makes sense!

components/compliance-service/reporting/util/csv.go
vjeffrey: interesting, thank you!! i'll play around with those two solutions

components/compliance-service/reporting/util/csv.go
```diff
@@ -82,3 +83,11 @@ func ReportToCSV(report *reportingapi.Report) (string, error) {
 	}
 	return content, nil
 }
+
+func maxCharLimit(item string) string {
+	if utf8.RuneCountInString(item) > 32000 {
+		countToReturn := 32000 - 52
+		return fmt.Sprintf("character limit exceeds 32,000. truncating message: %s", string([]rune(item)[0:countToReturn]))
```
phiggins: nitpick: Instead of hardcoding the length of the message, you could have the message in a variable and take its length instead. This would be useful if we ever wanted to tweak the message, etc.

Issue #1958, opened 2019-10-23
Update eas endpoint URLs 


kmacgugan:
## User Story

As a developer, I would like to change the eas endpoint urls to contain /v0 instead of /beta at the point that we prepare eas for GA.

PR #2272, opened 2019-11-19
Move applications/ URL paths from beta->v0

danielsdeleo:

In preparation for a GA release of the EAS applications functionality, the API URL paths are moved from `beta/` to the versioned/stable `v0` paths.

#1958 

* build automate-gateway and -ui; start all services
* the UI should function as before
* `export T=$(get_admin_token)`
* APIs should be accessible with curl via URLs under `api/v0`, e.g.: `curl -k -H "api-token: $T" https://a2-dev.test/api/v0/applications/services`

- [x] Docs added/updated?


danielsdeleo:
Discussed in meatspace that we can merge this one now since it's not changing default visibility.

Issue #2193, opened 2019-11-13
Integration tests for nodemanager resource project tagging


lancewf:
## User Story
As a developer, I will use this work when I want to verify through automated testing that the nodemanager resource ingestion project tagging, project update, and project filtering are working correctly.

Use the Client Runs cypress tests as an example to follow.

## Definition of Done
Automated testing for nodemanager ingestion project tagging, project update, and project filtering for is complete.

## Demo Script / Repro Steps
Run the cypress tests.

PR #2280, opened 2019-11-19
Adding nodemanager project update tests

lancewf:
Adding API level integration tests for node manager project update re-tagging.

#2193

### :+1: Definition of Done
Have a test for each possible attribute for nodemanager project update re-tagging.

Check the buildkite tests

- [x] Tests added/updated?
- [ ] Docs added/updated?


e2e/cypress/integration/api/iam/06_projects_api_compliance_update.ts
```diff
@@ -145,8 +145,6 @@ describeIfIAMV2p1('Compliance report project update tagging', () => {
 
   before(() => {
     cy.cleanupV2IAMObjectsByIDPrefixes(cypressPrefix, ['projects', 'policies']);
-    cy.cleanupV2IAMObjectsByIDPrefixes('test-projects-api-compliance-update',
```
lancewf: This should have been removed.

e2e/cypress/integration/api/iam/08_projects_api_nodemanager_update.ts
```diff
@@ -0,0 +1,288 @@
+import { describeIfIAMV2p1 } from '../../constants';
+import { uuidv4 } from '../../helpers';
+
+describeIfIAMV2p1('Nodemanager project update tagging', () => {
+  const cypressPrefix = 'test-nodemanager-update';
+
+
+  const projectsWithRule = [
+    // This test is commented out because there is a current limit of 6 projects allowed
```
lancewf: I am going to create a separate PR to increase the number of projects allowed and then uncomment all test with a seventh project.

e2e/cypress/integration/api/iam/08_projects_api_nodemanager_update.ts
```diff
@@ -0,0 +1,288 @@
...
+    // apply rules
+    cy.request({
+      headers: { 'api-token': Cypress.env('ADMIN_TOKEN') },
+      method: 'POST',
+      url: '/apis/iam/v2beta/apply-rules'
+    });
+
+    cy.waitUntilApplyRulesNotRunning(100);
+
+    // Ensure the both the compliance and client run nodes' tags are updated
```
bcmdarroch: s/the both/that both

e2e/cypress/integration/api/iam/08_projects_api_nodemanager_update.ts
```diff
@@ -0,0 +1,288 @@
...
+      }).then((response) => {
+        expect(response.body.nodes.length).to.greaterThan(0);
+        expect(nodeExist(complianceNodeId, response.body.nodes)).to.equal(true);
+        expect(nodeExist(clientRunsNodeId, response.body.nodes)).to.equal(true);
+      });
+    });
+  });
+});
+
+function nodeExist(nodeId: string, nodes: any[]): boolean {
```
bcmdarroch: [nit] nodeExists?

e2e/cypress/integration/api/iam/08_projects_api_nodemanager_update.ts
```diff
@@ -0,0 +1,288 @@
...
+    }
+  ];
+
+  before(() => {
+    cy.cleanupV2IAMObjectsByIDPrefixes(cypressPrefix, ['projects', 'policies']);
+  });
+
+  after(() => cy.cleanupV2IAMObjectsByIDPrefixes(cypressPrefix, ['projects', 'policies']));
+
+  it('tagging 7 projects with all the attributes on a compliance and client runs node', () => {
```
bcmdarroch: I'd second reorganizing the test cases into a map. Then we could have more specific test names, like `when a project has a rule that matches one compliance node's and one client runs node's ${attribute} attribute, successfully associates those nodes with that project`

e2e/cypress/integration/api/iam/08_projects_api_nodemanager_update.ts
```diff
@@ -0,0 +1,288 @@
...
+        name: 'project chef server'
+      },
+      rule: {
+        id: 'rule-chef-server',
+        name: 'rule CHEF_SERVER',
+        type: 'NODE',
+        project_id: `${cypressPrefix}-project-chef-server-${Cypress.moment().format('MMDDYYhhmm')}`,
+        conditions: [
+          {
+            attribute: 'CHEF_SERVER',
```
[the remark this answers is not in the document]
lancewf: I had it like that before, but I switched to one test case to limit the number of "apply rules" calls since they take about 30 seconds.

e2e/cypress/integration/api/iam/08_projects_api_nodemanager_update.ts
lancewf: I would like that too, but I could not think of a way to do that and only call "apply rules" once.

bcmdarroch: 🤔lemme ponder that

bcmdarroch: what I ended up doing in [this PR](https://github.com/chef/automate/pull/2322/files#diff-e1765dc0e83555d8980cd1d444bf612cR103) is running most of the actual API calls in the `before` block. then the test themselves just looped over each projectRule to assert the action got tagged correctly. it's a weird structure but I like having more specific test output

lancewf: Yeah, I kind of did the same thing. Take a look and tell me if you think this is better.

Issue #2256, opened 2019-11-18
tech debt, event-gateway: gnatsd dependency is pinned to a branch


vjeffrey:
## User Story

ssd:
```
Anyone up to speed on the events-gateway stuff know how problematic it would be to upgrade our gnatsd dependency to v2?  We are currently pinned to a branch because we were waiting for https://github.com/nats-io/gnatsd/pull/922 to merge. The merged it, but it appears to have only landed in the major version bump.
```

kyleen:
```
I can't remember if it would be problematic, but it would also affect the hab-automate link I think so we would want to test that as well.
```

## Definition of Done
No longer pinned to a branch!

PR #2282, opened 2019-11-19
Embed latest nats-server in event-gw and event-svc

danielsdeleo:

We had some patches to nats-server to support mutual TLS, which have
since been merged. In the meantime, the nats project shipped version 2
(which was very soon after our initial import/vendor) and we moved
automate from the `dep` tool to modules. This patch contains the
necessary changes to use the 2.x nats-server code in the event-gateway
and event-service. End-to-end functionality was verified using the
docker-based hab-client from applications-service's dev/ dir.

#2256

### :+1: Definition of Done

No longer pulling nats server from Dan's branch. ✅ 

* Build event-service and event-gateway
* start all services
* the development data generator still works: run `applications_populate_database` and see the data in the UI
* Works e2e with a real hab client: follow the instructions here: https://github.com/chef/automate/blob/master/components/applications-service/README.md


stevendanna:
@danielsdeleo Thanks for looking at this!  It looks like we are still on a version of `nats-io/nats-streaming-server` that depends on `nats-io/gnatsd` which I think is why this PR is all additions and still includes nats-io/gnatsd in the vendor dir.  I wonder if we could also upgrade to a nats-streaming-server to a version past this commit: https://github.com/nats-io/nats-streaming-server/commit/ff87d8570a8cf74151b76fec5fe90e9b7dd11315

danielsdeleo:
@stevendanna oh nice catch 💯 though that did explode the number of files affected by the revendor for whatever reason 😬

go.mod
```diff
@@ -54,36 +53,31 @@ require (
...
-	github.com/hashicorp/raft v1.0.0 // indirect
 	github.com/iancoleman/strcase v0.0.0-20190422225806-e506e3ef7365 // indirect
 	github.com/jaswdr/faker v1.0.2
 	github.com/jhump/protoreflect v1.5.0
 	github.com/jonboulle/clockwork v0.1.0 // indirect
 	github.com/konsorten/go-windows-terminal-sequences v1.0.2 // indirect
 	github.com/kylelemons/godebug v0.0.0-20170820004349-d65d576e9348
 	github.com/leanovate/gopter v0.2.4
-	github.com/lib/pq v0.0.0-20171126050459-83612a56d3dd
+	github.com/lib/pq v1.2.0
```
stevendanna: ⚠️ We've seen problems with this upgrade that we haven't been able to dig into for yet: https://github.com/chef/automate/pull/2115

danielsdeleo: @stevendanna so I removed this change to the pq version from this PR, but I am still seeing the same build kite failures in this PR as are occurring for the other (#2115) PR. Unless there's some build cache that's keeping bad builds around, it seems like there might be some other culprit for the failures. But I'm at a loss to explain it.

stevendanna: @danielsdeleo It looks like `lib/pq` files are still being updated in this PR.

danielsdeleo: 😬 indeed. I was looking only at `go.mod`, which I guess I was able to make out-of-sync with the actual content in `vendor/`. Do you have an idea of how we might detect that case? Seems like a thing we could put in the repo health ci job. `go mod verify` doesn't seem to solve the problem.

danielsdeleo: K, so after flailing at this a bit I finally learned that newer versions of nats-streaming-server have lib/pq 1.2.0 in their go.mod file and older versions of nats-streaming-server won't work with the renamed nats.go (formerly called go-nats) library. There are some recent-ish versions of nats-streaming-server that require lib/pq 1.1.1, but that version seems to also have the same behavior as 1.2.0.

stevendanna: > go mod verify doesn't seem to solve the problem.

Indeed it does not.  There is an open issue upstream:

https://github.com/golang/go/issues/27348

I think the only alternative is to run `make revendor` and see if there is a diff.  That might be a bit of a bummer as it means re-downloading everything...

> There are some recent-ish versions of nats-streaming-server that require lib/pq 1.1.1, but that version seems to also have the same behavior as 1.2.0.

OK.  We had delayed looking into this, but we can perhaps move that up since it is blocking other work.  My current theory is that the issue is here:

https://github.com/lib/pq/commit/6e2a335bdf9878301a663728f1b8b6483848beed

as I _think_ that results in certain types of errors that per previous retried not being retried now which is hitting us when our pg connections hit the idle timeout in pg-gateway.

stevendanna: > go mod verify doesn't seem to solve the problem.

Indeed it does not.  There is an open issue upstream:

https://github.com/golang/go/issues/27348

I think the only alternative is to run `make revendor` and see if there is a diff.  That might be a bit of a bummer as it means re-downloading everything...

> There are some recent-ish versions of nats-streaming-server that require lib/pq 1.1.1, but that version seems to also have the same behavior as 1.2.0.

OK.  We had delayed looking into this, but we can perhaps move that up since it is blocking other work.  My current theory is that the issue is here:

https://github.com/lib/pq/commit/6e2a335bdf9878301a663728f1b8b6483848beed

as I _think_ that results in certain types of errors that per previous retried not being retried now which is hitting us when our pg connections hit the idle timeout in pg-gateway.

vjeffrey:
@danielsdeleo should this pr be closed?  or open some other card to track the stuff that's blocking this?

danielsdeleo:
replaced with #2874

PR #2283, opened 2019-11-20
Jamie/control menu angular ui2

SEAjamieD:
An ongoing effort to remove stencil components.  The goal of this branch was to remove `chef-control-menu` from the codebase.  It is replaced here with an Angular Material component, specifically `mat-select` and `mat-option`.  In order to remain keyboard accessible, onClick functions within `mat-option` have been changed to onSelectionChange per angular material requirement.

https://github.com/chef/automate/issues/2061

### :+1: Definition of Done
All `chef-control-menus` have been replaced, slight styling updates to what was requested by UX. 

All Control menus are keyboard accessible.

build components/automate_ui_devproxy && start_all_services

Across the site, control menus can be found in:
Settings >
 notifications
 policies
 projects
 API Tokens
 Teams
 Users
 Node Credentials
 Node Integrations

Compliance >
  Reports > Nodes Tab
  Reports > Profiles Tab
  Reports > Controls Tab

- [x] Tests added/updated?
- [ ] Docs added/updated?


components/automate-ui/src/styles.scss
```diff
@@ -196,3 +196,49 @@ chef-loading-spinner[fixed] {
 input {
   line-height: normal;
 }
+
```
SEAjamieD: Angular Material requires a lot of over rides, but doesn't really seem to have a good place to target globally outside of the styles.css file.  Any suggestions on a better location for something like this.  Will get crowded after adding several more components.

SEAjamieD:
@susanev - from a UX/style perspective, this ticket is done.  I'm still trying to work through some Cypress testing issues, but I feel comfortable opening up this PR for any style and/or user 
 related issues you might find.

SEAjamieD:
@susanev - Major apologies! I failed to update to new spec pictured here.  On that now and will retag when done
Screen Shot 2019-11-25 at 3 20 59 PM

SEAjamieD:
@susanev - Ok, new colors and states are in.  Sorry about that!

components/automate-ui/src/styles.scss
```diff
@@ -196,3 +196,68 @@ chef-loading-spinner[fixed] {
...
+    &:after {
+      content: '...';
+      position: absolute;
+      color: black;
+      bottom: 33%;
+      left: 30%;
+    }
+
+    &:hover {
+      background-color: #DEE5FD;
```
SEAjamieD: @susanev several of these colors here are ones that I haven't seen used in the codebase yet.  Are there names you want to use for them when I make the variables?

susanev:
@SEAjamieD can we increase the icon size? maybe `font-size: 20px;` or similar. also can we get the icon centered in the button?

components/automate-ui/src/app/modules/policy/details/policy-details.component.html
```diff
@@ -65,9 +65,9 @@
                 {{ member.displayType }}
               </chef-table-cell>
               <chef-table-cell class="three-dot-column">
-                <chef-control-menu>
-                  <chef-option (click)="removeMember(member)">Remove Member</chef-option>
-                </chef-control-menu>
+                <mat-select panelClass="chef-control-menu-x">
+                  <mat-option (onSelectionChange)="removeMember($event, member)">Remove Member</mat-option>
```
SEAjamieD: mat-select requires the use of `onSelectionChange` in order to be keyboard accessible.  Unfortunately this will cause a double fire of connected function -> one for the selection and one for the deselection.  The current workaround is running an if check for whether on not this was user input using $event.isUserInput.  see: https://github.com/angular/components/issues/4094#issuecomment-363746964

components/automate-ui/src/app/modules/policy/details/policy-details.component.ts
```diff
@@ -20,6 +20,10 @@ import {
 
 export type PolicyTabName = 'definition' | 'members';
 
+export interface KeyboardEvent {
+  isUserInput: boolean;
+}
+
```
SEAjamieD: By default - isUserInput is not part of the KeyboardEvent interface - this extends it.

components/automate-ui/src/app/modules/policy/details/policy-details.component.ts
```diff
@@ -103,11 +107,13 @@ export class PolicyDetailsComponent implements OnInit, OnDestroy {
...
-      id: this.policy.id,
-      members: [member]
-    }));
+  removeMember($event: KeyboardEvent, member: Member): void {
+    if ($event.isUserInput) {
+      this.store.dispatch(new RemovePolicyMembers(<PolicyMembersMgmtPayload>{
+        id: this.policy.id,
+        members: [member]
+      }));
+    }
```
SEAjamieD: An example of using the `$event.isUserInput` to only get the selection and not the deselection.

components/automate-ui/src/app/modules/token/token-list/api-token-list.component.html
```diff
@@ -66,18 +66,18 @@
                 <ng-container *ngIf="token.active">Active</ng-container>
                 <ng-container *ngIf="!token.active">Inactive</ng-container>
               </chef-table-cell>
-              <chef-table-cell class="three-dot-column">
-                <chef-control-menu id="menu-{{token.id}}" data-cy="token-control">
-                  <chef-option (click)="notifyCopy()" data-cy="copy-token">
+              <chef-table-cell class="controls">
+                <mat-select panelClass="chef-control-menu-x" id="menu-{{token.id}}" data-cy="token-control">
+                  <mat-option (onSelectionChange)="notifyCopy($event)" data-cy="copy-token">
```
SEAjamieD: Again - instances of passing in the event to check agains `isUserInput`.

components/automate-ui/src/app/modules/token/token-list/api-token-list.component.ts
```diff
@@ -28,6 +28,10 @@ import { assignableProjects } from 'app/services/projects-filter/projects-filter
 import { Project, ProjectConstants } from 'app/entities/projects/project.model';
 import { ProjectsFilterOption } from 'app/services/projects-filter/projects-filter.reducer';
 
+export interface KeyboardEvent {
+  isUserInput: boolean;
+}
```
SEAjamieD: Same as above - extending Keyboard Event

SEAjamieD:
@susanev 
> also, we need to be able to tab to these buttons, and open them and select an option with keyboard only. you can tab to them, but there's no visible focus (maybe cause you set the outline to 0?).. and you cant open the menu with ur keyboard.

>flow should be..

>tab to item
space bar to open
arrow keys to navigate in the menu
enter to select
and if you tab off the menu should close.

How set are you on this flow?  With angular material we get most of this built in, only change is that instead of using space bar to open, enter opens.  Is this acceptable?

susanev:
@seajamied that's fine, thanks for checking!!

susanev:
@SEAjamieD it looks and works well.. a few more things
- for some reason, the height of a row with a control menu increases from `55px` to `76px`, since we will have some tables with rows with and without control menus we really want those to be the same height, any way to keep the row with a control menu at the same height as rows without?
- the menu style when opened, and with focus on the menu items looks a bit different than what these were originally designed to look like..
    - font appears to be roboto but should be muli like the rest of automate
    - on open, the first item has the gray background and im not sure why... would prefer no background color until something gets focus or hover
    - on hover/focus can we change the background color to our chef-primary-dark and change the font color to white? (i believe these were the styles before converting to angular material)

components/automate-ui/src/app/modules/policy/details/policy-details.component.ts
SEAjamieD: `isUserInput` is provided by Angular Material.  Every time an option is selected there is an event fired for selected and an event fired for deselected.  The selected event passed through `isUserInput = true` and the deselection event passes through `isUserInput = false`;

msorens:
I assume you are still working on this, but a couple other items I noticed:
1. roles-list.component.html (and roles-list.component.spec.ts) are still using `chef-control-menu`, so those should be removed.
2. Should also delete the Stencil component (chef-ui-library/src/molecules/chef-control-menu).

SEAjamieD:
@susanev 
>on open, the first item has the gray background and im not sure why... would prefer no background color until something gets focus or hover

by default, the first item in the panel is focused upon opening, which is why the first item is dark.  I've left for accessibility purposes and recolored.  If this is not ideal, flag it and figure out how to adjust 👍

components/automate-ui/src/app/pages/integrations/list/integrations-list.component.ts
```diff
@@ -47,8 +48,10 @@ export class IntegrationsListComponent {
     return status === 'loading';
   }
 
-  handleDelete(id) {
-    this.store.dispatch(new DeleteManager({id}));
+  handleDelete($event: ChefKeyboardEvent, id) {
```
msorens:
```suggestion
  handleDelete($event: ChefKeyboardEvent, id: string) {
```

components/automate-ui/src/app/pages/project/list/project-list.component.spec.ts
```diff
@@ -87,7 +87,8 @@ describe('ProjectListComponent', () => {
           inputs: ['visible', 'applyRulesStatus', 'stopRulesInProgress'],
           outputs: ['confirm', 'cancel']
         }),
-        MockComponent({ selector: 'chef-control-menu' }),
```
msorens: Also need to delete L95
```
MockComponent({ selector: 'chef-option' }),
```

components/automate-ui/src/styles.scss
```diff
@@ -234,3 +234,95 @@ input {
...
+// Angular Material Overrides
+.chef-control-menu-x {
+  position: absolute;
+  margin-top: 35px;
+  right: -48px;
+  margin-right: 35px;
+}
+
+.three-dot-column {
+  position: relative;
```
msorens: Ah, nice! Since you have this, could you please remove the definition for `three-dot-column` occurring in all the other scss files (8 files). All of them are identical, containing a single line: `text-align: right;`. See if that is still needed; if so, should be able to add it here.

components/automate-ui/src/styles.scss
```diff
@@ -234,3 +234,95 @@ input {
...
+
+  &:focus .mat-select-trigger {
+    outline: -webkit-focus-ring-color auto 5px;
+  }
+
+  .mat-select-trigger {
+    display: inline-block;
+    position: absolute;
+    right: 10px;
+    top: calc(50% - 16px); // subtracting half the height
```
msorens: Yes! Thank you for the comment to make it maintainable! 👍

components/automate-ui/src/styles.scss
```diff
@@ -234,3 +234,95 @@ input {
...
+    &:hover {
+      background-color: var(--chef-badge-primary-background);
+
+      &:after {
+        color: var(--chef-badge-primary-color);
+      }
+    }
+  }
+
+  &[aria-owns] {  // hacky alternative to active
```
msorens: The comment gives me only a vague idea of justification; might you have a URL reference you could add here?

components/chef-ui-library/src/global/variables.css
```diff
@@ -27,6 +27,7 @@
   --blue-de-france: #3DA5FF;
   --gray: #B7BCBC;
   --blue: #174AF0;
+  --palatinate-blue: #0E3DD5;
```
msorens: Wondering what your source is for this color?
Every reference I can find (e.g. https://encycolorpedia.com/search?q=palatinate+blue) shows palatinate blue as `#273be2`

components/automate-ui/src/app/pages/project/list/project-list.component.html
```diff
@@ -46,9 +46,9 @@
                   <chef-table-cell>{{ statusLabel[project.status] }}</chef-table-cell>
                   <chef-table-cell class="three-dot-column">
                     <!-- <app-authorized [allOf]="['/iam/v2/projects/{id}', 'delete', project.id]"> -->
-                      <chef-control-menu>
-                        <chef-option (click)="startProjectDelete(project)" data-cy="delete-project">Delete Project</chef-option>
-                      </chef-control-menu>
+                      <mat-select panelClass="chef-control-menu-x">
+                        <mat-option (onSelectionChange)="startProjectDelete(project)" data-cy="delete-project">Delete Project</mat-option>
```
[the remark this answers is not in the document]
msorens: Per our conversation, please add on this method and check that there are no other missing ones. Thanks!

components/automate-ui/src/app/pages/integrations/list/integrations-list.component.html
```diff
@@ -54,10 +54,10 @@ <h2 class="display4 strong">Cloud Platform Providers</h2>
             <chef-td class="column-date">{{ manager.date_added | datetime: RFC2822 }}</chef-td>
             <chef-td>{{ manager.status }}</chef-td>
             <chef-td class="controls">
```
msorens: Not your doing, but would you mind cleaning up this errant item to use `three-dot-column` class to make consistency complete?

components/automate-ui/src/app/modules/policy/list/policy-list.component.html
```diff
@@ -30,9 +30,9 @@
                       <chef-table-cell class="status-column">{{ status(policy) | capitalize }}</chef-table-cell>
                       <chef-table-cell class="three-dot-column">
                       <!-- <app-authorized [allOf]="[`/iam/v2/policies/{id}`, 'delete', policy.id]"> -->
-                        <chef-control-menu *ngIf="policy.type !== 'CHEF_MANAGED'">
-                          <chef-option (click)="startPolicyDelete(policy)">Delete Policy</chef-option>
-                        </chef-control-menu>
+                        <mat-select panelClass="chef-control-menu-x" *ngIf="policy.type !== 'CHEF_MANAGED'">
```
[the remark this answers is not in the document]
SEAjamieD: at the time it was about differentiating while building.  For the time being, i'll push through as `chef-control-menu-x` so that we don't delay, but i'll add to the list of cleanup items to be addressed.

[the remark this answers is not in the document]
SEAjamieD: actually - quick search and replace should handle this now.

components/automate-ui/src/app/pages/project/list/project-list.component.spec.ts
SEAjamieD: great catch!

components/automate-ui/src/styles.scss
SEAjamieD: FANTASTIC NOTE.  THANK YOU!  Deleting all that repetition was great

components/chef-ui-library/src/global/variables.css
SEAjamieD: I got if from the Automate Extended Color Palette.

components/automate-ui/src/styles.scss
SEAjamieD: I've updated the comment so it makes a little more sense.  Ultimately, I noticed "aria-owns" being applied and hooked in on it.  

`aria-owns` by definition tells assistive technology that another element being placed on the screen should be treated as a child element of the current element we're interacting with (https://developers.google.com/web/fundamentals/accessibility/semantics-aria/aria-labels-and-relationships#aria-owns)

I'm ultimately utilizing it out of scope because when `aria-owns` is added, we know that the dropdown panel is open.

susanev:
@SEAjamieD you're going to remove that .toml right?

SEAjamieD:
@susanev 
> @SEAjamieD you're going to remove that .toml right?

sorry thank you!

susanev:
thanks for your attention to detail @bcmdarroch!!
we are aware of the focus weirdness and we decided to move forward with it, but hoping that when the ui team has a chance to think more about how we are using material then we can revisit it then.

components/automate-ui/src/app/modules/policy/list/policy-list.component.ts
```diff
@@ -60,9 +61,11 @@ export class PolicyListComponent implements OnInit, OnDestroy {
     this.deleteModalVisible = false;
   }
 
-  public startPolicyDelete(policy: Policy): void {
-    this.deleteModalVisible = true;
-    this.policyToDelete = policy;
+  public startPolicyDelete($event: ChefKeyboardEvent, policy: Policy): void {
+    if ($event.isUserInput) {
```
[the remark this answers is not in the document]
SEAjamieD: This is effect of using Material UI's `mat-select` component.  When a selection is made, it fires off two events instead of just one.  One of the events is the item being selected, and the other event is something being DEselected.

Unfortunately, without this gate being added, we'd fire off two events.  The workaround is to check for the `isUserInput` property being true, if it is - that's the selected event we want.

PR #2291, opened 2019-11-20
Deploy all the products with terraform

jaym:
We're not deleting the old workflow one just yet because it currently
requires manually hacking up the terraform state file

### DIFF
--- a/terraform/test-environments/scenarios/single_local_all_inplace_upgrade.tf
+++ b/terraform/test-environments/scenarios/single_local_all_inplace_upgrade.tf
@@ -1,18 +1,18 @@
 #
-# A simple local install of A2 with Workflow using the chef-automate CLI. It currently uses a pseudo-inplace upgrade.
+# A simple local install of A2 with all supported products using the chef-automate CLI. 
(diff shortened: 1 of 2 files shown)


terraform/test-environments/scenarios/single_local_all_inplace_upgrade.tf
```diff
@@ -0,0 +1,65 @@
...
+  # Metadata
+  meta_title       = "Single Local (Inplace Upgrade) with all supported products"
+  meta_description = "A2 stack with all supported products (using SAML) deployed locally as Habitat packages on a single host using the chef-automate CLI."
+  meta_type        = "habitat"
+
+  # AWS Instance Configuration
+  vpc            = "${var.environment}"
+  platform       = "ubuntu-16.04"
+  key_name       = "cd-infrastructure"
+  instance_type  = "m4.large"
```
tduffield: we should use m5.large.

jaym: sorry, i didn't see this before i merged. is it still safe to change?

PR #2293, opened 2019-11-20
deployment: set TasksMax=infinity in unit file

stevendanna:
Users on SLES have seen Automate fails to start with with:

    Error: Failed to start service: Failed to load service chef/automate-load-balancer/0.1.0/20191031214341
    Output:
    thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }', libcore/result.rs:1009:5
    note: Run with `RUST_BACKTRACE=1` for a backtrace.
    : exit status 101
    Deploy failed
    DeployError: Unable to install, configure and start the service: deploy failed

and with the following in `dmesg`:

   [ 814.796021] cgroup: fork rejected by pids controller in
   /system.slice/chef-automate.service

Support tracked this down to the fact that SLES ships a low
DefaultMaxTasks. Setting it to infinity in our unit file should leave
us unconstrained by their default.

### DIFF
--- a/components/automate-deployment/pkg/target/local_target.go
+++ b/components/automate-deployment/pkg/target/local_target.go
@@ -81,6 +81,7 @@ Restart = on-failure
 LimitNOFILE = 65536
 LimitMEMLOCK=infinity
 UMask = 0022
+TasksMax = infinity
 Environment = "HAB_LICENSE=accept-no-persist"
 Environment = "HAB_SUP_BINARY=%s"
 Environment = "HAB_LAUNCH_BINARY=%s"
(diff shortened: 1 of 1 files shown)


components/automate-deployment/pkg/target/local_target.go
```diff
@@ -81,6 +81,7 @@ Restart = on-failure
 LimitNOFILE = 65536
 LimitMEMLOCK=infinity
 UMask = 0022
+TasksMax = infinity
```
ryancragun: Presumably this limit exists for a reason. Are we sure we want to set it to unlimited instead of a considered limit?

stevendanna: This is a good question.  According to https://www.freedesktop.org/software/systemd/man/systemd-system.conf.html:

> Defaults to 15%, which equals 4915 with the kernel's defaults on the host, but might be smaller in OS containers.

Given that we haven't seen this problem for non SLES users, we could set it to this, or some "round" number like `8192` to give us plenty of headroom.

One "unknown" bit here is that the new Builder component uses MANY threads in its default configuration.

PR #2306, opened 2019-11-20
Remove all applied search query filters when user clicks `clear all` button

SEAjamieD:
When attempting to remove all filters using the `clear all` button on the Compliance page, the function would not remove filters that include the `control_tag` filter.  This is because `control_tag`s have 3rd level filter available to them which incorporates a `:` delimiter and the initial function doesn't know how to handle them.

Since we are clearing all filters, it would make sense to use an empty object, instead of looping through all the filters.

### :+1: Definition of Done
the `clear all` button now removes all filters applied on compliance pages.

2. you may need to run `load_compliance_reports` in hab studio to have reports to filter by
3. Navigate to https://a2-dev.test/compliance/reports/overview
4. filter by anything involving `control tag`
5. filter by anything else other than `control tag`
6. click `clear all`
7. see them both go away and URL query should be clear;

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.ts
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.ts
@@ -386,13 +386,7 @@ export class ReportingComponent implements OnInit, OnDestroy {
   }
 
   onFiltersClear(_event) {
-    const queryParams = {...this.route.snapshot.queryParams};
-
-    const filteredParams = pickBy((_value, key) => {
-        return this.allowedURLFilterTypes.indexOf(key) < 0;
-      }, queryParams);
-
-    this.router.navigate([], {queryParams: filteredParams});
+    this.router.navigate([], {queryParams: {} });
   }
 
   getData(reportQuery: ReportQuery) {
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.ts
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.ts
@@ -386,7 +386,20 @@ export class ReportingComponent implements OnInit, OnDestroy {
   }
 
   onFiltersClear(_event) {
-    this.router.navigate([], {queryParams: {} });
+    const queryParams = {...this.route.snapshot.queryParams};
+
+    const filteredParams = pickBy((_value, key) => {
+
+      // Only handles control_tag filters
+      if (key.includes('control_tag')) {
+        return false;
+      }
+
+        return this.allowedURLFilterTypes.indexOf(key) < 0;
+      }, queryParams);
+
+
+    this.router.navigate([], {queryParams: filteredParams});
   }
 
   getData(reportQuery: ReportQuery) {


components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.ts
```diff
@@ -386,13 +386,7 @@ export class ReportingComponent implements OnInit, OnDestroy {
   }
 
   onFiltersClear(_event) {
-    const queryParams = {...this.route.snapshot.queryParams};
-
-    const filteredParams = pickBy((_value, key) => {
-        return this.allowedURLFilterTypes.indexOf(key) < 0;
-      }, queryParams);
-
-    this.router.navigate([], {queryParams: filteredParams});
```
SEAjamieD: If feels like maybe I'm missing something here - or maybe this served a dual purpose in the past?

Since we are clearing ALL the applied filters, it makes sense to me that if we send through an empty object, we'll clear all the applied filters, instead of looping through the filters one at a time to parse them into an object and then remove them.

SEAjamieD: @msorens @Shadae - would love your thoughts on this when you have a chance

vjeffrey: is it maybe that we were clearing all filters _except_ end time/start time?  i noticed those values aren't in the allowedURLFilterTypes function..and those are "default" filters

SEAjamieD: @vjeffrey ahhh Yes! thats it, thank you!  It doesn't add a date filter to the URL if not explicitly selecting a date other than today

components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.ts
```diff
@@ -389,9 +389,16 @@ export class ReportingComponent implements OnInit, OnDestroy {
     const queryParams = {...this.route.snapshot.queryParams};
 
     const filteredParams = pickBy((_value, key) => {
+
+      // Only handles control_tag filters
+      if (key.includes('control_tag')) {
+        return false;
+      }
+
```
SEAjamieD: While this solves the immediate problem in what I feel is a pragmatic way, I don't believe in its value long term.  It also begins to expose future issues.  Currently, in order to remove the applied query filters, we are looping through an array that contains all of our _first_ level filter keys, checking it agains the query parameters, and then removing if it is a match.  This does not and will not account for any second/third level filtering we choose to do in the future, because with second level filtering, (i.e. control_tag) we end up with a key that looks like `control_tag:something`.  I'm not sure we ever want that many keys in the array we're checking against.

SEAjamieD: a possible option might be to split the key at `:` and return the first element of that array.  If i'm thinking correctly, it will return the expected key even if a `:` doesn't exist.

PR #2308, opened 2019-11-20
Auth 2295/increase cypress project limit

blakestier:

We needed to increase the project limit to test full functionality in cypress.

### DIFF
--- a/integration/tests/cypress_e2e.sh
+++ b/integration/tests/cypress_e2e.sh
@@ -6,6 +6,16 @@ test_name=":cypress: e2e tests"
 test_deploy_inspec_profiles=()
 test_skip_diagnostics=true
 
+do_create_config() {
+    do_create_config_default
+
+    #shellcheck disable=SC2154
+    cat <<EOF >> "$test_config_path"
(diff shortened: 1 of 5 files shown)


integration/tests/cypress_e2e.sh
```diff
@@ -6,6 +6,16 @@ test_name=":cypress: e2e tests"
 test_deploy_inspec_profiles=()
 test_skip_diagnostics=true
 
+do_create_config() {
+    do_create_config_default
+
+    #shellcheck disable=SC2154
+    cat <<EOF >> "$test_config_path"
+[auth_z.v1.sys.service]
+  project_limit = 50
```
blakestier: I went to `50` bc that's what we're doing elsewhere 🤷‍♂

lancewf: The current tests only use 7, so 50 will work!

bcmdarroch: Not sure if it matters if we just bump it up to what we need 🤔 probably better to just give ourselves more leeway

Issue #1902, opened 2019-10-17
"flip the switch" for apps tab in automate


vjeffrey:
## User Story 
This issue covers the work for flipping the switch on the backend feature flag to default to on as well as removing the frontend feature flag. This should be done after https://github.com/habitat-sh/habitat/issues/7017 is complete.

## Definition of Done
backend flag for apps tab data defaults to "on"
feature flag for apps tab is removed/apps tab is always viewable

susanev:
any way we can get at a minimum the overlay from the event-feed onto the applications page when a user does not have view permissions, or ideally hide that page entirely when a user does not have permission to view it.

this is the current experience when a user does not have permission to view that page.
![sad](https://user-images.githubusercontent.com/5489125/67112594-714b8180-f18c-11e9-9c81-8c9c97d7a56a.png)

kmacgugan:
Technical details that I am leaving here for my future reference: 

Currently we have three config settings that the backend flag toggles. We should remove these config settings as they won't be needed and strip the code of any logic that refers to them. These are the config settings: 

  [applications.v1]
    [applications.v1.sys]
      [applications.v1.sys.service]
        enable_nats_feature = true

[event_service]
  [event_service.v1]
    [event_service.v1.sys]
      [event_service.v1.sys.service]
        enable_nats_feature = true
      [event_service.v1.sys.log]

[event_gateway]
  [event_gateway.v1]
    [event_gateway.v1.sys]
      [event_gateway.v1.sys.service]
        enable_nats_feature = true

kmacgugan:
Another detail, add port checks to preflight check for automate: 
Port 4222 must be available and accessible to external systems
Ports 10140, 10147, and 10148 must be available and accessible to internal Automate services.

PR #2318, opened 2019-11-21
Enable EAS Applications backend systems w/o feature flag

danielsdeleo:

Enables all EAS backend systems without the use of feature flags, so the feature will be available by default for all users, as part of a GA release. As part of this change, the `chef-automate applications` command category no longer contains any subcommands, so it is removed.

#1902 

### :+1: Definition of Done

EAS applications functionality works without use of CLI command or configuration to enable it.

Need to build:
* applications
* event-gateway
* event-service
* automate-gateway
* deployment-service
* automate-cli

After starting all services, EAS applications functionality should be available without further configuration (aside from the UI feature flag, see #2317)

Note that previously, the EAS applications functionality was enabled by default when using the studio because the necessary feature flags were set in `dev/config.toml`; those have been removed.


kmacgugan:
There is mention of enable_apps_feature and enable_nats_feature still left in /automate/terraform/test-environments/modules/chef_automate_install/main.tf

There's another mention of enable_nats_feature in /automate/components/event-service/habitat/default.toml

Not sure what the impact of leaving these is. Looks good otherwise.

danielsdeleo:
> There is mention of enable_apps_feature and enable_nats_feature still left in /automate/terraform/test-environments/modules/chef_automate_install/main.tf

I think this would have broken acceptance deploys with invalid automate config toml

> There's another mention of enable_nats_feature in /automate/components/event-service/habitat/default.toml

That one was harmless but I deleted it to keep things tidy

danielsdeleo:
Going to merge this ahead of GA and/or public beta because we still have the UI feature flag in place. This will reduce the complexity in starting to use the beta.

PR #2319, opened 2019-11-21
Auth 2304/hide unassigned if only one allowed

blakestier:

Users were confused by the `(unassigned)` label in the dropdown when it was the only "project" to which they had access.

### :+1: Definition of Done

All these scenarios are true:

- [x] Hide the (unassigned) label from the top-right if that's the only thing a user has access to
- [x] If they have access to only 1 project, that projects name should still appear in the top-right
- [x] If a user has access to unassigned and another project, unassigned should still show up in the top-right if the user filters to view only unassigned resources

Bring up the UI and sign in as users with varying project permissions (one project, one project + unassigned, only unassigned, no projects) and ensure that the above scenarios render correctly.

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/page-components/projects-filter/projects-filter.component.html
+++ b/components/automate-ui/src/app/page-components/projects-filter/projects-filter.component.html
@@ -5,5 +5,6 @@
   [selectionCountVisible]="projectsFilter.selectionCountVisible$ | async"
   [selectionCountActive]="projectsFilter.selectionCountActive$ | async"
   [dropdownCaretVisible]="projectsFilter.dropdownCaretVisible$ | async"
+  [filterVisible]="projectsFilter.filterVisible$ | async"
   (onSelection)="projectsFilter.saveOptions($event)">
 </app-projects-filter-dropdown>
--- a/components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.spec.ts
+++ b/components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.spec.ts
@@ -218,6 +218,11 @@ describe('projectsFilterReducer', () => {
         genProject('zz-proj')
       );
 
+      it('displays filter', () => {
+        const { filterVisible } = projectsFilterReducer(initialState, action);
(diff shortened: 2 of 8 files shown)


components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.spec.ts
```diff
@@ -239,9 +239,14 @@ describe('projectsFilterReducer', () => {
         genUnassignedProject()
       );
 
-      it('displays "(unassigned)"', () => {
+      it('does not display filter', () => {
+        const { filterVisible } = projectsFilterReducer(initialState, action);
+        expect(filterVisible).toEqual(false);
```
msorens: Suggest also add a check for this to be true in at least one case that is **not** just unassigned.

components/automate-ui/src/app/page-components/projects-filter/projects-filter.component.html
```diff
@@ -5,5 +5,6 @@
   [selectionCountVisible]="projectsFilter.selectionCountVisible$ | async"
   [selectionCountActive]="projectsFilter.selectionCountActive$ | async"
   [dropdownCaretVisible]="projectsFilter.dropdownCaretVisible$ | async"
+  [filterVisible]="projectsFilter.filterVisible$ | async"
```
msorens: I suspect you will need to set this in some unit tests for projects-filter-dropdown.component.ts, too.

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.spec.ts
blakestier: I added it everywhere to be consistent with the testing suite. Feels a little redundant, but such is testing.

components/automate-ui/src/app/page-components/projects-filter/projects-filter.component.html
blakestier: I did, thank you!

Issue #2258, opened 2019-11-18
add pagination and search to the profiles endpoints


vjeffrey:
## User Story

In https://github.com/chef/automate/issues/2144, @phiggins did some research to see if there was anything wrong with the postgres query or endpoint to cause the slow down. 
The conclusion was that the slowdown is most likely related to the json parsing. This endpoint does not have pagination and we are currently shipping over 300 "available" profiles, which is alot of objects to parse and return.
As a result, we decided to go ahead and add pagination and search to the endpoint (search was being done in the ui only).

The endpoint already takes a query object, so the pagination and filter parameters come for free:
https://github.com/chef/automate/blob/master/components/automate-gateway/api/compliance/profiles/profiles.proto#L61

The ui search existed on name, title, and version, so we'll need to make sure the api search covers these fields as well: 
https://github.com/chef/automate/blob/master/components/automate-ui/src/app/pages/%2Bcompliance/%2Bprofile/%2Bprofile-overview/profile-overview.component.ts#L99

## Definition of Done
List Profiles endpoint 
- respects provided pagination parameters
- when no pagination parameters are defined, returns all profiles
- provides text search capabilities for searching on profile name, title, and version.

PR #2320, opened 2019-11-21
Refactor latestVersion to avoid calling heavyweight codepath.

phiggins:
* I'm changing the behavior and signature of ListProfilesMetadata for #2258 so I wanted to avoid calling it here.
* This code was pulling a bunch of stuff from the database, parsing json, etc when it didn't need to.

### DIFF
--- a/components/compliance-service/profiles/db/store.go
+++ b/components/compliance-service/profiles/db/store.go
@@ -285,6 +285,10 @@ func (s *Store) latestVersion(namespace string, name string) (string, error) {
 		return "", err
 	}
 
+	if len(versions) == 0 {
+		return "", nil
+	}
+
 	semver.Sort(versions)
 	latest := versions[len(versions)-1]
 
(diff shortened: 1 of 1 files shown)


components/compliance-service/profiles/db/store.go
```diff
@@ -256,33 +256,49 @@ func (s *Store) ReadMarketTar(name string, version string) ([]byte, error) {
...
 		}
+		versions = append(versions, version)
+	}
+	err = rows.Err()
+	if err != nil {
+		return "", err
 	}
 
-	return &latest, nil
+	semver.Sort(versions)
```
stevendanna: Interesting.  I suppose this is also a behavior change/bug fix since previously we would have just gotten whatever the lexicographic sort order of the versions was?

phiggins: This function was doing an equivalent thing but more verbosely: https://github.com/chef/automate/blob/5651c419def4fd1fb56ec67c19509589d718b243/components/compliance-service/profiles/db/store.go#L266-L273

We do sort by the version string in the database in other places though.

stevendanna: Aha! Missed that bit, thanks!

PR #2329, opened 2019-11-21
Batch datafeed with cereal tasks

satellite15:

Removes the basic integration with the cereal service and adopts a new workflow to manage multiple tasks associated with the data-feed. The aggregation of data is now managed completely by this workflow and the associated tasks.

The workflow has 2 associated tasks. OnStart the work flow has a task to determine which nodes to gather data for - DataFeedPollTask collates a map of node ipaddress to node ID(to gather client run and attrribute data), report ID (to gather compliance report data). Once this is complete the workflow then batches the ID's before processing each batch with the remaining DataFeedAggregateTask which collates the data and sends to the registered endpoints (destinations). Each batch is processed sequentially and the default batch size is 100 nodes and is configurable. Batching manages the memory requirements of the data-feed-service

#### DataFeedPollTask
Checks for nodes that have had a client run in the polled interval
Stores the Node ID's in its the workflow payload in a map with the ipaddress as key to a node ID, report ID tuple
Checks for compliance reports in the polled interval
Stores the report ID's in its the workflow payload in a map with the ipaddress as key to a node ID, report ID tuple

#### DataFeedAggregateTask
Started when the poll task completes, takes the nodes from poll task
For each node builds the ohai and client run data for the node and it's most recent compliance report. 
Builds array of datafeedMessages to return in it's result. 
takes the batched datafeed message set and posts it to the destinations. 
If there are further batches to be processed the DatafeedClient task is scheduled to process the next batch. When all batches are complete the workflow completes.

Configure the poll interval in the config.toml for the data-feed-service to a short period 3m or 5m
Add a destination (service now or requestbin) in the Fata Feed UI page.
Perform client runs and compliance scans on nodes managed by automate.
Data should be periodically posted to the configured destination
In the config
node_batch_size = 100
updated_nodes_only = true
can be changed to change the size of node batches and whether to include data that has been updated in the interval only. Set to true only data that has been updated in the interval is added to the message set, set to false that data will include the latest client run or compliance report from a node that has changed. E.g. if there has been a client run in the interval it will also include the latest compliance report, if there is one, regardless of when it occurred. And vice versa if there was a compliance report in the interval it will also include the most recent client run data.

- [x] Tests added/updated? No new automated tests, tested using the ServiceNow dev/test environment
- [x] Docs added/updated? Code comments only at this stage


components/data-feed-service/service/data-feed-notifier-task.go
```diff
@@ -0,0 +1,91 @@
...
+	for destination := range destinations {
+		log.Debugf("Destination name %v", destinations[destination].Name)
+		log.Debugf("Destination url %v", destinations[destination].URL)
+		log.Debugf("Destination secret %v", destinations[destination].Secret)
+
+		username, password, err := GetCredentials(ctx, d.secrets, destinations[destination].Secret)
+
+		if err != nil {
+			log.Errorf("Error retrieving credentials, cannot send asset notification: %v", err)
+			// TODO error handling - need some form of report in automate that indicates when data was sent and if it was successful
```
vjeffrey: for future: would sending that message to event feed be enough or would we need something else?

components/data-feed-service/service/data-feed-poll-task.go
```diff
@@ -0,0 +1,209 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get destinations from db")
+	}
+	params.FeedStart = feedStartTime
+	params.FeedEnd = feedEndTime
+	params.PollTaskParams.NextFeedStart = feedEndTime
+	params.PollTaskParams.NextFeedEnd = feedEndTime.Add(params.PollTaskParams.FeedInterval)
+
+	if len(destinations) == 0 {
+		log.Info("DataFeedPollTask.Run no destinations returning")
```
vjeffrey: should this be an info log msg or warn/error?

jaym:
Thanks for the detailed description. Have not reviewed the code itself yet, maybe it'll answer this question:
why separate DataFeedAggregateTask and DataFeedNotifierTask. We could save copying a bunch of data (not sure how much this actually is) into postgres by combining them

components/data-feed-service/service/data-feed-notifier-task.go
satellite15: For now I am not considering anything other than logging the errors. For the future there needs to be a mechanism so that a user can review anything that has gone wrong, possibly with an errors page in the UI and errors available from an API. I'm happy to look at what the event feed can offer w.r.t error tracking and reporting. I'll add a ticket to my zenboard to review it for those purposes. Thanks!

components/data-feed-service/service/data-feed-poll-task.go
satellite15: It wouldn't be a warning or error, this is normal behaviour if a destination has not been added.

satellite15:
> Thanks for the detailed description. Have not reviewed the code itself yet, maybe it'll answer this question:
> why separate DataFeedAggregateTask and DataFeedNotifierTask. We could save copying a bunch of data (not sure how much this actually is) into postgres by combining them

Initially I simply split these tasks up logically. The aggregate task was initially 2 tasks that I combined as the logic was difficult to manage through two separate tasks. This is a fair point and I will combine the notification logic into the aggregate task.

components/data-feed-service/dao/destination.go
```diff
@@ -1,8 +1,175 @@
...
+func (db *DB) getDestinationRequestToDB(inDestination *datafeed.GetDestinationRequest) *Destination {
+	newDestination := Destination{}
+	newDestination.ID = inDestination.Id
+
+	return &newDestination
+}
+
+func (db *DB) AddDestination(destination *datafeed.AddDestinationRequest) (bool, error) {
+	dbDestination := db.addToDBDestination(destination)
+	var err error
```
stevendanna: [nit] I think we might want to avoid any confusion about shadowing here if possible by dropping this `var err error` and doing something like:

```
	err := Transact(db, func(tx *DBTrans) error {
		if err := tx.Insert(dbDestination); err != nil {
			return errors.Wrap(err, "AddDestination: unable to insert destination")
		}
		return nil
	})
```

similarly in the other examples like this.

components/data-feed-service/scripts/grpc.sh
```diff
@@ -0,0 +1,40 @@
...
+    -I vendor/github.com/grpc-ecosystem/grpc-gateway \
+    -I vendor/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis \
+    --go_out=plugins=grpc,paths=source_relative:/src \
+    --grpc-gateway_out=request_context=true,logtostderr=true:"${fauxpath}" \
+    --a2-config_out=paths=source_relative:/src \
+    "${i}" \
+    || exit $?
+
+  sync_from_fauxpath "${i}"
+done
```
stevendanna: I'm a bit confused on why we need this script.

components/data-feed-service/service/data-feed-aggregate-task.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+type DataFeedAggregateTaskResults struct {
+	DataFeedMessages map[string]map[string]interface{}
+	NodeIDs          map[string]NodeIDs
+}
+
+func NewDataFeedAggregateTask(dataFeedConfig *config.DataFeedConfig, connFactory *secureconn.Factory) (*DataFeedAggregateTask, error) {
+
+	cfgMgmtConn, err := connFactory.Dial("config-mgmt-service", dataFeedConfig.CfgmgmtConfig.Target)
```
stevendanna: I think it might be better if these constructors took the relevant connections as arguments, that way we don't need to create & manage multiple connection objects for the same upstream.

components/data-feed-service/service/data-feed-poll-task.go
```diff
@@ -0,0 +1,209 @@
...
+	params := DataFeedWorkflowParams{}
+	err := task.GetParameters(&params)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to parse task parameters")
+	}
+	log.Debugf("DataFeedPollTask.Run params %v", params)
+	log.Debugf("DataFeedPollTask.Run FeedStart %v", params.FeedStart)
+	log.Debugf("DataFeedPollTask.Run FeedEnd %v", params.FeedEnd)
+	log.Debugf("DataFeedPollTask.Run NextStart %v", params.PollTaskParams.NextFeedStart)
+	log.Debugf("DataFeedPollTask.Run NextEnd %v", params.PollTaskParams.NextFeedEnd)
```
stevendanna: [nit] For things like this we typically do a single log message using `WithFields` to print the bits we want in a structured fashion.

components/data-feed-service/service/data-feed-poll-task.go
```diff
@@ -0,0 +1,209 @@
...
+	now := time.Now()
+	params = d.getFeedTimes(params, now)
```
stevendanna: It might be a bit cleaner to have `getFeedTimes` to return start and end directly rather than mutating its input.

components/data-feed-service/service/data-feed-poll-task.go
```diff
@@ -0,0 +1,209 @@
...
+	feedStartTime := params.PollTaskParams.NextFeedStart
+	feedEndTime := params.PollTaskParams.NextFeedEnd
+	destinations, err := d.db.ListDBDestinations()
```
stevendanna: Moving this before or after the various time-related mutations might improve readability.

components/data-feed-service/service/data-feed-poll-task.go
```diff
@@ -0,0 +1,209 @@
...
+	params.PollTaskParams.NextFeedStart = feedEndTime
+	params.PollTaskParams.NextFeedEnd = feedEndTime.Add(params.PollTaskParams.FeedInterval)
+
+	if len(destinations) == 0 {
+		log.Info("DataFeedPollTask.Run no destinations returning")
+		return nil, nil
+	}
+
+	nodeIDs, err := d.GetChangedNodes(ctx, params.PollTaskParams.AssetPageSize, feedStartTime, feedEndTime)
+	log.Debugf("DataFeedPollTask nodes %v", nodeIDs)
```
stevendanna: While this is debug, I think we might still want to avoid log messages that potentially output a boatload of data.  Might we consider just logging the count of nodeIDs?

components/data-feed-service/service/data-feed-poll-task.go
```diff
@@ -0,0 +1,209 @@
...
+		return nil, err
+	}
+
+	params.NodeIDs = nodeIDs
+
+	log.Debugf("Updated next feed interval start, end: %s, %s", params.PollTaskParams.NextFeedStart.Format("15:04:05"), params.PollTaskParams.NextFeedEnd.Format("15:04:05"))
+	log.Debugf("Updated feed interval start, end: %s, %s", params.FeedStart.Format("15:04:05"), params.FeedEnd.Format("15:04:05"))
+
+	// we must update the workflow params to ensure the nest schedule has the update interval times
+	err = d.manager.UpdateWorkflowScheduleByName(context.Background(), dataFeedScheduleName, dataFeedWorkflowName,
```
stevendanna: Here you can use the context that we were passed as an argument rather than context.Background().

stevendanna: I'm wondering if this is the right place to update the schedule.  At this point we don't know if the work has actually been done yet. Should we wait to update the schedule until the data is actually sent to the destinations so that intervals are retried if they fail?

stevendanna:
Similar to @jaym's question on the aggregate and notification tasks, it seems like we might benefit from merging DataFeedPollTask & DataFeedListReportsTask.  Conceptually I like having those separate but we are already doing so much data shoveling between various systems that it might be good to combine them.

satellite15:
> Similar to @jaym's question on the aggregate and notification tasks, it seems like we might benefit from merging DataFeedPollTask & DataFeedListReportsTask. Conceptually I like having those separate but we are already doing so much data shoveling between various systems that it might be good to combine them.

No problem. Combing the aggregate and notifier tasks also gives us the potentially to add concurrent aggregate tasks to handle the batching rather than queuing them serially. However I'll leave that for another day.

components/data-feed-service/scripts/grpc.sh
satellite15: Marshalling the datafeed messages in and out of the cereal lib was prone to error. The messages were originally all in a go struct, since the reports and client run are proto messages there were issues with client runs being marshalled in and out of the cereal lib. json.unmarshal is not smart enough to determine that there are proto messages in the go struct and client runs could cause errors. To resolve this I made the DataFeedMessage a proto message. There is a data-feed-message.proto which this script compiles.

WIth the aggregate and notifier task combined however I might be able to remove it and this script if that doesn't introduce an error in the outgoing data.

components/data-feed-service/service/data-feed-aggregate-task.go
satellite15: fair enough

components/data-feed-service/service/data-feed-poll-task.go
satellite15: I think after as I've spotted a small would be bug (masked by the lag correction). The updated feed times should be returned in the task result when there are no destinations or there is an error. I'll improve this area.

components/data-feed-service/service/data-feed-poll-task.go
satellite15: At the moment yes. Current implementation is fail safely and  move on. Retries aren't trivial, we could get stuck on an interval with corrupt data, or fall behind and then the lag correction would skip intervals, we could try and add multiple intervals to the poll task so that retries are included with the current interval but then we could end up constantly retrying multiple intervals. We could end up retrying an interval that's data has already been superseded by a subsequent interval. I went for a keep it simple approach. Errors and retries are a roadmap feature! I'm actually going to do it slightly earlier so that the schedule is updated on error and if there is no work to do.

components/data-feed-service/service/data-feed-poll-task.go
satellite15: I'll look through the debug, I meant to take this out. There are probably a few others like it.

satellite15:
> Still looking over some of the details but I've left some inline comments from a first pass.

I have committed changes as per all the comments

satellite15:
I have pushed updates to resolve all comments so far. I also updated the PR description in line with the changes.

components/data-feed-service/scripts/grpc.sh
satellite15: I have removed the proto message and map. With the aggregate and notifier tasks combined this is no longer required to marshal in/out of the cereal lib witjout error

components/data-feed-service/service/data-feed-aggregate-task.go
```diff
@@ -0,0 +1,210 @@
...
+		buffer.Write(data)
+		buffer.WriteString("\n")
+	}
+
+	destinations, err := d.db.ListDBDestinations()
+	if err != nil {
+		return nil, err
+	}
+	for destination := range destinations {
+		log.Debugf("Destination name %v", destinations[destination].Name)
```
jaym:
```golang
log.WithFields(log.Fields{
"name": destinations[destination].Name,
"url:" destinations[destination].URL,
}).Debug("Something")
```

components/data-feed-service/service/data-feed-aggregate-task.go
```diff
@@ -0,0 +1,210 @@
...
+		log.Debugf("Destination url %v", destinations[destination].URL)
+		log.Debugf("Destination secret %v", destinations[destination].Secret)
```
jaym: consider not logging the secret

components/data-feed-service/service/data-feed-workflow.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+		/*
+		 * Update the workflow params
+		 */
+		params.NodeIDs = taskResults.NodeIDs
+		if len(taskResults.NodeIDs) > 0 {
+			payload.NodeIDs = batchNodeIDs(params.NodeBatchSize, taskResults.NodeIDs)
+			log.Debugf("payload.NodeIDs %v payload.NodeIDsIndex %v", payload.NodeIDs, payload.NodeIDsIndex)
+			params.NodeIDs = payload.NodeIDs[payload.NodeIDsIndex]
+			err = w.EnqueueTask(dataFeedAggregateTaskName, params)
```
jaym: not that this is wrong or that you should change it, but you could enqueue all the batches and have only one thread available to process them per machine. Unless you need exactly one batch to be processed at a time, then for sure leave it as is

components/data-feed-service/service/data-feed-workflow.go
```diff
@@ -0,0 +1,157 @@
...
+	switch ev.TaskName {
+	case dataFeedPollTaskName:
+		taskResults, err := getPollTaskResults(ev)
+		if err != nil {
+			return w.Fail(err)
+		}
+		/*
+		 * Update the workflow params
+		 */
+		params.NodeIDs = taskResults.NodeIDs
```
jaym: NodeIDs looks like it is not something the caller of the workflow passes in, and is derived while running the workflow. I found this a little confusing. It might be worth having the parameters for the task be distinct from the workflow parameters.

components/data-feed-service/service/data-feed-workflow.go
```diff
@@ -0,0 +1,157 @@
...
+			}
+		} else {
+			payload.TasksComplete = true
+		}
+		log.Debugf("AggregateTask %v, tasks complete: %v, time: %v", dataFeedAggregateTaskName, payload.TasksComplete, time.Now())
+	}
+
+	log.Debugf("TasksComplete %v, complete: %v", ev.TaskName, payload.TasksComplete)
+	if payload.TasksComplete {
+		return w.Complete()
```
jaym: Did you mean to save the last payload as the result? I ask because payload has `TasksComplete` on it. It seems like this could be a local variable if it doesn't need to be saved as part of the result since it being set to true always leads to completing the workflow. Maybe it wouldn't even be needed if saving the result was required

components/data-feed-service/service/data-feed-workflow.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	return w.Continue(&payload)
+
+}
+
+func (e *DataFeedWorkflowExecutor) OnCancel(w cereal.WorkflowInstance, ev cereal.CancelEvent) cereal.Decision {
+	return w.Complete()
+}
+
+func batchNodeIDs(batchSize int, nodeIDs map[string]NodeIDs) []map[string]NodeIDs {
```
jaym: it might not be horrible to write a few unit tests for this

components/data-feed-service/service/data-feed-workflow.go
```diff
@@ -0,0 +1,157 @@
...
+	numNodes := len(nodeIDs)
+	log.Debugf("number of nodes %v", numNodes)
+	var numBatches int
+	if numNodes < batchSize {
+		numBatches = 1
+	} else {
+		numBatches = numNodes / batchSize
+	}
+	if numNodes%1 > 0 {
```
jaym: I may be missing something, but isn't this always 0? I've never modded by 1 before, but 
`0%2=0` `1%2=1` `2%2=0`?

components/data-feed-service/service/data-feed-workflow.go
satellite15: Simply doing them in serial in a first pass to see how we perform. We don't need exactly one batch to be processed at a time unless we only have 1 machine.

components/data-feed-service/service/data-feed-workflow.go
satellite15: I'll raise an issue to look at this on our Zen board. I am using the workflow parameters as they need to be updated with the feed times for updating the schedule. Initially I had params per task but moved away for the simplicity with the feed times. It's worth looking at again now that the tasks have been consolidated.

components/data-feed-service/service/data-feed-workflow.go
satellite15: As above I'll add an issue to review this.

components/data-feed-service/service/data-feed-workflow.go
satellite15: Unit tests are an existing issue on our zenboard, there has been a lot of churn in this code and constantly refactoring unit tests would not have been ideal. We also have integration tests under development.

components/data-feed-service/service/data-feed-workflow.go
satellite15: not sure how that 1 got in there, it's fixed now. Simplified a little and it is numNodes%batchSize.

Example 25 nodes, batch size 10. `numBatches = 2` but 25 nodes requires 3 batches thus if `numNodes%batchSize > 0`, `25%10 = 5` we need another batch

Issue #1792, opened 2019-10-03
Deployment of a hab-builder is tested


ryancragun:
We need to deploy and verify that the hab builder works

## Acceptance Criteria:
* There is a pre-merge verification test that asserts that builder is minimally functional
* There is a post-merge acceptance environment that deploys builder with verification tests

## Questions:
* Are there any existing suites that we can re-use here?

sdelano:
Sprint Planning Notes:
- we have a test that runs pre-merge and post-merge in a one-off environment
  - logs in
  - creates token
  - creates an origin
  - creates a package
  - uploads a package
  - takes a backup
  - uninstalls automate
  - restores from backup (reinstalls)
  - downloads the package
- what we need now
  - deployed to dev and acceptance environments
    - move current workflow environments to full-stack (but a funnier name)
      - workflow
      - chef-server (infra)
      - bldr
  - run a subset of the premerge tests
    - login
    - get a token
    - create an origin
    - upload a package
    - promote a package?
    - download a package
  - try and run the hab bldr api tests
  - bonus points
    - how can we get a token easier?

PR #2330, opened 2019-11-21
Add builder smoke test to deploy pipeline

jaym:
step towards #1792


.expeditor/deploy.pipeline.yml
```diff
@@ -26,6 +26,15 @@ steps:
       executor:
         docker:
 
+  - label: "builder"
```
tduffield: There is a `:habicat:` emoji :) https://github.com/buildkite/emojis

PR #2337, opened 2019-11-21
Right panel positioning issue

### LINKED ISSUE #2238: right panel positioning issue
## Overview
The run history panel, compliance/scan-job error panel, and compliance/reports scan results panel are not positioned correctly. They seem to not correctly take up the right space vertically.

## Repro steps
- navigate to `infrastructure/client-runs` and go to a node detail page
- scroll down
- then open the run history panel
- notice that the top is cut off

#### Client runs
run-history-panel

#### Compliance scan jobs
scan-jobs-error-panel

tarablack01:
The run history panel, compliance/scan-job error panel, and compliance/reports scan results panel are not positioned correctly. They seem to not correctly take up the right space vertically.

https://app.zenhub.com/workspaces/ui-team-5cba23a3b14fdc05970c8f87/issues/chef/automate/2238

fixes #2238

### :+1: Definition of Done
Run history panel is not cut off on top and does extend to the bottom of page.

navigate to infrastructure/client-runs and go to a node detail page
scroll down
then open the run history panel
notice that the top is cut off

Screen Shot 2019-11-21 at 11 48 22 AM (3)

### DIFF
--- a/components/automate-ui/src/app/entities/layout/layout.facade.ts
+++ b/components/automate-ui/src/app/entities/layout/layout.facade.ts
@@ -44,6 +44,10 @@ export class LayoutFacadeService {
         });
     }
 
+    hasGlobalNotifications(): boolean {
+        return this.headerHeight === '110px';
+    }
+
     ShowPageLoading(showLoading: boolean) {
         this.store.dispatch( new ShowPageLoading(showLoading));
     }
--- a/components/automate-ui/src/app/page-components/resources/resources.component.scss
+++ b/components/automate-ui/src/app/page-components/resources/resources.component.scss
@@ -1,6 +1,11 @@
 @import "~styles/variables";
 @import "~styles/mixins";
 
+#resource-item-list {
+  height: 20vh;
+  overflow-y: scroll;
+}
+
 .rollup-item {
   display: inline-block;
   width: 15%;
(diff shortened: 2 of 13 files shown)


components/automate-ui/src/app/page-components/resources/resources.component.scss
```diff
@@ -1,6 +1,11 @@
 @import "~styles/variables";
 @import "~styles/mixins";
 
+#resource-item-list {
+  height: 20vh;
```
SEAjamieD: I'm not sure the requirement, but this limits us to only seeing 3 resources at any given time without having to scroll.  UX wise that doesn't seem ideal given that there could be a lot of resources in there and users would have to scroll deep.  Looping in @susanev for a look

components/automate-ui/src/app/entities/layout/layout.facade.ts
```diff
@@ -44,6 +44,10 @@ export class LayoutFacadeService {
         });
     }
 
+    hasGlobalNotifications(): boolean {
+        return this.headerHeight === '110px';
```
msorens: Reviewing after the fact:
This is rather brittle. Is it possible to check for an element rather than a specific (magic) size? If not, could you at least add a comment explaining the reason for this magic number?

PR #2339, opened 2019-11-21
[purge] always log purge actions

ryancragun:
Always log purge workflows and ElasticSearch sidecar purge operations.


components/es-sidecar-service/pkg/server/purge.go
```diff
@@ -39,34 +42,33 @@ func (server *EsSidecarServer) PurgeTimeSeriesIndicesByAge(ctx context.Context,
...
+	logctx := log.WithFields(log.Fields{
 		"id":                     in.GetId(),
 		"index":                  in.GetIndex(),
 		"delete-older-than-days": in.GetOlderThanDays(),
 		"custom-purge-field":     in.GetCustomPurgeField(),
-	}
+	})
 
-	log.WithFields(logFields).Debug("Initiating document purge from index")
+	logctx.Debug("Initiating document purge from index")
```
yzl: Debug or Info? (I don't know how often we expect the message to appear.)

ryancragun: I think Debug is fine here because we'll always log a success or failure later on

PR #2342, opened 2019-11-22
cereal: make EnqueuedAt time accessible to tasks & workflows

stevendanna:
This exposes the EnqueuedAt time for Tasks and Workflow events. This
allows individual workflow executors and task executors to make
decisions based on the age of the event they are processing.

We want to use this to implement some safety checks against processing
very old events.


lib/cereal/grpc/grpc.go
```diff
@@ -180,6 +180,22 @@ func (g *GrpcBackend) DequeueWorkflow(ctx context.Context, workflowNames []strin
 		return nil, nil, errors.New("unexpected")
 	}
 
+	tsProto := deq.GetEvent().GetEnqueuedAt()
+	ts := time.Time{}
+	if tsProto == nil {
+		// TODO(ssd) 2019-11-22: Determine what we want to do
```
jaym: zero value seems reasonable

api/interservice/cereal/cereal.proto
```diff
@@ -10,6 +10,11 @@ message Task {
   string name = 1;
   bytes parameters = 2;
   google.protobuf.Timestamp start_after = 3;
+  TaskMetadata metadata =4;
```
jaym: [nit] weird spacing

components/cereal-service/pkg/server/server.go
```diff
@@ -441,6 +441,11 @@ func (s *CerealService) DequeueTask(req cereal.Cereal_DequeueTaskServer) error {
 		return err
 	}
 
+	tsProto, err := ptypes.TimestampProto(task.Metadata.EnqueuedAt)
+	if err != nil {
+		return err
```
jaym: [nit] maybe log

lib/cereal/grpc/grpc.go
jaym: the godoc could mention that its probably best to test for 0

stevendanna:
> nice. i was going to have the task query the workflow schedule to see if it changed, but this is better

I could imagine we might still want to do that.

jaym:
i think if we made the window fairly small, we wouldn't need to. but could be argued that safety checks are important for the data lifecycle stuff and we should have more

jaym:
one more nit: we might be able to add a check to make sure the enqueued_at fields are being set in one of the tests

lib/cereal/grpc/grpc.go
stevendanna: Which doc are you looking at?  I'm not seeing a function for the Timestamp type that does a zero test similar to `time.IsZero`.  However, it _appears_ that zero-values for both types round-trip correctly:

```
package main

import (
        "fmt"
        "time"

        "github.com/golang/protobuf/ptypes"
        tspb "github.com/golang/protobuf/ptypes/timestamp"
)

func main() {
        fmt.Println("Test time.Time{} round-trip")
        v0, err := ptypes.TimestampProto(time.Time{})
        fmt.Printf("ptypes.TimestampProto(time.Time{}) = %v (err: %v)\n", v0, err)
        v1, err := ptypes.Timestamp(v0)
        fmt.Printf("ptypes.Timestamp(%v) = %#v (err: %v) (isZero: %v)\n", v0, v1, err, v1.IsZero())
        fmt.Println("")

        fmt.Println("Test tspb.Timestamp{} round-trip")
        v3, err := ptypes.Timestamp(&tspb.Timestamp{})
        fmt.Printf("ptypes.Timestamp(&tspb.Timestamp{}) = %#v (err: %v)\n", v3, err)
        v4, err := ptypes.TimestampProto(v3)
        fmt.Printf("ptypes.TimestampProto(%#v) = %#v (err: %v)\n", v3, v4, err)
}
```

stevendanna:
> one more nit: we might be able to add a check to make sure the enqueued_at fields are being set in one of the tests

done

lib/cereal/grpc/grpc.go
jaym: i think what i was trying to say was that in our cereal godoc, we could say that its possible you get a zero value (for example in the multinode case)

PR #2343, opened 2019-11-22
tools: add print-ds-config

stevendanna:
This tool allows us to print the TOML configuration directly from a
copy of the bolt-db database. Example usage:

```
cp /var/opt/chef-automate/backups/BACKUP_ID/deployment-service/bolt/bolt.db ./bolt.db 
./print-ds-config ./bolt.db 
```

### DIFF
--- a/tools/print-ds-config/main.go
+++ b/tools/print-ds-config/main.go
@@ -1,0 +1,93 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/boltdb/bolt"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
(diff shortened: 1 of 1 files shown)


tools/print-ds-config/main.go
```diff
@@ -0,0 +1,93 @@
...
+		return errors.Wrap(err, "could not initialize database")
+	}
+
+	d, err := deploymentStore.GetDeployment()
+	if err != nil {
+		return errors.Wrap(err, "could not get deployment from database")
+	}
+
+	c := d.GetUserOverrideConfigForPersistence()
+	c.Redact()
```
stevendanna: I want to commit this as is if possible since this is what the user has run, but as a follow up, let's add a flag to make the redact skippable in case we need to see the full thing at some point.

PR #2344, opened 2019-11-22
Add upgrade/backup/restore test

jaym:
There have been reports that having the retention settings set
and doing a backup and restore causes the restore to fail

### DIFF
--- a/api/interservice/deployment/override_config.go
+++ b/api/interservice/deployment/override_config.go
@@ -1,32 +1,63 @@
 package deployment
 
 import (
+	"github.com/sirupsen/logrus"
+
 	api "github.com/chef/automate/api/config/deployment"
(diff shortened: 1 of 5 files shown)


api/interservice/deployment/override_config.go
```diff
@@ -1,32 +1,63 @@
...
+	// They gave us nothing, return the existing config.
 	if req == nil {
-		return cfg
+		return nil
 	}
-	if req.PatchConfig != nil {
-		cfg = req.PatchConfig
+
+	// If they passed a config set we'll only use that
+	if req.GetSetConfig() != nil {
```
jaym: i feel like preventing people from setting deprecated config isn't right. As an example, you need to restore an old automate with the deprecated config to a new version, you probably still want the migrations to run

jaym:
:shipit:

PR #2347, opened 2019-11-22
nodemanager Read RPC project filtering

lancewf:
Adding project filtering to the "api/v0/nodes/<id>"/Read RPC function. This is needed because the nodemanager stores both ingested nodes and manually added nodes. These changes filter the ingested nodes only. The manually added nodes will not be filtered by the project tags in the context object.

If the ingested node is not allowed by the passed in projects in the request a 404 will be returned. 

#2350

### :+1: Definition of Done

All ingested client runs and compliance nodes are filtered by projects on the Read RPC. 

1. `build components/nodemanager-service && start_all_services`
1. Ingest some client run nodes with `send_chef_run_example`
1. Go to https://a2-dev.test/infrastructure/client-runs and copy the node ID for the new node created. 
1. Ensure the node can be retrieved with unassigned by running this command `curl -X GET -f --insecure -H "api-token: $(get_admin_token)" -H "projects: (unassigned)" https://localhost/api/v0/nodes/id/<node ID>`
1. Create a project with the name "project9" with one rule that matches environment = B
1. Apply the rules
1. Ensure the node can be retrieved with project9 with this command `curl -X GET -f --insecure -H "api-token: $(get_admin_token)" -H "projects: project9" https://localhost/api/v0/nodes/id/<node ID>`
1. Ensure the node can not be retrieved with unassigned by running this command again `curl -X GET -f --insecure -H "api-token: $(get_admin_token)" -H "projects: (unassigned)" https://localhost/api/v0/nodes/id/<node ID>`. You should see a 404. 

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/nodemanager-service/tests/project_filtering_test.go
+++ b/components/nodemanager-service/tests/project_filtering_test.go
@@ -17,6 +17,142 @@ import (
 	"github.com/chef/automate/lib/grpc/auth_context"
 )
 
+func TestReadProjectFilteringIngestedNodes(t *testing.T) {
+	timestamp, err := ptypes.TimestampProto(time.Now())
+	require.NoError(t, err)
+
+	db, err := createPGDB()
+	require.NoError(t, err)
+
+	nodeManager := nodesserver.New(db, nil, "")
+
+	// Adding a manual node
+	mgr1 := manager.NodeManager{Name: "mgr1", Type: "aws-ec2"}
+	mgrID1, err := db.AddNodeManager(&mgr1, "11111111")
+	require.NoError(t, err)
+	defer db.DeleteNodeManager(mgrID1)
+
+	node1 := manager.ManagerNode{Id: "i-1111111", Region: "us-west-2", Host: "Node1"}
+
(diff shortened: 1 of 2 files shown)


components/nodemanager-service/tests/project_filtering_test.go
```diff
@@ -17,6 +17,173 @@ import (
...
+		},
+		{
+			description:     "Node has no projects; requested for all projects",
+			requestProjects: []string{authzConstants.AllProjectsExternalID},
+			nodeProjects:    []string{},
+			isError:         false,
+		},
+		{
+			description:     "Node has no projects; requested projects is empty",
+			requestProjects: []string{},
```
bcmdarroch: I'm not sure if it should be possible for the domain to ever receive a request with empty projects 🤔I think authz-service should 403 if there are no allowed projects. maybe there should be an error if the request projects are empty

tylercloke: Made an issue to clean this up in all the domains!

https://github.com/chef/automate/issues/2369

PR #2360, opened 2019-11-24
waivers trends

rickmarry:
This change adds waivers to to the trend apis, it covers both node as well as control trends and it does so on the three supported depths of filtering (control/profile/report)

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/automate-gateway/api/authz/response/authz.swagger.json
```diff
@@ -4,6 +4,10 @@
     "title": "components/automate-gateway/api/authz/response/authz.proto",
     "version": "version not set"
   },
+  "schemes": [
+    "http",
+    "https"
+  ],
```
stevendanna: I think all these changes mean that you might have an old version of some of our gprc/swagger tooling.

vjeffrey: +1 on what @stevendanna said -- i think if you try running `compile_all_protobuf_components` from the studio this will all go away

rickmarry: interesting.. I did run `compile_all_protobuf_components`  I had to run that in order to support the new fields that I've added in both the compliance as well as the gateway protos.. but I'll double-check

rickmarry: That's quite possible but shouldn't be a problem for now.. notice, this pr is to merge this into ap/inspec-waivers-ingestion (not master)

Also, I did run `compile_all_protobuf_components` (had to in order for the new fields in the protos to gen the right stubs etc..)

rickmarry: I'm not sure how to fix this.. I have run `complile_all_protobuf_components` but that doesn't prevent the generation of that "schemes" block shown above.

Note, this block is also generated if i run `complile_all_protobuf_components` on master

Is there some tooling component that I need to install that lives outside of the repo?

stevendanna: My guess is that the binaries that do the generation are out of date.  They might be cached in `AUTOMATE_REPO_ROOT/bin/`.  You can try deleting them and then re-compiling `compile_all_protobuf_components`.

rickmarry: thanks @stevendanna 
How shall I upgrade the grpc/swagger tooling?
@vjeffrey, I have run `compile_all_protobuf_components` (in fact it was necessary in order to generate the fields for the latest proto fields in compliance-service and automate-gateway components).

N.B.:  I have also run `compile_all_protobuf_components` in master to compare outcome on that json file.. running on master resulted in the same exact "schemes" block being inserted into that json file.

I am using hab version: 0.90.6/20191112141314

rickmarry: solved.. thanks @alexpop! I tried doing `hab studio rm`, thinking that would clean things up but the real solution was to delete bin/* from within hab studio and then run `compile_all_protobuf_components`

rickmarry: @stevendanna - thanks!  I see that you and @alexpop had the same good advice, and @vjeffrey also had the correct recommendation to rerun compile..  I'm all set now.. thank you all!

Issue #2358, opened 2019-11-23
govcloud support for aws node integrations


vjeffrey:
## User Story
Users would like to be able to use node integrations with their govcloud aws accounts (api + ec2). 
At first glance, it seems like allowing a user to specify a region to target when creating the integrations should be sufficient.
Things to test:
- aws api integration with govcloud creds, run a scan and ensure results are as expected
- aws ec2 integration with govcloud creds, run a node scan and ensure results are expected
- ensure required perms suggested in our docs work for govcloud
- ensure ssm jobs work with govcloud

## Definition of Done
All aws integration functionality checked with govcloud

vjeffrey:
large t shirt size

jonong1972:
Update the label:
Specify a default region to target:

Pre-fill the text field with:
us-west-1 (or whatever our default is today)


Note: I would like to revisit this for a solution to pull the list of regions to put into a drop down.

vjeffrey:
noting repro/testing steps here for later

- create IAM user in aws govcloud with the following perms:
```
{"Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "ec2:DescribeInstances",
                "ssm:*",
                "ec2:DescribeRegions",
                "sts:GetCallerIdentity",
                "ec2:DescribeInstanceStatus",
                "iam:ListAccountAliases"
            ],
            "Resource": "*"
        }
    ]
}
```
create aws integration:
- api
- ec2

- ensure both work
- ensure nodes are attached to each one
- run a scan job on each one

vjeffrey:
from my testing so far, it looks like ssm scanning does not work in govcloud

vjeffrey:
validated in dev: https://a2-dev.test/settings/node-integrations

PR #2365, opened 2019-11-26
allow user to include region when creating aws mgr

vjeffrey:
When accessing the AWS api, a region value is required. We've been defaulting to the value of env var "AWS_REGION" and using "us-east-1" when the env var is not set.
This change modifies the code to also accept (and prefer) a region value by credential.
That means that a user can now add an integration with the following api body:
```
'{"name":"test (fake) aws-api integration","type":"aws-api","credential_data":[{"key":"AWS_ACCESS_KEY_ID","value":"test"},{"key":"AWS_SECRET_ACCESS_KEY","value":"test"},{"key":"AWS_REGION","value":"us-gov-west-1"}],"instance_credentials":[]}'
```
If the region is included in the api call, we use that value. If it is not specified, we go back to the original code path of looking for the env var and then falling back to us-east-1 as a default.

https://github.com/chef/automate/issues/2358

### :+1: Definition of Done
when creating an aws integration, the region value specified in the api is passed through to all aws api calls. 

- rebuild nodemanager service, compliance service, and ui
- create an IAM user in govcloud with the following perms:
```
                "ec2:DescribeInstances",
                "ssm:*",
                "ec2:DescribeRegions",
                "sts:GetCallerIdentity",
                "ec2:DescribeInstanceStatus",
                "iam:ListAccountAliases",
                "iam:GetAccountSummary",
                "iam:ListUsers",
```
- add an api integration and specify the region needed for gov cloud (`us-gov-west-1`)
- run a scan job with the AWS profile, take a look at the results
- add a node in the govcloud account
- add an ec2 integration and specify the region needed for gov cloud (`us-gov-west-1`)
- ensure the node was added and has all the correct info (public ip, tags, name)

**note: ssm scanning is not currently supported with govcloud**

- [x] Tests added/updated?
- [x] Docs added/updated?

Screen Shot 2019-11-26 at 05 54 39

### DIFF
--- a/components/automate-chef-io/content/docs/node-integrations.md
+++ b/components/automate-chef-io/content/docs/node-integrations.md
@@ -63,12 +63,15 @@ The service makes calls to these API:
 * `STS-GetCallerIdentity`
 * `SEC2-DescribeRegions`
 * `IAM-ListAccountAliases`
+* `IAM-GetAccountSummary`
+* `IAM-ListUsers`
 
 Permissions: You'll need at least a global read permission; `arn:aws:iam::aws:policy/ReadOnlyAccess`
 
 ## AWS Credential-less Scanning with Chef Automate
 
 For users running Chef Automate 2 in EC2, we invite you to try out our "AWS-EC2 Credential-less Scanning"!
+Please note that credential-less scanning is not supported for AWS govcloud.
 
 ### Ensure Minimum Permissions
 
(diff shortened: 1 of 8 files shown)


components/automate-chef-io/content/docs/node-integrations.md
```diff
@@ -63,12 +63,15 @@ The service makes calls to these API:
...
 * `IAM-ListAccountAliases`
+* `IAM-GetAccountSummary`
+* `IAM-ListUsers`
 
 Permissions: You'll need at least a global read permission; `arn:aws:iam::aws:policy/ReadOnlyAccess`
 
 ## AWS Credential-less Scanning with Chef Automate
 
 For users running Chef Automate 2 in EC2, we invite you to try out our "AWS-EC2 Credential-less Scanning"!
+Please note that credential-less scanning is not supported for AWS govcloud.
```
phiggins: Since these are the official docs (I think?), does `govcloud` need to be capitalized any certain way?

phiggins: It looks like [for AWS it's `GovCloud`](https://aws.amazon.com/govcloud-us/)

vjeffrey: good call! ya!

PR #2367, opened 2019-11-26
Paginate and filter profile list endpoint

phiggins:
Implements server side filtering and pagination on the profiles list endpoint for #2258.

* To maintain backwards compatibility, only requests with the "per_page" parameter set to a non-default value (ie, 0) will be paginated.
* In my unscientific testing in the browser, hard coding the "per_page" value to 50 improved the duration of the available profiles request from ~3.8s to 1.8s.
* This PR introduces the use of the squirrel sql generation library to cut down on the code necessary to generate the sql for filtering.


components/compliance-service/api/tests/01_market_spec.rb
```diff
@@ -6,6 +6,15 @@
...
   def profiles ; Profiles::ProfilesService ; end
 
+  it "errors out when an invalid order field is specified" do
+    assert_grpc_error(/order field '1337' is invalid. Use either 'ASC' or 'DESC'/) do
+      GRPC profiles, :list, Profiles::Query.new(
+        sort: 'name',
+        order: 1337
+      )
+    end
+  end
```
phiggins: I fixed a small bug in this error message in #2320 and didn't add a test, and I also wanted to see what happened if I tried assigning an invalid value for this parameter to see if the error checking was even necessary. 

Maybe I'm missing something but it seems weird that an enum that only defines 0 or 1 as valid values allows assigning arbitrary values...

components/compliance-service/api/tests/run.rb
```diff
@@ -100,7 +98,9 @@ def assert_grpc_error(details, code)
       assert_equal(details, error.details)
     end
 
-    assert_equal(code, error.code)
+    if code
+      assert_equal(code, error.code)
+    end
```
phiggins: This test isn't strictly necessary, but I've never been sure what these numbers represent and to find the correct value in the past I've just used 0 then seen what appears in the failure message.

components/compliance-service/profiles/db/store.go
```diff
@@ -376,89 +379,108 @@ func (s *Store) Delete(namespace string, name string, version string) error {
...
+		Select("sha256, info, count(*) OVER() AS total").
+		From("store_profiles")
+
+	if len(req.Namespace) == 0 {
+		sql = sql.
+			Where("exists(select 1 from store_market where store_profiles.sha256 = store_market.sha256)")
+	} else {
+		sql = sql.
+			Where("exists(select 1 from store_namespace where store_profiles.sha256 = store_namespace.sha256 and store_namespace.owner = ?)", req.Namespace)
+	}
```
phiggins: The original hardcoded SQL did `where sha256 in (select sha256....` instead of the exists query. This is because the squirrel library's `Where` will do escape parameters for you but it won't accept arbitrary SQL fragments. The two queries should be equivalent and according to this issue on the squirrel bug tracker, the `exists` version should be faster in postgres: https://github.com/Masterminds/squirrel/issues/70

phiggins: I think the idiomatic way to do this would be to use a left outer join, but that seemed like a bigger change.

components/compliance-service/profiles/db/store.go
```diff
@@ -376,89 +379,108 @@ func (s *Store) Delete(namespace string, name string, version string) error {
...
+	if req.PerPage > 0 {
+		perPage := uint64(req.PerPage)
+		page := uint64(req.Page)
+
+		sql = sql.
+			Limit(perPage).
+			Offset(perPage * page)
+	}
+
+	sql = sql.OrderBy(fmt.Sprintf("info->>'%s' %s", req.Sort, req.Order), "info->>'version' DESC")
```
phiggins: The latest released version of squirrel only has `OrderBy`, which only accepts entire order by clauses. On master there's a new function `OrderByClause` which will accept parameters so we could potentially avoid the `Sprintf` call here: https://github.com/Masterminds/squirrel/blob/d13326f0be730e8aa260f440b63fd00c858a03d0/select.go#L330-L342

phiggins:
I had an idea to simplify this code but it would be a pretty big change. Currently the code pulls a bunch of small json documents from the database, parses them into on type of go struct, then converts them to another type of go struct, then returns that to the caller. This could potentially be simplified by:

* Using the second type of struct directly when parsing all of the json documents.
* Generating a SQL query to get a json document that is the same general shape as the collection of structs returned, then parsing one big document instead of many smaller ones.

components/compliance-service/profiles/db/store.go
msorens: Nice; thanks for the comment @phiggins !
You can take this further, though, if you want: instead of the hybrid (part Squirrel, part SQL) you could do nested squirrel--see https://github.com/Masterminds/squirrel/blob/7fcba632b69f0b1c1c60e71cce11bcbf6320a406/select_test.go#L203-L211 for the relevant use case.

components/compliance-service/profiles/db/store.go
```diff
@@ -376,89 +379,108 @@ func (s *Store) Delete(namespace string, name string, version string) error {
...
+	if len(req.Namespace) == 0 {
+		sql = sql.
+			Where("exists(select 1 from store_market where store_profiles.sha256 = store_market.sha256)")
+	} else {
+		sql = sql.
+			Where("exists(select 1 from store_namespace where store_profiles.sha256 = store_namespace.sha256 and store_namespace.owner = ?)", req.Namespace)
+	}
+
+	if len(req.Name) != 0 {
+		sql = sql.Where("info->>'name' = ?", req.Name)
```
msorens: What is the `->>` notation here (and likewise the `#>>` in the original)? Not familiar with that.

phiggins: They are both postgres json operators: https://www.postgresql.org/docs/9.6/functions-json.html

`#>>` is a more powerful operator to get an element at a specified path (hence the `{}` brackets in the original), `->>` is used to get the value at a top-level key. The extra `>` in both of those operators (ie `->>` vs `->`) means it'll always return a string instead of null or a string.

components/compliance-service/profiles/db/store.go
phiggins: Just from a glance at that example it doesn't look like it gains us much over the sql fragment besides just consistency in always using the library as much as possible. I'll play around with it though, I thought I remembered seeing an `Exists` method (maybe only on master?).

components/compliance-service/api/tests/01_market_spec.rb
vjeffrey: that is super weird.
maybe it's b/c it's an integer?  still weird

phiggins: I found a technical explanation for some of what is happening, but it still seems like weird behavior to me:

> During deserialization, unrecognized enum values will be preserved in the message, though how this is represented when the message is deserialized is language-dependent. In languages that support open enum types with values outside the range of specified symbols, such as C++ and Go, the unknown enum value is simply stored as its underlying integer representation.

https://developers.google.com/protocol-buffers/docs/proto3#enum

components/compliance-service/profiles/db/store.go
```diff
@@ -376,89 +379,108 @@ func (s *Store) Delete(namespace string, name string, version string) error {
...
+	Sort      string
+}
+
+func (s *Store) ListProfilesMetadata(req ProfilesListRequest) ([]inspec.Metadata, int, error) {
+	if req.Order != "ASC" && req.Order != "DESC" {
+		return nil, 0, status.Errorf(codes.InvalidArgument, "order field '%s' is invalid. Use either 'ASC' or 'DESC'", req.Order)
+	}
+
+	if req.Sort != "name" && req.Sort != "title" && req.Sort != "maintainer" {
+		return nil, 0, status.Errorf(codes.InvalidArgument, "sort field '%s' is invalid. Use either 'name', 'title' or 'maintainer'", req.Sort)
```
vjeffrey: maintainer?  should that be namespace or owner?

phiggins: Wow, yeah, that's twice now I've blindly copied and pasted this code without noticing that! Maybe this table used to have that column?

Not everything has an owner so it would be somewhat complicated to add sorting on that. I'll just delete it.

phiggins: Now that I've actually looked at the data, there is a maintainer field included in the json document, so I'll just leave it. This does affect the followup PR though so thank you for pointing it out!

vjeffrey: maintainer comes in on the inspec yml for the profile -- if it means more work, i would default to not supporting that field for sort b/c it's not really a popular one/only shown on profile detail pages

PR #2381, opened 2019-12-02
InSpec waivers in Automate groundwork

alexpop:

This is the initial change in Automate to support InSpec waivers. It is:

1. Changing the ElasticSearch mappings with extra waiver fields.
2. Migrating old ElasticSearch data to prevent search errors when historic data is accessed in the trend or via the date picker
3. Add new integration test data to enable waivers API testing
4. Update the integration tests based on the updated test data
5. Process the new waivers in ingestion
6. Add unit tests for the new ingestion logic


components/compliance-service/ingest/ingestic/mappings/comp-rep-date.go
```diff
@@ -270,6 +278,29 @@ var ComplianceRepDate = Mapping{
                     }
                   }
                 },
+                "waiver_data": {
+                  "type": "object",
+                  "properties": {
+                    "expiration_date": {
+                      "type": "text"
```
rickmarry: should this be "type": "date" instead of "type": "text"?

components/compliance-service/reporting/relaxting/es_types.go
```diff
@@ -98,6 +100,14 @@ type ESInSpecReportControlsResult struct {
 	SkipMessage string  `json:"skip_message,omitempty"`
 }
 
+type ESInSpecReportControlsWaiverData struct {
+	ExpirationDate     string `json:"expiration_date"`
```
rickmarry: should this be time.Time instead of string type?

components/compliance-service/ingest/ingestic/mappings/comp-rep-date.go
alexpop: I thought about that. It comes from InSpec as a string(i.e. "1977-06-01") and we don't use it for ElasticSearch querying since it only matters when the scan is executed. After that, we rely on the `waiver_data` object. For this reason, I was thinking to keep it simple and avoid conversions.

alexpop:
Thank you Rick for the review. I added another commit (https://github.com/chef/automate/pull/2381/commits/bec8bfa80b7eb98219fc5a1757b29eeca5c01723) in order to remove the `waived` boolean flag added in the initial version for report and profiles. Since we have a `waived` status for reports and profiles, we no longer need this boolean flag.

components/compliance-service/api/status/server/status.go
```diff
@@ -18,13 +18,14 @@ const (
...
+	MigrationLabelESa2v5 = "ElasticSearch_A2_v5"
 	MigrationLabelPG     = "PostgreSQL"
 	MigrationLabelPRO    = "Profiles"
 	//MigrationLabelFEEDS = "ElasticSearch_Feeds_1"
 )
 
-const maxMigrations = 7        // Total migrations should match the number of constants above
-const totalMigrationSteps = 28 // Max number of migration LogEntry items we can have across all migrations
+const maxMigrations = 8        // Total migrations should match the number of constants above
+const totalMigrationSteps = 36 // Max number of migration LogEntry items we can have across all migrations
```
vjeffrey: 🎉 we've seen this number get out of date quite a bit - is there something we should be doing to increase it automagically or is manual the best way to go?
Screen Shot 2019-12-03 at 12 14 36

components/compliance-service/api/tests/04_stats_spec.rb
```diff
@@ -369,9 +369,11 @@ def stats
           {
             "name" => "linux-baseline",
             "id" => "b53ca05fbfe17a36363a40f3ad5bd70aa20057eaf15a9a9a8124a84d4ef08015",
-            "failures" => 22,
-            "criticals" => 22,
-            "passed" => 23
+            "failures" => 21,
+            "criticals" => 21,
+            "passed" => 22,
+            # COMMENT THIS OUT WHEN WAIVERS API SUPPORT IS ADDED FOR THIS API CALL
```
vjeffrey: nice! that'll make it easier later!

components/compliance-service/api/status/server/status.go
alexpop: I agree we should find an automatic method to determine the maximum number of migration steps. It will require a bit of a refactor in another PR.

components/compliance-service/ingest/ingestic/mappings/comp-sum-date.go
```diff
@@ -222,6 +230,14 @@ var ComplianceSumDate = Mapping{
                   },
                   "type": "object"
                 },
+                "waived": {
+                  "properties": {
+                    "total": {
```
vjeffrey: if we're only _adding_ es mappings, why do we need to do a migration?

components/compliance-service/reporting/relaxting/es_types.go
vjeffrey: discussed in standup that we should leave as string -- this is metadata about the waived control that comes in on the report, no need to query it later

components/compliance-service/ingest/ingestic/mappings/comp-rep-date.go
alexpop: We talked about this in standup and decided to keep it as a string since we don't foresee a need in the future to treat the field as a date in the elasticsearch queries.

components/compliance-service/api/tests/04_stats_spec.rb
alexpop: Thank you. Wanted in a way to highlight why those expected values had to be changed.

components/compliance-service/ingest/ingestic/mappings/comp-sum-date.go
alexpop: We are adding new aggregated fields in ElasticSearch, like:
```
            "waived": {
              "properties": {
                "total": {
                  "type": "integer"
                }
              },
```
which the elasticsearch queries (trend or back in time via date picker) for the compliance API calls will target. If those fields are missing for the old data, the queries will error out instead of saying 0 waived nodes 1 year ago.

vjeffrey: gotcha! thanks for the explanation!

alexpop:
Don't merge yet. I'm running over a few migrations manually from different versions. Verifying at the final elasticsearch data for the new fields.

alexpop:
I pushed another commit to make `controls_sums.waived.total` explicitly 0 for old data using the reindex script. I ran all the migrations that we have and inspected the data. Looks good and consistent. Once we start exposing the waived bits in the API calls, we should ensure the migration tests (i.e. `61_A2V1_migration_spec.rb`) have one or two tests that look for waivers info.

PR #2387, opened 2019-12-02
inspec-agent node add cleanup

vjeffrey:
1) do not add node from the resolver. it will get added to the db when
it is ingested
2) make sure the manager (type and id) is included in the process node
func to ensure we correctly display data for the nodes

I was seeing some weird messages in the logs, so I decided to try to track them down and read through some of the code. I discovered we were doing an `add node` function from the inspec-agent resolver that was calling ProcessNode. But that call was getting rejected b/c ProcessNode is supposed to be called with some run or scan data. 
The node will eventually get added to the db upon ingestion, so it is unnecessary to do this call.
Also making sure the manager id/type get included in the process node function so the nodes are correctly associated with a manager. 

N/A

### :+1: Definition of Done
no more of these msgs in the logs: 
Screen Shot 2019-11-25 at 15 36 29

- add a node in ec2. make sure you can ssh into it.
- create a node integration (aws-ec2), create a node credential
- ensure you do the whole "tags - node credentials" thing 
- delete all the nodes assoc with the integration
- run a scan job on the integration
- expect the node you can ssh into to show up in the scan job nodes with the correct information

- [ ] Tests added/updated?
- [n/a] Docs added/updated?


vjeffrey:
opened this in draft mode cause i still need to add tests

components/compliance-service/inspec-agent/resolver/resolver.go
```diff
@@ -125,10 +126,11 @@ func (r *Resolver) handleAzureApiNodes(ctx context.Context, m *manager.NodeManag
 				logrus.Debugf("handleManagerNodes unable to get node id: %s", err.Error())
 			}
 			if len(nodeUUID) == 0 {
-				nodeUUID, err = r.scannerServer.AddNodeToDB(ctx, &inspec.OSInfo{}, &baseJob, scanner.SourceInfo{SourceAccountID: m.AccountId, SourceRegion: node.ID, SourceID: node.ID}, time.Time{})
+				id, err := uuid.NewV4()
```
vjeffrey: question for myself: do we need to do anything about creating a uuid here? why? won't it get done later in `ProcessNode`?

components/nodemanager-service/pgdb/nodes.go
```diff
@@ -958,10 +962,12 @@ func (db *DB) UpdateNodeDetectInfo(ctx context.Context, nodeDetectInfo *nodes.No
 	if err != nil {
 		return errors.Wrap(err, "UpdateNodeDetectInfo unable to read rows affected")
 	}
-	if rows == 1 {
-		logrus.Debugf("updated node %s", nodeDetectInfo.NodeId)
+	if rows == 0 {
```
vjeffrey: i was seeing the "unexpected results" msg a lot. seems like that's actually because 0 nodes were being updated but we weren't explicitly checking for that case

components/compliance-service/inspec-agent/resolver/resolver.go
vjeffrey: answer: if there's no node uuid, the report will get rejected! so we need this

components/nodemanager-service/pgdb/node-managers-process-node.go
```diff
@@ -123,31 +129,39 @@ func (db *DB) ProcessIncomingNode(node *manager.NodeMetadata) error {
 		return errors.Wrap(err, "ProcessIncomingNode unable to marshal projects data")
 	}
 
+	var mgrType string
+	if node.ManagerId != "" {
+		mgrType, err = db.SelectStr(sqlGetManagerTypeFromId, node.ManagerId)
+		if err != nil {
+			logrus.Warnf("unable to find manager for node")
+		}
+	}
```
phiggins: It'd probably be more efficient to do this with a join, but modifying the 4 insert queries does not sound like fun. :(

vjeffrey: agreeeed

PR #2391, opened 2019-12-02
Fail purge if not executed within 10 mins of being enqueued

jaym:
There was a bug in a release of automate where purge workflows
would be queued up, but would not run because nothing was listening for
them. When we fixed this, people were confused by why automate was
deleting data. More confusingly, the policy that would be used was the
old one and not the current one as a workflow had been queued but not
executed. We don't want these to run, so if each step of the purge does
not happen within 10 mins of when it was requested to happen, we fail.

### DIFF
--- a/lib/datalifecycle/purge/workflow.go
+++ b/lib/datalifecycle/purge/workflow.go
@@ -2,14 +2,22 @@ package purge
 
 import (
 	"context"
+	"time"
 
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"github.com/teambition/rrule-go"
 
 	es "github.com/chef/automate/api/interservice/es_sidecar"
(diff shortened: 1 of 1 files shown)


lib/datalifecycle/purge/workflow.go
```diff
@@ -2,14 +2,20 @@ package purge
...
 
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 	"github.com/teambition/rrule-go"
 
 	es "github.com/chef/automate/api/interservice/es_sidecar"
 	"github.com/chef/automate/lib/cereal"
 )
 
+const failPurgeIfExecutedAfter = 10 * time.Minute
```
yzl: [nit] Future-us might like a comment about how we picked 10.

PR #2392, opened 2019-12-02
[automate-2366] Remove introspection from individual projects on list page

msorens:

This PR removes introspection around the control menu on each project for the time being due to two issues.
1. Parameterized introspection is not yet set up for projects.
2. A back-end call for each resource is too much overhead.

N/A

### :+1: Definition of Done
Network calls to the introspect endpoint do not occur for each project in the projects list.

Rebuild automate-ui.
Watch in the network tab when you reload the projects page that introspect calls on each specific project do not occur.

### DIFF
--- a/components/automate-ui/src/app/modules/policy/list/policy-list.component.html
+++ b/components/automate-ui/src/app/modules/policy/list/policy-list.component.html
@@ -29,7 +29,7 @@
                       <chef-table-cell class="type-column">{{ policy.type | iamType }}</chef-table-cell>
                       <chef-table-cell class="status-column">{{ status(policy) | capitalize }}</chef-table-cell>
                       <chef-table-cell class="three-dot-column">
-                      <!-- <app-authorized [allOf]="['/policies/{policy}', 'delete', policy]"> -->
+                      <!-- <app-authorized [allOf]="[`/iam/v2beta/policies/{id}`, 'delete', policy.id]"> -->
                         <chef-control-menu *ngIf="policy.type !== 'CHEF_MANAGED'">
                           <chef-option (click)="startPolicyDelete(policy)">Delete Policy</chef-option>
                         </chef-control-menu>
@@ -40,7 +40,7 @@
               </chef-table-new>
             </app-authorized>
           </section>
-      
+
           <app-delete-object-modal
             [visible]="deleteModalVisible"
             objectNoun="policy"
--- a/components/automate-ui/src/app/modules/roles/list/roles-list.component.html
+++ b/components/automate-ui/src/app/modules/roles/list/roles-list.component.html
@@ -38,9 +38,11 @@
                   <chef-table-cell></chef-table-cell>
                   <chef-table-cell>{{ role.type | iamType }}</chef-table-cell>
                   <chef-table-cell class="three-dot-column">
+                    <!-- <app-authorized [allOf]="[`/iam/v2beta/roles/{id}`, 'delete', role.id]"> -->
                     <chef-control-menu *ngIf="(role.type | iamType).toLowerCase() === 'custom'">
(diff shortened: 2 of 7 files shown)


components/automate-ui/src/app/modules/policy/list/policy-list.component.html
```diff
@@ -29,7 +29,7 @@
                       <chef-table-cell class="type-column">{{ policy.type | iamType }}</chef-table-cell>
                       <chef-table-cell class="status-column">{{ status(policy) | capitalize }}</chef-table-cell>
                       <chef-table-cell class="three-dot-column">
-                      <!-- <app-authorized [allOf]="['/policies/{policy}', 'delete', policy]"> -->
+                      <!-- <app-authorized [allOf]="[`/iam/v2beta/policies/{id}`, 'delete', policy.id]"> -->
```
msorens: PR Bonus: Added relevant consistency cleanup in this PR for the **_non_**-project pages.
Here, you will note, before and after are both commented out, but the content has been corrected.

components/automate-ui/src/app/modules/roles/list/roles-list.component.html
```diff
@@ -38,9 +38,11 @@
                   <chef-table-cell></chef-table-cell>
                   <chef-table-cell>{{ role.type | iamType }}</chef-table-cell>
                   <chef-table-cell class="three-dot-column">
+                    <!-- <app-authorized [allOf]="[`/iam/v2beta/roles/{id}`, 'delete', role.id]"> -->
```
msorens: Still part of the consistency cleanup, by adding this even as commented out code, it helps us remember that the same code around the control menu should go here eventually.

components/automate-ui/src/app/modules/token/token-list/api-token-list.component.html
```diff
@@ -47,7 +47,7 @@
               <chef-table-header-cell *ngIf="isIAMv2$ | async">Projects</chef-table-header-cell>
               <chef-table-header-cell>Created Date</chef-table-header-cell>
               <chef-table-header-cell>Status</chef-table-header-cell>
-              <chef-table-header-cell class="controls"></chef-table-header-cell>
+              <chef-table-header-cell class="three-dot-column"></chef-table-header-cell>
```
msorens: PR Bonus: A recent PR changed all the other resources to "three-dot-column". This one was likely added right about the same time, using the old name, so updated this here.

components/automate-ui/src/app/pages/project/list/project-list.component.html
```diff
@@ -45,11 +45,11 @@
                   <chef-table-cell>{{ project.id }}</chef-table-cell>
                   <chef-table-cell>{{ getRulesStatus(project) }}</chef-table-cell>
                   <chef-table-cell class="three-dot-column">
-                    <app-authorized [allOf]="['/iam/v2beta/projects/{id}', 'delete', project.id]">
+                    <!-- <app-authorized [allOf]="['/iam/v2beta/projects/{id}', 'delete', project.id]"> -->
```
msorens: And finally, the actual fix for this PR: comment out this call.

components/automate-ui/src/app/modules/policy/list/policy-list.component.html
tylercloke: Can we just delete this commented code for now? We don't even know if this is what the final form will be.

components/automate-ui/src/app/modules/roles/list/roles-list.component.html
tylercloke: Heh I was gonna ask for the opposite. We don't even know if this will look like and we know it needs to happen everywhere. Commented code sitting around is a pet peeve of mine though so w/e works.

Issue #2389, opened 2019-12-02
Domain services Project delete Cereal orchestration


lancewf:
## User Story

As a user of IAM and Projects, when I delete a project, I expect that the project I deleted will be removed from all resources. Currently, when I delete a project, a reference to that project remains on ingested resources, which causes those resources to NOT appear in the (unassigned) filter like I would expect them to.

This subtask is to kick off the project delete for all the compliance-service, ingest-service, and nodemanager-service domain services with cereal. Look into adding this to the graveyarding cereal task like the team-service is using. 

## Definition of Done

When a project is deleted the compliance-service, ingest-service, and nodemanager-service all get notified to start deleting the project from their resources.

Issue #2384, opened 2019-12-02
Client Runs Project Delete


lancewf:
## User Story

As a user of IAM and Projects, when I delete a project, I expect that the project I deleted will be removed from all Client Runs resources. Currently, when I delete a project, a reference to that project remains for the client runs nodes and actions, which causes those resources to NOT appear in the (unassigned) filter like I would expect them to.

The "node-state" and the "actions-*" indexes need to be updated. 

The overall issue is https://github.com/chef/automate/issues/2016

## Definition of Done
When a project is deleted all references to that project are removed from Client Runs nodes and Actions.
## Demo Script / Repro Steps
1. Create a Client Runs node and an action
1. Create a project with ingestion rules that match both the node and action
1. Apply the new project created. 
1. Ensure the node and action are tagged with the above project
1. Delete the project
1. Ensure the node and action do not have the deleted project associated with them.

PR #2394, opened 2019-12-02
Client Runs and Chef Server Actions project delete

lancewf:
Creating a function to delete a project from Client Runs and Chef Server actions.  

#2384

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/compliance-service/ingest/ingestic/ingestic.go
+++ b/components/compliance-service/ingest/ingestic/ingestic.go
@@ -191,6 +191,12 @@ func (backend *ESClient) setDailyLatestToFalse(ctx context.Context, nodeId strin
 	return errors.Wrap(err, "setDailyLatestToFalse")
 }
 
+func (backend *ESClient) DeleteProjectTag(ctx context.Context,
+	projectTagToBeDelete string) ([]string, error) {
+	// TODO implement
+	return []string{}, nil
+}
+
 func (backend *ESClient) UpdateProjectTags(ctx context.Context, projectTaggingRules map[string]*iam_v2.ProjectRules) ([]string, error) {
 	logrus.Debug("starting project updater")
 
--- a/components/ingest-service/backend/client.go
+++ b/components/ingest-service/backend/client.go
@@ -74,6 +74,8 @@ type Client interface {
 	JobCancel(context.Context, string) error
 
 	UpdateProjectTags(context.Context, map[string]*iam_v2.ProjectRules) ([]string, error)
+	// @param (context, projectTagToDelete)
+	DeleteProjectTag(context.Context, string) ([]string, error)
 
 	// Migration contracts
 	ReindexInsightstoConvergeHistory(context.Context, string, string) error
--- a/components/ingest-service/backend/client.go
+++ b/components/ingest-service/backend/client.go
@@ -64,6 +64,12 @@ type Client interface {
 	CreateBulkActionRequest(InternalChefAction) elastic.BulkableRequest
 	// @param (context, bulkableRequests)
 	SendBulkRequest(context.Context, []elastic.BulkableRequest) error
+
+	// IAM Projects contracts
+	// @param (context, projectTagToDelete)
+	DeleteActionProjectTag(context.Context, string) (string, error)
+	// @param (context, projectTagToDelete)
+	DeleteNodeProjectTag(context.Context, string) (string, error)
 	// @param (context, projectRules)
 	UpdateNodeProjectTags(context.Context, map[string]*iam_v2.ProjectRules) (string, error)
 	// @param (context, projectRules)
@@ -72,7 +78,7 @@ type Client interface {
 	JobStatus(context.Context, string) (project_update_lib.JobStatus, error)
 	// @param (context, jobID)
 	JobCancel(context.Context, string) error
-
+	// @param (context, projectRules)
 	UpdateProjectTags(context.Context, map[string]*iam_v2.ProjectRules) ([]string, error)
 	// @param (context, projectTagToDelete)
 	DeleteProjectTag(context.Context, string) ([]string, error)
--- a/components/ingest-service/backend/elastic/actions.go
+++ b/components/ingest-service/backend/elastic/actions.go
@@ -35,13 +35,7 @@ func (es *Backend) DeleteActionProjectTag(ctx context.Context,
 	projectTagToBeDelete string) (string, error) {
 	script := `
 		if (ctx._source['projects'] != null) {
-			def foundIndex = -1;
-			for (def index = 0; index < ctx._source.projects.size(); index++) {
-				if (ctx._source.projects[index] == params.project) {
-					foundIndex = index;
-					break;
-				}
-			}
+			def foundIndex = ctx._source.projects.indexOf(params.project);
 
 			if (foundIndex != -1) {
 				ctx._source.projects.remove(foundIndex);
--- a/components/ingest-service/backend/elastic/actions.go
+++ b/components/ingest-service/backend/elastic/actions.go
@@ -31,6 +31,33 @@ func (es *Backend) CreateBulkActionRequest(action backend.InternalChefAction) el
 	return es.createBulkUpdateRequestToTimeseriesIndex(mapping, time, action)
 }
 
+func (es *Backend) DeleteActionProjectTag(ctx context.Context,
+	projectTagToBeDelete string) (string, error) {
+	script := `
+		if (ctx._source['projects'] != null) {
+			ctx._source.projects.remove(params.project);
+		}
+	`
+
+	params := map[string]interface{}{"project": projectTagToBeDelete}
+
+	index := fmt.Sprintf("%s-%s", mappings.Actions.Index, "*")
+
+	startTaskResult, err := elastic.NewUpdateByQueryService(es.client).
+		Index(index).
+		Type(mappings.Actions.Type).
+		Script(elastic.NewScript(script).Params(params)).
+		WaitForCompletion(false).
+		ProceedOnVersionConflict().
+		DoAsync(ctx)
+
+	if err != nil {
+		return "", err
+	}
+
+	return startTaskResult.TaskId, nil
+}
+
 func (es *Backend) UpdateActionProjectTags(ctx context.Context,
 	projectTaggingRules map[string]*iam_v2.ProjectRules) (string, error) {
 	script := `
--- a/components/ingest-service/backend/elastic/actions.go
+++ b/components/ingest-service/backend/elastic/actions.go
@@ -35,7 +35,17 @@ func (es *Backend) DeleteActionProjectTag(ctx context.Context,
 	projectTagToBeDelete string) (string, error) {
 	script := `
 		if (ctx._source['projects'] != null) {
-			ctx._source.projects.remove(params.project);
+			def foundIndex = -1;
+			for (def index = 0; index < ctx._source.projects.size(); index++) {
+				if (ctx._source.projects[index] == params.project) {
+					foundIndex = index;
+					break;
+				}
+			}
+
+			if (foundIndex != -1) {
+				ctx._source.projects.remove(foundIndex);
+			}
 		}
 	`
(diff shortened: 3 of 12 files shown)


components/nodemanager-service/pgdb/project_update.go
```diff
@@ -95,6 +95,12 @@ func (projectUpdate *ProjectUpdate) updateFail(err error) {
 	projectUpdate.running = false
 }
 
+func (db *DB) DeleteProjectTag(ctx context.Context,
+	projectTagToBeDelete string) ([]string, error) {
+	// TODO implement
+	return []string{}, nil
```
lancewf: @msorens This is where the project delete would start for the nodemanager.

components/compliance-service/ingest/ingestic/ingestic.go
```diff
@@ -191,6 +191,12 @@ func (backend *ESClient) setDailyLatestToFalse(ctx context.Context, nodeId strin
 	return errors.Wrap(err, "setDailyLatestToFalse")
 }
 
+func (backend *ESClient) DeleteProjectTag(ctx context.Context,
+	projectTagToBeDelete string) ([]string, error) {
+	// TODO implement
```
lancewf: This will be completed with this issue https://app.zenhub.com/workspaces/automate-auth-5d1c1073e0945f63bfe3195b/issues/chef/automate/2388

components/ingest-service/backend/elastic/actions.go
```diff
@@ -31,6 +31,43 @@ func (es *Backend) CreateBulkActionRequest(action backend.InternalChefAction) el
...
+				if (ctx._source.projects[index] == params.project) {
+					foundIndex = index;
+					break;
+				}
+			}
+
+			if (foundIndex != -1) {
+				ctx._source.projects.remove(foundIndex);
+			}
+		}
```
lancewf: `ctx._source.projects.remove(projectTagToBeDelete)` does not work because the `remove` function only takes an index of the item to remove.

lib/authz/project_update_manager.go
```diff
@@ -14,10 +14,12 @@ const (
 	maxNumberOfConsecutiveFails = 10
 )
 
-type EsClient interface {
+// ProjectTaggedDomainService - Each domain service with tagged resources needs to provide these functions
+type ProjectTaggedDomainService interface {
```
lancewf: Refactoring because the name `EsClient` does not fit any longer.

components/ingest-service/backend/elastic/elastic.go
```diff
@@ -42,6 +42,24 @@ func (es *Backend) Initializing() bool {
 	return !es.initialized
 }
 
+func (es *Backend) DeleteProjectTag(ctx context.Context, projectTagToBeDelete string) ([]string, error) {
```
blakestier: s/projectTagToBeDelete/projectTagToBeDeleted

blakestier: or `projectTagToDelete`

components/ingest-service/backend/elastic/elastic.go
```diff
@@ -42,6 +42,24 @@ func (es *Backend) Initializing() bool {
...
+	log.Debugf("starting project delete project %q", projectTagToBeDelete)
```
blakestier: Wonder if this message should go later . . . or if we should distinguish initializing the job vs starting it? maybe i'm overthinking

components/ingest-service/backend/elastic/elastic.go
```diff
@@ -42,6 +42,24 @@ func (es *Backend) Initializing() bool {
...
+	if err != nil {
+		return []string{}, errors.Wrap(err, "Failed to start Elasticsearch Node project tags delete")
+	}
+
+	esActionJobID, err := es.DeleteActionProjectTag(ctx, projectTagToBeDelete)
+	if err != nil {
+		return []string{}, errors.Wrap(err, "Failed to start Elasticsearch Action project tags delete")
+	}
+
+	log.Debugf("Started Project delete with node job ID: %q action job ID %q", esNodeJobID, esActionJobID)
```
blakestier: When I read this, it looks like we could have successfully deleted the project tag from Elasticsearch nodes but failed to do so from Elasticsearch Actions. Does that leave things in a confusing state? Do they need to both happen, or should we at least have an indication that one succeeded? (also i guess none of this is suggestive of success, just starting)

components/ingest-service/integration_test/project_delete_actions_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+func TestProjectDeleteActions(t *testing.T) {
+	var (
+		ctx = context.Background()
+	)
+
+	cases := []struct {
+		description         string
+		existingProjectIDs  []string
+		projectIDToBeDelete string
```
blakestier: here, too (`projectIDToBeDelete(d)`)

components/ingest-service/integration_test/project_delete_actions_test.go
```diff
@@ -0,0 +1,93 @@
...
+				esJobID, err := suite.ingest.DeleteActionProjectTag(ctx, test.projectIDToBeDelete)
+				require.Nil(t, err)
+
+				waitForJobToComplete(t, esJobID)
+
+				suite.RefreshIndices(fmt.Sprintf("%s-%s", mappings.Actions.Index, "*"))
+
+				// assert the node's project IDs
+				actualActions, err := suite.GetActions(100)
+				require.Nil(t, err)
```
blakestier: we sometimes switch this out for `require.NoError` (https://godoc.org/github.com/stretchr/testify/require#NoError), fwiw

lib/authz/project_rules_update_workflow.go
```diff
@@ -193,7 +193,7 @@ func (m *ESCancelUpdateProjectTagsTask) Run(
 }
 
 type ESStartProjectTagUpdaterTask struct {
-	esClient            EsClient
+	domainService       ProjectTaggedDomainService
```
blakestier: 🎉

blakestier: What are the potential domain services using this? I wonder if a comment would be helpful.

components/ingest-service/backend/elastic/actions.go
blakestier: 😢

components/ingest-service/backend/elastic/actions.go
```diff
@@ -31,6 +31,43 @@ func (es *Backend) CreateBulkActionRequest(action backend.InternalChefAction) el
...
+	projectTagToBeDelete string) (string, error) {
+	script := `
+		if (ctx._source['projects'] != null) {
+			def foundIndex = -1;
+			for (def index = 0; index < ctx._source.projects.size(); index++) {
+				if (ctx._source.projects[index] == params.project) {
+					foundIndex = index;
+					break;
+				}
+			}
```
msorens: I think this might do your lookup more concisely (assuming projects is an `ArrayList`)--it also returns `-1` if not found:
```
def foundIndex = ctx._source.projects.indexOf(params.project)
```
Reference: https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#indexOf-java.lang.Object-

(Same comment for nodes.go below)

components/ingest-service/integration_test/project_delete_actions_test.go
```diff
@@ -0,0 +1,93 @@
...
+				// Send a project rules update event
+				esJobID, err := suite.ingest.DeleteActionProjectTag(ctx, test.projectIDToBeDelete)
+				require.Nil(t, err)
+
+				waitForJobToComplete(t, esJobID)
+
+				suite.RefreshIndices(fmt.Sprintf("%s-%s", mappings.Actions.Index, "*"))
+
+				// assert the node's project IDs
+				actualActions, err := suite.GetActions(100)
```
msorens: What is the 100?

components/ingest-service/backend/elastic/elastic.go
lancewf: If the starting of one of the jobs fails most likely the other will fail. This would normally happen when Elasticsearch is in a bad state. This does leave things in a confusing state, but if we can not start the delete job there is not much else we can do. The only way this confusing state will be cleaned up is by running an apply-rules request.

components/ingest-service/integration_test/project_delete_actions_test.go
```diff
@@ -0,0 +1,93 @@
...
+				action := iBackend.InternalChefAction{
+					Projects:   test.existingProjectIDs,
+					RecordedAt: time.Now(),
+				}
+				suite.IngestActions([]iBackend.InternalChefAction{action})
+				defer suite.DeleteAllDocuments()
+
+				// Send a project rules update event
+				esJobID, err := suite.ingest.DeleteActionProjectTag(ctx, test.projectIDToBeDelete)
+				require.Nil(t, err)
```
msorens: Prefer to use `NoError` and add a brief message, e.g.:
```suggestion
				require.NoError(t, err, "rules update event failed")
```

components/ingest-service/backend/elastic/elastic.go
lancewf: Sorry, I don't really know the difference between initializing and starting a job. This log is mostly to show that this function was called.

components/ingest-service/integration_test/project_delete_actions_test.go
lancewf: Good catch. I normally do also, but I did not notice because I copy this code from an old test.

lib/authz/project_rules_update_workflow.go
lancewf: Currently nodmanager, ingest-servce, and compliance-service. Right, I did not think about the auth domain services.

components/ingest-service/backend/elastic/actions.go
lancewf: Thank you, I will try it out.

components/ingest-service/integration_test/project_delete_actions_test.go
lancewf: The function needs a number so 2 or higher would work.

components/ingest-service/backend/elastic/actions.go
```diff
@@ -31,6 +31,37 @@ func (es *Backend) CreateBulkActionRequest(action backend.InternalChefAction) el
 	return es.createBulkUpdateRequestToTimeseriesIndex(mapping, time, action)
 }
 
+func (es *Backend) DeleteActionProjectTag(ctx context.Context,
+	projectTagToBeDelete string) (string, error) {
+	script := `
```
bcmdarroch: this is painless script?

lancewf: That is correct.

components/ingest-service/backend/client.go
```diff
@@ -72,8 +78,10 @@ type Client interface {
 	JobStatus(context.Context, string) (project_update_lib.JobStatus, error)
 	// @param (context, jobID)
 	JobCancel(context.Context, string) error
-
+	// @param (context, projectRules)
 	UpdateProjectTags(context.Context, map[string]*iam_v2.ProjectRules) ([]string, error)
+	// @param (context, projectTagToDelete)
+	DeleteProjectTag(context.Context, string) ([]string, error)
```
msorens: I can do this in the follow-up PR for wiring things together, but need to adjust the signature to this:
```suggestion
	PurgeProject(context.Context, string) error
```

lancewf: I think we will need the `[]string` return because that is the job IDs of the running task in Elasticsearch. We can use these IDs to check when the job is complete or for errors.

msorens: Thanks for that pointer. In my PR I will encapsulate that check within `PurgeProject` itself. TL;DR can merge yours as is and my PR will make the necessary modifications. Thanks!

PR #2399, opened 2019-12-03
Don't use hab installer script to install hab

jaym:


stevendanna:
I've rebased this.  There are a few things I'd like to do before merging:

- [ ] Write `hab` to a temporary file name to avoid potentially confusing bugs that could be caused by overwriting the same file.
- [ ] Look into whether we can effectively remove the need for the habViaHab path

PR #2402, opened 2019-12-03
Jamie/2380 lost input style

### LINKED ISSUE #2380: some inputs have lost their styles
## Overview
The profile search input, along with some of the inputs on node integrations seem to have lost there styles.

### Current
profile-search-current

### Expected
profile-serach-expected

SEAjamieD:
Somewhere during the layout updates we lost input styles on a random assortment of inputs.  When I dug deeper it turns out that it is because the chef-input directive styles only get loaded when at least one of the inputs on that page is wrapped in a `chef-form-field`.

In this branch I copied over the .chef-input styles to the global scope.  I feel that ultimately we want to have unified inputs all across the site and that global scope may be the best place for them.

fixes #2380 

### :+1: Definition of Done
input styles are uniform again

navigate to https://a2-dev.test/settings/node-integrations/add and see that all inputs are back to normal
navigate to https://a2-dev.test/compliance/compliance-profiles and see that profile search is back to normal.

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/automate-ui/src/styles.scss
```diff
@@ -204,3 +204,33 @@ chef-loading-spinner[fixed] {
 input {
   line-height: normal;
 }
+
+// chef-input directive styles here when NOT wrapped in chef-form-field
+.chef-input {
```
SEAjamieD: @Shadae - at this point I almost feel like we should get rid of chef-input directive and just place this under input.  We want all of our inputs to be consistent across the site - so changing this to all be under <input> would make sense to me.

PR #2404, opened 2019-12-03
Don't detect broken packages in tests

jaym:
Turning this off for now. Once we see it fail, we can come up with a
more targeted fix. The current detection and fix is too heavy handed and
is causing failures for other reasons


stevendanna:
I'm gonna close this for now.  We aren't seeing this fail often right now.

Issue #2383, opened 2019-12-02
Node Manager Project delete


lancewf:
## User Story

As a user of IAM and Projects, when I delete a project, I expect that the project I deleted will be removed from all resources in the Node Manager. Currently, when I delete a project, a reference to that project remains for ingested client runs and compliance nodes, which causes those resources to NOT appear in the (unassigned) filter like I would expect them to.

The overall issue is https://github.com/chef/automate/issues/2016

## Definition of Done

When a project is deleted all references in the Node Manager are removed

## Demo Script / Repro Steps
1. Create a compliance and client runs node
1. Create a project that matches both nodes
1. Apply the new project created. 
1. Ensure the two nodes are tagged with the above project in the node manager
1. Delete the project
1. Ensure the two nodes do not have the deleted project associated with them.

PR #2411, opened 2019-12-04
Adding project delete for the nodemanager

lancewf:

Creating a function to delete an IAMv2 project from nodemanager nodes.

#2383 

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/nodemanager-service/pgdb/project_delete.go
+++ b/components/nodemanager-service/pgdb/project_delete.go
@@ -5,18 +5,13 @@ import (
 
 	project_update_lib "github.com/chef/automate/lib/authz"
 	uuid "github.com/chef/automate/lib/uuid4"
-	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 )
 
 const deleteProjectNodeAssociation = `
 DELETE FROM nodes_projects np
-WHERE np.project_id = $1;
-`
-
-const deleteProject = `
-DELETE FROM projects p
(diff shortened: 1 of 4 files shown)


components/nodemanager-service/pgdb/project_delete.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func (db *DB) DeleteProjectTag(ctx context.Context,
+	projectTagToBeDelete string) ([]string, error) {
+	jobUUID := uuid.Must(uuid.NewV4()).String()
+	log.Debugf("Running node manager DeleteProjectTag %v ID: %s", projectTagToBeDelete, jobUUID)
+
+	db.ProjectDelete.start(jobUUID)
+
+	go db.deleteProject(projectTagToBeDelete)
```
lancewf: We do not really need to run this delete async, but we will to keep the same pattern as the other services.

components/nodemanager-service/pgdb/project_update.go
```diff
@@ -123,7 +117,20 @@ func (db *DB) UpdateProjectTags(ctx context.Context,
 
 // JobStatus - get the job status of the project update
 func (db *DB) JobStatus(ctx context.Context, jobID string) (project_update_lib.JobStatus, error) {
```
lancewf: This is where we can check the status of both a project update and project delete status.

Issue #2016, opened 2019-10-28
Project Deletion: Ingested Resources


susanev:
## Overview

As a user of IAM and Projects, when I delete a project, I expect that the project I deleted will be removed from all resources. Currently, when I delete a project, a reference to that project remains on ingested resources, which causes those resources to **NOT** appear in the `(unassigned)` filter like I would expect them to.

## Design Details

Ideally, we can remove those project ids from ingested resources without needing to do a project update.

sdelano:
I was under the impression that project deletion triggered removal of the project from all resources, including the ones that ran through ingestion. If that's the case, do we need this?

susanev:
It only triggers removal from non-infected resources

sdelano:
Implementation notes from discussion today:

```
- when a user deletes a project via the API or the UI:
  - there's a flag in the projects table that says the project is
    marked for delete
    - if marked, the project won't show up via the APIs
    - doesn't show up in GET APIs
    - maybe we do need a way to show it in the API so they know they
      what changed
  - there's NO API right now for determining whether we need to apply rules
    - we could implement this if we want the banner eventually
  - on 'Update Projects'
    - insert the project into the graveyard
    - delete the project first
      - disappears from the table
      - cascade delete of all the rules and other stuff
    - need to ensure that the listProjectRules API doesn't have the project
    - move the staged rules
      - is there an edge case here?
    - cereal jobs
      - updating the compliance and ingest domains
      - deletes project from teams and tokens
      - remove project from the graveyard after all the cereal things
        are done
      - note: calculation of done % is going to need to be updated
```

susanev:
i guess the other thing we discussed at some point was to prevent the deletion of a project until it had zero ingest rules and was up-to-date... that may be another approach to consider.

sdelano:
*NOTE*: The implementation detail discussion above relates to the previous iteration of this card that framed the issue around the 'Update Projects' button instead of the expectations of the user. Please disregard the implementation above and let's take a look at this again with a fresh approach.

sdelano:
Notes from sprint planning:
- need to delete the project from postgres and all the rules associated with it
- have a job that goes through and deletes the project references from the ES data
  - can be run via a painless script in ES
  - should be run via a cereal job by the services responsible for the data, e.g. ingest service, compliance, config mgmt
  
Failure cases and race conditions:
- what happens if project update is happening at the same time as deletion?


Painless update looks **something like**:

```
cat <<EOF | 
{
"script": {
[...]

PR #2412, opened 2019-12-04
[automate-2388] Deleting project tags from Compliance resources

bcmdarroch:
Adds a function to delete a project from Compliance reports and summaries.

branched off: https://github.com/chef/automate/pull/2394

### :+1: Definition of Done
- a deleted project can be removed from Compliance reports and summaries (without having to go through a whole project update)

- depends on https://github.com/chef/automate/issues/2389 to exercise
- until that's ready, we can just run tests locally:
```
# in the studio, with all services up
hab pkg install -b core/go
export ELASTICSEARCH_URL="http://localhost:10141"

# reports only
go test -v -p 1 --parallel=1 -count=1 github.com/chef/automate/components/compliance-service/integration_test -run TestReportProjectDelete

# summaries only
go test -v -p 1 --parallel=1 -count=1 github.com/chef/automate/components/compliance-service/integration_test -run TestSummaryProjectDelete

# both reports and summaries
go test -v -p 1 --parallel=1 -count=1 github.com/chef/automate/components/compliance-service/integration_test -run TestProjectDelete
```

- [x] Tests added/updated?

### DIFF
--- a/components/compliance-service/ingest/ingestic/ingestic.go
+++ b/components/compliance-service/ingest/ingestic/ingestic.go
@@ -233,6 +233,7 @@ func (backend *ESClient) DeleteReportProjectTag(ctx context.Context, projectTagT
 		Index(index).
 		Type(docType).
 		Script(elastic.NewScript(deleteProjectTagScript).Params(params)).
+		Refresh("true").
 		WaitForCompletion(false).
 		ProceedOnVersionConflict().
 		DoAsync(ctx)
@@ -255,6 +256,7 @@ func (backend *ESClient) DeleteSummaryProjectTag(ctx context.Context, projectTag
 		Index(index).
 		Type(docType).
 		Script(elastic.NewScript(deleteProjectTagScript).Params(params)).
+		Refresh("true").
 		WaitForCompletion(false).
 		ProceedOnVersionConflict().
 		DoAsync(ctx)
--- a/components/compliance-service/ingest/ingestic/ingestic.go
+++ b/components/compliance-service/ingest/ingestic/ingestic.go
@@ -242,7 +242,32 @@ func (backend *ESClient) DeleteReportProjectTag(ctx context.Context, projectTagT
 }
 
 func (backend *ESClient) DeleteSummaryProjectTag(ctx context.Context, projectTagToDelete string) (string, error) {
-	return "", nil
+	script := `
+	if (ctx._source['projects'] != null) {
+		def foundIndex = ctx._source.projects.indexOf(params.project);
+
+		if (foundIndex != -1) {
+			ctx._source.projects.remove(foundIndex);
+		}
+	}
+`
+	params := map[string]interface{}{"project": projectTagToDelete}
+	docType := mappings.DocType
+	mapping := mappings.ComplianceSumDate
+	index := fmt.Sprintf("%s-%s", mapping.Index, "*")
+
+	startTaskResult, err := elastic.NewUpdateByQueryService(backend.client).
+		Index(index).
+		Type(docType).
+		Script(elastic.NewScript(script).Params(params)).
+		WaitForCompletion(false).
+		ProceedOnVersionConflict().
+		DoAsync(ctx)
+
+	if err != nil {
+		return "", err
+	}
+	return startTaskResult.TaskId, nil
 }
 
 func (backend *ESClient) UpdateProjectTags(ctx context.Context, projectTaggingRules map[string]*iam_v2.ProjectRules) ([]string, error) {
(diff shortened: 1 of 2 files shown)


components/compliance-service/ingest/ingestic/ingestic.go
```diff
@@ -211,7 +211,34 @@ func (backend *ESClient) DeleteProjectTag(ctx context.Context,
...
+
+	params := map[string]interface{}{"project": projectTagToDelete}
+	docType := mappings.DocType
+	mapping := mappings.ComplianceRepDate
+	index := fmt.Sprintf("%s-%s", mapping.Index, "*")
+
+	startTaskResult, err := elastic.NewUpdateByQueryService(backend.client).
+		Index(index).
+		Type(docType).
+		Script(elastic.NewScript(script).Params(params)).
```
bcmdarroch: @lancewf under the UpdateReportProjectTag, the query includes `Refresh("true").` Do you know if we need to refresh the indices after deleting a project tag?

lancewf: I would say we would want to refresh to ensure the delete is live when we say this task is complete. Since deletes will not happen that often a refresh should not add too much load to Elasticsearch.

components/compliance-service/ingest/ingestic/ingestic.go
```diff
@@ -20,6 +20,18 @@ type ESClient struct {
 	initialized bool
 }
 
+var (
```
bcmdarroch: @lancewf do you know if there's any elasticsearch related code shared btwn ingest-service and compliance? this script seems like it could be a shared constant.

though will that be true if we add a fetch to only get resources with the project to delete 🤔

lancewf: I don't know anywhere that elasticsearch code is shared between the two services. The code should be the same even with filtering for the deleted project. 

My preference would be to have the duplicate code in both services because the only reason the code is the same now is because the project field is named the same and is in the same location on the ES document.

components/compliance-service/ingest/ingestic/ingestic.go
```diff
@@ -20,6 +20,138 @@ type ESClient struct {
...
+		if (ctx._source['projects'] != null) {
+			def foundIndex = ctx._source.projects.indexOf(params.project);
+
+			if (foundIndex != -1) {
+				ctx._source.projects.remove(foundIndex);
+			}
+		}
+	`
+
+	matchingProjectTagsScript = `ArrayList matchingProjects = new ArrayList();
```
bcmdarroch: not really related to the work of this PR but i noticed this script was the exact same in the helper functions that were using it.

components/compliance-service/ingest/ingestic/ingestic.go
```diff
@@ -192,9 +324,67 @@ func (backend *ESClient) setDailyLatestToFalse(ctx context.Context, nodeId strin
 }
 
 func (backend *ESClient) DeleteProjectTag(ctx context.Context,
```
msorens: Signature needs to change per #2389 -- see https://github.com/chef/automate/pull/2428/files#r354108267

components/compliance-service/ingest/ingestic/ingestic.go
```diff
@@ -192,9 +324,67 @@ func (backend *ESClient) setDailyLatestToFalse(ctx context.Context, nodeId strin
...
+	}
+
+	esSummaryJobID, err := backend.DeleteSummaryProjectTag(ctx, projectTagToDelete)
+	if err != nil {
+		return []string{}, errors.Wrap(err, "Failed to start Elasticsearch Summary project tags delete")
+	}
+
+	logrus.Debugf("Started Project delete with report job ID: %q, summary job ID %q", esReportJobID, esSummaryJobID)
+
+	return []string{esReportJobID, esSummaryJobID}, nil
```
msorens: As part of the signature change, instead of returning the two IDs here, then need to add a monitoring loop like that shown here: https://github.com/chef/automate/pull/2428/files#r354108737

Given it would be the same monitoring loop, sounds like a common helper is called for.

(I am happy to handle those changes in #2389 if this is merged as is, first.)

blakestier: I'm into the rule of 3s when DRYing things up . . . will there be 3? I've lost track

bcmdarroch:
after discussing the [tradeoffs](https://github.com/chef/automate/issues/2016#issuecomment-562385337) of this approach, we decided as a team to proceed in a [different direction](https://github.com/chef/automate/issues/2441)

Issue #2260, opened 2019-11-18
project selection may be lost due to polling


susanev:
## Overview
Currently, in the global projects filter and the project assignment dropdown, we poll for changes to projects. A user could get unlucky and have their in-progress work totally lost before they apply their changes if they start editing close to the end of the polling period. This creates a very frustrating experience for the user.

## Repro steps
- Open the global projects filter (and the project assignment dropdown), make some changes, then wait.. notice it refreshes and you lose ur unsaved work

## Expected behaviour
- When the user is editing the project filter checkboxes or the checkboxes in the project assignment dropdown their changes should not be lost when the polling period ends, ideally they would also not see a visible refresh

PR #2421, opened 2019-12-04
[AUTOMATE-2260] If the user has pending changes in the projects filter, do not use polled updates

tylercloke:

We have two minute polling to get updated projects populated in the projects filter dropdown. However, if that polling triggers while edits are pending in the dropdown, it would remove the pending changes. Only update the dropdown if there are not pending changes so that user's edits aren't blown away by polling updates. #2260

```
start_all_services
npm run serve:hab
```

Edit `POLLING_INTERVAL_IN_SECONDS` in `projects-filter.effects.ts` to be 10 seconds or something shorter so you don't have to wait two minutes.

Make a few projects

```
curl -H "api-token: $ADMINTOK" -XPOST -d '{"name":"project1", "id":"project1"}' https://a2-dev.test/apis/iam/v2beta/projects --insecure
curl -H "api-token: $ADMINTOK" -XPOST -d '{"name":"project2", "id":"project2"}' https://a2-dev.test/apis/iam/v2beta/projects --insecure
```

Log into the UI and drop down the projects filter, you should see unassigned, projects 1 and 2. Do not select anything.

```
curl -H "api-token: $ADMINTOK" -XPOST -d '{"name":"project3", "id":"project3"}' https://a2-dev.test/apis/iam/v2beta/projects --insecure
```

Wait 10ish seconds and project 3 should show up.

Select anything to see the `Apply Changes` button enabled. Now create another project:

```
curl -H "api-token: $ADMINTOK" -XPOST -d '{"name":"project4", "id":"project4"}' https://a2-dev.test/apis/iam/v2beta/projects --insecure
```

The project will not show up in the projects filter dropdown or blow away your changes until you close or apply your changes. Do either and re-open the projects filter to see the updated projects.

Also do the same while deleting projects.

```
curl -H "api-token: $ADMINTOK" -XDELETE https://a2-dev.test/apis/iam/v2beta/projects/project4 --insecure
```

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.ts
+++ b/components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.ts
@@ -78,13 +78,16 @@ export function projectsFilterReducer(
     }
 
     case ProjectsFilterActionTypes.SAVE_OPTIONS: {
+      // merge any added or removed options with the newly saved values
+      const mergedOptions = mergeOptions(state.options, action.payload);
+      const sortedOptions = sortOptions(mergedOptions);
       return pipe(
-        set('options', action.payload),
-        set('selectionLabel', selectionLabel(action.payload)),
-        set('selectionCount', selectionCount(action.payload)),
-        set('selectionCountVisible', selectionCountVisible(action.payload)),
-        set('selectionCountActive', selectionCountActive(action.payload)),
-        set('dropdownCaretVisible', dropdownCaretVisible(action.payload))
+        set('options', sortedOptions),
+        set('selectionLabel', selectionLabel(sortedOptions)),
+        set('selectionCount', selectionCount(sortedOptions)),
+        set('selectionCountVisible', selectionCountVisible(sortedOptions)),
+        set('selectionCountActive', selectionCountActive(sortedOptions)),
+        set('dropdownCaretVisible', dropdownCaretVisible(sortedOptions))
       )(state) as ProjectsFilterState;
     }
   }
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
@@ -34,7 +34,7 @@ export class ProjectsFilterDropdownComponent implements OnChanges {
   dropdownActive = false;
 
   ngOnChanges(changes) {
-    if (changes.options) {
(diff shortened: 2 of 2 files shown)


lancewf:
I don't know if I am testing it correctly, but I am seeing the new project pop up in the project filter. 
pending

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.ts
```diff
@@ -78,13 +78,16 @@ export function projectsFilterReducer(
...
-        set('selectionCount', selectionCount(action.payload)),
-        set('selectionCountVisible', selectionCountVisible(action.payload)),
-        set('selectionCountActive', selectionCountActive(action.payload)),
-        set('dropdownCaretVisible', dropdownCaretVisible(action.payload))
+        set('options', sortedOptions),
+        set('selectionLabel', selectionLabel(sortedOptions)),
+        set('selectionCount', selectionCount(sortedOptions)),
+        set('selectionCountVisible', selectionCountVisible(sortedOptions)),
+        set('selectionCountActive', selectionCountActive(sortedOptions)),
+        set('dropdownCaretVisible', dropdownCaretVisible(sortedOptions))
```
msorens: Since this block is now virtually identical to that for `LOAD_OPTIONS_SUCCESS`, seems like it would ease the future maintenance burden by extracting a common `setState` function? There are just two differences to account for -- `filterVisible` and `optionsLoadingStatus` -- and I believe it would essentially be a no-op if those were added here for `SAVE_OPTIONS`, too.
Then, this block would just become
```
case ProjectsFilterActionTypes.SAVE_OPTIONS: { 
  return setState(state.options, action.payload);
}
```

msorens:
@lancewf If I am interpreting your video correctly, you are just opening the dropdown but not making any changes, so that is OK to propagate changes. The intent here is that between the time you make changes and select "Apply Changes" then updates should be suppressed.

components/automate-ui/src/app/services/projects-filter/projects-filter.reducer.ts
tylercloke: I started to take a shot at this but I think it actually makes things more complex because LOAD_OPTIONS_SUCCESS has two additional fields to set and one of them wants the sorted options:

```
        set('optionsLoadingStatus', EntityStatus.loadingSuccess),
        set('filterVisible', filterVisible(sortedOptions))
```

PR #2423, opened 2019-12-04
[AUTOMATE-1654] Reset the project update state on update and properly render state

tylercloke:

Reset the project update status state in the store when we kick off the project update and also fix some small issues with the update logic.

Boot up the UI and run some project updates. You should no longer briefly see 100% get returned. This was because it was getting back 100% complete due to race condition between project update kickoff API request and status endpoint.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/pages/project/list/project-list.component.ts
+++ b/components/automate-ui/src/app/pages/project/list/project-list.component.ts
@@ -92,12 +92,13 @@ export class ProjectListComponent implements OnInit, OnDestroy {
       .subscribe(({ state, failed, cancelled, percentageComplete }: ApplyRulesStatus) => {
         if (this.applyRulesInProgress && state === ApplyRulesStatusState.NotRunning) {
           this.cancelRulesInProgress = false;
+          this.percentageComplete = 0;
           this.closeConfirmApplyStopModal();
         }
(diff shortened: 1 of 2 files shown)


components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -92,12 +92,13 @@ export class ProjectListComponent implements OnInit, OnDestroy {
...
         if (this.applyRulesInProgress && state === ApplyRulesStatusState.NotRunning) {
           this.cancelRulesInProgress = false;
+          this.percentageComplete = 0;
           this.closeConfirmApplyStopModal();
         }
         this.applyRulesInProgress = state === ApplyRulesStatusState.Running;
         this.updateProjectsFailed = failed;
         this.updateProjectsCancelled = cancelled;
-        if (!this.cancelRulesInProgress) {
+        if (!this.cancelRulesInProgress && state !== ApplyRulesStatusState.NotRunning) {
```
msorens: Is there a reason not to use this simpler form?
```suggestion
        if (!this.cancelRulesInProgress && state === ApplyRulesStatusState.Running) {
```

tylercloke:
@bcmdarroch yeah saw that as well, I believe that's API side after looking at the network requests though. thanks for making a card!

PR #2424, opened 2019-12-04
Adding project delete for the nodemanager

lancewf:

Creating a function to delete an IAMv2 project from nodemanager nodes.

#2383

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/nodemanager-service/pgdb/project_delete.go
+++ b/components/nodemanager-service/pgdb/project_delete.go
@@ -1,0 +1,89 @@
+package pgdb
+
+import (
+	"context"
+
+	project_update_lib "github.com/chef/automate/lib/authz"
+	uuid "github.com/chef/automate/lib/uuid4"
+	log "github.com/sirupsen/logrus"
+)
+
+const deleteProjectNodeAssociation = `
+DELETE FROM nodes_projects np
+USING projects p
+WHERE p.id = np.project_id AND p.project_id = $1;
+`
+
+type ProjectDelete struct {
+	jobStatus      project_update_lib.JobStatus
+	jobStatusError error
+	running        bool
+	ID             string
+}
+
+func (db *DB) DeleteProjectTag(ctx context.Context,
+	projectTagToBeDelete string) ([]string, error) {
+	jobUUID := uuid.Must(uuid.NewV4()).String()
+	log.Debugf("Running node manager DeleteProjectTag %v ID: %s", projectTagToBeDelete, jobUUID)
+
+	db.ProjectDelete.start(jobUUID)
+
+	go db.deleteProject(projectTagToBeDelete)
+
+	return []string{jobUUID}, nil
+}
+
+func (db *DB) deleteProject(projectTagToBeDelete string) {
+	_, err := db.Exec(deleteProjectNodeAssociation, projectTagToBeDelete)
+	if err != nil {
+		db.ProjectDelete.updateFail(err)
+	}
+
+	db.ProjectDelete.updateComplete()
+}
+
+func (projectDelete *ProjectDelete) start(jobUUID string) {
+	projectDelete.jobStatusError = nil
+	projectDelete.running = true
+	projectDelete.jobStatus = project_update_lib.JobStatus{
+		Completed:             false,
+		PercentageComplete:    0,
+		EstimatedEndTimeInSec: 0,
+	}
+	projectDelete.ID = jobUUID
+}
+
+func (projectDelete *ProjectDelete) getJobStatus(jobUUID string) (project_update_lib.JobStatus, error) {
+	if projectDelete.ID == jobUUID {
+		if projectDelete.jobStatusError != nil {
+			return project_update_lib.JobStatus{}, projectDelete.jobStatusError
+		}
+		return projectDelete.jobStatus, nil
+	}
+
+	// If the jobID does not match the currently running job then it must have completed before.
+	return project_update_lib.JobStatus{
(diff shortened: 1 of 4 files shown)


components/nodemanager-service/pgdb/project_delete.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func (db *DB) DeleteProjectTag(ctx context.Context,
+	projectTagToBeDelete string) ([]string, error) {
+	jobUUID := uuid.Must(uuid.NewV4()).String()
+	log.Debugf("Running node manager DeleteProjectTag %v ID: %s", projectTagToBeDelete, jobUUID)
+
+	db.ProjectDelete.start(jobUUID)
+
+	go db.deleteProject(projectTagToBeDelete)
```
lancewf: We do not really need to run this delete async, but we will to keep the same pattern as the other services.

components/nodemanager-service/pgdb/project_update.go
```diff
@@ -123,7 +123,20 @@ func (db *DB) UpdateProjectTags(ctx context.Context,
 
 // JobStatus - get the job status of the project update
 func (db *DB) JobStatus(ctx context.Context, jobID string) (project_update_lib.JobStatus, error) {
```
lancewf: This is where we can check the status of both a project update and project delete status.

components/nodemanager-service/pgdb/project_delete.go
```diff
@@ -0,0 +1,89 @@
...
+
+type ProjectDelete struct {
+	jobStatus      project_update_lib.JobStatus
+	jobStatusError error
+	running        bool
+	ID             string
+}
+
+func (db *DB) DeleteProjectTag(ctx context.Context,
+	projectTagToBeDelete string) ([]string, error) {
```
blakestier: s/projectTagToBeDelete/projectTagToBeDeleted

components/nodemanager-service/pgdb/project_delete.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"context"
+
+	project_update_lib "github.com/chef/automate/lib/authz"
+	uuid "github.com/chef/automate/lib/uuid4"
+	log "github.com/sirupsen/logrus"
+)
+
+const deleteProjectNodeAssociation = `
+DELETE FROM nodes_projects np
```
blakestier: This is not immediately related to this PR . . . but shouldn't the key in this relational table be the project's `id`, not the `project_id`?

components/nodemanager-service/pgdb/project_delete.go
lancewf: Not again!! 😠

components/nodemanager-service/pgdb/project_delete.go
```diff
@@ -0,0 +1,89 @@
...
+		Completed:             true,
+		PercentageComplete:    1,
+		EstimatedEndTimeInSec: 0,
+	}, nil
+}
+
+func (projectDelete *ProjectDelete) updateComplete() {
+	projectDelete.jobStatus = project_update_lib.JobStatus{
+		Completed:             true,
+		PercentageComplete:    1.0,
```
blakestier: Why is this `PercentageComplete` a float while the other is a bool?

lancewf: `PercentageComplete` is the range between 1.0 to 0.0 so it can not be an integer unless we change the range to 100 to 0. The `EstimatedEndTimeInSec` is the number of seconds since 1970.

blakestier: It appears to be an integer here: https://github.com/chef/automate/pull/2424/files#diff-87e11a1c44a032343e1ce6928c55b7c7R67 and here https://github.com/chef/automate/pull/2424/files#diff-35fc59428b9b5a554bcf8b413be709ddR131

Maybe converted?

lancewf: The underlying data is stored as a float defined here https://github.com/chef/automate/blob/b5f2710d53bbad79a7a4602bd0799f19d40407fb/lib/authz/project_rules.go#L9. 

So, it is the same if we have a `1.0` or a `1` it is stored the same.

components/nodemanager-service/pgdb/project_delete.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+func (db *DB) DeleteProjectTag(ctx context.Context,
+	projectTagToBeDeleted string) ([]string, error) {
+	jobUUID := uuid.Must(uuid.NewV4()).String()
+	log.Debugf("Running node manager DeleteProjectTag %v ID: %s", projectTagToBeDeleted, jobUUID)
+
+	db.ProjectDelete.start(jobUUID)
+
+	go db.deleteProject(projectTagToBeDeleted)
```
tylercloke: We are just using goroutines here? What if the service crashes?

lancewf: If the service crashes then when we ping the service to check the `getJobStatus` we will see that the job failed. 

Looking at the `getJobStatus` https://github.com/chef/automate/pull/2424/files#diff-87e11a1c44a032343e1ce6928c55b7c7R65 I do need to change it some.

lancewf: On second thought I see another problem. This is only going to allow one project delete at a time. 🤔

Issue #2252, opened 2019-11-18
escalation: scan jobs with thousands of nodes working but reporting blank node errors


vjeffrey:
## Describe the bug
Sorry if the title is confusing.
tl;dr customer is running scan jobs on about 2000 nodes.
reports are being generated and ingested for most of them, as expected.
but the scan job lists all the nodes as having failed, with no error message.

_original issues: https://chefio.atlassian.net/browse/TRI-901 https://getchef.zendesk.com/agent/tickets/23215_

## To Reproduce
- load lots of nodes, most reachable, a few unreachable
- create a scan job to scan all the nodes
- see all nodes reported as failed in the scan jobs view despite having ingested reports from the nodes

_please note: i have not been able to reproduce this yet_

## Expected behavior
Only the nodes that we could not reach marked as failed, with an error message.

## Screenshots
<img width="949" alt="Screen Shot 2019-11-18 at 14 37 16" src="https://user-images.githubusercontent.com/10341541/69096064-f25a9a80-0a10-11ea-9441-541e4fb76798.png">

## Versions (please complete the following information):
- Automate Build Number: 20190904132002

vjeffrey:
medium t shirt size, but unknown b/c no repro yet

vjeffrey:
closing as the issue is not reproducible and we believe the merged pr fixes will help with the lack of errors the user was seeing

PR #2426, opened 2019-12-04
inspec agent cleanup

vjeffrey:
a bit of cleanup/refactoring
was spending some time in this part of the code base for some other changes and wanted to spend some time making it a little easier to read.
hopefully this work is a step in that direction.
i'm gonna add some testing around my new function too.

n/a

### :+1: Definition of Done
it's a little easier to read through the resolver code

rebuild compliance, make sure scan jobs run as expected

- [ ] Tests added/updated?
- [n/a] Docs added/updated?

### DIFF
--- a/components/compliance-service/inspec-agent/resolver/resolver.go
+++ b/components/compliance-service/inspec-agent/resolver/resolver.go
@@ -87,7 +87,10 @@ func (r *Resolver) ResolveJob(ctx context.Context, job *jobs.Job) ([]*types.Insp
 		logrus.Warnf("Nothing to do for job %s, empty list of nodes/selectors", job.Id)
 		return []*types.InspecJob{}, fmt.Errorf("cannot process this job, nodes and selectors are both nil")
 	}
-	r.scannerServer.UpdateJobNodeCount(job.Id, len(nodeJobs))
+	err = r.scannerServer.UpdateJobNodeCount(job.Id, len(nodeJobs))
+	if err != nil {
+		logrus.Errorf("error trying to job node count for job %s (%s): %s", job.Name, job.Id, err.Error())
+	}
 	return nodeJobs, nil
 }
 
--- a/components/compliance-service/inspec-agent/resolver/resolver.go
+++ b/components/compliance-service/inspec-agent/resolver/resolver.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"strings"
-	"time"
 
 	"github.com/chef/automate/api/external/secrets"
 	"github.com/chef/automate/components/compliance-service/api/common"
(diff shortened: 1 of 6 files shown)


components/compliance-service/scanner/db.go
```diff
@@ -104,45 +105,44 @@ func isStatusUpdateValid(oldStatus, newStatus string) bool {
...
 
-func (s *Scanner) UpdateJobStatus(job_id string, newStatus string, startTime *time.Time, endTime *time.Time) {
+func (s *Scanner) UpdateJobStatus(job_id string, newStatus string, startTime *time.Time, endTime *time.Time) error {
 	// get current status of job
 	var status string
 	err := s.DB.QueryRow(sqlGetNodeStatus, job_id).Scan(&status)
 	if err != nil {
-		logrus.Errorf("UpdateJobStatus unable to get current job status: %s", err.Error())
-		return
+		return errors.Wrapf(err, "UpdateJobStatus unable to get current job status")
```
vjeffrey: i still need to follow up with a commit that ensures all the callers of these functions are at the very least logging the errors found

components/compliance-service/inspec-agent/resolver/resolver.go
```diff
@@ -91,6 +90,60 @@ func (r *Resolver) ResolveJob(ctx context.Context, job *jobs.Job) ([]*types.Insp
...
+		},
+		SSM:               node.SSM,
+		Timeout:           job.Timeout,
+		Retries:           job.Retries,
+		TargetConfig:      fullTc,
+		Profiles:          job.Profiles,
+		SourceID:          node.CloudID,
+		SourceAccountID:   node.CloudAccountID,
+		ManagerID:         node.ManagerID,
+		MachineIdentifier: node.MachineIdentifier,
```
vjeffrey: would be good to add a comment for what some of the non-obvious fields are/why they are used

components/compliance-service/scanner/db.go
vjeffrey: that's done now, many thanks to the linter!

components/compliance-service/scanner/db.go
```diff
@@ -217,28 +216,31 @@ func (s *Scanner) UpdateResult(ctx context.Context, job *types.InspecJob, output
 	if job.NodeStatus == types.StatusFailed || job.NodeStatus == types.StatusAborted {
 		if inspecErr != nil {
 			result.Result = inspecErr.Message
+		} else {
+			result.Result = fmt.Sprintf("unable to complete scan on node. please check the logs for more information. node id ref: %s", job.NodeID)
```
vjeffrey: i'm wondering if maybe the lack of else logic here contributed to https://github.com/chef/automate/issues/2252
i'm hoping this else logic + more detailed msgs in the logs will help

components/compliance-service/inspec-agent/resolver/resolver.go
vjeffrey: done on the struct definition

components/compliance-service/inspec/cli.go
```diff
@@ -250,6 +250,11 @@ func getInspecError(serr string, err error, target *TargetConfig, timeout time.D
 		return NewInspecError(UNREACHABLE_HOST,
 			"Failed to connect to "+target.Hostname+", host is unreachable.")
 	}
+	connTimeoutErr := "Net::SSH::ConnectionTimeout"
+	if strings.Index(serr, connTimeoutErr) >= 0 {
```
vjeffrey: this should help with the reports that there are a lot of "unknown error" error msgs

components/compliance-service/inspec-agent/runner/runner.go
```diff
@@ -293,6 +293,27 @@ type ResolveTask struct {
...
+		},
+		StartTime:  &now,
+		EndTime:    &now,
+		NodeStatus: types.StatusAborted,
+	}
+	err := t.scanner.UpdateJobStatus(nodeJob.JobID, "failed", nodeJob.StartTime, nodeJob.EndTime)
+	if err != nil {
+		logrus.Errorf("error updating job status for job %s (%s): %s", job.Name, job.Id, err.Error())
+	}
+	err = t.scanner.UpdateResult(context.TODO(), nodeJob, nil, &inspec.Error{Message: "no nodes found"}, "")
```
stevendanna: The method this is called from is `Run(ctx context.Context, task cereal.Task)` so perhaps we can thread that context through rather than using context.TODO

components/compliance-service/inspec-agent/scheduler/scheduler.go
```diff
@@ -93,15 +96,19 @@ func (a *Scheduler) PollForJobs(ctx context.Context) {
...
+	}
 	if len(dueJobs) > 0 {
 		logrus.Debugf("processDueJobs, %d recurring jobs are due for running...", len(dueJobs))
 	}
 	for _, job := range dueJobs {
 		err := a.pushWorkflow(job, false)
 		if err != nil {
 			if err == cereal.ErrWorkflowInstanceExists {
-				logrus.Infof("Job %q is still/already running", job.Id)
+				logrus.Debugf("Job %q is still/already running", job.Id)
```
stevendanna: Are we seeing this a lot? Seems like something we would want to keep at Info as it might mean compliance can't keep up with the schedule.  Ideally, we would surface this to the user in the UI

components/compliance-service/scanner/db.go
stevendanna: 👍 We weren't brave enough to fix this error handling when moving the code around, thanks for taking care of it.

components/compliance-service/inspec-agent/scheduler/scheduler.go
vjeffrey: ah ya that’s a good point. i haven’t seen it often, i was just over-aggressively changing info msgs to debug :) i’ll switch it back to info!

vjeffrey: ah ya that’s a good point. i haven’t seen it often, i was just over-aggressively changing info msgs to debug :) i’ll switch it back to info!

components/compliance-service/inspec-agent/resolver/resolver.go
```diff
@@ -88,9 +87,69 @@ func (r *Resolver) ResolveJob(ctx context.Context, job *jobs.Job) ([]*types.Insp
...
+	MachineIdentifier string // special identifier used when querying azure api, the resource group name
+	Tags              []*common.Kv
+}
+
+func assembleJob(job *jobs.Job, node nodeInfo, secrets []*inspec.Secrets, tc inspec.TargetBaseConfig) *types.InspecJob {
+	if len(node.UUID) == 0 {
+		id, err := uuid.NewV4()
+		if err != nil {
+			logrus.Warnf("unable to generate uuid for node")
+		}
```
phiggins: Is it ok to continue processing if this fails? Can this error get bubbled up?

vjeffrey: oh good point. if we continue on from here we’ll eventually get a scan job report that gets rejected. i’ll push an update for that!

PR #2427, opened 2019-12-05
fix nodemanager tests

vjeffrey:
noticed some tests in CI weren't failing where they should...
Screen Shot 2019-12-04 at 20 55 14

i added a failure commit so we expect to see that same nodemanager-service test suite fail 🤞

### DIFF
--- a/.studio/nodemanager-service
+++ b/.studio/nodemanager-service
@@ -6,6 +6,6 @@ document "nodemanager_scans_integration" <<DOC
   Runs the nodemanager scan jobs for aws and azure
 DOC
 function nodemanager_scans_integration() {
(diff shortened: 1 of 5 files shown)


phiggins:
The bash function that runs these tests runs two `go_test`s in a row, could the first one fail and the second pass making the whole thing a success?

https://github.com/chef/automate/blob/master/.studio/nodemanager-service#L8-L11

I wonder if we need to `&&` those together or just run all the tests in one command...

phiggins:
We probably don't want `failfast` everywhere since that is more useful in a local fast iteration scenario.

phiggins:
I opened a PR to try to fix this in #2435 but it's failing some tests.

PR #2435, opened 2019-12-05
Debug nodemanager CI non-failures

phiggins:
we discovered nodemanager tests were failing in CI but not failing the job.
pete noticed we were running two separate go_test commands, so if the second one passed it ignored the failure from the first, so he modified the command to run em both.
we saw weird errors b/c this changed the order of the tests.
modified filenames to get around that.


vjeffrey:
i'm taking a look at the test failures now

vjeffrey:
these tests are being really weird.
i'm gonna work on them locally this afternoon. i have no idea what's going on

phiggins:
> these tests are being really weird.

Maybe these tests weren't made to be run together? That is, I wonder if the two sets of tests pollute the environment when before the global state was being reset in between runs, so maybe running them like `go_test ... && go_test ...` instead of running them all in one `go_test` run.

vjeffrey:
ah ya good point. the scan jobs are finishing and inserting nodes so then we have a no nodes added thing, etc. crud. alright we’ll i’ll see if i can get it working running all of them together, or some other way to order them to ensure it works...

vjeffrey:
now the failure that's there is the same failure we're seeing on master/in CI

components/nodemanager-service/tests/project_filtering_test.go
```diff
@@ -375,11 +375,20 @@ func TestListProjectFilteringIngestedNodes(t *testing.T) {
...
+					if node.Manager == "" {
+						actualNodeIDs = append(actualNodeIDs, node.Id)
+					} else {
+						actualManuallNodeIDs = append(actualManuallNodeIDs, node.Id)
+					}
+				}
+
+				for _, nodeID := range manualNodeIds {
+					assert.Contains(t, actualManuallNodeIDs, nodeID)
 				}
```
phiggins: Should `manualNodeIds` and `actualManualNodeIDs` contain the same elements? Should this be a `ElementsMatch` assertion like below?

phiggins:
I can't approve this since this is technically my PR

components/nodemanager-service/tests/project_filtering_test.go
vjeffrey: there could be more nodes in the actualManualNodeIDs than just the ones added in this test -- from other tests that ran first/scans finishing

PR #2442, opened 2019-12-06
Let timeout log when timing out integration test

jaym:
It will print something like:
```
timeout: sending signal TERM to command ‘bash’
```


jaym:
guess our timeout doesn't support `-v`

Issue #2447, opened 2019-12-06
Prevent project deletion API


lancewf:
## User Story

Only allowing a user to delete a project when the project has no ingest rules and no pending edits. 

## Definition of Done

The API request for the endpoint "/iam/v2beta/projects/{id}" validates if a user is allowed to delete a project with the two requirements of
* No pending edits on that project
* The project has no ingest rules

## Demo Script / Repro Steps
1. Create a project with ingest rules.
1. Apply the rules. 
1. Try to delete the project with this endpoint "/iam/v2beta/projects/{id}". Ensure an error is returned and the project is not deleted
1. Remove the ingest rules from the project and apply rules. 
1. Try again to delete the first project and ensure no error is returned and the project is deleted.

susanev:
Note from conov with lance and michael
- if there are pending edits on any project, then you cannot delete **any** project. this is because we do not know if an individual project is up-to-date.
- project delete needs to be atomic, lance and michael are going to pair on this
    - they cannot add rules to a project that is being deleted

msorens:
Single user:
If there are rules (staged or applied) for a given project, that project cannot be deleted. The API should return an error to the effect of "cannot delete a project containing any applied or staged ingest rules". Once the user has removed all rules, they will be allowed to do the delete.

Multiple users:
Say user A starts a delete of project 1. Meanwhile, user B attempts to add a rule to project 1 simultaneously.

 - User A's `delete project` occurs first => user B will receive an error (either from authz or from the DB itself) saying that the project does not exist.

 - User B's `add rule` occurs first => the delete will then come along right after and delete the project. I believe this cascade deletes any rules, too. So User B will think their `add rule` succeeded, but if they refresh the browser the new rule will not be there. While this is not ideal UX it is a known limitation of our system for multiple users attempting to manipulate the same IAM resource (e.g. one editing a team name while another deleting the team).

PR #2450, opened 2019-12-06
[automate-2447] Preventing project deletion

lancewf:

From the API, prevent the deletion of a project if it has rules associated with it. There are two cases to prevent project deletion. One, if the project has any rules associated with it. Second, if the project is in the edits pending state. 

https://github.com/chef/automate/issues/2447

1. Go to the project page https://a2-dev.test/settings/projects
1. Create a project with a rule.
1. Ensure the project can not be deleted.
1. Run apply-rules. 
1. Again ensure the project can not be deleted.
1. Delete the rule from the project. 
1. Then one more time, ensure the project can not be deleted.
1. Run apply-rules again.
1. Then ensure the project can be deleted. 

- [x] Tests added/updated?
- [ ] Docs added/updated?


components/authz-service/server/v2/project.go
```diff
@@ -331,7 +331,16 @@ func (s *ProjectState) ListProjectsForIntrospection(
 func (s *ProjectState) DeleteProject(ctx context.Context,
 	req *api.DeleteProjectReq) (*api.DeleteProjectResp, error) {
 
-	err := s.ProjectPurger.Start(req.Id)
+	resp, err := s.ListRulesForProject(ctx, &api.ListRulesForProjectReq{Id: req.Id})
```
lancewf: This is the main change. Checking if there are any applied or staged rules for the project to be deleted.

components/authz-service/server/v2/project_test.go
```diff
@@ -229,7 +229,7 @@ func TestGetProject(t *testing.T) {
 
 func TestDeleteProject(t *testing.T) {
 	ctx := context.Background()
-	cl, store, cleanup := setupProjects(t)
+	cl, projects, rules, _, cleanup := setupProjectsAndRules(t)
```
lancewf: This allows the test to add a rule to the project.

components/authz-service/server/v2/project.go
```diff
@@ -331,7 +331,12 @@ func (s *ProjectState) ListProjectsForIntrospection(
 func (s *ProjectState) DeleteProject(ctx context.Context,
 	req *api.DeleteProjectReq) (*api.DeleteProjectResp, error) {
 
-	err := s.ProjectPurger.Start(req.Id)
+	err := s.validateProjectDelete(ctx, req.Id)
```
lancewf: This is the main change. 
We are validating that the project can be deleted.

components/authz-service/server/v2/project.go
```diff
@@ -568,6 +541,65 @@ func (s *ProjectState) DeleteRule(ctx context.Context, req *api.DeleteRuleReq) (
...
 
+const ruleSetEditsPendingStatus = "edits-pending"
+
+func (s *ProjectState) validateProjectDelete(ctx context.Context, projectID string) error {
+	projectResponse, err := s.GetProject(ctx, &api.GetProjectReq{Id: projectID})
+	if err != nil {
+		return err
+	}
+
+	if projectResponse.Project.Status == ruleSetEditsPendingStatus {
```
lancewf: The project can not be deleted if it has 'edits-pending'.

components/authz-service/server/v2/project.go
```diff
@@ -568,6 +541,65 @@ func (s *ProjectState) DeleteRule(ctx context.Context, req *api.DeleteRuleReq) (
...
+		return status.Errorf(codes.FailedPrecondition,
+			"Project %q can not be deleted because it has edits pending", projectResponse.Project.Name)
+	}
+
+	rulesResponse, err := s.ListRulesForProject(ctx, &api.ListRulesForProjectReq{Id: projectID})
+	if err != nil {
+		return err
+	}
+
+	if len(rulesResponse.Rules) > 0 {
```
lancewf: The project can not be deleted if it has any rules.

e2e/cypress/support/commands.ts
```diff
@@ -163,10 +163,81 @@ Cypress.Commands.add('cleanupV2IAMObjectsByIDPrefixes',
   (idPrefix: string, iamObjects: string[]) => {
 
   iamObjects.forEach((iamObject) => {
-    cleanupV2IAMObjectByIDPrefix(idPrefix, iamObject);
+    if (iamObject === 'projects') {
+      cleanupProjectsByIDPrefixes(idPrefix);
```
lancewf: We now need a special case for deleting projects, because of the new rules added in this PR.

components/authz-service/server/v2/project.go
```diff
@@ -568,6 +541,65 @@ func (s *ProjectState) DeleteRule(ctx context.Context, req *api.DeleteRuleReq) (
 	}
 }
 
+const ruleSetEditsPendingStatus = "edits-pending"
```
lancewf: Is there a better place to put this?

bcmdarroch: This might already be an enum somewhere..

bcmdarroch: I think you'll want to use this enum: https://github.com/chef/automate/blob/master/components/authz-service/storage/v2/project_rules_status.go#L11

lancewf: I could not find it in the authz-service only in the automate-gateway.

lancewf: How do I go from the string version to the enum?

components/authz-service/server/v2/project.go
bcmdarroch: like @blakestier mentioned at standup, it should be ok to delete a project if there's only staged rules, because they're not in effect and they get blown away on project delete. however, since we're worried about race conditions with the project update, it's probably best to be more restrictive, as you've done here.

components/authz-service/server/v2/project_test.go
```diff
@@ -247,36 +247,78 @@ func TestDeleteProject(t *testing.T) {
...
 		{"fails with NotFound when deleting from empty store", func(t *testing.T) {
-			require.Zero(t, store.ItemCount())
+			require.Zero(t, projects.ItemCount())
 			resp, err := cl.DeleteProject(ctx, &api.DeleteProjectReq{Id: "test"})
 			grpctest.AssertCode(t, codes.NotFound, err)
 			assert.Nil(t, resp)
 		}},
 		{"deletes custom project when one project is in database", func(t *testing.T) {
-			require.Zero(t, store.ItemCount())
+			require.Zero(t, projects.ItemCount())
```
bcmdarroch: let's update the test name to `deletes custom project with no roles when one project is in database`

components/authz-service/server/v2/project_test.go
```diff
@@ -247,36 +247,78 @@ func TestDeleteProject(t *testing.T) {
...
+			addProjectToStore(t, projects, id, "my foo", storage.Custom)
 
 			_, err := cl.DeleteProject(ctx, &api.DeleteProjectReq{Id: id})
 			require.NoError(t, err)
-			assert.Zero(t, store.ItemCount())
+			assert.Zero(t, projects.ItemCount())
 		}},
 		{"deletes custom project when several are in database", func(t *testing.T) {
-			require.Zero(t, store.ItemCount())
+			require.Zero(t, projects.ItemCount())
```
bcmdarroch: change this name too

components/authz-service/server/v2/project.go
lancewf: I think I figured out how to use that.

e2e/cypress/support/commands.ts
```diff
@@ -163,10 +163,81 @@ Cypress.Commands.add('cleanupV2IAMObjectsByIDPrefixes',
...
+      if (project.id.startsWith(idPrefix)) {
+        projectIds.push(project.id);
+      }
+    }
+
+    deleteProjects(projectIds, 0, false);
+  });
+}
+
+function deleteProjects(projectIdsToDelete: string[], index: number,
```
bcmdarroch: nice work on this helper :)

components/authz-service/server/v2/project_test.go
lancewf: Updated

components/authz-service/server/v2/project_test.go
lancewf: Updated

e2e/cypress/support/commands.ts
```diff
@@ -163,10 +163,81 @@ Cypress.Commands.add('cleanupV2IAMObjectsByIDPrefixes',
...
-    cleanupV2IAMObjectByIDPrefix(idPrefix, iamObject);
+    if (iamObject === 'projects') {
+      cleanupProjectsByIDPrefixes(idPrefix);
+    } else {
+      cleanupV2IAMObjectByIDPrefix(idPrefix, iamObject);
+    }
   });
 });
 
+function cleanupProjectsByIDPrefixes(idPrefix: string): void {
```
msorens: This was a challenging bit of code to grok. Consider this variation to make it a bit easier, using smaller functions and eliminating the need for recursion:
```
function cleanupProjectsByIDPrefixes(idPrefix: string): void {
  cy.request({
    headers: { 'api-token': Cypress.env('ADMIN_TOKEN') },
    method: 'GET',
    url: '/apis/iam/v2beta/projects'
  }).then((resp) => {
    const projectIds: string[] = resp.body.projects
      .filter(project => project.id.startsWith(idPrefix))
      .map(project => project.id);
    const rulesWereDeleted =
      projectIds.reduce((result, id) => deleteProjectRules(id) || result, false);
    if (rulesWereDeleted) {
      applyRules();
    }
    projectIds.forEach(project => deleteProject(project));
  });
}
```

with these very simple support functions:
```
function deleteProjectRules(id: string): boolean {
  . . .
  return rulesWereDeleted;
}

function deleteProject(id: string): void {
  . . .
}

function applyRules(): void {
  . . .
}
```

components/authz-service/server/v2/project_test.go
```diff
@@ -247,36 +247,78 @@ func TestDeleteProject(t *testing.T) {
...
+					Value:     []string{"opscode"},
+				},
+			}
+			projectID := fmt.Sprintf("test-project-%d", time.Now().UnixNano())
+			ruleID := "foo-applied-rule"
+			addProjectToStore(t, projects, projectID, "my foo", storage.Custom)
+			addRuleToStore(t, rules, ruleID, "my applied foo rule", applied, storage.Node, projectID, storageConditions)
+
+			_, err := cl.DeleteProject(ctx, &api.DeleteProjectReq{Id: projectID})
+			grpctest.AssertCode(t, codes.FailedPrecondition, err)
```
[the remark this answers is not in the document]
lancewf: I don't normally add tests for error strings, because it makes the code harder to change.

e2e/cypress/support/commands.ts
lancewf: I went down a similar path when I first wrote this, but I could not figure out how to write the `deleteProjectRules` function. The problem here is it does an async request but needs to return a value based on the response.

lancewf: I did clean up the function some but I could not figure out how to write the `deleteProjectRules` function.

PR #2453, opened 2019-12-08
do not allow users to update name or tags of ingested nodes

vjeffrey:

this change prevents users from updating the name or tags on ingested nodes (nodes with no manager)

users may update other fields, such as the credential id for the node or backend, port, host information.
the name and tags are collected at ingestion time from the chef-client and inspec reports.

also made the inability to update aws-ec2 or azure-vm node names, tags, and host more obvious.

n/a

### :+1: Definition of Done
user not allowed to update name or tags on ingested node

rebuild nodemanager
ingest some nodes
try to update name of ingested node

- [x] Tests added/updated?
- [n/a] Docs added/updated?

### DIFF
--- a/components/nodemanager-service/api/nodes/server/server.go
+++ b/components/nodemanager-service/api/nodes/server/server.go
@@ -308,30 +308,45 @@ func (srv *Server) BulkDelete(ctx context.Context, in *nodes.Query) (*nodes.Bulk
 }
 
 func (srv *Server) validateNodeUpdate(ctx context.Context, in *nodes.Node) (*nodes.Node, error) {
-	fullNode := in
 	node, err := GetNode(ctx, &nodes.Id{Id: in.Id}, srv.db, srv.secretsClient)
 	if err != nil {
-		return fullNode, errors.Wrapf(err, "unable to retrieve node from db %s", in.Id)
+		return in, errors.Wrapf(err, "unable to retrieve node from db %s", in.Id)
 	}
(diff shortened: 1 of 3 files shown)


components/nodemanager-service/api/nodes/server/server.go
```diff
@@ -308,30 +308,45 @@ func (srv *Server) BulkDelete(ctx context.Context, in *nodes.Query) (*nodes.Bulk
...
+			return in, &errorutils.InvalidError{Msg: fmt.Sprintf("invalid option. unable to update name of %s node", in.Manager)}
 		} else if in.GetTargetConfig().GetHost() != "" && in.GetTargetConfig().GetHost() != node.GetTargetConfig().GetHost() {
-			logrus.Warnf("invalid option. unable to update host of %s node", in.Manager)
+			return in, &errorutils.InvalidError{Msg: fmt.Sprintf("invalid option. unable to update host of %s node", in.Manager)}
+		} else if !equal(in.Tags, node.Tags) {
+			return in, &errorutils.InvalidError{Msg: fmt.Sprintf("invalid option. unable to update tags of %s node", in.Manager)}
+		}
+	}
+	return in, nil
+}
```
phiggins: This function doesn't appear to set anything on the node anymore, so it could just return an error or nil instead of returning the node.

vjeffrey: ah good point, ya!

PR #2457, opened 2019-12-08
moving authentication docs to their own section

susanev:

The config menu item is getting kinda big so I was thinking it might be helpful to users to have an authentication specific section.

`make serve` from `components/automate-chef-io`

- [ ] Tests added/updated?
- [x] Docs added/updated?


susanev:
chatted with the team, and for now we are going to leave this in config
it may be good to consider the number of pages in that section in the future
i think there is some cognitive load to finding the authentication pages
tertiary menus would be a rad solution here

PR #2460, opened 2019-12-09
[automate-2448] UI notification for project deletion

msorens:

---- First section obsolete; see revised strategy below ----

The user can at any time select `Delete Project` from the UI. However, this delete operation is only allowed if the project has no associated rules. If it does, the UI reports that failed precondition to the user within the delete modal dialog, thus keeping that dialog open. Note, however, that other types of errors are possible, too. For these other errors, the dialog is closed and the standard notification system is used to report the error.

Here is the set of requirements implemented as unit tests (see project-list.component.spec.ts):

The delete modal...
```
1. has empty error message upon opening
2. opens upon selecting delete from control menu
3. sets error message upon failed precondition when modal is open
4. keeps modal open upon failed precondition
5. does not set error message upon failed precondition if modal is closed
6. resets error message after failure when re-opened later
7. does not set error upon failure other than precondition failed
8. closes upon failure other than precondition failed
9. does not set error message upon success
10. closes upon success
```

Note that the distinction discussed above about failed preconditions vs. other errors are covered in items 3, 4, 7, and 8. (The tests do not explicitly include checking the standard notification banner content, though.)

Using the control menu for a given project, select the `Delete Project` choice and you get this modal dialog:
image

There are two categories of errors that may come back:
Those failing precondition (which includes the error of a project still having associated rules) and those having some other cause.

The former keeps the dialog open and reports the error (I have applied just the most basic styling--to be adjusted per UX guidance):
image

The latter closes the dialog and uses the standard notification system:
image

## Revised Strategy 

Per PR comments, I have updated the strategy to notify the user upon opening the dialog rather than upon sending the back-end request. This actually makes the work simpler. Here is the revised contract:

```
1. Opens new message dialog if any rules associated with project
2. Opens normal delete dialog if no rules associated with project
```
image

N/A

### :+1: Definition of Done
User is notified of rules still attached upon attempting to delete a project.

Rebuild automate-ui.
Create a project with staged rules.
Create a project with applied rules.
Create a project with no rules.
Open the delete dialog for each case; the first two should present a rules-still-attached message.

### DIFF
--- a/components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.html
+++ b/components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.html
@@ -4,10 +4,9 @@
     {{ objectAction | continuous }} {{objectNoun}} {{objectName}}
     <strong>cannot be undone</strong><span *ngIf="moreDetails"><br>{{moreDetails}}</span>.
   </div>
-  <div *ngIf="errorText" class="error">{{ errorText }}</div>
 
   <div class="options">
-    <chef-button primary caution [disabled]="!!errorText" (click)="deleteObject()">
+    <chef-button primary caution (click)="deleteObject()">
       {{objectAction}} {{objectNoun | titlecase}}
     </chef-button>
     <chef-button tertiary (click)="closeEvent()">Cancel</chef-button>
--- a/components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.html
+++ b/components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.html
@@ -1,13 +1,8 @@
 <chef-modal [visible]="visible" (closeModal)="closeEvent()">
   <h2 slot="title">{{objectAction}} {{objectNoun | titlecase}}?</h2>
   <div>
-    <span *ngIf="moreDetails">
-      {{ objectAction | continuous }} {{objectNoun}} {{objectName}} <strong>cannot be undone</strong><br>
-      {{moreDetails}}.
-    </span>
-    <span *ngIf="!moreDetails">
-      {{ objectAction | continuous }} {{objectNoun}} {{objectName}} <strong>cannot be undone</strong>.
-    </span>
+    {{ objectAction | continuous }} {{objectNoun}} {{objectName}}
+    <strong>cannot be undone</strong><span *ngIf="moreDetails"><br>{{moreDetails}}</span>.
   </div>
 
   <div class="options">
--- a/components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.html
+++ b/components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.html
@@ -4,6 +4,7 @@
     {{ objectAction | continuous }} {{objectNoun}} {{objectName}}
     <strong>cannot be undone</strong><span *ngIf="moreDetails"><br>{{moreDetails}}</span>.
   </div>
+  <div *ngIf="errorText" class="error">{{ errorText }}</div>
 
   <div class="options">
     <chef-button primary caution (click)="deleteObject()">
--- a/components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.html
+++ b/components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.html
@@ -7,7 +7,7 @@
   <div *ngIf="errorText" class="error">{{ errorText }}</div>
 
   <div class="options">
-    <chef-button primary caution (click)="deleteObject()">
+    <chef-button primary caution [disabled]="!!errorText" (click)="deleteObject()">
       {{objectAction}} {{objectNoun | titlecase}}
     </chef-button>
     <chef-button tertiary (click)="closeEvent()">Cancel</chef-button>
--- a/components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.scss
+++ b/components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.scss
@@ -12,4 +12,10 @@ chef-modal {
   .options {
     margin-top: 30px;
   }
+
+  .error {
+    margin-top: 10px;
+    font-weight: bold;
+    color: red;
+  }
 }
--- a/components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.scss
+++ b/components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.scss
@@ -12,10 +12,4 @@ chef-modal {
   .options {
     margin-top: 30px;
   }
-
-  .error {
-    margin-top: 10px;
-    font-weight: bold;
-    color: red;
-  }
 }
--- a/components/automate-ui/src/app/entities/projects/project.effects.ts
+++ b/components/automate-ui/src/app/entities/projects/project.effects.ts
@@ -6,7 +6,7 @@ import { interval as observableInterval, of as observableOf, Observable } from '
 import { catchError, mergeMap, map, filter, switchMap, withLatestFrom } from 'rxjs/operators';
 
 import { NgrxStateAtom } from 'app/ngrx.reducers';
-import { HttpStatus } from 'app/types/types';
+import { HttpStatus, GrpcErrorResponse, GrpcStatus } from 'app/types/types';
 import { CreateNotification } from 'app/entities/notifications/notification.actions';
 import { Type } from 'app/entities/notifications/notification.model';
 import { isIAMv2 } from 'app/entities/policies/policy.selectors';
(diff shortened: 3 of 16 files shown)


susanev:
@msorens i was hoping to have a different modal experience. is there any way for us to know that their project has rules when these select delete project from the control menu, instead of waiting for them to use the button in the modal?

lancewf:
I am not seeing the project removed from the table after deleting a project that does not have rules. When I refresh the page the project is removed. 
ui_project_delete

components/automate-ui/src/app/entities/projects/project.effects.ts
```diff
@@ -6,7 +6,7 @@ import { interval as observableInterval, of as observableOf, Observable } from '
 import { catchError, mergeMap, map, filter, switchMap, withLatestFrom } from 'rxjs/operators';
 
 import { NgrxStateAtom } from 'app/ngrx.reducers';
-import { HttpStatus } from 'app/types/types';
+import { HttpStatus, GrpcErrorResponse, GrpcStatus } from 'app/types/types';
```
bcmdarroch: feels like we're leaking some backend concerns with these GRPC responses. HTTP does have a [412 status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/412) for failed precondition, but I noticed the API returns a 400 (using the changes from your PR description)
```
< HTTP/2 400 
< server: nginx
< date: Mon, 09 Dec 2019 19:26:13 GMT
< content-type: application/json
< trailer: Grpc-Trailer-Content-Type
< x-xss-protection: 1; mode=block
< strict-transport-security: max-age=63072000; includeSubDomains
< x-content-type-options: nosniff
< 
{ [119 bytes data]

* Connection #0 to host localhost left intact
{
  "error": "project still has rules \"project1\"",
  "code": 9,
  "message": "project still has rules \"project1\"",
  "details": []
}
```

I'm thinking we change this UI code to look for code 412, then add a requirement to the backend card that it must return that HTTP status code

components/automate-ui/src/app/entities/projects/project.model.ts
```diff
@@ -10,10 +8,6 @@ export interface Project {
   status: ProjectStatus;
 }
 
-export function sortProjectsByName(projects: Array<Project>): Array<Project> {
```
bcmdarroch: no longer needed?

components/automate-ui/src/app/pages/project/list/project-list.component.html
```diff
@@ -92,7 +92,7 @@
         [visible]="deleteModalVisible"
         objectNoun="project"
         [objectName]="projectToDelete?.id"
-        [moreDetails]="inUseMessage()"
```
bcmdarroch: what was this for?

components/automate-ui/src/app/types/types.ts
```diff
@@ -818,7 +818,21 @@ export enum HttpStatus {
   FORBIDDEN = 403,
   NOT_FOUND = 404,
   CONFLICT = 409,
-  PRECONDITION_FAILED = 412,
+  PRECONDITION_FAILED = 412, // Warning! Grpc.FailedPrecondition maps to 400, not 412!
```
bcmdarroch: 🤔 ooh just saw this. can we make our API return the correct HTTP code?

msorens: Nope. That's what gRPC does.

msorens:
@susanev Ah, so there are two interpretations of the requirement "When a user tries to delete a project..." => I interpreted that to mean the request went to the backend; you were assuming it was when they select `Delete Project` from the control menu to open the dialog. And, I like your interpretation better. 😁 
So if you want to go that route:
1. I can revert a bit of this so that errors will always close the dialog and appear in the regular notification banner.
2. Add checking for project with rules upon opening the dialog -- and disable the button in the dialog if present.

components/automate-ui/src/app/pages/project/list/project-list.component.html
msorens: That was "for future use" in this project list view (it is used, though, for policy list) to give messaging when a resource is in use; did not need it with what I implemented (but now with Susan's changes, may have to revisit that).

components/automate-ui/src/app/entities/projects/project.effects.ts
msorens: See below. 🤷‍♂

msorens:
@susanev 	I have revised the approach per notes mentioned above.

components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.html
```diff
@@ -1,17 +1,13 @@
...
   <div>
-    <span *ngIf="moreDetails">
-      {{ objectAction | continuous }} {{objectNoun}} {{objectName}} <strong>cannot be undone</strong><br>
-      {{moreDetails}}.
-    </span>
-    <span *ngIf="!moreDetails">
-      {{ objectAction | continuous }} {{objectNoun}} {{objectName}} <strong>cannot be undone</strong>.
-    </span>
+    {{ objectAction | continuous }} {{objectNoun}} {{objectName}}
+    <strong>cannot be undone</strong><span *ngIf="moreDetails"><br>{{moreDetails}}</span>.
```
msorens: PR bonus: streamlined bit of code

components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.html
```diff
@@ -1,17 +1,13 @@
...
   </div>
+  <div *ngIf="errorText" class="error">{{ errorText }}</div>
 
   <div class="options">
-    <chef-button primary caution (click)="deleteObject()">
+    <chef-button primary caution [disabled]="!!errorText" (click)="deleteObject()">
```
msorens: This is where the action is: disable button when error is displayed.

components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.scss
```diff
@@ -12,4 +12,10 @@ chef-modal {
   .options {
     margin-top: 30px;
   }
+
+  .error {
+    margin-top: 10px;
+    font-weight: bold;
+    color: red;
+  }
```
msorens: Final styling to be determined.

components/automate-ui/src/app/entities/projects/project.effects.ts
```diff
@@ -58,116 +58,116 @@ export class ProjectEffects {
 
   @Effect()
```
msorens: This file is all whitespace!

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -57,6 +59,13 @@ export class ProjectListComponent implements OnInit, OnDestroy {
 
   private isDestroyed = new Subject<boolean>();
 
+  public statusLabel: Record<ProjectStatus, string> = {
+      'PROJECT_RULES_STATUS_UNSET': '',
+      'NO_RULES': 'No rules',
+      'EDITS_PENDING': 'Edits pending',
+      'RULES_APPLIED': 'Applied'
+  };
```
msorens: PR bonus: streamlined this accessor.

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -154,20 +162,19 @@ export class ProjectListComponent implements OnInit, OnDestroy {
   }
 
   public startProjectDelete(p: Project): void {
-    this.deleteModalVisible = true;
+    this.deleteErrorMessage =
+      ['EDITS_PENDING', 'RULES_APPLIED'].includes(p.status)
+        ? `oops ${p.id} has rules--dump them`
```
msorens: Final text of this message to be determined.

components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.html
lancewf: What does the two '!' do here?

blakestier:
This modal message is really confusing because the user can't actually delete the project, so the confirmation question is misleading. 

We're planning to fix that in a separate PR? Seems like we def need a different modal that _instructs_ the user what needs to be done. (Like, "This project cannot be deleted as it has active rules applied. First, delete those rules and then run "Update Projects" before deleting". . . or something)

components/automate-ui/src/app/types/types.ts
```diff
@@ -818,7 +818,9 @@ export enum HttpStatus {
...
   INTERNAL_SERVER_ERROR = 500
 }
 
+// Map of gRPC codes to HTTP codes is available at:
```
bcmdarroch: do we still need this?

components/automate-ui/src/app/components/delete-object-modal/delete-object-modal.component.html
msorens: Shorthand for convert-to-boolean

blakestier:
The code looks good- I think the "Are you sure you want to delete this" messaging (when you can't actually delete it) is super confusing-- enough so that I'm tempted to wait on this until we've resolved it, but 🤷‍♂

msorens:
Added DO-NOT-MERGE becuase this is waiting on UX review.

susanev:
@msorens this is what id like the modal experience to be in this case, let me know if this is possible.

cannot-delete-project

msorens:
@susanev Could we go with one of these variations (for grammatical reasons)? 
1. Before you can delete this project, delete all ingest rules and update projects.
2. Before you can delete this project, delete all ingest rules, then update projects.

susanev:
grammarly told me otherwise, could you run it by docs folks.. whatever they say sounds good to me.

components/automate-ui/src/app/pages/project/list/project-list.component.html
```diff
@@ -111,6 +110,13 @@
         (confirm)="confirmApplyStop()"
         (cancel)="cancelApplyStop()">
       </app-confirm-apply-stop-modal>
+
+      <app-message-modal
+        [title]="'Could Not Delete Project'"
+        [visible]="messageModalVisible"
+        (close)="closeMessageModal()">
+        Before you can delete this project, delete all ingest rules, and then update projects.
```
msorens: (Grammar confirmed with @mjingle .)

components/automate-ui/src/app/pages/project/list/project-list.component.html
```diff
@@ -111,6 +110,13 @@
         (confirm)="confirmApplyStop()"
         (cancel)="cancelApplyStop()">
       </app-confirm-apply-stop-modal>
+
+      <app-message-modal
+        [title]="'Could Not Delete Project'"
```
bcmdarroch: @susanev should weigh in on this, but should we change the color of the title or message so it's more clear something is wrong? Right now it looks similar enough to the regular delete modal that folks might click `close` mindlessly without reading the message.
Screen Shot 2019-12-17 at 3 44 29 PM
Screen Shot 2019-12-17 at 3 44 40 PM

msorens: Confirmed with @susanev to leave as is for now. Thanks for raising the question!

PR #2461, opened 2019-12-09
fix revendor target and ignore EINVAL in atomic writer

stevendanna:
This fixes two problems that led to `revendor` not working correctly.

Previously, we were copying files from grpc-ecosystem/grpc-gateway
manually into our vendor directory. However, to determine the path to
copy from, we needed to look up the the path after installation. But,
because of some confusion about how `$(eval)` works, this did not work
on systems where that go module was not already installed.

To fix this, I've moved this logic to a shell script since I think
more people will be able to reason about the behavior of the shell
script.

Additionally, we now ignore EINVAL from directory syncs in the atomic
writer. Vboxfs and potentially other filesystems don't support calling
sync on a directory, returning EINVAL. This has been a recurring
source of confusion. Other bits of code turn of sync'ing completely to
avoid this issue, but that is a bit overkill.

### DIFF
--- a/lib/io/fileutils/atomic_write.go
+++ b/lib/io/fileutils/atomic_write.go
@@ -1,11 +1,13 @@
 package fileutils
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"os"
 	"path"
 	"path/filepath"
+	"syscall"
 
 	"go.uber.org/multierr"
 )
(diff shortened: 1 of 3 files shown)


lib/io/fileutils/atomic_write.go
```diff
@@ -190,8 +192,13 @@ func syncDir(dirname string) error {
 	}
 
 	if err := dir.Sync(); err != nil {
-		return err
+		// Some filesystems (vboxfs) return EINVAL when trying
+		// to call sync on a directory. We ignore the error in
+		// this case since the directory sync doesn't impact
+		// atomicity (only durability).
+		if !errors.Is(err, syscall.EINVAL) {
+			return err
```
jaym: :)

PR #2462, opened 2019-12-09
minor facade cleanup

msorens:

As I was "near" this code, I noticed a few odds and ends to streamline and make slightly more maintainable.

_Follow the commits for specific details._

N/A

### :+1: Definition of Done
Everything still works.

rebuild automate-ui


components/automate-ui/src/app/entities/layout/layout.facade.ts
```diff
@@ -1,115 +1,107 @@
 import { Injectable } from '@angular/core';
 import { Store } from '@ngrx/store';
```
msorens: This file is best-viewed ignoring whitespace changes!!

PR #2467, opened 2019-12-09
automate-2281 automate-2278

### LINKED ISSUE #2278: setting sidebar no longer reacts to IAM version change without refresh
## Describe the bug
Previously, the settings sidebar was subscribed to the IAM Version on the store and would instantly react to IAM version changes, displaying the correct menu items for each version. 

Since the layout update, a refresh is required to display the correct sidebar.

## To Reproduce
**Reset: IAM v2 -> IAM v1**
- your local env will start on IAM v2.1 by default
- navigate to any settings page
- run `chef-automate iam reset-to-v1`
- you can still navigate to the V2-only pages (roles, projects, policies) but they will be empty
- to see the V1 sidebar, refresh

**Upgrade: IAM v1 -> IAM v2**
- now run `chef-automate iam upgrade-to-v2`
- you will not see any V2 sidebar items unless you refresh or navigate to a V2 URL directly (https://a2-dev.test/settings/policies)

## Expected behavior
When I upgrade to V2, the UI should react quickly and display the Access Management links.
```
Access Management
- Policies
- Roles
- Projects
```

When I reset to V1, the UI should react quickly and remove the Access Management links, so that I only see
```
Identity
- Users
- Teams
- API Tokens
```

### LINKED ISSUE #2281: workflow left nav item missing
## Overview
Previously, on the infrastructure left nav there was a workflow nav item, its gone now.

missing in dev: https://a2-workflow-local-fresh-install-dev.cd.chef.co/infrastructure/client-runs
there in acceptance: https://a2-workflow-local-fresh-install-acceptance.cd.chef.co/infrastructure/client-runs

tarablack01:
Automate-2281 - Workflow nav item is missing
Previously, on the infrastructure left nav there was a workflow nav item, its gone now.
missing in dev: https://a2-workflow-local-fresh-install-dev.cd.chef.co/infrastructure/client-runs
there in acceptance: https://a2-workflow-local-fresh-install-acceptance.cd.chef.co/infrastructure/client-runs

Automate-2278 - setting sidebar no longer reacts to IAM version change without refresh
Previously, the settings sidebar was subscribed to the IAM Version on the store and would instantly react to IAM version changes, displaying the correct menu items for each version.
Since the layout update, a refresh is required to display the correct sidebar.

fixes #2281

fixes #2278

### :+1: Definition of Done
Workflow nav shows when clientRunsWorkflowEnabled is enabled.

Settings sidebar menu items show correct links in IAMVersion.

chef-automate iam reset-to-v1
start_workflow_service

Screen Shot 2019-12-09 at 2 22 27 PM
Screen Shot 2019-12-09 at 2 21 53 PM
Screen Shot 2019-12-09 at 2 32 20 PM

### DIFF
--- a/components/automate-ui/src/app/components/authorized/authorized.component.ts
+++ b/components/automate-ui/src/app/components/authorized/authorized.component.ts
@@ -1,4 +1,4 @@
-import { Component, Input, OnInit, OnDestroy, ChangeDetectorRef } from '@angular/core';
+import { Component, Input, EventEmitter, OnInit, OnDestroy, ChangeDetectorRef, Output } from '@angular/core';
 import { Store } from '@ngrx/store';
 import { isEmpty, isArray, pipe } from 'lodash/fp';
 import { takeUntil, filter, debounceTime } from 'rxjs/operators';
@@ -30,12 +30,12 @@ export class AuthorizedComponent implements OnInit, OnDestroy {
 
   @Input()
   set allOf(val: Check[] | Check) {
-    this._allOf = this.normalizeInput(val);
+    this._allOf = val ? this.normalizeInput(val) : undefined;
   }
 
   @Input()
   set anyOf(val: Check[] | Check) {
-    this._anyOf = this.normalizeInput(val);
+    this._anyOf = val ? this.normalizeInput(val) : undefined;
   }
 
   // Include the bare `not` attribute in your HTML element to negate the check.
@@ -45,6 +45,9 @@ export class AuthorizedComponent implements OnInit, OnDestroy {
   // Defaults to false. Useful when trying to generalize components that contain
   // this component.
   @Input() overrideVisible = false;
+
+  // if you wish to output visible
+  @Output() isAuthorized: EventEmitter<boolean> = new EventEmitter();
 
   public visible = false;
   private isDestroyed: Subject<boolean> = new Subject<boolean>();
@@ -68,6 +71,7 @@ export class AuthorizedComponent implements OnInit, OnDestroy {
         const oldVisible = this.visible;
         const newVisible = this.authorizedChecker.evalPerms(perms);
         this.visible = this.not === undefined ? newVisible : !newVisible;
+        this.isAuthorized.emit(this.visible);
 
         // Allow this component to receive updates on pages with "OnPush" strategy.
         if (oldVisible !== this.visible) {
--- a/components/automate-ui/src/app/components/authorized/authorized.component.ts
+++ b/components/automate-ui/src/app/components/authorized/authorized.component.ts
@@ -30,12 +30,12 @@ export class AuthorizedComponent implements OnInit, OnDestroy {
 
   @Input()
   set allOf(val: Check[] | Check) {
-    this._allOf = val ? this.normalizeInput(val) : undefined;
+    this._allOf = val ? this.normalizeInput(val) : [];
   }
 
   @Input()
   set anyOf(val: Check[] | Check) {
-    this._anyOf = val ? this.normalizeInput(val) : undefined;
+    this._anyOf = val ? this.normalizeInput(val) : [];
   }
 
   // Include the bare `not` attribute in your HTML element to negate the check.
--- a/components/automate-ui/src/app/components/sidebar/sidebar.component.html
+++ b/components/automate-ui/src/app/components/sidebar/sidebar.component.html
@@ -6,34 +6,23 @@
     <div class="nav-items">
       <span *ngFor="let menuGroup of menuGroups$ | async">
         <div
-          *ngIf="isGroupVisible(menuGroup)"
+          *ngIf="(menuGroup.visible | async) || hasAuthroizedMenuItems(menuGroup)"
           class="menu-item-groups"
         >
           <div class="group">{{ menuGroup.name }}</div>
           <span *ngFor="let menuItem of menuGroup.items">
-            <span *ngIf="menuItem.visible && !menuItem.authorized">
+            <span *ngIf="!menuItem.authorized && (menuItem.visible | async)">
               <chef-sidebar-entry
                 [route]="menuItem.route"
                 [icon]="menuItem.icon"
                 [openInNewPage]="menuItem.openInNewPage"
               >{{ menuItem.name }}</chef-sidebar-entry>
             </span>
-            <span *ngIf="menuItem.visible && menuItem.authorized && menuItem.authorized.anyOf">
-              <app-authorized
-                [menuItem.authorized.name]
-                [anyOf]="menuItem.authorized.anyOf"
-              >
-                <chef-sidebar-entry
-                  [route]="menuItem.route"
-                  [icon]="menuItem.icon"
-                  [openInNewPage]="menuItem.openInNewPage"
-                >{{ menuItem.name }}</chef-sidebar-entry>
-              </app-authorized>
-            </span>
-            <span *ngIf="menuItem.visible && menuItem.authorized && menuItem.authorized.allOf">
+            <span *ngIf="menuItem.authorized && (menuItem.visible | async)">
                 <app-authorized
-                  [menuItem.authorized.name]
+                  [anyOf]="menuItem.authorized.anyOf"
                   [allOf]="menuItem.authorized.allOf"
+                  (isAuthorized)="isAuthorized($event, menuItem)"
                 >
                   <chef-sidebar-entry
                     [route]="menuItem.route"
--- a/components/automate-ui/src/app/components/sidebar/sidebar.component.html
+++ b/components/automate-ui/src/app/components/sidebar/sidebar.component.html
@@ -6,19 +6,19 @@
     <div class="nav-items">
       <span *ngFor="let menuGroup of menuGroups$ | async">
         <div
-          *ngIf="isVisible(menuGroup.visible) | async"
+          *ngIf="isGroupVisible(menuGroup)"
           class="menu-item-groups"
         >
           <div class="group">{{ menuGroup.name }}</div>
           <span *ngFor="let menuItem of menuGroup.items">
-            <span *ngIf="(isVisible(menuItem.visible) | async) && !menuItem.authorized">
+            <span *ngIf="menuItem.visible && !menuItem.authorized">
               <chef-sidebar-entry
                 [route]="menuItem.route"
                 [icon]="menuItem.icon"
                 [openInNewPage]="menuItem.openInNewPage"
               >{{ menuItem.name }}</chef-sidebar-entry>
             </span>
-            <span *ngIf="(isVisible(menuItem.visible) | async) && menuItem.authorized && menuItem.authorized.anyOf">
+            <span *ngIf="menuItem.visible && menuItem.authorized && menuItem.authorized.anyOf">
               <app-authorized
                 [menuItem.authorized.name]
                 [anyOf]="menuItem.authorized.anyOf"
@@ -30,7 +30,7 @@
                 >{{ menuItem.name }}</chef-sidebar-entry>
               </app-authorized>
             </span>
-            <span *ngIf="(isVisible(menuItem.visible) | async) && menuItem.authorized && menuItem.authorized.allOf">
+            <span *ngIf="menuItem.visible && menuItem.authorized && menuItem.authorized.allOf">
                 <app-authorized
                   [menuItem.authorized.name]
                   [allOf]="menuItem.authorized.allOf"
--- a/components/automate-ui/src/app/components/sidebar/sidebar.component.html
+++ b/components/automate-ui/src/app/components/sidebar/sidebar.component.html
@@ -6,7 +6,7 @@
     <div class="nav-items">
       <span *ngFor="let menuGroup of menuGroups$ | async">
         <div
-          [hidden]="!((menuGroup.visible$ | async) && menuGroup.hasAuthroizedMenuItems)"
+          [hidden]="!((menuGroup.visible$ | async) && menuGroup.hasAuthorizedMenuItems)"
           class="menu-item-groups"
         >
           <div class="group">{{ menuGroup.name }}</div>
--- a/components/automate-ui/src/app/components/sidebar/sidebar.component.html
+++ b/components/automate-ui/src/app/components/sidebar/sidebar.component.html
@@ -6,7 +6,7 @@
     <div class="nav-items">
       <span *ngFor="let menuGroup of menuGroups$ | async">
         <div
-          [hidden]="!((menuGroup.visible$ | async) && menuGroup.hasAuthorizedMenuItems)"
+          [hidden]="!((menuGroup.visible$ | async) && menuGroup.hasVisibleMenuItems)"
           class="menu-item-groups"
         >
           <div class="group">{{ menuGroup.name }}</div>
--- a/components/automate-ui/src/app/components/sidebar/sidebar.component.html
+++ b/components/automate-ui/src/app/components/sidebar/sidebar.component.html
@@ -6,23 +6,23 @@
     <div class="nav-items">
       <span *ngFor="let menuGroup of menuGroups$ | async">
         <div
-          *ngIf="(menuGroup.visible | async) || hasAuthroizedMenuItems(menuGroup)"
+          [hidden]="!((menuGroup.visible$ | async) && menuGroup.hasAuthroizedMenuItems)"
           class="menu-item-groups"
         >
           <div class="group">{{ menuGroup.name }}</div>
           <span *ngFor="let menuItem of menuGroup.items">
-            <span *ngIf="!menuItem.authorized && (menuItem.visible | async)">
+            <span *ngIf="!menuItem.authorized && (menuItem.visible$ | async)">
               <chef-sidebar-entry
                 [route]="menuItem.route"
                 [icon]="menuItem.icon"
                 [openInNewPage]="menuItem.openInNewPage"
               >{{ menuItem.name }}</chef-sidebar-entry>
             </span>
-            <span *ngIf="menuItem.authorized && (menuItem.visible | async)">
+            <span *ngIf="menuItem.authorized && (menuItem.visible$ | async)">
                 <app-authorized
                   [anyOf]="menuItem.authorized.anyOf"
                   [allOf]="menuItem.authorized.allOf"
-                  (isAuthorized)="isAuthorized($event, menuItem)"
+                  (isAuthorized)="isAuthorized($event, menuItem, menuGroup)"
                 >
                   <chef-sidebar-entry
                     [route]="menuItem.route"
--- a/components/automate-ui/src/app/components/sidebar/sidebar.component.ts
+++ b/components/automate-ui/src/app/components/sidebar/sidebar.component.ts
@@ -1,5 +1,6 @@
 import { Component, Inject } from '@angular/core';
-import { Observable, BehaviorSubject } from 'rxjs';
+import { Observable } from 'rxjs';
+import { find } from 'lodash';
 
 import { LayoutFacadeService } from 'app/entities/layout/layout.facade';
 import { MenuItemGroup } from 'app/entities/layout/layout.model';
@@ -18,9 +19,7 @@ export class SidebarComponent {
     this.menuGroups$ = layoutFacade.sidebar$;
   }
 
-  public isVisible(item: any) {
-    return item.subscribe
-      ? item
-      : new BehaviorSubject(item);
+  public isGroupVisible(menuGroup: MenuItemGroup) {
+    return menuGroup.visible && menuGroup.items && find(menuGroup.items, ['visible', true]);
   }
 }
--- a/components/automate-ui/src/app/components/sidebar/sidebar.component.ts
+++ b/components/automate-ui/src/app/components/sidebar/sidebar.component.ts
@@ -3,7 +3,7 @@ import { Observable } from 'rxjs';
 import { find } from 'lodash';
 
 import { LayoutFacadeService } from 'app/entities/layout/layout.facade';
-import { MenuItemGroup } from 'app/entities/layout/layout.model';
+import { MenuItemGroup, MenuItem } from 'app/entities/layout/layout.model';
 
 @Component({
   selector: 'chef-sidebar',
@@ -19,7 +19,7 @@ export class SidebarComponent {
     this.menuGroups$ = layoutFacade.sidebar$;
   }
 
-  public isGroupVisible(menuGroup: MenuItemGroup) {
+  public isGroupVisible(menuGroup: MenuItemGroup): MenuItem {
     return menuGroup.visible && menuGroup.items && find(menuGroup.items, ['visible', true]);
   }
 }
--- a/components/automate-ui/src/app/components/sidebar/sidebar.component.ts
+++ b/components/automate-ui/src/app/components/sidebar/sidebar.component.ts
@@ -15,7 +15,7 @@ export class SidebarComponent {
   constructor(
     @Inject(LayoutFacadeService) public layoutFacade: LayoutFacadeService
   ) {
-    this.menuGroups$ = layoutFacade.menuGroups$;
+    this.menuGroups$ = layoutFacade.sidebar$;
   }
 
   public isVisible(item: any) {
--- a/components/automate-ui/src/app/components/sidebar/sidebar.component.ts
+++ b/components/automate-ui/src/app/components/sidebar/sidebar.component.ts
@@ -1,9 +1,9 @@
 import { Component, Inject } from '@angular/core';
 import { Observable } from 'rxjs';
-import { find } from 'lodash';
+import { find } from 'lodash/fp';
 
 import { LayoutFacadeService } from 'app/entities/layout/layout.facade';
-import { MenuItemGroup, MenuItem } from 'app/entities/layout/layout.model';
+import { MenuItemGroup } from 'app/entities/layout/layout.model';
 
 @Component({
   selector: 'chef-sidebar',
@@ -19,7 +19,14 @@ export class SidebarComponent {
     this.menuGroups$ = layoutFacade.sidebar$;
   }
 
-  public isGroupVisible(menuGroup: MenuItemGroup): MenuItem {
-    return menuGroup.visible && menuGroup.items && find(menuGroup.items, ['visible', true]);
+  public hasAuthroizedMenuItems(menuItemGroup: any): boolean {
+    const authorizedItem = find(menuItemGroup.items, (menuItem) => {
+      return menuItem.authorized && menuItem.authorized.isAuthorized;
+    });
+    return authorizedItem !== undefined;
+  }
+
+  public isAuthorized($event, menuItem) {
+    menuItem.authorized.isAuthorized = $event;
   }
 }
--- a/components/automate-ui/src/app/components/sidebar/sidebar.component.ts
+++ b/components/automate-ui/src/app/components/sidebar/sidebar.component.ts
@@ -19,32 +19,32 @@ export class SidebarComponent {
     this.updateMenuGroupVisibility();
   }
 
-  public updateMenuGroupVisibility(): void {
+  public isAuthorized($event, menuItem, menuGroup) {
+    menuItem.authorized.isAuthorized = $event;
+    this.setGroupVisibility(menuGroup);
+  }
+
+  public hasMenuItemsNotRequiringAuthorization(menuItemGroup: any): boolean {
+    return menuItemGroup.items.some(menuItem =>
+      menuItem.authorized === undefined);
+  }
+
+  private updateMenuGroupVisibility(): void {
     this.menuGroups$.subscribe((menuGroups: MenuItemGroup[]) => {
       if (menuGroups) {
-        menuGroups.forEach(menuItemGroup => this.hasVisibleMenuItems(menuItemGroup));
+        menuGroups.forEach(menuItemGroup => this.setGroupVisibility(menuItemGroup));
       }
     });
   }
 
-  public hasVisibleMenuItems(menuItemGroup: any): void {
-    menuItemGroup.hasAuthorizedMenuItems =
-      this.hasAuthorizedMenuItems(menuItemGroup) ||
-      this.hasNoAuthorizedMenuItems(menuItemGroup);
+  private setGroupVisibility(menuItemGroup: MenuItemGroup): void {
+    menuItemGroup.hasVisibleMenuItems =
+      this.hasVisibleMenuItems(menuItemGroup) ||
+      this.hasMenuItemsNotRequiringAuthorization(menuItemGroup);
   }
 
-  public hasAuthorizedMenuItems(menuItemGroup: any): boolean {
-    return menuItemGroup.items.filter(menuItem =>
-        menuItem.authorized && menuItem.authorized.isAuthorized).length > 0;
-  }
-
-  public hasNoAuthorizedMenuItems(menuItemGroup: any): boolean {
-    return menuItemGroup.items.filter(menuItem =>
-        menuItem.authorized === undefined).length > 0;
-  }
-
-  public isAuthorized($event, menuItem, menuGroup) {
-    menuItem.authorized.isAuthorized = $event;
-    this.hasVisibleMenuItems(menuGroup);
+  private hasVisibleMenuItems(menuItemGroup: any): boolean {
+    return  menuItemGroup.items.some(menuItem =>
+          menuItem.authorized && menuItem.authorized.isAuthorized);
   }
 }
--- a/components/automate-ui/src/app/components/sidebar/sidebar.component.ts
+++ b/components/automate-ui/src/app/components/sidebar/sidebar.component.ts
@@ -16,16 +16,35 @@ export class SidebarComponent {
     @Inject(LayoutFacadeService) public layoutFacade: LayoutFacadeService
   ) {
     this.menuGroups$ = layoutFacade.sidebar$;
+    this.updateMenuGroupVisibility();
   }
 
-  public hasAuthroizedMenuItems(menuItemGroup: any): void {
+  public updateMenuGroupVisibility(): void {
+    this.menuGroups$.subscribe((menuGroups: MenuItemGroup[]) => {
+      if (menuGroups) {
+        menuGroups.forEach(menuItemGroup => this.hasVisibleMenuItems(menuItemGroup));
+      }
+    });
+  }
+
+  public hasVisibleMenuItems(menuItemGroup: any): void {
     menuItemGroup.hasAuthroizedMenuItems =
-      menuItemGroup.items.filter(menuItem =>
+      this.hasAuthroizedMenuItems(menuItemGroup) ||
+      this.hasNoAuthroizedMenuItems(menuItemGroup);
+  }
+
+  public hasAuthroizedMenuItems(menuItemGroup: any): boolean {
+    return menuItemGroup.items.filter(menuItem =>
         menuItem.authorized && menuItem.authorized.isAuthorized).length > 0;
+  }
+
+  public hasNoAuthroizedMenuItems(menuItemGroup: any): boolean {
+    return menuItemGroup.items.filter(menuItem =>
+        menuItem.authorized === undefined).length > 0;
   }
 
   public isAuthorized($event, menuItem, menuGroup) {
     menuItem.authorized.isAuthorized = $event;
-    this.hasAuthroizedMenuItems(menuGroup);
+    this.hasVisibleMenuItems(menuGroup);
   }
 }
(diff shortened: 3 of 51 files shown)


components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
```diff
@@ -1,278 +1,289 @@
...
-
     constructor(
         private store: Store<NgrxStateAtom>,
         private clientRunsStore: Store<fromClientRuns.ClientRunsEntityState>,
         private featureFlagsService: FeatureFlagsService
     ) {
-        this.isIAMv2$ = this.store.select(isIAMv2);
         this.applicationsFeatureFlagOn = this.featureFlagsService.getFeatureStatus('applications');
         this.ServiceNowfeatureFlagOn = this.featureFlagsService.getFeatureStatus('servicenow_cmdb');
+        this.store.select(isIAMv2).subscribe(
```
msorens: Need to add the standard `takeUntil(this.isDestroyed)` before a `subscribe`
both here and on L41 below.

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
```diff
@@ -1,278 +1,289 @@
...
-                        name: 'Policies',
-                        icon: 'security',
-                        route: '/settings/policies',
-                        authorized: {
-                            name: 'policies',
-                            allOf: ['/iam/v2beta/policies', 'get']
+            }],
+            settings: [
+                {
+                    name: 'Node Management',
```
msorens: Have not dug in far enough yet, but do you have a mechanism to display/hide this group name based on the collection of anyOf/allOf of its children?

components/automate-ui/src/app/modules/policy/details/policy-details.component.ts
```diff
@@ -46,7 +46,7 @@ export class PolicyDetailsComponent implements OnInit, OnDestroy {
   ) {  }
 
   ngOnInit(): void {
-    this.layoutFacade.showSettingsSidebar();
+    this.layoutFacade.showSidebar('settings');
```
msorens: I would prefer to see the arguments to `showSidebar` come from an enum, so this would look like:
```suggestion
    this.layoutFacade.showSidebar(Sidebar.Settings);
```
That makes it much less fragile (due to typos, etc.)

components/automate-ui/src/app/components/sidebar/sidebar.component.ts
```diff
@@ -15,7 +15,7 @@ export class SidebarComponent {
   constructor(
     @Inject(LayoutFacadeService) layoutFacade: LayoutFacadeService
   ) {
-    this.menuGroups$ = layoutFacade.menuGroups$;
+    this.menuGroups$ = layoutFacade.sidebar$;
   }
 
   public isVisible(item: any) {
```
msorens: Can stongly type the return value.
```suggestion
  public isVisible(item: any): Observable<boolean> {
```
I do not see any easy fix for the parameter, but this helps, I think:
```suggestion
  public isVisible(item: any /* boolean | Observable<boolean> */): Observable<boolean> {
```

That said, I do not see anywhere that you pass an `Observable` -- every occurrence is a plain boolean, so could simplify this function, I think, to just return a boolean
(and also change the MenuItemGroup.visible to be boolean instead of any in layout.model.ts).

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
```diff
@@ -1,278 +1,289 @@
...
 @Injectable({
     providedIn: 'root'
 })
-export class LayoutSidebarService {
+export class LayoutSidebarService implements OnInit, OnDestroy {
+    public sidebars = new BehaviorSubject<Sidebars>({});
     public applicationsFeatureFlagOn: boolean;
-    public isIAMv2$: Observable<boolean>;
+    public isIAMv2: boolean;
     public ServiceNowfeatureFlagOn: boolean;
```
msorens:
```suggestion
     public ServiceNowFeatureFlagOn: boolean;
```

components/automate-ui/src/app/modules/policy/details/policy-details.component.ts
tarablack01: @msorens Thanks for brining that up. I would agree. I'll add that! 😊

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
tarablack01: @msorens good point! I'll add that logic.

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
```diff
@@ -1,278 +1,289 @@
...
+import {
+    UpdateSidebars
+} from './layout.actions';
 
 @Injectable({
     providedIn: 'root'
 })
-export class LayoutSidebarService {
+export class LayoutSidebarService implements OnInit, OnDestroy {
+    public sidebars = new BehaviorSubject<Sidebars>({});
```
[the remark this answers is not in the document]
tarablack01: @msorens No it's not, it was left behind. I'll clean it up.

components/automate-ui/src/app/components/sidebar/sidebar.component.ts
tarablack01: @msorens It was previously passing either a boolean or observable. I now have control over that and can remove the observable logic.👍🏻

components/automate-ui/src/app/entities/layout/layout.facade.ts
```diff
@@ -8,32 +8,41 @@ import { Notification, Type } from 'app/entities/notifications/notification.mode
...
+  Compliance = 'compliance',
+  Settings = 'settings',
+  Profile = 'profile'
+}
+
 @Injectable({
   providedIn: 'root'
 })
 export class LayoutFacadeService {
   headerHeight = '70px';
```
SEAjamieD: Is this default headerHeight because by default we are not dealing with any notifications?  If that is the case, my thought is to use `Height.Navigation` so that there's only one location to modify.

tarablack01: @SEAjamieD Thanks for catching that! Updated👍🏻

susanev:
tested out the workflow part of this pr, and it works
still need the headings in settings to react to what pages users have access to

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
```diff
@@ -1,286 +1,295 @@
...
-        this.isIAMv2$ = this.store.select(isIAMv2);
         this.applicationsFeatureFlagOn = this.featureFlagsService.getFeatureStatus('applications');
-        this.ServiceNowfeatureFlagOn = this.featureFlagsService.getFeatureStatus('servicenow_cmdb');
+        this.ServiceNowFeatureFlagOn = this.featureFlagsService.getFeatureStatus('servicenow_cmdb');
+        this.store.select(isIAMv2).pipe(
+          takeUntil(this.isDestroyed)
+        ).subscribe(
+            (IAMv2) => {
+                this.isIAMv2 = IAMv2;
+                this.updateSidebars();
```
msorens: This appears to be wired up properly but it stops after this error occurs:
```
ERROR TypeError: You provided 'undefined' where a stream was expected. You can provide an Observable, Promise, Array, or Iterable.
    at subscribeTo (subscribeTo.js:27)
    at subscribeToResult (subscribeToResult.js:11)
    at SwitchMapSubscriber._innerSub (switchMap.js:46)
    at SwitchMapSubscriber._next (switchMap.js:36)
    at SwitchMapSubscriber.next (Subscriber.js:49)
    at WithLatestFromSubscriber._next (withLatestFrom.js:57)
    at WithLatestFromSubscriber.next (Subscriber.js:49)
    at WithLatestFromSubscriber._next (withLatestFrom.js:57)
    at WithLatestFromSubscriber.next (Subscriber.js:49)
    at FilterSubscriber._next (filter.js:33)
```
Unfortunately, that error trace is not very helpful in identifying where the error stems from. 🤷‍♂ 

To reproduce:
1. Change this line to 10 seconds for testing purpose: https://github.com/chef/automate/blob/70df1da94a0d8e40da232c4ca6e4b1382de57a84/components/automate-ui/src/app/entities/policies/policy.effects.ts#L53
2. rebuild automate-ui
3. Start with `chef-automate iam reset-to-v1`
4.  with settings page open, refresh browser and open dev tools (<kbd>cmd</kbd> + <kbd>opt</kbd> + <kbd>i</kbd>).
5. upgrade with `chef-automate iam upgrade-to-v2 --skip-policy-migration --beta2.1`
6. observe brower updates no more than 10 seconds later, **adding** "ACCESS MANAGEMENT" group in sidebar.
7. Downgrade again with `chef-automate iam reset-to-v1`
8. observe brower updates no more than 10 seconds later, **removing** "ACCESS MANAGEMENT" group in sidebar.
9. Also observe above error trace in browser console.


**Any further upgrades/downgrades will not update the sidebar.**

components/automate-ui/src/app/components/sidebar/sidebar.component.ts
```diff
@@ -15,12 +16,10 @@ export class SidebarComponent {
...
   ) {
-    this.menuGroups$ = layoutFacade.menuGroups$;
+    this.menuGroups$ = layoutFacade.sidebar$;
   }
 
-  public isVisible(item: any) {
-    return item.subscribe
-      ? item
-      : new BehaviorSubject(item);
+  public isGroupVisible(menuGroup: MenuItemGroup) {
```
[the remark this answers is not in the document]
tarablack01: @msorens it's returning the MenuItem, I will update the expected return value.

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
```diff
@@ -1,286 +1,295 @@
...
+      this.isDestroyed.next(true);
+      this.isDestroyed.complete();
     }
 
-    public getInfastructureSidebar(): MenuItemGroup[] {
-        return [{
-            name: 'Infrastructure',
+    public updateSidebars(sidebarName?: string): void {
+      this.activeSidebar = sidebarName || this.activeSidebar;
+      this.store.dispatch(new UpdateSidebars({
```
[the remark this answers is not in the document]
tarablack01: @msorens 
1. The sidebar object isn't constant, there is a dynamic value that changes each time the sidebar is updated to flag which section of the sidebar to display. I've removed it from the function into a private variable to separate the code and data.👍🏻
2. I've added the in-code comment.

components/automate-ui/src/app/entities/layout/layout.facade.ts
```diff
@@ -42,26 +51,26 @@ export class LayoutFacadeService {
...
+    });
   }
 
   hasGlobalNotifications(): boolean {
     return this.headerHeight === Height.Banner;
   }
 
-  ShowPageLoading(showLoading: boolean): void {
-    this.store.dispatch(new ShowPageLoading(showLoading));
+  ShowPageLoading(showLoading: boolean) {
```
blakestier: I'm an angular newbie-- can you explain why we drop the `void` for some, but not all of these? It looks like it clarifies whether we expect anything returned.

blakestier: (or is the thing that looks like the arg the return?)

components/automate-ui/src/app/entities/layout/layout.actions.ts
```diff
@@ -1,22 +1,29 @@
...
 }
 
 export class ShowPageLoading implements Action {
   readonly type = LayoutActionTypes.SHOW_PAGE_LOADING;
   constructor(public payload: boolean) {}
 }
 
-export class UpdateSidebarMenuGroups implements Action {
-  readonly type = LayoutActionTypes.UPDATE_SIDEBAR_MENU_GROUPS;
+export class GetSidebar implements Action {
```
blakestier: Again, I'm new here-- is there one sidebar or many? We have a `GetSidebar` but `UpdateSidebars` in the plural

components/automate-ui/src/app/entities/layout/layout.facade.ts
```diff
@@ -70,39 +79,7 @@ export class LayoutFacadeService {
...
-    this.store.dispatch(new UpdateSidebarMenuGroups(
-      this.layoutSidebarService.getSettingsSidebar()
-    ));
-  }
-
-  showUserProfileSidebar(): void {
-    this.store.dispatch(new UpdateSidebarMenuGroups(
-      this.layoutSidebarService.getUserProfileSidebar()
-    ));
+  showSidebar(sidebarName: string) {
```
blakestier: what's the difference between enabling a sidebar and showing a sidebar?

susanev:
@tarablack01 i tested this out, and was still having some trouble with the settings left nav.. i included curl commands for policies that might help in testing; let me know if i can help in any way

- when signed in as an admin, the access management section is totally missing even though i can navigate directly to the pages just fine
access-managment

- when signed in as a user with only permissions to node credentials, i still see identity
only-node-creds
```
curl -k -sSH "api-token: $TOKEN" -d '{"name": "node-creds", "id": "node-creds", "projects": [], "members": [], "statements": [{"effect": "ALLOW", "actions": ["secrets:*"], "projects": ["*"]}]}' -X POST https://a2-dev.test/apis/iam/v2beta/policies?pretty
```

- when signed in as a user with only permissions on users i still see node management
only-users

```
curl -k -sSH "api-token: $TOKEN" -d '{"name": "users", "id": "users", "projects": [], "members": [], "statements": [{"effect": "ALLOW", "actions": ["iam:users:*"], "projects": ["*"]}]}' -X POST https://a2-dev.test/apis/iam/v2beta/policies?pretty
```

- when signed in as a user with only permissions on roles, i see both identity and node managment headings
node-mang-and-ident

```
curl -k -sSH "api-token: $TOKEN" -d '{"name": "roles", "id": "roles", "projects": [], "members": [], "statements": [{"effect": "ALLOW", "actions": ["iam:roles:*"], "projects": ["*"]}]}' -X POST https://a2-dev.test/apis/iam/v2beta/policies?pretty
```

components/automate-ui/src/app/components/sidebar/sidebar.component.html
```diff
@@ -6,19 +6,19 @@
...
       <span *ngFor="let menuGroup of menuGroups$ | async">
         <div
-          *ngIf="isVisible(menuGroup.visible) | async"
+          *ngIf="isGroupVisible(menuGroup)"
           class="menu-item-groups"
         >
           <div class="group">{{ menuGroup.name }}</div>
           <span *ngFor="let menuItem of menuGroup.items">
-            <span *ngIf="(isVisible(menuItem.visible) | async) && !menuItem.authorized">
+            <span *ngIf="menuItem.visible && !menuItem.authorized">
```
bcmdarroch: why remove `async` here?

components/automate-ui/src/app/components/sidebar/sidebar.component.ts
```diff
@@ -15,12 +16,10 @@ export class SidebarComponent {
...
+    return menuGroup.visible && menuGroup.items && find(menuGroup.items, ['visible', true]);
```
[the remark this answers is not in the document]
tarablack01: @msorens I believe it's coded correctly, if you don't mind testing it :)

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
tarablack01: @msorens if you could check if you still have the same issue. 👍🏻

components/automate-ui/src/app/entities/layout/layout.facade.ts
tarablack01: @blakestier The void represents the type of value returned, in this case nothing.

components/automate-ui/src/app/entities/layout/layout.actions.ts
tarablack01: @blakestier we have sidebars and we get one of those sidebars.

components/automate-ui/src/app/entities/layout/layout.facade.ts
tarablack01: @blakestier true, we can rename it. We call this in the pages to specify which sidebar to display or enable. this.layoutFacade.showSidebar(Sidebar.Settings);

components/automate-ui/src/app/components/sidebar/sidebar.component.html
tarablack01: @bcmdarroch This was a back and forth refactor between static and observable values, which is added back.

tarablack01:
@susanev So I had troubles testing the credentials. I would successfully fo the curl command but the userPerms paths would still be enabled and the <app-auth> would still return true for user authorization. Can you test again?

susanev:
@tarablack01 still seems to be a bit of an issue, you can sign in as `viewer` with the password `chefautomate` and youll see the identity heading when you shouldnt.

viewer

msorens:
I am seeing a new issue:
Every time you select Settings or select Compliance on the top nav bar, or any entry under those two tabs in the side nav bar, a browser error occurs.
The message is just slightly different between Firefox and Chrome but the lines of code are the same.

### CHROME
```
SidebarComponent.html:22 ERROR TypeError: Cannot read property 'endpoint' of undefined
    at authorized.ts:79
    at Array.filter (<anonymous>)
    at AuthorizedChecker.toUserPermsPayload (authorized.ts:79)
    at AuthorizedChecker.setPermissions (authorized.ts:56)
    at AuthorizedComponent.ngOnInit (authorized.component.ts:81)
    at checkAndUpdateDirectiveInline (core.js:31910)
    at checkAndUpdateNodeInline (core.js:44367)
    at checkAndUpdateNode (core.js:44306)
    at debugCheckAndUpdateNode (core.js:45328)
    at debugCheckDirectivesFn (core.js:45271)
```
### FIREFOX
```
ERROR TypeError: "check is undefined"
    paths authorized.ts:79
    toUserPermsPayload authorized.ts:79
    setPermissions authorized.ts:56
    ngOnInit authorized.component.ts:81
    Angular 5
    View_SidebarComponent_5 SidebarComponent.html:22
    Angular 38
    RxJS 5
    Angular 8
SidebarComponent.html:22:16
    View_SidebarComponent_5 SidebarComponent.html:22
    Angular 15
    RxJS 5
    Angular 8
```
Occurs for either IAM v1 or v2.

components/automate-ui/src/app/entities/layout/layout.model.ts
```diff
@@ -1,21 +1,38 @@
...
 //   items: [],
-//   visible: true
+//   visible: Observable
 // }
 
 export interface MenuItemGroup {
   name: string;
   items: MenuItem[];
-  visible: any; // boolean or observable
+  visible?: any;
```
msorens: Hmm... looks like you are still using this as a boolean **or** an Observable, but that makes it very difficult to maintain. Should probably be of type `Observable<boolean>`, just like the `MenuItem` property.

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
```diff
@@ -1,294 +1,321 @@
...
 })
-export class LayoutSidebarService {
+export class LayoutSidebarService implements OnInit, OnDestroy {
     public applicationsFeatureFlagOn: boolean;
     public chefInfraServerViewsFeatureFlagOn: boolean;
-    public isIAMv2$: Observable<boolean>;
-    public ServiceNowfeatureFlagOn: boolean;
-    private workflowEnabled: boolean;
-
+    public isIAMv2: Observable<boolean>;
```
msorens: Please stay with the convention of always naming Observables with a trailing `$`. That convention is used throughout our code base and makes it easier to digest the code. (That's not just a Chef convention, but a general rxjs convention.)

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
```diff
@@ -1,294 +1,321 @@
...
+    public ServiceNowFeatureFlagOn: boolean;
+    private activeSidebar: string;
+    private workflowEnabled: Observable<boolean>;
```
msorens: So that would apply here, too. Thanks!

components/automate-ui/src/app/entities/layout/layout.model.ts
tarablack01: @msorens thanks for the catch, it should always be an observable! 👍🏻

components/automate-ui/src/app/entities/layout/layout.model.ts
```diff
@@ -24,17 +41,18 @@ export interface MenuItem {
   iconRotation?: number;
   route: string;
   authorized?: Authorized;
-  visible: boolean;
+  visible?: Observable<any>;
```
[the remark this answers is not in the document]
tarablack01: @msorens I guess I wasn't really sure what the flags would return but I think it's good to just force a boolean value 👍🏻

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
tarablack01: @msorens Yes, you're right! I had switched between an extracted value back to observable. I will correct it! Thanks 😊

components/automate-ui/src/app/components/sidebar/sidebar.component.html
```diff
@@ -6,34 +6,23 @@
...
       <span *ngFor="let menuGroup of menuGroups$ | async">
         <div
-          *ngIf="isVisible(menuGroup.visible) | async"
+          *ngIf="(menuGroup.visible | async) || hasAuthroizedMenuItems(menuGroup)"
           class="menu-item-groups"
         >
           <div class="group">{{ menuGroup.name }}</div>
           <span *ngFor="let menuItem of menuGroup.items">
-            <span *ngIf="(isVisible(menuItem.visible) | async) && !menuItem.authorized">
+            <span *ngIf="!menuItem.authorized && (menuItem.visible | async)">
```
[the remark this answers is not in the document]
tarablack01: @msorens I actually loop through the sidebar object and put the static values into observables, and add observables to empty values, so they're all consistent.

susanev:
signed in as `viewer`, settings is now displaying correctly.. but, the left nav for dashboards, applications, and infrastructure is not being shown at all when it should be.

also, noticed a new bug. scan jobs should be shown in the left nav of compliance for a viewer but its not being shown.

tarablack01:
@susan check out the viewer user again :)

susanev:
@tarablack01 its working great now, last issue to figure out is why scan jobs isnt showing up on the compliance left nav

components/automate-ui/src/app/components/sidebar/sidebar.component.ts
```diff
@@ -15,12 +15,36 @@ export class SidebarComponent {
...
+    });
+  }
+
+  public hasVisibleMenuItems(menuItemGroup: any): void {
+    menuItemGroup.hasAuthroizedMenuItems =
+      this.hasAuthroizedMenuItems(menuItemGroup) ||
+      this.hasNoAuthroizedMenuItems(menuItemGroup);
+  }
+
+  public hasAuthroizedMenuItems(menuItemGroup: any): boolean {
```
bcmdarroch:
```suggestion
  public hasAuthorizedMenuItems(menuItemGroup: any): boolean {
```

components/automate-ui/src/app/components/sidebar/sidebar.component.ts
```diff
@@ -15,12 +15,36 @@ export class SidebarComponent {
...
+    return menuItemGroup.items.filter(menuItem =>
+        menuItem.authorized && menuItem.authorized.isAuthorized).length > 0;
+  }
+
+  public hasNoAuthroizedMenuItems(menuItemGroup: any): boolean {
```
bcmdarroch:
```suggestion
  public hasNoAuthorizedMenuItems(menuItemGroup: any): boolean {
```

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
```diff
@@ -1,294 +1,321 @@
...
-        this.chefInfraServerViewsFeatureFlagOn = this.featureFlagsService.getFeatureStatus('chefInfraServerViews');
         this.applicationsFeatureFlagOn = this.featureFlagsService.getFeatureStatus('applications');
-        this.ServiceNowfeatureFlagOn = this.featureFlagsService.getFeatureStatus('servicenow_cmdb');
-        this.clientRunsStore.select(clientRunsWorkflowEnabled).subscribe(
-            (workflowEnabled) => this.workflowEnabled = workflowEnabled
-        );
+        this.ServiceNowFeatureFlagOn = this.featureFlagsService.getFeatureStatus('servicenow_cmdb');
+        this.isIAMv2$ = this.store.select(isIAMv2);
+        this.workflowEnabled$ = this.clientRunsStore.select(clientRunsWorkflowEnabled);
+        // this.store.select(isIAMv2).pipe(
```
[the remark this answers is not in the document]
bcmdarroch: That being said, v1 is going away in a few weeks anyway so I think it's ok if we don't get that working

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
```diff
@@ -1,294 +1,321 @@
...
+    ngOnInit() {
+      this.updateSidebars();
+    }
+
+    ngOnDestroy() {
+      this.isDestroyed.next(true);
+      this.isDestroyed.complete();
+    }
+
+    private updateVisibleValues(sidebar): Sidebars {
```
bcmdarroch: what type is `sidebar`?

components/automate-ui/src/app/entities/layout/layout.model.ts
```diff
@@ -1,21 +1,38 @@
+import { Observable } from 'rxjs';
+
+// Sidebar {
```
bcmdarroch: still needed?

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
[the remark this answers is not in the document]
susanev: +1 on this is an issue, but we dont have to fix it. because very soon there will be no v1.

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
tarablack01: @bcmdarroch It's type is Sidebars but it's not the correct model when it gets passed into the function as the function is actually manipulating the object to match the Sidebars model and returning it back.

components/automate-ui/src/app/entities/layout/layout.model.ts
tarablack01: @bcmdarroch still needed but noticed I needed to rename it and fix a few other comment examples 👍🏻

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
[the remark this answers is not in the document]
tarablack01: @bcmdarroch removed commented code 👍🏻

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
```diff
@@ -1,294 +1,321 @@
-import { Injectable } from '@angular/core';
+import { Injectable, OnInit, OnDestroy } from '@angular/core';
 import { Store } from '@ngrx/store';
-import { Observable } from 'rxjs';
+import { Subject, Observable, BehaviorSubject } from 'rxjs';
+import { forEach } from 'lodash';
+// import { takeUntil } from 'rxjs/operators';
```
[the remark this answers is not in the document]
tarablack01: @bcmdarroch not needed, removed 👍🏻

components/automate-ui/src/app/components/sidebar/sidebar.component.ts
tarablack01: @bcmdarroch nice catch! Thanks :)

components/automate-ui/src/app/components/sidebar/sidebar.component.ts
tarablack01: @bcmdarroch nice catch! Thanks :)

components/automate-ui/src/app/components/sidebar/sidebar.component.ts
```diff
@@ -15,12 +15,36 @@ export class SidebarComponent {
...
+        menuGroups.forEach(menuItemGroup => this.hasVisibleMenuItems(menuItemGroup));
+      }
+    });
+  }
+
+  public hasVisibleMenuItems(menuItemGroup: any): void {
+    menuItemGroup.hasAuthorizedMenuItems =
+      this.hasAuthorizedMenuItems(menuItemGroup) ||
+      this.hasNoAuthorizedMenuItems(menuItemGroup);
+  }
```
msorens: The logic is perfectly sound here, but the naming confused me for a bit. After several passes I came up with this to improve clarity; what do you think?

- hasVisibleMenuItems => setGroupVisibility
- hasAuthorizedMenuItems => hasVisibleMenuItems
- hasNoAuthorizedMenuItems => hasMenuItemsNotRequiringAuthorization

resulting in:
```
  public setGroupVisibility(menuItemGroup: MenuItemGroup): void {
    menuItemGroup.hasVisibleMenuItems =
      this.hasAuthorizedMenuItems(menuItemGroup) ||
      this.hasMenuItemsNotRequiringAuthorization(menuItemGroup);
  }
```
(Also added type instead of `any`, but then you will need to update the definition of MenuItemGroup to include hasVisibleMenuItems.)

Happy to chat about these suggested changes as this is of key importance, I believe.

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
```diff
@@ -1,294 +1,304 @@
...
+        this.ServiceNowFeatureFlagOn = this.featureFlagsService.getFeatureStatus('servicenow_cmdb');
+        this.isIAMv2$ = this.store.select(isIAMv2);
+        this.workflowEnabled$ = this.clientRunsStore.select(clientRunsWorkflowEnabled);
     }
 
-    public getDashboardsSidebar(): MenuItemGroup[] {
-        return [{
-            name: 'Dashboards',
+    populateSidebar() {
+      const sidebars = {
```
msorens: It is important to strongly type this to catches any typos, invalid assignments, or missing properties. I realize you will have to add a few values here and there to satisfy it, but I think it is worth it. 
```suggestion
      const sidebars: Sidebars = {
```
For instance, as soon as I tried, it revealed that L85 (`visible$: this.applicationsFeatureFlagOn`) has the wrong type.

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
```diff
@@ -1,294 +1,304 @@
...
+      }
+    }
+
+    private setMenuGroupVisibility(menuGroup: any): any {
+      let anyMenuItemsVisible = false;
+      forEach(menuGroup.items, (menuItem) => {
+        if (menuItem.visible$ === undefined) {
+          anyMenuItemsVisible = true;
+        } else if (menuItem.visible$.subscribe) {
+          menuItem.visible$.subscribe((value) => {
```
msorens: A subscription somewhere other than `ngOnInit` is a red flag to me, so I added a ``console.log(`adding subscription for ${menuItem.name}`);`` inside this subscription, then navigated around the app in the browser. The result: a continuously growing list of subscriptions, so this needs to be fixed:
```
layout-sidebar.service.ts:304 adding subscription for Reports
layout-sidebar.service.ts:304 adding subscription for Scan Jobs
layout-sidebar.service.ts:304 adding subscription for Profiles
layout-sidebar.service.ts:304 adding subscription for Notifications
layout-sidebar.service.ts:304 adding subscription for Data Feeds
layout-sidebar.service.ts:304 adding subscription for Node Integrations
layout-sidebar.service.ts:304 adding subscription for Node Credentials
layout-sidebar.service.ts:304 adding subscription for Node Lifecycle
layout-sidebar.service.ts:304 adding subscription for Users
. . .
```

components/automate-ui/src/app/components/authorized/authorized.component.ts
```diff
@@ -68,6 +71,7 @@ export class AuthorizedComponent implements OnInit, OnDestroy {
         const oldVisible = this.visible;
         const newVisible = this.authorizedChecker.evalPerms(perms);
         this.visible = this.not === undefined ? newVisible : !newVisible;
+        this.isAuthorized.emit(this.visible);
```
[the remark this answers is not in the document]
tarablack01: @msorens 😊

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
tarablack01: @msorens Updated change request 😊

components/automate-ui/src/app/components/sidebar/sidebar.component.ts
tarablack01: @msorens Updated change request 😊

components/automate-ui/src/app/components/sidebar/sidebar.component.ts
```diff
@@ -15,12 +15,36 @@ export class SidebarComponent {
...
+
+  public hasVisibleMenuItems(menuItemGroup: any): void {
+    menuItemGroup.hasAuthroizedMenuItems =
+      this.hasAuthroizedMenuItems(menuItemGroup) ||
+      this.hasNoAuthroizedMenuItems(menuItemGroup);
+  }
+
+  public hasAuthroizedMenuItems(menuItemGroup: any): boolean {
+    return menuItemGroup.items.filter(menuItem =>
+        menuItem.authorized && menuItem.authorized.isAuthorized).length > 0;
```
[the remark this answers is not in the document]
tarablack01: @msorens Updated change request 😊

components/automate-ui/src/app/components/sidebar/sidebar.component.ts
```diff
@@ -15,12 +15,36 @@ export class SidebarComponent {
...
+  }
+
+  public hasNoAuthroizedMenuItems(menuItemGroup: any): boolean {
+    return menuItemGroup.items.filter(menuItem =>
+        menuItem.authorized === undefined).length > 0;
```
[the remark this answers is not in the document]
tarablack01: @msorens Updated change request 😊

components/automate-ui/src/app/components/sidebar/sidebar.component.ts
```diff
@@ -15,12 +15,36 @@ export class SidebarComponent {
...
-    this.menuGroups$ = layoutFacade.menuGroups$;
+    this.menuGroups$ = layoutFacade.sidebar$;
+    this.updateMenuGroupVisibility();
   }
 
-  public isVisible(item: any) {
-    return item.subscribe
-      ? item
-      : new BehaviorSubject(item);
+  public updateMenuGroupVisibility(): void {
```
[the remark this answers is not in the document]
tarablack01: @msorens Updated change request 😊

components/automate-ui/src/app/components/sidebar/sidebar.component.ts
```diff
@@ -15,12 +15,36 @@ export class SidebarComponent {
...
+    this.menuGroups$.subscribe((menuGroups: MenuItemGroup[]) => {
+      if (menuGroups) {
+        menuGroups.forEach(menuItemGroup => this.hasVisibleMenuItems(menuItemGroup));
+      }
```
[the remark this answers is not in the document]
tarablack01: @msorens it will not always be available that's why i had to add it, it was breaking because the initial value is undefined.

components/automate-ui/src/app/components/sidebar/sidebar.component.ts
```diff
@@ -15,12 +15,36 @@ export class SidebarComponent {
...
+    this.menuGroups$.subscribe((menuGroups: MenuItemGroup[]) => {
+      if (menuGroups) {
+        menuGroups.forEach(menuItemGroup => this.setGroupVisibility(menuItemGroup));
+      }
+    });
+  }
+
+  private setGroupVisibility(menuItemGroup: MenuItemGroup): void {
+    menuItemGroup.hasVisibleMenuItems =
+      this.hasVisibleMenuItems(menuItemGroup) ||
```
msorens: Don't mean to belabor the point. 😁  This line probably should not have been changed: it is clearer if this reads 
```suggestion
      this.hasAuthorizedMenuItems(menuItemGroup) ||
```
Thus, we are then saying the `hasVisibleMenuItems` property (L41) is determined by whether there are authorized items (L42) or there are items that always show (L43).

tarablack01: @msorens updated change :)

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
tarablack01: @msorens Updated code

PR #2470, opened 2019-12-10
deployment: always restart service on reconfigure

stevendanna:
The previous code tried to determine when a restart was necessary to
provide online reconfiguration. Unfortunately, this code wasn't kept
up-to-date, and is less necessary now that we have the `debug` server
where we can put functionality such as `set-log-level`.

We currently believe that every configuration item in the "System"
configuration that is actually considered in the converger will render
a configuration file, so the self-hup code was removed.

We return to idle after 300 seconds just in case hab is in trouble.


susanev:
@stevendanna can this be merged?

stevendanna:
This was ready to merge at the time it was approved but we held off before "winter break", but it has now been a long time.  If there is interest in this I would like to reserve some time this week to re-test it against master.

kalroy:
Although there is not enough information how to test the changes, I tested by patching a configuration and it worked fine.

PR #2473, opened 2019-12-10
Unset PIDFILE in notifications-service

jaym:
Modern versions of systemd seem to export this as an environment
variable. This confuses distillery which thinks it is to write to that
file.

We should probably do this for all the services


chefsalim:
Addresses https://github.com/chef/automate/issues/2474

PR #2476, opened 2019-12-10
Try to avoid network access in test setup for `go install`

danielsdeleo:
...it doesn't work well with proxy tests


danielsdeleo:
More complete fix was included in #2477

PR #2479, opened 2019-12-10
Encourage people to use the CLI to uninstall

jaym:
The CLI does all the things

### DIFF
--- a/components/automate-chef-io/content/docs/troubleshooting.md
+++ b/components/automate-chef-io/content/docs/troubleshooting.md
@@ -74,15 +74,6 @@ With the `chef-automate` CLI:
 chef-automate uninstall
 ```
 
-or
-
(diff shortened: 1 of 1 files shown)


components/automate-chef-io/content/docs/troubleshooting.md
```diff
@@ -68,6 +68,14 @@ The following procedure will remove Chef Automate from your system,
 including all data. If you wish to preserve the data, make a backup
 before uninstalling.
 
+With the `chef-automate` CLI:
+
+```bash
+chef-automate uninstall
+```
+
+or
```
yzl: You don't have to do this, but I would probably comment out this alternative if we'd rather people not do it but we want it available as an escape hatch to point someone to.

PR #2315, opened 2019-11-20
Documentation for Automate-deployed on-prem Builder.

yzl:
### :nut_and_bolt: Description: What code changed, and why?
Document how to deploy on-prem Builder with Automate.

TODO:
- [x] the docs don't build locally due to relref issues with the IAM v2 docs.
- [x] cover package population
- [x] update existing oauth builder page

### :chains: Related Resources
#1795 

### :white_check_mark: Checklist

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### :camera: Screenshots, if applicable

PR #2481, opened 2019-12-11
Add configuration for builder-api logging

### LINKED ISSUE #2480: chef/automate-builder-api logging should be set to error by default
Currently after installing Builder via chef automate, and looking at the logs, it looks like the log level is set to DEBUG for automate-builder-api.

We should default this to the 'error' setting.

jaym:
The log level has the following mapping from automate to
rust logger:
debug => debug
default, info, warn, warning, panic, fatal => error

Configurable through global config:
```toml
[global.v1.log]
level = "debug"
```

or

```toml
[builder_api.v1.sys.log]
level = "debug"
scoped_levels = ["tokio_core=error"]
```
Fixes #2480


jaym:
cc @chefsalim

yzl:
@jaym I'll add something to #2315 about this mapping.

PR #2483, opened 2019-12-11
add backupRunning state to converger 

stevendanna:
The new backupRunning state represents the fact that we do not want to
run most operations while the backup is running. The converger enters
this state when then backup code runs StartBackup.  The motivation for this
change is recently observed test failures caused by attempts to start a backup
when a previously configuration change was still waiting to be applied.

Right now, the backup code remains outside of the compiler, so we've
also added a BackupComplete() call to signal when we should leave this
state.

Attempts to StartBackup() when we are waiting for a restart or a
reconfigure will fail. The client then retries these failures.

### DIFF
--- a/components/automate-deployment/pkg/backup/runner.go
+++ b/components/automate-deployment/pkg/backup/runner.go
@@ -10,14 +10,12 @@ import (
 	"sync"
 	"time"
 
-	"github.com/pkg/errors"
-
-	"github.com/chef/automate/lib/grpc/secureconn"
-
 	"github.com/golang/protobuf/ptypes"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
 	api "github.com/chef/automate/api/interservice/deployment"
+	"github.com/chef/automate/components/automate-deployment/pkg/converge"
 	"github.com/chef/automate/components/automate-deployment/pkg/deployment"
 	"github.com/chef/automate/components/automate-deployment/pkg/events"
 	"github.com/chef/automate/components/automate-deployment/pkg/manifest"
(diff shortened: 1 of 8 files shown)


stevendanna:
Built on https://github.com/chef/automate/pull/2470

components/automate-deployment/pkg/backup/runner.go
```diff
@@ -892,6 +893,30 @@ func (r *Runner) startBackupOperations(ctx context.Context) {
 	ctx, cancel := context.WithDeadline(context.Background(), deadline)
 	defer cancel()
 
+	if r.converger != nil {
```
jaym: should converger ever be nil?

components/automate-deployment/pkg/backup/runner.go
```diff
@@ -892,6 +893,30 @@ func (r *Runner) startBackupOperations(ctx context.Context) {
...
+		defer func() {
```
jaym: [nit] given how long this function already is, maybe pull this out into a function

components/automate-deployment/pkg/converge/converger.go
```diff
@@ -91,15 +87,6 @@ func WithCompiler(compiler Compiler) ConvergerOpt {
 	}
 }
 
-// WithSelfHuper sets the compiler to use when fulfilling a converge request.
-// By default, we will use the PhaseOrderedCompiler. Specifying a compiler
-// is useful in unit testing
-func WithSelfHuper(s SelfHuper) ConvergerOpt {
```
jaym: :)

components/automate-deployment/pkg/server/backup.go
```diff
@@ -39,9 +40,30 @@ func (s *server) CreateBackup(ctx context.Context, req *api.CreateBackupRequest)
...
+	}
+
+	err = s.converger.StartBackup(t)
+	if err != nil {
+		s.deployment.Unlock()
+		logrus.WithError(err).Error("Failed to start backup")
+		return nil, err
+	}
+
+	err = <-t.C
```
jaym: I'm not sure I'm following what's happening here. Why is a message expected on t.C?

jaym: nvm, i think i understand.. We call sendMessage, which doesn't wait for a response.
If there aren't any use cases for not waiting for the response, maybe the function should wait.
It would also get rid of the need for creating a task.

Also, `t.Wait()` might be clearer

PR #2485, opened 2019-12-11
Code review: refactor the workflow and tasks parameters

satellite15:

The DataFeedWorkflowParams have been refactored. These are no longer shared with the poll and aggregate tasks, these now have their own params 

DataFeedPollTaskParams
DataFeedAggregateTaskParams

I also moved the scheduling update into the workflow as it updates the workflow params which are no longer passed to the poll task.

Also removed the tasksComplete from the payload and it is now a local variable.

- [ ] Tests added/updated?
- [ ] Docs added/updated?


satellite15:
@jaym these are the updates I raised issues for on the previous PR.

PR #2490, opened 2019-12-11
[gateway] bind unknown backend gRPC clients to an unimplemented backend

ryancragun:
As we move to a world where Chef Automate becomes a platform and the
products deployed on any given instance will vary wildly, we need the
gateway to be able to initialize with several different combinations of
binds, depending on the services that are required by a given product.

Previously, the gateway had a hard-coded list of target addresses and
ports which it would use to dial in cases where target information for a
service was not present in the configuration. If we were to start up the
gateway with only the binds required by a small product it would still
attempt to dial all gRPC clients that it might ever need, regardless of if
the client was ever going to exist for the deployed product. This leads
to many failures and noisy logs.

This change introduces a few new concepts:

* Instead of always trying to dial all services from a known "best
  guess" list, we instead dial configured targets and for each service
  that is missing in the configuration, we'll dial the "null backend".
  The null backend is an aggregate Chef Automate backend that implements
  all client RPCS and returns "unimplemented" errors. This will allow
  the gateway to cleanly initialize regardless of the given binds and
  will return clean errors if an RPC is requested and the required
  backend is not available.

* The gateway now handles signals and attempts to cleanly shut down when
  terminated via SIGUSR1, SIGTERM, and SIGINT. It also now responds to
  SIGHUP and will reload the configuration without needing to restart the
  gateway. The initial reconfigure implementation almost always restarts each
  sub-system, but this can be more fine tuned if we deem useful.

This change also:

* Add optional binds for all services which the gateway might attempt to
  dial.

* Replaces functional options for `NewServer()` and replaces by embedding
  Config into the Server instance. In most cases it's much less work to
  reference the embedded Config than to use functional options to set
  equivalent fields on the Server struct.

* Decomposes `Start` into several smaller functions, allowing better
  control and re-usability when re-configuring.


components/automate-gateway/gateway/clients.go
```diff
@@ -480,21 +456,98 @@ func (c *clientsFactory) connectionByName(name string) (*grpc.ClientConn, error)
...
+			opts = append(opts,
+				grpc.WithUnaryInterceptor(grpc_prometheus.UnaryClientInterceptor),
+				grpc.WithStreamInterceptor(grpc_prometheus.StreamClientInterceptor))
+		}
+
+		if endpoint.Target == c.socketPath() {
+			opts = append(opts, grpc.WithBlock(), grpc.WithTimeout(1*time.Second))
+		}
+
+		if endpoint.Secure {
```
stevendanna: This is existing code so we can flag this for potential follow-up:

- I think we should probably make the _default_ "Secure" with "Insecure" as optional.
- We seem to only apply `tracing.GlobalClientInterceptor` if Insecure is set, seems like in that case we might consider just removing it.

components/automate-gateway/gateway/server.go
```diff
@@ -325,42 +125,314 @@ func (s *Server) NewGRPCServer() (*grpc.Server, error) {
...
+	}()
+
+	return nil
+}
+
+func (s *Server) stopNullBackendServer() {
+	s.logger.Info("stopping null backend server")
+
+	s.nullBackendServer.GracefulStop()
+	_ = os.Remove(s.Config.GrpcClients.NullBackendSock)
```
jaym: what happens if this file is not removed

components/automate-gateway/gateway/server.go
```diff
@@ -325,42 +125,314 @@ func (s *Server) NewGRPCServer() (*grpc.Server, error) {
...
+
+	err := s.stopHTTPServer()
+
+	s.stopGRPCServer()
+	s.stopNullBackendServer()
+
+	return err
+}
+
+func (s *Server) reconfigure() error {
```
jaym: I might be missing something obvious, but why doesn't this end up leaking memory. Does everything we use end up getting garbage collected if its no longer accessible? I feel like the notifications client was one of the things that ended up leaking a bunch of memory when it was accidentally getting recreated a lot

components/automate-gateway/gateway/server.go
ryancragun: Nothing really, because the hab run hook will delete it during start-up if the old socket gets left around. Here we're just being good stewards and cleaning up after ourselves.

ryancragun:
> I think we should remove reconfigure entirely, but I think this is mergable once the automate requirement for builder is put back in

I demoed the current implementation to Susan and Natalie and think it's okay to go ahead with this implementation and to allow deploying only builder with "just enough" automate to handle users with the UI. We're putting a card in the UI backlog to disable the parts of the UI that aren't required in this mode but they think that shouldn't hold this up. I also updated the docs with a full explanation.

components/automate-chef-io/content/docs/on-prem-builder.md
```diff
@@ -93,6 +93,70 @@ products on the command line. For example:
 
 Accept the license with `y`.
 
+To deploy Chef Habitat Builder with the Chef Automate UI and Auth services only, specify the `builder`
```
yzl: [suggestion] It'd be good to let the user know up-front that installing this way is an option. What comes to my mind is opening the "Deploy Chef Automate ..." section with something like "You can deploy Chef Habitat Builder either with a full Chef Automate installation or with the Chef Automate Auth stack only," then giving each type of deploy its own section header.

PR #2499, opened 2019-12-12
backup: Remove async operations

stevendanna:
Currently, we do not use any async operations so these should be safe
to remove. This is one in a number of steps we want to take to reduce
the scope of the backup/restore code.

### DIFF
--- a/components/automate-deployment/pkg/backup/executor.go
+++ b/components/automate-deployment/pkg/backup/executor.go
@@ -8,11 +8,9 @@ import (
 	api "github.com/chef/automate/api/interservice/deployment"
 )
 
-// Executor executes synchronous and asynchronous backup operations. It
-// listens for the operation events, annotates them, and publishes them to the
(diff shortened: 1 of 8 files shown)


components/automate-deployment/pkg/backup/executor.go
```diff
@@ -27,16 +25,15 @@ type Executor struct {
 	// an operation failure.
 	cancel func()
 
-	// Operation progress calculators. These are used to aggregate the overall
-	// progress of all operation that are to be executed.
-	syncProgress  *ProgressCalculator
-	asyncProgress *ProgressCalculator
+	// Operation progress calculator. Thit is used to aggregate
```
yzl: [nit] Thit -> this

stevendanna:
```suggestion
	// Operation progress calculator. This is used to aggregate
```

Issue #2149, opened 2019-11-07
Selecting a new row in the service group table causes the page to lose its scroll position


apriofrost:
## Describe the bug
In the service group page, scroll down the service group table, and select a new row there.
The table refreshes and goes back to the top of the table and loses its scroll position.

## Expected behavior
Expect the scroll position to remain the same whether a new row is selected or not.
Expect the table to not refresh when a new row is selected.
Expect another scroll bar to appear right near the table, in addition to the scroll bar for the side panel.

susanev:
is it possible to do that selection without refreshing the main table?

apriofrost:
@susanev that would be ideal. Only the content in the right-side panel needs refreshing.

Issue #1884, opened 2019-10-16
Refactor Service Groups page URL to be more consistent and meaningful


apriofrost:
## Two Problems
Firstly, when users land on the Applications tab, the URL looks like such:
![image](https://user-images.githubusercontent.com/9401459/66958153-7093dd80-f01c-11e9-8015-215addf5cc23.png)
It doesn't include the page name `service groups` to it, which could cause a problem in the future when we add more pages to the Applications tab that share the same URL parameters for status filters.

Secondly, after users select a service group in the table, then apply a status filter in the sidebar, a `sgStatus` is attached after the service group ID parameter. This is confusing as the status filter in the sidebar is filtering on individual service instance status. In addition, if we add more pages, such as deployment, to the Applications tab, we would like the sidebar applied filters parameter to be generic and not tied to specific content on the page.

## Solution

1. Add a page name in the URL after the `/applications`. For the service groups page, it will become `/applications/service-groups`. 
(Note: This is a pattern we would like to form for other pages as well: `/{domain-name}/{page-name}`)

_(the following solution is still under discussion, it should match with the API change in #1881 )_

2. Form a pattern for the applied filters for the main table and the sidebar across pages. For the applied filters for both the main table and the sidebar, attach the `status=value` to the URL. For the service groups page, when users filter on the "critical" service groups, and "warning" services in a specific group, the URL will become:
```
/applications/service-groups?status=critical&sgid={insert-id-here}&status=warning
```

PR #2503, opened 2019-12-12
Remove services view state from URL and update service groups URL

danielsdeleo:

Updates the URLs in the apps tab as specified in #1884

Removes services view state from the URL; instead it is managed in the application memory. This makes sense architecturally because a particular service group can come and go from a set of service group filters over time. For example, if the critical status filter is selected, and the services within a service group all change status to `OK`, then the URL `/applications/service-groups?status=critical&sgId=<the id>` no longer refers to the same content in the same way.

This change means that when users share URLs they won't have the same service group selection state, though if a user wishes to construct a URL that shows just a particular service group, they can still do so with the filtering functionality we have.

Making this change means that changes to service group view state will no longer trigger events from observables that watch the query parameters, which solves a bug where the service group list was being re-rendered for changes to the services view state and would lose the user's scroll position after flickering the content (#2149)

Addresses #1884 and #2149

*To Build:*
* build components/automate-ui
* start_all_services
* create the sample data with `applications_populate_database`

*Verify No Regressions:*
* Start at the event feed: `https://a2-dev.test/dashboards/event-feed`
* Click on "Applications" in the nav bar, the URL should not be wrong
* Click on "Service Groups" on the left nav, the URL should not be wrong (nothing should happen since it should be a link to the page you are on)
* Service groups state should all save to the URL:
  * use the 'OK' health filter
  * Sort by env or app or any non-default field
  * go to page 2
  * refresh and you should have the same service groups list on the left side list
* search bar filters filter the services in the sidebar
  * search for `Site` = `aaa`, should show the "multisite.default" service group, only the two services with site = aaa should show in the sidebar.
* services status filters in the sidebar work as expected
 * Select a filter with more than zero services in the sidebar, it should work; select a filter with zero services and you should see the "None of the services" message
* pagination in the sidebar works:
  * select the `redis.default` service group
  * filter for critical status
  * page 2 of services should have different IDs than those on page 1 and the "OK" services should not be shown.
* services status filter and pagination state is reset when a new service group is selected
  * with the critical filter active and viewing page 2 of `redis.default`, select a service group that will have multiple pages, like `mobile-api-frontend.default`. The health filter should reset to "total" and you should be on page 1.

While doing the above, you should only see the service groups list be refreshed when taking an action that affects the service groups content (service groups health filter, search bar filter, service groups sort, service groups pagination).

### DIFF
--- a/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.ts
+++ b/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.ts
@@ -63,30 +63,13 @@ export class ServicesSidebarComponent implements OnInit, OnDestroy {
   public timewizardMessage = this.serviceGroupsFacade.timewizardMessage;
 
   ngOnInit() {
-    console.log('oninit for a new ServicesSidebarComponent')
-
-    // FIXME: need to reset status filter and pagination state when the service group id changes
-    // FIXME: pagination controls borked, probably need to fix app-page-picker 
-
-    // this.services$.pipe(takeUntil(this.isDestroyed))
-    //   .subscribe( () => {
-    //   });
-
     this.svcHealthSummary$.pipe(takeUntil(this.isDestroyed))
         .subscribe((serviceGroupsHealthSummary) => {
-          // this.selectedHealth = 'total';
-          // this.currentPage = 1;
-          // this.pageSize = 25;
-
           this.serviceGroupsHealthSummary = serviceGroupsHealthSummary;
           this.totalServices = getOr(0, this.selectedHealth, this.serviceGroupsHealthSummary);
         });
 
     this.currentServicesFilters$.pipe(takeUntil(this.isDestroyed)).subscribe((servicesFilters) => {
-      console.log(`currentServicesFilters$ servicesFilters:`)
-      console.log(servicesFilters)
-
-
       this.selectedHealth = getOr('total', 'health', servicesFilters);
       this.currentPage    = getOr(1, 'page', servicesFilters);
       this.totalServices  = getOr(0, this.selectedHealth, this.serviceGroupsHealthSummary);
@@ -96,9 +79,6 @@ export class ServicesSidebarComponent implements OnInit, OnDestroy {
          totalServices: this.totalServices,
          statusFilter: this.selectedHealth
       });
-
-      console.log(`currentServicesFilters$ this:`)
-      console.log(this);
     });
   }
 
@@ -108,7 +88,6 @@ export class ServicesSidebarComponent implements OnInit, OnDestroy {
   }
 
   public updateHealthFilter(health: string) {
-    console.log(`updateHealthFilter to ${health}`)
     this.currentPage = 1;
     this.selectedHealth = this.serviceGroupsFacade
       .updateHealthFilter(health, 'applicationsStatusFilter');
@@ -124,7 +103,6 @@ export class ServicesSidebarComponent implements OnInit, OnDestroy {
 
   refresh() {
     if (this.serviceGroupsId) {
-      console.log(`doing dat refresh>>>`)
 
       const paramsForDispatch = {
           service_group_id: this.serviceGroupsId,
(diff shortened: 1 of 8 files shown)


components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.ts
```diff
@@ -85,15 +91,30 @@ export class ServicesSidebarComponent implements OnInit, OnDestroy {
...
       .updatePageNumber(pageNumber, this.totalServices, this.currentPage, 'applicationsPageChange');
-    this.serviceGroupsFacade
-      .updateServicesFilters(this.selectedHealth, this.currentPage);
+    this.refresh();
   }
+
+  refresh() {
+    if (this.serviceGroupsId) {
+
+      const paramsForDispatch = {
```
msorens: Strong-typing makes life easier. 😁 
```suggestion
      const paramsForDispatch: GroupServicesFilters = {
```

components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
```diff
@@ -211,40 +210,34 @@ export class ServiceGroupsComponent implements OnInit, OnDestroy {
       withLatestFrom(this.route.queryParamMap),
       takeUntil(this.isDestroyed)
     )
-    .subscribe(([serviceGroups, queryParams]) => {
+    .subscribe(([serviceGroups]) => {
```
msorens: More clear to be explicit:
```suggestion
    .subscribe(([serviceGroups, _]) => { 
```
(Otherwise someone might come along--like I almost did--and say "hey, you could just do `.subscribe(serviceGroups => {` " which at first glance seems equivalent but it is not.)

components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
```diff
@@ -211,40 +210,34 @@ export class ServiceGroupsComponent implements OnInit, OnDestroy {
...
-        const sgId = queryParams.get('sgId');
-        if ( !sgId ) {
-          this.router.navigate(
-            [],
-            { queryParams: { sgId: serviceGroups[0]['id'] }, queryParamsHandling: 'merge' }
-          );
+        // if we do not have a selected service group, pick the first sgId from
+        // the service group list
+        if ( !this.selectedServiceGroupId ) {
+          this.selectedServiceGroupId = serviceGroups[0]['id'];
```
msorens: In its existing form, you can replace `'id'` with anything and get no complaint. But if you change to this, you can only put valid field names:
```suggestion
          this.selectedServiceGroupId = serviceGroups[0].id;
```

components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
```diff
@@ -346,18 +339,13 @@ export class ServiceGroupsComponent implements OnInit, OnDestroy {
...
-    if (sgId) {
-      const servicesFilters: GroupServicesFilters = {
-        service_group_id: sgId,
-        page: parseInt(queryParams.get('sgPage'), 10) || 1,
-        pageSize: parseInt(queryParams.get('sgPageSize'), 10) || 25,
-        health: queryParams.get('sgStatus') || 'total',
+  refreshServicesSidebar() {
+    if (this.selectedServiceGroupId) {
+      this.store.dispatch(new UpdateSelectedSG({
+        service_group_id: this.selectedServiceGroupId,
```
msorens: Unlike the previous example I commented on, thi argument to `UpdateSelectedSG` is strongly typed already because it is inlined, so no changes needed.

components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
```diff
@@ -211,40 +210,34 @@ export class ServiceGroupsComponent implements OnInit, OnDestroy {
...
-        // the first service group from the list and navigate to it. this lets
-        // us maintain a service group selection regardless of other navigation
-        // as long as the service group is still in the list.
-        const selectedSGVisibleOnPage = serviceGroups.find(sg => sg.id === sgId);
+        // the first service group from the list. this lets us maintain a
+        // service group selection regardless of other navigation as long as
+        // the service group is still in the list.
+        const selectedSGVisibleOnPage = serviceGroups.find(sg => {
+          return sg.id === this.selectedServiceGroupId;
+        });
```
msorens: Could make slightly more readable/concise:
```
        const selectedSGVisibleOnPage =
          serviceGroups.find(sg => sg.id === this.selectedServiceGroupId);
 
```

danielsdeleo: Yeah, I like that better, too but the line length linter doesn't like it 🙃

Issue #2449, opened 2019-12-06
Prevent Project Delete e2e API tests


lancewf:
## User Story

Create a few cypress e2e tests to cover the three cases of project deletion. 
* Project deletion when edits are pending
* Project deletion when the project has ingestion rules
* Project deletion when there are no edits pending and the project has no ingestion rules

## Definition of Done
All three project deletion prevention cases have e2e tests. 

## Demo Script / Repro Steps
These should be in the suit of buildkite tests.

PR #2509, opened 2019-12-13
Adding project delete integration tests

lancewf:

Add integration tests for the cases preventing project delete. 

A user can not delete a project when:
* The project has applied rules
* The project has staged rules
* The project has deleted applied rules

#2449 

Look at the cypress test in buildkite

### DIFF
--- a/e2e/cypress/integration/api/iam/01_projects_api.spec.ts
+++ b/e2e/cypress/integration/api/iam/01_projects_api.spec.ts
@@ -267,12 +267,7 @@ describeIfIAMV2p1('projects API', () => {
           .forEach(({ status }) => expect(status).to.equal(editsPendingStr));
       });
 
-      cy.request({
-        headers: { 'api-token': Cypress.env('ADMIN_TOKEN') },
-        method: 'POST',
-        url: '/apis/iam/v2beta/apply-rules'
-      });
-      cy.waitUntilApplyRulesNotRunning(500);
+      cy.applyRulesAndWait(500);
 
       // confirm rules are applied
       for (const project of [avengersProject, xmenProject]) {
@@ -386,12 +381,7 @@ describeIfIAMV2p1('projects API', () => {
           .forEach(({ status }) => expect(status).to.equal(editsPendingStr));
       });
 
-      cy.request({
-        headers: { 'api-token': Cypress.env('ADMIN_TOKEN') },
-        method: 'POST',
-        url: '/apis/iam/v2beta/apply-rules'
-      });
-      cy.waitUntilApplyRulesNotRunning(100);
+      cy.applyRulesAndWait(100);
 
       cy.request({
         headers: {
@@ -467,12 +457,7 @@ describeIfIAMV2p1('projects API', () => {
           .forEach(({ status }) => expect(status).to.equal(editsPendingStr));
       });
 
-      cy.request({
-        headers: { 'api-token': Cypress.env('ADMIN_TOKEN') },
-        method: 'POST',
-        url: '/apis/iam/v2beta/apply-rules'
-      });
-      cy.waitUntilApplyRulesNotRunning(100);
+      cy.applyRulesAndWait(100);
 
       cy.request({
         headers: { 'api-token': Cypress.env('ADMIN_TOKEN') },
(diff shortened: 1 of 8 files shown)


e2e/cypress/integration/api/iam/01_projects_api.spec.ts
```diff
@@ -467,12 +457,7 @@ describeIfIAMV2p1('projects API', () => {
           .forEach(({ status }) => expect(status).to.equal(editsPendingStr));
       });
 
-      cy.request({
-        headers: { 'api-token': Cypress.env('ADMIN_TOKEN') },
-        method: 'POST',
-        url: '/apis/iam/v2beta/apply-rules'
-      });
-      cy.waitUntilApplyRulesNotRunning(100);
+      cy.applyRulesAndWait(100);
```
bcmdarroch: nice helper 👍

e2e/cypress/integration/api/iam/14_projects_delete_api.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+
+    // Add a rule to the project
+    cy.request({
+      headers: { 'api-token': Cypress.env('ADMIN_TOKEN') },
+      method: 'POST',
+      url: `/apis/iam/v2beta/projects/${project.id}/rules`,
+      body: rule
+    });
+
+    // Try to delete the project with a stagged rule
```
bcmdarroch: s/stagged/staged

e2e/cypress/integration/api/iam/14_projects_delete_api.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+    cy.request({
+      headers: { 'api-token': Cypress.env('ADMIN_TOKEN') },
+      method: 'DELETE',
+      url: `/apis/iam/v2beta/projects/${project.id}`,
+      failOnStatusCode: false
+    }).then((deleteResp) => {
+      expect(deleteResp.status,
+        'Did not fail deleting a project with a stagged rule').to.be.oneOf([400]);
```
bcmdarroch: s/stagged/staged

bcmdarroch: Also I think we could simplify the assertion to `expect(deleteResp.status).to.equal(400);`

e2e/cypress/integration/api/iam/14_projects_delete_api.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+    });
+
+    // Ensure the project was not deleted
+    cy.request({
+      headers: { 'api-token': Cypress.env('ADMIN_TOKEN') },
+      method: 'GET',
+      url: '/apis/iam/v2beta/projects'
+    }).then((resp: Cypress.ObjectLike) => {
+      expect(resp.body.projects && resp.body.projects.length > 0 &&
```
bcmdarroch: These read a bit easier as separate lines. Cypress has a [helpful list of all the assertions it allows](https://docs.cypress.io/guides/references/assertions.html#BDD-Assertions)

e2e/cypress/integration/api/iam/14_projects_delete_api.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+        resp.body.projects.some((p: any) => p.id === project.id)).to.be.equal(true);
+    });
+
+    // Apply the rules to make the rule applied
+    cy.applyRulesAndWait(100);
+
+    // Try to delete the project with the applied rule
+    cy.request({
```
bcmdarroch: can we write these as separate tests?

lancewf: I could not think of a way to do that and not run the apply rules any extra times.

e2e/cypress/integration/api/iam/14_projects_delete_api.spec.ts
lancewf: We can not use a lot of those assertions because the linter does not like them. For example `expect(myVar).to.exist` and `expect(true).to.be.true`. I get the error "expected an assignment or function call".

msorens: After considering the various operators, I believe what you have is optimal. 
(Here's one source that concurs: https://stackoverflow.com/a/34445397)
However, since you reuse this check several times, consider extracting it as a helper function:
```
      expect(responseContainsProject(resp.body.projects, project.id)).to.be.equal(true);
```

Issue #2284, opened 2019-11-20
escalation: data discrepancies between Automate GUI vs. exported data (client runs)


vjeffrey:
## Describe the bug
When exporting nodes from the Infrastructure tab, the results differ from what is seen in the WebUI (specifically, the total count of nodes is about 10% less when exporting)

## To Reproduce
_please note that support has been unable to reproduce this issue_
- load up a bunch of nodes in client runs
- export the data
- note any differences in node counts in the export vs gui

## Expected behavior
export and gui reflect same information

## Versions (please complete the following information):
- Automate Build Number:  20191024135531

## Additional context
https://github.com/chef/customer-bugs/issues/12
https://getchef.zendesk.com/agent/tickets/23731
https://getchef.zendesk.com/agent/tickets/23783

vjeffrey:
from @teknofire :
```
curl -k https://localhost/api/v0/cfgmgmt/nodes/export -H 'api-token: token' -d '{"output_type":"csv","filter":[],"sorting":{"field":"name","order":0}}'

curl -k https://localhost/api/v0/cfgmgmt/nodes/export -H 'api-token: token' -d '{"output_type":"csv","filter":[],"sorting":{"field":"checkin","order":0}}'

The first curl only returns 240ish nodes and the second one returns all 744 nodes.
```

phiggins:
tshirt size medium

PR #2510, opened 2019-12-14
Fix client runs export incomplete data bug

phiggins:
Attempted fix for #2284.

Sorting on the client runs export endpoint was being done case sensitively, and this made it so that the paginated could be incomplete with a large number of similarly named nodes. I believe this happened because the pagination for this endpoint works by remembering that last position in the list and searching for matching elements after that one; when lots of the nodes had similar names differing in case, this made it so finding the last element when fetching the next page was unreliable. This PR contains a fix for the incomplete data issue, but it necessarily includes a behavior change because it's unclear whether we can support a case sensitive sort with the way pagination is implemented.

Before the sorting changes in this PR, the test would reliably fail with a message like this:

```
=== RUN   TestNodeExportPaging
--- FAIL: TestNodeExportPaging (0.48s)
    node_export_test.go:771:
                Error Trace:    node_export_test.go:771
                Error:          Not equal:
                                expected: 150
                                actual  : 118
                Test:           TestNodeExportPaging
FAIL
coverage: [no statements]
FAIL    github.com/chef/automate/components/config-mgmt-service/integration_test        1.776s
FAIL
```

```
=== RUN   TestNodeExportPaging
--- FAIL: TestNodeExportPaging (0.47s)
    node_export_test.go:771:
                Error Trace:    node_export_test.go:771
                Error:          Not equal:
                                expected: 150
                                actual  : 120
                Test:           TestNodeExportPaging
FAIL
coverage: [no statements]
FAIL    github.com/chef/automate/components/config-mgmt-service/integration_test        1.987s
FAIL
```

There is a small amount of randomness to the results because random UUIDs are generated in the tests and the UUID is used as the secondary, tie-breaker sort.

### DIFF
--- a/components/config-mgmt-service/backend/client.go
+++ b/components/config-mgmt-service/backend/client.go
@@ -55,12 +55,12 @@ type Client interface {
 		string, int, string, bool) ([]InventoryNode, error)
 	// @params (ctx, start, end, filters, cursorField, cursorID, pageSize, sortField, ascending)
 	// returns (Node, error)
-	GetNodesPageByCurser(context.Context, time.Time,
+	GetNodesPageByCursor(context.Context, time.Time,
 		time.Time, map[string][]string, interface{},
 		string, int, string, bool) ([]Node, error)
 	// @params (ctx, nodeID, start, end, filters, cursorEndTime, cursorID, pageSize, ascending)
 	// returns (Runs, error)
-	GetRunsPageByCurser(context.Context, string, time.Time, time.Time, map[string][]string, time.Time,
+	GetRunsPageByCursor(context.Context, string, time.Time, time.Time, map[string][]string, time.Time,
 		string, int, bool) ([]Run, error)
 }
 
--- a/components/config-mgmt-service/backend/elastic/elastic.go
+++ b/components/config-mgmt-service/backend/elastic/elastic.go
@@ -31,7 +31,6 @@ const (
 	// Elasticsearch fields we use within this package
 	RunFieldTimestamp    = "start_time"
 	RunFieldEndTimestamp = "end_time"
-	CheckinTimestamp     = "checkin"
 	ActionFieldTimestamp = "recorded_at"
 	ActionFieldID        = "id"
 	NodeFieldID          = "entity_uuid"
--- a/components/config-mgmt-service/backend/elastic/nodes.go
+++ b/components/config-mgmt-service/backend/elastic/nodes.go
@@ -100,7 +100,7 @@ func (es Backend) GetInventoryNodes(ctx context.Context, start time.Time,
 	return nodes, nil
 }
 
-func (es Backend) GetNodesPageByCurser(ctx context.Context, start time.Time,
+func (es Backend) GetNodesPageByCursor(ctx context.Context, start time.Time,
 	end time.Time, filters map[string][]string, cursorField interface{},
 	cursorID string, pageSize int, sortField string,
 	ascending bool) ([]backend.Node, error) {
--- a/components/config-mgmt-service/backend/elastic/nodes.go
+++ b/components/config-mgmt-service/backend/elastic/nodes.go
@@ -55,7 +55,7 @@ func (es Backend) GetInventoryNodes(ctx context.Context, start time.Time,
 
 	mainQuery := newBoolQueryFromFilters(filters)
 
-	rangeQuery, ok := newRangeQueryTime(start, end, CheckinTimestamp)
+	rangeQuery, ok := newRangeQueryTime(start, end, backend.CheckIn)
 
 	if ok {
 		mainQuery = mainQuery.Must(rangeQuery)
@@ -67,7 +67,7 @@ func (es Backend) GetInventoryNodes(ctx context.Context, start time.Time,
 		Query(mainQuery).
 		Index(IndexNodeState).
 		Size(pageSize).
-		Sort(CheckinTimestamp, ascending).
+		Sort(backend.CheckIn, ascending).
 		Sort(NodeFieldID, ascending).
 		FetchSourceContext(fetchSource)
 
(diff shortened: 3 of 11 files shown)


components/config-mgmt-service/backend/elastic/elastic.go
```diff
@@ -31,7 +31,6 @@ const (
 	// Elasticsearch fields we use within this package
 	RunFieldTimestamp    = "start_time"
 	RunFieldEndTimestamp = "end_time"
-	CheckinTimestamp     = "checkin"
```
phiggins: The various codepaths used by the export endpoint were using "CheckinTimestamp and `backend.CheckIn` interchangeably, and it was confusing to me as someone unfamiliar with the code that this value could come from two places.

components/config-mgmt-service/backend/elastic/nodes.go
```diff
@@ -100,18 +100,18 @@ func (es Backend) GetInventoryNodes(ctx context.Context, start time.Time,
 	return nodes, nil
 }
 
-func (es Backend) GetNodesPageByCurser(ctx context.Context, start time.Time,
-	end time.Time, filters map[string][]string, cursorField interface{},
+func (es Backend) GetNodesPageByCursor(ctx context.Context, start time.Time,
+	end time.Time, filters map[string][]string, cursorValue interface{},
```
phiggins: nitpick: This code uses the term `field` to refer to eg a field on a struct, so using `field` to refer to the value assigned to a field on a struct was confusing to me.

components/config-mgmt-service/backend/client.go
```diff
@@ -55,12 +55,12 @@ type Client interface {
 		string, int, string, bool) ([]InventoryNode, error)
 	// @params (ctx, start, end, filters, cursorField, cursorID, pageSize, sortField, ascending)
 	// returns (Node, error)
-	GetNodesPageByCurser(context.Context, time.Time,
+	GetNodesPageByCursor(context.Context, time.Time,
```
phiggins: nitpick: "curser" is not a word.

components/config-mgmt-service/grpcserver/node_export.go
```diff
@@ -117,14 +117,10 @@ func (s *CfgMgmtServer) exportNodes(ctx context.Context, request *pRequest.NodeE
 	// even after the node no longer exists
 	nodeFilters["exists"] = []string{"true"}
 
-	if sortField == "" {
-		sortField = backend.CheckIn
-	} else {
-		sortField = params.ConvertParamToNodeRunBackend(sortField)
-	}
```
phiggins: This code was misleading because:

* `sortField` will never be blank by the time the code gets here, so the first branch is not necessary.
* `sortField` has already been passed through `ConvertParamToNodeRunBackend` in the call to `request.Sorting.GetParameters()` a few lines up.

components/config-mgmt-service/integration_test/node_export_test.go
```diff
@@ -689,3 +689,84 @@ func TestNodeExportLoopBug(t *testing.T) {
...
+	names := []string{
+		"Taco Bell",
+		"taco bell",
+		"TACO BELL",
+		"tACO bELL",
+		"TACO bell",
+	}
+
+	// page size is hardcoded to 100, so to test the pagination code we need more than that.
+	numberOfNodes := 150
```
phiggins: The number 150 was not chosen super specifically, but with 5 possible node names and 150 nodes I was able to see this test fail reliably and I couldn't always get it to fail with fewer nodes.

components/config-mgmt-service/integration_test/node_export_test.go
```diff
@@ -689,3 +689,84 @@ func TestNodeExportLoopBug(t *testing.T) {
...
+	for dec.More() {
+		var nodes []interface{}
+		err := dec.Decode(&nodes)
+		require.NoError(t, err)
+
+		actualNumberOfNodes += len(nodes)
+	}
+
+	assert.Equal(t, numberOfNodes, actualNumberOfNodes)
+}
```
phiggins: While debugging this code, I changed the hardcoded page size to 1 and every test calling `NodeExport` would fail because they only look at the first page of data. That is fairly safe to do since these tests use a lot fewer nodes than would be necessary to trigger pagination behavior, but it was still fairly confusing and now this test works differently in how it parses the result.

components/config-mgmt-service/integration_test/node_export_test.go
```diff
@@ -689,3 +689,84 @@ func TestNodeExportLoopBug(t *testing.T) {
...
+	dec := json.NewDecoder(bytes.NewReader(data))
+	for dec.More() {
+		var nodes []interface{}
+		err := dec.Decode(&nodes)
+		require.NoError(t, err)
+
+		actualNumberOfNodes += len(nodes)
+	}
+
+	assert.Equal(t, numberOfNodes, actualNumberOfNodes)
```
phiggins: This test doesn't verify that the correct data is returned, it follows the pattern of several other tests to test that the correct number of results were returned.

components/config-mgmt-service/params/parameters.go
```diff
@@ -68,39 +68,39 @@ func ConvertParamToNodeRunBackend(parameter string) string {
...
+	case "policy_group", backend.PolicyGroup:
 		return backend.PolicyGroup + ".lower"
-	case "policy_name":
+	case "policy_name", backend.PolicyName:
 		return backend.PolicyName + ".lower"
-	case "policy_revision":
+	case "policy_revision", backend.PolicyRevision:
 		return backend.PolicyRevision + ".lower"
 	default:
 		return parameter
```
phiggins: This change was necessary because the flow of data looks like this:
* The request to `NodeExport` is made with the sorting option set to `name`. 
* The parameters are processed and `name` is replaced with `node_name` in `ConvertParamToNodeRunBackend`. 
* Before calling elasticsearch, `node_name` needs to be replaced with `node_name.lower`, but this function will only do that if passed `name`.

The choices were to:

* duplicate a lot of parameter processing code (eg, setting the default when the sort parameter is not present, etc).
* Put the desired logic into another very similar function.
* Modify this function to handle already processed values.

(I went with the third option.)

phiggins:
Updated the description to hopefully describe the root problem a little better.

Issue #2511, opened 2019-12-14
duplicate node records in postgres (nodes) table


vjeffrey:
## User Story
There are a few different ways a node can get added to the nodes postgres table (this is the data that serves the `api/v0/nodes/search` api).
- a client run report is ingested, so we send information to the nodemanager about the node
- a compliance report is ingested, so we send information to the nodemanager about the node
- a user "manually" adds the node by accessing `compliance/scan-jobs/nodes/add` and entering information about some nodes
- a user creates an aws, azure, or gcp node integration by visiting `settings/node-integrations/add`

A user recently reported that when they created an aws-ec2 integration, they ended up with duplicate node records under the `nodes/search` api.  Since chef-client was running on those nodes, we already had a record of them, but failed to recognize those nodes as the same object when ingesting information.

The nodes table has two different unique constraints:
- the traditional id one, which correlates to the node uuid that is on the ingested report or created at time of object creation
- a three-field unique match on `source_id`, `source_region`, and `account_id`. those fields, when referencing a node in aws, correlate to the instance-id of the node, the region in which the node exists (e.g. us-east-1), and the id of the aws account in which it exists. for azure, this is the node id, region, and tenant id.

This specific user problem can be traced to a missing data field on the message sent from the [client run ingestion path to the nodemanager](https://github.com/chef/automate/blob/master/components/ingest-service/pipeline/publisher/node_manager_publisher.go#L81). We are only sending the instance id and region, no account id. The work to add the account id (and ensure we're storing all the correct information when ingesting the node data) is [in progress](https://github.com/chef/automate/compare/vj/try-no-dup-nodes).

But there are other cross-points where the duplicate node object problem still exists.
@kmacgugan wrote a thing about this: <fill in link later>

It's important that we address these duplicate node situations as much as possible for the one node view epic. As part of that epic, we'll be exposing all those nodes records in the ui, which will make any duplicate node object issues more apparent.

## Definition of Done
- read kyleen's doc
- figure out what node uuid work has already been done and what problems still exist
- see if there are other problems
- create issues for them so we can address them

danielsdeleo:
currently, the chef-client UUID can be set to the habitat ring member ID, but this is opt-in only.

The client.rb template is here: https://github.com/chef/chef/blob/cd444f5dfd39e4494e0a8495b20b39259dab923a/habitat/config/client.rb#L17-L19
and the default config is https://github.com/chef/chef/blob/master/habitat/default.toml#L36

[...]

PR #2512, opened 2019-12-16
adding an aws-ec2 integration should not create duplicate node records

vjeffrey:

Modified node manager code to ensure node identifier information is always up to date.
Modified ingest-service code to pass account id for the node to the node manager.

*why?*
A user recently reported that when they created an aws-ec2 integration, they ended up with duplicate node records under the nodes/search api. Since chef-client was running on those nodes, we already had a record of them, but failed to recognize those nodes as the same object when ingesting information.

https://github.com/chef/automate/issues/2511

### :+1: Definition of Done
when a user has nodes in aws running chef infra
and they want to add an aws-ec2 integration
automate correctly identifies those node objects and updates them [it *does not* create a duplicate entry for them]

- rebuild ingest, nodemanager
- create a node in ec2
- write down the instance id, region, and account id for the node
- grab a sample chef infra run report and change the ec2 information to match those details
- ingest the infra run report
- add an aws-ec2 integration for the same account you made the node in
    - _state without these changes:_ duplicate node entry in postgres for the node in question
    - _expected state with these changes:_ *no* duplicate node entry in postgres for the node in question

- [X] Tests added/updated?
- [N/A] Docs added/updated?


components/nodemanager-service/pgdb/node-managers-process-node.go
```diff
@@ -17,24 +19,24 @@ import (
 const sqlUpsertByIDRunData = `
 INSERT INTO nodes
 	(id, name, platform, platform_version, source_state, 
-		last_contact, source_region, source_account_id, last_run, projects_data, manager)
-VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11)
+		last_contact, source_id, source_region, source_account_id, last_run, projects_data, manager)
+VALUES ($1, $2, $3, $4, $5, $6, NULLIF($7,''), NULLIF($8,''), NULLIF($9,''), $10, $11, $12)
```
phiggins: I had never seen `NULLIF` before. Neat!

components/nodemanager-service/pgdb/node-managers-process-node.go
```diff
@@ -111,9 +115,19 @@ func (db *DB) ProcessIncomingNode(node *manager.NodeMetadata) error {
 		node.Uuid = id.String()
 	}
 
+	// trim the region field, b/c we want to just use "us-east-1", not more specific regions like "us-east-1a"
+	// this is to ensure that we don't create duplicate nodes
+	if match, err := regexp.MatchString(`^.*-.*-\d\D`, node.GetSourceRegion()); match {
```
phiggins: This seems like it could have some weird/unintended side effects, but I can't really think of a better way to do this. Anything like a hardcoded list would get outdated, etc. 🤔

components/ingest-service/backend/elastic/mappings/converge.go
```diff
@@ -244,6 +244,9 @@ var ConvergeHistory = Mapping{
 							},
 							"public_ipv4": {
 								"type": "ip"
+							},
```
kmacgugan: Should we make this a top level field? We could share the id with the tenant_id from azure and call it cloud_id or something for easier querying.

components/ingest-service/pipeline/publisher/node_manager_publisher.go
```diff
@@ -79,6 +79,7 @@ func gatherInfoForNode(node backend.Node) (*manager.NodeMetadata, error) {
 		LastContact:     timestamp,
 		SourceId:        node.Ec2.InstanceId,
 		SourceRegion:    node.Ec2.PlacementAvailabilityZone,
```
kmacgugan: Does node manager use placement availability zone or region? region would be like 'us-west-2' and placement availability zone is like 'us-west-2a'. Region is also available in the ohai data.

components/nodemanager-service/pgdb/node-managers-process-node.go
kmacgugan: The extra letter is because it is actually the placement availability zone. The region is also in the ohai data, and we could grab this directly and it would not have the extra letter on the end.

kmacgugan: Here is an ohai data example with region: https://github.com/chef/automate/blob/master/components/ingest-service/chef-load-data/ohai-data.json#L980

kmacgugan:
I think we should consider making some top level fields to query that are generic versions of these compound id fields. Like source_id, account_id and region_id. Then we could query without needing to know if we are looking at azure or ec2.

components/nodemanager-service/pgdb/node-managers-cloud.go
```diff
@@ -229,17 +229,24 @@ func (db *DB) AddManagerNodesToDB(nodesArr []*manager.ManagerNode, managerId str
 			}
 		}
 
+		var dbNodeID string
 		err = Transact(db, func(tx *DBTrans) error {
-			uuid, err = tx.SelectStr(sqlInsertManagerNode, uuid, mgrType, item.Id, jsonTc, item.Name, item.Region, managerAcctId)
+			logrus.Debugf("adding manager node %s with cloud details: %s %s %s ", item.Name, item.Id, item.Region, managerAcctId)
```
stevendanna: I'm cool with these kinda debugging log statements,  but I'd prefer something like:

```suggestion
			logctx := logrus.WithFields(logrus.Fields{
				"name": item.Name,
				"id": item.Id,
				"region": item.Region,
				"account_id": managerAcctId,
			})
			logctx.Debug("adding manager node")
```

so then we can reuse logctx at line 242 to ensure we get consistent logging.

components/nodemanager-service/pgdb/node-managers-process-node.go
```diff
@@ -103,6 +105,8 @@ func (db *DB) ProcessIncomingNode(node *manager.NodeMetadata) error {
 
 	lastContact := ptypes.TimestampString(node.LastContact)
 
+	// note: we shouldn't be able to ever get here without a node id, because this function is called from the
+	// ingestion process, and the ingestion process will reject a report with no node uuid
```
stevendanna: Given that, might it make sense to warn or fail here instead?

components/nodemanager-service/pgdb/node-managers-process-node.go
```diff
@@ -138,33 +152,26 @@ func (db *DB) ProcessIncomingNode(node *manager.NodeMetadata) error {
...
-					node.GetSourceAccountId(), node.GetJobUuid(), lastContactDataByte, projectsDataByte, mgrType)
-			} else if node.GetRunData() != nil {
-				id, err = db.SelectStr(sqlUpsertBySourceIDRunData, node.GetUuid(),
-					node.GetName(), node.GetPlatformName(), node.GetPlatformRelease(),
-					nodeState, lastContact, node.GetSourceId(), node.GetSourceRegion(),
-					node.GetSourceAccountId(), lastContactDataByte, projectsDataByte, mgrType)
-			}
-			if len(id) > 0 {
-				node.Uuid = id
+			node.Uuid, err = db.upsertByCloudDetails(node, nodeDetails)
```
stevendanna: I'm a bit confused on why we are using `db` rather than `tx` here.  I think this implies that are happening outside of the transaction?

components/nodemanager-service/pgdb/node-managers-process-node.go
vjeffrey: ya i don't love that i have to do this -- sounds like if we grab the actual region (i didn't know we had one from the chef client report) i will be able to delete this part!

components/ingest-service/pipeline/publisher/node_manager_publisher.go
vjeffrey: it uses region -- yes it would be awesome to grab the region instead - i didn't know it was available

vjeffrey:
> I think we should consider making some top level fields to query that are generic versions of these compound id fields. Like source_id, account_id and region_id. Then we could query without needing to know if we are looking at azure or ec2.

YES!!!! that sounds great!
Also: if you have some time, I would totally appreciate some help with that, as I am rather unfamiliar with the ohai data fields and ingest service bits.

components/nodemanager-service/pgdb/node-managers-cloud.go
vjeffrey: oooh nice!! ya ill change it to that!

components/nodemanager-service/pgdb/node-managers-process-node.go
vjeffrey: ah good point! yes!

components/nodemanager-service/pgdb/node-managers-process-node.go
vjeffrey: hmm ya. good point, i missed that. ill see about switching to tx

vjeffrey:
thank you for the commit @kmacgugan !! 🎉 ⭐️

vjeffrey:
rebasing on master

Issue #2436, opened 2019-12-05
Project percentage complete does not update


bcmdarroch:
## Describe the bug
Project percentage complete does not update. In the UI, it sits at 0% then turns to `Projects Up-to-Date` after a period of time.

The problem stems from the API, which is not returning updated percentage complete values.

## To Reproduce
API:
- `chef_load_nodes 500`
- ` curl -kH "api-token: $TOK" -XPOST -d '{"name":"project1", "id":"project1"}' https://a2-dev.test/apis/iam/v2beta/projects`
- `curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/projects/project1/rules -d "$(jq -n '{ id: "rule1", name: "my foo rule", type: "NODE", project_id: "project1", conditions: [{ operator: "EQUALS", attribute: "CHEF_ORGS", values: ["The Misfits"]}]}')" | jq .`

- start a project update then keep curling the status endpoint until it stops running
```
[40][default:/src:0]# curl -kH "api-token: $TOK" -X POST https://localhost/apis/iam/v2beta/apply-rules | jq
{}

[47][default:/src:0]# curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/apply-rules | jq
{
  "state": "running",
  "estimated_time_complete": "0001-01-01T00:00:00Z",
  "percentage_complete": 0,
  "failed": false,
  "failure_message": "",
  "cancelled": false
}

[48][default:/src:0]# curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/apply-rules | jq
{
  "state": "running",
  "estimated_time_complete": "0001-01-01T00:00:00Z",
  "percentage_complete": 0,
  "failed": false,
  "failure_message": "",
  "cancelled": false
}

# ... status response stays the same, until it stops running:

[49][default:/src:0]# curl -kH "api-token: $TOK" https://localhost/apis/iam/v2beta/apply-rules | jq
{
  "state": "not_running",
  "estimated_time_complete": "0001-01-01T00:00:00Z",
  "percentage_complete": 1,
  "failed": false,
  "failure_message": "",
  "cancelled": false
}
```

UI:
- chef_load_nodes 500
- create a project or two
- add some rules to those projects
- on the project list page, select `Update Projects`
- the percentage complete never changes on the update button or in the stop update modal

## Expected behavior
The `apply-rules` status API returns increasing values of percentage complete as the operation takes place.

[...]

PR #2532, opened 2019-12-17
Fixing the way the project update precentage is calculated

lancewf:

There was a bug in the way we were calculating the percentage complete for a project update. We were just taking the last value from all the domain services. This change cleans up the percentage complete function to remove the cereal code to make the code easier to read. 
 
#2436 

### :+1: Definition of Done

The percentage complete does not show 100% when the time is in the future. 

1. Add a lot of data to compliance with `chef_load_compliance_scans -N1000 -D10 -M2 -T20000`
1. Create a project
1. Apply the rules
1. Ensure the percentage complete is not always 100%

- [x] Tests added/updated?
- [ ] Docs added/updated?


components/authz-service/server/v2/project_update_manager.go
```diff
@@ -285,45 +284,20 @@ func (w *workflowInstance) PercentageComplete() float64 {
...
-		if subWorkflow.IsRunning() {
-			if err := subWorkflow.GetPayload(&payload); err != nil {
-				logrus.WithError(err).Errorf("failed to get payload for %q", d)
-				continue
-			}
-
-			if !payload.MergedJobStatus.Completed {
-				estimatedTime := time.Unix(payload.MergedJobStatus.EstimatedEndTimeInSec, 0)
-				if estimatedTime.After(longestEstimatedTimeComplete) {
-					percentComplete = float64(payload.MergedJobStatus.PercentageComplete)
```
lancewf: Here is where the bug was. We should have been setting `longestEstimatedTimeComplete = estimatedTime`.

lib/authz/project_rules.go
```diff
@@ -22,7 +22,7 @@ func FindSlowestJobStatus(jobStatuses []JobStatus) JobStatus {
 	for _, jobStatus := range jobStatuses {
 		if !jobStatus.Completed {
 			combinedJobStatus.Completed = false
-			if jobStatus.EstimatedEndTimeInSec > combinedJobStatus.EstimatedEndTimeInSec {
+			if jobStatus.EstimatedEndTimeInSec >= combinedJobStatus.EstimatedEndTimeInSec {
```
lancewf: This allows for the case when no statuses have a time estimate.

components/authz-service/server/v2/project_update_manager.go
```diff
@@ -285,45 +284,28 @@ func (w *workflowInstance) PercentageComplete() float64 {
...
 
-	percentComplete := 0.0
-	domainServicesUpdateInstance, err := w.GetUpdateDomainServicesInstance()
-	if err == cereal.ErrWorkflowInstanceNotFound {
-		return percentComplete
-	}
-
-	domainServices := domainServicesUpdateInstance.ListSubWorkflows()
-	if len(domainServices) == 0 {
+	domainJobStatuses := w.collectAllDomainServiceJobStatuses()
```
lancewf: Below is what the function looks like in this PR. 
```
func (w *workflowInstance) PercentageComplete() float64 {
	if !w.IsRunning() {
		return 1.0
	}

	domainJobStatuses := w.collectAllDomainServiceJobStatuses()
	if len(domainJobStatuses) == 0 {
		return 1.0
	}

	longestDomainJobStatus := project_update_tags.FindSlowestJobStatus(domainJobStatuses)

	return float64(longestDomainJobStatus.PercentageComplete)
}
```

components/authz-service/server/v2/project_update_manager_internal_test.go
```diff
@@ -1,51 +0,0 @@
-package v2
-
-import (
-	"testing"
-	"time"
-
-	"github.com/stretchr/testify/assert"
-)
-
-func TestFindLongestEstimatedTimeEmpty(t *testing.T) {
```
lancewf: This function was removed for the already existing `FindSlowestJobStatus` function.

components/authz-service/server/v2/project_update_manager.go
```diff
@@ -408,27 +390,14 @@ func (m *CerealProjectUpdateManager) Status() (ProjectUpdateStatus, error) {
 	return projectUpdateInstance, nil
 }
 
-// Find the longest estimate that is in the future. If no estimate is found return zero time.Time{}
-func findLongestEstimatedTime(estimatedTimes []time.Time) time.Time {
```
lancewf: Replaced with the already existing `FindSlowestJobStatus` function.

components/authz-service/server/v2/project_update_manager.go
```diff
@@ -408,27 +390,14 @@ func (m *CerealProjectUpdateManager) Status() (ProjectUpdateStatus, error) {
...
-			longestEstimatedTimeComplete = estimatedTime
-		}
-	}
-
-	return longestEstimatedTimeComplete
-}
-
-func (w *workflowInstance) collectAllDomainServiceEstimates() []time.Time {
-	estimates := make([]time.Time, 0)
+func (w *workflowInstance) collectAllDomainServiceJobStatuses() []project_update_tags.JobStatus {
```
lancewf: All the cereal logic is hidden in this function now.

msorens:
Some observations:
1. ❓Compliance percentage is **not** strictly monotonic increasing. Sample: 15%, 78%, 63%, 100%
2. ❓Compliance estimated time is **not** strictly monotonic increasing. Sample: 1576622615, 1576622536, 1576622570, 1576622537, 1576622565, 1576622537
3. ✔️ Compliance completed flag remains at `false` until done, then remains at `true`. 
4. ❓Compliance estimated completion time may still change after `Completed` becomes `true`.
5. ❌Compliance percentage can decrease after touching 100%. Sample: 15%, 100%, 93%, 100% (Note that `Completed` only became `true` on the **final** 100) 

Not sure if there is any significance but the logging calls show that `JobStatus`, whenever it is called, is invoked **twice**--and the reported values are different. If, for example, I only look at the first call of each pair, then several of the issues above go away.

lancewf:
@msorens 
❓Compliance percentage is not strictly monotonic increasing. Sample: 15%, 78%, 63%, 100%
**True**
❓Compliance estimated time is not strictly monotonic increasing. Sample: 1576622615, 1576622536, 1576622570, 1576622537, 1576622565, 1576622537
**True**
✔️ Compliance completed flag remains at false until done, then remains at true.
**True**
❓Compliance estimated completion time may still change after Completed becomes true.
**false** _I have not seen this behavior._ 
❌Compliance percentage can decrease after touching 100%. Sample: 15%, 100%, 93%, 100% (Note that Completed only became true on the final 100)
**True**

The Completed flag is the surest value. Where the percentage complete and the estimated time fluctuate when the project update is first starting. The estimated time is calculated of the percentage complete. 

Note: This PR is only trying to fix the problem that we were only displaying the percentage complete of the last job and not the job that is predicted to run the longest.

susanev:
@lancewf do you mind opening up a separate issue around 

> Note: This PR is only trying to fix the problem that we were only displaying the percentage complete of the last job and not the job that is predicted to run the longest.

lancewf:
Added https://github.com/chef/automate/issues/2540

msorens:
My observations were to try to determine if there are still other bugs requiring attention; I think you are saying that there are not. 
However, the one that concerns me is that the percentage might go to 100% even when `Completed` is still `false`. Seems like that directly ties in with the topic of this PR.

Issue #1780, opened 2019-10-03
Add Filter Buttons for Disconnected Services in the Applications View


apriofrost:
## Description

Allow users to find and remediate disconnected nodes by adding filter buttons for disconnected services for both the service group table and the service detail list(right side-panel).

## Definition of Done
![image.png](https://images.zenhubusercontent.com/5a721a104b5806bc2bc51768/210957fb-3d68-45dd-9b41-d030159721e6)
* Add a big filter button for disconnected services groups above the service groups table
* Add a small filter button for disconnected services above the service details list
* For the big filter, when it's selected, only service groups that have one or more disconnected services are shown in the service groups table
* For the small filter, when it's selected, only disconnected services are shown in the services detail list.
* Responsive filter size
  - For big filters: when resizing the screen, they respond to the new size while keeping the same padding and a consistent `8px` margin from each other. The total width of the big filters should match with that of the table underneath. The minimal width of the individual big filter is `115px`. Create a new line when the existing line is full with all filters at the minimum width.
<img src="https://images.zenhubusercontent.com/5a721a104b5806bc2bc51768/0c5bc69e-d98e-4cdd-8161-2fbec3712104" width="200">
  - For small filters: currently, the width of the side-bar is fix-width. The width of the smaller filters only responds to its content (the count numbers) while keeping the consistent padding and margin; create a new line when the existing one is full. 

### Backend 
Done, see [PR](https://github.com/chef/automate/pull/1641)

* API for service groups can filter for disconnected
* API for health counts includes the count of disconnected
* API for services by service group can filter for disconnected
* API for health counts for services in a service group can filter for disconnected

### UI Demo
View demo video in Google Drive: https://drive.google.com/open?id=1YMnYQq7ViS4yhgijVZqhtAREfMc2h342
#### Implementation notes
* Reach out to Liz to invite you to the InVision project so that you could use the inspect feature to see the styling details.
* Styles for both filters:
  * Reuse the existing `icon/disconnected service` status icon
  * when hovered, the icon color changes from `#C22DD5` to `$chef-info-alert`; the background color changes from `FFFFFF` to `$chef-info-alert-light`; the background border becomes hidden.
  * when selected, add the background border back to 1px, inside, `#C22DD5`.
* Mockup link: https://chef.invisionapp.com/share/C3T08NGNKEQ#/388380424_step_2-_add_single_select_filter

scottopherson:
[...]

PR #2535, opened 2019-12-17
Add disconnected filter to chef-ui-library

danielsdeleo:

To implement https://github.com/chef/automate/issues/1780 we need the fuchsia colored filters for the disconnected status. This PR adds them to the library.

I also changed the margins and wrapping at @apriofrost's request.

https://chef.invisionapp.com/share/C3T08NGNKEQ#/388380424_step_2-_add_single_select_filter

Not-selected state:
Screen Shot of filter buttons in the not-selected state

Selected state:
Screen Shot of filter button in the selected state

The lean version:
screenshot of the lean version of the filter button


components/chef-ui-library/src/molecules/chef-status-filter-group/chef-status-filter-group.scss
```diff
@@ -104,6 +118,11 @@ chef-status-filter-group {
         background: #E7E9E9;
         box-shadow: 0px 0px 0px 1px var(--chef-dark-grey) inset;
       }
+
+      &.disconnected {
+        background: #FADAFA;
```
danielsdeleo: This exists in automate-ui as "fuschia-light" https://github.com/chef/automate/blob/42fb0448535b9df5a8b1acc1b29000ab836875c6/components/automate-ui/src/styles/_colors.scss#L19 (which I guess is a misspelling and it should be "fuchsia" FWIW). Is it necessary to unify these things? How would I do that?

components/chef-ui-library/src/molecules/chef-status-filter-group/chef-status-filter-group.scss
```diff
@@ -104,6 +118,11 @@ chef-status-filter-group {
...
+        box-shadow: 0px 0px 0px 1px #A424B4 inset;
```
danielsdeleo: Likewise, this is `$fuschia` in the same file

components/chef-ui-library/src/molecules/chef-status-filter-group/chef-status-filter-group.scss
```diff
@@ -10,10 +11,10 @@ chef-status-filter-group {
     border-radius: 4px;
     font-size: 14px;
     display: inline-flex;
-    margin-bottom: 16px;
+    margin-bottom: 10px;
 
     &:not(:last-child) {
-      margin-right: 16px;
+      margin-right: 10px;
```
susanev: @apriofrost i think our standard for space between buttons in automate has been at least 16px of space to prevent folks from interacting with buttons that they don't intend. could you tell me a bit more about the problem you're solving here?

danielsdeleo: For reference, the "lean" version of these filter buttons is/has been a lot more compact: https://github.com/chef/automate/blob/783f03c60988c92a8b88e4dd1664b5fb85e03bb4/components/chef-ui-library/src/molecules/chef-status-filter-group/chef-status-filter-group.scss#L134-L153

apriofrost: @susanev , I prefer it to be 10 px so that it's consistent with the space measurement we have for other UI molecules in this view. For example, the rows and headers in the tables has a shared distance of 10px. Same for the cards in the side nav. 
The idea is that by reducing the distance, the individual atom within a molecule feels like a group that shares the same look, function, and interactivity. While in between molecules, I set the space as 16px so they feel further apart from each other. 
In this example, all the different status filters form a group of filters, that serve the same purpose of showing an aggregated count of service groups in that status. I would worry more about the space between buttons in a group since the button size is usually much smaller than the status filter here. But since the status filter is already large enough, I won't worry about that. 
What do you think?

apriofrost:
I missed one thing: Can we make the width of the disconnected filter the same with the rest? It seems to be wider than the rest now. 

The spec is like this (copied from #1780 )
* Responsive filter size
  - For big filters: when resizing the screen, they respond to the new size while keeping the same padding and a consistent `8px` margin from each other. The total width of the big filters should match with that of the table underneath. The minimal width of the individual big filter is `115px`. Create a new line when the existing line is full with all filters at the minimum width.

components/chef-ui-library/src/molecules/chef-status-filter-group/chef-status-filter-group.scss
```diff
@@ -21,7 +21,7 @@ chef-status-filter-group {
       display: flex;
       align-items: center;
       flex-wrap: wrap;
-      min-width: 100px;
+      width: 100px;
```
danielsdeleo: This was the only way I could figure to make the filters all the same size. There is something happening with the CSS box model that I don't completely understand. When I did not force the width to 100px, then the filter for disconnected was computed to be 113px wide, even though it had a lot of blank space on the right side

rendered element:
Screen Shot 2019-12-18 at 2 09 58 PM

box model debugger thingy:
Screen Shot 2019-12-18 at 2 10 20 PM

I'm guessing that one or more of the elements that display below the text are added to the text's width for the box sizing computation but I don't know what to do about it.

In any case, this works as long as the label is less than 100px so it's probably fine 🤷‍♀

danielsdeleo: This shows the not-padding-padding that is still there when forced to 100px

Screen Shot 2019-12-18 at 3 08 08 PM

components/chef-ui-library/src/molecules/chef-status-filter-group/chef-status-filter-group.scss
susanev: I haven't pulled down the pr so maybe I misunderstood I was mostly worried about the accessibility issues caused by not having enough space between buttons, if that's not relevant here feel free to ignore my comment.

apriofrost: No worries, feel free to take a look at it later when you have time and see how it feels to you :)

components/chef-ui-library/src/molecules/chef-status-filter-group/chef-status-filter-group.scss
danielsdeleo: This is what it looks like with the width forced to 100px change:

Screen Shot 2019-12-19 at 9 24 43 AM

components/automate-ui/src/styles/_colors.scss
```diff
@@ -15,9 +15,9 @@ $pink: #DC267F;
 $violet: #C22DD5;
 $indigo: #785EF0;
 $science-blue: #0075DB;
-$fuschia: #A424B4;
-$fuschia-light: #FADAFA;
-$fuschia-dark: #440F4B;
+$fuchsia: #A424B4;
```
Shadae: Leonhart Fuchs would be proud!

PR #2563, opened 2019-12-19
How to create an effortless project

### LINKED ISSUE #2552: Add necessary documentation to be able to use effortless nodes with ingest rules
## Overview
Add necessary documentation to be able to use effortless nodes with ingest rules. 
Use the information from the below investigation to update the docs. 
https://github.com/chef/automate/issues/2463

lancewf:
Add necessary documentation to be able to use effortless nodes with project ingesting rules.

This new text and image was added above the "Projects in the API" https://automate.chef.io/docs/iam-v2-overview/#projects-in-the-api

fixes #2552 

image


components/automate-chef-io/content/docs/iam-v2-overview.md
```diff
@@ -264,6 +264,12 @@ From there, you can select any individual rule to view its list of conditions, a
 
 ![](/images/docs/admin-projects.png)
 
+To create a project that contains all Effortless Infra nodes add a Node ingest rule with a condition that uses an attribute of Chef Server and a value of "localhost".
+
+![](/images/docs/effortless-project-rule.png)
+
+The above rule is matching on a node's Chef Server field, which is set to "localhost". This works because all Effortless Infra nodes do not have a Chef Server. Then, by adding a second condition that matches a specific Chef Policy Name, subgroups of Effortless Infra nodes can be created.
```
bcmdarroch: s/is matching/matches

bcmdarroch: also I'm not sure what you mean by `subgroups`

lancewf: I have gone back and forth on that word. I was thinking that the Effortless Infra nodes are a group, then the subgroup is the nodes matching a specific policy name.

lancewf:
I think @bcmdarroch is right that this should be move to the [IAM v2 guide section on ingested resources](https://github.com/chef/automate/blob/90d9370894eba2bdddc5b5f0777a181b7eb62abc/components/automate-chef-io/content/docs/iam-v2-guide.md#assigning-ingested-resources-to-projects) section.

components/automate-chef-io/content/docs/iam-v2-overview.md
blakestier: Filtered nodes? Or maybe explicitly call them "nodes matching a specified policy name"?

components/automate-chef-io/content/docs/iam-v2-guide.md
```diff
@@ -233,7 +233,7 @@ In order to find a token's ID, visit the *API Tokens* page.
 
 ### Projects
 
-Projects are used to group and permission Chef Automate resources as well as ingested data, specifically Compliance reports, Chef Server and Compliance events, and Infrastructure nodes.
+Projects are used to group and permission Chef Automate resources as well as ingested data, specifically Compliance reports, Chef Server events, and Infrastructure nodes.
```
lancewf: Removing the "Compliance events" part because is not implemented in Automate. Meaning that we are not tagging or filtering on projects for compliance events.

components/automate-chef-io/content/docs/iam-v2-guide.md
```diff
@@ -394,6 +394,13 @@ Rules of type `Node` can have conditions with attributes `Chef Organization`, `C
 Navigate to the project list page once more. Select `Update Projects`. Upon completion of the update, you should be able to
 filter by `project-devops` across Automate's dashboards and see only the ingested data that you expect.
 
+#### Effortless Infra Project
+To create a project that contains all Effortless Infra nodes add a Node ingest rule with a condition that uses an attribute of Chef Server and a value of "localhost".
```
blakestier:
```suggestion
To create a project that contains all Effortless Infra nodes, add a Node ingest rule with a condition that uses an attribute of Chef Server and a value of "localhost".
```

components/automate-chef-io/content/docs/iam-v2-guide.md
```diff
@@ -394,6 +394,13 @@ Rules of type `Node` can have conditions with attributes `Chef Organization`, `C
 Navigate to the project list page once more. Select `Update Projects`. Upon completion of the update, you should be able to
 filter by `project-devops` across Automate's dashboards and see only the ingested data that you expect.
 
+#### Effortless Infra Project
```
blakestier: This section has some inconsistent capitalization but I'm not sure what our standards are (Node Ingest Rule? node ingest rule? Do we capitalize 'node'? I have no idea.

lancewf: Thank you for catching that. I updated it to be consistent with the lower case 'node'.

components/automate-chef-io/content/docs/iam-v2-guide.md
```diff
@@ -394,6 +394,13 @@ Rules of type `Node` can have conditions with attributes `Chef Organization`, `C
...
+To create a project that contains all Effortless Infra nodes, add a node ingest rule with a condition that uses an attribute of Chef Server and a value of "localhost".
+
+![](/images/docs/effortless-project-rule.png)
+
+The above rule matches on a node's Chef Server field, which is set to "localhost". This works because all Effortless Infra nodes do not have a Chef Server. Then, by adding a second condition that matches a specific Chef Policy name, subgroups (i.e., nodes matching a specified policy name) of Effortless Infra nodes can be created.
```
mjingle: @lancewf To clarify, to create a project that contains all Effortless Infra nodes, you need 1) the condition shown in the included screenshot, and 2) a second condition that matches a specific Chef Policy name specific to the user's set-up? Or is including the second condition optional?

susanev: to get all effortless nodes you need a condition to match _localhost_ in the chef server attribute, then if you want to scope that down, even more, you could add additional conditions.

mjingle: @lancewf A question about "This works because all Effortless Infra nodes do not have a Chef Server."

 In this case, is 'localhost' considered the default value of a `Chef Server` attribute? And thus, it is not set to a different value because a `Chef Server` attribute isn't used in Effortless Infra?

lancewf: @mjingle You could think of "localhost" as the default value, but it is closer to when Effortless Infra is used that is the value that Chef Server is set to. This is because Effortless Infra does not have a Chef Server. When Effortless Infra is not used, the IP address or domain name of the Chef Server is used in place of the "localhost".

PR #2565, opened 2019-12-19
Filter service groups and services by disconnected status

danielsdeleo:

Users want to know which services are disconnected from automate so they can remediate problems. We currently have badges for disconnected services but no way to filter for them.

#1780 

Build:
* `build components/automate-ui`
* `start_all_services`

Sample Data:
Generating the sample data is time-sensitive since the data generator doesn't create health check messages with timestamps in the past. What I did was:
* run `applications_populate_database`
* Wait 5m
* run it again

Then within the next 5 minutes:
* Filter for `environment: production`, this reduces the data down to 5 service groups
* Click the service groups disconnected status filter, there should be only 4 matching service groups (the group `wizrd.default` should be filtered out and not visible)
* All the service groups visible on the page should have 1/2 of the services disconnected. Select one of the service groups, then click the services disconnected status filter, the services list should be refreshed with only the disconnected service(s) visible.

not-selected:
Screen Shot 2019-12-19 at 3 04 49 PM

hover:
Screen Shot 2019-12-19 at 3 05 13 PM

selected:
Screen Shot 2019-12-19 at 3 05 52 PM

lean version in the side panel:
Screen Shot 2019-12-19 at 3 06 15 PM

### DIFF
--- a/components/automate-ui/src/app/entities/service-groups/service-groups.facade.ts
+++ b/components/automate-ui/src/app/entities/service-groups/service-groups.facade.ts
@@ -45,7 +45,7 @@ export class ServiceGroupsFacadeService {
   public currentServicesFilters$: Observable<GroupServicesFilters>;
 
   // The collection of allowable status
-  public allowedStatus = ['ok', 'critical', 'warning', 'unknown'];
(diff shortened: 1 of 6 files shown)


apriofrost:
Dan demoed it to me, everything looks perfect! 🎆 🍾 🎊

Issue #2554, opened 2019-12-19
Applications: Display health check messages for services


apriofrost:
## User Story
When services are not in an OK status, users would like to know more about what exactly went wrong with the service beyond the status of critical, warning or unknown. In Habitat, we've provided a way for users to write their own health check messages which contain more details about the service health status. However, currently, we are not showing the messages anywhere on the service group page.

In addition, we have limitations as follows: 
We couldn't control what users write for the health check, or what content in the messages are returned to Automate. Considered the limitation of network bandwidth, we've limited the maximum size of a message to 1KB in Habitat. Besides, we couldn't control which type of messages, be it standard output or standout error, that users choose to use. We could, in the future, provide guidance on writing good health checks but they don't exist today.

## Solution
![image.png](https://images.zenhubusercontent.com/5a721a104b5806bc2bc51768/c7bddc0a-c2c5-489b-9674-f4e0cbce98cf)
We've come to a solution illustrated in the [mockup](https://chef.invisionapp.com/share/AJV5V0176PU#/screens/398055717_health_Check_Message-V4-Default) with the following thoughts in mind:

1. By default, we display the first three lines of the STDOUT or STDERR, it’s about 140 characters on a 1440px widescreen
2. We always show the health check messages received for any services in a critical, warning or unknow status. If there are two types of messages, we display the STDERR on top of STDOUT. If there is none to display, we show an empty status:

> Health check massages not available. When services specify a health-check hook, the hook's output will be shown here. Learn more about writing health-check hooks [link to docs].

  we don’t have the docs about the health check messages now, we would only show the first two sentences until we have it ready.
3.  We always hide the health check messages received for any services in an OK status
4.  Since each message is limited to 1KB, we load all messages on a page at once.
5.  Move the position of the disconnected service status banner, from above the service identifiers to underneath, and grouped together with the health check status.
6.  Move the position of the time wizard (the relative time showing when the status changed) to be part of the health check status banner, above the actual health check messages
7.  In the future, we plan to add two filters named “Health Check Error Message” and “Health Check Output Message” to the search bar, and allow users to filter on services based on the content of these messages.
[...]

PR #2572, opened 2019-12-20
Show Services' Health Check Messages

danielsdeleo:

Display services' health check messages in the applications tab UI. This enables expedited troubleshooting of problems when users emit helpful information from their health check scripts.

#2554

NOTE: this doesn't include the show more/show less accordion for long messages. That will be done later.

Build:
* `build components/automate-ui`

Test data:
There are some improvements to the sample data to show all of the different UI cases. Either restart the studio or `source .studio/applications-service`. Then `applications_populate_database`.

Verify Behavior:
* Filter for service name `pos-terminal`
* The four `pos-terminal.default` service groups show each of the following cases:
  * (production-us-west environment) no health check messages at all: shows informational message with link to documentation
  * (production-us-central environment)  stderr only: shows the stderr and no empty chrome for stdout
  * (production-us-mountain environment) stdout only: shows the stdout and no empty chrome for stderr
  * (production-us-east environment) stdout and stderr are both shown.
* filter for service name `nginx`
  * the bldr cache app in qa env shows messages of the maximum size. They wrap even without word breaks
  * the demo/demo one shows what we think a typical message might be.

Full page:
Screen Shot 2019-12-20 at 1 58 51 PM

Different health statuses (critical/warning/unknown/ok) with stdout, stderr, both, neither:
Screen Shot 2019-12-20 at 1 36 20 PM
Screen Shot 2019-12-20 at 1 36 34 PM
Screen Shot 2019-12-20 at 1 36 49 PM
Screen Shot 2019-12-20 at 2 00 58 PM

Screen Shot 2019-12-20 at 1 37 23 PM
Screen Shot 2019-12-20 at 1 37 45 PM
Screen Shot 2019-12-20 at 1 38 00 PM
Screen Shot 2019-12-20 at 1 38 11 PM
Screen Shot 2019-12-20 at 1 38 24 PM

### DIFF
--- a/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.scss
+++ b/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.scss
@@ -57,6 +57,7 @@ chef-alert {
     display: block;
     background-color: $chef-white;
     border-top: 1px solid $gray-pale;
+    // if changed then hr.health-info-divider needs to be adjusted also
     padding: 1em;
     position: relative;
     width: 100%;
@@ -104,6 +105,11 @@ chef-alert {
     align-items: center;
     margin-bottom: 8px;
     font-size: 14px;
+  }
+
+  hr.health-info-divider {
+    margin: 16px -1em 16px -1em;
+    background-color: $gray-pale;
   }
 
   .health-info-panel {
--- a/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.scss
+++ b/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.scss
@@ -160,8 +160,8 @@ chef-alert {
   }
 
   .no-healthcheck-messages {
-    border: 1px solid;
     margin: 1px;
+    border: 1px solid;
     border-radius: $global-radius;
     padding: 3px 8px 3px 8px;
 
--- a/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.scss
+++ b/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.scss
@@ -113,7 +113,7 @@ chef-alert {
   }
 
   .health-info-panel {
-    border-radius: $global-radius;
+    border-radius: 2px;
     align-items: center;
     padding: 8px;
     font-size: 12px;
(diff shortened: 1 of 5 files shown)


apriofrost:
Hi @danielsdeleo Overall this looks great!! I have some nitpick for you to consider iterating on:

1. **Text link**: Instead of having the entire sentence of "Learn more about writing health-check hooks" as a link to the doc, could we just have the "writing health-check hooks" with the link? This way, the text would match more closely with what the doc is about. It would look like this: image
2. **Radius size**: We talked about changing the radius size for all the new UI elements to be 4px. But looking at the result from the screenshots, I think the original 2px design would make it look sleeker and give higher priority to the content itself, rather than the shape of the container. I would suggest changing them all to 2px if it's possible or easy to do.
3. **Space between messages**: The space between the error and output messages when they co-exist seems a bit narrower than expected. In the image below, I heightened the areas where space should all be 8px.
image
4. **Missing divider**: In the mockup, there is also a divider between the service details info and the status-related notifiers. The distance from the divider to the element below and above it is 16px. 
image
Let me know if you have any questions about them. 😺

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.scss
```diff
@@ -106,24 +107,121 @@ chef-alert {
     font-size: 14px;
   }
 
-  .health-line {
-    display: flex;
+  hr.health-info-divider {
+    margin: 16px -1em 16px -1em;
```
danielsdeleo: So I think this semantically is more like a `hr` than a border on a div. But to do it that way means we have to negate the padding on the enclosing element which couples those two things 😬 🤷‍♀

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.scss
```diff
@@ -106,24 +107,121 @@ chef-alert {
...
+    &.OK {
+      background-color: var(--chef-badge-success-background);
+      color: var(--chef-badge-success-color);
+    }
+
   }
 
-  .channel-line {
-    padding-left: 18px;
+  .disconnect-notice {
```
danielsdeleo: Most of this isn't new, I moved it to be as nested as the other things so this file would be consistent with itself.

danielsdeleo:
K, I made the following changes:

* make the `disconnect-notice` and `health-info-panel` and `no-healthcheck-messages` border radius 2px
* add horizontal rule between basic service info and health/connectedness info panels
* change the text that's in/out of the link tags
* set space between two `healthcheck-message`s to 8px

PR #2579, opened 2019-12-23
[automate-2578] bump node version 10.15.2 -> 12.13.1

msorens:

Periodic updating of node.js version.

**See commit messages for specific details of the updates.**

Though Node.js is up to v14, production applications should use either "Active" or "Maintenance" phase versions, and v12 is the latest "Active LTS" version available:

image

_References:_
 - [Node.js Releases](https://nodejs.org/en/about/releases/)
 - [Node.js v12 - New Features You Shouldn't Miss](https://blog.risingstack.com/node-js-12-new-features/)

**Prior to this PR...** was a [core-plans PR](https://github.com/habitat-sh/core-plans/pull/3158) to install this version of node into the habitat depot.

**Once this merges...** need to provide notice in multiple Slack channels that if you build UI components locally you need to:
  (a) upgrade your own version of node (`nvm install 12.13.1`) and
  (b) update your node_modules directory (`npm install`).

**After this PR is merged and acceptance tested...** 
need a follow-up PR in [chef/release-engineering](https://github.com/chef/release-engineering) to update the docker file (a la https://github.com/chef/release-engineering/pull/843).

_Current:_
Epic #2553 Bump Angular from 8.x to 9.x
_Historical:_
#785 Bump Node.js version 8.9.4 -> 10.15.2 (automate-ui and chef-ui-library)
#2267 Bump Node.js version 5.6.0 -> 10.15.2 (automate-workflow-web)

### :+1: Definition of Done
 - Everything still builds and runs.
 - None of the `npm install` commands should generate any non-committed changes to the respective package-lock.json files.

1. Upgrade your local version of node.
```
nvm install 12.13.1
```

2. Update your packages corresponding to the node version change (order matters!).
```
cd automate/components/chef-ui-library
npm install
cd components/automate-ui
npm install
cd components/automate-workflow-web
npm install
```

3. Exercise your automate-ui server
```
make serve
```

4. Exercise workflow server

Detailed instructions for this legacy feature are available at components/automate-chef-io/content/docs/workflow_install.md.

(The "quick and dirty" approach is to run `start_workflow_server` in hab studio, then navigate to https://a2-dev.test/workflow.)


.studio/automate-ui
```diff
@@ -259,7 +259,7 @@ function _install_nodejs_from_nvmrc() {
   fi
 
   log_line "Installing NodeJS (v.$NODE_VERSION)"
-  OPTS="-b" install "core/node10/$NODE_VERSION" >/dev/null
+  OPTS="-b" install "core/node/$NODE_VERSION" >/dev/null
```
msorens: Two previous PRs (#785 and #2267) updated plan.sh
in a couple projects to be specific to a node.js major version.
I now believe that was an unnecessary, and in fact, brittle change.
A node.js package newly installed in core-plans makes its way
to the "main branch" (https://bldr.habitat.sh/#/pkgs/core/node).
References should remain invariant to that location
so they do not need updating when node.js major versions change.

components/automate-ui/package-lock.json
```diff
@@ -612,7 +612,8 @@
             "ansi-regex": {
               "version": "2.1.1",
               "bundled": true,
-              "dev": true
+              "dev": true,
+              "optional": true
```
msorens: All package-lock.json changes are adding the `optional` property to a large number of packages.

msorens:
@SEAjamieD There should be no change to package-lock.json. Can you either post the diff here or send me the file via slack?

SEAjamieD:
@msorens - yes - i'm gonna revert and rerun this morning for a sanity check and will update with diff or an OK

SEAjamieD:
@msorens here's the diff, it's pretty small
```
     "es6-promisify": {
       "version": "5.0.0",
-      "resolved": "http://registry.npmjs.org/es6-promisify/-/es6-promisify-5.0.0.tgz",
+      "resolved": "https://registry.npmjs.org/es6-promisify/-/es6-promisify-5.0.0.tgz",
```

msorens:
@SEAjamieD Based on your feedback on the http/https discrepancy, I did some digging and found that there was, in fact, a problem with the way packages were recorded in the npm registry that caused this, so some packages generated http addresses in  package-lock.json for a time. That has allegedly been fixed. I suspect the reason you were seeing it correctly as https and I am seeing http, is that my local npm cache was built quite a while before yours was and mine had the http problem embedded in it.

While that could be a security risk, it turns out it is not. Per this post (https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285/16), "npm cli will always use HTTPS when talking to the registry for tarballs even though it says HTTP [in the package-lock.json file]".

SEAjamieD:
@msorens - Thank you for educating me on this, I really appreciate it!

PR #2173, opened 2019-11-12
[infra-proxy-service] Add infra-proxy-service & manage servers & orgs APIs

vsingh-msys:
### :nut_and_bolt: Description: What code changed, and why?
Implement a new service that is managing the chef servers ends points and teach how to fetch data of the desired state. 

As of now, we are dealing with current state data of chef node which might fails and loss of data if the particular node/server fails during convergence.

In order to make a single source of truth for chef infra servers data here, we have added `infra-proxy-service` service.

#### In this PR following points I have covered:
- Connect `infra-proxy-service` to `deployment-service`.
- Provide gPRC and HTTP endpoints to manage chef infra servers & orgs.
- CRUD for infra servers.
- CRUD for infra server orgs.

#### Pending tasks:
1. Add IAM policies for admin users only.
2. Manage infra server user's KEYS using `secrets-service`.
3. Add Unit & integration test cases.
4. Documentations of `infra-proxy-service`

As I am a newbie to Automate and I have tried to cover as much as possible in order to get the understating of Automate code flow.

Please reviews and do provide the feedback.

Also, I am very thankful to @stevendanna  who guided me with proceedings.

Thanks

### :chains: Related Resources
https://github.com/chef/automate/issues/1544

### :+1: Definition of Done
Added `infra-proxy-service` a separate service in order to manage chef infra servers endpoints.

### :athletic_shoe: How to Build and Test the Change
1. `build components/infra-proxy-service`
2. `rebuild components/automate-gateway`
3. `rebuild components/automate-deployment && start_all_services`
4. run `chef-automate dev grpcurl automate-gateway list chef.automate.api.infra_proxy.InfraProxy` in order to get list of implemeted gRPC APIs.

### :white_check_mark: Checklist

- [ ] Tests added/updated?
- [ ] Docs added/updated?

Attached DB structure for reference:

chef_infra_proxy_db

Issue #1544, opened 2019-09-10
Automate Infra Views [read-only]


btm:
# Automate Infra Views
####  (Chef Manage in Automate)

Automate is the observability platform for Chef Infra, but only provides the outcome of Chef Infra client runs. If there was a failure, I currently need to leave Automate and use other tools to find the contents of the [Policy](https://docs.chef.io/chef_overview.html#policy) to understand the failure.

To be able to retrieve Policy information, Automate will need to know address information and credentials for one or more Chef Infra Servers. By creating an API that proxies to Chef Infra servers, we can separate the different Chef Infra Server authn/authz design from the Automate authn/authz design. We also retain the option to allow Automate to control the API going forward.

## Motivation

    As a System Administrator,
    I want to have Chef Infra Policy information in Automate,
    so that I can quickly understand the cause of Chef Infra client failures.

## Specification

This initiative covers *read-only* access and views to the Chef Infra Server only, and excepts for later a few cases such as viewing nodes, policyfile information, and users.

#### /infra/servers
get - returns list of known Chef Infra servers
post - add a Chef Infra server record
patch - update a Chef Infra server record
delete - remove a Chef Infra server record

fields: ID, name, fqdn/ip

#### /infra/servers/ID/orgs
get - returns list of configured organizations
post - configure an organization
patch - update an organization
delete - remove an organization

fields: id, name, admin_user, admin_key (secrets-service)

#### /infra/servers/ID/api

proxies requests to the identified server using the stored credentials for the organization
e.g. /infra/servers/ID/api/organizations/ORG_ID/cookbooks

### Automate Infra Server Views (UI)

- [X] Feature Flag (Chef Infra Server Views)

##### Pages 
- [X] Manage connected Infra Server organizations
- [X] Manage connected Infra Servers
- [x] View Cookbooks # Phase 2
- [x] View Roles # Phase 2
- [x] View Environments # Phase 3
- [x] View Data Bags # Phase 3
- [ ] View Clients # # Phase 4 - Stretch goal

##### APIs
- [X] Manage connected Infra Server organizations API
- [X] Manage connected Infra Servers API
- [X] Cookbooks API
- [x] Affetecd nodes API(cookbooks, roles, environments, policyfiles)
- [X] Roles API
- [x] Environments API
- [x] Data Bags API
- [x] Policyfiles API
- [x] Clients API

##### IAM
- [x] Limit adding/removing Infra Servers and organizations to Editors
- [x] Allow associating an organization to one or more projects
[...]

PR #2581, opened 2019-12-26
Sachin/automate infra proxy server ui

sachin-msys:
Code changes for introduction of `infra proxy module` in `automate-ui` In order to make a single source of truth for chef infra servers data here

**In this PR following points we have covered:**

1. Created `infra-proxy` module in automate-ui to keep all `infra-proxy` related component separate
2.  Designed `infra proxy`  UI flows and used ` Infra-proxy-service `

Refered Wireframes: https://chef.invisionapp.com/share/PZSLN5TJ8B9#/screens/369724917
For feature, description refer: #1544 

[ Updated Mock Ups Jan 17, 2020 ] - addition of Cookbooks to the InVision prototype
https://chef.invisionapp.com/share/3JVLVFMASDZ#/401258846_V0_-_Client_Runs_-_Cookbooks_-_Details_-_Description
Note: The side nav may different because ideally would love to aggregate the cookbooks from all orgs.

### :+1: Definition of Done
I have added infra proxy module and UI new pages
1. chef server's list page
2. create chef-server popup modal

### :athletic_shoe: How to Build 
- build components/infra-proxy-service
- rebuild components/automate-gateway
- rebuild components/automate-deployment
- build components/automate-ui

NOTE: This PR is dependent on PR https://github.com/chef/automate/pull/2173 so build steps are depending on the same for ` Infra-proxy-service `


components/automate-ui/src/app/modules/infra-proxy/chef-servers-list/chef-servers-list.component.ts
```diff
@@ -0,0 +1,135 @@
...
+    private fb: FormBuilder,
+    private layoutFacade: LayoutFacadeService
+  ) {
+    this.loading$ = store.pipe(select(getStatus), map(loading));
+
+    this.sortedChefServers$ = store.pipe(
+      select(allServers),
+      map(servers => ChefSorters.naturalSort(servers, 'name')));
+
+    this.createChefServerForm = this.fb.group({
```
msorens: Please add a comment after this line like this:
https://github.com/chef/automate/blob/fd54bb3f252d715e44a877daab48c52ffb554e4e/components/automate-ui/src/app/modules/token/token-list/api-token-list.component.ts#L72-L73

... using your file name, create-chef-server-modal.component.html

components/automate-ui/src/app/helpers/auth/regex.ts
```diff
@@ -20,7 +20,13 @@ export class Regex {
...
 
     // Allows all special characters except colon :
-    NO_MIXED_WILDCARD_ALLOW_SPECIAL: '^(\\*|[^:*]+)$'
+    NO_MIXED_WILDCARD_ALLOW_SPECIAL: '^(\\*|[^:*]+)$',
+
+    // Allows valid FQDN only
+    VALID_FQDN: '(https?://)?([\\da-z.-]+)\\.([a-z.]{2,6})[/\\w .-]*/?',
+
+    // Allows valid IP Address only (ipv4)
+    VALID_IP_ADDRESS: '^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$'
```
msorens: If you copied these from a published source it is important to add a comment noting the source.
If you created these from scratch, I think it would be really good to add tests (at least a couple dozen for each) that show what does and does not validate. That will let folks who are less familiar with reading regex's be able to judge whether it meets product needs.

Just a few problem examples off-hand for FQDN:

These should match but do not:
```
    http://fo_o.bar/
    https://bit.ly/2OWGwiL
```
These should not match but do:
```
    http://foo.bar com/
    http://foo.bar..com/
    http://...foo.bar.com/
    http://foo.bar-.-.
```

Probably better to find a reliable, documented source rather than create your own to avoid such problem areas. 😁

components/automate-ui/src/app/modules/infra-proxy/chef-servers-list/chef-servers-list.component.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(ChefServersListComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
```
msorens: Please add some tests. The functionality of the underlying component is simple, yes, but should still be tested. You should be able to adapt existing tests for other components that do very similar things.

components/automate-ui/src/app/helpers/auth/regex.ts
sachin-msys: Hello @msorens 
I have rewrote the ```REGX``` for FQDN but still not able to create a perfect one so can I get a help on the same please.

jonong1972:
I updated the description with a link to the Cookbooks section. We can discuss in next meeting of the small things.

components/automate-ui/src/app/helpers/auth/regex.ts
msorens: I did a bit of web searching and found a couple possibilities but because of the nature of an FQDN there does not seem to be a consensus of a good matching regex. So my suggestion is let's back up; you should chat with whoever is requesting this work (is that @jonong1972 ?) to find out what types of FQDN's are needed. It may be that for chef servers we can more precisely specify a URL.

msorens: I found that we use this in a couple spots in the code base:
https://github.com/chef/automate/blob/f8ca770fc9c7915a9658855d5130811e5875eae1/components/automate-ui/src/app/pages/notification-form/notification-form.component.html#L97
https://github.com/chef/automate/blob/f8ca770fc9c7915a9658855d5130811e5875eae1/components/automate-ui/src/app/pages/data-feed-form/data-feed-form.component.html#L37
which, in English means: _either `http://` or `https://` followed by something-dot-something._
It is a very forgiving pattern--it will not exclude any valid ones, which is important--though it will certainly allow invalid ones.

@jonong1972 Do you think that pattern would be sufficient here?

sachin-msys: @msorens created followed regx which is allowing most of valid ones and blocking most of invalid one 

`VALID_FQDN: /^(http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?[a-zA-Z0-9_]+([\-\.]{1}[a-zA-Z0-9]+)*\.[a-zA-Z]{2,5}(:[0-9]{1,5})?(\/.*)?$/,`

tried and added test cases for the examples which you have provided 
Valid Inputs
```
http://fo_o.bar/
https://bit.ly/2OWGwiL
```

Invalid Inputs

```
 http://foo.bar com/
 http://foo.bar..com/
 http://...foo.bar.com/
 http://foo.bar-.-.
```

working well with above all can we go with this ?

msorens: I want to get some product or UX guidance on this question--I will try to speak to @jonong1972 about this tomorrow.

vjeffrey:
yesterday we discussed trying to get this pr merged asap to get some bits merged and try to move towards a more iterative approach.
it looks like there's a conflicting file blocking this.
if there are more unanswered questions/feedback, we can create issues to address those next to support moving forward.
please let me know if you need anything from me to get that done!

components/automate-ui/src/app/helpers/auth/regex.ts
susanev: @sachin-msys lets not do any validation on this input right now, we can come back and iterate later after this work is out of discovery.

sachin-msys: @susanev yes sure sounds good, for now resolved conflict just now please have a look thanks

components/automate-ui/src/app/modules/infra-proxy/chef-servers-list/chef-servers-list.component.ts
```diff
@@ -0,0 +1,136 @@
...
+  private resetCreateModal(): void {
+    this.creatingChefServer = false;
+    this.createChefServerForm.reset();
+    this.conflictErrorEvent.emit(false);
+  }
+
+  public startServerDelete(server: Server): void {
+    this.serverToDelete = server;
+    this.deleteModalVisible = true;
+  }
```
SEAjamieD: related to my comment above, since we are now using `mat-select` instead of `chef-control-menu`, we need to put a guard in place or we'll get two event fires.  Suggest changing this function to look like this:

```
  public startServerDelete($event: ChefKeyboardEvent, server: Server): void {
    if ($event.isUserInput) {
      this.serverToDelete = server;
      this.deleteModalVisible = true;
    }
  }
```

The reason for this is that when using mat-select, if fires off two events, one for a selection and one for a deselection.  `$event.isUserInput = true` is the user selected event.

sachin-msys: @SEAjamieD Thanks for the review. I have incorporated all suggested changes please have a look

SEAjamieD: Thanks for the changes - they look great 👍

components/automate-ui/src/app/modules/infra-proxy/chef-servers-list/chef-servers-list.component.spec.ts
msorens: Nice tests; thank you!

components/automate-ui/src/app/modules/infra-proxy/create-chef-server-modal/create-chef-server-modal.component.ts
```diff
@@ -0,0 +1,44 @@
+import { Component, EventEmitter, Input, Output, OnInit } from '@angular/core';
+import { FormGroup } from '@angular/forms';
+
+@Component({
+  selector: 'app-create-chef-server-modal',
+  templateUrl: './create-chef-server-modal.component.html',
+  styleUrls: ['./create-chef-server-modal.component.scss']
+})
+export class CreateChefServerModalComponent implements OnInit {
```
msorens: _Comment for future work--not for now:_

This code follows our existing patterns fine. But we recently introduced a new, cleaner pattern that I would like you and your team to be aware of for modals--see https://github.com/chef/automate/pull/2704. It encapsulates the code for the modal in a much cleaner way.

Again, no need to apply to this PR; sounds like we want to get this merged ASAP. But please take a look in case of future modals.

PR #1172, opened 2019-08-06
allow underscore IDs in IAM resources

srenatus:
### :nut_and_bolt: Description: What code changed, and why?

Allow underscore char in the IDs of

- projects
- rules
- policies
- tokens
- roles
- teams

### :chains: Related Resources

- #800 

### :+1: Definition of Done

- underscores are allowed in the aforementioned entities' IDs

### :athletic_shoe: How to Build and Test the Change

- Graphically, start the stack including the UI,
- `rebuild components/authn-service` (tokens)
- `rebuild components/authz-service` (policies, projects, rules, roles)
- `rebuild components/teams-service` (teams)
- and attempt to create a token using different (in)valid IDs:

Screenshot 2019-08-06 at 15 32 13

Finally, `o_0` is a valid token ID!

...and `one_two` a valid project ID,

Screenshot 2019-08-06 at 15 55 51

For projects, use an API request to add one with an underscore in its ID:
```
[17][default:/src:113]# echo '{"id":"new_id", "name":"new policy", "projects":["one_two"], "statements": [ {"effect": "ALLOW", "actions": [ "iam:*" ], "projects": [ "one_two" ]}] }' | chef-automate dev grpcurl authz-service -- -d @ chef.automate.domain.authz.v2.Policies.CreatePolicy
{
  "name": "new policy",
  "id": "new_id",
  "type": "CUSTOM",
  "statements": [
    {
      "resources": [
        "*"
      ],
      "actions": [
        "iam:*"
      ],
      "projects": [
        "one_two"
      ]
    }
  ],
  "projects": [
    "one_two"
  ]
}
```
Screenshot 2019-08-06 at 16 11 56

...likewise for roles:
```
[18][default:/src:130]# chef-automate dev grpcurl authz-service --  -d "$(jq -n '{id: "foo_bar", name: "Foo", projects: ["one_two"], actions: ["svc:type:action"]}')" chef.automate.domain.authz.v2.Policies.CreateRole
{
  "name": "Foo",
  "id": "foo_bar",
  "type": "CUSTOM",
  "actions": [
    "svc:type:action"
  ],
  "projects": [
    "one_two"
  ]
}
```
Screenshot 2019-08-06 at 16 12 13

### :white_check_mark: Checklist

- [X] Tests added/updated?
- [x] Docs added/updated?

PR #2582, opened 2019-12-27
[automate-2570] preserve underscores in auto-generated IAM ids

msorens:

**Note: I was so tempted to make this the smallest PR on record -- a single character -- but I would not be able to face myself in the morning if I did not apply TDD and do the tests first. Sigh.**

Back in PR #1172 (allow underscore IDs in IAM resources) we expanded the allowable character set for IAM IDs to include an underscore character. However, one spot that was missed was the auto-generated IDs in the UI, specifically in the `CreateObjectModalComponent` that is (at present) used for creating new teams, tokens, and projects.

PR #1172 (allow underscore IDs in IAM resources)

### :+1: Definition of Done
When entering a name for a team, token, or project in the UI, the auto-generated ID that appears preserves any underscores that are typed (along with hyphens, digits, and alphabetic characters).

Rebuild automate-ui.
Navigate to `Settings >> Teams >> Create Team`.

Type a value for the name. As you type the ID is auto-generated, character for character, with non-supported characters being replaced by hyphens. But now underscores will be preserved.

image

Similar fix works for tokens and projects as well.

### DIFF
--- a/components/automate-ui/src/app/helpers/auth/id-mapper.spec.ts
+++ b/components/automate-ui/src/app/helpers/auth/id-mapper.spec.ts
@@ -4,11 +4,15 @@ import { IdMapper } from './id-mapper';
 describe('IdMapper', () => {
 
     using([
-      ['my-name', 'my-name', 'mirrors nominal input'],
-      ['-0a5l9z-', '-0a5l9z-', 'supports numbers, letters, and hyphens'],
-      ['my-name!', 'my-name-', 'replaces trailing out-of-range char with hyphen'],
-      ['#$my-name', '--my-name', 'replaces leading out-of-range chars with hyphens'],
-      ['my name', 'my-name', 'replaces space char with hyphen'],
+      ['-0a5l9z-', '-0a5l9z-', 'preserves numbers, lowercase letters, and hyphens'],
+      ['my_name__', 'my_name__', 'preserves underscores'],
+      ['my-name!', 'my-name-', 'replaces trailing unsupported single char with hyphen'],
+      ['my-name$/', 'my-name--', 'replaces trailing unsupported multiple chars with hyphen'],
+      ['my^%<>name', 'my----name', 'replaces middle unsupported chars with hyphen'],
+      ['#my-name', '-my-name', 'replaces leading unsupported single char with hyphens'],
+      [';=@my-name', '---my-name', 'replaces leading unsupported multiple chars with hyphens'],
+      [' my_name  ', '-my_name--', 'replaces leading/trailing spaces with hyphens'],
+      ['my   name', 'my---name', 'replaces middle spaces with hyphens'],
       ['My-Name', 'my-name', 'maps upper case to lower case']
     ], function (inputVal: string, outputVal: string, desc: string) {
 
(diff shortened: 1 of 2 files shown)


components/automate-ui/src/app/helpers/auth/id-mapper.ts
```diff
@@ -1,6 +1,6 @@
 export class IdMapper {
 
   public static transform(input: string): string {
-    return input.toLowerCase().replace(/[^a-z0-9]/g, '-');
+    return input.toLowerCase().replace(/[^a-z0-9_]/g, '-');
```
msorens: This IdMapper is used by `CreateObjectModalComponent` to generate the ID corresponding to the name as it is typed by the user.
See: https://github.com/chef/automate/blob/16316b40e8d9103a3157b66b6e96301c9a0b1780/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts#L68-L74

components/automate-ui/src/app/helpers/auth/id-mapper.spec.ts
```diff
@@ -4,11 +4,15 @@ import { IdMapper } from './id-mapper';
...
-      ['#$my-name', '--my-name', 'replaces leading out-of-range chars with hyphens'],
-      ['my name', 'my-name', 'replaces space char with hyphen'],
+      ['-0a5l9z-', '-0a5l9z-', 'preserves numbers, lowercase letters, and hyphens'],
+      ['my_name__', 'my_name__', 'preserves underscores'],
+      ['my-name!', 'my-name-', 'replaces trailing unsupported single char with hyphen'],
+      ['my-name$/', 'my-name--', 'replaces trailing unsupported multiple chars with hyphen'],
+      ['my^%<>name', 'my----name', 'replaces middle unsupported chars with hyphen'],
+      ['#my-name', '-my-name', 'replaces leading unsupported single char with hyphens'],
+      [';=@my-name', '---my-name', 'replaces leading unsupported multiple chars with hyphens'],
+      [' my_name  ', '-my_name--', 'replaces leading/trailing spaces with hyphens'],
```
msorens: Not the job of the IDMapper to trim spaces, so testing it works properly for that case even though calling functions should be doing trimming--see below.

components/automate-ui/src/app/helpers/auth/id-mapper.spec.ts
```diff
@@ -4,11 +4,15 @@ import { IdMapper } from './id-mapper';
 describe('IdMapper', () => {
 
     using([
-      ['my-name', 'my-name', 'mirrors nominal input'],
-      ['-0a5l9z-', '-0a5l9z-', 'supports numbers, letters, and hyphens'],
-      ['my-name!', 'my-name-', 'replaces trailing out-of-range char with hyphen'],
-      ['#$my-name', '--my-name', 'replaces leading out-of-range chars with hyphens'],
-      ['my name', 'my-name', 'replaces space char with hyphen'],
+      ['-0a5l9z-', '-0a5l9z-', 'preserves numbers, lowercase letters, and hyphens'],
+      ['my_name__', 'my_name__', 'preserves underscores'],
```
msorens: This is the new test case that started red and is now green.
(Other test cases were added/changed here to increase the depth of coverage.)

Issue #2255, opened 2019-11-18
specifying a control filter without a profile filter does not filter the exported report


vjeffrey:
## Describe the bug
When exporting the reports, all search filters should be respected.
Specifying a profile and control filter results in an exported report that only has the specified control.
However, specifying a control filter *without* the profile filter does not filter down to only that control.

We used to the profile and control filtering manually, but changed that with deep filtering. Since a single control filter does not trigger deep filtering, we'll need to do manual filters when a control id but no profile id filter is selected.
While we're at it, we should probably take a look at all those "only one profile/one control" filter rules. Are they really necessary? Could we allow the users to specify more than one? We've had questions about this.

## To Reproduce
- add a filter for a profile and control
- export the report, notice only that one control is included
- remove the profile filter, leave the control filter in place
- export the report, notice all controls are included

## Expected behavior
Specifying a control filter exports a report with only that single control

## Versions (please complete the following information):
- Automate Build Number: 20191112230549

phiggins:
tshirt size medium

PR #2583, opened 2019-12-27
specifying multiple control filters on export returns only those controls

vjeffrey:

This code change addresses two issues:
- when a single control filter is applied on export (and no profile filter is included), the report should be filtered by that control (only that control is returned)
- allow filtering by multiple controls

https://github.com/chef/automate/issues/2255

### :+1: Definition of Done
filtering by a single control filter returns a report with only that control
filtering by multiple control filters returns reports with only those controls included

rebuild compliance
filter by a single control, no profile
expect exported report/viewed report to only include that one control
filter by a multiple controls, no profile
expect exported report/viewed report to only include specified controls

- [x] Tests added/updated?
- [x] Docs added/updated?

### DIFF
--- a/components/compliance-service/api/tests/export_test.go
+++ b/components/compliance-service/api/tests/export_test.go
@@ -282,7 +282,7 @@ func TestJSONExportWithControlFilter(t *testing.T) {
 	assert.Equal(t, "nginx-02", report.GetProfiles()[0].GetControls()[0].GetId())
 }
 
-func TestJSONExportWithTwoControlFiltersReturnsError(t *testing.T) {
+func TestJSONExportWithTwoControlsOnlyReturnsThoseTwoControls(t *testing.T) {
 	// get reporting client
 	conn, err := getClientConn()
 	require.NoError(t, err)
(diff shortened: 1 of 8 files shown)


vjeffrey:
TODO: test it out in dev env

components/compliance-service/reporting/relaxting/reports.go
```diff
@@ -507,13 +507,28 @@ func (backend *ES2Backend) GetReport(esIndex string, reportId string,
...
+
 func convertControl(profileControlsMap map[string]*reportingapi.Control, reportControlMin ESInSpecReportControl, filters map[string][]string) *reportingapi.Control {
 	profileControl := profileControlsMap[reportControlMin.ID]
 
 	if profileControl == nil {
 		return nil
 	}
 
+	if len(filters["control"]) > 0 {
+		if !arrayIncludes(filters["control"], profileControl.Id) {
```
stevendanna: minor, but if you want, I think this is the same as:

```suggestion
		if !stringutils.SliceContains(filters["control"], profileControl.Id) {
```

which would allow you to remove lines 510-518 above.

vjeffrey: ah, thank you, yes!

components/compliance-service/api/tests/export_test.go
```diff
@@ -306,10 +306,17 @@ func TestJSONExportWithTwoControlFiltersReturnsError(t *testing.T) {
 	stream, err := reporting.Export(context.Background(), &profileFilterQuery)
 	require.NoError(t, err)
 
-	_, err = stream.Recv()
+	reports, err := getJSONReportsFromStream(stream)
+	require.NoError(t, err)
 
-	grpctest.AssertCode(t, codes.InvalidArgument, err)
-	assert.Equal(t, "rpc error: code = InvalidArgument desc = Invalid: Only one 'control' filter is allowed", err.Error())
+	report := reports[0]
```
phiggins: We probably want a `require.Len` above this, otherwise if this returns 0 results it'll crash the test run.

components/compliance-service/api/tests/export_test.go
```diff
@@ -306,10 +306,17 @@ func TestJSONExportWithTwoControlFiltersReturnsError(t *testing.T) {
...
+	assert.Equal(t, "bb93e1b2-36d6-439e-ac70-cccccccccc04", report.GetId())
+	assert.Equal(t, 3, len(report.GetProfiles()))
```
phiggins: Both of these len checks should probably also be require.

Issue #2019, opened 2019-10-28
document external manager proto for api docs


vjeffrey:
## User Story

In order to properly display api documentation for the manager apis, we need to document the externally exposed protobuf file(s).
For reference, please see: https://github.com/chef/automate/pull/1412
 
## Definition of Done
Protobuf file in https://github.com/chef/automate/tree/master/components/automate-gateway/api/nodes/manager is well documented, checked for content/accuracy/formatting in local docs build.

vjeffrey:
WIP here: https://github.com/chef/automate/compare/vj/manager-proto

PR #2584, opened 2019-12-27
document the node manager proto for api docs

vjeffrey:
Add documentation for the node manager proto (for api docs)

https://github.com/chef/automate/issues/2019

### :+1: Definition of Done
manager proto is well documented

pull the branch down
run `make serve` from components/automate-chef-io
visit localhost:1313/docs/api
look at the docs!


vjeffrey:
TODO: add examples

components/automate-gateway/api/nodes/manager/manager.proto
```diff
@@ -56,6 +89,13 @@ service NodeManagerService {
 			action: "infra:nodeManagers:update"
 		};
 	};
+
+	/*
+	Delete a nodemanager
+
+	Delete a nodemanager given an id. Note this only deletes the manager itself. Any nodes
+	associated with the manager will be abandoned.
```
msorens: Would "orphaned" be better than "abandoned" ? Not familiar enough with the ramifications.

components/automate-gateway/api/nodes/manager/manager.proto
```diff
@@ -82,6 +128,12 @@ service NodeManagerService {
 			action: "infra:nodeManagers:delete"
 		};
 	};
+
+	/*
+	Delete a nodemanager and set nodes to have a state of stopped
+
+	Delete a nodemanager and update all associated nodes to have a state of stopped.
```
msorens: Consider delineating keywords--such as "stopped" here--with a style; presumably, there is a convention for that.

components/automate-gateway/api/nodes/manager/manager.proto
```diff
@@ -108,6 +166,25 @@ service NodeManagerService {
...
 		};
 	};
+
+	/*
+	List of nodemanagers
+
+	Returns a list of nodemanagers matching the query.
+	Supports filtering, sorting, and pagination.
+	Valid filtering fields: 'manager_type'
+	Valid sorting fields: name, type, status, status_message, date_added
```
msorens: Should use consistent quoting between filtering and sorting fields (either both with quotes or both without).
Also, the above 4 lines are rendered as a paragraph. Those two "Valid..." sections in particular read better with line breaks, though.

components/automate-gateway/api/nodes/manager/manager.proto
```diff
@@ -122,6 +199,23 @@ service NodeManagerService {
 			action: "infra:nodeManagers:list"
 		};
 	};
+
+	/*
+	Search node fields
+
+	Searches the available values for a given field and nodemanager id.
```
msorens: Not sure "values" is a useful word there; might "nodes" be more appropriate (to mirror what "Search Nodes" uses) ?

components/automate-gateway/api/nodes/manager/manager.proto
```diff
@@ -150,6 +259,13 @@ service NodeManagerService {
 			action: "infra:nodeManagers:list"
 		};
 	};
+
+	/*
+	Connect
+
+	Attempts to reach the API for the given nodemanager id to check validity of the 
+	credentials associated with the nodemanager.
```
msorens: s/check validity of the credentials/validate the credentials/

components/automate-gateway/api/nodes/manager/manager.proto
```diff
@@ -122,6 +199,23 @@ service NodeManagerService {
...
+	Possible fields: regions, tags, name, subscription_id
+
+	Example:
+	```
+	{
+		"query": {
+			"filter_map":[]
+		},
+		"field": "name"
+	}
```
phiggins: It might be helpful to describe what these example queries are in prose. Does this return all available names? How would adding something to the filter_map affect the results?

components/automate-gateway/api/nodes/manager/manager.proto
```diff
@@ -136,6 +230,21 @@ service NodeManagerService {
...
+
+	Searches the available nodes for a given nodemanager id.
+
+	Example:
+	```
+	{
+		"query": {
+			"filter_map":[]
+		}	
+	}
```
phiggins: Are the empty parameters necessary? Does this do something different than not providing any parameters at all?

components/automate-gateway/api/nodes/manager/manager.proto
```diff
@@ -169,62 +285,91 @@ service NodeManagerService {
...
+	// List of available fields matching the requested criteria.
 	repeated string fields = 1;
 }
 
 message FieldQuery {
-		Query query = 20;
-		string field = 21;
-		string node_manager_id = 22;
+	// Query details (filters) to be applied to the results.
+	Query query = 20;
```
phiggins: A single, non-repeated sub message like this will have its description overridden by the description for where the message is defined. You'll need to move this comment here to have it show up in the UI: https://github.com/chef/automate/pull/2584/files#diff-4c4ab876caf413fdc04f835d615ccabdR296

Additionally, this should be changed to `// Intentionally blank.` to match our convention. https://github.com/chef/automate/blob/master/components/automate-chef-io/README.md#message-documentation

components/automate-gateway/api/nodes/manager/manager.proto
```diff
@@ -169,62 +285,91 @@ service NodeManagerService {
...
+	repeated string nodes = 1;
+	// Total count of node names matching the request.
+	int32 total = 20;
 }
 
 message NodeQuery {
-		Query query = 20;
-		string node_manager_id = 21;
+	// Query details (filters) to be applied to the results.
+	Query query = 20;
```
phiggins: The same comment applies here: https://github.com/chef/automate/pull/2584/files#r362128435

components/automate-gateway/api/nodes/manager/manager.proto
```diff
@@ -169,62 +285,91 @@ service NodeManagerService {
...
+	string name = 2;
+	// Type of nodemanager (aws-ec2, azure-vm, aws-api, azure-api, gcp).
+	string type = 3;
+	// UUID of credential containing the information to connect to aws, azure, or gcp.
+	string credential_id = 20;
+	// List of tag and credential uuid associations to make. These are ssh, winrm, and sudo creds used to access instances.
+	repeated CredentialsByTags instance_credentials = 21;
+	// Status of the nodemanager (reachable, unreachable).
+	string status = 22;
+	// Account id associated with the nodemanager.
```
phiggins: This is probably ok as a placeholder but do we know more about this? What account is this id? Is this something we get from aws or something?

components/automate-gateway/api/nodes/manager/manager.proto
susanev: could we say what abandoned means instead of using a different word?

components/automate-gateway/api/nodes/manager/manager.proto
vjeffrey: aws account id, azure tenant id, or gcp account id

components/automate-gateway/api/nodes/manager/manager.proto
vjeffrey: good idea!

kagarmoe:
Please examine the changes in this commit carefully. I used the option fields for example values and required fields. Example values provided through options overwrite values provided in the message, which (particularly for the `message Query` on line 265) might not be what we desire. I added `required` fields without executing curls, which means that these are at best educated guesses.

api/external/nodes/manager/manager.proto
```diff
@@ -56,6 +77,13 @@ service NodeManagerService {
 			action: "infra:nodeManagers:update"
 		};
 	};
+
+	/*
+	Delete a Node Manager
+
+	Delete a single node manager. This deletes the node manager itself and
+	re-assigneds its associated nodes to the Automate node manager.
```
phiggins:
```suggestion
	reassigns its associated nodes to the Automate node manager.
```

components/automate-chef-io/data/docs/api-static/03-tags.swagger.json
```diff
@@ -9,8 +9,8 @@
       "tags": ["ConfigMgmt"]
     },
     {
-      "name": "Nodes",
-      "tags": ["NodesService"]
+      "name": "Node Management Service",
```
phiggins: The other tags don't include "Service" in their names, should this be just `Node Management`?

api/external/nodes/manager/manager.proto
```diff
@@ -42,6 +56,13 @@ service NodeManagerService {
 			action: "infra:nodeManagers:get"
 		};
 	};
+
+	/*
+	Update a Node Manager
+
+	Update a node manager's metadata, such as its name, associated credential id, or data.
+	This is a PUT operation and it overwrites ALL of the existing node manager metadata. Include all fields, because a PUT operation overwrites any missing fields to empty ("").
```
phiggins: The last two sentences are kind of redundant, they could probably be combined into one.

api/external/nodes/manager/manager.proto
```diff
@@ -42,6 +56,13 @@ service NodeManagerService {
...
+	*/	
 	rpc Update(NodeManager) returns (google.protobuf.Empty) {
```
phiggins: Screen Shot 2020-02-14 at 11 02 04 AM

Not really related to your PR, but the output in the UI when an endpoint returns `google.protobuf.Empty` looks strange.

api/external/nodes/manager/manager.proto
```diff
@@ -82,6 +116,12 @@ service NodeManagerService {
 			action: "infra:nodeManagers:delete"
 		};
 	};
+
+	/*
+	Delete a Node Manager and Stop Nodes
+
+	Delete a node manager and update its associated nodes to `stopped`.
```
phiggins: Should this include the name of the field that gets updated?

api/external/nodes/manager/manager.proto
```diff
@@ -169,62 +246,141 @@ service NodeManagerService {
...
 }
 
 message Query {
+	option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+		json_schema: {
+			required: ["filter_map"]
+			}
+		example: {value: '{"filter_map": [{"key": "manager_type","values":["aws-ec2"]}],"sort": "date_added"}'}
+	};
+	// Filters for the query: "manager_type".
```
phiggins: I'm not sure what this comment is supposed to convey. Is `manager_type` the only valid filter?

api/external/nodes/manager/manager.proto
```diff
@@ -150,6 +220,13 @@ service NodeManagerService {
...
 		};
 	};
+
+	/*
+	Connect
+
+	Attempts to reach the API for the given nodemanager id to validate the 
+	credentials associated with the nodemanager.
+	*/
 	rpc Connect(Id) returns (ConnectResponse) {
```
phiggins: I'm not sure if there's anything we can do, but some text for the response here would be nice like "If the credentials are bad you'll get an error, an empty response means its fine." It currently looks like this:

Screen Shot 2020-02-14 at 11 11 11 AM

susanev:

api/external/nodes/manager/manager.proto
```diff
@@ -169,62 +246,141 @@ service NodeManagerService {
...
+	// you might want to start on 50.
 	int32 page = 23;
+	// Number of results on each page.
 	int32 per_page = 24;
 }
 
 message Fields {
+	option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+		json_schema: {
+			required: ["fields"]
```
vjeffrey: this is a response message -- i'm not sure "required" fields make sense for response messages?

api/external/nodes/manager/manager.proto
```diff
@@ -169,62 +246,141 @@ service NodeManagerService {
...
 }
 
 message Nodes {
-		repeated string nodes = 1;
-		int32 total = 20;
+	option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+		json_schema: {
+			required: ["nodes"]
+		}
+	};
```
vjeffrey: same question here, this is a response message

api/external/nodes/manager/manager.proto
```diff
@@ -169,62 +246,141 @@ service NodeManagerService {
...
+	// Total count of node names matching the request.
+	int32 total = 20;
 }
 
 message NodeQuery {
-		Query query = 20;
-		string node_manager_id = 21;
+	option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+		json_schema: {
+			required: ["node_manager_id"]
```
vjeffrey: :+1:

api/external/nodes/manager/manager.proto
```diff
@@ -169,62 +246,141 @@ service NodeManagerService {
...
-		string credential_id = 20; // aws or azure cred id
-		repeated CredentialsByTags instance_credentials = 21; // ssh, winrm creds
-		string status = 22;
-		string account_id = 24;
-		google.protobuf.Timestamp date_added = 25;
-		repeated chef.automate.domain.compliance.api.common.Kv credential_data = 26;
+	option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+		json_schema: {
+			required: ["id","credential_id","credential_data","type"]
+		}
```
vjeffrey: hmmm. id and type are required, and then we need *either* credential id or credential data. if we receive both, we will override the data in credential id with the data found in credential data

api/external/nodes/manager/manager.proto
```diff
@@ -169,62 +246,141 @@ service NodeManagerService {
...
+	// Date the nodemanager was created.
+	google.protobuf.Timestamp date_added = 25;
+	// Credential data for the nodemanager. This field is used when a credential
+	// has not yet been created, to be able to include credential data (such as AWS_ACCESS_KEY) inline.
+	repeated chef.automate.domain.compliance.api.common.Kv credential_data = 26;
 }
 message CredentialsByTags {
+	option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+		json_schema: {
+			required: ["tag_key","tag_value","credential_ids"]
```
vjeffrey: :+1:

api/external/nodes/manager/manager.proto
```diff
@@ -169,62 +246,141 @@ service NodeManagerService {
...
 }
 message Ids {
+	// List of node manager UUIDs.
 	repeated Id ids = 1;
 }
 
 message Query {
+	option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
+		json_schema: {
+			required: ["filter_map"]
```
vjeffrey: 👍

api/external/nodes/manager/manager.proto
vjeffrey: yupp: https://github.com/chef/automate/blob/master/components/nodemanager-service/pgdb/node-managers.go#L337

components/automate-gateway/api/nodes/manager/manager.proto
kagarmoe: Every key needs a value. This one just says "use an array", which is implied by "map", but every little bit helps.

api/external/nodes/manager/manager.proto
kagarmoe: We try to be alarmingly clear about destructive actions.

api/external/nodes/manager/manager.proto
kagarmoe: It really does!

api/external/nodes/manager/manager.proto
kagarmoe: Yah, I saw that in the code but we don't have a conditional case for marking things as required. We can 1) prefer one over the other and only use that in examples 2) mark both as required, use very clear language to explain the distinction, and only provide examples of one case or  3) mark neither as required but provide an example.

vjeffrey:
i'd like to get this pr merged soon. im guessing theres more work that could be done on it, but right now im facing major troubles trying to get it rebased on master and up to date, and that's only going to get worse with time. maybe we can push more updates to a future pr?

vjeffrey:
i'm gonna open a new pr based off master with these changes b/c the rebase is rather ugly.

vjeffrey:
https://github.com/chef/automate/pull/2975

PR #2585, opened 2019-12-28
[automate-2486] quell unit test noise

msorens:

There are a lot of extraneous console messages (over 1000!!) when you run `make unit` for automate-ui. 
```
WARN: 'The feature name "clientRunsEntity" does not exist in the state, therefore createFeatureSelector cannot access it.  Be sure it is imported in a loaded module using StoreModule.forRoot('clientRunsEntity', ...) or StoreModule.forFeature('clientRunsEntity', ...).  If the default state is intended to be undefined, as is the case with router state, this development-only warning message can be ignored.'
WARN: 'The feature name "notifications" does not exist in the state, therefore createFeatureSelector cannot access it.  Be sure it is imported in a loaded module using StoreModule.forRoot('notifications', ...) or StoreModule.forFeature('notifications', ...).  If the default state is intended to be undefined, as is the case with router state, this development-only warning message can be ignored.'
WARN: 'The feature name "notifications" does not exist in the state, therefore createFeatureSelector cannot access it.  Be sure it is imported in a loaded module using StoreModule.forRoot('notifications', ...) or StoreModule.forFeature('notifications', ...).  If the default state is intended to be undefined, as is the case with router state, this development-only warning message can be ignored.'
WARN: 'The feature name "clientRunsEntity" does not exist in the state, therefore createFeatureSelector cannot access it.  Be sure it is imported in a loaded module using StoreModule.forRoot('clientRunsEntity', ...) or StoreModule.forFeature('clientRunsEntity', ...).  If the default state is intended to be undefined, as is the case with router state, this development-only warning message can be ignored.'
. . .
```
This PR quashes those.

Notes: 
Those warnings are completely non-specific in terms of where they came from. The bulk of the work was identifying the problem files.
I did that by first starting the tests in watch-mode:
```
npm run test:watch
```
Then, start by subdividing the set of *.spec.ts files: switched to a subdirectory (e.g. `entities`) then used this to enable just those tests:
```
sed -i '' 's/^describe/fdescribe/' $(find . -name '*.spec.ts' -print)
```
If any failures in that group, subdivide again.
If all clean, revert the `fdescribe`:
```
sed -i '' 's/fdescribe/describe/' $(find . -name '*.spec.ts' -print)
```

The second challenge was figuring out what the fix was.
Comparing and contrasting files that were working with those that were not led me to the appropriate ngrx/store setup lines.
I readily addressed the issue in 10 of the 14 problem files.
The remaining four files required I get sidetracked for awhile--see PR #2595--and then returning here, I could finish up those last 4 files.

I then went through all uses of the `StoreModule.forRoot`, slicing it this way and that, and shook out a few more files that needed the same fix from PR #2595. Also identified another handful that enumerated state bits for setup yet were not even needed, so I removed all those. Follow the commits to see the logical groupings.

NA

### :+1: Definition of Done
Test output is just this, with no console warnings, and all tests pass:
```
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
................................................................................
...............................................................................
Chrome 79.0.3945 (Mac OS X 10.14.6): Executed 1359 of 1393 (skipped 34) SUCCESS (35.647 secs / 32.769 secs)
```

```
cd automate-ui
make unit
```

### DIFF
--- a/components/automate-ui/src/app/components/notifications/notifications.component.spec.ts
+++ b/components/automate-ui/src/app/components/notifications/notifications.component.spec.ts
@@ -2,7 +2,7 @@ import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 import { MockComponent } from 'ng2-mock-component';
 import { NO_ERRORS_SCHEMA } from '@angular/core';
 import { StoreModule, Store } from '@ngrx/store';
-import { NgrxStateAtom, ngrxReducers, defaultInitialState, runtimeChecks } from 'app/ngrx.reducers';
+import { NgrxStateAtom, ngrxReducers, runtimeChecks } from 'app/ngrx.reducers';
 
 import { ChefNotificationsComponent } from './notifications.component';
 
@@ -18,7 +18,7 @@ describe('ChefNotificationsComponent', () => {
         ChefNotificationsComponent
       ],
       imports: [
-        StoreModule.forRoot(ngrxReducers, { ...defaultInitialState, runtimeChecks })
+        StoreModule.forRoot(ngrxReducers, { runtimeChecks })
       ],
       schemas: [
         NO_ERRORS_SCHEMA
--- a/components/automate-ui/src/app/modules/team/team-details/team-details.component.spec.ts
+++ b/components/automate-ui/src/app/modules/team/team-details/team-details.component.spec.ts
@@ -2,33 +2,15 @@ import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 import { Router } from '@angular/router';
 import { RouterTestingModule } from '@angular/router/testing';
 import { ReactiveFormsModule } from '@angular/forms';
-import { routerReducer } from '@ngrx/router-store';
-import { MockComponent } from 'ng2-mock-component';
 import { StoreModule, Store, Action } from '@ngrx/store';
-
 import * as routerStore from '@ngrx/router-store';
-import { NgrxStateAtom, runtimeChecks } from 'app/ngrx.reducers';
-import {
-  projectsFilterReducer,
-  projectsFilterInitialState
-} from 'app/services/projects-filter/projects-filter.reducer';
-import {
-  policyEntityReducer, PolicyEntityInitialState
-} from 'app/entities/policies/policy.reducer';
-import {
-  projectEntityReducer,
-  ProjectEntityInitialState
-} from 'app/entities/projects/project.reducer';
+import { MockComponent } from 'ng2-mock-component';
+
+import { NgrxStateAtom, ngrxReducers, defaultInitialState, runtimeChecks } from 'app/ngrx.reducers';
+import { FeatureFlagsService } from 'app/services/feature-flags/feature-flags.service';
+import { PolicyEntityInitialState } from 'app/entities/policies/policy.reducer';
 import { Project } from 'app/entities/projects/project.model';
 import { GetProjectsSuccess, GetProjects } from 'app/entities/projects/project.actions';
-import {
-  userEntityReducer,
-  UserEntityInitialState
-} from 'app/entities/users/user.reducer';
-import {
-  teamEntityReducer,
-  TeamEntityInitialState
-} from 'app/entities/teams/team.reducer';
 import {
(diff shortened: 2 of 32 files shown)


components/automate-ui/src/app/modules/token/token-list/api-token-list.component.spec.ts
```diff
@@ -26,7 +28,9 @@ describe('ApiTokenListComponent', () => {
         StoreModule.forRoot({
           apiTokens: apiTokenEntityReducer,
           policies: policyEntityReducer,
-          projectsFilter: projectsFilterReducer
+          projectsFilter: projectsFilterReducer,
+          notifications: notificationEntityReducer, // not used here but needed to suppress warnings
+          clientRunsEntity: clientRunsEntityReducer // not used here but needed to suppress warnings
```
msorens: The specific cause of all the console warnings was that these two properties were missing from the state in 14 test files. This fix was used in the first 10 files.

components/automate-ui/src/app/modules/team/team-details/team-details.component.spec.ts
```diff
@@ -112,14 +91,7 @@ describe('TeamDetailsComponent v2', () => {
...
         RouterTestingModule,
-        StoreModule.forRoot({
-          router: routerReducer,
-          teams: teamEntityReducer,
-          users: userEntityReducer,
-          policies: policyEntityReducer,
-          projects: projectEntityReducer,
-          projectsFilter: projectsFilterReducer
-        }, { initialState, runtimeChecks })
+        StoreModule.forRoot(ngrxReducers, { initialState, runtimeChecks })
```
msorens: By the time I fixed the two bugs that prevented further progress here (in the second PR I mentioned), I realized a cleaner way was to get rid of the itemized properties and just use the globally defined `defaultInitialState`, which already includes the two needed lines (for client runs and notifications). Here it is using a customized version of that global state (`initialState`).

components/automate-ui/src/app/components/notifications/notifications.component.spec.ts
```diff
@@ -18,7 +18,7 @@ describe('ChefNotificationsComponent', () => {
         ChefNotificationsComponent
       ],
       imports: [
-        StoreModule.forRoot(ngrxReducers, { ...defaultInitialState, runtimeChecks })
+        StoreModule.forRoot(ngrxReducers, { runtimeChecks })
```
msorens: As mentioned in the other PR, applying spread to the initial state here is wrong... but also it is not even needed, so just removed it in all the files in this commit.

components/automate-ui/src/app/modules/team/team-details/team-details.component.spec.ts
```diff
@@ -65,21 +73,19 @@ describe('TeamDetailsComponent v2', () => {
...
-      },
-      navigationId: 0 // what's that zero?
+        params: { id: targetId }
+      }
     },
-    policies: v2PolicyEntityInitialState
+    policies: {
+      ...PolicyEntityInitialState, iamMajorVersion: 'v2' as IAMMajorVersion
+    }
   };
```
msorens: Removed more duplication for a few files in this commit, which also has the side benefit of showing just the fields that are relevant to the test.

components/automate-ui/src/app/pages/+compliance/+profile/+profile-details/profile-details.component.spec.ts
```diff
@@ -30,30 +30,16 @@ describe('ProfileDetailsComponent', () => {
...
-        url: '/',
-        params: {},
-        queryParams: {},
-        fragment: ''
-      }
-    },
-    clientRunsEntity: fromClientRuns.ClientRunsEntityInitialState,
-    notifications: fromNotifications.InitialState,
-    layout: fromLayout.InitialState
-  };
```
msorens: This initial state is not actually providing any customization. That was a smell that indicated it was not even needed, so this commit removes all those.

Issue #2486, opened 2019-12-11
More Automate-ui unit test noise


msorens:
## Describe the bug

These likely harmless warnings recently started appearing in UI unit test output. Should find what is needed to quell them.
```
WARN: 'The feature name "clientRunsEntity" does not exist in the state, therefore createFeatureSelector cannot access it.  Be sure it is imported in a loaded module using StoreModule.forRoot('clientRunsEntity', ...) or StoreModule.forFeature('clientRunsEntity', ...).  If the default state is intended to be undefined, as is the case with router state, this development-only warning message can be ignored.'
WARN: 'The feature name "notifications" does not exist in the state, therefore createFeatureSelector cannot access it.  Be sure it is imported in a loaded module using StoreModule.forRoot('notifications', ...) or StoreModule.forFeature('notifications', ...).  If the default state is intended to be undefined, as is the case with router state, this development-only warning message can be ignored.'
WARN: 'The feature name "notifications" does not exist in the state, therefore createFeatureSelector cannot access it.  Be sure it is imported in a loaded module using StoreModule.forRoot('notifications', ...) or StoreModule.forFeature('notifications', ...).  If the default state is intended to be undefined, as is the case with router state, this development-only warning message can be ignored.'
.............................
WARN: 'The feature name "clientRunsEntity" does not exist in the state, therefore createFeatureSelector cannot access it.  Be sure it is imported in a loaded module using StoreModule.forRoot('clientRunsEntity', ...) or StoreModule.forFeature('clientRunsEntity', ...).  If the default state is intended to be undefined, as is the case with router state, this development-only warning message can be ignored.'
WARN: 'The feature name "clientRunsEntity" does not exist in the state, therefore createFeatureSelector cannot access it.  Be sure it is imported in a loaded module using StoreModule.forRoot('clientRunsEntity', ...) or StoreModule.forFeature('clientRunsEntity', ...).  If the default state is intended to be undefined, as is the case with router state, this development-only warning message can be ignored.'
WARN: 'The feature name "notifications" does not exist in the state, therefore createFeatureSelector cannot access it.  Be sure it is imported in a loaded module using StoreModule.forRoot('notifications', ...) or StoreModule.forFeature('notifications', ...).  If the default state is intended to be undefined, as is the case with router state, this development-only warning message can be ignored.'
[...]

PR #1485, opened 2019-09-06
Moving the License Banner

Shadae:
### :nut_and_bolt: Description: What code changed, and why?
- Added logic to display notifications on different parts of the page from each other. 
- Move the license expiration banner to the top of the page to align with original designs and prepare for an additional banner on the bottom of the page for projects. 
*Note: This is for the expiring license banner only, all other user notifications should stay where they are.*

### :+1: Definition of Done
- [x] - Move the banner to the very top of the page, all other content should shift down
- [x] - Change the content to say: There are days remaining on your license. Apply a license or contact Chef to continue to use Chef Automate.
- [x] - The x-button should dismiss the banner and shift the rest of the content back up

### :athletic_shoe: How to Build and Test the Change
1.      build components/automate-ui-devproxy/
2.      start_automate_ui_background
3.      start_all_services 
4.      chef-automate dev psql chef_license_control_service -- -c "UPDATE licenses SET active=NULL"
5.  Add a user and see that other types of notifications work and remain at the original spot

### :white_check_mark: Checklist

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### :camera: Screenshots, if applicable

PR #2595, opened 2019-12-31
[Automate-2593] fix unit test ngrx setup

msorens:

Trying to resolve a lot of noise being generated by UI unit tests (see #2486)
was causing compilation errors. It turns out that the new work of #2486 (automate unit test noise) was exposing a hidden bug in the unit test setup that had been recently introduced from #1485 (moving the license banner)--some calls to `StoreModule.forRoot` were passing incorrect arguments.  That bug in turn was masking one more bug: two properties in the `defaultInitialState` used by ngrx store were defined incorrectly.

This PR corrects both of those issues.

#### First bug: calls to `forRoot` using `defaultInitialState`

Here's the signature for [forRoot](https://ngrx.io/api/store/StoreModule):
image

And here is the [RootStoreConfig](https://ngrx.io/api/store/RootStoreConfig) interface:
image

Notice that `RootStoreConfig` has an `initialState` property, which is where the `defaultInitialState` object needs to reside, rather than flatten its properties with the spread operator. Since all properties of `RootStoreConfig` are optional (!), this did not cause any errors (though it did cause me some grief 😁 ).

In a nutshell, then, needed to remove the spread operator and add the `initialState` property:
```
-        StoreModule.forRoot(ngrxReducers, { ...defaultInitialState, runtimeChecks })
+        StoreModule.forRoot(ngrxReducers, { initialState: defaultInitialState, runtimeChecks })
```

#### Second bug: Incorrect `defaultInitialState` definition

With the above fixed, I was then seeing compilation errors:
```
Type '{ router: { state: { url: string; params: { id: string; }; queryParams: {}; fragment: string; path: any[]; }; navigationId: number; }; policies: PolicyEntityState & { iamMajorVersion: string; }; ... 26 more ...; users: { ...; }; }' is not assignable to type 'InitialState<{ credentials: CredentialsState; scanner: ScannerState; router: RouterReducerState<{ url: string; queryParams: {}; params: {}; fragment: string; path: string[]; }>; ... 26 more ...; users: UserEntityState; }>'.
Type '{ router: { state: { url: string; params: { id: string; }; queryParams: {}; fragment: string; path: any[]; }; navigationId: number; }; policies: PolicyEntityState & { iamMajorVersion: string; }; ... 26 more ...; users: { ...; }; }' is not assignable to type 'Partial<{ credentials: CredentialsState; scanner: ScannerState; router: RouterReducerState<{ url: string; queryParams: {}; params: {}; fragment: string; path: string[]; }>; ... 26 more ...; users: UserEntityState; }>'.
Types of property 'users' are incompatible.
Type '{ (): EntityState<User>; <S extends object>(state: S): EntityState<User> & S; }' is missing the following properties from type 'UserEntityState': getStatus, updateStatus, deleteStatus, createStatus, and 2 more.
```

Should probably call those compilation warnings rather than errors, as they did **not** fail the build nor did they affect test results--_all tests passed_!

But the error displays on the console when running `make unit` or `npm run test`.

There were actually two problems:
1. The `router` section was missing the "router" key.
2. The `users` section was given a method name (not a method call), when it should have been given a constant.

The incorrect `router` section did not manifest, showing only that no test relied upon it.

The incorrect `users` value caused the error shown above once the `forRoot` bug was patched.

Here are the crucial changes:

```
-  ...defaultRouterState,
+  router: defaultRouterState,

-  users: userEntity.userEntityAdapter.getInitialState
+  users: userEntity.UserEntityInitialState

```

See above

### :+1: Definition of Done
All tests pass and none of the errors of the type shown above appear.
You will, however, still see many, many occurrences of console warnings that will be fixed in #2486.

```
cd components/automate-ui
make unit
```

### DIFF
--- a/components/automate-ui/src/app/components/settings-sidebar/settings-sidebar.component.spec.ts
+++ b/components/automate-ui/src/app/components/settings-sidebar/settings-sidebar.component.spec.ts
@@ -22,7 +22,7 @@ describe('SettingsSidebarComponent', () => {
   beforeEach(() => {
     TestBed.configureTestingModule({
       imports: [RouterTestingModule,
-        StoreModule.forRoot(ngrxReducers, { ...defaultInitialState, runtimeChecks })
+        StoreModule.forRoot(ngrxReducers, { initialState: defaultInitialState, runtimeChecks })
       ],
       declarations: [
         SettingsSidebarComponent,
--- a/components/automate-ui/src/app/ngrx.reducers.ts
+++ b/components/automate-ui/src/app/ngrx.reducers.ts
@@ -133,8 +133,10 @@ export function routerReducer(state = defaultRouterState, action) {
   }
 }
 
+// Note: this is used ONLY for unit tests, but better to be defined here since
+// it needs to mirror the ngrxReducers.
 export const defaultInitialState = {
-  ...defaultRouterState,
+  router: defaultRouterState,
   credentials: credentials.initialState,
   event_feed: eventFeed.initialState,
   scanner: scanner.initialState,
@@ -166,7 +168,7 @@ export const defaultInitialState = {
   serviceGroups: serviceGroups.ServiceGroupEntityInitialState,
   teams: teamEntity.TeamEntityInitialState,
   userperms: permEntity.initialState,
-  users: userEntity.userEntityAdapter.getInitialState
+  users: userEntity.UserEntityInitialState
 };
 
 export const ngrxReducers = {
--- a/components/automate-ui/src/app/ngrx.reducers.ts
+++ b/components/automate-ui/src/app/ngrx.reducers.ts
@@ -9,7 +9,7 @@ import * as projectsFilter from './services/projects-filter/projects-filter.redu
 import * as apiToken from './entities/api-tokens/api-token.reducer';
 import * as automateSettings from './entities/automate-settings/automate-settings.reducer';
 import * as clientRuns from './entities/client-runs/client-runs.reducer';
-import * as integationsAdd from './pages/integrations/add/integration-add.reducer';
+import * as integrationsAdd from './pages/integrations/add/integration-add.reducer';
 import * as integrationsDetail from './pages/integrations/detail/integrations-detail.reducer';
 import * as integrationsEdit from './pages/integrations/edit/integrations-edit.reducer';
 import * as license from './entities/license/license.reducer';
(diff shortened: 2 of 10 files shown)


components/automate-ui/src/app/components/settings-sidebar/settings-sidebar.component.spec.ts
```diff
@@ -22,7 +22,7 @@ describe('SettingsSidebarComponent', () => {
   beforeEach(() => {
     TestBed.configureTestingModule({
       imports: [RouterTestingModule,
-        StoreModule.forRoot(ngrxReducers, { ...defaultInitialState, runtimeChecks })
+        StoreModule.forRoot(ngrxReducers, { initialState: defaultInitialState, runtimeChecks })
```
msorens: Instance of bug #1 fix

components/automate-ui/src/app/ngrx.reducers.ts
```diff
@@ -133,8 +133,10 @@ export function routerReducer(state = defaultRouterState, action) {
   }
 }
 
+// Note: this is used ONLY for unit tests, but better to be defined here since
+// it needs to mirror the ngrxReducers.
 export const defaultInitialState = {
-  ...defaultRouterState,
+  router: defaultRouterState,
```
msorens: Bug #2 fix for router

components/automate-ui/src/app/ngrx.reducers.ts
```diff
@@ -166,7 +168,7 @@ export const defaultInitialState = {
   serviceGroups: serviceGroups.ServiceGroupEntityInitialState,
   teams: teamEntity.TeamEntityInitialState,
   userperms: permEntity.initialState,
-  users: userEntity.userEntityAdapter.getInitialState
+  users: userEntity.UserEntityInitialState
```
msorens: Bug #2 fix for users

components/automate-ui/src/app/ngrx.reducers.ts
```diff
@@ -9,7 +9,7 @@ import * as projectsFilter from './services/projects-filter/projects-filter.redu
 import * as apiToken from './entities/api-tokens/api-token.reducer';
 import * as automateSettings from './entities/automate-settings/automate-settings.reducer';
 import * as clientRuns from './entities/client-runs/client-runs.reducer';
-import * as integationsAdd from './pages/integrations/add/integration-add.reducer';
+import * as integrationsAdd from './pages/integrations/add/integration-add.reducer';
```
msorens: PR bonus: spelling fix

PR #2598, opened 2020-01-01
Remove possibly dead file with exposed API key.

phiggins:
This file does not appear to be used and has an API key in it. After deleting this file, I started the docs server and did not notice any differences.


phiggins:
It sounds like we still need this file. From #docs-support in slack:

> Deleting that baseof.html  file will make the site use the baseof.html from the chef-hugo-theme. The theme has its search set up for www.chef.sh so it will probably just return search results for the workstation docs but not automate.

and

> In other words, it will probably break search on the site.

PR #2599, opened 2020-01-01
Spellcheck for API docs

phiggins:
A few weeks ago @kagarmoe asked if there was a way to add spellcheck for the API docs. The existing spellcheck script has a bunch of special case code for different use cases and didn't really fit the use case I had, so I added a task in the `automate-chef-io` Makefile for straightforward spellchecking for doc writers. While investigating this, I noticed there were a lot of documentation files in the repo that were not being spellchecked so I changed the original spellcheck script to check those.


components/automate-chef-io/Makefile
```diff
@@ -10,7 +10,6 @@ themes/chef:
 	scripts/clone_hugo.sh
 
 clean:
-	rm -rf site/themes/chef
```
phiggins: This directory doesn't exist and we don't create anything with this path.

components/automate-chef-io/Makefile
```diff
@@ -53,3 +52,13 @@ lint: themes/chef
...
 	hugo
+
+# Consult scripts/spellcheck.sh in the root Automate directory for notes on using cspell.
+spellcheck:
+	@pushd ../.. > /dev/null && \
+		cspell \
+			components/automate-gateway/**/*.proto \
+			components/automate-chef-io/**/*.md \
+			components/automate-chef-io/layouts/**/*.html \
+			components/automate-chef-io/data/docs/api-static/**/* \
```
phiggins: This appeared to be all the files in this directory that contained spellcheck'able prose that was not a file generated elsewhere in the project.

components/automate-chef-io/data/docs/api-static/02-meta-description.yaml
```diff
@@ -26,13 +26,13 @@ info:
...
       "description": "New Token",
-      "value": "bww8EEpr39_eYMnQ2zybtrP9uzk=",
+      "value": "bww8EEpr39_eYMnQ2zyBtrP9uzk=",
       "active": false,
       "created": "2019-12-03T00:15:10Z",
       "updated": "2019-12-03T00:15:10Z"
     }
 
-    export TOKEN2=bww8EEpr39_eYMnQ2zybtrP9uzk=
+    export TOKEN2=bww8EEpr39_eYMnQ2zyBtrP9uzk=
```
phiggins: The spellchecker appears to consider any string of characters in the same case to be a potential word, so it thought `zybtr` was a misspelled word. This value does not appear to be significant, but rather than make up a fake token value I just changed one letter to uppercase to fool the spellchecker.

components/automate-chef-io/layouts/_default/data-api.html
```diff
@@ -18,7 +18,7 @@
     </style>
   </head>
   <body>
-    <redoc spec-url="/api-docs/all-apis.swagger.json"></redoc>
+    <ReDoc spec-url="/api-docs/all-apis.swagger.json"></ReDoc>
```
phiggins: I felt like we might actually want to catch instances of all lowercase `redoc` and HTML is generally case insensitive, so this seemed like the least bad solution to this being a spellcheck failure.

ec2/README.md
```diff
@@ -92,7 +92,7 @@ See the services logs, from studio shell:
 sl
 ```
 
-or from your favourite Linux shell:
+or from your Unix shell:
```
phiggins: Sorry Brits, the spellchecker doesn't like the superfluous "u".

scripts/spellcheck.sh
```diff
@@ -107,7 +107,7 @@ checkFiles() {
 
     if [[ "$CHECK_GLOBAL_DOCS" = "true" ]]; then
         >&2 echo -n "global docs [md]... "
-        run_cspell "dev-docs/**/*.md" "components/automate-chef-io/content/docs/**/*.md" || RETVAL=1
+        run_cspell $(find . -name '*.md' | grep -v node_modules | grep -v "^\(\./cache\|\./components\|\./vendor\)") || RETVAL=1
```
phiggins: It seemed easier to use find and grep than deal with cspell's slow/broken globbing.

msorens: My concern here is potentially hitting the command-line length limit. Though a quick search says it is 256KB, so I guess we're OK.

components/automate-chef-io/Makefile
```diff
@@ -53,3 +52,14 @@ lint: themes/chef
 
 build: themes/chef generate_swagger
 	hugo
+
+# Consult scripts/spellcheck.sh in the root Automate directory for notes on using cspell.
+spellcheck:
```
msorens: Glad to see you putting this in the makefile. It is long been on my "todo" list to eventually wire up spelling failures to fail the build, too.

components/automate-chef-io/data/docs/api-static/02-meta-description.yaml
phiggins: Is this a common format in this project? We could probably also replace this with a regex recognizing base64 values instead. WDYT?

scripts/spellcheck.sh
phiggins: There are not currently too many .md files in the project so this shouldn't cause any problems:

```
Petes-MBP:automate pete$ find . -name '*.md' | grep -v node_modules | grep -v "^\(\./cache\|\./components\|\./vendor\)" | wc -l
      53
```

components/automate-chef-io/data/docs/api-static/02-meta-description.yaml
phiggins: After mentioning a regex to recognize base64 values I went and looked if that was possible. It looks like it is, but none of the examples I saw included `_` like the above value does, so I'm not sure how these values are being generated. I'll use your solution instead.

msorens: Sounds good. Though, I recognize mine is a bit of kludge. 🤷‍♂

phiggins: Apparently the `_` thing is because of golang's implementation of base64:

```
The most common encoding is the "base64" encoding defined in RFC 4648 and used in MIME (RFC 2045) and PEM (RFC 1421). RFC 4648 also defines an alternate encoding, which is the standard encoding with - and _ substituted for + and /. 
```

https://golang.org/pkg/encoding/base64/

Supporting that would not be too difficult, but in a recent commit I made a change that caused more files to be spellchecked and those had failures from fragments of base64 encoded strings, which seems much more of a maintenance burden to support.

PR #2612, opened 2020-01-08
Periodic spellcheck

msorens:

This PR:
1. Updates the codebase to correct spelling errors and add spelling exceptions.
2. Instruments all makefiles with a new `make spell` capability.
3. Back-ported relevant updates from automate-chef-io makefile into spellcheck.sh script.

PR #2599 Spellcheck for API docs

### :+1: Definition of Done
A good chunk of the codebase (including some user-facing text) has been spell-corrected.
Can run `make spell` individually for Automate components (e.g. authz-service, automate-ui, etc.).

Example for automate-ui:
```
 $ cd components/automate-ui
 $ make spell
components/automate-ui [ts and tsx]... CSpell: Files checked: 590, Issues found: 10 in 6 files
components/automate-ui [sh]... CSpell: Files checked: 0, Issues found: 0 in 0 files
components/automate-ui [md]... CSpell: Files checked: 9, Issues found: 1 in 1 files
components/automate-ui [html]... CSpell: Files checked: 154, Issues found: 1 in 1 files
   1 HHMM
   1 Hmmss
   1 Infastructure
   1 YYYYMMDDZ
   1 catagory
   1 destiantions
   1 dont
   1 lkjasdl
   1 reuturned
   1 sarch
   2 invaild
```
The numbers above are counts of that particular misspelling; thus, "invaild" appeared twice while all the others appeared just once.

NA

- [x] I have read the [CONTRIBUTING document](../CONTRIBUTING.md).
- [ ] Tests added/updated?
- [ ] Docs added/updated?

NA


components/authn-service/Makefile
```diff
@@ -63,4 +63,11 @@ kill_docker_pg:
 proto:
 	cd ../../ && hab studio run 'source .studiorc; compile_go_protobuf_component authn-service'
 
+spellcheck:
+	@pushd ../.. > /dev/null; \
+	./scripts/spellcheck.sh --wordsOnly components/authn-service; \
+	EXIT_CODE=$$?; \
+	popd > /dev/null; \
+	exit $$EXIT_CODE
```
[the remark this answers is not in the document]
stevendanna: I wonder if there is a way to put this in Makefile.common_go so we don't have to copy it around as much?

cspell.json
```diff
@@ -244,14 +244,14 @@
...
         "httpd",
-        "httputils",
         "http's",
         "icmp",
         "idempotently",
         "idis",
         "iface",
         "iframes",
         "iirc",
+        "ILIKE",
```
blakestier: why is this all caps?

blakestier: oh nvmd, exceptions. lol

cspell.json
```diff
@@ -917,6 +927,7 @@
         "dkhttp",
         "dlsserver",
         "dmesg",
+        "doesnt",
```
phiggins: I know this probably got added for a reason, but I feel like we probably don't want to excuse instances of this without the apostrophe...

components/authn-service/Makefile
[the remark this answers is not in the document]
phiggins: I agree, I think I'd rather see this added to every project in a standard way rather than through copy-pasting.

cspell.json
SEAjamieD: I agree with this as well.

components/authn-service/Makefile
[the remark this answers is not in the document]
msorens: Figured out the variable machinations to do that and just pushed a commit.

cspell.json
msorens: I will have to agree, too. 🤷‍♂

components/authn-service/Makefile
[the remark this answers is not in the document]
msorens: @phiggins I used `make spell` for the common functionality. But note that now `automate-chef-io` has its own private `make spellcheck` plus this common one `make spell`. What do you think about renaming the private one to be `make spell-plus-protos` so it will be much clearer what the private one is for?

Makefile.common_go
```diff
@@ -2,6 +2,10 @@
 # components/xyz-service/Makefile
 REPOROOT?=../..
 
+# Convert, e.g., "src/github.com/chef/automate/components/teams-service" to "components/teams-service"
+REPO_DIR := $(shell cd $(REPOROOT) && pwd)/
+COMPONENT_DIR := $(shell echo $${PWD\#"$(REPO_DIR)"})
```
msorens: The tricky part here was using make variables as shell variables.
The first line gets the expanded path of the REPOROOT.
The second line then subtracts that prefix from the current directory using the shell variable syntax `${variable-name#prefix-to-strip}`

stevendanna: I wonder if we could use `$(CURDIR)` at line 46 instead of doing this conversion.  This works for me locally though, so dealers choice.

msorens: Thanks for the suggestion! I was not aware of `CURDIR`. That would work, yes, but it has the same (minor) drawback that using just `PWD` above would have; output is more verbose. One would see this:
```
/Users/msorens/code/go/src/github.com/chef/automate/components/teams-service [go]... CSpell: Files checked: 0, Issues found: 0 in 0 files
/Users/msorens/code/go/src/github.com/chef/automate/components/teams-service [rb]... CSpell: Files checked: 0, Issues found: 0 in 0 files
/Users/msorens/code/go/src/github.com/chef/automate/components/teams-service [md]... CSpell: Files checked: 0, Issues found: 0 in 0 files
/Users/msorens/code/go/src/github.com/chef/automate/components/teams-service [proto]... CSpell: Files checked: 0, Issues found: 0 in 0 files
/Users/msorens/code/go/src/github.com/chef/automate/components/teams-service [sh]... CSpell: Files checked: 0, Issues found: 0 in 0 files
```
instead of this:
```
components/teams-service [go]... CSpell: Files checked: 19, Issues found: 0 in 0 files
components/teams-service [rb]... CSpell: Files checked: 0, Issues found: 0 in 0 files
components/teams-service [md]... CSpell: Files checked: 3, Issues found: 0 in 0 files
components/teams-service [proto]... CSpell: Files checked: 0, Issues found: 0 in 0 files
components/teams-service [sh]... CSpell: Files checked: 1, Issues found: 0 in 0 files
```

components/authn-service/Makefile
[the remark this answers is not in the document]
phiggins: I'd be ok renaming, but I think I'd prefer something revealing the why more than the what, eg `make spell-all` or something.

Would it be possible to allow each service to customize what was checked like how the go linter does? Definitely not something you need to figure out today. :)

Issue #2622, opened 2020-01-14
[BETA][EAS] Filters are Misleading/Undocumented


sirajrauff:
Making use of `/v1/applications` as part of the Beta Automate API for EAS and encountered a few bumps trying to get it working.

First and foremost is there's no documentation regarding the endpoints other than in [V1 Policies](https://automate.chef.io/docs/iam-v1-policies/#applications-beta) (there's nothing for V2). This provides the permissions required but nowhere else can pagination or filtering arguments be found, which led to me initially think the endpoint was bugged (#2610). I ended up using the source code [[1](https://github.com/chef/automate/blob/master/components/automate-chef-io/data/docs/api_chef_automate/applications/applications.swagger.json#L70)],[[2](https://github.com/chef/automate/blob/7d821f9ddf957ef9dee7413874710e9a323fe433/components/applications-service/pkg/storage/postgres/services.go#L424)] to get my code working.

Second, the filters seem to correspond more to Compliance than to Habitat Services. Consider this return object:
```
{
  "services": [
    {
      "supervisor_id": "0710a75e58fc460ca5861559deac1d14",
      "release": "sirajrauff-public/sample-node-app/0.1.4/20191218214052",
      "group": "sample-node-app.default",
      "health_check": "OK",
      "status": "RUNNING",
      "application": "sample-node-app",
      "environment": "local-test",
      "fqdn": "2531d46609d9",
      "channel": "stable",
      "update_strategy": "AT-ONCE",
      "site": "laptop",
      "previous_health_check": "NONE",
      "current_health_since": "15 hours 16 minutes",
      "health_updated_at": "2020-01-14T01:05:58.885441Z",
      "disconnected": true,
      "last_event_occurred_at": "2020-01-14T01:07:28.965647Z",
      "last_event_since": "15 hours 14 minutes"
    },
    ...
  ]
}
```
- The `status` filter corresponds to [health check](https://github.com/chef/automate/blob/7d821f9ddf957ef9dee7413874710e9a323fe433/components/applications-service/pkg/storage/postgres/services.go#L428-L429), which is misleading as the object has both `status` and `health_check`. i.e. the first time I saw the `status` filter, I tried to do `status:RUNNING` which is incorrect.
[...]

PR #2627, opened 2020-01-14
Apps tab: Remove unused field from applications api

danielsdeleo:

The applications API has some inconsistencies. Most of them require substantial effort to fix, but in one case it's a field that's hardcoded to one value because the thing it was intended to represent was never created. So we can improve clarity by removing it.

https://github.com/chef/automate/issues/2622


api/external/applications/applications.proto
```diff
@@ -398,7 +398,7 @@ message Service {
   // Intentionally blank.
   HealthStatus health_check = 4;
   // Intentionally blank.
```
phiggins:
```suggestion
```

This comment isn't necessary anymore.

danielsdeleo: When we originally annotated all the things in this file we put `// Unused.` for the other `reserved` thing so I did that instead for consistency. 👍

Issue #2616, opened 2020-01-13
Profiles search endpoint pagination does not work


phiggins:
From slack:

> My servicenow environment is running from the acceptance channel and I'm seeing some compliance API issues. The compliance/profiles/search API doesn't appear to return results when I search by name
> example search query

```
{
  "name": "ssh-baseline",
  "order": "ASC",
  "sort": "name",
  "page": 1,
  "per_page": 100
}
```

phiggins:
Most likely a regression from #2367

phiggins:
I don't think there is a bug here, just poor (non-existent) documentation leading to failed expectations. We just recently implemented the pagination behavior for that endpoint, previously the parameters were accepted for requests to that endpoint but they were just non-functional. The example request parameters posted above include `page` of 1 and `per_page` of 100, which means the request is for the second page of 100 results, that is results 101-200. Since there is only one profile matching that name, empty results are returned.

To fix this in the short term, there are two options:
* change the `page` parameter to 0 to request the first page of results.
* Omit the `per_page` parameter to turn off the pagination behavior and just get everything back.

vjeffrey:
I know we've previously discussed that this endpoint uses page 0 as a starting point in standup. This might be a good example of this functionality biting us. A user of our api (@satellite15 ) made assumptions about the api's pagination params based on other compliance endpoints, and that api call was broken when pagination was implemented on the endpoint b/c it was implemented assuming a 0 index for the pagination.  
While I understand some folks aren't big fans of starting at page 1 for pagination, and we haven't been great about standardizing fully in our api, it's a user facing feature, and if we think about it in literal terms, we're asking for the first page of content, not the zero-th. Let's talk with ux in standup -- we can always still change things so page 1 is the first page

PR #2628, opened 2020-01-14
Use 1 for the first page of paginated profile results.

phiggins:
As we discovered in #2616, using 0 to represent the first page of paginated results is not necessarily as intuitive as I had originally thought. Given that using 1 for the first page is [somewhat of a convention in automate](https://github.com/chef/automate/blob/d8f6e7482050126e83906d242f56c6c69dbdc7c9/components/applications-service/pkg/params/pagination.go#L10) and is [also used in popular pagination libraries](https://github.com/kaminari/kaminari/blob/master/README.md#the-page-scope), it seems like a good idea to change this.

### DIFF
--- a/components/compliance-service/api/tests/01_market_spec.rb
+++ b/components/compliance-service/api/tests/01_market_spec.rb
@@ -181,7 +181,7 @@ describe File.basename(__FILE__) do
 
   it "limits list results when per_page parameter is present" do
     actual_data = GRPC profiles, :list, Profiles::Query.new(
-      page: 0,
+      page: 1,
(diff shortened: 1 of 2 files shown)


phiggins:
I had originally implemented this differently trying to preserve the original behavior (only paginating when `PerPage` was set in the request parameters) in order to keep backwards compatibility, but it felt pretty awkward and didn't seem very valuable given this is undocumented and only committed a few weeks ago. This PR now maintains the optional pagination behavior if either `PerPage` or `Page` is set.

We'll definitely need to document this in #2356.

PR #2637, opened 2020-01-15
Fixed back button prompts Not Found error

### LINKED ISSUE #1700: Back button prompts Not Found error
## Describe the bug
When on a team's detail page, I click the link to view a user and am navigated to that user's details page. In order to see the team again, I use the browser's <kbd><--</kbd> (back button). I am navigated back to the team details page, but I see a "User Not Found" error banner.

## Expected behavior
I should be able to click the back button, see the previous view without an error banner.

- Chrome browser
Screen Shot 2019-09-24 at 8 20 12 AM

vinay033:

When on a team's detail page, someone clicks the link to view a user and am navigated to that user's details page. In order to see the team again, I use the browser's <-- (back button). I am navigated back to the team details page, but I see a "User Not Found" error banner.

fixes #1700
### :+1: Definition of Done
I have added some changes to fix this issue.
### :athletic_shoe: How to Build 
build components/automate-ui
### :camera: Screenshots
ezgif com-video-to-gif

### DIFF
--- a/components/automate-ui/src/app/modules/user/user-details/user-details.component.ts
+++ b/components/automate-ui/src/app/modules/user/user-details/user-details.component.ts
@@ -55,13 +55,8 @@ export class UserDetailsComponent implements OnInit, OnDestroy {
       pluck('id'),
       filter(identity),
       takeUntil(this.isDestroyed))
-      .subscribe((id: string) => {
-        if (this.preUrlUserId === '') {
-          this.store.dispatch(new GetUser({ id }));
-          this.preUrlUserId = id;
-        } else {
-          this.store.dispatch(new GetUser({ id: this.preUrlUserId }));
-        }
+      this.route.params.subscribe(params => {
+        this.store.dispatch(new GetUser({ id: params.id }));
       });
 
     combineLatest([
--- a/components/automate-ui/src/app/modules/user/user-details/user-details.component.ts
+++ b/components/automate-ui/src/app/modules/user/user-details/user-details.component.ts
@@ -37,7 +37,6 @@ export class UserDetailsComponent implements OnInit, OnDestroy {
   private isDestroyed = new Subject<boolean>();
   private deletingUser = false;
   public updatingUser = false;
-  public preUrlUserId = '';
 
   constructor(
     private store: Store<NgrxStateAtom>,
--- a/components/automate-ui/src/app/modules/user/user-details/user-details.component.ts
+++ b/components/automate-ui/src/app/modules/user/user-details/user-details.component.ts
@@ -37,6 +37,7 @@ export class UserDetailsComponent implements OnInit, OnDestroy {
   private isDestroyed = new Subject<boolean>();
   private deletingUser = false;
   public updatingUser = false;
+  public preUrlUserId = '';
 
   constructor(
     private store: Store<NgrxStateAtom>,
@@ -55,7 +56,12 @@ export class UserDetailsComponent implements OnInit, OnDestroy {
       filter(identity),
       takeUntil(this.isDestroyed))
       .subscribe((id: string) => {
-        this.store.dispatch(new GetUser({ id }));
+        if (this.preUrlUserId === '') {
+          this.store.dispatch(new GetUser({ id }));
+          this.preUrlUserId = id;
+        } else {
+          this.store.dispatch(new GetUser({ id: this.preUrlUserId }));
+        }
       });
 
     combineLatest([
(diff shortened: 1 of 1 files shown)


components/automate-ui/src/app/modules/user/user-details/user-details.component.ts
```diff
@@ -55,7 +56,12 @@ export class UserDetailsComponent implements OnInit, OnDestroy {
       filter(identity),
       takeUntil(this.isDestroyed))
       .subscribe((id: string) => {
-        this.store.dispatch(new GetUser({ id }));
+        if (this.preUrlUserId === '') {
```
bcmdarroch: I'd prefer a fix that prevents the extra call to the user details API when we go back to the team-details page instead of letting it go through with an old ID. 

It appears that when we click the back button, the routeParams subscription on the user-details components grabs the team ID and passes it to the GetUser action.
I'm wondering if we can add another step of filtering to this block that checks the contents of the ActivatedRoute ([more info on that here](https://angular.io/guide/router#activated-route)). Then we could ensure that GetUser is dispatched only if the route includes `users`.

Let me know if you have questions about that idea!

msorens: I concur with @bcmdarroch and her suggested approach is a good one and should work. However, I think we can simplify a bit further by considering the context of the situation.
What Brenna correctly describes is that the route change is firing an Observable on the page we are in the process of leaving; we have no need to do that. Thinking further, we really only need to fire this subscription exactly once--when we arrive on the page.
So adding the `first()` operator should address the scenario:
```
    this.store.pipe(
      select(routeParams),
      pluck('id'),
      filter(identity),
      first())
      .subscribe((id: string) => {
          this.store.dispatch(new GetUser({ id }));
      });
```

components/automate-ui/src/app/modules/user/user-details/user-details.component.ts
```diff
@@ -37,6 +37,7 @@ export class UserDetailsComponent implements OnInit, OnDestroy {
   private isDestroyed = new Subject<boolean>();
   private deletingUser = false;
   public updatingUser = false;
+  public preUrlUserId = '';
```
bcmdarroch: looks like we can drop this variable

components/automate-ui/src/app/modules/user/user-details/user-details.component.ts
```diff
@@ -54,8 +55,8 @@ export class UserDetailsComponent implements OnInit, OnDestroy {
       pluck('id'),
       filter(identity),
       takeUntil(this.isDestroyed))
-      .subscribe((id: string) => {
-        this.store.dispatch(new GetUser({ id }));
+      this.route.params.subscribe(params => {
+        this.store.dispatch(new GetUser({ id: params.id }));
```
msorens: It looks like you are trying to nest a second `subscribe` inside the first `subscribe`, which does not quite make sense. But I don't believe the code as written is doing what you think. You actually have two separate subscriptions--the first one stops at L57 and does nothing (you will see the linter complaining about a missing semicolon there). The second one starts at L58 and does what the code did originally, but it is missing the vital `takeUntil`.

There are two possibilities for this block. Using the approach you were suggesting would look like this:
```
    this.route.params.pipe(
      takeUntil(this.isDestroyed))
      .subscribe(params => {
        this.store.dispatch(new GetUser({ id: params.id }));
      });
```
Or, the approach I previously suggested would also work:
```
    this.store.pipe(
      select(routeParams),
      pluck('id'),
      filter(identity),
      first())
      .subscribe((id: string) => {
          this.store.dispatch(new GetUser({ id }));
      });
```
As the first is a bit shorter, probably makes sense to use that one.

vinay033: Thanks for the suggestions, I have updated the changes accordingly, please have a look :)

msorens: Hmm... because of the failing tests, we might have to go with the second version, which I think should allow the tests to pass. Try that and see if it works.

vinay033: yes, It works, Thanks :)

components/automate-ui/src/app/modules/user/user-details/user-details.component.ts
susanev: @vinay033 can you take this unused variable out?

vinay033: yes, I removed.

susanev: hmm, but its still showing in the files changed??

PR #2647, opened 2020-01-15
[automate-2520] support S3 snapshots when using external elasticsearch

### LINKED ISSUE #2520: External Elasticsearch breaks S3 snapshot backups.
## Describe the bug
If external-elasticsearch servers are configured and S3 snapshots are setup they will only work for Postgres. When services attempt to create their repository they will fail as they are using an empty string for a bucket name.

This seems to be caused by this block being dropped when external es is configured. 
https://github.com/chef/automate/blob/9e0022b71c146794e594c4f598464e06f5c141f4/components/es-sidecar-service/habitat/config/config.toml#L52-L85

I did notice there are two different docs describing how to set this up.

This one works on standalone.
https://automate.chef.io/docs/backup/#backup-to-s3

This one moves the same keys to a different place, but seems to be ignored when you set it.
https://automate.chef.io/docs/install/#backup-to-s3

I don't think we need separate backup configuration for external es, as all snapshots are taken through the API, and the settings are set by the user and are generally the same either way.

## To Reproduce
Patch in a config with external Elasticsearch servers and S3 backups configured similar to this.

```
[global]
  [global.v1]
    [global.v1.backups]
      location = "s3"
      [global.v1.backups.s3]
        [global.v1.backups.s3.bucket]
          endpoint = "https://s3.us-west-2.amazonaws.com"
          base_path = "a2ha"
          name = "tcate-a2-backup-test"
    [global.v1.external]
      [global.v1.external.elasticsearch]
        enable = true
        nodes = ["https://10.1.0.79:9200", "https://10.1.1.96:9200", "https://10.1.2.41:9200"]
        [global.v1.external.elasticsearch.backup]
          enable = true
          location = "s3"
          [global.v1.external.elasticsearch.backup.s3]
            bucket = "tcate-a2-backup-test"
            base_path = "a2ha"
```

Attempt to run chef-automate backup and notice that event-feed fails and deployment service returns this error.

```
Dec 16 15:45:48 ip-10-1-0-105.eu-west-1.compute.internal hab[3196]: deployment-service.default(O): time="2019-12-16T15:45:48Z" level=error msg="Backup failed" backup_id=20191216154546 error="es-sidecar-service failed to create snapshot: rpc error: code = Unknown desc = failed to create snapshot repository chef-automate-es6-event-feed-service for snapshot \"20191216154546\" of service \"event-feed-service\": elasticsearch repository create request failed for repo \"chef-automate-es6-event-feed-service\"; request data: '{s3 map[base_path:a2ha/chef-automate-es6-event-feed-service bucket: compress:false]}': elastic: Error 500 (Internal Server Error): [chef-automate-es6-event-feed-service] cannot create blob store [type=repository_exception]"
```
Notice bucket is an empty string in that error. If you check ES the repo was created however,
```
curl -s http://localhost:9200/_snapshot/chef-automate-es6-event-feed-service?pretty
{
  "chef-automate-es6-event-feed-service": {
    "type": "s3",
    "settings": {
      "bucket": "",
      "compress": "false",
      "base_path": "a2ha/chef-automate-es6-event-feed-service"
    }
  }
}
```

If you look at /hab/svc/es-sidecar-service/config/config.toml you'll see this section for backups,
```
[backups]
  backend = "disable"
```

## Expected behavior
In every configuration Elasticsearch should read the keys in [global.v1.backups] for it's backup settings, similar to how postgres backups work.

CentOS Linux release 7.7.1908 (Core)

Version: 2
CLI Build: 20191128000858
Server Build: 20191129172405

ryancragun:
* Add new external es with s3 integration test
* Rewrite SetGlobalConfig() for the es-sidecar-service
* Add unit tests for the various external, global and es-sidecar level
  configuration combinations
* Extract common s3 integration test functions into helpers
* Update the external Es to 6.8.3 to match what is currently deployed
* Add and configure repository-s3 for external es integration tests

Fixes #2520

### DIFF
--- a/api/config/es_sidecar/config_request.go
+++ b/api/config/es_sidecar/config_request.go
@@ -85,16 +85,15 @@ func (c *ConfigRequest) SetGlobalConfig(g *ac.GlobalConfig) {
 		backupsCfg = &ConfigRequest_V1_System_Backups{}
 	}
 
-	// If we're in external mode, apply and global and external config
+	c.applyGlobalBackupConfig(backupsCfg, g.GetV1().GetBackups())
+
 	if externalCfg.GetEnable().GetValue() {
-		c.applyGlobalBackupConfig(backupsCfg, g.GetV1().GetBackups())
 		c.applyExternalConfig(backupsCfg, externalCfg.GetBackup())
+		// Early return as we don't need to migrate to the backup-gateway with
+		// external es
 		return
 	}
 
-	// If we're in internal mode, apply global config and then ensure that old
-	// internal fs config is migrated to use the backup gateway in S3 mode.
-	c.applyGlobalBackupConfig(backupsCfg, g.GetV1().GetBackups())
 	c.migrateToBackupGateway(backupsCfg)
 
 	return
@@ -173,8 +172,8 @@ func (c *ConfigRequest) applyGlobalBackupConfig(
 		return
 	}
 
-	if backend := global.GetLocation(); backend != nil {
-		cfg.Backend = backend
+	if backend := global.GetLocation().GetValue(); backend != "" {
+		cfg.Backend = w.String(backend)
 	}
 
 	if globalS3 := global.GetS3(); globalS3 != nil {
--- a/api/config/es_sidecar/config_request.go
+++ b/api/config/es_sidecar/config_request.go
@@ -26,6 +26,7 @@ func NewConfigRequest() *ConfigRequest {
 }
 
 // DefaultConfigRequest returns a new instance of ConfigRequest with default values.
+// nolint: gomnd
 func DefaultConfigRequest() *ConfigRequest {
 	c := NewConfigRequest()
 	c.V1.Sys.Service.Host = w.String("0.0.0.0")
@@ -34,7 +35,7 @@ func DefaultConfigRequest() *ConfigRequest {
 	c.V1.Sys.Log.Level = w.String("info")
 	c.V1.Sys.Log.Format = w.String("text")
 
-	c.V1.Sys.Backups.Backend = w.String("fs")
+	c.V1.Sys.Backups.Backend = w.String("s3")
 	c.V1.Sys.Backups.S3.Es.Compress = w.Bool(true)
 	return c
 }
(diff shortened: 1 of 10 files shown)


integration/base.sh
```diff
@@ -373,6 +373,8 @@ export A2_ROOT_DIR="$source_dir/../"
 # Load the shared code
 # shellcheck source=./helpers/log.sh
 source "${source_dir}/helpers/log.sh"
+# shellcheck source=./helpers/s3.sh
+source "${source_dir}/helpers/s3.sh"
```
jaym: [nit] maybe source this in the tests that need it instead

api/config/es_sidecar/config_request.go
```diff
@@ -141,4 +67,173 @@ func (c *ConfigRequest) SetGlobalConfig(g *ac.GlobalConfig) {
...
+	externalCfg := g.GetV1().GetExternal().GetElasticsearch()
+
+	backupsCfg := c.GetV1().GetSys().GetBackups()
+	if backupsCfg == nil {
+		backupsCfg = &ConfigRequest_V1_System_Backups{}
+	}
+
+	// If we're in external mode, apply and global and external config
+	if externalCfg.GetEnable().GetValue() {
+		c.applyGlobalBackupConfig(backupsCfg, g.GetV1().GetBackups())
```
jaym: can this be pulled out of the if statement? it looks like its done first in both branches

api/config/es_sidecar/config_request.go
```diff
@@ -141,4 +67,173 @@ func (c *ConfigRequest) SetGlobalConfig(g *ac.GlobalConfig) {
...
+// into the backup config
+func (c *ConfigRequest) applyGlobalBackupConfig(
+	cfg *ConfigRequest_V1_System_Backups,
+	global *ac.Backups) {
+
+	if global == nil {
+		return
+	}
+
+	if backend := global.GetLocation(); backend != nil {
```
jaym: should we be checking `GetValue()` here? would be things happen if it was an empty string?

integration/helpers/s3.sh
```diff
@@ -0,0 +1,88 @@
...
+
+do_test_restore_s3_default() {
+  do_test_restore_default || return 1
+  delete_backup_and_assert_idempotent
+  verify_packages
+
+  hab pkg install core/hab
+  artifacts_dir=$(mktemp -d --suffix="bldr_smoke")
+  # TODO(jaym): remove when we upgrade hab to a version
+  # that supports hab pkg download
```
yzl: Is this comment still needed, or is there something about the `hab pkg download` command we're running on the next line that isn't what it's going to be in the next version?

ryancragun: I have no idea, I literally extracted the code from the existing test into a helper.

Issue #2488, opened 2019-12-11
API and CLI command to support manual deletion of services


kmacgugan:
## User Story

As a user I want to be able to delete services when I know I have destroyed them. 

Provide an API and CLI command to delete services by service group name, app name and environment (OR special hash of these that makes up the ID in the db, which can be used by the UI)

Issue #1192, opened 2019-08-07
Manual Deletion of Services in the Applications View


danielsdeleo:
_(Note: the following content was edited by @apriofrost, reach out to her if anything is not clear.)_

## Problem
Currently, after users have set up the Applications Dashboard and be able to view the service health status in the dashboard, there isn't an easy way for them to remove the services from the dashboard if they were intentionally decommissioned. 
We have provided users the ability to automatically remove disconnected services after a period set by the user. However, if they want to remove a service immediately, no matter if it has been marked as disconnected or not, they are unable to accomplish it. 

## Solution

Allow users to delete any services in the Applications Dashboard, so they are not alerted about disconnected services that were intentionally decommissioned.

## Acceptance Criteria
- [x] Replace the current `node` icon in the service instance card with a checkbox, update the position to align with the node FQDN on the top (the current is central vertically).
_(Note: the checkbox state only persist for one page, if users navigate to another page, the checkbox states on the prior page is reset to default/unselected.)_
<img src="https://images.zenhubusercontent.com/5a721a104b5806bc2bc51768/92cac177-8608-446e-b9e7-f7532c43345d" width="300">

- [x] Add a bulk action bar above the service instance cards on the right-side panel. The bulk action bar is consists of two types of elements: 1. the `Select All` selector, 2. the available actions, in this case,`Delete Services`. See screenshot below:
![image.png](https://images.zenhubusercontent.com/5a721a104b5806bc2bc51768/b97ff303-a0b9-4beb-8f3d-721a67f58ff6)
The bulk action bar can have three states:
_(Note: the bulk action bar state only persists for one page, if users navigate to another page, the bulk action bar state is reset to default.)_
1. When no service instance is selected, the `Select All` selector is in the `unselected` state. The available actions are disabled.
![image.png](https://images.zenhubusercontent.com/5a721a104b5806bc2bc51768/95b0eb29-c09a-4d3b-8d27-54bca47fab9b)
2. When there is any service instance selected but not all on a page, the `Select All` selector is in the `intermediate` state. The available actions are enabled. In this case, the `Delete Service` button label updates to reflect the number of service instances selected.
![image.png](https://images.zenhubusercontent.com/5a721a104b5806bc2bc51768/e79d6e8b-56e9-46f9-a3d3-99fd1d686ef6)
[...]

PR #2651, opened 2020-01-17
Dan/manual svc delete cli

danielsdeleo:

Adds the necessary API to support the manual service deletion flow described in #1192 and provides a CLI to use it.

#1192
#2488 

* *Build:* automate-gateway, automate-cli, applications-service
* Start all services and run `applications_populate_database`
* `chef-automate applications show-svcs` shows all the services
* `chef-automate applications show-svcs -a "*" -b "*" -c "*" -e "*" -g "*" -o "*" -n "*" -s "*" -v "*" -D` uses all the filtering options (this will probably not return any services unless you waited for them to become disconnected)
* Experiment with any of the filter options (see what they mean with `--help`)
* Delete services with `chef-automate applications remove-svcs`, they should not show up in the `show-svcs` command after.

- [x] Tests added/updated?
- [x] Docs added/updated?

```
chef-automate applications show-svcs -a "*" -b "*" -c "*" -e "*" -g "*" -o "*" -n "*" -s "*" -v "*" | head -5
id   	svc.group    	release                          	FQDN                	app:env  	status  
17741	redis.default	custom/redis/1.0.1/20190115189830	ebnru2er.example.net	demo:demo	CRITICAL
17695	nginx.default	custom/nginx/2.0.0/20190115189976	7810o1k8.example.net	demo:demo	CRITICAL
17734	redis.default	custom/redis/1.0.1/20190115189830	djfjxnrq.example.net	demo:demo	CRITICAL
17676	nginx.default	custom/nginx/2.0.0/20190115189976	23hacbg2.example.net	demo:demo	CRITICAL
17670	billing-svc.default	custom/billing-svc/1.3.4/20190115182382	jb0dbg8y.example.net	billing-backend:qa	CRITICAL
```

```
[258][default:/src:0]# chef-automate applications remove-svcs -a "m*" -b "*" -c "*" -e "p*" -g "*" -o "*" -n "mobile-api-auth*" -s "*" -v "*" 
id   	svc.group              	release                                    	FQDN                	app:env                  	status
18542	mobile-api-auth.default	custom/mobile-api-auth/2.0.8/20190115181192	bllhcayc.example.net	mobile-app-api:production	OK

The above 1 services will be deleted. Do you wish to continue? (y/n)
y
Removed 1 services
[259][default:/src:0]# chef-automate applications show-svcs -a "m*" -b "*" -c "*" -e "p*" -g "*" -o "*" -n "mobile-api-auth*" -s "*" -v "*" 
id	svc.group	release	FQDN	app:env	status
```


vjeffrey:
i really appreciate the detailed repro steps, **thank you** for always including those!!
i'll go through and review locally after demos

components/applications-service/pkg/server/server.go
```diff
@@ -335,6 +335,29 @@ func (app *ApplicationsServer) DeleteDisconnectedServices(ctx context.Context,
...
 }
 
+func (app *ApplicationsServer) DeleteServicesByID(ctx context.Context,
+	request *applications.DeleteServicesByIDReq) (*applications.ServicesRes, error) {
+	svcIDs := request.GetIds()
+	if svcIDs == nil {
+		return new(applications.ServicesRes),
+			status.Error(codes.InvalidArgument, "Must specify service IDs to delete")
+	}
+	if len(svcIDs) == 0 {
```
vjeffrey: i'm curious why the nil check == return error but the 0 id check returns empty?

.studio/applications-service
```diff
@@ -240,6 +240,12 @@ function applications_publish_supervisor_message() {
   applications-publisher --internal-nats --port 10140 "$@"
 }
 
+function applications_random_fqdn() {
```
afiune: tenor-130446749

components/applications-service/pkg/server/server.go
afiune: I had the same question, I remember defaulting to return `new()` for consistency.

components/applications-service/pkg/storage/postgres/services.go
```diff
@@ -164,6 +165,34 @@ RETURNING s.id
 	, s.health_check_exit_status as health_check_exit_status
 `
 
+	deleteServicesByID = `
+   DELETE FROM service_full AS s
```
afiune: (nit): extra tab.
```suggestion
DELETE FROM service_full AS s
```

components/applications-service/pkg/storage/postgres/services.go
```diff
@@ -164,6 +165,34 @@ RETURNING s.id
...
+    WHERE s.id IN (%s)
```
afiune: (nit): extra tab.
```suggestion
WHERE s.id IN (%s)
```

components/applications-service/pkg/server/server.go
danielsdeleo: @afiune I think the question is why I chose to return `status.Error(codes.InvalidArgument, ....` when `request.GetIds()` is nil, but return an empty response for an empty request. The answer to that is that the nil case feels more like a malformed/incompletely formed request, whereas an empty request is kinda silly but correctly formed. You could also imagine that a user writes some periodic cleanup script and doesn't think to skip the delete request when there's zero services to cleanup; failing the request might trigger some alerts or something for little reason.

I'll change the ` &applications.ServicesRes{}` to `new` so it matches.

vjeffrey: that makes sense, thanks for explaining!

vjeffrey: that makes sense, thanks for explaining!

PR #2657, opened 2020-01-17
Run compliance scanner tests against inspec master

phiggins:
In order to have better awareness of breaking changes in the integration of automate and inspec, this PR adds a buildkite task to build the latest inspec from master and use that to run the inspec focused tests.

### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -587,6 +587,34 @@ steps:
       - gopath-checkout#v1.0.1:
           import: github.com/chef/automate
 
+  - label: "compliance-service scanner with inspec master"
+    command:
+      - sudo sysctl -w vm.max_map_count=262144 # Needed for ElasticSearch 6
+      - export PATH=\$PATH:/usr/local/go/bin:\$HOME/.local/bin:\$HOME/bin
+      - sudo scripts/install_inspec_master.sh
+      - sudo scripts/install_golang.sh
+      - sudo scripts/install_grpcurl.sh
+      - cd components/compliance-service
+      - sudo -E  PATH=\$(hab pkg path core/ruby)/bin:\$PATH make test-integration-scanner
+      - cd ../.. && sudo git clean -fxd
+    artifact_paths:
+      - /tmp/secrets-service.log
+      - /tmp/event-service.log
+      - /tmp/compliance.log
+      - /tmp/nodemanager-service.log
+    env:
+      ES_VER: 6
+      LANG: en_US.UTF-8
+    timeout_in_minutes: 20
+    expeditor:
+      executor:
+        linux:
+          single-use: true
+          privileged: true
+    plugins:
+      - gopath-checkout#v1.0.1:
+          import: github.com/chef/automate
+
   - label: "compliance-service reporting"
     command:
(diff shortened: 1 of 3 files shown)


.expeditor/verify.pipeline.yml
```diff
@@ -587,6 +587,34 @@ steps:
...
+      LANG: en_US.UTF-8
+    timeout_in_minutes: 20
+    expeditor:
+      executor:
+        linux:
+          single-use: true
+          privileged: true
+    plugins:
+      - gopath-checkout#v1.0.1:
+          import: github.com/chef/automate
```
phiggins: This is a copy of the existing task for the scanner tests that calls a different script to install inspec.

components/compliance-service/Makefile
```diff
@@ -464,7 +464,7 @@ test: generate-ruby-grpc wait-for-compliance-service
 test-prep: $(MARKET_PATH) $(PROFILES_PATH) start-ssh-node download-sample-market-profiles install-inspec start-es-pg
 
 install-inspec:
-	@inspec --version | grep -q $(shell cat ../../INSPEC_VERSION) || gem install inspec-bin --no-document --version $(shell cat ../../INSPEC_VERSION)
+	@inspec --version 2> /dev/null | grep -q $(shell cat ../../INSPEC_VERSION) || gem install inspec-bin --no-document --version $(shell cat ../../INSPEC_VERSION)
```
phiggins: This change prevents the "inspec: command not found" message when you don't already have inspec installed.

phiggins:
Should this be allowed to fail without failing the entire build?

vjeffrey:
> Should this be allowed to fail without failing the entire build?

My opinion: yes!
Reasoning: if the whole build fails based on this failure, the entire Automate team will be blocked from merging until the inspec issue is fixed. While I believe some gates of this kind are super useful/needed, they should not be at the merge level, they should be at the release level.

.expeditor/verify.pipeline.yml
```diff
@@ -587,6 +587,34 @@ steps:
       - gopath-checkout#v1.0.1:
           import: github.com/chef/automate
 
+  - label: "compliance-service scanner with inspec master"
+    command:
```
vjeffrey: add `soft_fail: true` ?

scripts/install_inspec_master.sh
```diff
@@ -0,0 +1,19 @@
...
+
+install_inspec() {
+  hab pkg install core/gcc --binlink --force
+  hab pkg install core/ruby --binlink --force
+
+  tmp_dir=$(mktemp -d)
+  git clone --depth=1 https://github.com/inspec/inspec "$tmp_dir"
+  cd "$tmp_dir"
+  bundle
+  bundle exec rake install
```
zenspider: This looks great. I have some scripts in inspec/inspec's expeditor dir to make it really easy to cache bundle artifacts. We can get that on another iteration. Cuts down ~3 minutes per. Maybe it's also possible to cache the hab pkgs?

.expeditor/verify.pipeline.yml
zenspider: For now, totally... but at some point, once trust is built up and the system is stabilized, I'd like that removed and eyeballs to hit it when it goes red.

PR #2665, opened 2020-01-21
Increase project limit default to 300

blakestier:

With the IAM release, we are now supporting up to 300 projects. 

Customers using the default limit of 6 will have their limit automatically increased. Customers who have increased their limit will still have that limit applied.

`cat /hab/svc/authz-service/config/config.yml` will show you the current limit of 6

`rebuild components/automate-deployment
  rebuild components/authz-service`

`cat /hab/svc/authz-service/config/config.yml` will show you the new limit of 300

Try creating 301 projects-- you should get an error on the 301st.

per @msorens on quick project creation:

> 
> You can easily generate (or remove) hundreds of projects from the command-line with these commands:
> 
> ### Generate 500 starting with id 101
> authGen projects create 500 101
> ### Remove them after testing
> authGen projects delete 500 101
> That will work inside or outside hab studio with a couple prerequisites:
> 1. Point at your A2 instance and authorize it:
> 
> export TOK=`chef-automate iam token create TEST --admin`
> export TARGET_HOST=https://a2-dev.test
> Cherry-pick commit 0784ca05 into your PR that tweaks the existing script.
> Source the script (. ./scripts/auth_collections.sh)
> 

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/api/config/authz/config_request.go
+++ b/api/config/authz/config_request.go
@@ -48,9 +48,11 @@ func (c *ConfigRequest) Validate() error {
 
 	if projectLimit != nil {
 		if limit := projectLimit.GetValue(); limit < constants_v2.MinConfigurableProjects {
-			// MinConfigurableProjects supports customers who previously increased
-			// their limit, but not as high as the new limit. Can be removed when
-			// we no longer limit projects.
+			// Previously, users could not configure their project_limit to be below the
+			// default.
+			// MinConfigurableProjects supports customers who previously
+			// increased their limit to a number that is lower than the new default.
+			// Can be removed when we no longer limit projects.
 			e := errors.Errorf("project limit must be at least %v", constants_v2.MinConfigurableProjects)
 			cfgErr.AddInvalidValue("auth_z.v1.sys.service.project_limit", e.Error())
 		}
--- a/api/config/authz/config_request.go
+++ b/api/config/authz/config_request.go
@@ -47,8 +47,11 @@ func (c *ConfigRequest) Validate() error {
 	projectLimit := c.GetV1().GetSys().GetService().GetProjectLimit()
 
 	if projectLimit != nil {
-		if limit := projectLimit.GetValue(); limit < constants_v2.MaxProjects {
-			e := errors.Errorf("project limit must be at least %v", constants_v2.MaxProjects)
+		if limit := projectLimit.GetValue(); limit < constants_v2.MinConfigurableProjects {
+			// MinConfigurableProjects supports customers who previously increased
+			// their limit, but not as high as the new limit. Can be removed when
+			// we no longer limit projects.
+			e := errors.Errorf("project limit must be at least %v", constants_v2.MinConfigurableProjects)
 			cfgErr.AddInvalidValue("auth_z.v1.sys.service.project_limit", e.Error())
 		}
 	}
--- a/api/config/authz/config_request.go
+++ b/api/config/authz/config_request.go
@@ -29,7 +29,7 @@ func DefaultConfigRequest() *ConfigRequest {
 
 	c.V1.Sys.Service.Host = w.String("0.0.0.0")
 	c.V1.Sys.Service.Port = w.Int32(10130)
-	c.V1.Sys.Service.ProjectLimit = w.Int32(constants_v2.MaxProjects)
+	c.V1.Sys.Service.ProjectLimit = w.Int32(constants_v2.DefaultProjectLimit)
 	c.V1.Sys.Logger.Level = w.String("info")
 	c.V1.Sys.Logger.Format = w.String("text")
 	c.V1.Sys.Storage.Database = w.String("chef_authz_service")
--- a/api/config/authz/config_request.go
+++ b/api/config/authz/config_request.go
@@ -1,10 +1,11 @@
 package authz
 
 import (
+	fmt "fmt"
+
 	config "github.com/chef/automate/api/config/shared"
 	w "github.com/chef/automate/api/config/shared/wrappers"
 	constants_v2 "github.com/chef/automate/components/authz-service/constants/v2"
-	"github.com/pkg/errors"
 )
 
 // NewConfigRequest returns a new instance of ConfigRequest with zero values.
@@ -52,9 +53,9 @@ func (c *ConfigRequest) Validate() error {
 			// default.
 			// MinConfigurableProjects supports customers who previously
 			// increased their limit to a number that is lower than the new default.
-			// Can be removed when we no longer limit projects.
-			e := errors.Errorf("project limit must be at least %v", constants_v2.MinConfigurableProjects)
-			cfgErr.AddInvalidValue("auth_z.v1.sys.service.project_limit", e.Error())
+			// It should be removed when we no longer limit projects.
+			failureStr := fmt.Sprintf("project limit must be at least %d", constants_v2.MinConfigurableProjects)
+			cfgErr.AddInvalidValue("auth_z.v1.sys.service.project_limit", failureStr)
 		}
 	}
 
--- a/components/authz-service/constants/v2/constants.go
+++ b/components/authz-service/constants/v2/constants.go
@@ -50,4 +50,10 @@ const (
 // Business logic constants
 const (
 	DefaultProjectLimit = 300
+	// MinConfigurableProjects is based on our original default for customers who have
+	// previously set their project_limit to greater than 6, but less than the
+	// new default of 300.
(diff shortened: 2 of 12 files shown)


components/authz-service/server/v2/project_test.go
```diff
@@ -489,8 +491,9 @@ func setupProjectsAndRules(t *testing.T) (api.ProjectsClient, *cache.Cache, *cac
 	l, err := logger.NewLogger("text", "error")
 	require.NoError(t, err, "init logger for storage")
 
-	mem_v2 := memstore_v2.New()
-	pg, testDb, _, _, _ := testhelpers.SetupTestDB(t)
+	mem_v2 := memstore_v2.NewWithProjectLimit(projectLimitForTesting)
+	// TODO: we should be able to optionally use PG. We're only using memstore
```
blakestier: @chef/auth-team  is that what we want?

terraform/test-environments/modules/chef_automate_install/main.tf
```diff
@@ -234,9 +234,6 @@ EOF
 [global.v1]
   fqdn = "${var.alb_fqdn == "" ? element(var.instance_fqdn, count.index) : var.alb_fqdn}"
 
-[auth_z.v1.sys.service]
```
blakestier: We could optionally make this another number greater than 300, but I think we're unlikely to hit that.

components/authz-service/server/v2/project_test.go
blakestier: (we are testing the limit elsewhere, but I do think this suite *looks* like it's using PG when it's not . . .)

components/automate-chef-io/content/docs/iam-v2-guide.md
```diff
@@ -241,7 +241,7 @@ so they are not relevant when assigning IAM resources such as teams or roles.
 
 #### Configuring Project Limit
 
-By default, Chef Automate limits you to six projects. You can increase the project limit using the command line.
+By default, Chef Automate is currently limited to 300 projects. If you would like to increase that limit, you will need to do so using the Chef Automate CLI:
```
mjingle: So this line was originally reworked to decrease the use of passive tense and simplify the sentence structure. I recommend just changing the number to `300`.
```suggestion
By default, Chef Automate limits you to 300 projects. You can increase the project limit using the command line.
```

blakestier: I strictly swapped the numbers-- not sure how this old text took over. Maybe a bad rebase 🤷‍♀

mjingle: Totally understand. I was mildly suspicious the text change was a computer conspiracy, not actually by a user. It's so frustrating when that happens!

api/config/authz/config_request.go
```diff
@@ -47,8 +47,13 @@ func (c *ConfigRequest) Validate() error {
...
-		if limit := projectLimit.GetValue(); limit < constants_v2.MaxProjects {
-			e := errors.Errorf("project limit must be at least %v", constants_v2.MaxProjects)
+		if limit := projectLimit.GetValue(); limit < constants_v2.MinConfigurableProjects {
+			// Previously, users could not configure their project_limit to be below the
+			// default.
+			// MinConfigurableProjects supports customers who previously
+			// increased their limit to a number that is lower than the new default.
+			// Can be removed when we no longer limit projects.
+			e := errors.Errorf("project limit must be at least %v", constants_v2.MinConfigurableProjects)
 			cfgErr.AddInvalidValue("auth_z.v1.sys.service.project_limit", e.Error())
```
tylercloke: [supernit] Looks like you are just passing a string (`e.Error()`) so you don't need an error at all:

```
failureStr := fmt.Sprintf("project limit must be at least %d", constants_v2.MinConfigurableProjects)
cfgErr.AddInvalidValue("auth_z.v1.sys.service.project_limit", failureStr)
```

components/authz-service/server/v2/project_test.go
tylercloke: I'd personally like us to ditch all the memstore tests altogether and just use PG where necessary but I might be in the minority there :)

api/config/authz/config_request.go
```diff
@@ -47,8 +47,13 @@ func (c *ConfigRequest) Validate() error {
 	projectLimit := c.GetV1().GetSys().GetService().GetProjectLimit()
 
 	if projectLimit != nil {
-		if limit := projectLimit.GetValue(); limit < constants_v2.MaxProjects {
-			e := errors.Errorf("project limit must be at least %v", constants_v2.MaxProjects)
+		if limit := projectLimit.GetValue(); limit < constants_v2.MinConfigurableProjects {
```
msorens: While I agree with this change to a lower limit check, not clear that it should ever be removed--the system needs to allow for at least 1 project I would think. So consider setting it to 1 now and be done with it...?

components/authz-service/constants/v2/constants.go
```diff
@@ -49,5 +49,11 @@ const (
 
 // Business logic constants
 const (
-	MaxProjects = 6
+	DefaultProjectLimit = 300
+	// Previously, users could not configure their project_limit to be below the
+	// default.
+	// MinConfigurableProjects supports customers who previously
+	// increased their limit to a number that is lower than the new default.
+	// It should removed when we no longer limit projects.
```
msorens:
```suggestion
	// It should be removed when we no longer limit projects.
```
(But also see my earlier comment.)

components/authz-service/server/v2/project_test.go
```diff
@@ -33,6 +33,8 @@ import (
 	"github.com/chef/automate/lib/tls/test/helpers"
 )
 
+const projectLimitForTesting = 10 // DefaultProjectLimit is 300
```
msorens: This becomes a maintenance issue. Maybe:
```suggestion
const projectLimitForTesting = 10 // DefaultProjectLimit is higher
```

components/authz-service/server/v2/rules_property_test.go
```diff
@@ -24,7 +24,7 @@ import (
 const (
 	idRegex          = "^[a-z0-9-_]{1,64}$"
 	conditionLimit   = 10 // help avoid the grpc limit of 4194304
-	projectLimit     = 6  // system limitation
+	projectLimit     = 6  // system limitation is 300
```
msorens: Same suggestion as above:
```suggestion
	projectLimit     = 6  // system limitation is higher
```

integration/tests/cypress_e2e.sh
```diff
@@ -6,16 +6,6 @@ test_name=":cypress: e2e tests"
...
 
-do_create_config() {
-    do_create_config_default
-
-    #shellcheck disable=SC2154
-    cat <<EOF >> "$test_config_path"
-[auth_z.v1.sys.service]
-  project_limit = 50
-EOF
-}
```
msorens: By removing this does that mean that cypress just inherits the 300?

components/automate-chef-io/content/docs/iam-v2-guide.md
```diff
@@ -241,7 +241,7 @@ so they are not relevant when assigning IAM resources such as teams or roles.
 
 #### Configuring Project Limit
 
-By default, Chef Automate limits you to six projects. You can increase the project limit using the command line.
+By default, Chef Automate limits you to 300 projects. You can increase the project limit using the command line.
```
msorens: Global search for "six projects" turns up one more occurrence in iam-v2-overview.md.

api/config/authz/config_request.go
blakestier: @msorens We could, but once we do away with the limit, we won't allow customers to configure their limit at all, and this will all go away.

components/authz-service/server/v2/project_test.go
bcmdarroch: 💯 These should either run both with memstore and pg or even just pg. Definitely worth making a tech debt card to discuss as a team. The only reason I can think of for using that we can easily seed the store with whatever we need for the test. But we can do that in pg too so it's not a great reason :/

components/authz-service/server/v2/rules_property_test.go
bcmdarroch: any reason not to update this test with the current default limit?

api/config/authz/config_request.go
blakestier: Huh. Good catch.

PR #2683, opened 2020-01-22
Regen CLI docs after they failed for a while

danielsdeleo:

The pipeline normally runs `.expeditor/generate-automate-cli-docs.sh` automatically when there are changes to the CLI, but there was an error with this script for the past month or so and we are behind. I would like to reference some of the new functionality in the CLI elsewhere in the docs so I need the CLI docs to be updated.

https://github.com/chef/automate/pull/2632


phiggins:
Would it make sense to add the script itself to the list of files watched to see if the script needs to be run?

danielsdeleo:
@phiggins good suggestion, I made #2684 for that

PR #2687, opened 2020-01-22
Apps Tab docs for Habitat 1.5 Release

danielsdeleo:

* In the upcoming habitat 1.5 release, the event stream functionality is no longer gated behind a feature flag (https://github.com/habitat-sh/habitat/pull/7335). I updated the minimum version of habitat required in the docs to 1.5 and removed the feature flag stuff from the code examples.
* In the acceptance deployment of the docs, the "applications" page seems to be shadowed by the "applications" navigation category. Renaming the page to "applications dashboard" both fixes the shadowing and matches the page title. In the screenshot here, we should see "EAS" and "applications" items in the "Applications" category but they are shadowed:
Screen Shot 2020-01-22 at 12 11 53 PM
With the change in this PR you see this instead:
Screen Shot 2020-01-22 at 12 44 50 PM

* Change instructions for fetching the TLS/SSL cert from automate to use the `external-cert` command which we added for this purpose
* Add "long" description to some CLI commands which will improve the generated documentation.


components/automate-chef-io/content/docs/applications-setup.md
```diff
@@ -106,8 +103,8 @@ Chef Automate's TLS certificate file is located at `/hab/svc/automate-load-balan
 
 From Chef Automate:
 
-1. On the command line, run `chef-automate config show`
-1. Copy the `[[global.1.frontend_tls]]` certificate contents from the certificate list
+1. On the command line, run `chef-automate external-cert show`
```
kagarmoe: Is this change in the dev channel?

danielsdeleo: Yes, it's in dev and also in acceptance. It comes from https://github.com/chef/automate/pull/2504

kagarmoe: Thanks!

components/automate-chef-io/content/docs/applications-setup.md
```diff
@@ -106,8 +103,8 @@ Chef Automate's TLS certificate file is located at `/hab/svc/automate-load-balan
...
+1. Copy the certificate contents from the output
 1. Make sure to copy it entirely, including `-----BEGIN CERTIFICATE...END CERTIFICATE-----`
 1. Save the file as `automate.cert` somewhere safe and accessible to you
```
kagarmoe: Does this need to be `ssl-certificate.crt` if the user is copying it to `/hab/cache/ssl`? See the [certs](https://github.com/habitat-sh/on-prem-builder/blob/master/on-prem-docs/builder-automate.md#step-three-put-the-certs-with-the-install-script) section in the on-prem-docs.

components/automate-chef-io/content/docs/applications-setup.md
kagarmoe: Works like a dream!

components/automate-cli/cmd/chef-automate/applications.go
```diff
@@ -168,7 +171,14 @@ func newApplicationsRemoveSvcsCmd() *cobra.Command {
 	c := &cobra.Command{
 		Use:   "remove-svcs",
 		Short: "Remove services from the applications database",
-		RunE:  runApplicationsRemoveSvcsCmd,
+		Long: `
+Remove services from the applications database.
+
+You must decommission the services prior to running this command. Services
```
kagarmoe: What does "decommission" mean in this context? `hab svc stop` or `hab svc unload`? Something else? Wat specific actions does the user need to take?

components/automate-cli/cmd/chef-automate/applications.go
```diff
@@ -168,7 +171,14 @@ func newApplicationsRemoveSvcsCmd() *cobra.Command {
...
+removed with this command will be added to the database again if Automate
+receives a health check from the service.
```
kagarmoe: Asking for clarification. Does "Services removed with this command will be added to the database again if Automate receives a health check from the service" mean "Unless a service is fully decommissioned before it is removed with this command, it will send a health-check at the appointed time, at which point Automate will re-add it to the services (nodes?) database"?

components/automate-chef-io/content/docs/applications-setup.md
danielsdeleo: The referenced docs are for installing bldr, this procedure is for the hab supervisor. As far as I'm aware, the supervisor will pick up any file with a `.cert` name and use it for SSL connections.

components/automate-cli/cmd/chef-automate/applications.go
danielsdeleo: The user needs to make the service not run any more. How they do that could be retiring physical hardware, terminating a VM or container, or a `hab svc unload`.

The idea here is that the user is decommissioning some habitat services (for example, shutting down a test environment) and they wish for the services to no longer appear in the automate UI and they do not want to wait the normal amount of time before Automate removes the services automatically (default is 7 days).

components/automate-cli/cmd/chef-automate/applications.go
danielsdeleo: > Unless a service is fully decommissioned before it is removed with this command, it will send a health-check at the appointed time, at which point Automate will re-add it to the services (nodes?) database

Yes, exactly this.

components/automate-chef-io/content/docs/applications-setup.md
danielsdeleo: The behavior we're relying on here is documented in this section: https://www.habitat.sh/docs/using-builder/#sts=Using%20Custom%20Certificates the part starting with "The Habitat 0.85.0 release in September 2019..." -- which unfortunately doesn't say anything about what the filenames should be. But in any case, if this was working when you and Kyleen first wrote this, it should still be working now.

susanev:
unsure if this is relevant to this pr, but should we remove the beta label from the applications section of the iam v1 policies docs page?
docs/iam-v1-policies/#applications-beta

components/automate-cli/cmd/chef-automate/applications.go
kagarmoe:
```suggestion
You must fully decommission services by retiring physical hardware, terminating the VM or container, or by using 'hab svc unload' before using the 'remove-svcs' command. Services that are not fully decommissioned will send a health-check at the appointed time and Automate will re-add them to the services database.
```

kagarmoe: It is wordy, but tells a new user what needs done and the consequences of doing it badly.

components/automate-chef-io/content/docs/applications-setup.md
```diff
@@ -121,10 +120,19 @@ Use **one** of these three options:
...
 
-#### Add the Tls Certificate to Your Event Stream Command
+```shell
+cp /hab/svc/automate-load-balancer/data/automate.cert /hab/cache/ssl/automate.cert
+```
+
+**On Windows systems**, store your certs in `C:\hab\cache\ssl`. For example, copy the file with:
+
+```powershell
+Copy-Item "C:\hab\svc\automate-load-balancer\data\automate.cert" -Destination "C:\hab\cache\ssl"
```
kagarmoe: This doesn't use the new step. To do so, change to 
```suggestion
Copy-Item "C:\path\to\automate.cert" -Destination "C:\hab\cache\ssl"
```

components/automate-chef-io/content/docs/applications-setup.md
```diff
@@ -121,10 +120,19 @@ Use **one** of these three options:
 
 Chef Habitat automatically searches that directory and uses the certificate at start up.
 
-* On Linux systems, copy the cert to`/hab/cache/ssl` if you are the root user and `~/.hab/cache/ssl` if you are a non-root user. If you are a non-root user, the full file path is `/Users/username/.hab/cache/ssl/automate.cert`, located in your local user's home directory.
-* On Windows systems, store your certs in `C:\hab\cache\ssl`
+**On Linux systems**, copy the certificate to`/hab/cache/ssl` if you are the root user and `~/.hab/cache/ssl` if you are a non-root user. If you are a non-root user, the full file path is `/Users/username/.hab/cache/ssl/automate.cert`, located in your local user's home directory. Non-root users may need to use `sudo` with the command. For example, as the root user, copy the file with:
 
-#### Add the Tls Certificate to Your Event Stream Command
+```shell
+cp /hab/svc/automate-load-balancer/data/automate.cert /hab/cache/ssl/automate.cert
```
kagarmoe: his doesn't use the new step. To do so, change to:
```suggestion
cp /path/to/automate.cert /hab/cache/ssl/automate.cert
```

components/automate-chef-io/content/docs/applications-setup.md
```diff
@@ -121,10 +120,19 @@ Use **one** of these three options:
 
 Chef Habitat automatically searches that directory and uses the certificate at start up.
 
-* On Linux systems, copy the cert to`/hab/cache/ssl` if you are the root user and `~/.hab/cache/ssl` if you are a non-root user. If you are a non-root user, the full file path is `/Users/username/.hab/cache/ssl/automate.cert`, located in your local user's home directory.
-* On Windows systems, store your certs in `C:\hab\cache\ssl`
+**On Linux systems**, copy the certificate to`/hab/cache/ssl` if you are the root user and `~/.hab/cache/ssl` if you are a non-root user. If you are a non-root user, the full file path is `/Users/username/.hab/cache/ssl/automate.cert`, located in your local user's home directory. Non-root users may need to use `sudo` with the command. For example, as the root user, copy the file with:
```
mjingle:
```suggestion
**On Linux systems**, copy the certificate to`/hab/cache/ssl` if you are the root user, or to `~/.hab/cache/ssl` if you are a non-root user. If you are a non-root user, the full file path is `/Users/username/.hab/cache/ssl/automate.cert`, located in your local user's home directory. Non-root users may need to use `sudo` with the command. For example, as the root user, copy the file with:
```

PR #2286, opened 2019-11-20
Document applications-service protobufs

kmacgugan:
Signed-off-by: kmacgugan <kmacgugan@chef.io>

### :white_check_mark: Checklist

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### :camera: Screenshots, if applicable

PR #2689, opened 2020-01-22
[AUTOMATE-2614] Add API documentation for IAM policy

tylercloke:

Added API documentation for `policy.proto`.

```
cd components/autoamte-chef-io

### if you need to install stuff
brew install hugo
brew tap go-swagger/go-swagger && brew install go-swagger
brew install jq

make sync_swagger_files && make serve
```

Go to http://localhost:1313/docs/api/#tag/Policies

- [ ] Tests added/updated?
- [x] Docs added/updated?


tylercloke:
Why did this change every single `swagger.json`? 🤔

components/automate-chef-io/README.md
```diff
@@ -261,6 +261,22 @@ Block comments support markdown tables:
...
+import "protoc-gen-swagger/options/annotations.proto";
+```
+
+And adding this annotation to the rpc in question:
+
+```go
+option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+  tags: "hidden";
+};
+```
```
phiggins: This is already mentioned in the docs, but got kind of buried: https://github.com/chef/automate/blob/ea926991bf9e335812713a7a8585cff3ae504bc9/components/automate-chef-io/README.md#notes

components/automate-chef-io/content/docs/iam-v2-api-reference.md
```diff
@@ -64,184 +64,9 @@ Type               | May be empty ?    | Allows "*" ?         | May specify (una
...
-Roles that you create are *Custom* roles.
-
-HTTP request              | Description
---------------------------|------------
-GET /roles                | list both *Chef-managed* and *Custom* roles
-GET /roles/{**id**}       | get the specified role
-POST /roles               | create a *Custom* role
-PUT /roles/{**id**}       | update a *Custom* role
-DELETE /roles/{**id**}    | delete a *Custom* role
+These endpoints are now documented [here](https://automate.chef.io/docs/api/#tag/Policies).
```
susanev: no here links please, maybe `API Documentation for Policies` or `Policies API Documentation` for the link text
https://webaim.org/techniques/hypertext/#link_to_link

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -14,6 +15,38 @@ import "components/automate-grpc/protoc-gen-policy/api/annotations.proto";
 import "components/automate-grpc/protoc-gen-policy/iam/annotations.proto";
 
 service Policies {
+  /*
+  Create a new policy
+
+  Creates a new IAM policy used to control permissions in Automate.
+  A policy is composed of one or more statements that act on a set of members.
```
bcmdarroch: s/act on/grant permissions to

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -14,6 +15,38 @@ import "components/automate-grpc/protoc-gen-policy/api/annotations.proto";
...
+  Each statement contains a role as well as a list of projects.
```
bcmdarroch: s/a role/a role or a list of actions

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -14,6 +15,38 @@ import "components/automate-grpc/protoc-gen-policy/api/annotations.proto";
...
+
+  The role defines a set of actions that the statement is scoped to.
+  The project list defines the set of resources that the statement is scoped to.
+  Pass `"projects": ["*"]` to scope a statement to every project.
+
+  The top-level projects list for the policy defines which project(s) the policy belongs to,
```
bcmdarroch: maybe add: for use when filtering policies themselves by the projects they belong to

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -14,6 +15,38 @@ import "components/automate-grpc/protoc-gen-policy/api/annotations.proto";
...
+  whereas the statement level projects list defines which project(s) the statement applies to.
+
+  This example creates a new policy in the (unassigned) project that grants the viewer role
+  on a few projects for all LDAP teams:
+
+  ```
+  {
+    "name": "My Viewer Policy",
```
bcmdarroch: maybe we could use a different example, maybe add another statement with some in-line actions? Since this kind of Project-Viewer policy would be generated when the project is created (though not for both projects I spose)

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -28,6 +61,9 @@ service Policies {
     option (chef.automate.api.iam.policy).resource = "iam:policies";
     option (chef.automate.api.iam.policy).action = "iam:policies:create";
     };
+  /*
+  Get an IAM policy
```
bcmdarroch: do these all need `IAM`? Seems a little redundant, unless we've already been referring to them like that in the docs.

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -78,30 +125,45 @@ service Policies {
...
+        get: "/iam/v2beta/policy_version"
       }
     };
     option (chef.automate.api.policy).resource = "auth:policies";
     option (chef.automate.api.policy).action = "read";
     option (chef.automate.api.iam.policy).resource = "iam:policyVersion";
     option (chef.automate.api.iam.policy).action = "iam:policies:get";
     };
+  /*
+  List policy members
```
bcmdarroch: I'd either drop `IAM` like we do here or keep it in front of every use of `policy`

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -144,6 +217,13 @@ service Policies {
     option (chef.automate.api.iam.policy).resource = "iam:policies:{id}:members";
     option (chef.automate.api.iam.policy).action = "iam:policyMembers:create";
     };
+  /*
+  Create a new role
+
+  Creates a new role to be used in the IAM policies that control permissions in Automate.
+
+  A role defines the scope for a set of actions.
```
bcmdarroch: s/defines the scope for a set of actions/defines the scope of actions in a policy statement

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -116,6 +178,12 @@ service Policies {
     option (chef.automate.api.iam.policy).resource = "iam:policies:{id}:members";
     option (chef.automate.api.iam.policy).action = "iam:policyMembers:update";
     };
+  /*
+  Remove policy members
+
+  Remove specific members from the member list for a specific policy. Silently ignores
+  members that are not part of the member list.
```
bcmdarroch:
```suggestion
  members that are not already part of the member list.
```

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -208,6 +305,13 @@ service Policies {
     option (chef.automate.api.iam.policy).resource = "iam:roles:{id}";
     option (chef.automate.api.iam.policy).action = "iam:roles:update";
   };
+  /*
+  Create a project
+
+  Creates a new project to be used in the IAM policies that control permissions in Automate.
+
+  A project defines the scope for a set of resources. A resource can be in more than one project.
```
bcmdarroch: Not sure if we want to use `resources` here or not. Do we define resources anywhere else in the doc?
```suggestion
  A project defines the scope for a set of resources in a policy statement. A resource can be in more than one project.
```

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -14,6 +15,38 @@ import "components/automate-grpc/protoc-gen-policy/api/annotations.proto";
...
+    "members": ["team:ldap:*"],
+    "statements" : [
+      {
+        "role": "viewer",
+        "projects": ["project1", "project2"]
+      }
+    ],
+    "projects": []
+  }
+  ```
```
phiggins: I'm not sure how to fix this, but the whitespace in this code fence is getting eaten.

Screen Shot 2020-01-24 at 11 04 35 AM

vjeffrey: i think that's true for _all_ the code examples we have in api docs

phiggins:
IDs are inconsistently documented. An ID is kinda self explanatory but it seems odd to documents some and not others.

Screen Shot 2020-01-24 at 11 14 08 AM
Screen Shot 2020-01-24 at 11 14 00 AM

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -78,30 +125,45 @@ service Policies {
...
     option (chef.automate.api.policy).action = "read";
     option (chef.automate.api.iam.policy).resource = "iam:policyVersion";
     option (chef.automate.api.iam.policy).action = "iam:policies:get";
     };
+  /*
+  List policy members
+
+  List all members of a specific policy.
+  */
   rpc ListPolicyMembers (ListPolicyMembersReq) returns (ListPolicyMembersResp) {
```
phiggins: I'm not sure where `ListPolicyMembersResp` is defined but it is not documented in the UI.

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -78,30 +125,45 @@ service Policies {
...
     option (chef.automate.api.policy).action = "read";
     option (chef.automate.api.iam.policy).resource = "iam:policies:{id}:members";
     option (chef.automate.api.iam.policy).action = "iam:policyMembers:get";
     };
+  /*
+  Replace policy members
+
+  Replace the entire member list for a specific policy with a new list.
+  */
   rpc ReplacePolicyMembers (ReplacePolicyMembersReq) returns (ReplacePolicyMembersResp) {
```
phiggins: I'm not sure where `ReplacePolicyMembersResp` is defined but it is not documented in the UI.

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -130,6 +198,11 @@ service Policies {
     option (chef.automate.api.iam.policy).resource = "iam:policies:{id}:members";
     option (chef.automate.api.iam.policy).action = "iam:policyMembers:delete";
     };
+  /*
+  Add policy members
+
+  Add specific members to the member list for a specific policy.
+  */
   rpc AddPolicyMembers (AddPolicyMembersReq) returns (AddPolicyMembersResp) {
```
phiggins: I'm not sure where `AddPolicyMembersResp` is defined but it is not documented in the UI.

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -78,30 +125,45 @@ service Policies {
     option (chef.automate.api.iam.policy).resource = "iam:policies:{id}";
     option (chef.automate.api.iam.policy).action = "iam:policies:update";
     };
+  /*
+  Get IAM version
+
+  Returns the major and minor version of IAM that your automate installation is running.
+  */
   rpc GetPolicyVersion (GetPolicyVersionReq) returns (GetPolicyVersionResp) {
```
phiggins: I'm not sure where `GetPolicyVersionResp` is defined but it needs a `.` at the end of the line for the `version` field.

phiggins:
We might want to consider breaking this into multiple groups, the list in the sidebar is difficult to scan with so many things in one group. With my browser window maximized I can't see everything without scrolling.

Screen Shot 2020-01-24 at 11 34 20 AM

susanev:
i think breaking into `Policies`, `Roles`, and `Projects` sections would be rad!

components/automate-gateway/api/iam/v2/policy.proto
tylercloke: That's what I've been thinking.

components/automate-gateway/api/iam/v2/policy.proto
tylercloke: Been trying to avoid inline actions but I'll add another statement with a custom role.

tylercloke:
@phiggins the main inconsistencies in how ID is documented is on create vs update. Mostly trying to document that it can't be changed on create and can't be updated on update.

components/automate-gateway/api/iam/v2/policy.proto
tylercloke: What specifically are you looking for here? For a lot of these, the request and response woulda just been copy pasta'ed content. Is that what we want?

tylercloke:
@phiggins it looks like the groupings are based on protobuf service? Is there a different way to break things up? Breaking the service up is a much bigger change.

components/automate-gateway/api/iam/v2/policy.proto
phiggins: The API docs are a reference so a user will likely seek out a particular endpoint rather than reading the doc top to bottom. If a user seeks out a specific endpoint they will likely not have seen a particular description or explanation if it is elsewhere in the doc.

phiggins:
> @phiggins it looks like the groupings are based on protobuf service? Is there a different way to break things up? Breaking the service up is a much bigger change.

The default is one group per service, but this is controllable [with tags](https://github.com/chef/automate/blob/master/components/automate-chef-io/README.md#tags-and-tag-groups). An example of this can be found in the `applications.proto`. #2286

components/automate-gateway/api/iam/v2/common/policy.proto
```diff
@@ -24,42 +31,53 @@ message Statement {
         ALLOW = 0;
         DENY = 1;
     }
+    // Whether the statement allows or denies.
```
blakestier: Above, you designate what the options are for Type. I think that makes sense here, too.

components/automate-gateway/api/iam/v2/common/policy.proto
```diff
@@ -11,11 +11,18 @@ enum Type {
 }
 
 message Policy {
+    // Name for the policy.
     string name = 1;
+    // Unique ID for the policy. Cannot be changed after creation.
     string id = 2;
```
blakestier: Hi i'm late here but I think we should indicate that these IDs are user-specified. That might be part of @phiggins questions around ID definition, as well.

tylercloke: I have `Cannot be changed after creation.` for every create request. Is it not clear that it's user specified from that? Anyway, I don't have a strong opinion about how the IDs should be annotated, but I'm not clear on what people are asking for specifically.

msorens: IMHO the existing text is spot on because _everything_ should be assumed to be user-specified--we should only need to point out something if it is *not*. And that is just what the additional note of "Cannot be changed after creation" does.

components/automate-gateway/api/iam/v2/common/policy.proto
```diff
@@ -11,11 +11,18 @@ enum Type {
...
+    // Whether this policy is user created or chef managed.
+    // One of `CUSTOM`, or `CHEF_MANAGED`, respectively.
```
msorens: Should not have a comma when there are two items--this applies throughout the PR.
```suggestion
    // One of `CUSTOM` or `CHEF_MANAGED`, respectively.
```

components/automate-gateway/api/iam/v2/common/policy.proto
```diff
@@ -24,42 +31,54 @@ message Statement {
         ALLOW = 0;
         DENY = 1;
     }
+    // Whether the statement allows or denies.
+    // One of `ALLOW`, or `DENY`, respectively.
     Effect effect = 1;
 
-    // inline definitions
+    // DEPRECATED: Actions defined inline. Use role instead.
```
msorens: Sorry--when did inline actions become deprecated?

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -14,6 +15,42 @@ import "components/automate-grpc/protoc-gen-policy/api/annotations.proto";
...
+  Each statement contains a role as well as a list of projects.
+
+  The role defines a set of actions that the statement is scoped to.
+  The project list defines the set of resources that the statement is scoped to.
+  Pass `"projects": ["*"]` to scope a statement to every project.
+
+  The policy's top-level projects list defines which project(s) the policy belongs to, for filtering policies by their projects,
+  whereas the statement level projects list defines which project(s) the statement applies to.
+
+  This example creates a new policy in the (unassigned) project that grants the viewer role
```
msorens: Should clearly delineate such objects, here and throughout:
```suggestion
  This example creates a new policy in the `(unassigned)` project that grants the `viewer` role
```

components/automate-gateway/api/iam/v2/common/policy.proto
susanev: if a policy has a role it should not have actions, idk if that needs to be communicated here or not. inline actions are allowed if not using a role, but i would prefer not to encourage there use whenever possible.

components/automate-gateway/api/iam/v2/policy.proto
susanev: folks, do users need to know that unassigned with implemented as a project internally? if not, id prefer to not be referring to unassigned as a project.

components/automate-gateway/api/iam/v2/common/policy.proto
susanev: i think IDs are tricky because it many parts of automate we using global unique identifier.. i think auth stuff is pretty unique in regards to allowing users to specify IDs. id be interested in hearing more from the docs team on this.

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -14,6 +15,42 @@ import "components/automate-grpc/protoc-gen-policy/api/annotations.proto";
...
+
+  Creates a new IAM policy used to control permissions in Automate.
+  A policy is composed of one or more statements that grant permissions to a set of members.
+  Each statement contains a role as well as a list of projects.
+
+  The role defines a set of actions that the statement is scoped to.
+  The project list defines the set of resources that the statement is scoped to.
+  Pass `"projects": ["*"]` to scope a statement to every project.
+
+  The policy's top-level projects list defines which project(s) the policy belongs to, for filtering policies by their projects,
```
msorens:
```suggestion
  A policy's top-level projects list defines which project(s) the policy belongs to (for filtering policies by their projects),
```

components/automate-gateway/api/iam/v2/policy.proto
msorens: Good point, @susanev . So I will amend my comment above:
```suggestion
  This example creates a new policy not associated with any project (because the top-level `projects` property is empty) that grants the `viewer` role
```

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -27,7 +64,13 @@ service Policies {
     option (chef.automate.api.policy).action = "create";
     option (chef.automate.api.iam.policy).resource = "iam:policies";
     option (chef.automate.api.iam.policy).action = "iam:policies:create";
-    };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+			tags: "policies";
+		};
+  };
+  /*
+  Get a policy
```
msorens: We still need to give at least one sentence beyond just the heading--even if it adds no real value--because the docs look broken without it.
image

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -181,7 +307,16 @@ service Policies {
...
     option (chef.automate.api.iam.policy).action = "iam:roles:get";
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+			tags: "roles";
+		};
   };
+  /*
+  Delete a role
+
+  This will cascade delete any statements that only contain this role.
+  If the resulting policy has no statements, it is also deleted.
```
msorens: I would prefer not to say "cascade delete" here; not everyone knows that jargon. Also, it is not consistent because the second line is also talking about a cascade delete yet suggests it is doing something different.
My suggestion:
```
Delete a specified role and removes it from any statements that may have been using it.
If such a statement has no other associated actions, the statement is deleted as well.
Similarly, if that statement removal results in a policy with no other statements,
that policy is removed as well.
```

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -259,7 +432,20 @@ service Policies {
...
   };
+  /*
+  Delete a project
+
+  Deletes a project. Cascade deletes the project from any resources tagged with it.
+  If a resulting resource no longer belongs to any project, it goes into the (unassigned) project.
+
+  Also cascade deletes this project from any project list in all statements.
+  If the resulting project list for a given statement is empty, it is deleted.
+  If the resulting policy has no statements, it is also deleted.
```
msorens: Similar to above, suggest remove "cascade delete".
Per @susanev's suggestion to remove "(unassigned)", I think deleting the second sentence would be fine.

components/automate-gateway/api/iam/v2/policy.proto
blakestier: 🤔

blakestier: I agree that this looks odd, but some of our API calls don't require any parameters at all (The list calls, and they look really weird), and the ID in fetching seems self-explanatory.

I agree we should be consistent tho 🤷‍♂

tylercloke: I'm gonna copy-paste `Get a policy` twice then. There's not much else to say here.

msorens:
Outstanding questions:
1. @tylercloke Similar to your own question, why do we want to say each endpoint supports http and https? I thought it was just https.
image

2. I think you missed the comment above about no "here" links. https://github.com/chef/automate/pull/2689#discussion_r370168853

3. Can we do anything about `additional_bindings` that are generating what seem to be duplicate endpoints (but it is really one for v2 and one for v2beta) ? See https://chefio.slack.com/archives/CB4TMAWCQ/p1580178143015000
image

4. Can we do anything about marking deprecated fields? In the image, the text clearly labels it but the sample at the right does not, making it misleading.
image

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -63,7 +124,15 @@ service Policies {
...
     option (chef.automate.api.iam.policy).action = "iam:policies:delete";
-    };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+			tags: "policies";
+		};
+  };
+  /*
+  Update an existing policy
+
+  Very similar to create except the ID cannot be changed.
```
msorens: Actually, there is more to say about updates--this is adapted from what was deleted in the .md file and should be included here:

"When updating a policy you must supply all of that policy's properties, not just the ones you wish to update.
Properties that you do not include are reset to empty values!

You can use this endpoint to modify _custom_ policies but not _Chef-managed_ policies.

If you just wish to update members on a policy (either custom or Chef-managed), see: Add policy members, Remove policy members, and Replace policy members"

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -77,31 +146,55 @@ service Policies {
...
     option (chef.automate.api.iam.policy).action = "iam:policyMembers:get";
-    };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+			tags: "policies";
+		};
+  };
+  /*
+  Replace policy members
+
+  Replace the entire member list for a specific policy with a new list.
```
msorens: Add after this:
"Ensure each element of the `members` array is in the correct [Member Expression]({{< relref "iam-v2-guide.md#member-expressions" >}}) format."

Is that hugo link going to work here? I don't know. 🤷‍♂ 

That sentence should be added to Replace, Add, and Remove members.

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -115,7 +208,16 @@ service Policies {
...
-    };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+			tags: "policies";
+		};
+  };
+  /*
+  Remove policy members
+
+  Remove specific members from the member list for a specific policy. Silently ignores
+  members that are not already part of the member list.
```
msorens: In addition to my comment above about format, also need this:
"The removed members will still exist within Chef Automate, but are no longer associated with this policy."

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -143,7 +253,17 @@ service Policies {
...
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+			tags: "policies";
+		};
+  };
+  /*
+  Create a new role
+
+  Creates a new role to be used in the policies that control permissions in Automate.
+
+  A role defines the scope of actions in a policy statement.
```
msorens: Adapted from the text deleted from the .md file; please add:
"There are several default *Chef-managed* roles that are essential to the operation of Chef Automate and cannot be altered.
Roles that you create are *Custom* roles, and you have full control about modifying or deleting them.

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -157,7 +277,15 @@ service Policies {
...
     option (chef.automate.api.iam.policy).action = "iam:roles:create";
-    };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+			tags: "roles";
+		};
+  };
+  /*
+  List all roles
+
+  List all roles.
```
msorens:
```suggestion
  List all *Chef-managed* and *Custom* roles.
```

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -193,7 +340,15 @@ service Policies {
...
     option (chef.automate.api.iam.policy).action = "iam:roles:delete";
-    };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+			tags: "roles";
+		};
+  };
+  /*
+  Update a role
+
+  Update an existing role. Very similar to create except the ID cannot be changed.
```
msorens: To be consistent with "Update a policy", delete the first fragment:
```suggestion
  Very similar to create except the ID cannot be changed.
```
And then add:
"When updating a role, supply all of a role's properties, not just the ones you wish to update. Properties that you do not include are reset to empty values."

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -129,7 +231,15 @@ service Policies {
...
     option (chef.automate.api.iam.policy).action = "iam:policyMembers:delete";
-    };
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+			tags: "policies";
+		};
+  };
+  /*
+  Add policy members
+
+  Add specific members to the member list for a specific policy.
```
msorens: Add after this:

"Ensure each element of the `members` array is in the correct [Member Expression]({{< relref "iam-v2-guide.md#member-expressions" >}}) format."

susanev:

kagarmoe:
@susanev @tylercloke I rebased, but the file count looks really off. I have all my changes locally, so it is OK to roll back to a state before any of my commits.

susanev:
@msorens will add a commit to indicate what fields are required when time allows

susanev:
@kagarmoe required fields are now labeled, please let us know if you need us to do anything else or if you have any more questions

susanev:
closing in favor of #2982

Issue #2635, opened 2020-01-14
Document Input Parameters to Applications Tab APIs


danielsdeleo:
## User Story

This is for end-users who integrate Automate with their own internal systems via API usage.

In a previous documentation pass, we added annotations for each of the elements in the API protobuf messages, but for some inputs, the docs are not complete enough for a user to understand how to use the API. In particular, our filter parameters are kinda opaque enums that need comprehensive documentation.

## Definition of Done

An end user can construct URL query parameters to filter service groups and/or services by all of the properties that we support in the UI by reading the documentation and not needing to use the source code or use developer tools to inspect the UI.

PR #2705, opened 2020-01-23
Document Apps API filter parameters

danielsdeleo:

The `filters` fields in the Applications API are free-form text at the protobuf level but then are validated like an enum inside the applications service. We need to tell users what the valid values are and what they mean so they can use the API without having to reverse engineer it from the browser.

#2635
#2622

### DIFF
--- a/api/external/applications/applications.proto
+++ b/api/external/applications/applications.proto
@@ -307,6 +307,28 @@ service ApplicationsService {
 // Request message for listing services.
 message ServicesReq {
   // Applies search filters, in the format of `fieldname:value`.
+  // Valid filter fieldnames are:
+  // * `origin`: origin component of the service's package identifier
+  // * `service`: the name component of the service's package identifier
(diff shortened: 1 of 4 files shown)


api/external/applications/applications.proto
```diff
@@ -307,6 +307,28 @@ service ApplicationsService {
...
+  // status filter parameters are:
+  // * `status:disconnected`: only return services in the disconnected state
+  // * `status:critical`: only return services that are returning a "critical"
+  //   healthcheck result
+  // * `status:unknown`: only return services that are returning an "unknown"
+  //   healthcheck result
+  // * `status:warning`: only return services that are returning a "warning"
+  //   healthcheck result
+  // * `status:ok`: only return services that are returning "ok" health check
+  //   results
```
phiggins: These lists would be easier to scan if they were alphabetized.

PR #2706, opened 2020-01-23
Fix broken patch config during restore

jaym:
It seems to have broken in #2344
Basically, in most cases, the code ends up discarding the requested
patch config because it returns early. The only case when this doesn't
happen is if you're doing backups to S3


thomascate:
Just a note that this is extremely dangerous, as if the user is using an external Postgres or Elasticsearch server. The restore will use the credentials and IPs that are in the backup, not the ones provided. So if a customer is restoring a backup from prod into a test environment, it will connect to their production database servers and start dropping tables and restoring data.

PR #2708, opened 2020-01-24
Two notifications received when removing a user from a team

msorens:

Upon removing a user, we were getting two notifications in the UI:
image

This is due to the fact that the actual delete operation was dispatched twice, not that the response was posted twice.

Technical details:
The problem here with "onSelectionChange" is that when you select a new option, it both selects the new option and then deselects the old one. So you get one event with the new option and a second event with the old option.
The fix: Check for the new option selection only, the one with `isUserInput` being true.

Reference: https://github.com/angular/components/issues/4094

### :+1: Definition of Done
Only one dispatch, and thus one notification in the UI.

Rebuild automate-ui
Navigate to Users
Using the control menu for any user in the list, select <kbd>Remove User</kbd>.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/modules/user/user-table/user-table.component.ts
+++ b/components/automate-ui/src/app/modules/user/user-table/user-table.component.ts
@@ -1,4 +1,5 @@
 import { Component, Input, Output, OnInit, EventEmitter } from '@angular/core';
+import { ChefKeyboardEvent } from 'app/types/material-types';
 import { User } from 'app/entities/users/user.model';
 
 @Component({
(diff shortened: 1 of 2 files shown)


components/automate-ui/src/app/modules/user/user-table/user-table.component.ts
```diff
@@ -30,4 +31,10 @@ export class UserTableComponent implements OnInit {
     this.getPermissionsPath = [this.baseUrl, 'get'];
     this.createPermissionsPath = [this.baseUrl, 'post'];
   }
+
+  deleteUser($event: ChefKeyboardEvent, user: User) {
```
msorens: See the PR preamble for the context of this change.

PR #2704, opened 2020-01-23
[automate-2629] add users for team workflow

msorens:
### :nut_and_bolt: Description: What code changed, and why?

Here's the user story:
We want to add users to a particular team. We drill down to the team details, select "Add Users" and, lo and behold, all users that could be added have already been added, so there are no users to add.
But now there is a new message, with a big, enticing button to press.
image

The goal of this PR is for that button to open the self-same user creation modal just recently updated for the users list page. Only now, the modal opens on top of the team's "Add User" dialog (left) -- and it is a dialog in a dialog (right).
image

Problem: that recently finished modal was not self-contained. No fault of that development--none of our existing creation/deletion modals are. They all require intricate wiring and handlers on the base page invoking them.
In order to avoid duplicating all that identical boilerplate, I am introducing a new design pattern: **the self-contained modal**. 

* Instead of 5 or 6 or 7 inputs and outputs, it needs just one: a signal to open and the modal manages itself. 
```
<app-create-user-modal [openEvent]="openUserModal"></app-create-user-modal>
```
* Instead of several methods, variables, and event emitters to manage communication between the modal and the base page, there is just one:
```
openModal(): void {
    this.openUserModal.emit(true);
  }
```
These improve (a) encapsulation and (b) adherence to the Single Responsibility Principle (the users' list should not need to know details about adding a user).

Once I had settled on the pattern implementation was very straightforward, thanks to the new modal (thanks @bcmdarroch and @blakestier !). Most of it was simply moving code from `UserFormComponent` and `UserManagementComponent` into the new `CreateUserModalComponent`. The rest was simplifying because we no longer need to manage the larger set of inputs/outputs to the modal.

### :chains: Related Resources
PR #2654 User modal update

### :+1: Definition of Done
You can add new local users from the teams "Add Users" page when the current list has been exhausted.

### :athletic_shoe: How to Build and Test the Change

Rebuild stencil library and automate-ui server:
```
cd components/automate-ui
make update-ui-lib
make serve
```

1. Navigate to Teams >> (team) >> Add Users.
2. If there are any users available to add, add them all, then re-open Add Users.
3. Select <kbd>Create User</kbd> and exercise the dialog. Do not forgot to try error conditions (attempting to add a user with an existing username, adding a password of "eeeeeeee", etc.)
4. For a successful addition, the dialog will close and the new user will appear in the list, ready to be added to the team.
4. Navigate to Users, select <kbd>Create User</kbd>, and exercise the dialog.

### :white_check_mark: Checklist

[...]

PR #2713, opened 2020-01-24
[automate-1523] fix password match error

bcmdarroch:
Fixes when the `password must match` error shows up

branched off: https://github.com/chef/automate/pull/2704 (merged and rebased)

### :+1: Definition of Done
on both the user create modal and user details page:
- [x] Whenever the two password fields do not match an error should be presented to the user.
- [x] When they match (whether the first is corrected to match the second or the reverse), the error should go away and the save button should be enabled.
- [x] bonus: each field throws a Required error if the user enters all spaces then leaves the input

Create User Modal
1. Sign in as admin
2. Navigate to Settings > Users
3. Use the Create User button
4. Enter a password, and a matching password in the confirm password box
5. Now go back to the first password box and add/delete a letter
6. Keep changing the password and confirm password fields until they match and the save button is enabled

User Details/User Profile Page

1. Click on a user to view their details page and click on the Reset Pasword tab (or click on the Profile link in the top right)
2. Enter in a password
3. Enter the identical password with one extra character on the end
4. You will see the "Passwords must match" error and the save is disabled
5. Add the additional character to the above password field
6. Now they match and the save button is enabled

- [x] Tests added/updated?

### DIFF
--- a/components/automate-ui/src/app/page-components/create-user-modal/create-user-modal.component.ts
+++ b/components/automate-ui/src/app/page-components/create-user-modal/create-user-modal.component.ts
@@ -39,7 +39,7 @@ export class CreateUserModalComponent implements OnInit, OnDestroy {
     fb: FormBuilder
   ) {
     this.createUserForm = fb.group({
-      // Must stay in sync with error checks in user-form.component.html
+      // Must stay in sync with error checks in create-user-modal.component.html
       displayName: ['', [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]],
       username: ['', [Validators.required, Validators.pattern(USERNAME_PATTERN)]],
       // length validator must be consistent with
@@ -126,6 +126,10 @@ export class CreateUserModalComponent implements OnInit, OnDestroy {
   }
 
   handlePasswordInput(event: KeyboardEvent): void {
+    // This handles the case where a user first enters the password and confirmPassword
+    // then goes back and changes the original password.
+    // Since the match validator is only activated by changes to confirmPassword,
+    // we have to manually revalidate confirmPassword here
     this.createUserForm.get('confirmPassword').updateValueAndValidity();
     if (!this.isNavigationKey(event)) {
       this.passwordError = false;
@@ -141,7 +145,7 @@ export class CreateUserModalComponent implements OnInit, OnDestroy {
       this.createUserForm.get(field).dirty);
   }
 
-  hasAttemptedInputWithError(field: string): boolean {
+  hasAttemptedInputWithRequiredError(field: string): boolean {
     return (this.createUserForm.get(field).hasError('required') &&
       (this.createUserForm.get(field).touched || this.createUserForm.get(field).dirty));
   }
--- a/components/automate-ui/src/app/page-components/create-user-modal/create-user-modal.component.ts
+++ b/components/automate-ui/src/app/page-components/create-user-modal/create-user-modal.component.ts
@@ -125,6 +125,13 @@ export class CreateUserModalComponent implements OnInit, OnDestroy {
     }
   }
 
+  handlePasswordInput(event: KeyboardEvent): void {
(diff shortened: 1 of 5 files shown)


blakestier:
Yay totally works.

components/automate-ui/src/app/page-components/create-user-modal/create-user-modal.component.ts
```diff
@@ -125,6 +125,13 @@ export class CreateUserModalComponent implements OnInit, OnDestroy {
     }
   }
 
+  handlePasswordInput(event: KeyboardEvent): void {
+    this.createUserForm.get('confirmPassword').updateValueAndValidity();
```
msorens: I was wondering why this situation is unique, requiring bringing in this extra call. But I commented it out locally and all errors reacted just the same, so i could not find a need for it. Do you have a use case where this becomes significant?

(Same question for user-details.component -- commenting it out does not seem to change any behavior.)

components/automate-ui/src/app/page-components/create-user-modal/create-user-modal.component.ts
```diff
@@ -135,10 +142,8 @@ export class CreateUserModalComponent implements OnInit, OnDestroy {
   }
 
   hasAttemptedInputWithError(field: string): boolean {
-    return ((this.createUserForm.get(field).hasError('required') &&
-      this.createUserForm.get(field).touched) ||
-      (this.createUserForm.get(field).hasError('pattern') &&
```
msorens: You deleted `pattern` here (and that's good, because it is redundant with the `minLength` check) but you also need to make the corresponding change in the Validator itself, i.e., delete this line: https://github.com/chef/automate/blob/2e5e0ec16d4bda3b3f8b5f1e601181bc71a4fa27/components/automate-ui/src/app/page-components/create-user-modal/create-user-modal.component.ts#L49

components/automate-ui/src/app/page-components/create-user-modal/create-user-modal.component.ts
```diff
@@ -125,6 +125,13 @@ export class CreateUserModalComponent implements OnInit, OnDestroy {
     }
   }
```
msorens: Please fix--this was my oversight:
Change this:
https://github.com/chef/automate/blob/2e5e0ec16d4bda3b3f8b5f1e601181bc71a4fa27/components/automate-ui/src/app/page-components/create-user-modal/create-user-modal.component.ts#L42
to this:
```
// Must stay in sync with error checks in create-user-modal.component.html
```

components/automate-ui/src/app/page-components/create-user-modal/create-user-modal.component.ts
bcmdarroch: This specifically handles the case where a user enters a `password` and `confirmPassword`, then goes back and changes the original password. Since the match validator is only activated by changes to `confirmPassword`, we have to manually trigger a new check on confirmPassword here

components/automate-ui/src/app/page-components/create-user-modal/create-user-modal.component.ts
bcmdarroch: I'm going to keep the NON_BLANK validator in since it prevents passwords that are all spaces. However, it wasn't triggering any error alerts so I also updated the `field is required` error conditional so it will display for all space values.

components/automate-ui/src/app/page-components/create-user-modal/create-user-modal.component.ts
msorens: Ah, quite so! Thanks for pointing that out.
Could you add a brief in-code comment to that effect so others will know as well?

components/automate-ui/src/app/page-components/create-user-modal/create-user-modal.component.ts
msorens: That's even better! 😁

Issue #2660, opened 2020-01-18
User Display Name not changing in Menu


lancewf:
## User Story
When a user updates their display name it will not update in the menu until they sign out and sign back in. The Menu is shown below.
 ![menu.png](https://images.zenhubusercontent.com/5d1a3fbf1449416ed898598a/dbae98e3-1187-478a-a681-7f770d59de30)

## Definition of Done
When changing a users display name from the user's profile page or admin user's page the name should be updated on the menu. 

## Demo Script / Repro Steps
1. Open the users profile page by clicking the menu on the top right. ![image.png](https://images.zenhubusercontent.com/5d1a3fbf1449416ed898598a/6f2179ed-553b-4f63-a8eb-f04476298797). Then clicking Profile. 
1. Update the Display Name and save. 
1. Check the name in the menu to ensure that it has changed.

lancewf:
This is the component where the name is displayed. 
https://github.com/chef/automate/blob/6b3840d29126d224a816ac118666193cfb3fb131/components/automate-ui/src/app/page-components/profile/profile.component.ts#L100

PR #2714, opened 2020-01-24
Syncing changes to the user menu display name

lancewf:
When a user changes their user display name from the profile page or setting-> users page, the display name is updated in the menu.

This one simple bug needed a lot of changes to complete. These changes add the concept of the logged-in user to the UI. A new section to the NGRX store was created that stores information about this user. The logged-in user's ID is first placed in the store from the chef-seesion.service. Where the ID is updated by each session refresh. 

The user-details component has been updated to have three different states: admin editing other users, profile page, and admin editing their own user. The admin editing other users has not changed from before. The profile page is different in that it updates the newly created logged-in user NGRX store. The admin editing their own user is different in that when a user tries to update their password they must enter the previous password. This also updates the logged-in NGRX store. 

The main change is to the profile component that displays the logged-in user's information in the top-right menu. Before the user's display name was pulled from the session when they first logged in. The problem was that it could only be changed by logging out and back in. Now, the display name is pulled from the newly created logged-in NGRX store. Therefore when changes to the display name are made the menu updates. 

A flag to say if the user is a local or non-local user was also added. For non-local users (for example SAML) they do not have a configurable display name through Automate. Because of this, we have to remove the profile page link and not try to pull the logged-in user's information from the APIs.  To figure out if a user is local or not we are looking at the federated claims field on the session ID token. The federated claims have a field called connector_id that has the term 'local' for local users and 'saml' for non-local users. 

https://github.com/chef/automate/issues/2660

### :+1: Definition of Done
* Change display name from the user profile page
* Change display name from the setting -> user page
* Change the logged-in user's password from the setting -> user page needs the user's previous password

1. Start the UI
1. Go to the user profile page and update the display name
1. Ensure the menu name is updated. 
1. Go to the logged-in user's page from the settings -> users page and update the display name. 
1. Ensure the menu name is updated. 
1. Go to the logged-in user's page from the settings -> users page and update the password. 
1. Ensure the previous password is needed. 

- [x] Tests added/updated?
- [ ] Docs added/updated?


components/automate-ui/src/app/services/chef-session/chef-session.service.ts
```diff
@@ -120,12 +126,17 @@ export class ChefSessionService implements CanActivate {
     }
     this.user = <ChefSessionUser>JSON.parse(localStorage.getItem(sessionKey));
     this.user.telemetry_enabled = this.fetchTelemetryPreference();
+    this.store.dispatch(new SetUserSelfID({ id: this.user.username }));
```
lancewf: Setting the user ID when pulled from the local storage.

components/automate-ui/src/app/services/chef-session/chef-session.service.ts
```diff
@@ -120,12 +126,17 @@ export class ChefSessionService implements CanActivate {
...
   }
 
   // setSession sets ChefSession's session data in localStorage for having it
   // available for setSessionOrRedirectToLogin() (part of ChefSession's
   // constructor)
   setSession(uuid, fullname, username, id_token: string, groups: Array<string>): void {
+    if (!this.user || username !== this.user.username) {
+      this.store.dispatch(new SetUserSelfID({ id: username }));
+    }
+
```
lancewf: Updating the user ID when the session is set.

components/automate-ui/src/app/entities/users/user.actions.ts
```diff
@@ -2,7 +2,6 @@ import { Action } from '@ngrx/store';
 import { HttpErrorResponse } from '@angular/common/http';
 
 import { User } from './user.model';
-import { SelfUser } from './userself.model';
```
lancewf: Removing all the userself code from the user ngrx store files. This code was moved to the newly created userself ngrx store files.

components/automate-ui/src/app/entities/users/userself.actions.ts
```diff
@@ -0,0 +1,79 @@
+import { Action } from '@ngrx/store';
```
lancewf: The userself logic was moved from the user ngrx store files because we need to register for changes to the logged-in user.

components/automate-ui/src/app/ngrx.effects.ts
```diff
@@ -23,6 +23,7 @@ import { ScannerEffects } from './pages/+compliance/+scanner/state/scanner.effec
 import { SidebarEffects } from './services/sidebar/sidebar.effects';
 import { TeamEffects } from './entities/teams/team.effects';
 import { UserEffects } from './entities/users/user.effects';
+import { UserSelfEffects } from './entities/users/userself.effects';
```
lancewf: This file and the next, setup the new userself ngrx store logic with the global storage.

components/automate-ui/src/app/page-components/profile/profile.component.html
```diff
@@ -13,10 +13,15 @@
...
   <chef-dropdown class="dropdown" [attr.visible]="dropdownVisible" (keydown.esc)="closeDropdown()">
     <ul class="dropdown-list">
-      <li class="dropdown-list-item">
-        Signed in as <strong class="dropdown-display-name">{{ displayName }}</strong>
-      </li>
+      <app-authorized not [allOf]="['/auth/users/{username}', 'get', userName]">
+        <li class="dropdown-list-item">
+          Signed in as <strong class="dropdown-display-name">{{ userName }}</strong>
+        </li>
+      </app-authorized>
```
lancewf: Non-local user doe not have access to '/auth/users/'. For these users, we use the userName as the display name and do not show the profile page.

components/automate-ui/src/app/helpers/jwt/jwt.ts
```diff
@@ -19,6 +19,11 @@ export interface IDToken {
   email_verified: boolean;
   name: string;
   groups: Array<string>;
+  federated_claims: FederatedClaims;
+}
+
+interface FederatedClaims {
+  connector_id: string;
```
lancewf: Adding the `federated_claims.connector_id` to decern if the logged-in user is a local user or non-local user. Example values for this field are 'local' and 'saml'.

components/automate-ui/src/app/modules/user/user-details/user-details.component.ts
```diff
@@ -34,76 +41,51 @@ export type UserTabName = 'password' | 'details';
...
-  public displayNameForm: FormGroup;
-  public passwordForm: FormGroup;
   private isDestroyed = new Subject<boolean>();
+  public loading = true;
   public tabValue: UserTabName = 'details';
   private url: string;
-  public saveSuccessful = false;
-  public saveInProgress = false;
-  public loadingGetUser$: Observable<boolean>;
+  public userDetails: UserDetails;
```
lancewf: The UserDetails represents three cases: Admin user editing other users, admin user editing their own user, and profile page.

components/automate-ui/src/app/modules/user/user-details/user-details.component.ts
```diff
@@ -34,76 +41,51 @@ export type UserTabName = 'password' | 'details';
...
+            // Get the user ID from the URL
+            this.store.pipe(
+              select(routeParams),
+              pluck('id'),
+              filter(identity),
+              first()),
+            // Get the current logged in user's ID
+            this.store.select(userSelfId)
+          ]).subscribe(([routeId, currentUserId]: [string, string]) => {
+            if (routeId === currentUserId) {
```
lancewf: Checking if the user logged in is the user being edited on this page.

components/automate-ui/src/app/modules/user/user-details/user-details.component.ts
```diff
@@ -120,13 +102,107 @@ export class UserDetailsComponent implements OnInit, OnDestroy {
...
+  }
+
+  public onSelectedTab(event: { target: { value: UserTabName } }): void {
+    this.tabValue = event.target.value;
+    // Drop the previous fragment and add the incoming fragment.
+    this.router.navigate([this.url.split('#')[0]], { fragment: event.target.value });
+  }
+}
+
+interface UserDetails {
```
lancewf: Below there are three classes that implement this interface that represent the three cases from above.

e2e/cypress/support/commands.ts
```diff
@@ -331,8 +331,6 @@ function LoginHelper(username: string) {
   cy.server();
   // the gloabl permissions for the user that populates the initial permissions cache
   cy.route('GET', '/api/v0/auth/introspect').as('getAuthPopulateCache');
-  // the parameterized permissions, called for the specific page loaded
-  cy.route('POST', '/api/v0/auth/introspect').as('getAuthParameterized');
```
lancewf: This call was removed in the profile component.

components/automate-ui/src/app/modules/user/user-details/user-details.component.ts
msorens: I like the approach! Have you considered abstract classes? Seems like that would help you make it even more tidy. What it gives you "for free" is that the methods that are invariant do not have to be duplicated. You define just those methods in the abstract class, leaving the customized methods to be defined in the derived classes. Reference: https://www.typescriptlang.org/docs/handbook/classes.html#abstract-classes

components/automate-ui/src/app/page-components/profile/profile.component.html
```diff
@@ -16,16 +16,14 @@
       <li class="dropdown-list-item">
         Signed in as <strong class="dropdown-display-name">{{ displayName }}</strong>
       </li>
-      <app-authorized [allOf]="['/auth/users/{username}', 'get', userName]">
```
msorens: Was the permissioning not appropriate here? Wondering why it was removed.

components/automate-ui/src/app/page-components/profile/profile.component.ts
```diff
@@ -29,18 +42,45 @@ export class ProfileComponent implements OnInit, OnDestroy {
...
+        map((status: EntityStatus) =>  status !== EntityStatus.loadingSuccess));
+
+      combineLatest([
+        this.loading$,
+        this.store.select(userSelf)
+      ]).pipe(
+        filter(([loadingUserData, user]) => !loadingUserData && !isNil(user)),
+        map(([_, user]) => user.name),
+        distinctUntilChanged(),
+        takeUntil(this.isDestroyed)).subscribe((displayName) => {
```
msorens:
```suggestion
        takeUntil(this.isDestroyed)).subscribe(displayName => {
```

components/automate-ui/src/app/page-components/profile/profile.component.html
lancewf: I was under the impression that this was only here to check if the user was a local user or not. This was to decide if we should show the profile link. We now have the `isLocalUser` variable that does this check.

components/automate-ui/src/app/entities/users/userself.effects.ts
```diff
@@ -0,0 +1,107 @@
...
+      return new CreateNotification({
+        type: Type.error,
+        message: `Could not get logged in user: ${msg || error}`
+      });
+    }));
+
+  @Effect()
+  setUserSelfId$ = this.actions$.pipe(
+    ofType(UserSelfActionTypes.SET_ID),
+    map(() => new GetUserSelf()));
```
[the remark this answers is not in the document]
lancewf: Yes, trigger an update of the userself when the ID is updated.

components/automate-ui/src/app/entities/users/userself.effects.ts
```diff
@@ -0,0 +1,107 @@
...
+    }));
+
+  @Effect()
+  updateNameSelf$ = combineLatest([
+    this.actions$.pipe(ofType<UpdateNameSelf>(UserSelfActionTypes.UPDATE_NAME_SELF)),
+    this.store$.select(iamMajorVersion).pipe(filter(identity))])
+    .pipe(
+      mergeMap(([action, version]: [UpdateNameSelf, IAMMajorVersion]) =>
+      this.requests.updateSelf(action.payload, version).pipe(
+        map((resp: SelfUser) => new UpdateNameSelfSuccess(resp)),
```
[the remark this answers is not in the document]
lancewf: Susan did not want this for a successful name change. There is a notification next to the save button.

components/automate-ui/src/app/entities/users/userself.reducer.ts
```diff
@@ -0,0 +1,71 @@
+import { set, pipe } from 'lodash/fp';
+
+import { EntityStatus } from '../entities';
```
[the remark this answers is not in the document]
lancewf: Good to know :+1:

components/automate-ui/src/app/modules/user/user-details/user-details.component.spec.ts
```diff
@@ -31,115 +36,264 @@ describe('UserDetailsComponent', () => {
...
-
-  const initialState = {
-    ...defaultInitialState,
-    router: {
-      ...defaultRouterState,
-      state: {
-        ...defaultRouterRouterState,
-        url: '/settings/users/alice',
-        params: { id: 'alice' }
+  describe('logged-in user opening thier user page', () => {
```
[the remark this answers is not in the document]
lancewf: I have been doing that a lot today. Good catch.

components/automate-ui/src/app/modules/user/user-details/user-details.component.spec.ts
```diff
@@ -31,115 +36,264 @@ describe('UserDetailsComponent', () => {
...
+        expect(component).toBeTruthy();
+      });
+
+      it('defaults to showing details section', () => {
+        expect(component.tabValue).toBe('details');
+      });
 
-    it('picks up the user from state', () => {
-      expect(component.user).toEqual(user);
+      it('show password section when password tab is selected', () => {
```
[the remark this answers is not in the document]
lancewf: :thinking: I get it. It took me forever to put my English hat on to understand what you were saying.

components/automate-ui/src/app/modules/user/user-details/user-details.component.ts
lancewf: I don't think I have done abstract classes in javascript before. I will give it a try.

lancewf:
@msorens Thank you for the review. I took you up on switching to using an abstract class. It does remove more of the duplicated code.

components/automate-ui/src/app/entities/users/userself.effects.ts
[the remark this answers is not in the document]
msorens: Ah, that's fine... then I think it would be prudent to add an in-code to that effect (pun intended 🤣 ).

components/automate-ui/src/app/modules/user/user-details/user-details.component.spec.ts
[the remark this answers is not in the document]
msorens: Oh, sorry, if I was being too terse. 🐑 😁

components/automate-ui/src/app/page-components/profile/profile.component.html
msorens: No, it is not checking local-or-not status.  That was a standard permission check to determine if the current user (local or not) have permission to do a get on the specified user (which must be a local user).

components/automate-ui/src/app/entities/users/userself.effects.ts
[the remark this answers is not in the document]
lancewf: Add a comment saying why there is not a Success effect?

components/automate-ui/src/app/modules/user/user-details/user-details.component.ts
```diff
@@ -34,99 +41,183 @@ export type UserTabName = 'password' | 'details';
...
+// This view is used when a user goes to another users detail page.
+class UserAdminDetails extends UserDetails {
+  public loading$: Observable<boolean>;
+  public user: User;
+  public saveSuccessful = false;
+  public saveInProgress = false;
+  public displayNameForm: FormGroup;
+  public passwordForm: FormGroup;
+  public showBreadcrumbs = true;
+  public showPreviousPassword = false;
```
msorens: Yes--looks better with an abstract class! 👍 
One more streamlining note here; while you can declare these as abstract properties and then instantiate them, there is no need because you are instantiating them identically. So you can move this block into the abstract class and not have to repeat it in each derived class.

components/automate-ui/src/app/page-components/profile/profile.component.html
lancewf: What is the reason for the check? It looked like to me to see if the profiles page could be displayed.

components/automate-ui/src/app/modules/user/user-details/user-details.component.ts
```diff
@@ -34,99 +41,183 @@ export type UserTabName = 'password' | 'details';
...
+    this.saveInProgress = true;
+    const password = this.passwordForm.get('newPassword').value;
+    this.store.dispatch(new UpdatePasswordUser({ ...this.user, password }));
+  }
+
+  public saveDisplayName(): void {
+    this.saveSuccessful = false;
+    this.saveInProgress = true;
+    const name = this.displayNameForm.get('displayName').value.trim();
+    this.store.dispatch(new UpdateNameUser({ ...this.user, name }));
```
msorens: Last tidbit about streamlining abstract classes: by defining the method first in the base class...
```
 protected saveDisplayName(): void {
    this.saveSuccessful = false;
    this.saveInProgress = true;
    const name = this.displayNameForm.get('displayName').value.trim();
 }
```
Then this method in the derived class becomes this--which very clearly shows what is unique about it. 
```
public saveDisplayName(): void {
  super.saveDisplayName();
  this.store.dispatch(new UpdateNameUser({ ...this.user, name }));
}
```
Just a suggestion.

lancewf: I think this makes what is going on more confusing because I would have to look between the two functions to really see what is going on. 

I could see doing the below though. 
```
// base class
 public saveDisplayName(): void {
    this.saveSuccessful = false;
    this.saveInProgress = true;
    const name = this.displayNameForm.get('displayName').value.trim();
    this.store.dispatch(this.createUpdateUserNameAction( name ));
 }

abstract createUpdateUserNameAction(name: string): Action;
```
```
// derived class. 
createUpdateUserNameAction(name: string): Action {
   return new UpdateNameUser({ ...this.user, name });
}
```

msorens: Agreed--I like that better! 👍

components/automate-ui/src/app/modules/user/user-details/user-details.component.ts
lancewf: Thanks. I moved all the fields to the abstract class but the last two because they are different for each derived class.

components/automate-ui/src/app/entities/users/userself.effects.ts
```diff
@@ -0,0 +1,111 @@
...
+      mergeMap(([_action, userId]) =>
+        this.requests.getUser(userId).pipe(
+        map((resp: SelfUser) => new GetUserSelfSuccess(resp)),
+        catchError((error: HttpErrorResponse) => of(new GetUserSelfFailure(error)))))
+      );
+
+  @Effect()
+  getUserSelfFailure$ = this.actions$.pipe(
+    ofType(UserSelfActionTypes.GET_FAILURE),
+    // when first logging in, we see this failure with no error.
```
bcmdarroch: 🤔 does this always happen?

components/automate-ui/src/app/ngrx.reducers.ts
```diff
@@ -170,7 +172,8 @@ export const defaultInitialState = {
   serviceGroups: serviceGroups.ServiceGroupEntityInitialState,
   teams: teamEntity.TeamEntityInitialState,
   userperms: permEntity.initialState,
-  users: userEntity.UserEntityInitialState
+  users: userEntity.UserEntityInitialState,
+  userself: userSelfEntity.UserSelfEntityInitialState
```
bcmdarroch: should this be camel cased like below?

components/automate-ui/src/app/entities/users/userself.effects.ts
lancewf: Yeah, it always happens on log in.

components/automate-ui/src/app/ngrx.reducers.ts
lancewf: I don't know. It should be one way or the other though.

components/automate-ui/src/app/entities/users/userself.effects.ts
lancewf: I think it has something to do with the signin.component setting the user self ID which kicks off getting the user self before automate is fully started. I am guessing the request to get user self is canceled.

components/automate-ui/src/app/ngrx.reducers.ts
lancewf: I think you are right. I will switch to "userSelf".

lancewf: Good catch! I am not sure why that did not break things.

bcmdarroch: hrm ya i would have expected something to break. oh well 🤷‍♀

PR #2721, opened 2020-01-27
Auth 2694/autgen docs tokens proto

blakestier:
# :+1: Definition of Done

Docs are generated from proto files. All tokens protos in the `api` directory are annotated as needed.

Related to https://github.com/chef/automate/pull/2689 

make sync_swagger_files && make serve


components/automate-chef-io/data/docs/api-static/03-tags.swagger.json
```diff
@@ -15,14 +15,22 @@
     {
       "name": "Applications",
       "tags": ["service_groups", "retention"]
-    }
+    },
+    {
+      "name": "Identity and Access Management",
+      "tags": ["policies", "roles", "projects", "tokens"]
```
blakestier: I copied this from https://github.com/chef/automate/pull/2689 but would like them alphabetized when that merges 🤓

phiggins: If it were a multiline list it might help merge conflicts if more things are going to get added.

components/automate-gateway/api/iam/v2/common/tokens.proto
```diff
@@ -4,11 +4,19 @@ package chef.automate.api.iam.v2;
 option go_package = "github.com/chef/automate/components/automate-gateway/api/iam/v2/common";
 
 message Token {
+    // Unique, optionally user-specified ID. Cannot be changed.
```
bcmdarroch: for v2 I think we always want the user to specify the ID. not sure if these docs are supposed to be written with v2 only in mind

susanev: yes v2 only.

components/automate-gateway/api/iam/v2/tokens.proto
```diff
@@ -13,6 +14,26 @@ import "components/automate-grpc/protoc-gen-policy/api/annotations.proto";
 import "components/automate-grpc/protoc-gen-policy/iam/annotations.proto";
 
 service Tokens {
+
+  /*
+  Create a token
+
+  Creates a token.
+  Name, ID are required.
```
bcmdarroch:
```suggestion
  Name and ID are required.
```

components/automate-gateway/api/iam/v2/request/tokens.proto
```diff
@@ -6,22 +6,33 @@ option go_package = "github.com/chef/automate/components/automate-gateway/api/ia
...
 }
 
+
 message GetTokenReq {
     string id = 1;
 }
 
 message UpdateTokenReq {
-    // ID can't be changed; ID used to discover token
+    // Unique, user-specified ID. Cannot be changed.
```
bcmdarroch: Wish there was a way to specify this as a path parameter only so it didn't show up twice 🤔 But that might be Next Level open API

components/automate-gateway/api/iam/v2/common/tokens.proto
blakestier: Oh yes, I thought I changed that 🤔

susanev:
@blakestier oh, a thought. can we refer to these as `API Tokens` instead of just tokens.. or `IAM API Tokens` when appropriate. that will align with how we describe them in the other docs and in the ui.

components/automate-gateway/api/iam/v2/common/tokens.proto
```diff
@@ -4,11 +4,19 @@ package chef.automate.api.iam.v2;
...
     string id = 1;
+    // User-specified name.
```
msorens: Here you are using "user-specified" frequently but the policy endpoints do not. Though my personal preference is to omit that phrase (because seems like it should be on every field), but I would say it is more important to be consistent, one way or the other through all endpoints.

components/automate-gateway/api/iam/v2/common/tokens.proto
```diff
@@ -4,11 +4,19 @@ package chef.automate.api.iam.v2;
...
+    // Unique, optionally user-specified value.
     string value = 3;
+    // Active state. Defaults to true.
+    // If set to false, token will not authenticate.
     bool   active = 4;
+    // Created timestamp.
     string created_at = 5;
+    // Updated timestamp.
     string updated_at = 6;
+    // Array of projects token is in. Empty by default.
```
msorens: Again we should strive for one voice, one way or the other.
Policy endpoints use "The list of projects this policy belongs to." But I like your "Empty by default", too.

(Same consistency comments on the other tokens.proto file too)

components/automate-gateway/api/iam/v2/tokens.proto
```diff
@@ -38,7 +68,18 @@ service Tokens {
...
+      tags: "tokens"
+    };
   };
+
+  /*
+  Update a token
+
+  This PUT operation will overwrite all fields excepting ID, timestamps, and value,
+  including those omitted from the request. Include `name`, `active` and `projects`
+  in the request.
```
msorens: I find it difficult to understand what these two sentences are conveying, and I am familiar with the code. I think users would be lost.

components/automate-gateway/api/iam/v2/tokens.proto
```diff
@@ -26,7 +47,16 @@ service Tokens {
...
     option (chef.automate.api.iam.policy).action = "iam:tokens:create";
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+      tags: "tokens"
+    };
   };
+
+  /*
+  Get a token
+
+  Returns the details for a token.
```
msorens: I commented in the policy PR that the docs looked a bit broken without a one-line like this (though it does not really add value otherwise). You add it for `get` here, but not for `delete` or `list` below. If you want to argue for not having it, that is fine, too; my preference is not a strong preference. But again, let's be consistent throughout all endpoints.

components/automate-gateway/api/iam/v2/tokens.proto
```diff
@@ -64,7 +112,14 @@ service Tokens {
     option (chef.automate.api.policy).action = "delete";
     option (chef.automate.api.iam.policy).resource = "iam:tokens:{id}";
     option (chef.automate.api.iam.policy).action = "iam:tokens:delete";
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+      tags: "tokens"
+    };
   };
+
+  /*
+  List all tokens
```
msorens: Suggest add:
`This command lists both admin and non-admin tokens.`

components/automate-gateway/api/iam/v2/tokens.proto
```diff
@@ -13,6 +14,26 @@ import "components/automate-grpc/protoc-gen-policy/api/annotations.proto";
...
+  Value is auto-generated when not provided.
+  ```
+  {
+    "name": "token 1",
+    "id": "token-1",
+    "active": true,
+    "projects": [
+      "east-region",
+      "west-region"
+    ]
```
msorens: Sorry to harp on a theme: we have examples for these token endpoints but the policy endpoints show no examples. Would suggest to go the route of no examples, because the doc generator automatically provides a sample (without data, but I think it is sufficient and saves all the maintenance burden).

components/automate-gateway/api/iam/v2/tokens.proto
```diff
@@ -13,6 +14,26 @@ import "components/automate-grpc/protoc-gen-policy/api/annotations.proto";
...
 
 service Tokens {
+
+  /*
+  Create a token
+
+  Creates a token.
+  Name, ID are required.
+  Active defaults to true when not specified.
+  Value is auto-generated when not provided.
```
msorens: Please add back this vital text that was removed from the .md file:

Note that this creates **non**-admin tokens that may then be assigned permissions via policies just like users or teams (unless you have already created policies that encompass all tokens using `tokens:*`).

You cannot create admin tokens via the REST API.
Admin tokens can only be created by specifying the `--admin` flag to this `chef-automate` sub-command:

```bash
  chef-automate iam token create <your-token-name> --admin
```

components/automate-gateway/api/iam/v2/common/tokens.proto
blakestier: Yep @tylercloke and I are on it

susanev:

susanev:
@msorens will add a commit to indicate what fields are required when time allows

susanev:
@kagarmoe required fields are now labeled, please let us know if you need us to do anything else or if you have any more questions

susanev:
closing in favor of #2982

PR #2726, opened 2020-01-28
Applying license should immediately be reflected in license notification banner

### LINKED ISSUE #1509: Applying license should immediately be reflected in license notification banner
_Originally reported by @bcmdarroch_

## Overview
Whether getting a new trial license (in which case a license notification banner should appear) or applying a full license (in which case the banner should disappear), that banner change should occur immediately, not at the next polling interval.

## Repro Steps
- remove your license with `remove_dev_license` in the studio
- open the ui
- fill in the form to get a trial license, and register it
- notice that after you register there is not expiration banner at the top, you have to do a browser refresh to see it
    - we want the banner to show up right after you register and the registering modal closes
- similarly, if you see the banner, and use the apply license button in the banner and submit a valid license then the banner should go away without needing to refresh
    - for this part you can get a valid license from the auth team, reach out in #a2-auth-team

vinay033:

Whether getting a new trial license (in which case a license notification banner should appear) or applying a full license (in which case the banner should disappear), that banner change should occur immediately, not at the next polling interval.

fixes #1509
### :+1: Definition of Done
I have added some changes to reflect the license notification banner immediately. 
build components/automate-ui
### :camera: Screenshots
ezgif com-video-to-gif (1)

### DIFF
--- a/components/automate-ui/src/app/modules/license/license-apply/license-apply.component.spec.ts
+++ b/components/automate-ui/src/app/modules/license/license-apply/license-apply.component.spec.ts
@@ -40,7 +40,7 @@ function genLicenseResp(licenseEndDate: moment.Moment): LicenseStatus {
 }
 
 describe('LicenseApplyComponent', () => {
-  let component;
+  let component: LicenseApplyComponent;
 
   it('should be created', () => {
     setup(genLicenseApplyReducer());
--- a/components/automate-ui/src/app/modules/license/license-apply/license-apply.component.spec.ts
+++ b/components/automate-ui/src/app/modules/license/license-apply/license-apply.component.spec.ts
@@ -11,7 +11,7 @@ import { TelemetryService } from 'app/services/telemetry/telemetry.service';
 import { MockChefSessionService } from 'app/testing/mock-chef-session.service';
 import { EntityStatus } from 'app/entities/entities';
 import { runtimeChecks } from 'app/ngrx.reducers';
-import { ApplyStatus, LicenseStatus, FetchStatus } from 'app/entities/license/license.model';
+import { ApplyStatus, LicenseStatus } from 'app/entities/license/license.model';
 import { LicenseApplyComponent } from './license-apply.component';
 
 function genErrorResp(status: number, msg: string): any /* HttpErrorResponse */ {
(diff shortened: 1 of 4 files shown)


susanev:
@vinay033 the banner does not seem to allow dismissal at all anymore... we should still allow folks to dismiss the banner.

vinay033:
> @vinay033 the banner does not seem to allow dismissal at all anymore... we should still allow folks to dismiss the banner.

allow a user to dismiss banner after applying for license?

vinay033:
> toHaveBeenCalled

@msorens  I understand, sure I will add, 
Thanks for the suggestion :)

components/automate-ui/src/app/modules/license/license-apply/license-apply.component.spec.ts
```diff
@@ -40,7 +40,7 @@ function genLicenseResp(licenseEndDate: moment.Moment): LicenseStatus {
 }
 
 describe('LicenseApplyComponent', () => {
-  let component: LicenseApplyComponent;
+  let component;
```
msorens: Please revert this--should always use specific types rather than declaring variables as type `any`, which is what this does.

Same change in license-lockout.component.spec.ts, too.

vinay033: actually `licenseFacade` is private and `let component: LicenseApplyComponent;` is not allowed to call directly, if I use `let component: LicenseApplyComponent;` then I am getting
`Property 'licenseFacade' is private and only accessible within class 'LicenseLockoutComponent'.ts(2341)
`

msorens: Ah, so that means the code was actually hiding that quite valid error message--you should not be able to access `licenseFacade` in the test. I understand why you made the change--ideally one should not have to modify the code-under-test just for the benefit of the test.
However, it is more important not to hide any valid errors--a somewhat better solution in this case is to make `licenseFacade` public. (Yes, it is a not-ideal change to make, but it is the lesser of two evils.)

vinay033: Yes, I had this solution before, but I didn't change it.

vinay033: I have updated the changes.

PR #2730, opened 2020-01-28
Migrate TLS from load balancer config to global config

jaym:
The TLS frontend config is used by event-gateway. It reads it from the
global config. This config was never migrated, so old installs spew
errors when trying to configure NATS:
```
Jan 28 16:10:16 a2-local-inplace-upgrade-acceptance.cd.chef.co hab[30015]: event-gateway.default(O): [11427] [ERR] 127.0.0.1:58280 - cid:19445 - TLS handshake error: tls: no certificates configured
Jan 28 16:10:16 a2-local-inplace-upgrade-acceptance.cd.chef.co hab[30015]: event-gateway.default hook[health-check]:(HK): checking connectivity: connecting to NATS server "nats://HEALTHCHECK-4o28Fgfu3od9NLA6BDnjU4BtE44=@localhost:4222": remote error: tls: internal error
```

### DIFF
--- a/components/automate-deployment/pkg/server/server.go
+++ b/components/automate-deployment/pkg/server/server.go
@@ -760,6 +760,8 @@ func (s *server) ConfigureDeployment(ctx context.Context,
 }
 
 func (s *server) persistDeployment() error {
+	runConfigMigrations(s.deployment)
+
 	_, err := s.deploymentStore.UpdateDeployment(func(d *deployment.Deployment) error {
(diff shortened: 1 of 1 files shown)


danielsdeleo:
In https://github.com/chef/automate/pull/2719 I am adding diagnostics tests for applications data. Some of those scenarios are catching the problem this PR is solving, so I cherry-picked it there. This should help make the deep upgrades pass.

jaym:
Thanks

Issue #2733, opened 2020-01-28
NSS-MyHostname dep fails license scout


danielsdeleo:
## Describe the bug

The habitat ruby package depends on core/nss-myhostname as of https://github.com/habitat-sh/core-plans/commit/7761ccef724292640d14143ccb0cd5466e877fa5

This fails our license scout check with 

```
+---------+------------------------------------------+-------------------+-------------+
--
  | \| Type    \| Dependency                               \| License(s)        \| Results     \|
  | +---------+------------------------------------------+-------------------+-------------+
  | \| habitat \| core/nss-myhostname (0.3-20191025155905) \| LGPL-2.1-or-later \| Not Allowed \|
  | +---------+------------------------------------------+-------------------+-------------+

```

We can either make this dependency ok in policy or get rid of it somehow.

danielsdeleo:
Fixed in https://github.com/chef/automate/pull/2738

PR #2738, opened 2020-01-28
Request legal exception for LGPL nss-myhostname

danielsdeleo:

The core/ruby package depends on nss-myhostname at runtime. This library is LGPL 2.1+ and fails license scout checks. This patch adds an exception for the library. DO NOT MERGE unless/until approved by legal.

The library in question is described in it's documentation this way:

> nss-myhostname is a plugin for the GNU Name Service Switch (NSS)
> functionality of the GNU C Library (glibc) providing host name
> resolution for the locally configured system hostname as returned by
> gethostname(2).

Therefore the "work that uses the library" (which is subject to LGPL requirements) in this case is glibc, which is already packaged with Automate and excepted by legal.

* https://github.com/chef/automate/issues/2733
* https://github.com/chef/automate/pull/2734
* https://github.com/nomeata/libnss-myhostname

### DIFF
--- a/.license_scout.yml
+++ b/.license_scout.yml
@@ -257,6 +257,8 @@ exceptions:
       reason: (sd) Binary use only. Already shipped in chef-server. Exception made by Chef Legal
     - name: chef/automate-openjdk
       reason: Exception made by Chef Legal
+    - name: core/nss-myhostname
+      reason: (dan) Plugin for nss/glibc, doesn't link to our software.
 
   ruby:
     - name: highline
--- a/.license_scout.yml
+++ b/.license_scout.yml
@@ -258,7 +258,7 @@ exceptions:
     - name: chef/automate-openjdk
       reason: Exception made by Chef Legal
     - name: core/nss-myhostname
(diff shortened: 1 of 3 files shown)


jaym:
may or may not be useful information: to use it, `/etc/nsswitch.conf` needs to have a specific value. If it does, the file was likely already shipped with the system

.license_scout.yml
```diff
@@ -257,6 +257,8 @@ exceptions:
       reason: (sd) Binary use only. Already shipped in chef-server. Exception made by Chef Legal
     - name: chef/automate-openjdk
       reason: Exception made by Chef Legal
+    - name: core/nss-myhostname
+      reason: (dan) linked only in ruby interpreter.
```
btm: core/nss-myhostname isn't being linked in ruby, it's a library for nss/glibc.

danielsdeleo: Correct, I learned this between when I wrote this comment and when I wrote the PR description. Will update.

PR #2749, opened 2020-01-29
Library code for managing builder incremental backups

jaym:
This PR adds a library that is capable of managing the builder incremental backup repository. It is capable of taking snapshots, restoring snapshots, and deleting snapshots. This is not yet integrated into the deployment-service backup procedure

### The directory structure:
```
backups
|-- automate-elasticsearch-data
|-- shared
|   |-- builder
|   |   |-- artifacts
|   |   |   |-- core
|   |   |   |   |-- glibc
|   |   |   |   |   |-- ....
|   |   |   |   |   |   |-- core-glibc...hart
|   |   |-- snapshots
|   |   |   |-- 20200113201712.snapshot
|-- 20200113201712
|   |-- automate-builder
|   |   |-- metadata.json
|   |   |-- artifacts
|   |   |   |-- snapshot.meta
```

### Important Files:
- `shared` directory: This is the place where we're going to start putting
   things shared across backups. For example, `automate-elasticsearch-data`
   should go there, but that's difficult to do now

- `shared/builder/artifacts` directory: This actual harts will go in here.
  We need to keep the directory structure origin/pkg/version/release/platform/
  because just the hart name is ambiguous.

- `shared/builder/snapshots` directory: This file will contain files representing
  each snapshot taken. Each .snapshot file contains a list of all of the packages
  that were present in that snapshot. Also, this file should be compressed.

- `backupID/automate-builder/metadata.json`: This file needs to point at artifacts/snapshot.meta

- `backupID/automate-builder/artifacts/snapshot.meta`: This file contains the snapshot name in the repo along with the checksum
```json
{
  "name": "20200113201712",
  "checksum": "4c013726d095081e9bb86cd62601020dfaa3436318f135b1a830bb1f11e2a238"
}
```

### Assumptions:
- Lists are/remain sorted, so merge and diff on large lists is fast. 
- Deletes do not happen in parallel with other deletes and backups.
  - maybe we can lock objects to make this work?
- People aren't mucking with the repo manually

### Operations:
- Backup: It will first connect to
          the builder database and list all of the expected packages. Then, download and
          merge all the .snapshot files. This will be the list of files that exist in the
          snapshot repository. Diff the list of packages from the database and the list of
          packages from the snapshot repository unioned with the list of files from the database.
          This is the list of packages we need to upload. Upload them. Maybe we can have a upload
          manager that uploads a few in parallel at a time. Save the list of files from that 
          database as the .snapshot file.

- Restore: It downloads the snapshot file. This has the list of files we 
           need. Copy the files to where they need to be. Maybe we can have a download manager that
           can download these in parallel. Maybe we skip restoring if the file already exists.

- Delete: It downloads the snapshot file. Call this set B. It downloads all
          the other snapshot files and merges them. Call this set R. Find all the files that are in
          set B but missing in set R. This is the list of files that needs to be deleted. Move the
          .snapshot file that created B to something like .snapshot.deleting. It might be useful if
          things fail. Delete the files from the list of files we have to delete. Delete the 
          .snapshot file.

## Some numbers:
- A list of all the packages in hosted builder is 3MB. This is of the format
  `origin/package/version/release`, with one package per line. It compresses
  down to 500K gzip compressed. We can double this because we need to also write
  the file name as part of our line.


components/automate-deployment/pkg/backup/artifact_repo_helpers_test.go
```diff
@@ -0,0 +1,169 @@
...
+func sortStrings(in []string) []string {
+	sort.Strings(in)
+	return in
+}
+
+func isSorted(in []string) bool {
+	return sort.IsSorted(sort.StringSlice(in))
+}
+
+func isSortedAnUnique(in []string) bool {
```
yzl: `isSortedAndUnique`?

components/automate-deployment/pkg/backup/artifact_streams_propert_test.go
```diff
@@ -0,0 +1,125 @@
+package backup
+
```
yzl: Should the filename for this test be `artifact_streams_property_test.go`?

components/automate-deployment/pkg/backup/artifact_streams.go
```diff
@@ -0,0 +1,449 @@
...
+	// Peek returns the next element that will be returned
+	// when Next is called. It can be called multiple times
+	// and will keep generating the same value until Next
+	// is called
+	Peek() (string, error)
+}
+
+type peekableArtifactStream struct {
+	stream ArtifactStream
+	next   *string
```
stevendanna: Are we using `*string` here primarily to differentiate between a valid `""` value and a non-existent value or is there an additional reason?

components/automate-deployment/pkg/backup/artifact_streams.go
```diff
@@ -0,0 +1,449 @@
...
+	"io"
+	"strings"
+
+	"go.uber.org/multierr"
+)
+
+// ArtifactStream streams a list of artifacts keys
+// by name. A stream may not produce the same value
+// twice.
+// Returns io.EOF when no more elements will be produced
```
stevendanna: Might be good to document either here or on the implementations around what our expectations are around copying of these stream structs and thread-safety.  From reading the implementation, it appears that these aren't concurrency-safe (which is fine) so we might want to document that.

components/automate-deployment/pkg/backup/artifact_streams.go
```diff
@@ -0,0 +1,449 @@
...
+}
+
+func (d *mergeStream) Close() error {
+	err1 := d.a.Close()
+	err2 := d.b.Close()
+	return multierr.Combine(err1, err2)
+}
+
+// Merge returns a stream which is the union of all the given
+// streams
```
stevendanna: We may want to document that Merge returns unique elements?

components/automate-deployment/pkg/backup/artifact_streams.go
stevendanna: > A stream may not produce the same value twice.

I think another assumption is that a stream produces sorted values?

components/automate-deployment/pkg/backup/artifact_streams.go
jaym: i didn't want to think about if an empty string was valid or non-existent. But i think we could say empty string is the same as non-existent as other iterators filter them out

components/automate-deployment/pkg/backup/artifact_streams.go
jaym: yep, both true

PR #2758, opened 2020-01-30
team-add-users: add app-authorized guards

bcmdarroch:
Fixes a tiny issue I noticed while acceptance testing. As a project owner, I'm allowed to add users to teams in my project. However, I can't create users. On the team-add-users modal, once all users are added to a team, we show a button to create new users. But as a project owner, I get a 403 error when I try to create a user because I don't have permission to.

This PR adds an app-authorized guard to the `Create User` buttons on the team-add-users modal so a project owner (or any user without create user permissions) can't see them.

It also slightly changes the message the non-permissioned user will see.

### :+1: Definition of Done
- [x] any logged-in user who lacks permission to create users will not see a `Create User` button of the team-add-users modal

- start Automate + UI
- create a project
- add some user to the auto-generated `<project name> Project Owners` policy
- create a team inside that project
- log in as the project-owner user
- navigate to https://a2-dev.test/settings/teams/editors/add-users
- you should see the screen below:

Screen Shot 2020-01-29 at 5 26 54 PM

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/modules/team/team-add-users/team-add-users.component.html
+++ b/components/automate-ui/src/app/modules/team/team-add-users/team-add-users.component.html
@@ -12,12 +12,17 @@
   <div *ngIf="!(loading$ | async)">
     <app-create-user-modal [openEvent]="openUserModal"></app-create-user-modal>
     <div id="no-users-container" *ngIf="usersNotFiltered().length === 0">
-      <p>All local users have already been added; create some more!</p>
-      <chef-button primary (click)="openModal()">Create User</chef-button>
+      <app-authorized not [allOf]="['/iam/v2/users', 'post']">
+        <p>All local users have already been added.</p>
+      </app-authorized>
+      <app-authorized [allOf]="['/iam/v2/users', 'post']">
+        <p>All local users have already been added; create some more!</p>
+        <chef-button primary (click)="openModal()">Create User</chef-button>
+      </app-authorized>
     </div>
     <div id="table-container" *ngIf="usersNotFiltered()?.length > 0">
       <div id="table-controls">
-        <app-authorized [anyOf]="['/iam/v2/teams/{id}/users:add', 'post', team?.id]">
+        <app-authorized [allOf]="['/iam/v2/users', 'post']">
           <chef-button primary (click)="openModal()">Create User</chef-button>
         </app-authorized>
         <div id="users-selected">


components/automate-ui/src/app/modules/team/team-add-users/team-add-users.component.ts
```diff
@@ -169,9 +169,7 @@ export class TeamAddUsersComponent implements OnInit, OnDestroy {
   }
 
   getConfirmBtnText(): string {
-    if (this.usersNotFiltered().length === 0) {
-      return '';
-    } else if (this.addingUsers) {
+    if (this.addingUsers) {
```
bcmdarroch: this was a bonus fix. I noticed in the case of `if (this.usersNotFiltered().length === 0)`, the `Add Users` button just became a shortened blank button which looked off to me. So now when there are no users left to add you'll see a disabled `Add Users` button

susanev: oh was ur chef-ui-library up to date? those buttons should be totally hidden if there are no more users left to add.

components/automate-ui/src/app/modules/team/team-add-users/team-add-users.component.html
```diff
@@ -12,12 +12,17 @@
...
+      </app-authorized>
+      <app-authorized [allOf]="['/iam/v2/users', 'post']">
+        <p>All local users have already been added; create some more!</p>
+        <chef-button primary (click)="openModal()">Create User</chef-button>
+      </app-authorized>
     </div>
     <div id="table-container" *ngIf="usersNotFiltered()?.length > 0">
       <div id="table-controls">
-        <app-authorized [anyOf]="['/iam/v2/teams/{id}/users:add', 'post', team?.id]">
+        <app-authorized [allOf]="['/iam/v2/users', 'post']">
```
msorens: While I agree that having the users permission there is good, I would suggest also include the original teams users:add as well, because that is not being checked prior to getting here. (It should be, but that looks like a bit larger can of worms to fix that.)

components/automate-ui/src/app/modules/team/team-add-users/team-add-users.component.ts
msorens: @susanev is correct; this change should not be needed.

bcmdarroch: oh good catch thanks!

components/automate-ui/src/app/modules/team/team-add-users/team-add-users.component.html
bcmdarroch: I switched it because the app-authorized is only hiding the create button. 
Started looking into how we could start guarding the `Add Users` buttons on the team-details page and realized it would take some refactoring of the app-user-table. I'll make an issue for that fix.

bcmdarroch: issue: https://github.com/chef/automate/issues/2762

PR #2761, opened 2020-01-30
[go] update go to 1.13.7

ryancragun:


ryancragun:
We currently ship binaries built with 13.1, this bumps to 13.7, which has several `net/http` fixes, among many others, that seem like a good idea. Should we hold off another 2-4 weeks for 14.0 and rebuild the world then? Waiting will require a more involved change as there are updates to the module system that will likely require a bit of finessing.

PR #2671, opened 2020-01-21
Jamie/1841 search filter global projects

SEAjamieD:
### :nut_and_bolt: Description: What code changed, and why?
This branch adds a filtering feature to our global projects dropdown menu.  This allows users to sort through their open projects by name and check/uncheck which projects they would like to view.

### :chains: Related Resources
https://github.com/chef/automate/issues/1841

### :+1: Definition of Done
User can filter through their projects inside the dropdown, select, deselect, and apply selection to view their chosen projects limited by their filter choice.

Please note: in the interest of deadlines, the Create Project and Edit Project buttons have been moved to a later release.

### :athletic_shoe: How to Build and Test the Change
To Build
1. build components/automate-ui-devproxy && start_all_services
2. make serve

To Test
Its easier to visualize if you navigate to https://a2-dev.test/settings/projects
if you do not have several projects, please create at least 3
Click the projects dropdown button in the top right of the screen
in the dropdown, you should see an input box, begin to type the name of a project
you should see the list begin to filter closer to your project name as you type

select one project by clicking the checkbox next to its name
you should see a blue check mark, the project count raise by 1
the name of the project be the new title on the dropdown.

select two projects
you should see the project count = 2
the new title should read "multiple projects"

click the "apply" button
you should see page refresh
only your chosen projects should now be viewable

click the "Clear selection" button
you should see page refresh
all projects are now visible again

### :white_check_mark: Checklist

- [x] I have read the [CONTRIBUTING document](https://github.com/chef/automate/blob/master/CONTRIBUTING.md).
- [x] Tests added/updated?
- [ ] Docs added/updated?
- [x] All commits have been signed-off for the [Developer Certification of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

### :camera: Screenshots, if applicable

Screen Shot 2020-01-21 at 2 11 01 PM
Screen Shot 2020-01-17 at 1 56 47 PM
Screen Shot 2020-01-21 at 2 10 53 PM

PR #2770, opened 2020-01-31
Jamie/10-filters-ngrx-conversion

SEAjamieD:
The Global Projects Dropdown filter was built using a two way binding, ngModel pattern, however we would prefer a one way binding, ngrx pattern.  This branch modifies the Global project filter dropdown to use a one way binding ngrx pattern.

https://github.com/chef/automate/pull/2671

### :+1: Definition of Done
ngModel is gone from the global project dropdown and and ngrx pattern takes its place

To build:
2. make serve

To test:
1. if you do not have projects, go to https://a2-dev.test/settings/projects and create several projects
2. Click on the top right dropdown that will read something like "All Projects"
3.  In the dropdown, type some letters and watch your projects filter.

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
@@ -1,4 +1,4 @@
-import { Component, EventEmitter, Input, Output } from '@angular/core';
+import { Component, EventEmitter, Input, Output, OnInit } from '@angular/core';
 import { cloneDeep } from 'lodash/fp';
 import { ProjectsFilterOption } from 'app/services/projects-filter/projects-filter.reducer';
 import { ProjectsFilterService } from 'app/services/projects-filter/projects-filter.service';
@@ -11,7 +11,7 @@ const { ALL_PROJECTS_LABEL } = ProjectConstants;
   templateUrl: './projects-filter-dropdown.component.html',
   styleUrls: [ './projects-filter-dropdown.component.scss' ]
 })
-export class ProjectsFilterDropdownComponent {
+export class ProjectsFilterDropdownComponent implements OnInit {
 
   constructor(public projectsFilterService: ProjectsFilterService) { }
 
@@ -37,11 +37,16 @@ export class ProjectsFilterDropdownComponent {
   // so they can be filtered while maintaining the actual options.
   filteredOptions: ProjectsFilterOption[] = [];
 
-  filterValue = '';
-
   optionsEdited = false;
 
   dropdownActive = false;
+
+
+  ngOnInit() {
+    this.projectsFilterService.filterValue$.subscribe((filterValue) => {
+      this.filteredOptions = this.filterOptions(filterValue);
+    });
+  }
 
   resetOptions() {
     if (!this.optionsEdited) {
@@ -50,13 +55,16 @@ export class ProjectsFilterDropdownComponent {
   }
 
   handleFilterKeyUp(filterValue: string): void {
-    // this.filteredOptions = this.filterOptions(this.filterValue);
     this.projectsFilterService.updateFilterValue(filterValue);
   }
 
   filterOptions(value: string): ProjectsFilterOption[] {
     return this.editableOptions.filter(option =>
       option.label.toLowerCase().indexOf(value.toLowerCase()) > -1);
+  }
+
+  clearFilterValue(): void {
+    this.projectsFilterService.updateFilterValue('');
   }
 
   handleLabelClick() {
@@ -69,8 +77,8 @@ export class ProjectsFilterDropdownComponent {
   handleEscape(): void {
     this.optionsEdited = false;
     this.resetOptions();
-    this.filterValue = '';
     this.dropdownActive = false;
+    this.clearFilterValue();
     this.onOptionChange.emit(this.editableOptions);
   }
 
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
@@ -1,6 +1,7 @@
 import { Component, EventEmitter, Input, Output } from '@angular/core';
 import { cloneDeep } from 'lodash/fp';
 import { ProjectsFilterOption } from 'app/services/projects-filter/projects-filter.reducer';
+import { ProjectsFilterService } from 'app/services/projects-filter/projects-filter.service';
 import { ProjectConstants } from 'app/entities/projects/project.model';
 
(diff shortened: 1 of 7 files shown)


components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
```diff
@@ -33,27 +37,36 @@ export class ProjectsFilterDropdownComponent {
...
   optionsEdited = false;
 
   dropdownActive = false;
 
+
+  ngOnInit() {
+    this.projectsFilterService.filterValue$.subscribe((filterValue) => {
+      this.filteredOptions = this.filterOptions(filterValue);
+    });
+  }
```
SEAjamieD: ngOnInit seemed like a logical place to put this at the time, but now I'm thinking that since this is global - are subscriptions expensive?  I'm also not currently unsubscribing at any point.

I'm wondering if maybe a better route is to kick off this subscription when the dropdown opens, and then unsubscribe when the dropdown closes... thoughts?

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
```diff
@@ -33,27 +37,36 @@ export class ProjectsFilterDropdownComponent {
...
     if (!this.optionsEdited) {
       this.filteredOptions = this.editableOptions = cloneDeep(this.options);
     }
   }
 
-  handleFilterKeyUp(): void {
-    this.filteredOptions = this.filterOptions(this.filterValue);
+  handleFilterKeyUp(filterValue: string): void {
+    this.projectsFilterService.updateFilterValue(filterValue);
   }
```
SEAjamieD: this service is kicking off the update to filterValue that now lies in state

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
SEAjamieD: I should also mention, this subscription is listening for the filterValue in state, which is what a user types in the filter input box, and then updates the projects dropdown list according to what is being filtered.

msorens: You ask a good question ("are subscriptions expensive?") but that is really _two_ questions:

1. Is it expensive to create/destroy a subscription?
I would say "no" because best practice espoused by many articles is to create/destroy subscriptions as tightly as possible (create just-in-time and destroy as soon as you are done with it)

2. Is it expensive to maintain a subscription?
Not much in and of itself, as I imagine the rxjs design to monitor source Observables is rather efficient. But imagine if we never destroyed any subscriptions: We have a lot of "Observable overlap" in our code, since each component subscribes to what it needs, independent of what other components are doing. Particularly for those overlapping ones, every subscription will process the emitted Observable--even if it is not on a visible page and its result will never be used! (And sometimes leading to hard-to-find bugs.)

So while the global project filter is rather unique (it is created once and remains in existence for the life of the application) compared to almost every other component that is created and destroyed as things happen, we can still institute a tight create/destroy paradigm because of the lifecycle you described. And you have already led yourself to a good answer: open dropdown => subscribe, close dropdown => unsubscribe.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
```diff
@@ -33,27 +37,36 @@ export class ProjectsFilterDropdownComponent {
...
 
-  filterValue = '';
-
   optionsEdited = false;
 
   dropdownActive = false;
 
+
+  ngOnInit() {
+    this.projectsFilterService.filterValue$.subscribe((filterValue) => {
```
msorens:
```suggestion
    this.projectsFilterService.filterValue$.subscribe(filterValue => {
```

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
```diff
@@ -33,27 +37,36 @@ export class ProjectsFilterDropdownComponent {
...
   }
 
   filterOptions(value: string): ProjectsFilterOption[] {
     return this.editableOptions.filter(option =>
       option.label.toLowerCase().indexOf(value.toLowerCase()) > -1);
   }
 
+  clearFilterValue(): void {
+    this.projectsFilterService.updateFilterValue('');
+  }
```
msorens: Because this is only used once it seems like it would be better inlined... but that led me thinking about this.
1. Create this method:
```
  private setFilterValue(newValue: string): void {
    this.projectsFilterService.updateFilterValue(newValue);
  }
```
2. Use `setFilterValue` in both `handleFilterKeyUp` and `handleEscape`.

(Just a suggestion; you might prefer otherwise.)

SEAjamieD:
> Couple comments below.
> But as I thought about it, do we actually need the ngrx added here?
> If you just delete the ngModel entirely and do not reset the field to empty upon escape--the only place it was trying to send data back--doesn't that work with the original code? 🤔

I'm not sure why I didn't catch that in the first place... I just opened a new PR here (https://github.com/chef/automate/pull/2779)  and will close this one based on whether or not you agree.  Seems way simpler to not have ngrx or ngModel....

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
SEAjamieD: Thanks for taking the time to answer this so thoroughly.  While I suspect we'll end up using my other PR - this was my first time using a subscription and it was a good introduction.  It makes sense that we would want to unsubscribe as soon as we don't need that subscription anymore and I'll keep that tight coupling in mind moving forward.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
bcmdarroch: interesting question! 💭 I think this is a case where we might not need a subscription.
[This article](https://medium.com/angular-in-depth/when-to-subscribe-a83332ae053) seems to discourage subscriptions unless absolutely necessary. It cites this [other article](https://medium.com/angular-in-depth/angular-question-rxjs-subscribe-vs-async-pipe-in-component-templates-c956c8c0c794) which argues that `async` in the template is the better option for using observable data, since it works with change detection out of the box and you don't have to remember to unsubscribe.

msorens:
Agreed; go ahead and close this one. ✔️

SEAjamieD:
Closed in favor of https://github.com/chef/automate/pull/2779

Issue #2717, opened 2020-01-27
auth pages and button toolbar issues


susanev:
## Overview
All buttons in the toolbar should have their left, top, and bottom margins removed. It seems some of the auth pages have lost this.

## Design Details
Update the toolbars on various pages, so that the buttons are correctly placed. The buttons in the toolbar should have no top, left, or bottom margin. And any toolbar with multiple buttons should have 12px of space between those buttons.
- [x] projects list table
- [x] ingest rules table
- [x] teams list table
- [x] users on team table
- [x] users table
- [x] policy members

## Screenshot 
Tokens list page has the correct working example.
<img width="698" alt="correct-button-placement" src="https://user-images.githubusercontent.com/5489125/73187329-1b18a380-40d6-11ea-9bcb-ad23bb9883e7.png">

PR #2772, opened 2020-01-31
adjusting toolbar button margins

susanev:

When we switched the angular table component we started using border-spacing which added 8px to the top of our tables, which also pushed the buttons 8px farther away from the tables. This is fixing it by setting the bottom margin of the toolbar buttons to 0 as well. We could also just stop using border-spacing.. but this seemed like a smaller change.

#2717

### :+1: Definition of Done
tables and toolbar buttons are 8px away from each other

`make update-ui-lib` from `components/automate-ui`

- [ ] Tests added/updated?
- [ ] Docs added/updated?

#### before
before

#### after
after

### DIFF
--- a/components/chef-ui-library/src/molecules/chef-toolbar/chef-toolbar.scss
+++ b/components/chef-ui-library/src/molecules/chef-toolbar/chef-toolbar.scss
@@ -1,10 +1,5 @@
 chef-toolbar {
   chef-button {
-    margin-top: 0;
-
-    &:first-child {
-      margin-top: 0;
-      margin-left: 0;
-    }
+    margin: 0 8px 0 0;
   }
 }


components/chef-ui-library/src/molecules/chef-toolbar/chef-toolbar.scss
```diff
@@ -1,10 +1,5 @@
 chef-toolbar {
   chef-button {
-    margin-top: 0;
-
-    &:first-child {
```
susanev: the way things are setup right now all buttons in the toolbar are first childs so this was being applied to all the buttons anyway

susanev:
and no this wont work, because we havent, switched all tables yet... sigh. i think we prolly need to stop using border-spacing.

PR #2773, opened 2020-01-31
changing angular table component to use flex

susanev:

Recently, we started using a new table component on the auth tables. The styles on this new table used border-spacing which is pretty inflexible and caused an additional 8px to be added to the top of the table. To address this, I changed the component to use flex instead.

I also considered removing the bottom margin from the buttons in the toolbar but since this new table component has not replaced all tables yet that would cause the pages with the chef ui library table component to have zero space between the toolbar buttons and the table.

#2717

### :+1: Definition of Done
tables and toolbar buttons are 8px away from each other

- [ ] Tests added/updated?
- [ ] Docs added/updated?

#### before
before

#### after
after

### DIFF
--- a/components/automate-ui/src/app/components/table/table-cell/table-cell.component.scss
+++ b/components/automate-ui/src/app/components/table/table-cell/table-cell.component.scss
@@ -1,7 +1,6 @@
 @import '../table.component';
 
 :host {
-  display: table-cell;
   padding: $size_16;
   vertical-align: middle;
   text-align: left;
--- a/components/automate-ui/src/app/modules/token/token-list/api-token-list.component.scss
+++ b/components/automate-ui/src/app/modules/token/token-list/api-token-list.component.scss
@@ -1,9 +1,5 @@
 @import "~styles/variables";
 
-#create-button {
-  margin: 0;
-}
-
 .empty-state {
   text-align: center;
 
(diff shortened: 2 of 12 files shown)


components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.scss
```diff
@@ -70,7 +70,7 @@ chef-loading-spinner {
     chef-table-cell:last-child {
       chef-button {
         margin: -0.5em 0;
-        float: right;
+        margin-left: auto;
```
susanev: this was needed for the get buttons because floats dont work in flex.

components/automate-ui/src/app/modules/token/token-list/api-token-list.component.scss
```diff
@@ -1,9 +1,5 @@
 @import "~styles/variables";
 
-#create-button {
```
susanev: this is not needed anymore

components/automate-ui/src/app/components/table/table-cell/table-cell.component.scss
```diff
@@ -1,7 +1,6 @@
 @import '../table.component';
 
 :host {
-  display: table-cell;
```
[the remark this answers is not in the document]
susanev: should i leave the empty scss file there for table body and table header?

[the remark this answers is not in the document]
msorens: No, better to remove those (and the reference to them in the .ts files).

components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.scss
SEAjamieD: [this is a total nitpick]: while I don't think there's anything wrong with this, since everything else is going flex, It might be a bit more "flex-like" to use `align-self: flex-end;` here.

PR #2774, opened 2020-01-31
delete old and empty indices

rickmarry:

This change fixes the panic that occurs when empty orphaned compliance elasticsearch indices are left in ES.  The problem with empty indices for a given index version is that when compliance comes up and tries to migrate them, they have not start_time (because they have no docs), and this results in a nil pointer on the min_date value.  Normally, indices are deleted once they are migrated but if for whatever reason, the deletion does not occur, these orphaned indices are problematic.  The solution is to remove these empty indices when this condition is detected.

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/compliance-service/reporting/relaxting/migrations.go
```diff
@@ -41,6 +42,31 @@ func StoreExists(client *elastic.Client, indexName string) (bool, error) {
...
+	}
+
+	indices := make([]string, len(catIndicesResponse))
+
+	for _, catIndicesResponseRow := range catIndicesResponse {
+		indexName := catIndicesResponseRow.Index
+		logrus.Debugf("\nIndex: %s ", indexName)
+
+		dateAsStr := strings.SplitAfterN(indexName, "-", 4)
+		dateToRemove, err := time.Parse("2006.01.02", dateAsStr[len(dateAsStr)-1])
```
lancewf: Is there a chance that the `dateAsStr` will have zero length?

components/compliance-service/reporting/relaxting/migrations.go
```diff
@@ -354,6 +383,17 @@ func (backend ES2Backend) getScanDateRange(indexPrefix string) (*time.Time, *tim
 		return nil, nil, errors.Wrap(err, fmt.Sprintf("%s unable to read max_date for range", myName))
 	}
 
+	if minDate.Value == nil {
```
lancewf: Would `minDate.Value` be nil if there were no indices with this prefix?

components/compliance-service/reporting/relaxting/migrations.go
```diff
@@ -354,6 +383,17 @@ func (backend ES2Backend) getScanDateRange(indexPrefix string) (*time.Time, *tim
...
+		//if we are here, it means that there are indices with this prefix but not any elasticsearch documents within those indices
+		//if there were, then mindDate.value would be non-nil
+		//in this case, we need to remove all of the indices with this prefix.. they are all empty anyway
+		_, err := cleanupEmptyIndicesForPrefix(client, esIndex, migratable)
+		if err != nil {
+			return nil, nil, errors.Wrap(err, fmt.Sprintf("%s unable to delete the empty indices that need to be removed", myName))
+		}
+		return nil, nil, errors.Errorf("%s there were not any reports in elasticsearch indices that are prefixed with %s, these indices, therefore, have been successfully deleted", myName, esIndex)
```
lancewf: Should we return an error if there were no indices to migrate? We could return `nil, nil, nil` and the migration would show a success.

rickmarry: yes. thanks, @lancewf.  We no longer return an error here, instead, we log that as a warning and move on

components/compliance-service/reporting/relaxting/migrations.go
vjeffrey: ^^ that's a good point. probably worth a guard on that

rickmarry: good catch.  If that dateAsStr is zero len, we now log a warning and move on.

PR #2779, opened 2020-02-03
Remove ngmodel without ngrx

SEAjamieD:
The Global Projects Dropdown filter was built using a two way binding, ngModel pattern, however we would prefer to not use ngModel.  This branch removes the use of ngModel.

https://github.com/chef/automate/pull/2671

### :+1: Definition of Done
ngModel is gone from the global project dropdown

To build:
2. make serve

To test:
1. if you do not have projects, go to https://a2-dev.test/settings/projects and create several projects
2. Click on the top right dropdown that will read something like "All Projects"
3.  In the dropdown, type some letters and watch your projects filter.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
@@ -21,8 +21,7 @@
           <input 
             chefInput
             type="text"
-            [(ngModel)]="filterValue"
-            (keyup)="handleFilterKeyUp()"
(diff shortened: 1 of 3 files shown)


msorens:
For a richer experience, it is easy to generate lots of projects from the command line:


Generate 500 starting with id 101
```
    authGen projects create 500 101
```
Remove them after testing
```
    authGen projects delete 500 101
```
That will work inside **or** outside hab studio with a simple prerequisite--point at your A2 instance and authorize it, then source the script:
```
    export TOK=chef-automate iam token create TEST --admin
    export TARGET_HOST=a2-dev.test
    . scripts/auth_collections.sh
```

PR #2781, opened 2020-02-04
remove ngmodel

SEAjamieD:
Removing ngModel from the filter dropdown component

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.html
+++ b/components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.html
@@ -14,7 +14,7 @@
   </button>
   <chef-click-outside (clickOutside)="closeDropdown()">
     <chef-dropdown [attr.visible]="active">
-
+      <p>{{ filterValue$ | async }}</p>    <!-- this is just for sanity checking -->
       <div id="filter-container">
         <input 
           chefInput
--- a/components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.html
+++ b/components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.html
@@ -20,7 +20,7 @@
           chefInput
           type="text"
           placeholder="Filter projects..."
-          [ngModel]="filterValue"
+          [ngModel]="filterValue$ | async"
           (keyup)="handleFilterKeyUp($event.target.value)"
           />
       </div>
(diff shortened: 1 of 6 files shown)


SEAjamieD:
@msorens - Since you're familiar with what i'm looking accomplish here.  Unfortunately it looks to me like the more reusable version of the projects dropdown filter can't simply remove `ngModel` in the same manner as the global version, solely because when the dropdown is closed, the input isn't cleared and whatever was typed into that input remains indefinitely until the user removes it.

Before I spend the time to do it, do you think this would be a candidate for ngRx?

I could make that input a `#ref` or something, but that still seems like 2 way binding to me.

msorens:
Thanks for asking the question, @SEAjamieD !
The TL;DR is "yes!" this is a good candidate for using ngrx/store. (So what you did in the last PR and discarded was not wasted effort. 😁 )

Before you dive in, though, let me suggest you review a couple things.
My first thought was, "well, I should just be able to point you to an example in the code base"... But from a quick search, we actually do what I will call a "full-circle input field" very seldomly in our code base, and those that do, seem to use `[(ngModel)]` 🤷‍♂ which, as we have discussed, violates the clean one-way binding benefit of ngrx/store.

So, first let me point you to my slide deck (this is a short one 😁 ): https://docs.google.com/presentation/d/13ut5p2dWSEKYy4Xt3sREZenfQiPtZLc_9uA_70GlRxg/edit#slide=id.g20011f5b3f_0_0
that I put together to introduce ngrx to our engineering community some time ago.
That illustrates one-way binding but, alas, still not the full-circle input field.
So for that, I found this tutorial that looks quite coherent: https://malcoded.com/posts/angular-ngrx-guide/#displaying-state-in-components-with-the-async-pipe

The whole article is well-written, but the link I have given jumps right to the key example of how to do a full-circle input. To give away the ending, it is:
```
<input [ngModel]="amount$ | async" (ngModelChange)="onAmountChange($event)" />
```

susanev:
where are we on this work? starting to get a bit worried about the age of the pr...

SEAjamieD:
Oh I’ll close the PR for this.  It’s just a chore and needed to be rethought later.  Other work is more pressing at the moment, but I’d like to keep the ticket open if that’s OK

susanev:
sure yes of course, totally cool to keep the issue open.

PR #2796, opened 2020-02-05
[Automate-1890] gateway v1 scrub

msorens:

The intent of this PR is to remove all v1/v2 distinctions in gateway code; I would say it achieves about 90% of that level; see my notes below where I felt like I was getting diminishing returns.

Now if you do a search for V1 or V2 in automate-gateway, there are very limited occurrences. 😁 

NA

### :+1: Definition of Done
Builds, tests, and running code all work.
(Waiting on v2 migrations to be in place before buildkite will be happy.)

```
rebuild components/automate-gateway/
```

- [x] Tests added/updated?
- [ ] Docs added/updated?


components/automate-gateway/gateway/middleware/auth_interceptor.go
```diff
@@ -41,18 +41,16 @@ type SwitchingAuthorizationHandler interface {
 	GRPCAuthorizationHandler
 	SwitchingFilterHandler
 	IsAuthorized(ctx context.Context, subjects []string,
-		resourceV1, actionV1, resourceV2, actionV2 string, projects []string) (AnnotatedAuthorizationResponse, error)
+		resource, action string, projects []string) (AnnotatedAuthorizationResponse, error)
 }
 
 type SwitchingFilterHandler interface {
```
msorens: Though I stripped out the v1/v2 distinction, it seems that there are further simplifications to be done: removing this SwitchingFilterHandler and the SwitchingAuthorizationHandler, for instance, since they are no longer "switching".

But I feel like I have stared at them long enough for now; if you view this file (auth_interceptor.go) in its entirety, you will begin to the see the... entanglements that would need to be unravelled.

components/automate-gateway/handler/authz.go
```diff
@@ -18,19 +18,18 @@ import (
...
-	"github.com/chef/automate/components/automate-gateway/authz/policy_v2"
+	policy "github.com/chef/automate/components/automate-gateway/authz/policy_v2"
 	"github.com/chef/automate/components/automate-gateway/gateway/middleware"
 	"github.com/chef/automate/lib/grpc/auth_context"
 )
 
 // AuthzServer is the server interface
 type AuthzServer struct {
-	client        authz.AuthorizationClient
+	client        authz.AuthorizationClient // now only needed for GetVersion
```
msorens: Another bit of future work. We still need a v1 client here for GetVersion and for introspection calls; but the v1 client also handles policies here, too(!), hence my comments below in this file.

tylercloke: Can we make a card for this?

components/automate-gateway/handler/authz_test.go
```diff
@@ -36,8 +31,8 @@ import (
 	_ "github.com/chef/automate/components/automate-gateway/api/compliance/profiles"
 )
 
-func TestIntrospectAllV1(t *testing.T) {
-	authzSrv, _, s, hdlr := testServerAndHandler(t)
+func TestIntrospectAllV2(t *testing.T) {
```
bcmdarroch: Is it necessary to keep `V2` in the name?

components/automate-gateway/handler/authz_test.go
```diff
@@ -226,7 +154,7 @@ func TestIntrospectSome(t *testing.T) {
 	for name, tc := range cases {
 		t.Run(name, func(t *testing.T) {
 			ctx := auth_context.NewContext(
-				context.Background(), []string{"user:local:admin"}, []string{"project"}, "some:resource", "some:action", "AuthV1")
+				context.Background(), []string{"user:local:admin"}, []string{"project"}, "some:resource", "some:action", "Auth")
```
bcmdarroch: What is `"Auth"`?

components/automate-gateway/pkg/authorizer/authorizer.go
```diff
@@ -4,113 +4,48 @@ import (
...
-		return newCtx, nil
-	case codes.FailedPrecondition:
-		if a.fromStatus(st) {
-			return a.Handle(ctx, subjects, projects, req)
-		}
-		fallthrough
-	default: // any other error status
-		return ctx, err
-	}
+	return a.authHandler.Handle(ctx, subjects, projects, req)
```
bcmdarroch: This is probably best left for another card but do we want to keep this Handler abstraction?

bcmdarroch:
At what point will we be able to remove the [v1 introspection code](https://github.com/chef/automate/tree/master/components/automate-gateway/api/authz/pairs) in the gateway? Will that require porting introspection over to the v2 API first?

components/automate-gateway/handler/authz_test.go
msorens: No; good catch.

components/automate-gateway/handler/authz_test.go
msorens: Good question...found an answer, patched it here, and added further cleanup to https://github.com/chef/automate/issues/2807

components/automate-gateway/handler/authz.go
msorens: Done: https://github.com/chef/automate/issues/2808

components/automate-gateway/pkg/authorizer/authorizer.go
msorens: No, should probably remove it... but that is part of the work referred to in my comment above about `SwitchingFilterHandler`.

bcmdarroch: I'm confused - do you plan to do that in this PR or a follow-up?

PR #2806, opened 2020-02-06
Improvements to acceptance testing guide

phiggins:
Some improvements to the acceptance testing guide I made while going through it for the first time.


components/compliance-service/smokin/README.md
```diff
@@ -21,8 +20,14 @@ COLLECTOR_URL='https://A2-HOST/data-collector/v0/' COLLECTOR_TOKEN='TOKEN' kitch
...
-```
-or from the Automate UI: `Admin` > `API Tokens` > `Add API Token` > `Generate API Token` > Copy the value
+## Generating a token
+
+* The value for COLLECTOR_TOKEN can be obtained from Automate's hab studio via command:
+  ```bash
+  get_admin_token
+  ```
+
+* Or from the Automate UI:
```
susanev: i think yall must still be using iam v1 for acceptance testing, soon on iam v2 after creating from the ui it will need to be added to the policy as a member to make it an admin token (this can also be done from the ui)

PR #2820, opened 2020-02-07
Add app-authorized to applications nav link

SEAjamieD:
We need to remove the Applications link from navigation when the user does not have permissions to see the applications tab.  This branch adds an app authorized container around Applications using introspection to check if the user has permission or not.

### :+1: Definition of Done
When a user does not have permission to use the Applications tab, it is not accessible

make serve

In order to test this properly, you'll need to:
1. log in as admin
2. create a new user
3. create a policy that does not allow a user to view applications
4. assign your new user that policy
5. log out of admin
6. log in as the user you created.

If you're testing this, please let me know if you see the applications tab when you expect not to. - thank you!

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/page-components/navbar/navbar.component.html
+++ b/components/automate-ui/src/app/page-components/navbar/navbar.component.html
@@ -7,7 +7,7 @@
       <div class="navigation-menu">
         <div role="menuitem" class="nav-link"><a routerLink="/dashboards/event-feed" routerLinkActive="active" tabindex="0">Dashboards</a></div>
 
-        <app-authorized [anyOf]="[['/applications/services', 'get'], ['/applications/service-groups', 'get']]">
+        <app-authorized [allOf]="[['/applications/services', 'get'], ['/applications/service-groups', 'get']]">
           <div role="menuitem" class="nav-link"><a routerLink="/applications/service-groups" routerLinkActive="active" tabindex="0">Applications</a></div>
         </app-authorized>
 
--- a/components/automate-ui/src/app/page-components/navbar/navbar.component.html
+++ b/components/automate-ui/src/app/page-components/navbar/navbar.component.html
@@ -7,7 +7,9 @@
       <div class="navigation-menu">
         <div role="menuitem" class="nav-link"><a routerLink="/dashboards/event-feed" routerLinkActive="active" tabindex="0">Dashboards</a></div>
 
-        <div role="menuitem" class="nav-link"><a routerLink="/applications/service-groups" routerLinkActive="active" tabindex="0">Applications</a></div>
+        <app-authorized [anyOf]="['/auth/introspect', 'get']">
+          <div role="menuitem" class="nav-link"><a routerLink="/applications/service-groups" routerLinkActive="active" tabindex="0">Applications</a></div>
+        </app-authorized>
 
         <app-authorized [allOf]="[['/cfgmgmt/nodes', 'get'], ['/cfgmgmt/stats/node_counts', 'get']]">
           <div role="menuitem" class="nav-link"><a routerLink="/infrastructure" routerLinkActive="active" tabindex="0">Infrastructure</a></div>
--- a/components/automate-ui/src/app/page-components/navbar/navbar.component.html
+++ b/components/automate-ui/src/app/page-components/navbar/navbar.component.html
@@ -7,7 +7,7 @@
       <div class="navigation-menu">
         <div role="menuitem" class="nav-link"><a routerLink="/dashboards/event-feed" routerLinkActive="active" tabindex="0">Dashboards</a></div>
 
-        <app-authorized [anyOf]="['/auth/introspect', 'get']">
+        <app-authorized [anyOf]="['/applications/service-groups', 'get']">
           <div role="menuitem" class="nav-link"><a routerLink="/applications/service-groups" routerLinkActive="active" tabindex="0">Applications</a></div>
         </app-authorized>
 
--- a/components/automate-ui/src/app/page-components/navbar/navbar.component.html
+++ b/components/automate-ui/src/app/page-components/navbar/navbar.component.html
@@ -7,7 +7,7 @@
       <div class="navigation-menu">
         <div role="menuitem" class="nav-link"><a routerLink="/dashboards/event-feed" routerLinkActive="active" tabindex="0">Dashboards</a></div>
 
-        <app-authorized [anyOf]="['/applications/service-groups', 'get']">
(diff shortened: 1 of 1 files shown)


components/automate-ui/src/app/page-components/navbar/navbar.component.html
```diff
@@ -7,7 +7,9 @@
       <div class="navigation-menu">
         <div role="menuitem" class="nav-link"><a routerLink="/dashboards/event-feed" routerLinkActive="active" tabindex="0">Dashboards</a></div>
 
-        <div role="menuitem" class="nav-link"><a routerLink="/applications/service-groups" routerLinkActive="active" tabindex="0">Applications</a></div>
+        <app-authorized [anyOf]="['/auth/introspect', 'get']">
```
SEAjamieD: @bcmdarroch @tyler-ball @msorens - I think I have this setup correctly, but Its doesn't seem to be working as I expect it to.  

If i'm reading the readme correctly, a `GET` call to the `/auth/introspect` endpoint should check all of the non parameterized endpoints
>1. GET - Does an "IntrospectAll", returning data for all non-parameterized endpoints
   (i.e. concrete endpoints like "/auth/teams" but not abstract endpoints like "/auth/users/{email}").

So I'm expecting this would check all of the users authorizations/permisssions, but I'm still seeing the Applications link shown when a user does not have access to Applications.  Does this look off to you?

susanev:
```suggestion
        <app-authorized [anyOf]="['/applications', 'get']">
```

pretty sure thats what you want there..

susanev: oh also tyler ball is a different engineer at chef, hes on the workstation team.. tyler cloke is the tyler on the auth team

SEAjamieD: haha whoops!  def meant tyler cloke!  Thanks susan!

components/automate-ui/src/app/page-components/navbar/navbar.component.html
```diff
@@ -7,7 +7,9 @@
       <div class="navigation-menu">
         <div role="menuitem" class="nav-link"><a routerLink="/dashboards/event-feed" routerLinkActive="active" tabindex="0">Dashboards</a></div>
 
-        <div role="menuitem" class="nav-link"><a routerLink="/applications/service-groups" routerLinkActive="active" tabindex="0">Applications</a></div>
+        <app-authorized [anyOf]="['/applications/service-groups', 'get']">
```
msorens: I know this is still a draft, but you will probably want to check with the applications team to see what paths should be included in the `app-authorized`. It is not necessarily a one-to-one with the URL below. Looking at the applications tab, it could easily be that they need to have all permissions (`allOf`) for GetServices, GetServiceGroups, , and GetServiceGroupsHealthCounts, for example. Or it might be an `anyOf` for those. But it is the applications team that needs to inform you on that.

components/automate-ui/src/app/page-components/navbar/navbar.component.html
msorens: @susanev is correct; you will want this one... at least. See my notes below.

components/automate-ui/src/app/page-components/navbar/navbar.component.html
SEAjamieD: Thanks @msorens - I'll reach out to the applications team tomorrow to clarify

components/automate-ui/src/app/page-components/navbar/navbar.component.html
susanev: also, the reason this appears to work is because your policy has `applications:*` in it. i imagine though that that page would have errors if the user (by way of the policy) only has access to services and not the service groups.

components/automate-ui/src/app/page-components/navbar/navbar.component.html
SEAjamieD: @vjeffrey - would you mind having a look at this and seeing if it covers what you intended?  otherwise I may need to add additional routes.

SEAjamieD:
> worked for me locally!
> 
> i tested in my local dev env by doing the following:
> 
> * sign into local dev env as admin
> * go to settings/users, create a user
> * go to settings/policies, select ingest policy, member tab, add the user that was just created
> * sign in as said user. before this change i saw the applications link in the header, and after did not! 🎉
> * when accessing as viewer/admin, the applications link was there as expected
> 
> thank you!

Awesome!  Thank you for checking!

components/automate-ui/src/app/page-components/navbar/navbar.component.html
```diff
@@ -7,7 +7,9 @@
       <div class="navigation-menu">
         <div role="menuitem" class="nav-link"><a routerLink="/dashboards/event-feed" routerLinkActive="active" tabindex="0">Dashboards</a></div>
 
-        <div role="menuitem" class="nav-link"><a routerLink="/applications/service-groups" routerLinkActive="active" tabindex="0">Applications</a></div>
+        <app-authorized [anyOf]="[['/applications/services', 'get'], ['/applications/service-groups', 'get']]">
```
susanev: are we doing any of was that the decision? i thought dan has mentioned needing both?

SEAjamieD: I know the comment you're talking about and I am mind-boggled right now as to where it went.  @danielsdeleo / @vjeffrey can one of you confirm if we need _both_ of the endpoints or _at least one_ ?

danielsdeleo: You would want to have permission on both or else the best you could get is 1/2 of the page. In the back-end, these are both gated by the same permission (`applications:serviceGroups:list` for IAMv2) so it's not possible to have permissions to only one. So it makes sense to me to do one of these:
* only check `['/applications/services', 'get']`, knowing that anyone authorized for that will have permissions on the other one as well - OR -
* require permissions on both endpoints, which would defend you against the case that we changed the permissions of things but forgot to make the corresponding change in this part of the UI. "Famous last words" and all but I don't expect to make a change like that any time soon.

SEAjamieD: Thank you! 

@susanev - after further info from @danielsdeleo, I've changed it to `allOf`, but left out `service groups id` because in this context, it would make any difference in terms of permissions.

PR #2824, opened 2020-02-07
Jamie/2573 accordions for long service messages

SEAjamieD:
We are attempting to make the outputs from health checks more scannable, so text should be sitting in an accordion if longer than around 140 characters.  This branch adds show more buttons to long health messages that can expand and collapse

https://github.com/chef/automate/pull/2572
https://github.com/chef/automate/issues/2554

### :+1: Definition of Done
show more buttons are in place and toggling properly

source .studio/applications-service

make serve

navigate to https://a2-dev.test/applications/service-groups
toggle open various health messages on the right side of the screen

Verify Behavior:
Filter for service name: pos-terminal
There should be no "show more" buttons shown in any of the production environments

Filter for service name: nginx
the bldr-cache app in qa env shows messages of the maximum size. They wrap even without word breaks

- [ ] Tests added/updated?
- [ ] Docs added/updated?

Accordions

### DIFF
--- a/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
+++ b/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
@@ -96,11 +96,13 @@
                 </div>
                 <div class="healthcheck-content" [class.active]="activeErrorAccordions.includes(i)">
                   {{service.health_check_result.stderr}}
+                  <div class="show-more-container">
+                    <button 
+                    *ngIf="service.health_check_result.stderr.length > 138"
+                    class="show-more-button"
+                    (click)="toggleMoreErrorMsg(i)">Show {{ activeErrorAccordions.includes(i) ? "Less" : "More"}}</button>
+                  </div>
                 </div>
-                <button 
-                *ngIf="service.health_check_result.stderr.length > 138"
-                class="show-more-button"
-                (click)="toggleMoreErrorMsg(i)">Show {{ activeErrorAccordions.includes(i) ? "Less" : "More"}}</button>
               </div>
               <div class="healthcheck-message" *ngIf="service.health_check_result.stdout !== ''">
                 <div class="healthcheck-header {{service.health_check}}">
@@ -108,11 +110,13 @@
                 </div>
                 <div class="healthcheck-content" [class.active]="activeHealthAccordions.includes(i)">
                   {{service.health_check_result.stdout}}
+                  <div class="show-more-container">
+                    <button 
+                    *ngIf="service.health_check_result.stdout.length > 138"
+                    class="show-more-button"
+                    (click)="toggleMoreHealthMsg(i)">Show {{ activeHealthAccordions.includes(i) ? "Less" : "More"}}</button>
+                  </div>
                 </div>
-                <button 
-                *ngIf="service.health_check_result.stdout.length > 138"
-                class="show-more-button"
-                (click)="toggleMoreHealthMsg(i)">Show {{ activeHealthAccordions.includes(i) ? "Less" : "More"}}</button>
               </div>
             </div>
           </div>
--- a/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
+++ b/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
@@ -99,8 +99,10 @@
                   <div class="show-more-container">
                     <button 
                     *ngIf="service.health_check_result.stderr.length > 138"
-                    class="show-more-button"
-                    (click)="toggleMoreErrorMsg(i)">Show {{ activeErrorAccordions.includes(i) ? "Less" : "More"}}</button>
+                    (click)="toggleMoreErrorMsg(i)"
+                    class="show-more-button">
+                    <chef-icon class="unfold">{{ activeErrorAccordions.includes(i) ? "unfold_less" : "unfold_more"}}</chef-icon>
+                    Show {{ activeErrorAccordions.includes(i) ? "Less" : "More"}}</button>
                   </div>
                 </div>
               </div>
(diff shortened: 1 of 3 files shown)


components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
```diff
@@ -94,18 +94,30 @@
                 <div class="healthcheck-header {{service.health_check}}">
                   Health Check Error Message
                 </div>
-                <div class="healthcheck-content">
+                <div class="healthcheck-content" [class.active]="activeErrorAccordions.includes(i)">
                   {{service.health_check_result.stderr}}
+                  <div class="show-more-container">
+                    <button 
+                    *ngIf="service.health_check_result.stderr.length > 138"
```
SEAjamieD: UX requested ~140 characters in the short form of the output.  By my calculation - 138 characters is set to 3 lines of output.

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.ts
```diff
@@ -117,4 +119,20 @@ export class ServicesSidebarComponent implements OnInit, OnDestroy {
...
+  }
+
+  public toggleMoreHealthMsg(index: number) {
+    if (this.activeHealthAccordions.includes(index)) {
+      this.activeHealthAccordions = this.activeHealthAccordions.filter(n => n !== index);
+    } else {
+      this.activeHealthAccordions.push(index);
+    }
+  }
+
```
SEAjamieD: I wanted a way to expand/collapse accordions and to be able to have multiple accordions open/collapsed at different times.  I felt like the best way to accomplish this was to store the index of active accordions in a public array, and then have the template check for whether or not the array contained that particular index and attach an `active` class accordingly.  This way, the CSS handles the actual height of the active accordion.

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
```diff
@@ -94,18 +94,30 @@
                 <div class="healthcheck-header {{service.health_check}}">
                   Health Check Error Message
                 </div>
-                <div class="healthcheck-content">
+                <div class="healthcheck-content" [class.active]="activeErrorAccordions.includes(i)">
```
SEAjamieD: We attach an active class here if this particular accordion is included in the active accordions array.

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.scss
```diff
@@ -210,11 +210,42 @@ chef-alert {
...
 
   .healthcheck-content {
+    position: relative;
     border-radius: 0 0 2px 2px;
     background-color: var(--chef-primary-dark);
     color: var(--chef-white);
-    padding: 8px;
+    padding: 8px 8px 40px 8px;
     overflow-wrap: break-word;
+    max-height: 94px;
```
SEAjamieD: By default, we're setting the max height to 94px... this is 3 lines of output plus the expand button.

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.scss
```diff
@@ -210,11 +210,42 @@ chef-alert {
...
+    overflow: hidden;
+    transition: .25s max-height;
+
+    &.active {
+      max-height: 540px; // temp placeholder
```
SEAjamieD: we only expand as far as the text needs to go by using max-height instead of height, tapping out at 540px max because we will only ever show 1024 characters (need to confirm with applications)

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.scss
```diff
@@ -210,11 +210,42 @@ chef-alert {
...
+    position: relative;
     border-radius: 0 0 2px 2px;
     background-color: var(--chef-primary-dark);
     color: var(--chef-white);
-    padding: 8px;
+    padding: 8px 8px 40px 8px;
     overflow-wrap: break-word;
+    max-height: 94px;
+    overflow: hidden;
+    transition: .25s max-height;
```
SEAjamieD: I think i want to add a specific bezier curve here for looks - the reason we're transitioning to specific max-heights is so that we can animate it.  This css animation is cheap and lets us slide the text open and closed for a less jarring effect.

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.ts
SEAjamieD: I should add that although these functions accomplish the exact same task - I felt it was easier to follow/maintain by keeping them separate and not having to pass in Arrays or follow any kind of extended logic.  Open to suggestions though.

SEAjamieD:
> Unfortunately, I was not able to show any service groups or messages after `applications_populate_database`. Just got a message in the UI "Health check messages not available." Let me know if you have any ideas.

@msorens - Thanks for having a look.  I think I missed a step, so I added it in to the instructions above.  You may need to run `source .studio/applications-service` in hab studio prior to running `applications_populate_database`

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
SEAjamieD: good point - thanks for the suggestion!

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.ts
SEAjamieD: good idea - I chose to use close/open accordion for a bit more specificity of whats goin on in the template.

apriofrost:
Looking great! @SEAjamieD I love the sliding up and down animation you added for the transitional state.
I wonder if we could get rid of the blue box outline as the focus state once the button is clicked?

SEAjamieD:
@apriofrost 
> I wonder if we could get rid of the blue box outline as the focus state once the button is clicked?

Generally its considered poor for accessibility to remove that blue focus outline (womp wommmp), so I don't recommend removing it.  _however_ there are some alternative plans if you're open to that?

1.  We could apply, across the whole application, that if a user is using the mouse, the focus ring doesn't show.  If the user if using the keyboard to navigate - we can make sure it _does_ show. 

2.  We could style it, across the whole application, so that it at least looks nice.

3.  We could do both 1 + 2 (you can see an example of this on www.zumiez.com).  When you are tabbing you'll see a styled focus ring, but if you touch the mouse while tabbing the ring will go away.

If you're really attached to removing it - we can do that, but it's a bad look for accessibility.  There's also lots of different options to styling the focus ring we can look into.

apriofrost:
@SEAjamieD Hi Jamie! Thanks for the great suggestions. I talked to the UX team about it. Since this would be a global change, and we need more time to discuss the best approach for accessibility. I would suggest that we ignore this feedback for this issue for now.
I'll make another card after the UX team figure out a solution. Thank you again!

Issue #2828, opened 2020-02-09
Automate api Docs site broken in Chrome


msorens:
## Describe the bug

Ensure Chrome is updated to version 80.

Note: not 100% sure Chrome v80 is the culprit but several factors lend credence to the assertion:
(a) Chrome 80 caused other rendering problems in Automate itself (see https://github.com/chef/automate/pull/2825),
(b) docs are working fine in Firefox, and
(c) no one has reported this as a problem previously indicating it is likely very recent and thus coincident with the Chrome 80 update.

Load https://automate.chef.io/docs/api/#operation/Node%20Export in Chrome.
At first glance the page looks OK, but as soon as you try to interact you should see odd behaviors.

Example:
Scroll the main panel a little bit up or down and the navigation panel seems to react properly, changing the current topic indicator. But scroll the main panel up to the top of the Reporting section and the navigation panel goes away. In reality, it has jumped to a fixed position (now offscreen) at the top of the viewport.

Scroll all the way to the top to bring the navigation panel into view. Try expanding items in the navigation panel and it does not respond.

## Expected behavior
Interaction works correctly, as it does in Firefox.

## Versions (please complete the following information):
- OS: macOS
- Browser: Chrome v80

susanev:
@msorens can you add a video? im not able to repro this but also dont fully understand your description.

susanev:
i am able to repro. seems though refreshing fixes this as well. ill add that the api docs are affected as well in the known issues of the release notes.

phiggins:
It sounds like this might have been fixed upstream in redoc. I can investigate tomorrow AM. https://github.com/Redocly/redoc/issues/1168

phiggins:
I've got a PR open to fix this https://github.com/chef/automate/pull/2835

PR #2835, opened 2020-02-10
Track latest redoc release in API docs

phiggins:
To fix problems with the API docs not behaving properly with Chrome 80 we need to track the latest release of redoc. The problem is described in #2828.

We were already attempting to track the latest release but for some reason were one behind so this is not really a change in the desired behavior. We should probably be more explicit about the version of redoc used but that is another discussion.

### DIFF
--- a/components/automate-chef-io/layouts/_default/data-api.html
+++ b/components/automate-chef-io/layouts/_default/data-api.html
@@ -19,7 +19,7 @@
   </head>
   <body>
     <ReDoc spec-url="/api-docs/all-apis.swagger.json"></ReDoc>
-    <script src="https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js"></script>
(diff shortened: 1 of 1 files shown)


components/automate-chef-io/layouts/_default/data-api.html
```diff
@@ -19,7 +19,7 @@
   </head>
   <body>
     <ReDoc spec-url="/api-docs/all-apis.swagger.json"></ReDoc>
-    <script src="https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/redoc/bundles/redoc.standalone.js"></script>
```
phiggins: Removing the `@next` tag loads 2.0.0-rc.23 rather than 2.0.0-rc.22. According to [the jsdelivr docs)[https://www.jsdelivr.com/features] it should be possible to do something like `@2.0` to guard against accidentally upgrading to the next major version but I couldn't get that to work.

PR #2836, opened 2020-02-10
Restart hab sup after upgrading it on pipeline boxes

danielsdeleo:

* restart habitat supervisor on the `a2-perf-test-chef-load` machines after we upgrade it
* set the env VAR to turn off progress bars in hab so the build logs are more readable

Chef employees can see the first two (failing) deployment builds here: https://buildkite.com/chef/chef-automate-master-deploy-acceptance/builds/72


jaym:
Merged in #2844

PR #2691, opened 2020-01-22
Automate-835 bottom user notifications

tarablack01:
### :nut_and_bolt: Description: What code changed, and why?
For IAM v2.1 beta, we need a way to globally communicate to all users that a long background process is currently happening. This task is to add a user-notification banner to the bottom of the page, there will be 2 designs, one to display a progress bar and one to display other information.

REMAINING ITEMS:
1. info icon modal details
2. Page layout resize to show bottom user notifications without scrolling
3. Code cleanup

### :chains: Related Resources
https://app.zenhub.com/workspaces/ui-team-5cba23a3b14fdc05970c8f87/issues/chef/automate/835

### :+1: Definition of Done
https://app.zenhub.com/workspaces/ui-team-5cba23a3b14fdc05970c8f87/issues/chef/automate/835

### :athletic_shoe: How to Build and Test the Change
hab studio enter
export DEVPROXY_URL=localhost
build components/automate-ui-devproxy
start_automate_ui_background |  stop_automate_ui_background
start_all_services

### :camera: Screenshots, if applicable
Screen Shot 2020-01-22 at 4 16 23 PM
Screen Shot 2020-01-22 at 4 16 37 PM
Screen Shot 2020-01-22 at 4 16 50 PM
Screen Shot 2020-01-22 at 4 16 57 PM

PR #2838, opened 2020-02-10
Correcting disposition of ListProjects for IAM v1

msorens:
### UPDATED DESCRIPTION

Just treating v1 as a problem; V2 behavior is sufficient as is--without this commit, to wit:
viewer, editor, or better will not see an error, and will see projects status banner
user with no permissions will see an error--but that's OK. It is not a real use case (at most, a user is just in the process of being set up)

image

### ORIGINAL DESCRIPTION
Initially, this surfaced as a v1 issue but it turns out it affects both v1 and v2, though differently.

#### v2 (now obsolete)
_Actual (prior to this PR):_
Only admin users (or users explicitly given permission) could access `ListProjects`; other users would see an error banner.

_Desired (with this PR):_
All users need to see the projects status bar when projects are not up to date (though only folks with permission to update should see the <kbd>update projects</kbd> and <kbd>stop update</kbd> buttons).

#### v1
_Actual (prior to this PR):_
The system attempts to invoke `ListProjects` for all users; an error banner ensues for non-admin users.

_Desired (with this PR):_
The system no longer attempts to invoke `ListProjects` for any user.

#### Trade-Off (now obsolete)
This PR is the lesser of two evils. It fixes the issue as outlined above, but this quick-fix requires giving the user permission to view the projects list page, which is not desirable.
A better fix would be to introduce a new endpoint (called perhaps `AnyProjectsPending`) that returns a simple boolean, as that is all the front-end actually needs. Though fairly straightforward, this requires some team discussion and a bit more thought, so I deemed it not appropriate when trying to get acceptance unblocked.

#### Details

- Scope: affects all non-admin users, IAM v1
- In dev? 2020-02-06
- In acceptance? 2020-02-10
- In current? no

- PR of origin: "Automate-835 bottom user notifications" (https://github.com/chef/automate/pull/2691)
- Commit of origin: a02150a5d

### :+1: Definition of Done
v1: No error and no project update bar.

rebuild automate-ui (e.g. make serve)

Create a non-admin user. No need to add any particular policies.
Run the system in v1 (`chef-automate iam reset-to-v1`) mode logged in as the non-admin user.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/authz-service/server/v2/system.go
+++ b/components/authz-service/server/v2/system.go
@@ -66,6 +66,12 @@ func SystemPolicies() []*storage.Policy {
 				Resources: []string{"iam:rules"},
 				Projects:  []string{constants.AllProjectsID},
 			},
+			{
+				Effect:    storage.Allow,
+				Actions:   []string{"iam:projects:list"},
+				Resources: []string{"iam:projects"},
+				Projects:  []string{constants.AllProjectsID},
+			},
 		},
 	}
 
--- a/components/authz-service/server/v2/system.go
+++ b/components/authz-service/server/v2/system.go
@@ -66,12 +66,6 @@ func SystemPolicies() []*storage.Policy {
 				Resources: []string{"iam:rules"},
 				Projects:  []string{constants.AllProjectsID},
 			},
-			{
-				Effect:    storage.Allow,
-				Actions:   []string{"iam:projects:list"},
-				Resources: []string{"iam:projects"},
-				Projects:  []string{constants.AllProjectsID},
-			},
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


blakestier:
The code seems fine, but I'm wondering: We have an `ApplyRulesStatus` call, right? Could we orient permissions around that, rather than `ListProjects`?

susanev:
im super confused about parts of ur preamble, both the editor and viewer roles have `iam:projects:list`... so is that not working correctly?

components/authz-service/server/v2/system.go
```diff
@@ -66,6 +66,12 @@ func SystemPolicies() []*storage.Policy {
 				Resources: []string{"iam:rules"},
 				Projects:  []string{constants.AllProjectsID},
 			},
+			{
+				Effect:    storage.Allow,
```
susanev: we shouldnt need this as a system policy, ill be around later tonight if you want to chat more. i think i miscommunicated earlier.

PR #2845, opened 2020-02-11
[Automate-2830] Auto-document IAM actions in API documentation section

msorens:

We want to document all IAM actions because users need to know these when crafting their own policies. Arguably it is more important to document for the UI, this PR starts off with the low-hanging fruit of documenting the API, which can be done programmatically. 

#### Algorithm
```
 for each protobuf file
   for each endpoint
     if not already action-annotated but has manually-create doc section
       grab action from the IAM v2 action (`(chef.automate.api.iam.policy).action`)
       and add a notation to that manual doc section 
```

#### Result
image

#### Sausage Making Details
The engine for this work is an awk script (scripts/add_action_doc_to_proto.awk) that does the inner loop of the pseudo-code above.
The outer loop is handled by updates to the grpc.sh for `automate-gateway` and for `api`, the two components containing public and externally documented endpoints.

Specifically, the awk script is embedded in this shell function whose net effect is to rewrite the file in place:
```
doc_action() {
  awk -f ./scripts/add_action_doc_to_proto.awk $1 > data.tmp && mv data.tmp $1 ;
}
export -f doc_action
```
You can run this function manually on a **single** proto file to see it in action (pun intended). Be sure you are in the root of the automate repo and give the path to the proto file:
```
doc_action components/automate-gateway/api/compliance/reporting/stats/stats.proto
```
You can run this function manually on **all** proto files as well, again from the root of the automate repo. This finds all proto files with endpoints (containing `rpc`) and requisite manual documentation (the presence of a multi-line comment token `*/`) and calls `doc_action` on each. False positives are tolerated; the awk script will ignore endpoints or whole files that do not need any annotation.
```
grep --include=\*.proto -rwl rpc . | xargs grep -l '\*\/' | \
xargs -t -n1 -P1 bash -c 'doc_action "$@"' _
```

Once the proto files have been annotated, they then need to go through the normal process of regeneration: running in hab studio either `compile_go_protobuf_component automate-gateway` or `compile_go_protobuf_component api` will update necessary *.pb.go, *_swagger.json, and *.pb.swagger.go files.
Because you will always need to do these two steps one right after the other, that `grep... xargs... doc_action` has been embedded right into  `compile_go_protobuf_component` so it is done automatically.

But wait, not done yet! We still have to update the base documentation files (*.swagger.json in automate-chef-io/data/docs/api_chef_automate):
```
cd automate-chef-io
make sync_swagger_files
```
Everything up to here has been already done and included in this PR.
To actually see the generated documentation with the new action documentation you need to start the documentation server locally:
```
cd automate-chef-io
make serve
```
and finally load `http://localhost:1313/docs/api/` in your browser.

NB: For reviewing, there are only three code files changed:

- scripts/add_action_doc_to_proto.awk
- components/automate-gateway/scripts/grpc.sh
- api/scripts/grpc.sh

All the remaining files are the now-annotated protobuf files and the files regenerated from them.

NA

### :+1: Definition of Done
1. All previously documented protobuf files now have IAM actions documented in the API docs. For production this is found at https://automate.chef.io/docs/api/, for acceptance, https://a2-docs-acceptance.cd.chef.co/docs/api/.
2. Whenever `compile_go_protobuf_component automate-gateway` or `compile_go_protobuf_component api` is used to rebuild boilerplate from protobuf files, they are first checked for any endpoints requiring annotation per the above pseudo-code, then the normal regeneration process continues, automatically incorporating the action documenting with no additional manual steps.

See above commentary.

- [ ] Tests added/updated?
- [x] Docs added/updated?


api/scripts/grpc.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
 # set -x
 
+echo "Annotate authorization actions"
+
+doc_action() { awk -f ./scripts/add_action_doc_to_proto.awk $1 > data.tmp && mv data.tmp $1 ; }
+export -f doc_action
+grep --include=\*.proto -rwl rpc api | xargs grep -l '\*\/' | xargs -t -n1 -P1 bash -c 'doc_action "$@"' _
```
msorens: The PR preamble describes in detail what this does. Note that here L8 has been customized to run only the files in the root-level `api` directory.

components/automate-gateway/scripts/grpc.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
 # set -x
 
+echo "Annotate authorization actions"
+
+doc_action() { awk -f ./scripts/add_action_doc_to_proto.awk $1 > data.tmp && mv data.tmp $1 ; }
+export -f doc_action
+grep --include=\*.proto -rwl rpc components/automate-gateway | xargs grep -l '\*\/' | xargs -t -n1 -P1 bash -c 'doc_action "$@"' _
```
msorens: And here it is again for gateway.

scripts/add_action_doc_to_proto.awk
```diff
@@ -0,0 +1,69 @@
+BEGIN {
```
msorens: Who doesn't love a good awk script? I have documented the key points, but assumes some basic familiarity with awk.

scripts/add_action_doc_to_proto.awk
```diff
@@ -0,0 +1,69 @@
...
+
+!collecting {
+  print
+}
+
+function replayBuffer(){
+  for (i = 1; i < collecting; i++) print buffer[i]
+}
+
+# Define shell function to run this awk script for one file, sitting in automate root directory:
```
msorens: Since the code will soon be far removed from the PR, I have re-iterated the instructions on using this here in the awk script itself.

components/automate-chef-io/README.md
```diff
@@ -124,15 +124,15 @@ http://localhost:1313/docs/api
...
+Chef Automate generates API documentation from proto files comments using
 
 * [grpc-gateway protoc-gen-swagger](https://github.com/grpc-ecosystem/grpc-gateway/tree/master/protoc-gen-swagger) as the protoc compiler
 * [OpenAPI 2.0 specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md) as the API schema
 * [ReDoc](https://github.com/Redocly/redoc) as the front-end display
 
 To build the API documentation during development:
 
-1. Make any edits to the .proto files necessary.
+1. Make any edits to the .proto files necessary. Note that an "Authorization Action:" section will be added automatically in the next step.
```
phiggins: Should this get mentioned in the next step?

msorens: Great question! My reason for attaching to this step is that this is where the engineer is editing and, since most of our engineers have a pretty keen eye, they would likely have noticed that annotation in other files and naturally would just copy that format. That's OK (because if it is manually put in there the generator will see it and not add it again) but the point here is that the engineer would see that they do not need to waste time doing that as it will be done for them. (Sorry for rambling.)

scripts/add_action_doc_to_proto.awk
phiggins: Is there an advantage to using awk rather than something like ruby that more folks at the company will likely feel comfortable working with?

msorens: Just saw your comment after I merged. Probably no advantage, no. Given the linux expertise in the company I reasoned that awk would be an acceptable choice; I did not really consider ruby because I have not personally used it much.

PR #2847, opened 2020-02-11
Don't strip applications service binary

jaym:
This will make it possible for us to debug in the perf test environment.
That environment is showing 7k and growing goroutines, but we cannot
analyze the core dump with debug symbols

### DIFF
--- a/components/applications-service/habitat/plan.sh
+++ b/components/applications-service/habitat/plan.sh
@@ -44,3 +44,7 @@ do_install() {
   mkdir "${pkg_prefix}/schema"
   cp -r pkg/storage/postgres/schema/sql/* "${pkg_prefix}/schema"
 }
+
+do_strip() {
+  return 0
+}


danielsdeleo:
FWIW there's only 3 occurrences of `go thing()` in applications-service: 

https://github.com/chef/automate/blob/c1757004d242546edfa16e06fd339fc87086fdaf/components/applications-service/cmd/applications-service/commands/serve.go#L59
https://github.com/chef/automate/blob/c1757004d242546edfa16e06fd339fc87086fdaf/components/applications-service/cmd/applications-service/commands/serve.go#L62
https://github.com/chef/automate/blob/c1757004d242546edfa16e06fd339fc87086fdaf/components/applications-service/pkg/ingester/v1/ingest.go#L77-L79 (this makes 25 workers)

jaym:
ya, its unclear where they are all coming from

PR #2858, opened 2020-02-12
Dan/bump pq

danielsdeleo:

Giving it another go to see what happens. Probably won't actively work on it a lot if it's not easy so feel free to take over if you happen upon this PR and need it done.

- [ ] Tests added/updated?
- [ ] Docs added/updated?


danielsdeleo:
To explain the approach here:

It seems the automatic reconnect and retry logic in lib/pq is kinda dangerous as-is: https://github.com/lib/pq/issues/939 and attempts to bring back the previous retrying behavior petered out because of safety concerns: https://github.com/lib/pq/pull/871

So it seems the best path forward is to assume the retrying behavior is not coming back in lib/pq, or at least not soon. I also think it's fine for Automate to return one 500 after a database disconnect/reconnnect. If you agree with that then the sensible thing to do is to make our tests a little more forgiving of the occasional 500 that we would expect to see from the database resets.

integration/tests/cluster.sh
```diff
@@ -92,8 +92,13 @@ do_test_deploy() {
     export ELASTICSEARCH_URL="http://$frontend1_ip:10144"
     test_notifications_endpoint="http://$test_container_ip:15555"
 
-    run_inspec_tests "${A2_ROOT_DIR}" "a2-iam-v2-integration"
+    # Inspec tests are less tolerant of transient 500s; we expect a few of
+    # those as bad postgres connections get purged from the various services.
+    # Restart everything before running the tests
+    docker exec -t "$_frontend1_container_name"  "$cli_bin" restart-services
```
jaym: i'm not sure i understand whats happening here. We've just done a deploy. Why does restarting everything help

danielsdeleo: The frontends come up with broken Postgres connections, so the Inspec tests get exactly 2 500s, see for example https://buildkite.com/chef/chef-automate-master-verify-private/builds/9454#d349a780-69be-4a0f-ac17-10f6ae8f2475. Retrying 500s in the Inspec would be the better solution but it's a lot harder.

danielsdeleo: For anyone who might stumble on this later, the code comment above is the correct answer--an HAProxy somewhere has the idle timeout set to 5 minutes, which kills the connections on frontend1 during the more-than-five-minutes when frontend2 is being deployed.

components/automate-pg-gateway/habitat/default.toml
```diff
@@ -5,3 +5,5 @@ status_port = 10146
 
 [timeouts]
 connect = 5
+# 12 hours
+idle = 43200
```
jaym: should we wire this up to the config so its easy to change

Issue #2348, opened 2019-11-22
Nodemanager Read node duplicate projects


lancewf:
## User Story

When requesting a node from the nodemanager duplicate projects are returned. Below is an example where the node is tagged with "project9".
```
{
  "id": "92ee51bd-831c-4c5b-b738-239f47dd7296",
  "name": "insights.B",
  "platform": "redhat 8.9",
  "platform_version": "8.9",
  "manager": "",
  "tags": [
    {
      "key": "chef-tag",
      "value": "slave"
    },
    {
      "key": "chef-tag",
      "value": "slave"
    },
    {
      "key": "chef-tag",
      "value": "supermarket"
    },
    {
      "key": "chef-tag",
      "value": "supermarket"
    },
    {
      "key": "chef-tag",
      "value": "builder"
    },
    {
      "key": "chef-tag",
      "value": "builder"
    }
  ],
  "last_contact": "2018-06-28T15:19:22Z",
  "status": "unknown",
  "last_job": null,
  "target_config": null,
  "manager_ids": [],
  "connection_error": "",
  "state": "",
  "name_prefix": "",
  "projects": [
    "project3",
    "project3",
    "project3",
    "project9",
    "project9",
    "project9"
  ],
  "run_data": {
    "id": "e2d05569-dfeb-4ff0-a5a2-863ce12cc134",
    "status": "PASSED",
    "penultimate_status": "UNKNOWN",
    "end_time": "2018-06-28T15:19:22Z"
  },
  "scan_data": {
    "id": "",
    "status": "UNKNOWN",
    "penultimate_status": "UNKNOWN",
    "end_time": null
  }
}
```

## Definition of Done

Duplication projects are not returned when requesting a node

## Demo Script / Repro Steps
1. `start_all_services`
1. Ingest a client run node with `send_chef_run_example`
1. Create a project with an ingestion rule that matching on environment = "B"
1. Apply the project rules. 
1. Go to https://a2-dev.test/infrastructure/client-runs and copy the node ID
1. To retrieve the node from the node manager run `curl -X GET -f --insecure -H "api-token: $(get_admin_token)" -H "projects: project9" https://localhost/api/v0/nodes/id/92ee51bd-831c-4c5b-b738-239f47dd7296` Replace "project9" with the project created above. Replace the UUID at the end with the node ID. 
1. Ensure the projects field does not have duplicates.

PR #2859, opened 2020-02-12
Prevent duplicate data from aggregated data when fetching nodes

phiggins:
Fix for #2348.

As far as I can tell, this bug was caused by the [sql query that fetches nodes](https://github.com/chef/automate/blob/master/components/nodemanager-service/pgdb/nodes.go#L40-L76) doing multiple joins and aggregations at once. I initially solved this by changing the query to do each aggregation separately and joining against teh result, but some of the code we use for filtering depends on the query doing the joins how it is currently done, and redoing that did not seem like a great use of time.

Using `DISTINCT` like I did in this PR solves the immediate issue, but from what I've gathered this is sort of a big hammer that only papers over the original problem. Using `DISTINCT` like this has the following downsides:

* The query is bigger and more complicated than it needs to be. The sql is kind of straigtforward to read, but results in complicated behavior for postgres because it builds up a big data set before anything is done to whittle it down. In other words, if these tables were big this query might not work very well.
* There might be other behavior or side effects that are obscured. For example, for the aggregations this query does, grouping manager ids, project ids, and tags, unique values are desired, but it wouldn't be possible to add an aggregation where repeated values were the correct behavior.

### DIFF
--- a/components/compliance-service/api/tests/run.rb
+++ b/components/compliance-service/api/tests/run.rb
@@ -16,6 +16,7 @@ require 'json'
 require 'pry'
 require 'grpc'
 require 'deepsort'
+require 'set'
 
 module VulcanoTest
 
@@ -109,6 +110,10 @@ module VulcanoTest
 
   def assert_equal_json_sorted(expected, actual)
     assert_equal JSON.parse(expected).deep_sort, JSON.parse(actual).deep_sort
+  end
+
+  def assert_same_elements(expected, actual)
+    assert_equal(expected.to_set, actual.to_set)
   end
 end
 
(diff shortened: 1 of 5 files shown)


components/compliance-service/api/tests/run.rb
```diff
@@ -110,6 +111,10 @@ def assert_equal_json_content(expected, actual)
   def assert_equal_json_sorted(expected, actual)
     assert_equal JSON.parse(expected).deep_sort, JSON.parse(actual).deep_sort
   end
+
+  def assert_same_elements(expected, actual)
+    assert_equal(expected.to_set, actual.to_set)
+  end
```
phiggins: Adding `DISTINCT` seemed to have the side effect of making postgres alphabetize the values(?), so rather than change the expected values for the tests or add `.sort` calls all over the place, I added this test helper that tests that the same elements were returned without caring about the order.

PR #2862, opened 2020-02-12
[elasticsearch] Make max shards per node a configurable option

ryancragun:
After an ElasticSearch cluster failure and recovery in the long-lived performance environment we noticed that new shards were being created but remained unallocated. After investigation it was determined that the new shards were created as part of new CREATE INDEX requests failing due to the cluster exceeding the default number of max allowable shards (1000). In ElasticSearch 6 the default number of shards per index is 5, therefore with a default cluster we can only retain ~50 days worth of data assuming the 4 daily timeseries are being created daily.

Here we bump the default on the performance environments to a more reasonable number (1500, of which about 1250 should allocated on a regular basis). Due to the hab package pinning this change won't make it to those environments until we decide to bump ElasticSearch again, so the persistent configuration has been set manually until that time.


terraform/test-environments/modules/chef_automate_install/main.tf
```diff
@@ -264,6 +264,8 @@ ${var.saml == "true" ? local.saml_config : ""}
 
 [elasticsearch.v1.sys.runtime]
   heapsize = "${var.enable_cloudwatch_metrics == "true" ? "16g" : "2g"}"
+[elasticsearch.v1.sys.cluster]
+  max_shards_per_node = ${var.enable_cloudwatch_metrics == "true" ? "1500" : "1000"}
```
jaym: what about cloudwatch metrics is causing problems. Or is that an indicator for our perf environments

ryancragun: The latter

PR #2869, opened 2020-02-13
Update notifications service dependencies

jaym:
Bring in updated grpc, protobuf, and cowboy. The new protobuf library
changed default enum values. Unfortunately, both atoms and integers are
allowed, and there's no easy way to enforce consistency.

Should address the remaining of #1245


components/notifications-service/server/lib/data/store.ex
```diff
@@ -211,8 +211,23 @@ defmodule Notifications.Data.Store do
      unqualified_type_name(type), url, secret_id, critical_controls_only]
   end
 
-  defp rule_event_to_value(event), do: Rule.Event.value(String.to_atom(event))
-  defp rule_value_to_event(value), do: Rule.Event.key(value)
+  defp rule_event_to_value(event) when is_integer(event) do 
+    event
+  end
+
+  defp rule_event_to_value(event) do
```
ryancragun: :warning: I don't know elixir.

Does this need to be guarded with `is_string()`? Otherwise it'll match before the atom matcher.

jaym: i think that case never happens, but ill add one for string and one for atom just in case

PR #2871, opened 2020-02-14
[elasticsearch] configure the default replicas to 0

ryancragun:
* Make the refresh_interval and number_of_replicas configurable
* Change the number_of_replicas from 1 to 0 for single node installs
  because it's impossible to replicate a shard with a single node


components/automate-elasticsearch/habitat/hooks/post-run
```diff
@@ -0,0 +1,36 @@
...
+
+{{#if svc.me.cfg.http-host ~}}
+HOST="{{svc.me.sys.ip}}"
+{{else}}
+HOST="127.0.0.1"
+{{/if ~}}
+
+source {{pkg.svc_config_path}}/health_check
+
+wait_until_healthy 300 # seconds
```
jaym: i think past experience has shown blocking in hooks is not good. https://github.com/habitat-sh/habitat/pull/6705 indicates that it will retry if you exit with non-zero

components/automate-elasticsearch/habitat/hooks/post-run
```diff
@@ -0,0 +1,36 @@
+#!{{pkgPathFor "core/bash"}}/bin/bash
```
jaym: i think this post-run hook will also get run in the HA deployment

components/automate-elasticsearch/habitat/hooks/post-run
```diff
@@ -0,0 +1,41 @@
...
+
+{{#if svc.me.cfg.http-host ~}}
+HOST="{{svc.me.sys.ip}}"
+{{else}}
+HOST="127.0.0.1"
+{{/if ~}}
+
+source {{pkg.svc_config_path}}/health_check
+
+if [[ $(wait_until_healthy 300) -ne 0 ]]; then
```
jaym: i think past experience has shown blocking in hooks is not good. habitat-sh/habitat#6705 indicates that it will retry if you exit with non-zero. The ha services also did something similar to this but ended up ditching it because it caused some problems

components/automate-elasticsearch/habitat/hooks/post-run
```diff
@@ -0,0 +1,41 @@
+#!{{pkgPathFor "core/bash"}}/bin/bash
+
```
jaym: do we need to set -x

stevendanna: Do you mean `set -x`?

components/automate-elasticsearch/habitat/hooks/post-run
```diff
@@ -0,0 +1,41 @@
...
+exec 2>&1
+
+{{#if cfg.disable ~}}
```
jaym: we probably also want to check if ES is configured to be a proxy node and not mess with the cluster as well

components/automate-elasticsearch/habitat/hooks/post-run
jaym: yes, that, the error one

components/automate-elasticsearch/habitat/hooks/post-run
ryancragun: For posterity: previous versions of hab ran the hooks in the main supervisor thread so any long operation in them would block the supervisor. Now hooks are run async so it's okay to block.

components/automate-elasticsearch/habitat/hooks/post-run
ryancragun: I'm intentionally not failing at all in the post-run hook because it's a best effort attempt. Since there is no backoff of retries and hab doesn't currently reap hooks when a service is unloaded, that seems like the safest approach.

components/automate-elasticsearch/habitat/hooks/post-run
```diff
@@ -0,0 +1,69 @@
...
+# We're using this hook to configure the default index settings after the service
+# has initialized enough to service queries.
+
+# The post-run hook is fired immediately after the run hook, therefore it's
+# unlikely that elasticsearch is healthy. We call the health-check wait_until_healthy
+# function which waits until ElasticSearch returns green or yellow.
+
+# We give it a default timeout of 600 seconds to ensure that the post-run
+# hook will eventually time out. We do this because Habitat does not currently
+# reap hooks if the service is unloaded. We don't want to end up in a place
```
jaym: I think weird stuff is going to happen here if ES is in a failure loop in that these post-run hooks are going to likely run longer than it takes ES to start again and fire off another post-run hook. It'll probably be fine if it gets fixed quickly, but if it goes for a while, we're going to have a lot of health checks against ES running every second

jaym: A better strategy might be to check the health, sleep for a few seconds if it fails, then return 1. That should get around 7494

PR #2874, opened 2020-02-14
Upgrade nats and nats-streaming clients and servers to latest

danielsdeleo:

replaces #2282

This allows us to stop using patched versions of nats and nats-streaming servers (our patches have long since been merged), and gets us on versions which are updated for go modules. This has been blocked on upgrading `lib/pq`, so this PR is initially branched from a PR which updates that library.

### :+1: Definition of Done

Tests pass with updated dependencies


vjeffrey:
@danielsdeleo anything in particular that reviewers should test out in their dev envs related to the changes made here?

danielsdeleo:
The tests should really cover most of what you'd test manually, but if you like, there are a few routes to test it locally:

#### Diagnostics

- Build automate-cli, event-service, event-gateway, and applications service
- start all services
- `chef-automate diagnostics run applications --skip-cleanup`
- In the applications tab UI, you should see 4 service groups with names like `db.diagnostics-run-e98840648183e43aa8eeb2f0c1c38b7bc548ba1831c26f85f0e59ee51f041130`
- You can delete these by running `chef-automate diagnostics run applications --skip-generate`

#### End to End with Habitat in Docker

- Build all the same services as above and start all services
- Follow the instructions here to run the docker-based end to end tester: https://github.com/chef/automate/blob/master/components/applications-service/dev/hab-client/README.md

PR #2891, opened 2020-02-18
fix project update progress percentage

susanev:
We were displaying a fractional percentage, this fixes it.

### :+1: Definition of Done
The percentage format that is displayed in stop modal matches the one displayed in the banner.

- `rebuild components/automate-ui-devproxy`
- load 500 nodes with either `chef_load_nodes 500` or `chef_load_compliance_nodes 500`
- create a project
- add an ingest rule to it, it doesnt matter the content of the ingest rule
- run a project update
    - note: if you only ever see 0% or 100% youll need to add more nodes

- [ ] Tests added/updated?
- [ ] Docs added/updated?

#### before
incorrect-percentage

#### after
correct-percentage

### DIFF
--- a/components/automate-ui/src/app/components/process-progress-bar/process-progress-bar.component.ts
+++ b/components/automate-ui/src/app/components/process-progress-bar/process-progress-bar.component.ts
@@ -38,7 +38,7 @@ export class ProcessProgressBarComponent implements OnInit {
         }
         this.applyRulesInProgress = state === ApplyRulesStatusState.Running;
         if (!this.cancelRulesInProgress && state === ApplyRulesStatusState.Running) {
-          this.percentageComplete = percentageComplete;
+          this.percentageComplete = Math.floor(percentageComplete * 100);
         }
         this.layoutFacade.layout.userNotifications.updatesProcessing = this.applyRulesInProgress;
         this.layoutFacade.updateDisplay();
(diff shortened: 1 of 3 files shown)


components/automate-ui/src/app/components/process-progress-bar/process-progress-bar.component.ts
```diff
@@ -38,7 +38,7 @@ export class ProcessProgressBarComponent implements OnInit {
         }
         this.applyRulesInProgress = state === ApplyRulesStatusState.Running;
         if (!this.cancelRulesInProgress && state === ApplyRulesStatusState.Running) {
-          this.percentageComplete = percentageComplete;
+          this.percentageComplete = Math.floor(percentageComplete * 100);
```
msorens: This will work, in that it completes the task of all accesses to "percentageComplete" handling it the same way, multiplying by 100. However, we could alternately make the fix more robust by changing the source in the UI rather than each use of it in the UI.

That would be changing this line:
https://github.com/chef/automate/blob/8afdaf15b72920fe5af046b18ee3ebe25f194f13/components/automate-ui/src/app/entities/projects/project.reducer.ts#L129
to this:
```
      return set(
        APPLY_RULES_STATUS,
        mapKeys(key => camelCase(key),
          {
            ...action.payload,
            percentage_complete: action.payload.percentage_complete * 100
          }),
        state);
```

susanev: okay, i will try.. thank you!

PR #2895, opened 2020-02-18
Jdm/bump pq

jaym:


danielsdeleo:
Thanks for your help! We finished this in https://github.com/chef/automate/pull/2858 and some other PRs

PR #2903, opened 2020-02-18
[automate-2703] spinner forever on project updates

msorens:

Intermittently we were getting this error upon starting a project update--spinner appears forever and a console error (ignore the part about it saying you have no projects yet 😁 ):

Screen Shot 2020-01-22 at 4 08 01 PM

### :+1: Definition of Done
1. Spinner shows up only briefly then the projects list is redisplayed on the projects list page.
2. no console error

rebuild automate-ui
create some projects and rules that need updated
(not sure if it is technically needed, but issue seems to occur more when there are nodes in the system, e.g. `chef_load_nodes 1000` in hab studio).
Set your brower to a slow connection ("Slow 3G"), then start a project update.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/entities/userperms/userperms.effects.ts
+++ b/components/automate-ui/src/app/entities/userperms/userperms.effects.ts
@@ -10,7 +10,7 @@ import {
   withLatestFrom,
   switchMap
 } from 'rxjs/operators';
-import { of, Observable } from 'rxjs';
+import { of } from 'rxjs';
 import * as moment from 'moment';
 
 import { NgrxStateAtom } from 'app/ngrx.reducers';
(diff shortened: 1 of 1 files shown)


components/automate-ui/src/app/entities/userperms/userperms.effects.ts
```diff
@@ -65,18 +64,15 @@ export class UserPermEffects {
...
       ([[action, lastTime], list]: [[GetSomeUserPerms, Date], IndexedEntities<UserPermEntity>]) => {
-        let obs: Observable<UserPermsActions>;
         const isFresh = !!list && !this.stale(lastTime);
-        if (isFresh) {
+        return isFresh ?
           // TODO: This works, but it has extraneous 'id' property in the payload.
           // Should strip that from each entry in the list.
-          obs = of(new GetSomeUserPermsSuccess(list));
-        } else {
-          this.requests.fetchSome(action.payload).pipe(
```
msorens: The error was that the `else` block was not returning anything, so it returned "undefined".

Issue #2703, opened 2020-01-23
Update Projects button hits race condition on slow connections


bcmdarroch:
## Describe the bug
On slow connections, the `Update Projects` button sometimes hits a race condition error.

## To Reproduce
- Log in to the UI as an admin
- create several projects
- add a rule or two to some projects
- go to the `Network` tab of dev tools and set your network speed to `Slow 3G` (option under `Online`)
- click Update Projects on the project list page
- you should get stuck on this screen with a console error:
![Screen Shot 2020-01-22 at 4 08 01 PM](https://user-images.githubusercontent.com/21015366/72947829-33846900-3d38-11ea-8dc0-3d168aaa820b.png)
```
ERROR TypeError: You provided 'undefined' where a stream was expected. You can provide an Observable, Promise, Array, or Iterable.
    at subscribeTo (subscribeTo.js:27)
    at subscribeToResult (subscribeToResult.js:11)
    at SwitchMapSubscriber._innerSub (switchMap.js:46)
    at SwitchMapSubscriber._next (switchMap.js:36)
    at SwitchMapSubscriber.next (Subscriber.js:49)
    at WithLatestFromSubscriber._next (withLatestFrom.js:57)
    at WithLatestFromSubscriber.next (Subscriber.js:49)
    at WithLatestFromSubscriber._next (withLatestFrom.js:57)
    at WithLatestFromSubscriber.next (Subscriber.js:49)
    at FilterSubscriber._next (filter.js:33)
```
(you may to try a few times)

## Expected behavior
No error, project progress bar should pop up at the bottom of the screen.

## Versions (please complete the following information):
- Browser: chrome
- Automate Build Number: 20200122235425

PR #2905, opened 2020-02-18
[automate-2904] Remove state-clearing upon initiating "get all"

msorens:

Closely tied to #2703, during that project update--if you were on the projects list page, notice that the background behind the spinner shows there are no projects.

(Weird! Unable to load the image here.)

That is due to the list of projects being erased at the start of the project fetch that is coincident with a Project Update.

That list should not have actually been erased--we may have been a bit over-eager in state clearing in a couple
earlier PRs, notably https://github.com/chef/automate/pull/2158.
I think it still makes sense for individual "get" calls to do some state clearning,
but "get all" does not need it; when the result is returned and
added to the state with an `addAll` call, that completely replaces
the section of state at hand.

Reference: https://ngrx.io/guide/entity/adapter#adapter-collection-methods

Technically, there is only one page affected--the projects list, because a `GetAllProjects` is done with a Project Update. Which means only the project reducer needs to be fixed. However, if it is a legitimate fix for that file, then it should be done across the board (as I have done), to keep consistent in what we consider best practices.

See above

### :+1: Definition of Done
Project list remains behind the spinner during project update

rebuild automate-ui
create some projects and rules that need updated
(not sure if it is technically needed, but issue seems to occur more when there are nodes in the system, e.g. `chef_load_nodes 1000` in hab studio).
Set your brower to a slow connection ("Slow 3G"), then start a project update.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/entities/projects/project.reducer.ts
+++ b/components/automate-ui/src/app/entities/projects/project.reducer.ts
@@ -65,7 +65,7 @@ export function projectEntityReducer(
 
   switch (action.type) {
     case ProjectActionTypes.GET_ALL:
-      return set(GET_ALL_STATUS, EntityStatus.loading, projectEntityAdapter.removeAll(state));
+      return set(GET_ALL_STATUS, EntityStatus.loading, state);
 
     case ProjectActionTypes.GET_ALL_SUCCESS:
       return set(GET_ALL_STATUS, EntityStatus.loadingSuccess,
--- a/components/automate-ui/src/app/entities/roles/role.reducer.ts
+++ b/components/automate-ui/src/app/entities/roles/role.reducer.ts
@@ -24,7 +24,7 @@ export function roleEntityReducer(state: RoleEntityState = RoleEntityInitialStat
 
   switch (action.type) {
     case RoleActionTypes.GET_ALL:
-      return set('getAllStatus', EntityStatus.loading, roleEntityAdapter.removeAll(state));
+      return set('getAllStatus', EntityStatus.loading, state);
 
     case RoleActionTypes.GET_ALL_SUCCESS:
       return set('getAllStatus', EntityStatus.loadingSuccess,
--- a/components/automate-ui/src/app/entities/api-tokens/api-token.reducer.ts
+++ b/components/automate-ui/src/app/entities/api-tokens/api-token.reducer.ts
@@ -45,13 +45,8 @@ export function apiTokenEntityReducer(
   switch (action.type) {
 
     case ApiTokenActionTypes.GET_ALL: {
-      return set(
-        STATUS,
-        EntityStatus.loading,
-        // clear token state to ensure we fetch the latest data
-        apiTokenEntityAdapter.removeAll(state)
-      ) as ApiTokenEntityState;
-    }
+      return set(STATUS, EntityStatus.loading, state);
+   }
 
     case ApiTokenActionTypes.GET_ALL_SUCCESS: {
       return set(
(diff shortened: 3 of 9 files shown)


components/automate-ui/src/app/entities/projects/project.reducer.ts
```diff
@@ -65,7 +65,7 @@ export function projectEntityReducer(
 
   switch (action.type) {
     case ProjectActionTypes.GET_ALL:
-      return set(GET_ALL_STATUS, EntityStatus.loading, projectEntityAdapter.removeAll(state));
+      return set(GET_ALL_STATUS, EntityStatus.loading, state);
```
msorens: As mentioned in the preamble, this one in particular causes the problem.
We start a `GET_ALL`, here we erase the list--and that manifests in the UI.
Sometime later--much later at slow network speed--the list gets restored via `GET_ALL_SUCCESS`.

We still need the `removeAll` for single `GET` calls I believe, but it does nothing for `GET_ALL` calls and, in fact, causes the problem addressed by this PR, so let's remove it.

SEAjamieD: If i'm understanding this correctly, what we're essentially doing now is _replacing_ the state instead of clearing it out completely and then writing in new state.... so on slow connections, the users will see the old state until it is replaced?

components/automate-ui/src/app/entities/roles/role.reducer.ts
```diff
@@ -24,7 +24,7 @@ export function roleEntityReducer(state: RoleEntityState = RoleEntityInitialStat
 
   switch (action.type) {
     case RoleActionTypes.GET_ALL:
-      return set('getAllStatus', EntityStatus.loading, roleEntityAdapter.removeAll(state));
+      return set('getAllStatus', EntityStatus.loading, state);
```
bcmdarroch: I tested this out to make sure none of the [data leakage issues](https://github.com/chef/automate/issues/2138) came back with the revert. 
I only saw leakage with roles. To repro, make this policy and log in as the test user:

```
$ echo '{
    "name": "testing",
    "id": "test1",
    "members": [
      "user:local:test"
    ],
    "statements": [
      {
        "effect": "ALLOW",
        "actions": ["iam:*:list"],
        "projects": [
          "*"
        ]
      }
    ],
    "projects": []
  }' >> pol.json

$ curl -kH "api-token: $TOK" -X POST -d @pol.json https://localhost/apis/iam/v2beta/policies | jq
```

If you navigate to roles, then click on one to go to its details page, you'll still be able to see all the role details despite the GET request returning 403

components/automate-ui/src/app/entities/projects/project.reducer.ts
msorens: Exactly; which is at it should be. In many cases, the data would not have changed so they will only see a spinner appear, then disappear, with the underlying table unchanged.

components/automate-ui/src/app/entities/roles/role.reducer.ts
msorens: That does not quite make sense. The only thing I did was not erase data in the middle of a ListRoles. At the completion of the ListRoles call everything is exactly as it was before. The GetRole call is unchanged. Can discuss further tomorrow. 🤔

bcmdarroch: Yeah looks like this is showing up on master too so must be a separate bug! I'll file it

PR #2911, opened 2020-02-19
studio: use vendor-state specific directory for GOBIN

stevendanna:
The binaries we build and put in /src/bin are typically dependent on
the state of our vendored modules. This sets up GOBIN to be based on
the hash of `vendor/modules.txt` to help ensure we don't see errors
caused by old binaries.

We don't clean up the old bin directories by default, but we do warn
devs as they enter the studio.

### DIFF
--- a/.studiorc
+++ b/.studiorc
@@ -53,10 +53,19 @@ export CGO_ENABLED=0
 export GO111MODULE=on
 
 # Specify where to put 'go installed' binaries
-export GOBIN=/src/bin
+setup_gobin() {
+    local vendor_hash
+    vendor_hash=$(sha256sum /src/vendor/modules.txt | cut -c1-16)
+    export GOBIN_BASE="/src/bin"
+    GOBIN="${GOBIN_BASE}/${vendor_hash}"
+    export GOBIN
+    mkdir -p "$GOBIN"
+}
+
+setup_gobin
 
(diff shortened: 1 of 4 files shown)


.studiorc
```diff
@@ -53,10 +53,19 @@ export CGO_ENABLED=0
 export GO111MODULE=on
 
 # Specify where to put 'go installed' binaries
-export GOBIN=/src/bin
+setup_gobin() {
+    local vendor_hash
+    vendor_hash=$(sha256sum /src/vendor/modules.txt | cut -c1-16)
```
stevendanna: `cut -c1-16`

Adjust based on how unlucky you think we are :D

.studiorc
```diff
@@ -53,10 +53,19 @@ export CGO_ENABLED=0
...
+    export GOBIN_BASE="/src/bin"
+    GOBIN="${GOBIN_BASE}/${vendor_hash}"
```
msorens: Am I correct that should I want to, say, update the grpc vendor lib next time and I want to test on my branch then test on master, the two subdirs of /src/bin will peacefully coexist?

stevendanna: Yes, but you might have to re-enter your studio or call setup_gobin when you change branches back and forth.

Issue #1044, opened 2019-07-23
Don't show _Dashboards_ when you don't have permission


susanev:
## Overview

Currently, we always show the Event Feed even if a user does not have permission. Instead, this page should be fully hidden from the user when they do not have permission to see it. If a user does not have permission to see the Event Feed page, they should not see the Dashboards top nav and they should be directed to the next available page (left-to-right).

![event-feed-permission-state.png](https://images.zenhubusercontent.com/5c58849f0b133e639c2b7211/684d0655-5ea7-4464-8a11-1d336a7d3d83)

 ## Design Details
### User has access to at least one page
* If a user does not have permission to see the Event Feed page they should be directed to the next available page (left-to-right).

### User has access to no pages
[Abstract designs](https://share.goabstract.com/4a1ce511-6d73-4b6b-9b64-78648b9143de?mode=design&sha=cd7079a0b77dd40e9416b6c1fc3da3ccba8cffd6)
* Hide the left-nav
* Hide the projects dropdown
* Change background color to `#fff`
* Display message (with both chef support and automate documentation being external links)...
```
It looks like you do not have permission to access any data in Automate.
Reach out to your administrator or contact Chef Support for help.

In the meantime, feel free to explore the Automate documentation.
```

![1044-no-permissions](https://user-images.githubusercontent.com/5489125/64186519-d12bda00-ce23-11e9-959b-37d64002ece7.png)

## References
* [Convert SettingsLandingComponent to reusable landing component](https://github.com/chef/a2/pull/5039)
* [Settings landing page for permission-based routing](https://github.com/chef/a2/pull/4642)

srenatus:
💭 Could we have a similar behaviour for when the backend can't be reached? Right now, we've got a pop up saying that the license cannot be verified, I think.

susanev:
content needs to be updated to:

```It looks like you do not have permission to access any data in Automate.
Reach out to your administrator or contact Chef Support for help.

In the meantime, feel free to explore the Automate documentation.```

as part of feedback from @mjingle

susanev:
@susanev find out if auth can more simply communicate if a user has access to no pages

susanev:
note: the screenshots are old, the top nav is now named _Dashboards_

srenatus:
✅ I'll accept this. Logging in as a user with no perms whatsoever, I see the blank page. When then making my user a member of the "Compliance Viewers" policy, I see the topnav containing only the Compliance tab.

msorens:
[...]

PR #2913, opened 2020-02-19
Streamline notifications (cosmetic)

msorens:

Noticed some redundancy while investigating a related issue.

### :+1: Definition of Done
Info and error notifications still work the same way (directly affected by this change).
License notifications still work the same way (not affected by this change)

Logged in as admin, create a new user, e.g.. "bob".
Upon successful creation you should see a "created user" message lasting for 8 seconds.
image

Logout and login as "bob".
Attempt to visit the policies page via URL (https://a2-dev.test/settings/policies). You should see an error lasting for 30 seconds:
image

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/components/notifications/notifications.component.html
+++ b/components/automate-ui/src/app/components/notifications/notifications.component.html
@@ -1,14 +1,7 @@
 <div class="notifications">
   <ng-container *ngFor="let notification of notifications">
-    <chef-notification *ngIf="notification.type === 'error'"
-      [type]="notification.type === 'license' ? 'error' : notification.type"
-      [timeout]="notification.timeout" (dismissed)="handleNotificationDismissal(notification.id)">
-      {{ notification.message }}
-    </chef-notification>
-  </ng-container>
-  <ng-container *ngFor="let notification of notifications">
-    <chef-notification *ngIf="notification.type === 'info'"
-      [type]="notification.type === 'license' ? 'error' : notification.type"
+    <chef-notification *ngIf="notification.type !== 'license'"
+      [type]="notification.type"
       [timeout]="notification.timeout" (dismissed)="handleNotificationDismissal(notification.id)">
       {{ notification.message }}
     </chef-notification>
--- a/components/automate-ui/src/app/entities/notifications/notification.model.ts
+++ b/components/automate-ui/src/app/entities/notifications/notification.model.ts
@@ -1,9 +1,6 @@
 export enum Type {
   error = 'error',
   info = 'info',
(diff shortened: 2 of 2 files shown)


components/automate-ui/src/app/components/notifications/notifications.component.html
```diff
@@ -1,14 +1,7 @@
...
-    </chef-notification>
-  </ng-container>
-  <ng-container *ngFor="let notification of notifications">
-    <chef-notification *ngIf="notification.type === 'info'"
-      [type]="notification.type === 'license' ? 'error' : notification.type"
+    <chef-notification *ngIf="notification.type !== 'license'"
+      [type]="notification.type"
       [timeout]="notification.timeout" (dismissed)="handleNotificationDismissal(notification.id)">
       {{ notification.message }}
     </chef-notification>
```
msorens: Not changing anything here; logic should be exactly equivalent. Here's why:
In the original code there's an `ngIf` looking for `error` and another looking for `info`. So neither of them will render for type `license`, so it is sufficient to check for `!== license`, delete the check for `type === license` (which would never have been true), and use a single block because the two original blocks are identical.

components/automate-ui/src/app/entities/notifications/notification.model.ts
```diff
@@ -1,9 +1,6 @@
 export enum Type {
   error = 'error',
   info = 'info',
-  /* TODO (tc) This is a stopgap because product decided it was
-     too risky to ship the full experience of a license banner.
-     This code will be removed first thing after ChefConf. */
```
msorens: No longer relevant; license type is now being used by the license notification banner (license-notifications.component.html

msorens:
@SEAjamieD Just a couple thoughts: 
(1) It is not necessarily a direct URL; could be clicking something in the app, and we cannot necessarily tell. So going back to "home" would not always be the correct answer.
(2) We discussed recently (in ui-team I think) that the entire top nav bar was going to be fluid/permission-based, so "event feed" would not always be "home". (But come to think of it, this point does not matter: once it is dynamic, it could route to the dynamic "home".)

susanev:
we really really need pages for 404 and 403.
but yah, hopefully, we can [stop showing the event feed if they dont have permission](https://github.com/chef/automate/issues/1044) sometime soon.

SEAjamieD:
@msorens - Good points, more to think about than what I was imagining and it will be great when nav is fully dynamic.

@susanev - it'll be awesome when we get to that!

PR #2919, opened 2020-02-19
Remove bintray from preflight checks

danielsdeleo:

habitat is now fetched from packages.chef.io in most cases

Testing for bintray now exposes us to false positives in the preflight checker.


danielsdeleo:
FYI, this code still refers to bintray for hab downloads, we probably need to keep it for now: https://github.com/chef/automate/blob/1fb807fd6f74c5cee09624c6cbe18c4ebed9376a/components/automate-deployment/pkg/airgap/bundle_creator.go#L69-L85

PR #2902, opened 2020-02-18
Add button for back to docs

SEAjamieD:
### :nut_and_bolt: Description: What code changed, and why?

The API docs are on an outside site and getting back to docs was a pretty bad experience. I added a "return to docs" button on the API page. 

There are limitations to styling the API pages. Even so, this button could be prettier. 

Screen Shot 2020-02-18 at 1 04 53 PM

### :chains: Related Resources
https://github.com/chef/automate/pull/2890

### :athletic_shoe: How to Build and Test the Change

Checkout this branch
Build the API docs by running the following all-in-one command from the top-level automate directory:

```
hab studio run "source .studiorc && compile_all_protobuf_components" && pushd components/automate-chef-io/ && make sync_swagger_files generate_swagger && make serve || popd
```
Review the API docs in your browser:

http://localhost:1313/docs/api

### :white_check_mark: Checklist

- [x] I have read the [CONTRIBUTING document](https://github.com/chef/automate/blob/master/CONTRIBUTING.md).
- [ ] Tests added/updated?
- [ ] Docs added/updated?
- [x] All commits have been signed-off for the [Developer Certification of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

Issue #2855, opened 2020-02-12
Automate API docs should not rely on a CDN and track latest version of deps


phiggins:
Followup to #2835. The Automate docs rely on a library called ReDoc and currently load the latest released version from a CDN. Reducing the dependence on this would have a few benefits:

* Each release of Automate could stay with the version of ReDoc it was tested with which would help keep the docs working in the face of ReDoc updates breaking backwards compatibility.
* Downtime or slowness of the CDN we currently load ReDoc from would not affect customers of Automate.
* ReDoc appears to be able to build a self-contained version of the documentation which would help environments with limited internet connectivity and would push some failure states into package building time rather than run time: https://github.com/Redocly/redoc/blob/master/cli/README.md

PR #2921, opened 2020-02-19
Fetch ReDoc js file when building hab package rather than rely on CDN.

phiggins:
Fix for #2855.

I had originally planned to use [redoc-cli](https://github.com/Redocly/redoc/blob/master/cli/README.md) to package the docs, but that had a few downsides:
* The combined HTML + js file didn't work with hugo the way we're doing things currently, and I don't understand hugo enough to rework things to get this to work.
* This would have added npm to the docs development process and package building, which seems a bit onerous.
* While trying to get this going there was work done that would have conflicted and needed to be redone (#2902).
* Some other smaller concerns that can basically be summed up as redoc-cli doesn't do much to justify depending on it.

This still relies on the CDN, but would move potential failures to package build time rather than when a customer views the docs, which seems like a much better trade off.

### DIFF
--- a/components/automate-chef-io/habitat/plan.sh
+++ b/components/automate-chef-io/habitat/plan.sh
@@ -6,6 +6,7 @@ pkg_license=('Chef-MLSA')
 pkg_description="HTML for automate.chef.io"
 pkg_upstream_url="https://www.chef.io/automate"
 pkg_build_deps=(
+  core/curl
   core/git
   core/hugo
   core/jq-static
--- a/components/automate-chef-io/.gitignore
+++ b/components/automate-chef-io/.gitignore
@@ -8,3 +8,4 @@ package.json
 public/
 themes/
 static/api-docs
+static/redoc.standalone.js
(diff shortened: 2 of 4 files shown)


components/automate-chef-io/habitat/plan.sh
```diff
@@ -6,6 +6,7 @@ pkg_license=('Chef-MLSA')
 pkg_description="HTML for automate.chef.io"
 pkg_upstream_url="https://www.chef.io/automate"
 pkg_build_deps=(
+  core/curl
```
phiggins: This was already pulled in as a transitive dep so this wasn't necessary, but it seemed better to be explicit.

phiggins:
> What impact does this have on the loading time for the api docs, if any?

I haven't measured, but it should be slightly faster not having to contact the CDN every time the docs loads.

phiggins:
I just tried loading the API docs at chef.automate.io with the firefox inspector open and downloading redoc took 12ms, so this probably won't have much effect.

Screen Shot 2020-02-19 at 2 23 46 PM

kagarmoe:
If it isn't slower, then it must be faster :-D

phiggins:
This means one fewer network request, so it definitely should not be slower.

PR #2927, opened 2020-02-20
Fix client-runs export breaking on page barrier

phiggins:
Fix for https://github.com/chef/customer-bugs/issues/70.

The problem with the exported JSON was that each page of results from elasticsearch was being streamed as a fully formed document. The go stdlib's json streaming parser is apparently lenient enough to interpret this as one big collection of objects, but the document parser is not. the solution ended up being changing the existing integration test to be more strict in the output allowed, then adding `[`, `]`, and `,` in the appropriate places to create one big document.

### DIFF
--- a/components/config-mgmt-service/grpcserver/node_export.go
+++ b/components/config-mgmt-service/grpcserver/node_export.go
@@ -157,12 +157,15 @@ func (s *CfgMgmtServer) nodePager(ctx context.Context, request *pRequest.NodeExp
 func jsonExport(stream service.CfgMgmt_NodeExportServer) exportHandler {
 	initialRun := true
 	return func(nodes []backend.Node) error {
+		// If this is the first set of nodes to export, prepend "[" to start a JSON document.
 		if initialRun {
 			err := stream.Send(&response.ExportData{Content: []byte("[")})
 			if err != nil {
 				return err
 			}
 			initialRun = false
+			// If this is not the first set of nodes and the collection has elements,
+			// prepend the "," that couldn't get added in the previous call to this function.
 		} else if len(nodes) != 0 {
 			err := stream.Send(&response.ExportData{Content: []byte(",")})
 			if err != nil {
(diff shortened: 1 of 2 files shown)


components/config-mgmt-service/integration_test/node_export_test.go
```diff
@@ -758,15 +759,86 @@ func TestNodeExportPaging(t *testing.T) {
...
-		var nodes []interface{}
-		err := dec.Decode(&nodes)
-		require.NoError(t, err)
+	var outputNodes []interface{}
+	err = json.Unmarshal(data, &outputNodes)
+
+	require.NoError(t, err)
+
+	assert.Equal(t, numberOfNodes, len(outputNodes))
+}
```
phiggins: This change switches from the lenient json stream parser to the more strict json document parser. The original code before this PR passed with the streaming parser but not with the document parser.

components/config-mgmt-service/integration_test/node_export_test.go
```diff
@@ -758,15 +759,86 @@ func TestNodeExportPaging(t *testing.T) {
...
+		data = append(data, tdata.GetContent()...)
+	}
+
+	reader := csv.NewReader(bytes.NewReader(data))
+	outputNodes, err := reader.ReadAll()
+
+	require.NoError(t, err)
+	// Add 1 for the CSV header
+	assert.Equal(t, numberOfNodes+1, len(outputNodes))
 }
```
phiggins: This test is 90% the same as the JSON test. Adding this test wasn't strictly necessary, but there weren't any tests for paginated CSV exports so it made sense to add one while I was messing with the pagination code.

components/config-mgmt-service/grpcserver/node_export.go
```diff
@@ -109,64 +94,119 @@ func (s *CfgMgmtServer) exportNodes(ctx context.Context, request *pRequest.NodeE
...
 
+		if len(nodes) == 0 {
+			break
+		}
+	}
+
+	return nil
+}
+
+func (s *CfgMgmtServer) nodePager(ctx context.Context, sortField string, sortAsc bool, nodeFilters map[string][]string) func() ([]backend.Node, error) {
```
lancewf: This is really slick. Good job.

vjeffrey:
rebased to get past CI error

PR #2934, opened 2020-02-20
Auth 2926/teams v2 migrations

blakestier:
# :+1: Definition of Done 

System builds, is on iam v2.1, and has the following teams: `admins`, `editors`, `viewers` and *not* `operators`. 

Note: for existing deployments, the team names will have different capitalization. We've decided not to address that inconsistency at the moment.

Note: These instructions rebuild all the things, but you should be able to just rebuild teams and see the change.

For a fresh install:

from this branch, enter the studio

build: `build components/automate-deployment && build components/automate-cli && build components/authz-service  && build components/teams-service && build components/local-user-service && build components/automate-gateway`
`start_all_services`

when everything is up,
`chef-automate dev psql chef_teams_service`
`select * from schema_migrations;` (should be on 10)
`select * from teams;` (should have admins, editors, viewers)

For an existing v1 install:

from this branch, enter the studio

`start_all_services`
rebuild: `rebuild components/automate-deployment && rebuild components/automate-cli && rebuild components/authz-service  && rebuild components/teams-service && rebuild components/local-user-service && rebuild components/automate-gateway`

when everything is up,
`chef-automate dev psql chef_teams_service`
`select * from schema_migrations;` (should be on 10)
`select * from teams;` (should have admins, editors, viewers)

For an existing v2 install:

from master, enter the studio
`start_all_services`

switch to this branch
rebuild: `automate-deployment, automate-cli, authz-service, teams-service, local-user-service, automate-gateway`

when everything is up,
`chef-automate dev psql chef_teams_service`
`select * from schema_migrations;` (should be on 10)
`select * from teams;` (should have admins, editors, viewers)

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/teams-service/storage/postgres/migration/sql/10_upgrade_to_v2_migrations.up.sql
```diff
@@ -0,0 +1,23 @@
+BEGIN;
```
bcmdarroch: what's the difference between these two?

components/teams-service/storage/postgres/migration/sql/10_upgrade_to_v2_migrations.sql
```diff
@@ -0,0 +1,23 @@
+BEGIN;
+
+UPDATE teams
+    SET
+        name = 'editors',
+        description = 'Editors'
+    WHERE
+        name = 'operators'
```
bcmdarroch: should we also set the team's projects to empty too, like in the old 2nd data migration?

blakestier: I don't think we need to do it twice-- I couldn't think of a scenario in which they could have had their projects dropped, then added more, because I think we enforced permissions in the meantime. 

But if you can think of one, we can do it!

components/teams-service/storage/postgres/migration/sql/10_upgrade_to_v2_migrations.up.sql
blakestier: ugh good catch. This one should have just been a rename of the other (added the `up`) but somehow both were preserved.

PR #2938, opened 2020-02-21
Add some logging around forced migration

msorens:

Additional logging could prove useful in the event of
(a) customer issues once the force upgrade is released, and 
(b) helping us to make sure it is working as expected before release

PR #2793 Forced migration upgrade

### :+1: Definition of Done

Supervisor log shows something like the output below.

#### REVISED PER REVIEW FEEDBACK

Got rid of the "bracketing" notion; not enough value added.

```
"applying database migrations from /hab/pkgs/msorens/authz-service/0.1.0/20200224202924/migrations"
"Initializing DB migrations..."
"Current schema version: 73"
"Migrating schema from version 73 to 74"
"Setting up IAM data basics..."
"Checking for data migrations..."
"Checking for remaining schema migrations..."
"DB initialization complete at version 74"
```

#### OUTDATED
Notice that the logging is instrumented as brackets. The "open bracket" has the form `x...` and the "close bracket" has the form `x -- complete`. That covers the eventuality of failures in the middle.
```
"applying database migrations from /hab/pkgs/msorens/authz-service/0.1.0/20200221033251/migrations"
"Running db migrations..."
"Current schema version: 73"
"Migrating schema from version 73 to 74"
"Migrating schema from version 73 to 74 -- complete"
"Migrating from IAM v1 to v2..."
"Migrating from IAM v1 to v2 -- complete"
"Migrating data if needed"
"Migrating data if needed -- complete"
"Migrating schema post-version 74..."
"Migrating schema post-version 74 -- no changes necessary"
"Running db migrations -- complete"
```

### :athletic_shoe: How to Build and Test the Change (maybe?)

```
rebuild components/authz-service && sl | grep authz-service
```

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/authz-service/storage/postgres/migration/migration.go
+++ b/components/authz-service/storage/postgres/migration/migration.go
@@ -36,7 +36,7 @@ func (c *Config) Migrate(dataMigConf datamigration.Config) error {
 	migrationsTable := ""
 	ctx := context.Background()
 
-	l.Infof("Running db migrations from %q", migrationsPath)
+	l.Info("Running db migrations...")
 	purl, err := addMigrationsTable(pgURL, migrationsTable)
 	if err != nil {
 		return errors.Wrap(err, "parse PG URL")
@@ -63,7 +63,9 @@ func (c *Config) Migrate(dataMigConf datamigration.Config) error {
 	}
 
 	if version < PreForceUpgradeMigration {
+		l.Infof("Migrating schema from version %v to %v...", version, PreForceUpgradeMigration)
 		err = m.Migrate(PreForceUpgradeMigration)
+		l.Infof("Migrating schema from version %v to %v -- complete", version, PreForceUpgradeMigration)
 		if err != nil && err != migrate.ErrNoChange {
 			return errors.Wrap(err, "migration up to IAM-V2-force-upgrade failed")
 		}
@@ -90,6 +92,7 @@ func (c *Config) Migrate(dataMigConf datamigration.Config) error {
 	if isOnV1 {
 		if isDirty { // we've attempted to migrate and not finished
 			// get IAM db in original, clean state to avoid conflicts
+			l.Infof("Detected unfinished migration; resetting IAM DB to original, clean state")
 			err = legacy.ResetIAMDb(ctx, db)
 			if err != nil {
 				return errors.Wrap(err, "reset IAM V2 database")
@@ -106,7 +109,9 @@ func (c *Config) Migrate(dataMigConf datamigration.Config) error {
 		if err != nil {
 			return errors.Wrapf(err, "failed to set IAM v2 migration_status to %s", constants_v2.EnumInProgress)
 		}
+		l.Info("Migrating from IAM v1 to v2...")
 		err = legacy.MigrateToV2(ctx, db, migrateV1Policies)
+		l.Info("Migrating from IAM v1 to v2 -- complete")
 		if err != nil {
 			statusErr := legacy.RecordMigrationStatus(ctx, constants_v2.EnumFailed, db)
 			if statusErr != nil {
@@ -121,18 +126,26 @@ func (c *Config) Migrate(dataMigConf datamigration.Config) error {
 	}
 
 	// idempotent
+	l.Info("Migrating data if needed...")
 	err = dataMigConf.Migrate()
+	l.Info("Migrating data if needed -- complete")
 	if err != nil {
 		return errors.Wrap(err, "IAM data migrations failed")
 	}
 
 	// perform remaining migrations
+	l.Infof("Migrating schema post-version %v...", PreForceUpgradeMigration)
 	err = m.Up()
 	if err != nil && err != migrate.ErrNoChange {
 		return errors.Wrap(err, "migrations failed")
 	}
+	resolution := "complete"
+	if err == migrate.ErrNoChange {
+		resolution = "no changes necessary"
+	}
+	l.Infof("Migrating schema post-version %v -- %s", PreForceUpgradeMigration, resolution)
 
-	l.Infof("Completed db migrations")
+	l.Info("Running db migrations -- complete")
 
 	err = db.Close()
 	if err != nil {
--- a/components/authz-service/storage/postgres/migration/migration.go
+++ b/components/authz-service/storage/postgres/migration/migration.go
@@ -36,7 +36,7 @@ func (c *Config) Migrate(dataMigConf datamigration.Config) error {
 	migrationsTable := ""
 	ctx := context.Background()
 
-	l.Info("Running db migrations...")
+	l.Info("Initializing DB migrations...")
 	purl, err := addMigrationsTable(pgURL, migrationsTable)
 	if err != nil {
 		return errors.Wrap(err, "parse PG URL")
@@ -63,11 +63,10 @@ func (c *Config) Migrate(dataMigConf datamigration.Config) error {
 	}
 
 	if version < PreForceUpgradeMigration {
-		l.Infof("Migrating schema from version %v to %v...", version, PreForceUpgradeMigration)
+		l.Infof("Migrating schema version %v to %v...", version, PreForceUpgradeMigration)
 		err = m.Migrate(PreForceUpgradeMigration)
-		l.Infof("Migrating schema from version %v to %v -- complete", version, PreForceUpgradeMigration)
 		if err != nil && err != migrate.ErrNoChange {
-			return errors.Wrap(err, "migration up to IAM-V2-force-upgrade failed")
+			return errors.Wrapf(err, "migration up to version %v failed", PreForceUpgradeMigration)
 		}
 	}
 
@@ -92,13 +91,13 @@ func (c *Config) Migrate(dataMigConf datamigration.Config) error {
 	if isOnV1 {
 		if isDirty { // we've attempted to migrate and not finished
 			// get IAM db in original, clean state to avoid conflicts
-			l.Infof("Detected unfinished migration; resetting IAM DB to original, clean state")
+			l.Infof("Detected unfinished migration; resetting IAM DB to clean state")
 			err = legacy.ResetIAMDb(ctx, db)
 			if err != nil {
-				return errors.Wrap(err, "reset IAM V2 database")
+				return errors.Wrap(err, "reset database")
 			}
 			if err := dataMigConf.Reset(); err != nil {
-				return errors.Wrap(err, "reset v2 data migrations")
+				return errors.Wrap(err, "reset data migrations")
 			}
 			if version != 0 {
 				migrateV1Policies = true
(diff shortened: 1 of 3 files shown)


components/authz-service/storage/postgres/migration/migration.go
```diff
@@ -36,7 +36,7 @@ func (c *Config) Migrate(dataMigConf datamigration.Config) error {
 	migrationsTable := ""
 	ctx := context.Background()
 
-	l.Infof("Running db migrations from %q", migrationsPath)
+	l.Info("Running db migrations...")
```
msorens: This migration path was already output in the log from the caller so don't need to duplicate it here. That lets me use my notion of making this an "open bracket" as explained above. The "close bracket" is down at the end of the function.

components/authz-service/storage/postgres/migration/migration.go
```diff
@@ -121,18 +126,26 @@ func (c *Config) Migrate(dataMigConf datamigration.Config) error {
...
+	l.Infof("Migrating schema post-version %v...", PreForceUpgradeMigration)
 	err = m.Up()
 	if err != nil && err != migrate.ErrNoChange {
 		return errors.Wrap(err, "migrations failed")
 	}
+	resolution := "complete"
+	if err == migrate.ErrNoChange {
+		resolution = "no changes necessary"
+	}
+	l.Infof("Migrating schema post-version %v -- %s", PreForceUpgradeMigration, resolution)
```
msorens: Notice the `resolution` on the end; it is either "complete" or "no changes necessary", which I think is very useful to share.
Could do the same thing for data migrations (L129-L131) above, but that requires changing what `dataMigConf.Migrate()` returns. Is that worth doing in your opinions?

components/authz-service/storage/v1/postgres/postgres.go
```diff
@@ -67,7 +67,7 @@ func (m *policyMap) Scan(src interface{}) error {
 
 // New instantiates the postgres IAM v1 storage backend.
 func New(ctx context.Context, l logger.Logger, migConf migration.Config, dataMigConf datamigration.Config) (storage.Storage, error) {
-	l.Infof("applying database migrations from %s", migConf.Path)
+	l.Infof("[v1] applying database migrations from %s", migConf.Path)
```
msorens: Adding this vital distinction makes it obvious where the process starts. The v2 one does not have the "[v2]" tag so you can distinguish them.

components/authz-service/storage/postgres/migration/migration.go
bcmdarroch: Will the `resolution` change once we add new migrations?

What would have to change about `dataMigConf.Migrate()`? Not sure I feel strongly either way. What does `complete` vs. `no changes necessary` tell us?

components/authz-service/storage/postgres/migration/migration.go
blakestier: While we're here, maybe "Initializing migrations" makes more sense. 🤷‍♂

components/authz-service/storage/postgres/migration/migration.go
```diff
@@ -63,7 +63,9 @@ func (c *Config) Migrate(dataMigConf datamigration.Config) error {
 	}
 
 	if version < PreForceUpgradeMigration {
+		l.Infof("Migrating schema from version %v to %v...", version, PreForceUpgradeMigration)
 		err = m.Migrate(PreForceUpgradeMigration)
+		l.Infof("Migrating schema from version %v to %v -- complete", version, PreForceUpgradeMigration)
```
blakestier: I think we'll want this after the err check

components/authz-service/storage/postgres/migration/migration.go
```diff
@@ -106,7 +109,9 @@ func (c *Config) Migrate(dataMigConf datamigration.Config) error {
 		if err != nil {
 			return errors.Wrapf(err, "failed to set IAM v2 migration_status to %s", constants_v2.EnumInProgress)
 		}
+		l.Info("Migrating from IAM v1 to v2...")
```
blakestier: Is it weird to draw all this attention to v1 -> v2? For new users, there has never been a v1. I considered renaming this function and it fell through the cracks, but we're only "migrating v1 -> v2" within the `MigrateToV2` function _if_ it's not a fresh install. Otherwise, we're really, er, "Creating base IAM roles and policies" . . . ? If you'd like to do a rename and push this logging into that function, I suppose that is fine. Otherwise, I'd modify this message to be something more like "Setting up IAM basics" or something.

components/authz-service/storage/postgres/migration/migration.go
```diff
@@ -121,18 +126,26 @@ func (c *Config) Migrate(dataMigConf datamigration.Config) error {
 	}
 
 	// idempotent
+	l.Info("Migrating data if needed...")
```
blakestier: This is also a little confusing, but it helps if we were debugging. More and more I'm wondering if these should be set to debug by default.

components/authz-service/storage/v1/postgres/postgres.go
blakestier: Do we need this change if v1 is going away? Or is it here for the interim to clarify things?

components/authz-service/storage/postgres/migration/migration.go
blakestier: I don't think we need the addition `resolution` in datamigrations. We can always add it if we need to.

blakestier:
I think the logging is helpful for debugging but I'm not confident we need the granularity generally, at least for some of the smaller steps (this does point out something of a smell in our naming/function crafting, but that's a big constrained by how we've built the system). What do you think? Would `debug` be sufficient?

components/authz-service/storage/v1/postgres/postgres.go
msorens: The latter--that is what clued me in to the fact that the v1 server was running migrations.

components/authz-service/storage/postgres/migration/migration.go
msorens: But his will only get run if someone is on v1, right? Or are you saying that out-of-the-box a fresh install will still start in v1, and will run this bit of code?

components/authz-service/storage/postgres/migration/migration.go
msorens: I think it is quite important to have them on by default, because if a customer runs into a problem we will at least have a clue what is going on. Asking them to re-run after setting to debug is not an adequate substitute.

components/authz-service/storage/postgres/migration/migration.go
msorens: @bcmdarroch 
(1) Yes, as we add new migrations and the customer restarts authz, they will either see "complete" or "no changes necessary" as appropriate. 
(2) For dataMigConf, it would need to also return the `migrate.ErrNoChange` error so we could do the same check.

components/authz-service/storage/postgres/migration/migration.go
msorens: Hmm... tried that and it makes it more complicated. Have to have one log statement for completed successfully, then another for completed with error since the err-check does a return. Then same changes on other blocks of code, where some have multiple error checks... So I am inclined to leave it as is.

components/authz-service/storage/postgres/migration/migration.go
msorens: I agree the "if needed" is confusing--that is what could be alleviated by the change to `dataMigConf.Migrate()` I spoke of.

msorens:
Additional note for posterity:

Here is an actual run with a v1 system, showing that the v1 server runs the migration first, then the v2 server runs them again. (The v1 server will go away soon.)
```
"[v1] applying database migrations from /hab/pkgs/msorens/authz-service/0.1.0/20200221234646/migrations"
"Running db migrations..."
"Current schema version: 73"
"Migrating schema from version 73 to 74..."
"Migrating schema from version 73 to 74 -- complete"
"Migrating from IAM v1 to v2..."
"Migrating from IAM v1 to v2 -- complete"
"Migrating data if needed..."
"1/u v2_data_migrations (25.7838ms)\n"
"2/u v2_rename_all_actions (41.2622ms)\n"
"3/u project_admin (62.6571ms)\n"
"4/u v2_set_projects_default (88.6839ms)\n"
"5/u remove_default_project (107.0955ms)\n"
"6/u v2_roles (124.04ms)\n"
"7/u update_v2_roles (142.1491ms)\n"
"8/u update_v2_projects (162.7142ms)\n"
"9/u update_project_admin (182.1245ms)\n"
"10/u add_owner_role_to_admin_pol (202.1975ms)\n"
"11/u remove_iam-members-viewer_role (224.8039ms)\n"
"Migrating data if needed -- complete"
"Migrating schema post-version 74..."
"Migrating schema post-version 74 -- no changes necessary"
"Running db migrations -- complete"
"initializing OPA store with 29 V1 policies"


"applying database migrations from /hab/pkgs/msorens/authz-service/0.1.0/20200221234646/migrations"
"Running db migrations..."
"Current schema version: 74"
"Migrating data if needed..."
"Migrating data if needed -- complete"
"Migrating schema post-version 74..."
"Migrating schema post-version 74 -- no changes necessary"
"Running db migrations -- complete"
```

components/authz-service/storage/postgres/migration/migration.go
bcmdarroch: Hmm, still thinking about the difference between `complete` vs. `no changes necessary`. By `complete` you mean to say that all migrations that should have been applied were actually applied, right? That might be better communicated by `latest migrations applied`.

msorens: I really want the verb signifying, well, completion, to be identical for all items, so I really want it to be "complete". If you want to tweak the wording on the left side, we could do that, but that wording needs to be the same for the opening and closing "brackets".

components/authz-service/storage/postgres/migration/migration.go
```diff
@@ -121,18 +126,26 @@ func (c *Config) Migrate(dataMigConf datamigration.Config) error {
...
 
 	// perform remaining migrations
+	l.Infof("Migrating schema post-version %v...", PreForceUpgradeMigration)
 	err = m.Up()
 	if err != nil && err != migrate.ErrNoChange {
 		return errors.Wrap(err, "migrations failed")
 	}
+	resolution := "complete"
+	if err == migrate.ErrNoChange {
+		resolution = "no changes necessary"
```
susanev:
```suggestion
		resolution = "no migration necessary"
```

components/authz-service/storage/postgres/migration/migration.go
susanev: what if we replaced `changes` with `migrations`?

susanev: also maybe instead of `complete` we use `migration complete`

components/authz-service/storage/postgres/migration/migration.go
blakestier: Hmm, I'm happy to have a go at it myself since I don't totally understand the scope of the problem you're mentioning. It looks to me like you could add a log statement before the err return along the lines of "migration to 74 failed" and then the "complete" log after that. 

But at a minimum, I think the phrase "complete" is misleading here, as it may have failed, so if we can't move it, I'd opt to delete it.

components/authz-service/storage/postgres/migration/migration.go
blakestier: Yes, the system understands "isOnV1" as "hasNotUpgradedToV2", including fresh installs.

blakestier: You could use the bool `migrateV1Policies` as an indicator for when we'll be doing *actual* v1-v2 migration work (the v1 policies)

components/authz-service/storage/postgres/migration/migration.go
blakestier: what about "migrating any unmigrated data" or something (then it's always true)

components/authz-service/storage/postgres/migration/migration.go
```diff
@@ -63,9 +63,10 @@ func (c *Config) Migrate(dataMigConf datamigration.Config) error {
 	}
 
 	if version < PreForceUpgradeMigration {
+		l.Infof("Migrating schema version %v to %v...", version, PreForceUpgradeMigration)
 		err = m.Migrate(PreForceUpgradeMigration)
 		if err != nil && err != migrate.ErrNoChange {
-			return errors.Wrap(err, "migration up to IAM-V2-force-upgrade failed")
+			return errors.Wrapf(err, "migration up to version %v failed", PreForceUpgradeMigration)
```
msorens: Review sentiment seemed to favor removing v1 and v2 references, so updated a few error messages in that regard.

components/authz-service/storage/postgres/migration/migration.go
```diff
@@ -106,33 +108,41 @@ func (c *Config) Migrate(dataMigConf datamigration.Config) error {
...
 		return errors.Wrap(err, "migrations failed")
 	}
 
-	l.Infof("Completed db migrations")
+	version, _, err = m.Version()
+	if err != nil {
+		return errors.Wrap(err, "failed to get version post-migration")
+	}
+
+	l.Infof("DB initialization complete at version %v", version)
```
msorens: Thought it would be a bonus to actually show the final migration level in the log, too.

components/authz-service/storage/postgres/migration/migration.go
```diff
@@ -106,33 +108,41 @@ func (c *Config) Migrate(dataMigConf datamigration.Config) error {
...
 		}
 		err = legacy.RecordMigrationStatus(ctx, constants_v2.EnumSuccessful, db)
 		if err != nil {
-			return errors.Wrapf(err, "failed to set IAM v2 migration_status to %s", constants_v2.EnumSuccessful)
+			return errors.Wrapf(err, "failed to set IAM migration_status to %s", constants_v2.EnumSuccessful)
 		}
 	}
 
 	// idempotent
+	l.Info("Checking for data migrations...")
```
blakestier: 😎

components/authz-service/storage/postgres/migration/migration.go
```diff
@@ -106,33 +108,41 @@ func (c *Config) Migrate(dataMigConf datamigration.Config) error {
...
+		l.Info("Setting up IAM data basics...")
 		err = legacy.MigrateToV2(ctx, db, migrateV1Policies)
 		if err != nil {
 			statusErr := legacy.RecordMigrationStatus(ctx, constants_v2.EnumFailed, db)
 			if statusErr != nil {
-				return errors.Wrapf(statusErr, "failed to set IAM v2 migration_status to %s:%s", constants_v2.EnumFailed, err.Error())
+				return errors.Wrapf(statusErr, "failed to set IAM migration_status to %s:%s", constants_v2.EnumFailed, err.Error())
 			}
-			return errors.Wrap(err, "IAM v2 force-upgrade failed")
+			return errors.Wrap(err, "IAM IAM data basics failed")
```
susanev: should it say IAM twice?

msorens: Oops. Thanks!

Issue #2568, opened 2019-12-20
Backup deletion fails because of slow ES snapshot delete


stevendanna:
# Describe the bug

We've now had at least 2 customers report that deletion of old backups failed to complete.  In both cases, we hit the default 2 minute timeout. (For customer-specific information see Zendesk tickets 23999 and 24030):

```
deployment-service.default(O): time="2019-12-13T05:51:16Z"
level=info msg="Running backup operation" action=delete
backup_id=20191211040000 operation=elasticsearch_snapshot
service_name=compliance-service

deployment-service.default(O): time="2019-12-13T05:53:15Z"
level=error msg="Failed to delete backup" error="failed to delete
backup 20191211040000: failed to delete elasticsearch backup
20191211040000 for compliance-service: rpc error: code = Canceled
desc = context canceled"
```

Attempts to re-run the deletion then hit the following error:

```
deployment-service.default(O): time="2019-12-13T05:54:29Z"
level=error msg="finished unary call with code Internal"
error="rpc error: code = Internal desc = failed to delete backup
20191211040000: failed to delete elasticsearch backup
20191211040000 for event-feed-service: rpc error: code = Unknown
desc = failure deleting snapshot \"20191211040000\" from repo
\"chef-automate-es6-event-feed-service\": elastic: Error
503 (Service Unavailable):
[chef-automate-es6-event-feed-service:20191211040000/VhRl6yfPSHqCQSIjj4iuhw]
cannot delete - another snapshot is currently being deleted
[type=concurrent_snapshot_execution_exception]"
grpc.code=Internal
```

It is not yet clear whether simply increasing the timeout via the `-t` flag would have resulted in a successful deletion.  So far, simply waiting some unspecified amount of time and then retrying allowed the backup to finish deleting.

Unknowns:

- [ ] The snapshot delete API appears asynchronous. I don't see code in es-sidecar-service to check on the status of the deletion.  Thus, what is timing out?
- [ ] Their doesn't appears to be an API to get the status of a snapshot delete.  Various internet search results claim we should be able to see it in the `_cat/tasks` output, but support reports they didn't see it.  However, it could be that we just missed the relevant output.
- [ ] If snapshot deleting is async, and if 2 can't be run at the same time, why don't we hit this all the time?  Are snapshot deletes just almost always fast enough?

PR #2948, opened 2020-02-24
Make ES snapshot deletes retry on 5xx 

jaym:
ES snapshot delete are async. You can only have one inflight
at a time, otherwise you'll get an error:
```
{
  "error" : {
    "root_cause" : [
      {
        "type" : "concurrent_snapshot_execution_exception",
        "reason" : "[snapshotXX] another snapshot is currently running cannot delete"
      }
    ],
    "type" : "concurrent_snapshot_execution_exception",
    "reason" : "[snapshotXX] another snapshot is currently running cannot delete"
  },
  "Status. 503.
}
```

This change keeps retrying until the delete is acknowledged. Hopefully, some of the pain
from #2568 will be eased.

### DIFF
--- a/components/es-sidecar-service/pkg/elastic/snapshot.go
+++ b/components/es-sidecar-service/pkg/elastic/snapshot.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"net/url"
 	"path"
 	"regexp"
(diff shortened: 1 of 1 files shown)


jaym:
i made an assumption. now i'm convinced its wrong

jaym:
looking at curator, it doesn't check if the snapshot is deleted. It just tries to delete the next one, and retries if that fails. Some discussion in https://discuss.elastic.co/t/curator-delete-snapshot-always-timeout/147637/3

PR #2954, opened 2020-02-24
[infra-proxy-service] API to get cookbook affected nodes

vsingh-msys:

In order to fetch nodes that are affected by the cookbook. Hit the chef server partial search API with required params.

```
params: name, chef_guid, url, check_in, uptime, platform, environment, policy_group
```

[MSYS-1241](https://chefio.atlassian.net/browse/MSYS-1241)
### :+1: Definition of Done
Able to fetch the nodes with desired params.

- rebuild `components/infra-proxy-service`
- rebuild `components/automate-gateway`
```
chef-automate dev grpcurl infra-proxy-service chef.automate.domain.infra_proxy.service.InfraProxy.GetCookbookAffectedNodes -- -d '{"server_id": "SERVER_ID", "org_id": "ORG_ID", "name": "COOKBOOK_NAME", "version":"COOKBOOK_VERSIOB"}'
```

- [ ] Tests added/updated?
- [ ] Docs added/updated?


vsingh-msys:
gRPC API:

```
chef-automate dev grpcurl infra-proxy-service chef.automate.domain.infra_proxy.service.InfraProxy.GetCookbookAffectedNodes -- -d '{"org_id": "8f1646d3-35ae-4562-84d7-ff013f3973f2", "server_id": "c7d36e02-4816-4bfa-a1b9-d239f56c9b8d", "name": "starter", "version": "1.0.13"}'
```
Response:
```
{
  "nodes": [
    {
      "name": "node1.test",
      "chefGuid": "d9af01f5-342b-4329-8986-2647b6383a63",
      "uptime": "10 minutes 55 seconds",
      "platform": "ubuntu",
      "environment": "_default"
    }
  ]
}

```

components/infra-proxy-service/server/v1/cookbooks.go
```diff
@@ -182,3 +219,57 @@ func fromAPIToListAvailableCookbooks(al chef.CookbookListResult) []*response.Coo
...
+}
+
+func getNodeAttributeFromRes(data interface{}) NodeAttribute {
+	m := data.(map[string]interface{})["data"].(map[string]interface{})
+	node := NodeAttribute{}
+	if name, ok := m["name"].(string); ok {
+		node.Name = name
+	}
+
+	if checkin, ok := m["ohai_time"].(string); ok {
```
vsingh-msys: At present `ohai_time` is not being populated in response. Looking for help?

Also, any code optimization are super welcome :)

kmacgugan:
Is the intent to get all nodes within an ecosystem that could be affected by a cookbook? Theoretically there could be nodes not in chef-server or nodes not in automate that are affected by a cookbook. I'm not sure where this is intended to be displayed or what the context is that this API supports. If we want to show nodes in automate that ran with a cookbook version we can query automate directly without going to chef-server.

vsingh-msys:
@kmacgugan here the intent is to get all nodes which stored on chef-server and affected by the cookbook. I am trying to wire these nodes with Automate's stored nodes possibly with `chef_guid`. Ref: http://bit.ly/2vpIz5N

components/infra-proxy-service/server/v1/cookbooks.go
vjeffrey: we discussed this a bit in slack - superteam channel:
```
lancewf: 
The "ohai_time" is stored in the node-attribute index.
We don't place it in the run, but we could without much work.
You can see the ohai_time on the "Attributes" tab on the node runs view if you search for it.  

dan 
I'm not sure we need it if we have some other timestamp.
```

vjeffrey: could you speak to why you were looking to use ohai_time instead of just the check in time @vsingh-msys ?

lancewf: Sorry, I forgot he was pulling this directly from Chef Server. The "ohai_time" is not at the root of the data element it is at "data" -> "automatic" -> "ohai_time". Here is an example. 
https://github.com/chef/automate/blob/82772a8c0a82da983bb9b329fb31716447d66609/components/ingest-service/examples/converge-success-report.json#L2684

vsingh-msys: In the cookbook affected node list check-in time is one of the attributes that require to show.

In client-runs nodes, we consider `check-in` time as last run/end time and it sends over the data collector.

https://github.com/chef/chef/blob/633c4445e00f3174d156ac69a05be117a3fd0456/lib/chef/data_collector/run_end_message.rb#L63-L64

https://github.com/chef/automate/blob/0de8065086367dadc8e99b42d2b9b54b5b1d1657/components/ingest-service/backend/chef_client_run.go#L117-L119



In Chef Server we have only `ohai_time` which is close to check-in time. Also in chef manage, we showed ohai_time as a check-in time.

Ref: 
node-check-in

vsingh-msys: Here the problem is ohai_time is float and I am not able to propolate in gRPC API or golang parsing side.

In the demo, script https://gist.github.com/vsingh-msys/f53921a2d4eb89e3a41ad772bff253a0 is getting populated.

lancewf: This could be one of the problems, the demo is using "github.com/go-chef/chef" and the automate code is using "github.com/chef/go-chef".

vjeffrey: okk i see. i can't fully run the example code b/c there's a key needed, but if the problem is getting the float to a string, you should be able to use strconv for that:
```
	ohai := 1469581742.9826262
	fmt.Println(reflect.TypeOf(ohai))

	stringOhai := strconv.FormatFloat(ohai, 'E', -1, 64)
	fmt.Println(reflect.TypeOf(stringOhai))
```

vjeffrey: in any case, i'm putting my +1 on the pr. feel free to merge as is and make a card to come back to this!

vsingh-msys: @vjeffrey thankyou so much!!! it is working as expected.

Issue #2955, opened 2020-02-24
authn-service fails due to A1 data collector token projects error


bcmdarroch:
## Describe the bug
[Reported by `@axelrtgs` in the community slack](https://chefcommunity.slack.com/archives/CAZSL536K/p1582575944004700)

On updating to the latest Automate: 
```sudo chef-automate status
Status from deployment with channel [current] and type [local]
Service Name            Process State  Health Check  Uptime (s) PID  
deployment-service      running        ok            10060      29512
backup-gateway          running        ok            94         12640
automate-postgresql     running        ok            93         12684
automate-pg-gateway     running        ok            92         12767
automate-elasticsearch  running        ok            91         12810
automate-es-gateway     running        ok            91         12784
automate-ui             running        ok            91         12796
pg-sidecar-service      running        ok            91         12888
cereal-service          running        ok            90         12946
event-service           running        ok            90         13043
es-sidecar-service      running        ok            72         14896
event-feed-service      running        ok            71         14913
authz-service           running        ok            88         13175
automate-dex            running        ok            88         13298
teams-service           running        ok            87         13336
authn-service           DOWN           unknown       1          0    
secrets-service         running        ok            86         13511
applications-service    running        ok            85         13544
notifications-service   running        ok            84         13669
nodemanager-service     running        ok            85         13636
compliance-service      running        CRITICAL      18         16255
license-control-service running        ok            83         13959
local-user-service      running        ok            82         14143
session-service         running        ok            82         14090
ingest-service          running        ok            73         14839
infra-proxy-service     running        ok            82         14203
config-mgmt-service     running        ok            71         14935
data-feed-service       running        ok            80         14324
event-gateway           running        ok            79         14516
automate-gateway        running        unknown       79         14548
automate-load-balancer  running        ok            46         15584
12:26
```
According to `@axelrtgs`:
"the symptom of this is after login it can no longer validate our license and all our clients are reporting a 403 sending data to the datacollector"

## To Reproduce
The issue only shows up when there is an `a1_data_collector_token` configured.
```
rm -rf results/*.hart
hab studio enter
start_all_services
echo '[auth_n]
  [auth_n.v1]
    [auth_n.v1.sys]
      [auth_n.v1.sys.service]
        a1_data_collector_token = "LEGACY_TOKEN_HERE"' >> a1.toml
chef-automate config patch a1.toml
sl | grep authn-service
```
[...]

PR #2961, opened 2020-02-25
Fix IAM V2 bug for new A1 data collector token entries

tylercloke:

Since this commit https://github.com/chef/automate/commit/a2e1a87bc749cd89d5011c82b5379d6fdf8fdc7f we have been enforcing project permissions validation on token entries. When the authn server starts up, if it's configured with a legacy A1 data collector token, it will first check if it exists, then insert a token with the config specified value if it doesn't.

However, on IAM V2, it performs validation that the subject from context has permissions on the projects specified. In this case, we do not wish to validate projects because 1) we are just inserting the token in the default project (will always exist) and 2) this is a domain specific operation being initiated by the server which shouldn't need to round-trip to authz to ask for permission.

```
hab studio enter
build components/authn-service
start_all_services
echo '[auth_n]
  [auth_n.v1]
    [auth_n.v1.sys]
      [auth_n.v1.sys.service]
        a1_data_collector_token = "LEGACY_TOKEN_HERE"' >> a1.toml
chef-automate config patch a1.toml
sl | grep authn-service
# verify that token exists
```

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/authn-service/tokens/pg/adapter.go
+++ b/components/authn-service/tokens/pg/adapter.go
@@ -37,7 +37,7 @@ func (a *adapter) CreateTokenWithValue(ctx context.Context,
 		id = uid.String()
 	}
 
-	return a.insertToken(ctx, id, description, value, active, projects)
+	return a.insertToken(ctx, id, description, value, active, projects, true)
 }
 
 func (a *adapter) CreateLegacyTokenWithValue(ctx context.Context, value string) (*tokens.Token, error) {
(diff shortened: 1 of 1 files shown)


components/authn-service/tokens/pg/adapter.go
```diff
@@ -72,24 +72,30 @@ func (a *adapter) ResetToV1(ctx context.Context) error {
...
-		NewProjects:     projects,
-		IsUpdateRequest: false,
-	})
-	if err != nil {
-		return nil, err
+
+	// we do not wish to validate projects for the A1 legacy token case
+	// since that is an internal DB call kicked off as part of the server startup code
+	// with no subjects
+	if validateProjects {
```
blakestier: I think since we are only skipping validation for this single use, we could name this bool `isLegacyToken` or something to indicate how isolated this case is.

sdelano:
Fixed #2955

PR #2963, opened 2020-02-25
Jamie/bug table widths

SEAjamieD:
Long content in tables is causing layout issues

### :+1: Definition of Done
Content is ellipsed and no longer breaks layout/looks the same as it did prior to having long content in it.

cd components/automate-ui
make serve

To Test
navigate to https://a2-dev.test/settings/tokens
create a token with a really long name
observe that the long names in the api token list are now ellipsed instead of affecting layout

- [ ] Tests added/updated?
- [ ] Docs added/updated?

Before

broken

Screen Shot 2020-02-25 at 4 47 27 PM

### DIFF
--- a/components/automate-ui/src/app/components/table/table-cell/table-cell.component.scss
+++ b/components/automate-ui/src/app/components/table/table-cell/table-cell.component.scss
@@ -10,6 +10,10 @@
   border-top: 1px solid $chef-grey;
   overflow: hidden;
   text-overflow: ellipsis;
+
+  &.text-right {
+    text-align: right;
+  }
 }
 
 :host:first-child {
--- a/components/automate-ui/src/app/components/table/table-cell/table-cell.component.scss
+++ b/components/automate-ui/src/app/components/table/table-cell/table-cell.component.scss
@@ -10,7 +10,6 @@
   border-top: 1px solid $chef-grey;
   overflow: hidden;
   text-overflow: ellipsis;
-  white-space: nowrap;
 }
 
 :host:first-child {
--- a/components/automate-ui/src/app/components/table/table-cell/table-cell.component.scss
+++ b/components/automate-ui/src/app/components/table/table-cell/table-cell.component.scss
@@ -1,12 +1,16 @@
 @import '../table.component';
 
 :host {
+  display: inline-block;
   padding: $size_16;
   vertical-align: middle;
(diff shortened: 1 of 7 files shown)


SEAjamieD:
Whoops...this is currently a draft - please hold on reviews

components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.html
```diff
@@ -199,7 +199,7 @@ <h2 id="upload-modal-title" slot="title">Upload an archived profile (.tar.gz or
                   </a>
                 </chef-table-cell>
                 <chef-table-cell>{{ profile.version }}</chef-table-cell>
-                <chef-table-cell>
+                <chef-table-cell class="text-right">
```
msorens: Slight concern over why this table is unique...?

components/automate-ui/src/app/components/table/table-cell/table-cell.component.scss
```diff
@@ -1,12 +1,19 @@
 @import '../table.component';
 
 :host {
```
msorens: These changes fix all the "list of x" pages.

components/automate-ui/src/app/modules/policy/details/policy-details.component.scss
```diff
@@ -40,9 +40,12 @@ tbody > .detail-row {
 }
 
 .id-column {
```
msorens: Each of the "x-details" pages that shows a one-line table of the item needs these changes individually, so you will see this pattern applied in several files.
I only checked the auth-pages, though; are there others...?

components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.html
SEAjamieD: in order to make the ellipses apply, content needs to be `inline-block`, which then caused - in this elements case - it to no longer be right aligned.  A simple and low code fix is to use `text-align: right` which I wanted to be reusable incase it needed to be used elsewhere.
The reason it was happening only here (that I know of yet) is that most of the columns contain text, where as this column contain an icon so it was thrown off in a way the others weren't

PR #2965, opened 2020-02-25
add sql.Open wrapper that sets ConnectionMaxLifetime and pinging

jaym:
The db.PGOpen and db.PGOpenContext() helpers create a database
connection with a default maximum connection lifetime to something
less than automates pg gateway timeout. The goal of this change is to
avoid errors caused by attempts to use connections that have already
been closed by pg-gateway (which will close connections after 12 hours
of inactivity) or other middleboxes.

Since ConnectionMaxLifetime limits the total lifetime of the
connection, we likely want to avoid setting it too low since it will
force us to re-create db connections even on a busy server.

The periodic db pinger also serve to keep the idle connections
alive. Because we don't control the pooling though, it is likely we'll
still see the occasional timed-out connection.

Co-authored-by: Steven Danna <steve@chef.io>
Signed-off-by: Jay Mundrawala <jmundrawala@chef.io>


lib/db/open.go
```diff
@@ -0,0 +1,51 @@
...
+func PGOpenContext(ctx context.Context, dataSourceName string) (*sql.DB, error) {
+	opts := dbOpts{
+		maxConnectionLifetime:       10 * time.Minute,
+		maxConnectionLifetimeJitter: 10 * time.Minute,
+	}
+	db, err := sql.Open("postgres", dataSourceName)
+	if err != nil {
+		return nil, err
+	}
+	db.SetConnMaxLifetime(opts.maxConnectionLifetime + jitter(opts.maxConnectionLifetimeJitter))
```
stevendanna: `max` is a bit of a misnomer since we are _adding_ the jitter.  But as long as we document it, it's fine.  What is the reasoning behind wanting a jitter here?

lib/db/open.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func PGOpen(dataSourceName string) (*sql.DB, error) {
+	return PGOpenContext(context.Background(), dataSourceName)
+}
+
+func PGOpenContext(ctx context.Context, dataSourceName string) (*sql.DB, error) {
+	opts := dbOpts{
+		maxConnectionLifetime:       10 * time.Minute,
+		maxConnectionLifetimeJitter: 10 * time.Minute,
```
danielsdeleo: Is the idle connection timeout for the HA cluster 5 or 6 minutes? Should this total to less than that?

jaym: oh, yea
forgot about that

jaym: maybe we do 5 - jitter(1min)?

lib/db/open.go
jaym: try to spread out when things reconnect a little bit. Maybe it doesn't matter

lib/db/open.go
stevendanna: It looks like the timeout in HA cluster might be 50s:

https://github.com/chef/a2-ha-backend/blob/master/components/automate-backend-haproxy/config/haproxy.conf#L10-L13

lib/db/open.go
```diff
@@ -0,0 +1,96 @@
...
+
+func PGOpen(dataSourceName string, opts ...PGOpenOption) (*sql.DB, error) {
+	return PGOpenContext(context.Background(), dataSourceName, opts...)
+}
+
+func PGOpenContext(ctx context.Context, dataSourceName string, userOpts ...PGOpenOption) (*sql.DB, error) {
+	opts := &dbOpts{
+		maxConnectionLifetime:       15 * time.Minute,
+		maxConnectionLifetimeJitter: 15 * time.Minute,
+		pingInterval:                20 * time.Second,
```
stevendanna: This enables the pinger by default.  Given that this only happens after 12 hours of inactivity, I'm not sure it is worth it.  But I suppose people using a2-cluster will hit this more often since their timeout is lower and this might help there.  I'm kinda loath to set maxConnectionLifetime to less than a minute...

danielsdeleo: IMO enabling the pinger gets us closer to the optimal outcome and the price is low. It should help keep our tests and such stable as well which is a pretty solid benefit.

PR #2956, opened 2020-02-25
[infra-proxy-service] Add Roles list & detail gRPC & HTTP API

vsingh-msys:
### :nut_and_bolt: Description: What code changed, and why?

- Add infra role list API with attributes name.

gRPC API:
```
chef-automate dev grpcurl infra-proxy-service chef.automate.domain.infra_proxy.service.InfraProxy.GetRoles -- -d '{"server_id": "c5ff62b5-bbfa-4e8c-892f-638ebd6387ec", "org_id": "7e343a0a-f2cd-4fa7-95d8-044396d7e280"}'
```
HTTP API:
```
curl -sSkH "api-token: $TOK" https://a2-dev.test/api/v0/infra_proxy/servers/{server_id}/orgs/{org_id}/roles

Response format:

{
  "roles": [
    {
      "name": "string"
    }
  ]
}
```

- Add infra role detail API with attributes name, chef_type, description, run_list, json_class

gRPC API:
```
chef-automate dev grpcurl infra-proxy-service chef.automate.domain.infra_proxy.service.InfraProxy.GetRole -- -d '{"server_id": "c5ff62b5-bbfa-4e8c-892f-638ebd6387ec", "org_id": "7e343a0a-f2cd-4fa7-95d8-044396d7e280", "name": "vivek23"}'
```
HTTP API:
```
curl -sSkH "api-token: $TOK" https://a2-dev.test/api/v0/infra/servers/{server_id}/orgs/{org_id}/roles/name

Response format:

{
  "name": "string",
  "chef_type": "string",
  "description": "string",
  "run_list": [
    "string"
  ],
  "json_class": "string"
}
```

### :chains: Related Resources
#1544 

### :+1: Definition of Done
 - Able to fetch roles list and detail API.

### :athletic_shoe: How to Build and Test the Change
 - `rebuild components/infra-proxy-service`
 - `rebuild components/automate-gateway`

### :white_check_mark: Checklist

- [x] I have read the [CONTRIBUTING document](https://github.com/chef/automate/blob/master/CONTRIBUTING.md).
- [ ] Tests added/updated?
- [ ] Docs added/updated?
- [x] All commits have been signed-off for the [Developer Certification of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

### :camera: Screenshots, if applicable
Signed-off-by: Vivek Singh <vivek.singh@msystechnologies.com>

PR #2976, opened 2020-02-27
go.sum: go mod tidy

stevendanna:
The go.sum change was produced by running

    go mod tidy -v

from the root of the repo. To try to avoid letting this drift in the
future, we now also run go mod tidy in our repo_health script.

### DIFF
--- a/scripts/repo_health.sh
+++ b/scripts/repo_health.sh
@@ -6,6 +6,7 @@ hab pkg install -b core/go core/git core/ruby core/jq-static core/shellcheck cor
 
 echo "Checking Go Dependencies"
 go mod verify
+go mod tidy -v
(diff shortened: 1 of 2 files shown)


vsingh-msys:
Have committed these changes in PR https://github.com/chef/automate/pull/2956/files#r383690721 will rebase once this PR will get merged.

Thanks!

scripts/repo_health.sh
```diff
@@ -6,6 +6,7 @@ hab pkg install -b core/go core/git core/ruby core/jq-static core/shellcheck cor
 
 echo "Checking Go Dependencies"
 go mod verify
+go mod tidy -v
```
stevendanna: I think the problem with this is that it'll download all the dependencies and thus open us up to more network flakiness

PR #2979, opened 2020-02-27
Allow user to configure load balancer's client certificate checking 

### LINKED ISSUE #2968: Nginx SSL Config Options
Issue details [here](https://github.com/chef/customer-bugs/issues/33)

## Issue Summary

Customer is using LibreSSL on workstations which appears to not support the ssl_client_certificate handshake and needs to disable it for the load balancer in Automate 2

They are currently commenting out the the following settings in "/hab/svc/automate-load-balancer/config/nginx.conf", but the system will periodically restore the original config, maybe due to automatic updates, requiring them to continually modify the config by hand to restore access to A2

ssl_client_certificate /hab/svc/automate-load-balancer/config/root_ca.crt;
ssl_verify_client optional;

## Expected Results

Disabling ssl_client_certificate and ssl_verify_client configs should stay disabled.
Actual Results

The disabled option in the nginx config gets automatically restored

## Repro Steps (if possible)

Setup a system to use LibreSSL instead of OpenSSL and try to use CURL to access the A2 API.

yzl:
Expose `ssl_verify_client` and `ssl_client_certificate` in the load balancer's nginx config so that they can be set by a user via the Automate config.

Fixes #2968.

- [ ] Tests added/updated?
- [ ] Docs added/updated?


api/config/load_balancer/config_request.go
```diff
@@ -61,6 +61,8 @@ func DefaultConfigRequest() *ConfigRequest {
 	c.V1.Sys.Ngx.Http.LargeClientHeaderBuffersSize = w.String("8k")
 	c.V1.Sys.Ngx.Http.Sendfile = w.String("on")
 	c.V1.Sys.Ngx.Http.SslCiphers = w.String(config.ExternalCipherSuite)
+	c.V1.Sys.Ngx.Http.SslClientCertificateFilename = w.String("root_ca.crt")
+	c.V1.Sys.Ngx.Http.SslVerifyClient = w.String("optional")
```
stevendanna: I wonder, while we are here, do we want to change this default to off?  Turning it off will make us compatible with libressl by default and I doubt many people are using client certs at the lb.

components/automate-chef-io/content/docs/configuration.md
```diff
@@ -355,7 +355,11 @@ To configure your Chef Automate installation's load balancer, create a TOML file
 # large_client_header_buffers_number = 4
 # sendfile = "on"
 # ssl_ciphers = "ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:AES256-GCM-SHA384:!aNULL:!eNULL:!EXPORT"
+# Setting ssl_client_certificate_filename to "" will result in ssl_client_certificate not
+# being set in nginx.conf.
+# ssl_client_certificate_filename = "root_ca.crt"
```
kagarmoe: Is setting ssl_client_certifivate_filename to a non-empty value the desired user action?

yzl: No. We expect most users to accept the default of root_ca.crt.

kagarmoe: Thanks! That's the answer I hoped for.

kagarmoe:
```suggestion
# We recommend against setting ssl_client_certificate_filename to an empty string (""). 
# Using an empty string causes the nginx.conf to inherit an empty setting for the ssl_client_certificate, which prevents the load balancer from communicating with the server.
# ssl_client_certificate_filename = "root_ca.crt"
```

kagarmoe: Not sure that I have the correct thing this breaks.

api/config/load_balancer/config_request.go
yzl: Since I changed the default for `ssl_verify_client` to off, I also made the default for `ssl_client_certificate_filename` to `""` for consistency.

components/automate-load-balancer/habitat/config/nginx.conf
```diff
@@ -111,8 +111,10 @@ http {
     ssl_certificate {{../pkg.svc_data_path}}/{{tls.server_name}}.cert;
     ssl_certificate_key {{../pkg.svc_data_path}}/{{tls.server_name}}.key;
 
-    ssl_client_certificate {{../pkg.svc_config_path}}/root_ca.crt;
-    ssl_verify_client optional;
+    {{#if ../cfg.ngx.http.ssl_client_certificate_filename ~}}
+    ssl_client_certificate {{../pkg.svc_config_path}}/{{ ../cfg.ngx.http.ssl_client_certificate_filename }};
```
jaym: have you tested this. I'm fairly certain there is no way to make this work with authz

yzl:
We've decided to turn off client certificate checking at the load balancer, as it was done as a research spike for the automate-deployed chef infra server. Closing.

Issue #2694, opened 2020-01-22
Annotate tokens.proto with API docs


tylercloke:
See the parent issue for more details.

# Acceptance Criteria

* API documentation is updated for all endpoints in [tokens.proto](https://github.com/chef/automate/blob/master/components/automate-gateway/api/iam/v2/tokens.proto) and related common / request / response files.
* Link to these new docs and remove the old ones [from here](https://automate.chef.io/docs/iam-v2-api-reference/).

susanev:
closing in favor of #2982

Issue #2614, opened 2020-01-13
Annotate policy.proto with API docs


susanev:
# User Story 

We have a new [auto-generated API docs](https://automate.chef.io/docs/api/). We eventually want to add all of the IAM APIs to this new section, but for now we have deemed the Policy API the highest priority, since it is the most complex of our APIs and cannot be used via the Automate UI.

# Technical Details
In order to make use of the auto-generation capabilities, we must add the appropriate annotations to the [policy.proto](https://github.com/chef/automate/blob/master/components/automate-gateway/api/iam/v2/policy.proto) file. Follow this [nodes.proto](https://github.com/chef/automate/blob/master/components/automate-gateway/api/nodes/nodes.proto) as an example.

We also need to set up an IAM tag to be used as a heading in the table of contents. Follow [these tag setup instructions](https://github.com/chef/automate/tree/master/components/automate-chef-io#tags-and-tag-groups).

## Resources
- [How to build API docs locally](https://github.com/chef/automate/tree/master/components/automate-chef-io#building-api-docs)
- channel in slack: #a2-api-documentation

susanev:
@tylercloke lets also remove the policy stuff from the manually created api reference, and link to the api docs in this work
https://automate.chef.io/docs/iam-v2-api-reference/#policies

if we can get that example policy over in the api docs that would be rad

susanev:
closing in favor of #2982

PR #2982, opened 2020-02-28
Policy and token api docs

msorens:

This PR combines the previously separate PRs for policy and token API documentation.

Added to those PRs are the following:

1. New sections for "Identity" and for "Access Management".
2. Removed duplication in the left navigation panel (duplication was due to presence of both v2 and v2beta paths, but the v2beta path is no more).
3. Addition of "Authorization Actions" for each endpoint. These annotations are automatically added (idempotently) whenever an API endpoint is documented. 
4. Annotations have been manually added to indicate required fields in the `request`.
5. As to the response, see the next diagram...

image

1. If you expand the response...
2. Then expand the top-level object in the response, the response is thoroughly documented below it (actions, id, name, etc.)
3. But notice here, that this top-level of the response (`CreateRoleResp` in this example) does not get documentation applied. That is the "blah blah blah" here would be ignored.
```
message CreateRoleResp {
  // blah blah blah
  Role role = 1;
}
```
That is mostly OK, because everything in the `Role` and below is well-documented, which is to say everything inside here is documented:
```
message Role {
  . . .
}
```
4. In the response here, notice there are no `required` markers like the request had. To compensate for that, for any list of things I denote like this for "required"...
5. And like this for "optional".
6. Finally, here we see that enumerated types are nicely documented automatically. That's great... except that you cannot add any other custom notes on that field; they are just ignored. (So you will see that I have noted that in the proto files with "This doc-comment is ignored for an enum.")

image

On the other hand, remember in (3) just above I said the top-level response object ignores doc-comments? That is true for complex objects; for simple types, the doc-comments are published, as shown here:

image

The last new piece (thanks @kagarmoe!) is that I added "section introductions". When you select a topic in the left navigation panel (1), there is a brief introduction on the main page (2). The text of that paragraph is obscurely located in automate-chef-io/data/docs/api-static/03-tags.swagger.json, not in the proto file(!), which is why I wanted to call it out explicitly here. The second point about this is that it is specified as a JSON attribute, so the markdown it supports will, by necessity, be rather limited. Sigh.

image

PR #2614 Annotate policy API
PR #2694 Annotate token API

### :+1: Definition of Done
API documentation in place for policies and tokens

Install Hugo if needed
```
make serve
open http://localhost:1313/docs/api/
```

- [ ] Tests added/updated?
- [x] Docs added/updated?


kagarmoe:
Thank you for your careful write-up @msorens ! The response is indeed...complicated? idiosyncratic? strange? Whatever it is, you've caught its nuances well.

PR #2992, opened 2020-03-02
Add waivers to report api call

alexpop:
With this change, a downloaded compliance report will contain controls waivers information.


api/external/compliance/reporting/reporting.proto
```diff
@@ -655,6 +655,20 @@ message Control {
 	repeated Ref refs = 8;
 	// Metadata defined on the control in key-value format.
 	map<string, string> tags = 9;
+	// Indicates if the control has been waived or not
+	string waiver_str = 10;
```
vjeffrey: @alexpop i think the mismatch here is what's causing all the failures -- here waiver_str is 10 and orig data is 11; on the internal proto waiver_str is 11 and orig data is 12 -- mismatches there will cause the api requests to fail

api/external/compliance/reporting/reporting.proto
```diff
@@ -655,6 +655,20 @@ message Control {
 	repeated Ref refs = 8;
 	// Metadata defined on the control in key-value format.
 	map<string, string> tags = 9;
+	// Indicates if the control has been waived or not
+	string waived_str = 10;
```
vjeffrey:
```suggestion
	string waived_str = 11;
```

api/external/compliance/reporting/reporting.proto
```diff
@@ -655,6 +655,20 @@ message Control {
...
+	// Additional details for waived controls
+	OrigWaiverData waiver_data = 11;
```
vjeffrey:
```suggestion
	OrigWaiverData waiver_data = 12;
```

api/external/compliance/reporting/reporting.proto
vjeffrey: ^ external api calls won't work without that change

vjeffrey:
i pushed a change for the external proto to match the internal one, and rebased since pipeline failures are resolved now

Issue #2804, opened 2020-02-06
Cannot manually remove more than 3000ish disconnected services via CLI


danielsdeleo:
## User Story

We provide the `chef-automate applications remove-svcs` command to manually remove decommissioned services. This command doesn't work for some value above 3000 services. There are probably two or more underlying problems:

* To get the services for display (and to collect the IDs), we don't use pagination so the max gRPC message size can be exceeded.
* Queries can time out when selecting services for deletion.

## Definition of Done

Should be able to delete 10k or more disconnected services.

PR #3005, opened 2020-03-02
Use a streaming API so CLI can show/rm large numbers of services

danielsdeleo:

The `chef-automate applications show-svcs` and `remove-svcs` commands are limited to 3000ish services at a time because they are based on a unary RPC and hit the grpc message size limit when returning lots of services (show is the first step of remove so remove hits this limit then).

This change uses a streaming RPC to get around the grpc message size limit without introducing the consistency problems you can get with pagination (e.g., we don't cursor over a consistent result set so items can be missing, duplicated, etc.).

See #2804

### :+1: Definition of Done

Acceptance criteria says 10k services can be deleted; I have tested this with 20123 services using the procedure below (with the data generation script changed to 20123)

Build `automate-gateway`, `automate-cli` and `applications-service`, start all services.

`chef-automate applications show-svcs` will now be using the streaming API. To show that this lets you show or delete more than 3000 services, you can create a lot of services with a script like this:

```
for _ in {1..5000} ; do applications_publish_supervisor_message --sup-id "$(uuidgen)" --application example; done
```

After that, you can confirm they were all created with a command like:

```
chef-automate applications show-svcs | wc -l
```

They can be deleted with something like

```
chef-automate applications remove-svcs --all --yes  | tail
```

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/applications-service/pkg/server/server.go
```diff
@@ -262,6 +262,35 @@ func (app *ApplicationsServer) GetServices(
...
+	if err != nil {
+		return status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	sortField, sortAsc, err := params.GetSortParamsForServices(request.GetSorting())
+	if err != nil {
+		return status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	services, err := app.storageClient.GetServices(sortField, sortAsc, 1, 0, filters)
```
danielsdeleo: I took the approach here of ignoring any incoming pagination params and then disabling pagination in this call. We could instead check for pagination params and return a 400 or possibly even allow pagination with different defaults (though I don't know why you would do that when the unary RPC is already there as an alternative).

lancewf:
This branch might need rebased because I am getting the below error when trying to build automate-gateway. 
```
go: inconsistent vendoring in /src:
	golang.org/x/text@v0.3.1-0.20171218113626-eb22672bea55: is explicitly required in go.mod, but vendor/modules.txt indicates golang.org/x/text@v0.3.2

run 'go mod vendor' to sync, or use -mod=mod or -mod=readonly to ignore the vendor directory
```

PR #2935, opened 2020-02-21
[automate-2914] force-upgrade integration scenario: v1 -> v2 with migrated legacy policies

bcmdarroch:
### :nut_and_bolt: Description: What code changed, and why?

This PR adds integration testing for the following scenario:
1. deploys an older version of Automate using IAM v1 (this will be the case for any customer using Automate who has not upgraded to IAM v2)
2. upgrades Automate to the latest build, which force-upgrades the system to IAM v2

♻️ For the force-upgrade, we have removed the following v2-specific integration tests scripts:

- iam_v2_no_legacy_policies.sh
- iam_v2_upgrade_diagnostics.sh
- iam_v2_with_v1_policies.sh
- upgrade_reset_iam_v2.sh
- reset_admin_access_v2.sh

These scripts all used `upgrade-to-v2` and `reset-to-v1`, commands which will be deprecated by the force-upgrade.

🎉  Now we have the following integration tests, all prefixed by `iam`: 
- iam v1 to v2 force-upgrade
- iam v1 to v2 force-upgrade (diagnostics only)
- iam reset admin access
- iam automate-gateway integration
- iam v2 with legacy to v2 force-upgrade (coming soon! https://github.com/chef/automate/pull/3008)
- iam v2 with no legacy to v2 force-upgrade (coming soon! https://github.com/chef/automate/pull/3009)
- iam v2 with no legacy to v2 force-upgrade (diagnostics only) (coming soon! https://github.com/chef/automate/pull/3009)

The inspec test suites have also been reorganized:
- `a2-api-integration` tests API behavior. This now includes all IAM v2 APIs. To distinguish from the iam inspec suites, these tests all assume admin-level access.
- `a2-iam-v1-integration` tests permissions on APIs specific to v1 (default policies)
- `a2-iam-no-legacy-integration` tests permissions on APIs on a fresh system without v1 legacy policies
- `a2-iam-legacy-integration` tests permissions on APIs on a system with v1 legacy policies

### :chains: Related Resources
follow-ups:
v2 force-upgrade to v2 with legacy policies: https://github.com/chef/automate/pull/3008
v2 force-upgrade to v2 with no legacy policies: https://github.com/chef/automate/pull/3009

### :+1: Definition of Done
- [ ] all buildkite tests are passing, except for the ones related to ingest permissions (to be fixed by #2987)

### :athletic_shoe: How to Build and Test the Change
when you check out this branch and start in a fresh studio, you can mimic the force-upgrade integration scenario following these steps:

[...]

PR #3009, opened 2020-03-02
[automate-2916] v2 with no legacy policies -> force-upgrade to latest v2 integration test

bcmdarroch:

This test verifies that anyone currently using IAM v2 who skipped migrating legacy policies when they upgraded will not have their system disrupted by the force-upgrade. No legacy policies should reappear following the force-upgrade.

leverages new inspec tests added in https://github.com/chef/automate/pull/2935
- [chef_managed_access.rb](https://github.com/chef/automate/pull/2935/files#diff-cc29b6a8b025e5d1f1853507b4fe899e)
- [custom_access.rb](https://github.com/chef/automate/pull/2935/files#diff-4858357bfa9ea1ef549514e4f39812d2)

### :+1: Definition of Done
- [x] `iam v2 force-upgrade to v2 with no legacy policies` passes buildkite
- [ ] `iam v2 force-upgrade to v2 with no legacy policies (diagnostics only)` passes buildkite

- see https://github.com/chef/automate/pull/2935

- [x] Tests added/updated?
- [ ] Docs added/updated?


integration/tests/iam_v2_no_legacy_to_v2_force_upgrade.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+#shellcheck disable=SC2034
+test_name="iam_force_upgrade_to_v2_with_no_legacy"
+test_upgrades=true
+test_upgrade_strategy="none"
+test_deploy_inspec_profiles=(a2-iam-no-legacy-integration)
+test_upgrade_inspec_profiles=(a2-iam-no-legacy-integration a2-api-integration)
```
msorens: Your purpose states that these changes verify two things:

_(a) The system will "not be disrupted" by the force-upgrade._
I am wondering if you could make this more concrete. Does "not disrupted" mean something like: it runs a standard battery of tests that all pass? If so, how do we know what constitutes that standard suite? Or does it mean something else?

_(b) No legacy policies will be present following the force-upgrade._
Which of the tests embedded in the controls above confirm this?

integration/tests/iam_v2_no_legacy_to_v2_force_upgrade_diagnostics.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+#shellcheck disable=SC2034
+test_name="iam_force_upgrade_to_v2_with_no_legacy"
+test_upgrades=true
+test_upgrade_strategy="none"
+test_diagnostics_pre_upgrade_filters="~skip-for-deep-upgrade"
+
+# Note: we can't run diagnostics AND inspec, so we don't include any inspec tests
+test_skip_diagnostics=false
```
msorens: So here we are saying to run "diagnostics" instead of inspec. Not being familiar with diagnostics, could you elaborate on what this is doing, what test are being run if not inspec tests, etc. Thanks!

integration/tests/iam_v2_no_legacy_to_v2_force_upgrade.sh
bcmdarroch: `a2-api-integration` verifies that our APIs continue to behave as we expect. I'm actually going to swap that for `a2-deployment-integration`, which includes `a2-api-integration` and some system checks (as in, all the services are up and running, etc).

`a2-iam-no-legacy` verifies that permissions work as expected on fresh v2 install with no legacy policies. The key test here is `chef_managed_access.rb` which verifies permissions for all Chef-managed roles

bcmdarroch: I've added some comments to this effect

.expeditor/verify_private.pipeline.yml
```diff
@@ -448,6 +448,32 @@ steps:
         linux:
           privileged: true
 
+  - label: "iam v2 force-upgrade to v2 with no legacy"
```
blakestier: this name is so long we may as well add the "policies" after "legacy"

integration/tests/iam_v1_to_v2_force_upgrade.sh
```diff
@@ -1,11 +1,26 @@
 #!/bin/bash
 
+# this test script:
+# 1. deploys an older version of Automate on IAM v1.
+# 2. runs inspec tests to verify IAM v1 behavior.
```
tylercloke: What's the point of running the inspec tests on the old install? Are we using it as a health check?

bcmdarroch: Yeah I was thinking they'd be helpful as a sanity check and a set of v1 behavior expectations we could refer to.

PR #3010, opened 2020-03-03
Allow auth for server streams

danielsdeleo:

I am hoping to implement a server-side streaming RPC; i.e., single client request returns a stream of messages in response. I found that auth was hard-coded to return an error for this case, with a nearby code comment saying:
> we haven't (conceptually) figured out authz for streaming services

I can understand how this may be the case for client-side streaming or bi-directional streams, but I think server-side streams are conceptually similar to the non-streaming case from an auth perspective. Thus, this PR implements auth for server streams.

### DIFF
--- a/components/automate-gateway/gateway/middleware/auth_interceptor.go
+++ b/components/automate-gateway/gateway/middleware/auth_interceptor.go
@@ -98,6 +98,60 @@ type authInterceptor struct {
 	authz GRPCAuthorizationHandler
 }
 
+func (a *authInterceptor) combinedAuth(ctxIn context.Context, req interface{}) (context.Context, error) {
+	// extract request-scoped logger
+	log := ctxlogrus.Extract(ctxIn)
+
+	// this context is only used for authenticating the request: we need
+	// headers!
+	// grpc-gateway translates these into metadata, and we'll forward that to
+	// authn-service below. we don't want that metadata on _every_ request,
+	// though, so this context is separated out
+	authCtx := ctxIn
+
+	// transfer incoming metadata to outgoing metadata
+	// if !ok, this will default to the ctx set above, and most likely be
+	// rejected by authn-service. However, we don't want to duplicate the
+	// "when is a request potentially good for authentication" logic here,
+	// so this will just go to authn-service as-is.
+	var md metadata.MD
+	var ok bool
+	if md, ok = metadata.FromIncomingContext(authCtx); ok {
+		authCtx = metadata.NewOutgoingContext(authCtx, md)
+	}
+
+	var subs []string
+	var authResponse *authn.AuthenticateResponse
+	var err error
+
+	certSubject, ok := headerAuthValidForClientAndPeer(ctxIn)
+	if ok && !fromGateway(certSubject) {
+		subs = []string{certSubject}
+		log.Debugf("using client cert to authenticate request: %q", certSubject)
+	} else { // ordinary authn by calling authn-service
+		authResponse, err = a.authn.Authenticate(authCtx, &authn.AuthenticateRequest{})
+		if err != nil {
+			log.Debugf("error authenticating request: %s", err)
+			return nil, err
+		}
+		subs = append(authResponse.Teams, authResponse.Subject)
+	}
+
+	projects := auth_context.ProjectsFromMetadata(md)
+
+	ctxOut, err := a.authz.Handle(authCtx, subs, projects, req)
+	if err != nil {
+		return nil, err
+	}
+
+	// pass on all auth metadata to domain services
+	// services can use that metadata in authz-related decisions
+	// like allowing project assignment
+	log.Debug("injecting auth context to downstream")
+	ctxOut = auth_context.NewOutgoingContext(ctxOut)
+	return ctxOut, nil
+}
+
 // UnaryInterceptor returns a grpc UnaryServerInterceptor that performs AuthN/Z.
 func (a *authInterceptor) UnaryServerInterceptor() grpc.UnaryServerInterceptor {
 	return func(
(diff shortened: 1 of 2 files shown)


components/automate-gateway/gateway/middleware/auth_interceptor.go
```diff
@@ -98,6 +98,60 @@ type authInterceptor struct {
...
+		if err != nil {
+			log.Debugf("error authenticating request: %s", err)
+			return nil, err
+		}
+		subs = append(authResponse.Teams, authResponse.Subject)
+	}
+
+	projects := auth_context.ProjectsFromMetadata(md)
+
+	ctxOut, err := a.authz.Handle(authCtx, subs, projects, req)
```
danielsdeleo: Would appreciate some double-checking that I have the right contexts going in and coming out of the various places

components/automate-gateway/gateway/middleware/auth_interceptor.go
```diff
@@ -106,87 +160,67 @@ func (a *authInterceptor) UnaryServerInterceptor() grpc.UnaryServerInterceptor {
...
+				return handler(req, ss)
 			}
 			return status.Errorf(codes.PermissionDenied, "unauthorized access to %s", info.FullMethod)
 
 		}
-		return errors.New("to be implemented")
+
+		// A client stream kinda resembles a sequence of several requests. We do
+		// not have consensus about how we want do deal with that so these are
+		// disallowed for now.
```
danielsdeleo: I added this here to make the change more conservative, but I'm also just as happy to remove it, if we are comfortable with client-side streams now.

vjeffrey:
hmmm, would this pr mean that we can change all of our mux streaming routes to normal grpc/gateway ones? 
(eg: https://github.com/chef/automate/blob/master/components/automate-gateway/gateway/services.go#L647)

i know we have a note on one endpoint about gateway also not supporting the stream, but maybe that's only for client side? i honestly don't remember: https://github.com/chef/automate/blob/master/api/external/compliance/profiles/profiles.proto#L45

srenatus:
@vjeffrey I don't think so, there's https://github.com/grpc-ecosystem/grpc-gateway/issues/958 holding us back there. Haven't followed recent developments on that issue, though.

components/automate-gateway/gateway/middleware/auth_interceptor.go
tylercloke: That looks right to me, but maybe worth adding some testing to that effect. It'd also be nice to see this in action with some e2e tests / somewhere we could see the context we are expecting show up. Hard to validate this behavior without an upstream.

danielsdeleo:
@tylercloke check out https://github.com/chef/automate/pull/3005 -- the second commit makes use of this PR. It's a simple server-side streaming endpoint that we access via automate-cli. It _could_ have been done with pagination but then you have to worry about the result set/order changing in between requests, which gives the client an incorrect result at the end.

components/automate-gateway/gateway/middleware/auth_interceptor.go
bcmdarroch: I'm good with keeping this a bit conservative

components/automate-gateway/gateway/middleware/auth_interceptor_internal_test.go
```diff
@@ -74,6 +102,38 @@ func TestHeaderAuthValidForClientAndPeer(t *testing.T) {
 	}
 }
 
+func TestComboAuthForwardsMetadataCorrectly(t *testing.T) {
```
bcmdarroch: I'd like to see a failure case added here, where authz-service returns an unauthorized error ([here](https://github.com/chef/automate/blob/master/components/automate-gateway/gateway/middleware/authv2/authv2.go#L112)) and it bubbles that error up correctly.

components/automate-gateway/gateway/middleware/auth_interceptor_internal_test.go
```diff
@@ -74,6 +111,82 @@ func TestHeaderAuthValidForClientAndPeer(t *testing.T) {
...
+			runtime.MetadataPrefix+"projects", "project1,project2",
+		))
+
+	t.Run("It returns error for authn failure", func(t *testing.T) {
+		a := authInterceptor{
+			authn: testAuthnClient{err: errors.New("sample authn failure")},
+			authz: testAuthzHandler{},
+		}
+		_, err := a.combinedAuth(reqFromALBCtx, nil)
+		require.Error(t, err)
```
bcmdarroch: [nit] this should be an `assert` since it's testing for the error

components/automate-gateway/gateway/middleware/auth_interceptor_internal_test.go
```diff
@@ -74,6 +111,82 @@ func TestHeaderAuthValidForClientAndPeer(t *testing.T) {
...
+	})
+	t.Run("It returns error for authz failure", func(t *testing.T) {
+		a := authInterceptor{
+			authn: testAuthnClient{},
+			authz: testAuthzHandler{err: errors.New("sample authz failure")},
+		}
+		_, err := a.combinedAuth(reqFromALBCtx, nil)
+		require.Error(t, err)
```
bcmdarroch: this could assert that the error is [grpc code 7 Permission Denied](https://github.com/grpc/grpc-go/blob/master/codes/codes.go#L72)

components/automate-gateway/gateway/middleware/auth_interceptor_internal_test.go
danielsdeleo: The actual `err` here is the `errors.New("sample authn failure")` from above because the combo auth function is just passing the error through. So we have to trust that the authenticator is doing the right thing.

components/automate-gateway/gateway/middleware/auth_interceptor_internal_test.go
danielsdeleo: Same deal as above, this error is whatever the mocks provide.

components/automate-gateway/gateway/middleware/auth_interceptor_internal_test.go
danielsdeleo: That said, I added a check to assert that we don't break something that would otherwise be working (e.g., by using `errors.Wrap`) in the `combinedAuth` method, below.

PR #3024, opened 2020-03-03
Modify project update to update 1 ES index at a time

jaym:
Previously, the project update tasks were trying to run an
update across all the indexes at the same time. ES wasn't
doing a good job of throttling this. In the perf environment,
updates for most documents failed because of queues filling up.

The errors looked something like this:
```
compliance-service.default(O): time="2020-02-26T14:04:23Z" level=error msg="Unable to ingest inspec_summary object" error="InsertInspecSummary: elastic: Error 429 (Too Many Requests): rejected execution of processing of [185283930][indices:data/write/bulk[s][p]]: request: BulkShardRequest [[comp-7-s-2020.02.26][4]] containing [index {[comp-7-s-2020.02.26][_doc][a7075581-884a-48d8-aa3b-980cce3dabdc], source[n/a, actual length: [2.1kb], max length: 2kb]}], target allocation id: VovYu0VRTf-v3iSZYVMexw, primary term: 1 on EsThreadPoolExecutor[name = 8FHoLNx/write, queue capacity = 200, org.elasticsearch.common.util.concurrent.EsThreadPoolExecutor@5b299396[Running, pool size = 8, active threads = 8, queued tasks = 200, completed tasks = 26703240]] [type=es_rejected_execution_exception]"
```

This way should be much more efficient. It avoids relying on ES to
throttle correctly when using wildcard index names which match a lot
of indexes. We will waste less work trying to aggregate results across
all of the indexes and instead focus on one at a time. Don't have to
think about what will happen with a lot of scrolls open at the same
time.

The previous method also didn't correctly account for conflicting
documents when trying to figure out if an update query succeeded. This
meant that any indexes actively receiving traffic would likely end up
failing. The new way of checking for failure is to check the response
key in the returned task information. For the update by query task, it
has a failures field.


components/authz-service/server/v2/project_update_manager.go
```diff
@@ -390,31 +402,77 @@ func (m *CerealProjectUpdateManager) Status() (ProjectUpdateStatus, error) {
...
+			} else {
+				if err := subWorkflow.GetResult(&state); err != nil {
+					logrus.WithError(err).Errorf("failed to get result for %q", d)
+					continue
+				}
+			}
+
+			if state.Phase == project_update_tags.SerializedProjectUpdateWorkflowPhaseRunning && state.Running.TotalTasks > 0 {
+				progress = append(progress,
+					state.Running.RunningTask.Status.PercentageComplete*(1.0/float32(state.Running.TotalTasks)))
```
jaym: this should just be `state.Running.RunningTask.Status.PercentageComplete`

components/authz-service/server/v2/project_update_manager.go
```diff
@@ -390,7 +415,82 @@ func (m *CerealProjectUpdateManager) Status() (ProjectUpdateStatus, error) {
...
+	if totalWeight == 0 {
+		return 0, nil
+	}
+	var prog float32
+	for i := range weight {
+		prog += (progress[i] * float32(weight[i]) / float32(totalWeight))
+	}
+
+	return prog, nil
+}
```
lancewf: I am not exactly sure how this is working, but I put in example numbers and I got a good estimate of completion. I have also run this many times and the percentage looks correct.

jaym: Ignoring nodemanager:
The general idea is each index is a task. We give each task the same weight, 1.
So the total weight would be the number of indexes.
To get there:
The current running task has a weight of 1, and its progress is whatever we get from looking at the ES task (some number less than 1).
Then you have the indexes that are complete: the progress of  these is 1 because they are all complete, but it has a weight of the number completed indexes
Then you have some number of indexes that are not running yet. These have a progress of 0, and a weight of the number of remaining indexes.

So: 
total weight = denominator =  1 (which is the current running index) + # completed index + # remaining indexes
numerator = progress of running task + 1 (progress of completed tasks which is 100%) * # completed tasks + 0 (which is the progress of the incomplete tasks) * # incomplete tasks

components/authz-service/server/v2/project_update_manager.go
```diff
@@ -390,7 +415,82 @@ func (m *CerealProjectUpdateManager) Status() (ProjectUpdateStatus, error) {
...
+	domainServicesUpdateInstance, err := w.GetUpdateDomainServicesInstance()
+	if err != nil {
+		return 0, err
+	}
+
+	progress := []float32{}
+	weight := []int{}
+
+	for _, d := range domainServicesUpdateInstance.ListSubWorkflows() {
+		subWorkflow, err := domainServicesUpdateInstance.GetSubWorkflow(d)
```
bcmdarroch: what's an example of a `subWorkflow`? does it represent something like an ES job on compliance? or is it more specific or more general than that?

components/compliance-service/ingest/ingestic/ingestic.go
```diff
@@ -580,7 +596,7 @@ func getPercentageComplete(status interface{}) (float64, bool) {
 		return 1, true
 	}
 
-	return (created + deleted + updated) / total, true
+	return (created + deleted + updated + conflicts + noops) / total, true
```
bcmdarroch: does this make our time estimate more accurate?

components/compliance-service/ingest/ingestic/project_update.go
```diff
@@ -0,0 +1,128 @@
...
+		tasks = append(tasks, project_update_lib.SerializedProjectUpdateTask{
+			Priority: priority,
+			Params:   params,
+		})
+	}
+
+	return tasks, nil
+}
+
+func (backend *ESClient) RunProjectUpdateTask(ctx context.Context, projectUpdateID string,
```
bcmdarroch: was this code moved from somewhere else? what was running the project update before?

lib/authz/serialized_project_rules_update_workflow.go
```diff
@@ -0,0 +1,638 @@
...
+)
+
+var errCanceled = errors.New("Canceled")
+
+// SerializedProjectUpdateWorkflowExecutor is a workflow that allows running 1
+// task at a time across many services.
+// Given a list of services, it first asks each service to list what tasks it
+// would like to run. For example, a service would return a task to run for
+// each elasticsearch index to update.
+// After the tasks have been gather for all services, the workflow executes them
```
bcmdarroch:
```suggestion
// After the tasks have been gathered for all services, the workflow executes them
```

lib/authz/serialized_project_rules_update_workflow.go
```diff
@@ -0,0 +1,638 @@
...
+
+	iam_v2 "github.com/chef/automate/api/interservice/authz/v2"
+	"github.com/chef/automate/lib/cereal"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+var errCanceled = errors.New("Canceled")
+
+// SerializedProjectUpdateWorkflowExecutor is a workflow that allows running 1
```
bcmdarroch: we already had a type of workflow for serialized tasks, `SerializedWorkflowExecutor`, right? How is this one different?

lib/authz/serialized_project_update_workflow_test.go
```diff
@@ -0,0 +1,1260 @@
...
+					},
+					RemainingTasks: []serializedProjectUpdateTask{},
+				},
+			})
+
+		executor.OnTaskComplete(instance, ev)
+		instance.AssertFailed()
+	})
+
+	t.Run("fails if error returned", func(t *testing.T) {
```
bcmdarroch: what's the difference btwn this test and the one above? is it where the error occurs?

components/authz-service/server/v2/project_update_manager.go
```diff
@@ -74,36 +80,40 @@ func NewWorkflowExecutor() (*patterns.ChainWorkflowExecutor, error) {
...
-	   |  +------------+      P| +--------------------+ | |    +-----------------+
-	   |                      a|                        | |
-	   |                      r| +--------------------+ | |    +-----------------+
-	   |                      a| |Domain Service      | cereal |Start Update     |
-	   |                      l| |Update Workflow     <-+-+---->Update Status    |
-	   |                      l| |(ingest)            | | |    |           ingest|
-	   |                      e| +--------------------+ | |    +-----------------+
-	   |                      l|                        | |
-	   |                       | +--------------------+ | |    +-----------------+
+	   |  +------------+       +------------------------+ |    +-----------------+
```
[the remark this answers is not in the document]
jaym: https://asciiflow.com/

components/authz-service/server/v2/project_update_manager.go
jaym: there are 2 subworkflows: the serialized one and the nodemanager domain service one. both these are allowed to run in parallel

components/compliance-service/ingest/ingestic/ingestic.go
jaym: the previous logic ignored these values and freaked out when a task was completed but didn't have `created + deleted + updated = total`
We need to account for the other things because conflicts are normal and its possible that the script decides it doesn't want to change a doc, which is a noop (though none of our scripts noop)

components/compliance-service/ingest/ingestic/project_update.go
jaym: This is new code. Search for `UpdateProjectTags` for the old code

lib/authz/serialized_project_rules_update_workflow.go
jaym: You might be thinking of the chain workflow executor. The problem with that is you need to know exactly how many things are in a chain. Here, we want to gather a list of all the things to be done before doing them so we can report progress in a more granular fashion. Also, i never anticipated that we would do chain -> parallel -> chain in terms of gluing together workflows, so thats not a thing that would just work

components/authz-service/server/v2/project_update_manager.go
lancewf: Thank you for explaining it.

components/authz-service/server/v2/project_update_manager.go
```diff
@@ -114,18 +124,28 @@ func NewWorkflowExecutor() (*patterns.ChainWorkflowExecutor, error) {
...
-	domainSvcExecutor := patterns.NewParallelWorkflowExecutor(func(key string) (cereal.WorkflowExecutor, bool) {
+	parallelWorkflowExecutor := patterns.NewParallelWorkflowExecutor(func(key string) (cereal.WorkflowExecutor, bool) {
 		lock.Lock()
 		defer lock.Unlock()
 		if workflowExecutor, ok := workflowMap[key]; ok {
 			return workflowExecutor, true
 		}
+
+		var workflowExecutor cereal.WorkflowExecutor
+		if key == serialSubworkflowName {
```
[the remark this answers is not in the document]
jaym: This workflow is a parallel workflow. When we enqueue it, it will have subworkflows. Before, these were just the names of the services ("ingest", "nodemanager", "compliance"). Now, one of the subworkflows is the new workflow that does all the ES stuff in a serialized way, which is called `serialSubworkflowName`. This all just makes sure we use the correct logic for each subworkflow.

lib/authz/serialized_project_update_workflow_test.go
jaym: Just 2 different ways of indicating error. The run task is expected to return a status. You would probably return an error in the task status if you ran a thing synchronously and it ran to completion with a failing result. You would return an error if for example you couldn't talk to elasticsearch.

I'm thinking one day, we could retry this type of error as we don't know what happened, but we wouldn't retry the previous one

PR #3034, opened 2020-03-04
Use credentials from vault to talk to s3

jaym:


ryancragun:
We shouldn't _have_ to set this in the config. If the right environment variables are set then credential resolver should find them. If that's not working then we have a bug.

jaym:
i would expect systemd to clear out the environment

Issue #2993, opened 2020-03-02
provide date when calling /reporting/reports/id/:id api


rickmarry:
Currently, all elasticsearch indices are being searched when requesting a compliance report by id.  Using all indices really starts to get slow in direct proportion to the amount of days/scans that we have in our system.  

## To Reproduce
- load a2 with many days of many compliance reports
- call /reporting/reports/id API 

## Expected behavior
The response either takes a very long time to come back or it times out completely.
The response takes longer and longer as our number of scans increases.

rickmarry:
The solution to this is to send the time range in which to search for the report.  In many cases, this time range is available prior to when calls are to be made to this API.  

For instance:
**example 1:**
  From the UI, when we land on the nodes page in compliance, we search for nodes using whichever filter criteria are selected.  In response, we get back a summary listing of info for each node, including the latest run id and its respective time/date (end_time).  When the user clicks on the node they're interested in, we make the call to the backend and with the node_uuid:
1) request a list of all reports for the said node for the date range in the original filter, sorted by end_time desc (most recent ends up first in the retrieved list)
2) request the full report of that first report

Because we are always getting the most recent report and since we know the report's end_time ahead of time, we may be much more efficient in our request for it by simply providing the end_time with the request.

PR #3037, opened 2020-03-05
use date that when searching for compliance report

rickmarry:
This change will result in a much more performant retrieval of a single compliance report from automate.  In the past, we were searching across the entire Elasticsearch base when a request for a single report was received.  Two changes are included in this change:

1.  We use the start_time and end_time that is passed in with filters to compute a more narrow set of es indices to include in our search for reports.
2.  When we retrieve the list of run_ids by calling: /reporting/report-ids, we include the end_time on which each run_id actually occurred.  This will boost any retrieval of reports by providing the exact date of the time-series es index within which to look.

This performance issue is not apparent when we have only a few days worth of data and few runs within those days.  The performance degradation is in direct proportion to the number of scans and the number of days in elasticsearch.

this addresses the issue: #3070 
it partially addresses #2993 (a new PR is forthcoming that addresses the remainder of #2993)

 

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/compliance-service/reporting/relaxting/reports.go
```diff
@@ -100,44 +104,75 @@ func (backend ES2Backend) getNodeReportIdsFromTimeseries(esIndex string,
...
 			for _, hit := range topHits.Hits.Hits {
-				nodeReport[nodeID] = hit.Id
+				type endTimeSource struct {
+					EndTime time.Time `json:"end_time"`
+				}
+				var et endTimeSource
+				if hit.Source != nil {
+					err = json.Unmarshal(*hit.Source, &et)
+					if err == nil {
+
```
danielsdeleo: you meant to do some error handling here?

components/compliance-service/reporting/relaxting/reports.go
```diff
@@ -100,44 +104,75 @@ func (backend ES2Backend) getNodeReportIdsFromTimeseries(esIndex string,
...
+					}
+				}
+				endTimeTimestamp, _ := ptypes.TimestampProto(et.EndTime)
+
+				repData := reportingapi.ReportData{}
+				repData.Id = hit.Id
+				repData.EndTime = endTimeTimestamp
+
+				nodeReport[nodeID] = repData
```
danielsdeleo: since this function is now pushing 100 lines it might be nice to extract some chunks of it to private functions and/or move the type def out.

ec2/Vagrantfile
```diff
@@ -291,9 +291,9 @@ Vagrant.configure('2') do |config|
     aws.secret_access_key = "#{aws_secret_access_key}"
     aws.session_token = "#{aws_session_token}" if aws_session_token
     aws.keypair_name = ENV['AWS_SSH_KEY_NAME']
-    aws.instance_type = 'm5.large'      # 2CPU, 8GB RAM
+    #aws.instance_type = 'm5.large'      # 2CPU, 8GB RAM
     #aws.instance_type = 'm5.xlarge'    # 4CPU, 16GB RAM
-    #aws.instance_type = 'm5.2xlarge'   # 8CPU, 32GB RAM
+    aws.instance_type = 'm5.2xlarge'   # 8CPU, 32GB RAM
```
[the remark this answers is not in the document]
rickmarry: yes, thanks @danielsdeleo

components/compliance-service/reporting/relaxting/reports.go
rickmarry: yes.. still WIP.. getting to that cleanup now.. thanks

components/compliance-service/reporting/relaxting/reports.go
```diff
@@ -68,19 +68,23 @@ func (backend ES2Backend) getDocIdHits(esIndex string,
...
 	boolQuery := backend.getFiltersQuery(filters, latestOnly)
 
+	fsc := elastic.NewFetchSourceContext(true).Include("end_time")
+
 	// aggs
 	aggs := elastic.NewTermsAggregation().Field("node_uuid").Size(reporting.ESize).
 		SubAggregation("distinct", elastic.NewTopHitsAggregation().Size(1).
-			FetchSource(false).
+			FetchSourceContext(fsc).
+			//FetchSource(false).
```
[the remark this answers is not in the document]
rickmarry: removed

components/compliance-service/reporting/relaxting/reports.go
rickmarry: I moved that typedef out

components/compliance-service/reporting/relaxting/reports.go
```diff
@@ -100,44 +104,70 @@ func (backend ES2Backend) getNodeReportIdsFromTimeseries(esIndex string,
...
-				nodeReport[nodeID] = hit.Id
+				var et EndTimeSource
+				if hit.Source != nil {
+					err = json.Unmarshal(*hit.Source, &et)
+					if err != nil {
+						logrus.Errorf("could not get the end_time for runid: %s %v", hit.Id, err)
+						continue
+					}
+				}
+				endTimeTimestamp, _ := ptypes.TimestampProto(et.EndTime)
```
vjeffrey: error check?

vjeffrey:
@rickmarry i recall you mentioning a ui change that could happen once this pr merges to further speed things up, could you please write that up in an issue so we can take care of it?

vjeffrey:
here are screenshots of the timing on retrieval of a single node report from the dev perf env, before this change merges:
Screen Shot 2020-03-09 at 19 32 07
 
Screen Shot 2020-03-09 at 19 45 42

rickmarry:
> @rickmarry i recall you mentioning a ui change that could happen once this pr merges to further speed things up, could you please write that up in an issue so we can take care of it?

I spoke with Dan last night.. he and I are going to pair up to look at that one quickly today

components/compliance-service/reporting/relaxting/reports.go
rickmarry: I didn't do anything with this error because it marshaled without error in the line previously, so it's value won't be nil but, worst case, the zero value of go's "time.Time" which is the first moment in time (January 1, year 1, 00:00:00.000000000 UTC)... I figured that by continuing on here, we'd allow the user decide what to do with that.. for example, they could notice that in the case that the end_time is the zero value and then decide to look over a wide range, etc.. but at least they'd have the run_id.   I was torn about how and to handle this one.

Thanks for pointing it out.  I will print a log message when an error occurs here.

vjeffrey:
👍 let's do the ui change in a separate issue in the interest of getting this change into the next automate release

rickmarry:
> +1 let's do the ui change in a separate issue in the interest of getting this change into the next automate release

rickmarry:
closed by accident

rickmarry:
> here are screenshots of the timing on retrieval of a single node report from the dev perf env, before this change merges:
> Screen Shot 2020-03-09 at 19 32 07
> 
> Screen Shot 2020-03-09 at 19 45 42

not sure how dramatic the diff will be on this since that system only has about a months worth of indices atm

vjeffrey:
> not sure how dramatic the diff will be on this since that system only has about a months worth of indices atm

ya -- but there are some api report fetch calls there taking 5-6 seconds, so im hoping we'll see _some_ difference

vjeffrey:
i would say this is an improvement 🎉 
Screen Shot 2020-03-10 at 10 03 41

PR #3041, opened 2020-03-05
AWS instance updates

jaymalasinha:

Re-size and remove AWS instances

- [ ] Tests added/updated?
- [ ] Docs added/updated?


terraform/test-environments/scenarios/single_local_inplace_upgrade.tf
```diff
@@ -18,7 +18,7 @@ module "single_local_inplace_upgrade" {
   vpc            = "${var.environment}"
   platform       = "ubuntu-16.04"
   key_name       = "cd-infrastructure"
-  instance_type  = "m4.large"
+  instance_type  = "m5.large"
```
jaym: does this mean we will lose the data that is currently on these instances?

jaymalasinha: No, I already resized them to m5.large this morning during the maintenance window.  During the update I just restarted the instance, so you did not loose any data

vendor/github.com/aws/aws-sdk-go/service/ec2/api.go
```diff
@@ -70475,6 +70475,24 @@ const (
 	// InstanceTypeT22xlarge is a InstanceType enum value
 	InstanceTypeT22xlarge = "t2.2xlarge"
 
+	// InstanceTypeT3Micro is a InstanceType enum value
+	InstanceTypeT3Micro = "t3.micro"
```
stevendanna: Are these changes you made yourself or are they from the upstream aws/aws-sdk-go package? If the former, we need to chat about how we are going to manage these patches as we typically do not change code in the vendor directory.  If the latter, we will mant to make sure we get go.mod and go.sum updated to account for what version these are in.

jaymalasinha: I made those manually since I was changing the type to t3.micro.  I will revert this file.

PR #3047, opened 2020-03-06
Fix docker local tests to be compatible with newer versions of pg image

tylercloke:

Recently, the docker image's default security config was changed because it defaulted to super insecure settings:

https://github.com/docker-library/postgres/issues/580

We didn't care about that since we were only using it locally for testing. Updating the Makefiles for auth related services to work with newer versions of the docker postgres image. Also cleaned up some of authn's use of PG_URL so there was less inconsistency and hard-codedness.

Get on the latest docker image:

```
docker image ls
docker image rm -f <imageid>
```

In teams, authz, and authn services:

```
make setup_docker_pg
make test_with_db
make kill_docker_pg
```

- [x] Tests added/updated?
- [ ] Docs added/updated?


components/authn-service/tokens/integration/token_integration_test.go
```diff
@@ -61,14 +85,9 @@ func TestChefClientAuthn(t *testing.T) {
 		t.Fatalf("parse test upstream URL %+v: %v", upstream, err)
 	}
 
-	pgURLGiven := false
-	pgCfg := pg.Config{
-		PGURL:          constants.TestPgURL,
```
bcmdarroch: is this what was passing the `postgres` user? 🤔

tylercloke: Yep

components/authn-service/tokens/integration/token_integration_test.go
```diff
@@ -96,11 +115,11 @@ func TestChefClientAuthn(t *testing.T) {
 	serv, err := server.NewServer(ctx, config, authorizationClient)
 	if err != nil {
 		// SKIP these tests if there's no PG_URL given -- and never skip during CI!
-		if pgURLGiven || os.Getenv("CI") == "true" {
+		if pgCfg == nil {
 			t.Fatalf("opening connector: %s", err)
 		} else {
 			t.Logf("opening database: %s", err)
-			t.Logf(testconstants.SkipPGMessageFmt, pgCfg.PGURL)
+			t.Logf("failed to open test database with PG_URL: %q", pgCfg.PGURL)
```
blakestier: These subsequent logs seem confusing-- if we know we're going to log a failure, why log the attempt?

tylercloke: I think we are just logging the error and the URL it failed with?

tylercloke: Could be more clear though. Open to whatever.

components/authn-service/Makefile
```diff
@@ -50,11 +50,11 @@ test:
 
 test_with_db:
 	@docker ps | grep authn-postgres || (echo "Docker postgres not up. Run 'make setup_docker_pg' and try this command again."; exit 1)
-	@PG_URL="postgresql://postgres@127.0.0.1:5432/authn_test?sslmode=disable&timezone=UTC" go test -cover -count=1 -parallel=1 -p 1 $(packages)
+	@PGTZ=UTC PG_URL="postgresql://postgres@127.0.0.1:5432/authn_test?sslmode=disable&password=docker" go test -cover -count=1 -parallel=1 -p 1 $(packages)
```
blakestier: any idea what the timezone parameter was doing for us?

blakestier: Oh I see you've moved it above 🤔

components/authn-service/tokens/integration/token_integration_test.go
blakestier: We hit this line if the `PG_URL` wasn't provided, right? Can we log that and indicate that we are skipping pg tests (something more akin to the old message)?

blakestier: Basically I would be led astray by the word "failed"

bcmdarroch: I think L119 is where we're handling a nil PG_URL, so maybe we could change the messaging on that one.

bcmdarroch: Yeah same

components/authn-service/tokens/pg/testconstants/testconstants.go
```diff
@@ -1,27 +0,0 @@
-package testconstants
-
-// SkipPGMessageFmt is the friendly message the developer sees when any tests
-// requiring postgres are skipped because the connection failed.
-//
-// Yes those lines are long, but if we linebreak, the output ends up starting
-// with TAB characters, which makes it harder to copy-paste the commands.
-const SkipPGMessageFmt = `
```
bcmdarroch: Like @blakestier was saying, this is a much more actionable error message. Do we need to remove it?

tylercloke: don't think that's what we want people to do? this made sense when we weren't automating with the makefile.

PR #3051, opened 2020-03-06
Check stdout and stderr for errors in inspec detect

jaym:


components/automate-gateway/integration/nodes_test.go
```diff
@@ -173,7 +173,7 @@ func (suite *GatewayTestSuite) TestGatewayNodesClient() {
 		Name:            "test node",
 		Manager:         "automate",
 		Platform:        "redhat",
-		PlatformVersion: "7.5",
+		PlatformVersion: "7.7",
```
vjeffrey: 👍

components/compliance-service/inspec/cli.go
```diff
@@ -132,6 +132,13 @@ func run(args []string, conf *TargetConfig, timeout time.Duration, env map[strin
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
+	which, _ := exec.LookPath(args[0]) // nolint: errcheck
+	logrus.WithFields(logrus.Fields{
```
vjeffrey: we should remove this if we merge the pr

components/compliance-service/inspec/cli.go
```diff
@@ -234,64 +241,71 @@ func Archive(profilePath string, outputPath string) error {
 	return nil
 }
 
-func getInspecError(serr string, err error, target *TargetConfig, timeout time.Duration) *Error {
+func getInspecError(stdOut string, stdErr string, err error, target *TargetConfig, timeout time.Duration) *Error {
```
vjeffrey: this change makes me sad b/c it means that we're somehow getting the error in stdout instead of stderr.....but i also want the pipeline to get unblocked AND we're all very uncertain of how this started happening...

danielsdeleo: I found that config like this would force the logging to stderr:

vjeffrey: i...wonder if this has anything to do with the fact that we haven't changed our code to use reporter (instead of the original "format") ?https://github.com/chef/automate/blob/master/components/compliance-service/inspec/cli.go#L109

PR #3061, opened 2020-03-09
[go] Upgrade go to 1.14

ryancragun:
* Upgrade go to 1.14
* Upgrade the olivere package to the lastest version and use the
  gopkg.in address
* Fix aggregate parsing in config-mgmt and compliance where a raw json.Number
  was incorrectly casting to a string.
* Run goimports across the repo


components/compliance-service/reporting/relaxting/profiles_control_level.go
```diff
@@ -93,13 +94,11 @@ func (depth *ControlDepth) getProfileMinsFromNodesResults(
...
 					profileInfoBucket := profileInfoResult.Buckets[0]
-					name := profileInfoBucket.KeyNumber
-					summary.Name = string(name)
+					summary.Name = profileInfoBucket.Key.(string)
 
 					if profileShaResult, found := profileInfoBucket.Terms("sha"); found &&
 						len(profileShaResult.Buckets) > 0 {
-						sha := profileShaResult.Buckets[0].KeyNumber
-						summary.Id = string(sha)
+						summary.Id = profileShaResult.Buckets[0].Key.(string)
```
yzl: maybe a comment here?

ryancragun:
Yeah, it's because the `Key` field is an `interface{}`, therefore we have to cast it to a string. Fixing this tells me that having an elastic aggregator parsing library would probably be the ideal solution, but I think that's probably outside the scope of this PR.

PR #3067, opened 2020-03-09
upgrade cypress to latest version

bcmdarroch:

Cypress tests have been failing for a while in the [deploy pipeline](https://buildkite.com/chef/chef-automate-master-deploy-acceptance/builds/82#5a3cf97b-c9ed-4007-9681-8cb83a78dc5c/95-131)

After some digging, I found out we aren't pinning the Cypress version. So the docker image we're running the test on is [installing the latest version (v4.1.0)](https://github.com/chef/release-engineering/blob/master/components/dockerfiles/chefes/releng-base/Dockerfile#L191), but our tests expect v3.1.4.

### DIFF
--- a/.expeditor/buildkite/cypress.sh
+++ b/.expeditor/buildkite/cypress.sh
@@ -25,6 +25,8 @@ do
   export CYPRESS_BASE_URL="https://$instance"
   export CYPRESS_RECORD_KEY="$CYPRESS_RECORD_KEY"
 
+  npm install # get dependencies defined in e2e/package.json
+
  if ! npm run cypress:record; then
       buildkite-agent artifact upload "cypress/videos/*;cypress/videos/**/*;cypress/screenshots/*;cypress/screenshots/**/*"
       exit 1
--- a/.expeditor/buildkite/cypress.sh
+++ b/.expeditor/buildkite/cypress.sh
@@ -25,9 +25,7 @@ do
   export CYPRESS_BASE_URL="https://$instance"
   export CYPRESS_RECORD_KEY="$CYPRESS_RECORD_KEY"
 
-  npm install # get dependencies defined in e2e/package.json
-
-  if ! npm run cypress:record; then
+ if ! npm run cypress:record; then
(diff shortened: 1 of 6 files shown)


.expeditor/buildkite/cypress.sh
```diff
@@ -25,9 +25,7 @@ do
   export CYPRESS_BASE_URL="https://$instance"
   export CYPRESS_RECORD_KEY="$CYPRESS_RECORD_KEY"
 
-  npm install # get dependencies defined in e2e/package.json
```
msorens: Why is this step being deleted? I am surprised it would still work without this.

e2e/package-lock.json
```diff
@@ -53,7 +53,6 @@
       "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.5.0.tgz",
       "integrity": "sha512-1TTVrt7J9rcG5PMjvO7VEG3FrEoEJNHxumRq66GemPmzboLWtIjjcJgk8rokuAS7IiRSpgVSu5Vb9lc99iJkOA==",
       "dev": true,
-      "optional": true,
```
msorens: This is a problem; these `optional` properties were added recently with a package update PR (https://github.com/chef/automate/pull/2611).
You need to upgrade your version of node to current (12.13.1) then re-run the `npm install` to fix this.

.expeditor/buildkite/cypress.sh
bcmdarroch: This step is redundant. I think it was a stop-gap solution while releng updated the base docker image to install cypress and its dependencies.

Testing with the docker container locally, it passed without `npm install`.

```
# check out this branch
# run from automate/
docker run -it --rm --volume $(PWD):/workdir --workdir /workdir chefes/releng-base bash
export CHANNEL=acceptance
.expeditor/buildkite/cypress.sh

# should see output like this:
true
true
false
--- Executing Cypress tests against a2-perf-test-single-local-inplace-upgrade-acceptance.cd.chef.co
> automate-e2e@1.0.0 cypress:run /workdir/e2e
> cypress run
It looks like this is your first time using Cypress: 4.1.0
  ✔  Verified Cypress! /root/.cache/Cypress/4.1.0/Cypress
Opening Cypress...
```

PR #3069, opened 2020-03-10
Update builder

jaym:


jeremymv2:
To include the new UI changes for Creating a Package manipulating Package Visibility settings, we would need to toggle this [feature flag](https://github.com/habitat-sh/builder/blob/13cf6a6d89e0a7524207d5e6db9fec63ae66c185/components/builder-web/app/reducers/features.ts#L27). However, that enables more than we would want to for Onprem.. 

For now, I don't see any problems with the updates in the PR, just keeping in mind that it doesn't present much of a change in the UI experience as of right now - until we re-define the boundaries of the Onprem vs SaaS experience.

eeyun:
To add context - the `enable_builder` flag does more or less what it says - it enables all of the Builder related features that exist outside of depot functionality. Historically, this has included _any_ package settings as the on-prem depot does not allow for connecting plan files and the like. In the future this may change however for today, the content of this PR should be enough to get the up-to-date behaviours that would affect any on-prem users.

Issue #2995, opened 2020-03-02
check-in counts api


vjeffrey:
## User Story
In order to support the desktop offering view, we need to provide a "daily check in" count.
This means we need an API that returns:
- total count of desktops (nodes) that have checked in to Automate
- total count of desktops (nodes) that have checked in to Automate over the last 24 hours

## Considerations
Since this is useful information that can be used in views that are not desktop-only, we should make sure we design this API in a generic way.
Please refer to the epic description, q&a doc, and designs to validate requirements. Chat with @apriofrost  @chef/superteam about questions.
API endpoint should be discussed with team and agreed upon with UX and Product.
Let's try to keep most discussion/conclusions documented in this issue so we can reference later

https://chef.invisionapp.com/share/E9W3TEFXV87#/screens/406241157_Desktop_Dashboard-V2-Changes_Selected

vjeffrey:
NOTE: in the current designs, there is a view that provides count of nodes checked in per 24 hrs compared to total nodes; we may want to combine those two APIs? let's chat about it

vjeffrey:
```
request/
days ago: 1
response/
[ {start date: 2020-03-09[time], end date: 2020-03-10[time], checked in: 45, total: 50} ]
request/
days ago: 3
response/
[ 
    {start date: 2020-03-07[time], end date: 2020-03-08[time], checked in: 30, total: 40},
    {start date: 2020-03-08[time], end date: 2020-03-09[time], checked in: 40, total: 50} 
    {start date: 2020-03-09[time], end date: 2020-03-10[time], checked in: 45, total: 50} 
]
```
discussed in slack about doing a request/response like the above

lancewf:
**Option 1:**
**Daily Total:**
From the node-state index, for each day bucket find the number of nodes created before the end of the day. Subtract the number of nodes deleted before the end of the day. 

**Daily Check-in:**
From the converge-history indexes, get the number of unique nodes that have reports. 


If we can not find away to create aggregations to calculate these values with only two or three requests then this method would have a slow response. This would be flexible in the start and end dates of the day bucket. 


**Option 2:** 
Create a daily summary index. The day would start and end on UTC time. The index would store the total number of of nodes for each day and the total number of check-ins. This would be a faster response, but inflexalbe for the start and end of the day bucket. 


I am going to look into Option 1 first to find out if we can create aggregations to collect these values.

lancewf:
[...]

PR #3074, opened 2020-03-10
Adding date time filter to node count RPC

lancewf:
To provide support for the "daily check-in" count for the desktop offering a start and end date has been added to the NodesCounts RPC. The start and end date filters on the node's check-in DateTime. 

https://github.com/chef/automate/issues/2995

### :+1: Definition of Done
Have an API that provides the number of nodes that have checked in and the number of nodes that have check-in within the last 24 hours. 

to be added before going out of Draft. 

- [x] Tests added/updated?
- [ ] Docs added/updated?


api/external/cfgmgmt/request/stats.proto
```diff
@@ -6,6 +6,10 @@ option go_package = "github.com/chef/automate/api/external/cfgmgmt/request";
 message NodesCounts {
   // List of filters to be applied to the node count results.
   repeated string filter = 1;
+  // Earliest node checkin time to return for the node counts.
```
vjeffrey: i know this is still in draft, i'm just doing some early review..
can you add an example of the format of the time string?

vjeffrey: also: api requires us to say
- count of nodes that have checked in over the given time period
- total count of nodes that have ever checked in

will the API response, as is, return that information for us?  so in other words, is the total nodes count  returned unaffected by the time range? and then just the success/failure counts are affected by the time range?

lancewf: No, that is not how it is working. The start and end dates are not required. So, to get all the nodes check in ever, send a request with no start and end date. To get the number of nodes that have check in in the last 24 hours send a request with the start with a timestamp from 24 hours ago and the end with the timestamp from now. This would require two requests to get both. 
I will add the example date format to the proto.

PR #3084, opened 2020-03-12
Dont trace log requests to ES

jaym:
Nothing good can come of it


ryancragun:
When debugging goes wrong

PR #3091, opened 2020-03-13
[automate-3030] Enhance edits pending bar to show failure

msorens:

When a project update fails, we want to make that much more apparent to the user.
This replaces the standard "edits pending" message with an error message (and coloring) that makes it evident to the user.

The illustration shows the two possible states of the pending edits bar, showing both wording and color change, along with the subtle color change of the button itself when hovering or focusing it.

image

NA

### :+1: Definition of Done
Here is the contract as defined by unit tests. Most of these were already in place. The new tests are after the `-----` divider.
```
 update-start-confirmation modal...
   when it emits a cancellation, hides the modal
   when it emits a confirmation, hides the modal
   when it emits a confirmation, the projectService starts the rule updates
 pending edits bar...
   is hidden if no project has changes
   is visible if some project has changes
   is visible if some project has changes... UNLESS progress bar is active
   is visible if rules are not being applied
   is visible if update fails
   is visible if update is cancelled
   is visible if update is cancelled but update is still running
----
   displays edits pending message before update is applied
   displays edits pending message if an update is cancelled
   displays failed message if an update fails
   displays failed message if both an update fails and was cancelled
```

In hab studio: `rebuild components/authz-service`

Remove the last commit to expose a testing harness:
```
git revert -n 3df50d5
```
Then rebuild automate-ui per your preferred method.
Create a project and a rule in the browser.
Navigate to the projects page to expose the edits pending bar in its nominal condition (due to another issue, the edits pending bar will not start appearing until you visit that particular page).
Now just watch: every 5 seconds a random failure--or not--will occur. You should see the blue (normal) or pink (failure) bars both appear over time.

That shows the "normal" edits pending state--before an update is applied,
and the result of a failed update. To see the result of a cancelled update, first you have to reset the code--pull down a fresh copy of the branch so it no longer has that revert mentioned above. Then just proceed normally: start an update and cancel it. You will see the normal edits pending bar.

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/components/pending-edits-bar/pending-edits-bar.component.html
+++ b/components/automate-ui/src/app/components/pending-edits-bar/pending-edits-bar.component.html
@@ -1,8 +1,15 @@
 <div id="pending-edits-bar" [hidden]="isBarHidden">
-  <span class="info">Project edits pending, update <a [routerLink]="['/settings/projects']">projects</a> to apply changes.</span>
-  <app-authorized [allOf]="['/iam/v2/apply-rules', 'post']">
-    <button mat-button (click)="openConfirmUpdateStartModal()">Update Projects</button>
-  </app-authorized>
+  <div [className]="updateProjectsFailed ? 'failure-container' : 'normal-container'">
+    <span *ngIf="!updateProjectsFailed" class="info">
+      Project edits pending, update <a [routerLink]="['/settings/projects']">projects</a> to apply changes.
+    </span>
+    <span *ngIf="updateProjectsFailed" class="info">
+      Project update failed. Review system status from the command line with <pre>chef-automate status</pre> and try again.
+    </span>
+    <app-authorized [allOf]="['/iam/v2/apply-rules', 'post']">
+      <button mat-button (click)="openConfirmUpdateStartModal()">Update Projects</button>
+    </app-authorized>
+  </div>
 </div>
 
 <app-confirm-apply-start-modal
--- a/components/automate-ui/src/app/components/pending-edits-bar/pending-edits-bar.component.scss
+++ b/components/automate-ui/src/app/components/pending-edits-bar/pending-edits-bar.component.scss
@@ -3,8 +3,16 @@
 #pending-edits-bar {
   background-color: $chef-primary-bright;
   color: $chef-white;
-  padding: 16px 10px;
   position: relative;
+
+  .normal-container {
+    padding: 16px 10px;
+  }
+
+  .failure-container {
+    padding: 16px 10px;
+    background-color: $chef-critical;
+  }
 
   .info {
     font-size: 16px;
@@ -12,6 +20,11 @@
 
     a {
       color: inherit;
+    }
+
+    pre {
+      display: inline;
+      padding: 0;
     }
   }
 
@@ -23,8 +36,13 @@
     top: 0;
   }
 
-  button:hover,
-  button:focus {
+  .failure-container button:hover,
+  .failure-container button:focus {
+    background-color: $chef-critical-hover;
+  }
+
+  .normal-container button:hover,
+  .normal-container button:focus {
     background-color: $chef-primary-hover;
   }
 }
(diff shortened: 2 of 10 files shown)


components/automate-ui/src/app/components/pending-edits-bar/pending-edits-bar.component.html
```diff
@@ -1,8 +1,15 @@
...
-  <app-authorized [allOf]="['/iam/v2/apply-rules', 'post']">
-    <button mat-button (click)="openConfirmUpdateStartModal()">Update Projects</button>
-  </app-authorized>
+  <div [className]="updateProjectsFailed ? 'failure-container' : 'normal-container'">
+    <span *ngIf="!updateProjectsFailed" class="info">
+      Project edits pending, update <a [routerLink]="['/settings/projects']">projects</a> to apply changes.
+    </span>
+    <span *ngIf="updateProjectsFailed" class="info">
+      Project update failed. Review system status from the command line with <pre>chef-automate status</pre> and try again.
+    </span>
```
msorens: Unlike the separation of pending edits vs. update in progress, that really have different functionality, the distinction between pending edits and pending-edits-due-to-failed-update is really the same functionality. That is why I just made this component handle both variations rather than creating a new component (which would have addded significant overhead for its infrastructure in layout facade).
In fact, here we already have the variable we need to distinguish--this.updateProjectsFailed.

components/automate-ui/src/app/components/pending-edits-bar/pending-edits-bar.component.scss
```diff
@@ -3,16 +3,29 @@
...
-  padding: 16px 10px;
   position: relative;
 
+  .normal-container {
+    padding: 16px 10px;
+  }
+
+  .failure-container {
+    padding: 16px 10px;
+    background-color: $chef-critical;
```
msorens: These two are in a new child div of `pending-edits-bar`, which starts with the default blue background. If it is a normal-container, that does not change. But if it is a failure-container, that completely overlays the blue background with the pink background.

components/automate-ui/src/app/components/pending-edits-bar/pending-edits-bar.component.scss
```diff
@@ -23,8 +36,13 @@
...
-  button:hover,
-  button:focus {
+  .failure-container button:hover,
+  .failure-container button:focus {
+    background-color: $chef-critical-hover;
+  }
+
+  .normal-container button:hover,
+  .normal-container button:focus {
     background-color: $chef-primary-hover;
```
msorens: Besides the background color, we also want the hover and focus colors to adjust.

components/automate-ui/src/app/entities/projects/project.reducer.ts
```diff
@@ -126,14 +126,11 @@ export function projectEntityReducer(
...
-        mapKeys(key => camelCase(key),
-          {
-            ...action.payload,
-            percentage_complete: action.payload.percentage_complete * 100
-          }),
-        state);
+      const payload: GetApplyRulesStatusSuccessPayload = {
+        ...action.payload,
+        percentage_complete: action.payload.percentage_complete * 100
+      };
```
msorens: Putting this in a variable provides strong typing.

components/automate-ui/src/styles/_colors.scss
```diff
@@ -19,6 +19,7 @@ $science-blue: #0075DB;
 $fuchsia: #A424B4;
 $fuchsia-light: #FADAFA;
 $fuchsia-dark: #440F4B;
+$burgundy: #BA1E6A;
```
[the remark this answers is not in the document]
msorens: Nice! A search of automate-ui did not show that, but I see it in a different component, chef-ui-library. Updating...

msorens:
@susanev Displaying the failure banner on cancel-project-update was not a bug in this PR; it just exposed that bug that we had not known about in the service. I adjusted the authz-service to correct that.

PR #3098, opened 2020-03-13
fix failing project update test

bcmdarroch:
The `StartTime` and `LastUpdated` in `payload.json` times were just some hard-coded dates grabbed from the perf env. Once those dates passed, the project_update helper `FindSlowestJobStatus` started returning null time and failing the test. 

This fixes the issue by turning `payload.json` into a template and setting `StartTime` and `LastUpdated`, respectively, to the current time and current time + 1 min.

### DIFF
--- a/components/authz-service/server/v2/project_update_manager_test.go
+++ b/components/authz-service/server/v2/project_update_manager_test.go
@@ -1,10 +1,13 @@
 package v2
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"os"
 	"testing"
+	"text/template"
+	"time"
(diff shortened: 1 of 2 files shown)


components/authz-service/server/v2/project_update_manager_test.go
```diff
@@ -46,7 +49,25 @@ func loadWorkflowFromDisk(t *testing.T, name string) *workflowInstance {
 	} else {
 		defer fPayload.Close()
```
msorens: You should get rid of all occurrences of `fPayload` which is no longer being used.

PR #3100, opened 2020-03-15
[automate-3031] Show pending edits bar promptly upon login

msorens:
We need the pending edits bar to be present promptly upon page refresh or upon login. Before this fix, it would only start appearing upon visiting selected pages (e.g. project list page), which is any page that did a `GetProjects` call. There was seemingly a call to `GetProjects` in the LayoutFacadeService but therein lies the problem! That call was in an `ngOnInit` function but that `ngOnInit` hook was never being called! That is because LayoutFacadeService is neither a Component nor a Directive (reference: Reference: https://angular.io/guide/lifecycle-hooks). Moving that to the constructor made the code actually run, resolving the issue.
The `noOnDestroy` hook code was never used either, for the same reason, so that is removed here as well.

Finally, the very same things were true of the LayoutSidebarService, so changes for that as well.

#3103 (Fixes from auditing Angular lifecycle hooks)

### :+1: Definition of Done
Pending edits bar now appears immediately upon login or page refresh,
and remains present throughout a session.

Setup: rebuild automate-ui, login, and create a project and a rule. Observe the pending edits bar appears, as it did previously. 
Now go to a non-auth page, e.g. events feed or compliance. Then logout and log back in. You should see the pending edits bar immediately.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/components/pending-edits-bar/pending-edits-bar.component.spec.ts
+++ b/components/automate-ui/src/app/components/pending-edits-bar/pending-edits-bar.component.spec.ts
@@ -5,18 +5,15 @@ import { ReactiveFormsModule } from '@angular/forms';
 import { StoreModule, Store } from '@ngrx/store';
 import { MockComponent } from 'ng2-mock-component';
 
-import { NgrxStateAtom, runtimeChecks } from 'app/ngrx.reducers';
+import { NgrxStateAtom, runtimeChecks, ngrxReducers } from 'app/ngrx.reducers';
 import { ChefPipesModule } from 'app/pipes/chef-pipes.module';
 import { using } from 'app/testing/spec-helpers';
 import { FeatureFlagsService } from 'app/services/feature-flags/feature-flags.service';
-import { notificationEntityReducer } from 'app/entities/notifications/notification.reducer';
-import { clientRunsEntityReducer } from 'app/entities/client-runs/client-runs.reducer';
 import { GetIamVersionSuccess } from 'app/entities/policies/policy.actions';
-import { policyEntityReducer } from 'app/entities/policies/policy.reducer';
 import { Project } from 'app/entities/projects/project.model';
 import { ProjectStatus } from 'app/entities/rules/rule.model';
 import { ProjectService } from 'app/entities/projects/project.service';
-import { projectEntityReducer, ApplyRulesStatusState } from 'app/entities/projects/project.reducer';
+import { ApplyRulesStatusState } from 'app/entities/projects/project.reducer';
 import {
   GetProjectsSuccess,
   GetApplyRulesStatusSuccess,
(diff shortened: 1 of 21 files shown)


components/automate-ui/src/app/entities/layout/layout.facade.ts
```diff
@@ -65,9 +64,6 @@ export class LayoutFacadeService implements OnInit, OnDestroy {
     this.sidebar$ = store.select(sidebar);
     this.showPageLoading$ = store.select(showPageLoading);
     this.updateDisplay();
-  }
-
-  ngOnInit(): void {
```
msorens: This is the crux: getting rid of `ngOnInit` because this class is not a Component nor a Directive, so that `OnInit` hook is completely ignored. Moved it into the constructor instead so it gets executed.

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
```diff
@@ -243,15 +243,6 @@ export class LayoutSidebarService implements OnInit, OnDestroy {
       this.sidebars = this.updateVisibleValues(sidebars);
     }
 
-    ngOnInit() {
-      this.updateSidebars();
-    }
```
msorens: Bonus to this PR: not related to the bug being addressed, but for the same reasons this code would never be executed, so applied the same fix.

components/automate-ui/src/app/entities/policies/policy.selectors.ts
```diff
@@ -16,9 +16,10 @@ export const iamMajorVersion = createSelector(
   (state) => state.iamMajorVersion
 );
 
+// This is used early enough that we have to guard against undefined (at least for unit tests)
 export const isIAMv2 = createSelector(
   policyState,
-  (state) => (state.iamMajorVersion === 'v2')
+  (state) => (state && state.iamMajorVersion === 'v2')
```
msorens: The fixes above somehow caused this to fail a couple unit tests with state being undefined here.

components/automate-ui/src/app/components/pending-edits-bar/pending-edits-bar.component.spec.ts
```diff
@@ -79,13 +76,7 @@ describe('PendingEditsBarComponent', () => {
...
         RouterTestingModule,
         ChefPipesModule,
-        StoreModule.forRoot(
-          {
-          policies: policyEntityReducer,
-          projects: projectEntityReducer,
-          notifications: notificationEntityReducer, // not used here but needed to suppress warnings
-          clientRunsEntity: clientRunsEntityReducer // not used here but needed to suppress warnings
-        }, { runtimeChecks })
+        StoreModule.forRoot(ngrxReducers, { runtimeChecks })
```
msorens: All the 18 other files have this change: moving from specific reducers being called out to using the master collection en masse, because the above fixes caused hundreds of warnings when running the unit tests from the command line--see the individual commit for details.

bcmdarroch: Thanks for this cleanup! So much cleaner now ⭐️

PR #3104, opened 2020-03-15
Robust Deprecation of IAM v1

kagarmoe:

This PR makes the IAM v1 deprecation more robust. The intention is two-fold: To guide users to the correct IAM documentation with ease, and to reduce future docs work. 

* Strengthens deprecation notices
* Adds (Deprecated) to the v1 title
* Renames "IAM v2 Overview" to  "IAM Overview" 
* Renames "IAM v2 Guide" to "IAM Guide"
* Fix references to these pages 
* Fixes a few broken links that I discovered along the way

The pages are retitled, but the urls are unchanged.

### DIFF
--- a/components/automate-chef-io/content/docs/iam-v1-overview.md
+++ b/components/automate-chef-io/content/docs/iam-v1-overview.md
@@ -16,7 +16,7 @@ toc = true
 +++
 
 {{< warning >}}
-This documentation and the feature it covers, IAM v1, are deprecated. This documentation will be removed soon. Most Chef Automate installations upgraded automatically to the current IAM implementation. Contact your customer service representative for help upgrading airgapped installations.
+This documentation and the feature it covers, IAM v1, are deprecated. This documentation will be removed soon. Most Chef Automate installations upgraded automatically to the current IAM implementation. The [Airgap Installation documentation {{< relref "airgapped-installation.md/#upgrades" >}} covers the manual upgrade process for systems without internet connectivity.
 The current IAM documentation pages are: the [IAM Overview]({{< relref "iam-v2-overview" >}}), the [IAM Guide]({{< relref "iam-v2-guide" >}}) and the [Chef Automate API]({{< relref "docs/api/#tag/teams" >}}).
 {{< /warning >}}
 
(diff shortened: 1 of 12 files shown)


components/automate-chef-io/content/docs/iam-v1-overview.md
```diff
@@ -47,7 +48,7 @@ Policies can contain wildcards to match a range of values.
 
 ## Structure
 
-A policy is made up of three essential components: the **subjects**, which is
+A policy has three essential components: the **subjects**, which is
```
stevendanna: Does a policy have more than these 3 components? If not then perhaps:
```suggestion
A policy has three components: the **subjects**, which is
```

components/automate-chef-io/content/docs/iam-v1-overview.md
```diff
@@ -16,10 +16,11 @@ toc = true
 +++
 
 {{< warning >}}
-IAM v1 is deprecated and we recommend updating to the latest version of Automate to make use of [IAM v2]({{< relref "iam-v2-overview" >}}).
+This documentation and the feature it covers, IAM v1, are deprecated. This documentation will be removed soon. Most Chef Automate installations upgraded automatically to the current IAM implementation. Contact your customer service representative for help upgrading airgapped installations.
```
stevendanna: This bit leaves me with a couple questions. These questions are probably more for engineering, but I think we should try to answer them and include them here:

- Do we know the version that the force upgrade will happen in? Many users do not have automatic upgrades turned on.  If I was an Automate users, I'd like a sentence here that was a bit more precise about what version this document applies to and how I check which one I need.

- Why do airgapped installations need to contact someone? We have an upgrade procedure for airgapped upgrades that users should be able to follow without assistance.  Are we anticipating problems here?

susanev: >Do we know the version that the force upgrade will happen in

Not yet, but we can add the version when we do.

components/automate-chef-io/content/docs/iam-v2-overview.md
```diff
@@ -10,7 +10,7 @@ toc = true
     weight = 10
 +++
 
-## Features in IAM v2
+## Features of IAM
```
mjingle: @kagarmoe We're going to have a merge conflict here. This header was already removed in 3033. :)

components/automate-chef-io/content/docs/iam-v1-overview.md
kagarmoe: refined phrasing based on @stevendanna's feedback.

components/automate-chef-io/content/docs/iam-v1-overview.md
kagarmoe: done

Issue #2640, opened 2020-01-15
ssm scan nodes coming in without a node name


vjeffrey:
## Describe the bug
When running a ssm scan job, the nodes scanned are coming in without a node name on the report.

## To Reproduce
- scan some nodes via ssm (add aws integration for ec2, ensure some nodes have ssm access enabled, run a scan job)

## Expected behavior
Node has a name

## Versions (please complete the following information):
- Automate Build Number 20200115001116

vjeffrey:
confirmed in fresh-install env:
![Screen Shot 2020-03-21 at 18.22.26.png](https://images.zenhubusercontent.com/5cfebc28d99ac71fbfbdcb73/bb660399-9293-465b-a6a3-bc7051b6f4ec)

PR #3110, opened 2020-03-16
bugfix: ensure node name is correctly assigned when running ssm scan

vjeffrey:
I noticed there was a bug with ssm jobs during some acceptance testing -- all nodes showed up nameless after an ssm scan.

Figured out we had some faulty logic in resolver.go -- we were looking for a key value to determine the name and completely ignoring the name of the instance as we had discovered it.
 
This code change fixes that!

https://github.com/chef/automate/issues/2640

### :+1: Definition of Done
Cloud scans of ec2 result in nodes with names!

rebuild compliance
run an ssm scan of the nodes (via aws or azure)
* this requires running the automate instance in aws, or having azure creds

- [x] Tests added/updated?
- [x] Docs added/updated?

### DIFF
--- a/components/compliance-service/inspec-agent/resolver/resolver.go
+++ b/components/compliance-service/inspec-agent/resolver/resolver.go
@@ -485,7 +485,7 @@ func (r *Resolver) handleAzureVmNodes(ctx context.Context, m *manager.NodeManage
 	return r.handleManagerNodes(ctx, m, nodeCollections, job)
 }
 
-func handleNodeInfo(node *manager.ManagerNode) nodeInfo {
+func nodeInfoFromManagerNode(node *manager.ManagerNode) nodeInfo {
 	var nodeDetails nodeInfo
 	if len(node.Name) == 0 {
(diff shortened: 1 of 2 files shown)


components/compliance-service/inspec-agent/resolver/resolver.go
```diff
@@ -485,17 +485,28 @@ func (r *Resolver) handleAzureVmNodes(ctx context.Context, m *manager.NodeManage
 	return r.handleManagerNodes(ctx, m, nodeCollections, job)
 }
 
-func (r *Resolver) handleInstanceCredentials(ctx context.Context, instanceCreds []*manager.CredentialsByTags, node *manager.ManagerNode) ([]*inspec.Secrets, nodeInfo) {
-	var nodeInfo nodeInfo
+func handleNodeInfo(node *manager.ManagerNode) nodeInfo {
```
stevendanna: I know we use `handle` in many of the other function names, but I'm wondering if `nodeInfoFromManagerNode` would be a more clear name?

PR #3141, opened 2020-03-18
FormControlDirective fix

msorens:

The purpose of the `FormControlDirective` is to notice when a field's value matches its original value, thereby allowing a form to recognize that the value has not changed and take an appropriate action, typically contributing to disabling the <kbd>Save</kbd> button. 

I discovered a case where this was not working correctly. If one loads a page containing an `<input>` element governed by `FormControlDirective`,
the initial "original value" will be set to null. (In other words, the directive is being triggered too soon.) This PR allows capturing the slightly delayed "original value" so the control works more reasonably, recognizing as the "original value" the first non-empty value.

This worked fine if you were navigating within the app, which would be occurring most of the time. The fix here is only needed when refreshing the browser page.

### :+1: Definition of Done
Original field value is recognized whether navigating within the app or refreshing a page.

Rebuild automate-ui.

#### Baseline Case
Navigate to some detail page within the auth pages, be it project details, team details, token details, etc, by way of Settings >> Projects >> `<project id>`.
Select the `Details` tab. Notice the <kbd>Save</kbd> button is disabled because there is nothing to save.
Say you had a project `proj1`. If you append a `2` so it now reads `proj12`, this will enable the <kbd>Save</kbd> button.
Now, press <kbd>Backspace</kbd> so it again reads `proj1`. Notice the <kbd>Save</kbd> button is disabled again. That is the previously working case.

#### New Case
Now refresh the details page.
You should again be at `proj1` with the <kbd>Save</kbd> button disabled.
Now repeat the exercise of changing the value then returning the value to `proj1` and you should see it enable then disable the <kbd>Save</kbd> button, respectively.

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/components/form-control/form-control.directive.ts
+++ b/components/automate-ui/src/app/components/form-control/form-control.directive.ts
@@ -5,6 +5,13 @@ import { distinctUntilChanged, takeUntil } from 'rxjs/operators';
 
 // Getting the most from an Angular form
 // *************************************
+//
+// WHAT YOU GET BY JUST ADDING 'formControlName' TO YOUR CONTROL
+//
+// Instrument a control by adding the directive with the form field name. Example:
+//
+//     <input chefInput name="firstName" formControlName="firstName" type="text" autofocus>
+//
 // Out of the box, this directive automatically provides "memory" to your form
 // controls (typically <chef-input>, <chef-radio>, and <chef-select> elements)
 // when a focussed control is being edited.
@@ -17,6 +24,9 @@ import { distinctUntilChanged, takeUntil } from 'rxjs/operators';
 // If, upon a successful operation, the user is taken elsewhere
 // (e.g. the modal closes, or the page navigates to a summary, etc.)
 // nothing further is required.
+//
+// ADDITIONAL INSTRUMENTATION NEEDED IF YOU HAVE RELEVANT STATE CHANGES
+//
 // If, however, the form (or modal) remains open, you typically want
 // to have the "pristine state" of the control be updated
 // to the newly processed value. Here's how to make that happen:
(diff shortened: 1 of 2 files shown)


components/automate-ui/src/app/components/form-control/form-control.directive.ts
```diff
@@ -72,6 +72,13 @@ export class FormControlDirective implements OnInit, OnDestroy, OnChanges {
         distinctUntilChanged()
       )
       .subscribe(newValue => {
+        // If null, undefined, or empty string, we came in to early last time.
```
blakestier:
```suggestion
        // If null, undefined, or empty string, we came in too early last time.
```

components/automate-ui/src/app/components/form-control/form-control.directive.ts
```diff
@@ -72,6 +72,13 @@ export class FormControlDirective implements OnInit, OnDestroy, OnChanges {
...
+        // So take this value as the originalValue.
+        if (!this.originalValue) {
+          this.originalValue = newValue;
+        }
```
blakestier: Should this  be an `else`, or a single conditional? When I read this, it looks like we first check for the original value. If it's not present, we set the original value to the new value . . . and then we immediately check if the new value is equal to the original value.

msorens: Good question! They need to be separate conditionals. Here is why: say we come in with the first non-empty value after having come in earlier with an empty value. We need to first set the originalValue to this first non-empty value, then perform the normal check of seeing if the new value === original value. Of course, it will always be true at that point, so we do a `reset()`--which is vital to do at that point. But now consider values that come in later. We will no longer adjust the originalValue, but we will continue to check if the new value === original value.

PR #3142, opened 2020-03-18
Refetch rules when project becomes applied

msorens:

If you are on the Projects List page and invoke a Project Update, the projects marked "pending" get promptly updated to "applied" because we re-fetch the projects in a timely fashion. The same thing should happen if you happen to be on the Ingest Rules tab of a project details page at the time you invoke a Project Update. Prior to thi PR, you had to do either a page refresh or re-navigate to the rules page to instigate that update. 

NA

### :+1: Definition of Done
Upon invoking "Update Projects" when displaying an edited list of project rules, the status promptly changes from "pending" to "applied".

Rebuild automate-ui
Create a project.
Create a rule.
Upon creating the rule you should be sitting back on the list of rules page for the project.
Invoke "Update Projects".
Observe that the page updates the status of the edited rule(s).

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/pages/project/details/project-details.component.ts
+++ b/components/automate-ui/src/app/pages/project/details/project-details.component.ts
@@ -3,7 +3,7 @@ import { FormBuilder, FormGroup, Validators } from '@angular/forms';
 import { Router } from '@angular/router';
 import { Store } from '@ngrx/store';
 import { Subject, combineLatest } from 'rxjs';
-import { filter, pluck, map, takeUntil, distinctUntilChanged } from 'rxjs/operators';
+import { filter, pluck, takeUntil, distinctUntilChanged } from 'rxjs/operators';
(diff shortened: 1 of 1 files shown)


components/automate-ui/src/app/pages/project/details/project-details.component.ts
```diff
@@ -132,6 +132,16 @@ export class ProjectDetailsComponent implements OnInit, OnDestroy {
       .subscribe(() => {
         this.store.dispatch(new GetProject({ id: this.id }));
       });
+
+    // when project becomes applied need to also mark rules applied
+    this.store.select(allProjects).pipe(
```
bcmdarroch: is there a reason we have to check all projects in the store instead of checking the current project?

msorens: Good observation! Changed.

PR #3148, opened 2020-03-18
[automate-2710] Remove system:* perms, modify infra:* perms

blakestier:

System actions are intended for internal use and are needed for Automate to work correctly (i.e. must be able to check the license status, the IAM version, etc. for the UI to render things correctly). They are not related to permissions a user might perform on Automate resources (i.e. create a user, delete a missing node). To keep system policies hidden from end users, we load them directly into OPA cache, keeping them out of policy database.

Since system actions are not intended for end users to use in their policies, actions like `system:*:list`, `system:*:get` are not needed on the viewer role

The only thing the editor gains with the `system:*` is the ability to apply a license, but we want that to be admin-only anyway.

We need to adjust our editor and viewer roles to be more explicit about what is allowed under infra because chef servers is getting added.

SCOPE CREEP:

In the making of this PR, we noticed 1) Roles were only accidentally getting access to telemetry due to the `system:*` action (the existing `telemetry:*` action was retired w/ v2 and 2) Ppl outside of Admin had the capacity to request/apply license and see the gateway endpoint due to `system:*`. . . we've tightened those perms here.

Ideally, we have a follow-up to move telemetry and license _out_ of the system namespace, but this gives us the desired functionality.

### :+1: Definition of Done

- [x]  system prefixed actions are removed from Chef-managed roles
- [x]  Update documentation, specifically..
- [x]  iam-v2-overview/#role-based-access-control: note that editors cant apply a license
- [x]  iam-v2-api-reference/#default-chef-managed-roles: update the table
- [x]  in the editor role, change `infra:*` to `infra:nodes:*, infra:nodeManagers:*`
- [x]  in the viewer role, change infra:*:get, infra:*:list to infra:nodes:get, infra:nodes:list, infra:nodeManagers:get, infra:nodeManagers:list
- [x]  change the infra legacy policy to migrate as infra:nodes:* and infra:nodeManagers:* instead of infra:*
- [x] only admin have access to requesting/applying licenses
- [x] all users have access to the telemetry endpoint

1. `start_all_services`
2.  force-upgrade: `rebuild components/authz-service/ && rebuild components/automate-cli && rebuild components/automate-deployment/ && rebuild components/automate-gateway && rebuild components/teams-service/`
3. `export TOK=chef-automate iam token create adm --admin`
4. curl updated legacy policy:

```
curl -kH "api-token: $TOK"  https://localhost/apis/iam/v2/policies/infrastructure-automation-access-legacy | jq

{
  "policy": {
    "name": "[Legacy] Infrastructure Automation Access",
    "id": "infrastructure-automation-access-legacy",
    "type": "CUSTOM",
    "members": [
      "user:*"
    ],
    "statements": [
      {
        "effect": "ALLOW",
        "actions": [
          "infra:nodes:*",
          "infra:nodeManagers:*"
        ],
        "role": "",
        "resources": [
          "*"
        ],
        "projects": [
          "*"
        ]
      }
    ],
    "projects": []
  }
}
```

the `'infrastructure-automation-access-legacy` policy no longer has `infra:*` statement actions (but `infra:nodes:*` and `infra:nodeManagers:*` instead).

Builds:
[Verify Private](https://buildkite.com/chef/chef-automate-master-verify-private/builds/10598)

[Verify](https://buildkite.com/chef-oss/chef-automate-master-verify/builds/10289#8b70a6fd-16eb-481d-a07b-ab688b3611cc)


components/authz-service/storage/postgres/migration/sql/75_update_roles_pols.up.sql
```diff
@@ -0,0 +1,71 @@
...
+        actions = '{
+            secrets:*:get,
+            secrets:*:list,
+            infra:nodes:get,
+            infra:nodes:list,
+            infra:nodeManagers:get,
+            infra:nodeManagers:list,
+            compliance:*:get,
+            compliance:*:list,
+            system:*:get,
```
bcmdarroch: i thought we were removing `system` actions from these roles?

components/authz-service/server/v2/system.go
```diff
@@ -48,6 +48,12 @@ func SystemPolicies() []*storage.Policy {
 				Resources: []string{"system:service:version"},
 				Projects:  []string{constants.AllProjectsID},
 			},
+			{
+				Effect:    storage.Allow,
+				Actions:   []string{"system:telemetryConfig:get"},
```
blakestier: everyone should have this access, and we were inadvertently denying it to all users, and providing it to certain roles by accident due to the `system:*` namespace.

integration/tests/iam_v2_no_legacy_to_v2_force_upgrade.sh
```diff
@@ -15,7 +15,7 @@ test_upgrade_strategy="none"
 
 # a2-iam-no-legacy-integration verifies permissions on an IAM v2 system 
 # without v1 legacy policies
-test_deploy_inspec_profiles=(a2-iam-no-legacy-integration)
+test_deploy_inspec_profiles=(a2-deploy-smoke)
```
blakestier: expected functionality for v2 has changed, so the `a2-iam-no-legacy-integration` won't pass on both the pre and post upgrade. Smoke tests ensure the pre-upgrade is running.

components/authz-service/storage/postgres/migration/sql/75_update_roles_pols.up.sql
```diff
@@ -0,0 +1,78 @@
...
+            infra:nodeManagers:get,
+            infra:nodeManagers:list,
+            compliance:*:get,
+            compliance:*:list,
+            event:*:get,
+            event:*:list,
+            ingest:*:get,
+            ingest:*:list,
+            iam:projects:list,
+            iam:projects:get
```
msorens: I think you missed `applications` for viewer (you have it correctly for editor below):
https://github.com/chef/automate/blob/7392f96f1e05b6009d38cd903436663d6d95c1d0/components/authz-service/storage/postgres/datamigration/sql/07_update_v2_roles.up.sql#L5

components/authz-service/storage/postgres/migration/sql/75_update_roles_pols.up.sql
```diff
@@ -0,0 +1,78 @@
...
+            iam:projects:get,
+            iam:projects:assign,
+            iam:policies:list,
+            iam:policies:get,
+            iam:policyMembers:*,
+            iam:teams:list,
+            iam:teams:get,
+            iam:teamUsers:*,
+            iam:users:get,
+            iam:users:list
```
msorens: If the docs are valid, project owner is supposed to be a superset of editor, so it also needs  `applications:*`.

components/authz-service/storage/postgres/migration/sql/75_update_roles_pols.up.sql
```diff
@@ -0,0 +1,81 @@
...
+            iam:teams:get,
+            iam:teamUsers:*,
+            iam:users:get,
+            iam:users:list,
+            applications:*
+        }'
+    WHERE
+        id = 'project-owner';
+
+UPDATE iam_statements
```
bcmdarroch: Even though it's not necessary with this migration, could we change the [migration code](https://github.com/chef/automate/blob/force-upgrade/components/authz-service/storage/postgres/migration/legacy/iam.go#L206) to reflect the correct migration for reference?

blakestier: I'd prefer not to *just* change the migration for the policy in the migration code and not also change the roles, and keep this migration.

If it's really clutch that try to preserve the state of the world in the `force-upgrade` branch, I'd rather modify the migration code for policies AND roles and associated tests, and the datamigrations, and not add an additional SQL migration. Or I can pull this work out and point it to master, and then we'll need to be careful to fix the things in `force-upgrade`. 

FWIW, this is why I didn't just change the migration like we'd discussed (these mostly assume that if I change the policy in the migration code, I should also change the roles (if the reason is to preserve the state of the world))

1- since this is a *post* migration migration (not just a data migration), then it's kind of weird that we're changing both-- if I'm strictly looking at the code in the future, I'm like "why do we have this migration if all the migration code already handles this?"

2- The other thing is that the datamigrations will *overwrite* the migration code-- so we'll create the roles in the v2 upgrade, then overwrite them in the datamigrations, then overwrite them _again_ in the post v2 migrations. Not a huge deal, but feels off. 

3- this change isn't really immediately related to the force upgrade-- I've put it here only bc we'd have to do a few modifications to `force-upgrade` in order to make it happy with this change. 

I agree that it's annoying that we don't have a single source of truth, but I think that this will only be a source of truth for a moment in time, and we'll need to reference the sql migrations after that. 

What do you think?

tylercloke: If there's not a functional reason to update the migration code, I'm a :bigfan: of only ever touching that code again if necessary. It's a lot of high risk code we don't wanna maintain long-term.

bcmdarroch: sounds good

susanev:
there is a do not merge label on this pr because its currently pointed at master
it needs to merge into the force-upgrade branch

Issue #2601, opened 2020-01-02
[ec2/vagrant setup] Update Vagrantfile according to new license model


vsingh-msys:
## Describe the bug
Seems that we are still using the old license model in order to set up the license for Automate dev setup.
https://github.com/chef/automate/blob/ff428ded612930645b9772511012017406dd4ad7/ec2/Vagrantfile#L13-L14

Also minor refactor required in order to setup habitat actual auth token.

## To Reproduce
- Follow the steps given in https://github.com/chef/automate/blob/master/ec2/README.md
- while running `vagrant up`

Error logs:
````
There was an error loading a Vagrantfile. The file being loaded
and the error message are shown below. This is usually caused by
a syntax error.

Path: /dev/automate/ec2/Vagrantfile
Line number: 0
Message: RuntimeError: A valid Automate JWT license is required. Provide the content in ENV variable AUTOMATE_LICENSE or save it in this file: ../dev/license.jwt
````

## Expected behavior
 it should created a desired EC2 instance with Automate dev setup.

## Versions (please complete the following information):
- OS: ubuntu 18
- Automate: master branch

PR #3151, opened 2020-03-19
Fix and improve ec2/Vagrantfile

alexpop:
Also adds an experimental `ssh/Vagrantfile` in the `ec2` directory to be able to setup Automate on a remote Linux server.

### DIFF
--- a/ec2/Vagrantfile
+++ b/ec2/Vagrantfile
@@ -229,6 +229,7 @@ else
 fi
 
 cat<<EOF >>dev/config.toml
+
 [compliance.v1.sys.logger]
   level = "debug"
 
--- a/ec2/Vagrantfile
+++ b/ec2/Vagrantfile
@@ -255,7 +255,6 @@ if [ "\\$USER" == "ubuntu" ]; then
     mkdir -p ~/.hab/etc
     cat<<'EOT' > ~/.hab/etc/cli.toml
 origin = "ubuntu"
-auth_token = "dev444BBB999"
 ctl_secret = "dev333ZZZ111"
 EOT
     echo " * Running: hab origin..."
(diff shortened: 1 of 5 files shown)


vsingh-msys:
Probably this PR will fix https://github.com/chef/automate/issues/2601

Thanks @alexpop

alexpop:
Yes Vivek, it fixes the setup after recent changes in Automate. You would still need to have an Automate license either in the `ENV['AUTOMATE_LICENSE']` variable or as a file `dev/license.jwt` in the repo.

ec2/Vagrantfile
```diff
@@ -109,6 +109,10 @@ if ARGV[0] == "up"
     stop_hours = ENV['STOP_HOURS']
   end
 
+  if ENV['ES_HEAP_SIZE'].nil?
+    ENV['ES_HEAP_SIZE'] = "3"
```
rickmarry: we changed this to 3 just so that we could see that it actually changed when we were testing.. now that we know it works, should we change it to 2 instead of 3?

ec2/ssh/Vagrantfile
```diff
@@ -0,0 +1,222 @@
+require 'json'
```
rickmarry: why do we need this ssh/Vagrantfile?

ec2/Vagrantfile
alexpop: Was thinking to leave it to 3gb since we recommend 8gb or larger for the setup. And will avoid changing from 2 to 2 :)

ec2/ssh/Vagrantfile
alexpop: as a backup or for anyone that's using some other type of cloud (Azure, Tata, GCP, etc). It's a generic linux setup

rickmarry: sweet!

ec2/Vagrantfile
rickmarry: cool

rickmarry:
cool.. thanks @alexpop!

PR #2157, opened 2019-11-07
[data-lifecyle] add data-lifecycle API

ryancragun:
Expose data lifecycle configuration, status, and run hooks in the gateway.

We need to expose the data lifecycle interfaces in the gateway so that we can configure them in the UI. During the design we noticed that the existing lifecycle page required 4 API calls to get the configuration and 4 to set it. Rather than build another ad-hoc API to provide the purge interfaces to the UI we determined that it would be better to build a top-level `data-lifecycle` resource and to move the complexity into the gateway instead of the UI.

The current shape of the API is as follows:

```
/data-lifecycle
  /status # GET: gets aggregate configuration along with other status metadata
  /config # PUT: updates aggregate configuration
  /run # POST: run aggregate lifecycle operations
  /infra
    /status # GET: gets configuration and status for infra lifecycle operation
    /config # PUT: update the infra lifecycle config
    /run # POST: run all the nodes {missing,delete} lifecycle events and purges infra
  /event-feed 
    /status # GET: get the configuration and status for event-feed data lifecycle operations
    /config # PUT: update the event-feed lifecycle config
    /run # POST: purges event feed
  /compliance
    /status # GET: get the configuration and status for compliance data lifecycle operations
    /config # PUT: update the compliance lifecycle config 
    /run # POST: purge compliance reports and scans
```

There's also a shim for applications services:

```
  /services # Not currently implemented as the applications service is still in beta
    /status
    /config
    /run
```

Example aggregate status response
```json
{
  "infra": {
    "jobs": [
      {
        "name": "delete_nodes",
        "disabled": false,
        "recurrence": "FREQ=SECONDLY;DTSTART=20191106T180240Z;INTERVAL=600",
        "threshold": "2d",
        "purge_policies": null,
        "last_elapsed": "0.041154s",
        "next_due_at": "2019-11-07T22:32:40Z",
        "last_enqueued_at": "2019-11-07T22:22:40.017947Z",
        "last_started_at": null,
        "last_ended_at": null
      },
      {
        "name": "missing_nodes",
        "disabled": false,
        "recurrence": "FREQ=SECONDLY;DTSTART=20191106T180240Z;INTERVAL=600",
        "threshold": "2d",
        "purge_policies": null,
        "last_elapsed": "0.043205s",
        "next_due_at": "2019-11-07T22:32:40Z",
        "last_enqueued_at": "2019-11-07T22:22:40.026315Z",
        "last_started_at": null,
        "last_ended_at": null
      },
      {
[...]

PR #3158, opened 2020-03-19
Jamie/2523 data lifecycle refactor

SEAjamieD:
This branch updates the Node Lifecycles Page to now be named Data Lifecycles, and will allow the user more options in controlling their data retention choices through the UI

Finishes: https://github.com/chef/automate/issues/1208
Makes use of: https://github.com/chef/automate/pull/2157
UX Design File: https://app.abstract.com/share/47da49f4-55df-4be0-be66-018952857ae3?mode=design&sha=84902a48f104fe6bb5f4cefa3754cc83a72ccbe0

### :+1: Definition of Done
User now has the ability to update all areas available through the API.  Currently this includes Event Feed, Service Groups, Compliance, and will soon include Services Groups

make serve

navigate to https://a2-dev.test/settings/data-lifecycle
make any changes you wish to see in the form and submit - see a "success" notification at top of screen.

You should not:
Be able to enter any negative numbers into inputs
Be allowed to submit the form until there has been some sort of change made at least once

- [x] Tests added/updated?
- [ ] Docs added/updated?

Screen Shot 2020-03-19 at 11 38 57 AM


SEAjamieD:
Please note this PR is only currently ready for UX review

SEAjamieD:
This is now open for review to all - thanks!

@msorens - please note: I have not yet changed the name of the component to something more specific.  This components was already in the system and a search on it linked me to ~152 files.  I'd like to separate that out as a separate ticket so as not to hold this one back.

components/automate-ui/src/app/entities/automate-settings/automate-settings.effects.ts
```diff
@@ -32,18 +32,8 @@ export class AutomateSettingsEffects {
...
-
-  @Effect()
-  configureSettingsSuccess$ = this.actions$.pipe(
-    ofType(AutomateSettingsActionTypes.CONFIGURE_SETTINGS_SUCCESS),
-    map((_action) => new GetSettings({})));
+    mergeMap((action: ConfigureSettings) => this.requests.configureIngestJobs(action.payload.jobs)),
+    map((_resp) => (new ConfigureSettingsSuccess({}))),
+    catchError((error) => of(new ConfigureSettingsFailure(error))
+    ));
 }
```
SEAjamieD: Removing this effect saved us an unnecessary API call after the new configurations are made, as we've already got them in the form.

components/automate-ui/src/app/entities/automate-settings/automate-settings.requests.ts
```diff
@@ -3,16 +3,19 @@ import { HttpClient } from '@angular/common/http';
...
+  RespJobSchedulerStatus,
+  UnfurledJob,
+  JobRequestBody
 } from './automate-settings.model';
 
-import { environment } from '../../../environments/environment';
-const RETENTION_URL = environment.retention_url;
+// import { environment } from '../../../environments/environment';
+
+// const RETENTION_URL = environment.retention_url;
```
SEAjamieD: whoops!  I'll fix this ASAP

components/automate-ui/src/app/entities/automate-settings/automate-settings.requests.ts
```diff
@@ -22,49 +25,111 @@ export class AutomateSettingsRequests {
...
+
+        case 'periodic_purge_timeseries': // fallthrough
+        case 'periodic_purge':
+          thisJob = this.unfurlIngestJob(job, true);
+          matchedNest = body[job.category].job_settings.find(item => item.name === job.name);
+          matchedNest.purge_policies.elasticsearch.push(thisJob);
+          break;
+
+        default:
+          break;
```
SEAjamieD: In this API Response, as well as PUT call, some elements are nested, and some aren't, so I needed a way to configure the items at different levels based on whether on not they were nested, which sends them to the UnfurlJob function.

components/automate-ui/src/app/entities/automate-settings/automate-settings.requests.ts
```diff
@@ -22,49 +25,111 @@ export class AutomateSettingsRequests {
...
+    } else {
+      return {
+        name: job.name,
+        threshold: job.threshold,
+        disabled: job.disabled
+      };
+    }
+  }
+
+
```
SEAjamieD: This function is taking the job, and placing in the required format based on whether or not it is nested in the API Calls

components/automate-ui/src/app/entities/automate-settings/automate-settings.requests.ts
```diff
@@ -22,49 +25,111 @@ export class AutomateSettingsRequests {
...
+    Object.keys(respJobSchedulerStatus).forEach((category: string) => {
+      if (respJobSchedulerStatus[category]) {
+        const catJobs = respJobSchedulerStatus[category].jobs
+          .map((respJob: RespJob) => new IngestJob(category, respJob));
+        allJobs.push(...catJobs);
+      }
+    });
+
+    return new JobSchedulerStatus(allJobs);
   }
```
SEAjamieD: This functions purpose is to take the response we get back when requesting the configurations, and untangling it so that we can place it back in the form for the user to see.  I had to start by separating them by category, `infra`, `compliance`, `services`, `event_feed`.

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.spec.ts
```diff
@@ -88,78 +145,121 @@ describe('AutomateSettingsComponent', () => {
...
+
+        // expect(component.notificationVisible).toBe(true);
         expect(component.notificationType).toEqual('info');
         expect(component.notificationMessage)
-          .toEqual('All settings have been updated successfully');
-        expect(component.notificationVisible).toEqual(true);
+        .toEqual('All settings have been updated successfully');
+        expect(component.formChanged).toEqual(false);
+        // expect(component.saving).toEqual(false);
       });
```
SEAjamieD: this is a testing issue that I'm not yet sure how to resolve and could use some help.  Visually this is working as expected and i've located the reason why the test is failing but not sure how to resolve yet

susanev:
this is totally my fault, but I think notifications and data feeds belong in the same section in the left nav. I've got a question out to the belfast team to figure out if they are best to be in general settings or in node management and ill follow up with their answer tomorrow.

susanev:
following up, lets move data feeds up to the general settings section, so notifications, data feeds, data lifecycle for the order

SEAjamieD:
Came across some a11y issues today for keyboard navigation, adding to list of items to remedy

components/automate-ui/e2e/deprecated-routes.e2e-spec.ts
```diff
@@ -8,7 +8,7 @@ describe('Deprecated routes redirect to the correct new route for backwards comp
   // Note: This is "old path" -> "new path", and the new path will be checked
   // using a regular expression: the string, anchored to the right by '$'.
   [
-    ['/admin/settings',                        '/settings/node-lifecycle'],
+    ['/admin/settings',                        '/settings/data-lifecycle'],
```
msorens: Need to add a new one, too:
```
   ['/settings/node-lifecycle',                        '/settings/data-lifecycle'],
```

components/automate-ui/src/app/app-routing.module.ts
```diff
@@ -73,7 +73,7 @@ const routes: Routes = [
           component: SettingsLandingComponent
         },
         {
-          path: 'node-lifecycle',
+          path: 'data-lifecycle',
```
msorens: This change in nomenclature warrants changes to node-lifecycle.md and/or data-lifecycle.md.

components/automate-ui/src/app/entities/automate-settings/automate-settings.effects.ts
msorens: Nice catch!

components/automate-ui/src/app/entities/automate-settings/automate-settings.model.ts
```diff
@@ -1,73 +1,141 @@
...
+  };
+  compliance: {
+    jobs: RespJob[];
+  };
+  event_feed: {
+    jobs: RespJob[];
+  };
+  services: {
+    jobs: RespJob[];
+  };
```
msorens: If these top-level properties only have a single child each, wondering if we could compress them to something like this? If the original is what you get from the backend, then there's not much we can do, of course.
```suggestion
  infra_jobs: RespJob[];
  compliance_jobs: RespJob[];
  event_feed_jobs: RespJob[];
  services_jobs: RespJob[];
```
(Same consideration for JobRequestBody)

SEAjamieD:
@susanev 


✅ Tooltip over services checkboxes --> let me know if you'd like me to change anything about these

✅ there's 24 px of space between the nested ones, can we reduce that to 12px?

✅ darken disband dropdown color --> is the disabled dropdown text lighter than input or are my eyes bad?

✅ can you remove all the margins from that save changes button?

✅ if there's anyway we can remove this gray background on this page, that would be rad. but not critical.

✅ Rearrange Sidebar

Can you give me another look on the Save Button being enabled on enter?  This is working on my computer reliably and curious if its still happening.

components/automate-ui/src/app/entities/automate-settings/automate-settings.model.ts
```diff
@@ -1,73 +1,141 @@
...
+  nested_name?: string;
+  recurrence?: string;
   threshold: string;
-  every?: string;
-  lastRun?: Date;
-  nextRun?: Date;
-  lastElapsed?: Date;
-  startedOn?: Date;
+  disabled: boolean;
+  purge_policies?: any; // needs specific type
```
msorens: Suggest add a TODO so it is more clear to everyone, plus it makes it easier to find such loose ends.
```suggestion
  purge_policies?: any; // TODO needs specific type
```

components/automate-ui/src/app/entities/automate-settings/automate-settings.model.ts
```diff
@@ -1,73 +1,141 @@
...
+export interface JobRequestComponent {
+  name: string;
+  purge_policies: {
+    elasticsearch: UnfurledJob[];
+  };
+}
+
+export interface JobRequestBody {
+  infra: {
+    // Infra is a special case and con contain a JobRequestComponent
```
msorens:
```suggestion
    // Infra is a special case and can contain a list of JobRequestComponents
```

components/automate-ui/src/app/entities/automate-settings/automate-settings.model.ts
```diff
@@ -1,73 +1,141 @@
...
+}
+
+export enum InfraJobName {
+  MissingNodes = 'missing_nodes',
+  MissingNodesForDeletion = 'missing_nodes_for_deletion',
+  DeleteNodes = 'delete_nodes',
+  PeriodicPurgeTimeseries = 'periodic_purge_timeseries'
+}
+
+export enum NestedJobName {
```
msorens: Might be worth a comment to the effect that jobs that are nested within the _xyz_ property of an infra job must be one of these.

components/automate-ui/src/app/entities/automate-settings/automate-settings.requests.ts
```diff
@@ -22,49 +25,111 @@ export class AutomateSettingsRequests {
   // fetchJobSchedulerStatus sends an HTTP GET Request to read the status of the JobScheduler
   // living inside the ingest-service
   public fetchJobSchedulerStatus(_params): Observable<JobSchedulerStatus> {
-    const url = `${RETENTION_URL}/nodes/status`;
+    // NEW ENDPOINT FOR DATA-LIFECYCLE
+    const url = '/api/v0/data-lifecycle/status';
```
msorens: 1. Probably do not need the comment there; that's only useful while reading this PR.
2. For the URL, we want to use the same convention as what you are supplanting: replace `/api/v0/data-lifecycle` with a value from the environment files. (In fact, if the name still makes sense, you could just update the definition of `RETENTION_URL` in the environment files and you would not need any changes in this file.

components/automate-ui/src/app/entities/automate-settings/automate-settings.requests.ts
```diff
@@ -22,49 +25,111 @@ export class AutomateSettingsRequests {
...
 
-    switch (job.name) {
-      case IngestJobs.MissingNodes: {
-        url = `${RETENTION_URL}/nodes/missing-nodes/config`;
-        break;
-      }
-      case IngestJobs.MissingNodesForDeletion: {
-        url = `${RETENTION_URL}/nodes/missing-nodes-deletion/config`;
-        break;
+    const body: JobRequestBody = {
```
msorens: 🎉 Thanks for the strong typing.

components/automate-ui/src/app/entities/automate-settings/automate-settings.requests.ts
```diff
@@ -22,49 +25,111 @@ export class AutomateSettingsRequests {
...
+            }
+          }
+        ]
       }
-      case IngestJobs.DeleteNodes: {
-        url = `${RETENTION_URL}/nodes/delete-nodes/config`;
-        break;
+    };
+
+    jobs.forEach((job: IngestJob): void => {
```
msorens: Could simplify a bit:
```suggestion
    jobs.forEach(job => {
```
The type is already inferred by machine and easily determined by humans.

components/automate-ui/src/app/entities/automate-settings/automate-settings.requests.ts
```diff
@@ -22,49 +25,111 @@ export class AutomateSettingsRequests {
...
+      let thisJob;
+      let matchedNest;
+
+      switch (job.name) {  // ADD STRONG TYPING
+        case 'delete_nodes':      // fallthrough
```
msorens: Use the enums, Luke. ⚡️ 
```suggestion
        case InfraJobName.DeleteNodes:      // fallthrough
```

components/automate-ui/src/app/entities/automate-settings/automate-settings.model.ts
```diff
@@ -1,73 +1,141 @@
...
-  // have been marked for deletion
-  DeleteNodes = 'delete_nodes'
+  // Compliance
+  ComplianceRemoveReports = 'complianceRemoveReports',
+  ComplianceRemoveScans = 'complianceRemoveScans'
 }
 
 export class IngestJob {
-  running: boolean;
+  category: string;
```
msorens: Here is a case where strong types are vital. To see why, before you make this change open this up in VSCode: https://github.com/chef/automate/blob/c2552860717826455b86591c8d821f33ac67b0f8/components/automate-ui/src/app/entities/automate-settings/automate-settings.requests.ts#L75-L98. If you hover over `job_settings` you will see it has type `any` even though `JobRequestBody` itself is strongly typed.

This needs to be either…
```suggestion
  category: 'infra' | 'compliance' | 'event_feed' | 'services';
```
or preferred, introduce an enum, e.g.:
```suggestion
  category: JobCategories;
```
(Once you make this change it will reveal an issue in that same block of code in the requests file.)

components/automate-ui/src/app/entities/automate-settings/automate-settings.requests.ts
```diff
@@ -22,49 +25,111 @@ export class AutomateSettingsRequests {
...
+        ]
       }
-      case IngestJobs.DeleteNodes: {
-        url = `${RETENTION_URL}/nodes/delete-nodes/config`;
-        break;
+    };
+
+    jobs.forEach((job: IngestJob): void => {
+      let thisJob;
+      let matchedNest;
```
msorens: More strong types needed here to avoid "compiler forgiveness". Before you make this change, you could have a typo in `matchedNest.purge_policies` and the compiler would not complain. Once you strongly type these, it is much safer.
```suggestion
      let thisJob: UnfurledJob;
      let matchedNest: JobRequestComponent;
```

components/automate-ui/src/app/entities/automate-settings/automate-settings.requests.ts
```diff
@@ -22,49 +25,111 @@ export class AutomateSettingsRequests {
...
+        case 'missing_nodes_for_deletion':
+          thisJob = this.unfurlIngestJob(job);
+          body[job.category].job_settings.push(thisJob);
+          break;
+
+        case 'periodic_purge_timeseries': // fallthrough
+        case 'periodic_purge':
+          thisJob = this.unfurlIngestJob(job, true);
+          matchedNest = body[job.category].job_settings.find(item => item.name === job.name);
+          matchedNest.purge_policies.elasticsearch.push(thisJob);
```
msorens: Anything to guarantee that matchedNest is never undefined?

components/automate-ui/src/app/entities/automate-settings/automate-settings.requests.ts
```diff
@@ -22,49 +25,111 @@ export class AutomateSettingsRequests {
...
   private convertResponseToJobSchedulerStatus(
     respJobSchedulerStatus: RespJobSchedulerStatus): JobSchedulerStatus {
-      const jobs = respJobSchedulerStatus.jobs.map((respJob: RespJob) => new IngestJob(respJob));
-      return new JobSchedulerStatus(respJobSchedulerStatus.running, jobs);
+
+    const allJobs = [];
+
+    Object.keys(respJobSchedulerStatus).forEach((category: string) => {
+      if (respJobSchedulerStatus[category]) {
+        const catJobs = respJobSchedulerStatus[category].jobs
```
msorens: This restricted type will allow hovering over `.jobs` in the last line to show a real type instead of `any`.
```suggestion
    Object.keys(respJobSchedulerStatus).forEach((category: JobCategories) => {
      if (respJobSchedulerStatus[category]) {
        const catJobs = respJobSchedulerStatus[category].jobs
```

susanev:
@SEAjamieD heres a video of the save button stuff I mentioned
data-lifecycle-save-button

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.ts
```diff
@@ -102,63 +197,102 @@ export class AutomateSettingsComponent implements OnInit {
...
+    if ( !allowedKeys.includes(key) ) {
+      return false;
+    }
   }
 
-  // patchDisableValue is the workaround for the chef-checkbox molecule since it is not
-  // an input annotation we need to patch the value inside the FormGroup
-  public patchDisableValue(form, checked: boolean) {
-    form.controls['disable'].setValue(checked);
+  public toggleInput(form, checked: boolean): void {
```
msorens: It would be more clear to rename this to either `setInput` or `setState`. You are not toggling the input; you are setting it to the given value. (The giveaway is that a true toggle method would not need a boolean parameter because it is just toggling.)

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.spec.ts
```diff
@@ -63,21 +85,56 @@ describe('AutomateSettingsComponent', () => {
...
+      // Service Groups on not currently uncheckable through the UI
+      // ['serviceGroupNoHealthChecks', true, false ],
+      // ['serviceGroupRemoveServices', true, false ],
+      ['clientRunsRemoveData', true, false ],
+      ['clientRunsLabelMissing', true, false ],
+      ['clientRunsRemoveNodes', true, false ],
+      ['complianceRemoveReports', true, false ],
+      ['complianceRemoveScans', true, false ]
+    ], function (form: string, currentState: boolean, expectedState: boolean) {
+      it('deactivates the asociated form', () => {
```
msorens: Confused by the test name compared to the last test; should this be "activates" instead of "deactivates"?
(Also typo: asociated)

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.spec.ts
```diff
@@ -88,78 +145,121 @@ describe('AutomateSettingsComponent', () => {
...
 
-    it('does not updates the "eventFeed" form group', () => {
-      component.updateForm(mockJobSchedulerStatus);
+    using([
+      ['eventFeedRemoveData', false, 53 ],
+      ['eventFeedServerActions', true, undefined ], // Infra purge_timeseries -> server_actions
+      // ['serviceGroupNoHealthChecks', false, 5 ], // Services not enabled yet
+      // ['serviceGroupRemoveServices', false, 5 ], // Services not enabled yet
+      ['clientRunsRemoveData', false, '7' ], // Infra Remove data
+      ['clientRunsLabelMissing', false, '14' ], // Infra label as missing data
```
msorens: Either adjust the input like so or adjust the parameter type in the function signature.
```suggestion
      ['clientRunsRemoveData', false, 7 ], // Infra Remove data
      ['clientRunsLabelMissing', false, 14 ], // Infra label as missing data
```

susanev:
@SEAjamieD 
- theres an `rgba(0, 0, 0, 0.38);` on the dropdown when its disabled, so i think thats why the text color appears lighter than the numerical input
- gray background, still appears at the very bottom of the page, we may be able to leverage whatever they did on the applications page to stop this from happening, but this is super low priority so if its not quick to fix dont worry about it
background-woes

components/automate-ui/src/app/entities/automate-settings/automate-settings.model.ts
SEAjamieD: Unfortunately - this is what i'm getting from the backend and matches the API, so I feel it makes more sense to leave it like this.

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.spec.ts
```diff
@@ -63,21 +85,56 @@ describe('AutomateSettingsComponent', () => {
...
+      ['complianceRemoveScans', false, true ]
+    ], function( form: string, currentState: boolean, expectedState: boolean) {
+      it('deactivates the asociated form', () => {
+        expect(component[form].value.disabled).toEqual(currentState);
+        component.toggleInput(component[form], currentState);
+        expect(component[form].value.disabled).toEqual(expectedState);
+        expect(component[form].get('unit').disabled).toBe(expectedState);
+        expect(component[form].get('threshold').disabled).toBe(expectedState);
+      });
+    });
```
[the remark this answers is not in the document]
SEAjamieD: 😲 This is awesome!

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.spec.ts
SEAjamieD: great catch!

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.spec.ts
```diff
@@ -88,78 +145,121 @@ describe('AutomateSettingsComponent', () => {
...
+        clientRunsLabelMissing,
+        complianceForms
+      ]);
     });
 
-    it('does not updates the "eventFeed" form group', () => {
-      component.updateForm(mockJobSchedulerStatus);
+    using([
+      ['eventFeedRemoveData', false, 53 ],
+      ['eventFeedServerActions', true, undefined ], // Infra purge_timeseries -> server_actions
```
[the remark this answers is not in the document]
SEAjamieD: I think this test can be improved from what you've mentioned here, and I'll attempt to make some improvements, but I waned to provide some clarity first.

`eventFeedServerActions` is not a special case, I was using it to test out a different scenario, i.e. the form is disabled.  What I should really be doing is testing all of them in both fashions.

When a formControl is disabled in the UI, Angular's formGroup actually removes that formControl from the formGroup, so the value comes through as undefined instead of as a value, which is what we expect.  When a group is disabled, we don't actually need that value for the time being.

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.spec.ts
SEAjamieD: In this particular case, that number is a string for the "older" part of the API and a number fo the "newer" parts of the API.  Would you recommend normalizing this in the frontend somewhere?

SEAjamieD: Actually - you sort of answered this down below, as this could be handled in factory functions.  I think I'm all set

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.spec.ts
```diff
@@ -88,78 +130,173 @@ describe('AutomateSettingsComponent', () => {
...
+                                           job.threshold.slice(-1)
+          ];
+          expect(newFormValues.threshold).toEqual(jobThreshold);
+          expect(newFormValues.unit).toEqual(jobUnit);
+        }
+
+        // all forms share the same expectation of being disabled
+        expect(newFormValues.disabled).toEqual(jobData.disabled);
+      });
     });
```
SEAjamieD: @msorens - before I go further on this, do you think it makes sense to separate out non-nested and nested forms into their own separate test blocks for clarity?  I'm using an `if/else` inside this test block and I feel like this is going more difficult for others to read.

SEAjamieD:
@susan - when UX reviewing - I just noticed checkboxes are not reactive to the enter key, so I'll be sure to fix that

components/automate-ui/src/app/app-routing.module.ts
SEAjamieD: @kagarmoe / @mjingle - Making sure I don't step on any toes here, is this part of documentation or is this a change that I should handle?

mjingle: I think this is being taken care of already? We're removing any instance of node-lifecycle in the docs per  https://github.com/chef/automate/pull/3170/ 
Also, I'm pretty sure that we're deleting node-lifecycle.md and creating data-lifecycle.md in its place.

susanev: data lifecycle docs page already existed, but yah deleting the node lifecycle page.. with a redirect.

SEAjamieD: ok great - thats the info I was looking for.  Thank you!

SEAjamieD:
> @susan - when UX reviewing - I just noticed checkboxes are not reactive to the enter key, so I'll be sure to fix that

Apologies, please disregard, chef-checkbox uses spacebar, not enter as I had thought

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.ts
SEAjamieD: I renamed it to 'handleFormActivation' because in essence it is either activating, or deactivating the form based on the click

components/automate-ui/src/app/entities/automate-settings/automate-settings.requests.ts
SEAjamieD: @msorens - I began to patch this, but I could use your opinion on my approach in your next run through;

components/automate-ui/src/app/entities/automate-settings/automate-settings.requests.ts
```diff
@@ -22,49 +27,123 @@ export class AutomateSettingsRequests {
...
+      switch (job.name) {
+        case InfraJobName.DeleteNodes:              // fallthrough
+        case InfraJobName.MissingNodes:             // fallthrough
+        case InfraJobName.MissingNodesForDeletion:
+          thisJob = this.unfurlIngestJob(job);
+          // Protect against an undefined value
+          if (body[job.category].job_settings) {
+            body[job.category].job_settings.push(thisJob);
+          } else {
+            console.log('TODO: send an error');
```
SEAjamieD: Not sure if this is the right approach here - Open to opinions on how to protect against an undefined value.

components/automate-ui/src/app/entities/automate-settings/automate-settings.model.ts
```diff
@@ -1,73 +1,182 @@
...
+  disabled: boolean;
+  policy_name?: string;
+  older_than_days?: number;
+  name?: string;
+  threshold?: string;
+}
+
+// A JobRequestComponent is very flexible
+  // It needs to be able to contain an older API Job object
+  // And a newer API Job Object
```
tylercloke: [ubernit] Never seen the comment tabbed out before.

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.scss
```diff
@@ -1,35 +1,104 @@
...
+// temporary - needs to go in theme file for material
+mat-form-field {
+  margin: 0 8px 0 0;
+  width: 120px;
+
+  ::ng-deep .mat-form-field-wrapper {
     padding-bottom: 0;
   }
+
+  ::ng-deep .mat-form-field-infix {
```
tylercloke: What's our stance on ng-deep for my own reference? I thought we were trying to not use it?

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.scss
```diff
@@ -1,35 +1,104 @@
...
+  transition: border 0.4s ease;
+
+  &:focus,
+  &:active {
+    border-color: $chef-primary-bright;
+    outline: none;
+  }
 }
+
+// end temporary
```
tylercloke: What's this communicating?

SEAjamieD: Good catch - that was a note for me because I needed to move some Sass over to an application wide level

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.scss
SEAjamieD: In this case - It's necessary for us to override some of the Angular Material styles that are applied by default.  otherwise, I avoid using.

components/automate-ui/src/app/entities/automate-settings/automate-settings.model.ts
SEAjamieD: I generally tab out a comment when there is information that extends out from the first line, but by bringing this up, I'm realizing I could probably right this comment better so that that isn't required.  Good lookin out!

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.ts
```diff
@@ -102,65 +197,109 @@ export class AutomateSettingsComponent implements OnInit {
...
+      const job = new IngestJob(null, null);
+      job.category = jobForm.category;
+      job.name = jobForm.name;
+      job.disabled = jobForm.disabled;
+
+      // If the user doesn't enter any number at all - this defaults to 0
+      if (jobForm.threshold === null) {
+        job.threshold = '0' + jobForm.unit;
+      } else {
+        job.threshold = jobForm.threshold + jobForm.unit;
```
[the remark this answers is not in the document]
SEAjamieD: This is one of those cases that is needed because of the difference between the older pieces of the API and the newer pieces of the API.  

The older pieces of the API store/read the threshold value as a string with both the unit and threshold values.  For example, if the value is "4 days", the threshold on this portion of the api is stored as "4d".

The newer portions of the API are only storing the value for the amount of days to between retention.  and they are storing them as a number.  ie. numeric 4.

From how i'm looking at it, we can't really avoid packing and splitting.  But if theres some other way that i'm not seeing - i'm all ears.

If possibly you're suggesting simply renaming `job.threshold` to something like 'job.packed_threshold` so that ultimately that line would read as
`job.packed_threshold = jobForm.threshold + jobForm.unit` or something like that - maybe that would be easier to read?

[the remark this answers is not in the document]
msorens: Roger that; sounds like you do have justification for the pack/split. ✔️ 

Your rename idea is probably a good thought, though. (IIRC you were using the variable `threshold` in some places to be the unpacked value and in others to be the packed value, which is a bit confusing.)

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.ts
```diff
@@ -211,53 +344,59 @@ export class AutomateSettingsComponent implements OnInit {
...
+          break;
+      }
     });
   }
 
-  private getJobForm(job: string) {
-    switch (job) {
-      case IngestJobs.MissingNodes: {
-        return this.automateSettingsForm.value['missingNodes'];
+  private getJobForm(jobName: string) {
```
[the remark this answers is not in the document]
msorens: In case I was not clear, was just showing this as an intermediate step. 😁 Since you only call this function in one place, consider getting rid of it completely. Just use `this[jobName]` where it is called from.

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.spec.ts
msorens: Great question, and yes I do. Should strive for cyclomatic complexity === 1  for test cases whenever possible. (That may or may not mean anything to you; if not, happy to chat!)

components/automate-ui/src/app/entities/automate-settings/automate-settings.model.ts
```diff
@@ -1,73 +1,182 @@
...
-  every: string;
+  nested_name?: string;
+  disabled: boolean;
+  recurrence: string;
   threshold: string;
-  last_run: Date;
-  next_run: Date;
-  last_elapsed: Date;
-  started_on: Date;
+  purge_policies: any; // TODO needs specific type
```
msorens, line 21: Looks like a couple outstanding "TODO" items.

components/automate-ui/src/app/entities/automate-settings/automate-settings.model.ts
```diff
@@ -1,73 +1,182 @@
...
+export class UnfurledJob {
+  disabled: boolean;
+  policy_name?: string;
+  older_than_days?: number;
+  name?: string;
+  threshold?: string;
+}
+
+// A JobRequestComponent is very flexible so that it may contain
+// contain an older API Job object and a newer API Job Object
```
msorens, line 131:
```suggestion
// an older API Job object or a newer API Job Object
```

components/automate-ui/src/app/entities/automate-settings/automate-settings.requests.ts
```diff
@@ -3,15 +3,20 @@ import { HttpClient } from '@angular/common/http';
...
-  IngestJobs,
-  RespJobSchedulerStatus
+  RespJobSchedulerStatus,
+  UnfurledJob,
+  JobRequestBody,
+  InfraJobName,
+  JobCategories
 } from './automate-settings.model';
 
 import { environment } from '../../../environments/environment';
```
msorens, line 18: Prefer to avoid such "parented" paths.
```suggestion
import { environment } from 'environments/environment';
```

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.spec.ts
```diff
@@ -63,21 +86,40 @@ describe('AutomateSettingsComponent', () => {
...
-    ]);
+    expect(Object.keys(component.automateSettingsForm.controls)).toEqual(ALL_FORMS);
   });
 
-  describe('patchDisableValue(form, value)', () => {
-    it('updates the value of the disable control from the provided form', () => {
-      expect(component.clientRunsForm.value.disable).toEqual(false);
-      component.patchDisableValue(component.clientRunsForm, true);
-      expect(component.clientRunsForm.value.disable).toEqual(true);
+  describe('handleFormActivation(form, value)', () => {
```
msorens, line 92: Not sure those params add value here...
```suggestion
  describe('handleFormActivation', () => {
```

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.spec.ts
```diff
@@ -63,21 +86,40 @@ describe('AutomateSettingsComponent', () => {
...
+
+    using(ALL_FORMS
+        // Service Groups on not currently uncheckable through the UI
```
msorens, line 95:
```suggestion
        // Service Groups are not currently uncheckable through the UI
```
Same applies in next `using` block below.

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.ts
```diff
@@ -102,65 +197,115 @@ export class AutomateSettingsComponent implements OnInit {
...
+
+      // If the user doesn't enter any number at all - this defaults to 0
+      jobForm.threshold === null
+        ? job.threshold = '0' + jobForm.unit
+        : job.threshold = jobForm.threshold + jobForm.unit;
+
+      if ( isNested ) {
+        job.nested_name = jobForm.nested_name;
+        job.threshold = jobForm.threshold;  // Automatically becomes a 0 from
+                                            // parseInt in reqeust if left blank
```
msorens, line 302:
```suggestion
                                            // parseInt in request if left blank
```

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.ts
```diff
@@ -211,61 +350,135 @@ export class AutomateSettingsComponent implements OnInit {
...
-        return this.automateSettingsForm.value['deleteMissingNodes'];
-      }
-    }
+  // private getJobForm(jobName: string): RawJobFormGroupValues {
+  private getJobForm(jobName: string): FormGroup {
+    return this[jobName];
   }
 
-  private splitThreshold(threshold: string) {
+  private splitThreshold(threshold: string): [string, string] {
```
msorens, line 383: Please add a comment, perhaps even giving an example, e.g.
```
splits packed threshold into number and unit, where unit is a single, final character. Example: "2495d" => ["2495", "d"]
```

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.ts
```diff
@@ -102,65 +197,115 @@ export class AutomateSettingsComponent implements OnInit {
...
+    this.isDestroyed.next(true);
+    this.isDestroyed.complete();
   }
 
-  // patchUnitValue is the workaround for the chef-select molecule since it is not
-  // an input annotation we need to patch the value inside the FormGroup
-  public patchUnitValue(form, event) {
-    form.controls['unit'].setValue(event.target.value);
+  // This prevents a user from being allowed to enter negative numbers
+  // or other actions that we dont want to allow
```
msorens, line 231:
```suggestion
  // or other actions that we don't want to allow
```

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.spec.ts
SEAjamieD: good catch! 😂 was on autopilot there

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.ts
```diff
@@ -24,75 +31,163 @@ import { TelemetryService } from '../../services/telemetry/telemetry.service';
...
 
-  // Has the form changed?
-  formChanged = false;
+  // Are settings currently saving
+  saving = false;
 
   // Notification bits
   notificationVisible = false;
   notificationType = 'info';
   notificationMessage = 'All settings have been updated successfully';
```
susanev, line 134:
```suggestion
  notificationMessage = 'Settings saved.';
```

SEAjamieD: no prob!

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.spec.ts
```diff
@@ -88,78 +130,230 @@ describe('AutomateSettingsComponent', () => {
...
         component.updateForm(mockJobSchedulerStatus);
         component.applyChanges();
-        expect(component.formChanged).toEqual(false);
+
+        // expect(component.notificationVisible).toBe(true);
         expect(component.notificationType).toEqual('info');
         expect(component.notificationMessage)
-          .toEqual('All settings have been updated successfully');
-        expect(component.notificationVisible).toEqual(true);
+        .toEqual('All settings have been updated successfully');
```
susanev, line 356: sorry, shoulda commented on the test too

```suggestion
        .toEqual('Settings saved.');
```

SEAjamieD: nah, that's not your responsibility - that was on me.  I pushed up and went to bed without checking the tests.  Sorry about that!

PR #3159, opened 2020-03-19
Add warning about running Manage with Server and A2

IanMadd:
Signed-off-by: IanMadd <imaddaus@chef.io>

This adds a warning about deploying Infra Server with Chef Manage and A2.

- [ ] Tests added/updated?
- [X] Docs added/updated?

### DIFF
--- a/components/automate-chef-io/content/docs/infra-server.md
+++ b/components/automate-chef-io/content/docs/infra-server.md
@@ -14,10 +14,15 @@ toc = true
 
 {{% warning %}}
 
-Chef Automate will not deploy the following Chef Infra Server add-ons:
-- Chef Manage
-- Push Jobs Server
-- Supermarket
+Chef Automate will not deploy the Chef Infra Server add-ons Chef Manage
+and Push Jobs Server.
+
+{{% /warning %}}
+
+{{% warning %}}
+
+Supermarket cannot authenticate users on Chef Infra Server that is deployed
+with Chef Automate.
 
 {{% /warning %}}
 
--- a/components/automate-chef-io/content/docs/infra-server.md
+++ b/components/automate-chef-io/content/docs/infra-server.md
@@ -12,6 +12,12 @@ toc = true
     weight = 60
 +++
 
+{{% warning %}}
+
+We do not support using the Chef Automate 2 deployment service to deploy Chef Infra Server with Chef Manage.
+
+{{% /warning %}}
+
 Use Chef Automate to install Chef Infra Server either for a single-host installation that contains both Chef Infra Server and Chef Automate, or for a standalone Chef Infra Server instance.
 See the [Chef Infra Server documentation](https://docs.chef.io/server_overview/) for instructions and guidance on using and managing your Chef Infra Server.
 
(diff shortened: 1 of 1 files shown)


components/automate-chef-io/content/docs/infra-server.md
```diff
@@ -12,6 +12,12 @@ toc = true
     weight = 60
 +++
 
+{{% warning %}}
+
+We do not support using the Chef Automate 2 deployment service to deploy Chef Infra Server with Chef Manage.
```
susanev: is there any way to get _Chef Manage_ closer to the beginning of the sentence? because up until that point it almost sounds like we are saying that we don't support deploying chef infra server with automate.

IanMadd: Latest draft:
`Chef Manage cannot be used to deploy Chef Infra Server using Chef Automate`

IanMadd: Or : `We do not support using Chef Manage with Chef Automate to deploy Chef Infra Server.`

yzl: I would say something like "Chef Automate does not deploy Chef Manage alongside Chef Infra Server."

components/automate-chef-io/content/docs/infra-server.md
```diff
@@ -12,6 +12,12 @@ toc = true
     weight = 60
 +++
 
+{{% warning %}}
+
+Chef Automate does not deploy Chef Manage alongside Chef Infra Server.
```
btm: > Chef Automate does not support installing any Chef Infra Server add-ons, such as Chef Manage or Push Jobs Server.

susanev: do we want to mention that supermarket is not supported as well?

IanMadd: rewrite:

Chef Automate will not deploy the following Chef Infra Server add-ons: 
- Chef Manage
- Push Jobs Server
- Supermarket

btm: technically supermarket isn't an "addon" as the addons were packages that could be installed with `chef-server-ctl install NAME`. I was happy to confirm this morning that the `chef-server-ctl` command shipped with automate is already well pruned and doesn't have the `install` command at all.

https://docs.chef.io/ctl_chef_server/#install

https://github.com/chef/chef-server/blob/master/src/chef-server-ctl/plugins/install.rb#L4-L8

(reporting is already EOL a while ago).

It's more technically correct that Supermarket cannot integrate with Automate Infra Server because we don't ship the `oc-id` service which is what supermarket uses to authenticate. We haven't ruled out doing something with Automate on-premises and Supermarket but nothing is currently planned either.

IanMadd: So maybe two separate warnings. 

```
Chef Automate will not deploy the Chef Infra Server add-ons Chef Manage
and Push Jobs Server.
```
and 
```
Supermarket cannot authenticate users on Chef Infra Server that is deployed 
with Chef Automate.
```

susanev: sounds good to me!

PR #3163, opened 2020-03-20
Make sure passwords with percent work

jaym:
The existing code that we have to parse the platform config and generate postgresql connection strings contains a bug that results in unparseable URL strings when the password contains a `%` character.

The cause of this is that we generate the bulk of the URL connection string (user, password, ip, port, options) and then run `fmt.Sprintf()` on THAT string to insert the database name later on.

This change uses a function pointer to make the `fmt.Sprintf` call only once with all of the data.

### DIFF
--- a/integration/services/ha_backend/init.sh
+++ b/integration/services/ha_backend/init.sh
@@ -10,6 +10,7 @@ ha_backend_container2=$(service_container_name "ha_backend_2")
 
 ha_backend_private=$(service_config_path "ha_backend_private")
 ha_backend_config=$(service_config_path "ha_backend.toml")
+ha_admin_pg_password='thisisapassword%u'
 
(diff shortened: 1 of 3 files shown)


sdelano:
I like the approach to testing this case in the full-stack integration test! It's currently failing with a urlescape error:

```
Mar 20 14:13:32 automate-9173-9185 hab[3033]: pg-sidecar-service.default(O): time="2020-03-20T14:13:32Z" level=fatal msg="parse \"postgresql://admin:thisisapassword%u@172.17.0.4:10145/postgres?sslmode=verify-ca&sslrootcert=/hab/svc/pg-sidecar-service/config/_a2_platform_external_pg_root_ca.crt\": invalid URL escape \"%u\""
```

PR #3166, opened 2020-03-22
Rm/fix ipaddress pointer race condition

rickmarry:
This fixes a bug that was caused by setting a pointer to an attribute on an incoming inspec report.
By making a copy of the string from the incoming struct and then pointing to that instead, we avoid having the pointer get overwritten by subsequent messages in the channel.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/compliance-service/ingest/pipeline/processor/report.go
+++ b/components/compliance-service/ingest/pipeline/processor/report.go
@@ -46,11 +46,13 @@ func ComplianceReport(in <-chan message.Compliance) <-chan message.Compliance {
 				ChefTags:         msg.Report.ChefTags,
 				FQDN:             msg.Report.Fqdn,
 			}
+
+			copyOfIpAddress := msg.Report.Ipaddress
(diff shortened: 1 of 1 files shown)


rickmarry:
> Good catch and thank you for digging into this. This looks reasonable to me as the near term fix. I quickly scanned the `ESInSpecReport` structure and didn't immediately spot any other problematic fields.
> 
> In the long run we might want to look into whether we could fix this in the type system in a way that makes it harder to make this mistake. For instance via something like the sql package's NullString type. This would likely require a change to the ES library we use though.
> 
> Thanks again. Great debuggering.

thank you @stevendanna.. In order to aid  in the debugging, @alexpop and I also made some really useful additions to the data that chef-load posts, one of which is the ipaddress field, formatted as ipv4 and coded to be equal to the number of the node that it corresponds with.. that pr will be up soon too.

alexpop:
Great fix in less than 1 business day after looking at the issue with the customer over Zoom.

PR #3181, opened 2020-03-23
Refer to `infra-server` rather than `chef-server`

### LINKED ISSUE #3176: Expose the chef-automate deploy --product flag
`chef-automate deploy --product` is still a hidden flag. Behind it lies the ability to install an infra server and/or automate, and presumably now on-premises hab depot.

We're expecting Chef Desktop users to have separate Infra Server and Automate infrastructure, being a different department in most enterprises. We'd like them to start by using the Automate Infra Server rather than a separate instance.

Acceptance Critiera:
* Add 'infra-server' as a product. Can be aliased to 'chef-server' or vice versa. But we want to promote the newer more specific Chef Infra server name.
* The 'chef-server' product continues to work without any changes. Must not break OWCA.
* `chef-automate deploy --help` shows `--product` and valid options.
* docs updated to use `infra-server` instead of `chef-server`. #2851

yzl:

Fixes #3176 

- [ ] Tests added/updated?
- [x] Docs added/updated?

### DIFF
--- a/components/automate-chef-io/content/docs/workflow_install.md
+++ b/components/automate-chef-io/content/docs/workflow_install.md
@@ -72,7 +72,7 @@ Then, on the Chef server:
 
 2. Creating a user in previous step displays a private key to stdout. Copy this content and and save it on the **Chef Automate** server as: `/hab/svc/automate-workflow-server/var/etc/delivery.pem`
 
-3. If this is a new chef-server, create an organization on the chef server with:
+3. If this is a new Chef Server, create an organization on the chef server with:
 
     ```shell
     `sudo chef-server-ctl org-create workflow workflow
@@ -168,7 +168,7 @@ Selecting **Workflow** opens the legacy Workflow screen.
 
     ```text
       192.168.0.20    automate-test
-      192.168.1.100   chef-server
+      192.168.1.100   chef-infra-server
     ```
 
 #### On the Automate 2 server
--- a/products.meta
+++ b/products.meta
@@ -150,7 +150,7 @@
     {
       "name": "chef-server",
       "type": "product",
-      "aliases": ["chef-infra-server"],
+      "aliases": ["chef-infra-server", "infra-server"],
       "dependencies": ["core", "postgresql", "elasticsearch"],
       "services": [
         "chef/automate-cs-bookshelf",
--- a/components/automate-chef-io/content/docs/client-runs.md
+++ b/components/automate-chef-io/content/docs/client-runs.md
@@ -154,7 +154,7 @@ curl -sSX POST "https://automate-url/api/v0/ingest/events/chef/nodedelete" -d
 '{
   "node_name": "somenode",
   "organization_name": "yourorg",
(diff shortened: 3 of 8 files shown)


products.meta
```diff
@@ -150,7 +150,7 @@
     {
       "name": "chef-server",
       "type": "product",
-      "aliases": ["chef-infra-server"],
+      "aliases": ["chef-infra-server", "infra-server"],
```
sdelano: I like the aliases part. Thanks for ensuring that existing deployments will continue to work!

kagarmoe:
Hi,
@chef/docs-team needs some time to sort out the ramifications of this rename. This is likely a much more significant set of changes than are in this PR, but as this rename is news to us, we need time for discovery.
Thanks!
Kimberly

kagarmoe:
I've looked at it & discussed the PR with @btm . The changes are fewer than I feared, so you just need to find and change any uses of the `chef-server` cli in the Automate docs. @chef/docs-team will handle fixing the uses of `chef-server` that should be `Chef Infra Server` in another PR.

yzl:
@btm I ran `git grep chef-server` on the docs to make sure I changed everything that needs changing, and I have a couple of questions:

1. Is the name of the config file still `/etc/opscode/chef-server.rb`?
1. Is the name of the command still `chef-server-ctl`?
1. Are these still the attribute names in the audit cookbook?
```
default['audit']['reporter'] = 'chef-server-automate'
default['audit']['fetcher'] = 'chef-server'
```
Otherwise, the only other instances of `chef-server` I found in the docs are in markdown links, which will be handled when @chef/docs-team makes the Chef Server -> Chef Infra Server change.

btm:
> 1. Is the name of the config file still `/etc/opscode/chef-server.rb`?
> 2. Is the name of the command still `chef-server-ctl`?
> 3. Are these still the attribute names in the audit cookbook?
> 
> ```
> default['audit']['reporter'] = 'chef-server-automate'
> default['audit']['fetcher'] = 'chef-server'
> ```

Yeah, all of those stay the same. There would be a number of specific issues to change each of those that would require cross-product work. 

Of course this is the first time we've documented this flag so there really aren't many/any uses of `chef-server` in this way, as you found.

I did take a stab at `Chef Server` updates this morning in #3202.

components/automate-chef-io/content/docs/workflow_install.md
```diff
@@ -72,7 +72,7 @@ Then, on the Chef server:
 
 2. Creating a user in previous step displays a private key to stdout. Copy this content and and save it on the **Chef Automate** server as: `/hab/svc/automate-workflow-server/var/etc/delivery.pem`
 
-3. If this is a new chef-server, create an organization on the chef server with:
+3. If this is a new Chef Server, create an organization on the chef server with:
```
btm: This should be Chef Infra Server

yzl: I'm leaving that for the docs team change, in order to avoid the inconsistency of the new term amid a sea of the old ones.

PR #3188, opened 2020-03-24
[automate-1886] farewell chef-automate admin-token

bcmdarroch:

Removes the last few references to `chef-automate admin-token`

- [ ] Tests added/updated?
- [x] Docs added/updated?


dev-docs/acceptance-testing/domain_services_acceptance_checklist.md
```diff
@@ -20,7 +20,7 @@ settings: (node integrations) a few integrations\ already created, (node credent
 
     ```
     ssh your-ad-user@a2-local-inplace-upgrade-acceptance.cd.chef.co
-    sudo chef-automate admin-token
+    sudo chef-automate iam token create admin-token-1 --admin
     ```
 
     * Or [Use the Automate UI](https://github.com/chef/automate/blob/master/components/compliance-service/smokin/README.md#generating-a-token)
```
susanev: i think the steps linked to here are gonna need to be updated, to indicate that after the token is created that they need to add it to a policy... idk if we want to pass that over to the super team though?

kagarmoe:
@bcmdarroch Wow, this is quite the PR! I'm ready to 👍 the API changes, but is there any way we can move the Hugo docs changes into a separate PR? I don't want to delay getting a PR of this size merged on while we edit.

susanev:
@kagarmoe for the edits to the config.md? or something else?

susanev:
added the do-not-merge label cause docs has some linting work to do on this pr

components/automate-chef-io/content/docs/configuration.md
```diff
@@ -60,28 +62,29 @@ upgrade_strategy = "at-once"
...
+The only supported `deployment_type` is `local`.
 
 #### Settings
 
-Currently, you cannot change the admin username, name, and password set during initial deployment.
+You cannot change the admin username, name, and password set during initial deployment.
 
 To change the admin password after deployment, use the Automate UI.
-Log in as the admin user, navigate to the _User_ page under the **Settings**  tab. Selecting "Local Administrator" opens a form for updating the password. Enter and confirm your new password in the interface, and then select the **Update Password** button to save your changes.
+Log in as the admin user, navigate to the _User_ page under the **Settings** tab.
```
susanev:
```suggestion
Sign in as the admin user, navigate to the _Users_ page under the **Settings** tab.
```

components/automate-chef-io/content/docs/configuration.md
```diff
@@ -60,28 +62,29 @@ upgrade_strategy = "at-once"
...
+Selecting "Local Administrator" opens a form for updating the password.
```
susanev: this is outdated, after going to the user detail page they now have to go to the reset password tab to reset the password

components/automate-chef-io/content/docs/configuration.md
```diff
@@ -60,28 +62,29 @@ upgrade_strategy = "at-once"
...
+Enter and confirm your new password in the interface, and then select the **Update Password** button to save your changes.
```
susanev:
```suggestion
Enter and confirm your new password in the interface, and then select the **Reset Password** button to save your changes.
```

components/automate-chef-io/content/docs/configuration.md
```diff
@@ -60,28 +62,29 @@ upgrade_strategy = "at-once"
...
 
-To change the admin password from the command-line, first
-[fetch the admin user record]({{< relref "users.md#fetching-users" >}})
-and copy the User ID, then use:
+To change the admin password from the command-line, first [fetch the admin user record]({{< relref "users.md#fetching-users" >}}), copy the User ID, and then use:
```
susanev: this should prolly link to the api docs instead

components/automate-chef-io/content/docs/configuration.md
kagarmoe: @IanMadd Let's add this to our linter list.

components/automate-chef-io/content/docs/configuration.md
```diff
@@ -60,31 +62,32 @@ upgrade_strategy = "at-once"
 
 Then run `chef-automate config patch </path/to/your-file.toml>` to deploy your change.
 
-To upgrade a Chef Automate installation when the `upgrade_strategy` is set to `none`, run:
+To upgrade a Chef Automate installation with the `upgrade_strategy` set to `none`, run:
```
stevendanna: This is the only of these cleanups that I'm unsure about.  The intent here is to say something like:

> If your Chef Automate installation's `upgrade_strategy` is set to `none`, you can still upgrade to the latest version by running:

I think "with" sounds strange to me here because the underlying command works the same regardless of upgrade_strategy and thus it also works "without" the upgrade strategy set to none.

components/automate-chef-io/data/docs/cli_chef_automate/commands/chef-automate_admin-token.yaml
```diff
@@ -1,31 +0,0 @@
-name: chef-automate admin-token
```
stevendanna: Heads up: I think this file will get re-committed to the repository after merge because we have automation that autogenerates it.

components/automate-chef-io/content/docs/configuration.md
mjingle: @stevendanna I re-wrote the sentence to get rid of the passive verb tense. Anything with "is + verb" - for example: "is set" - falls under passive verb tense. Would this be corrected by the following correction? I realized "the" could probably be removed for better flow:

> To upgrade a Chef Automate installation with `upgrade_strategy` set to `none`, run:

components/automate-chef-io/data/docs/cli_chef_automate/commands/chef-automate_admin-token.yaml
bcmdarroch: the `admin-token` command has already been removed from the CLI. should I remove it from anywhere else?

components/automate-chef-io/content/docs/configuration.md
stevendanna: I had assumed that is why you made the change.  My issue is not with the flow of the sentence (although I like the removal of "the"), but rather that we are being too subtle here about _why_ the user should care about the following command. Further, to some it might _imply_ that the following command might not work with an `upgrade_strategy` of `at-once` even though it does.  That said, I think it is best that we merge this change and I can open a PR to address my concerns if they continue to bother me.

The good news is that as part of this, I re-read this section and found that we overlooked the much more serious problem.  The command below is incorrect.  It should be

```bash
chef-automate upgrade run
```

mjingle: I am following your 🚋 of thought now. I'll have to think a bit further on this. 

In the meantime, let me incorporate the correct command. I am so glad you caught that!

components/automate-chef-io/data/docs/cli_chef_automate/commands/chef-automate_admin-token.yaml
stevendanna: I'm a moron and checked the wrong branch. I thought this was against master 🤦‍♂

bcmdarroch: No worries! Don't want to miss anything 👀

PR #3192, opened 2020-03-24
Add CIDR filtering to restrict data-feed to certain nodes

satellite15:

Two new configuration options with defaults

disable_cidr_filter = true
cidr_filter = "0.0.0.0/0"

disable_cidr_filter: true/false
cidr_filter: string of csv CIDRS e.g: "172.18.2.120/29,172.18.2.128/29,172.18.2.140/30,172.18.2.144/30,172.18.2.148/31,172.18.2.150/32"

Thus Allowing a user to configure the data-feed-service for a restricted number of nodes using a CIDR block or a number of CIDR blocks.  

The poll task will only include nodes whose ip address is in the defined CIDR ranges, thus only those nodes will be reported in the data feed

Update the config manually and restart the service to test CIDR ranges

- [x] Tests added/updated?
- [ ] Docs added/updated? NA no docs yet

### DIFF
--- a/components/data-feed-service/habitat/default.toml
+++ b/components/data-feed-service/habitat/default.toml
@@ -21,7 +21,7 @@ accept = false
 
 [log]
 format = "text"
-level = "debug"
+level = "info"
 
 [storage]
 database = "data_feed_service"
--- a/components/data-feed-service/habitat/default.toml
+++ b/components/data-feed-service/habitat/default.toml
@@ -6,10 +6,8 @@ asset_page_size = 100
 reports_page_size = 1000
 node_batch_size = 100
 updated_nodes_only = true
-#disable_cidr_filter = true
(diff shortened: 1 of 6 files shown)


components/data-feed-service/habitat/default.toml
```diff
@@ -6,6 +6,8 @@ asset_page_size = 100
 reports_page_size = 1000
 node_batch_size = 100
 updated_nodes_only = true
+disable_cidr_filter = true
```
vjeffrey: this is a cool new feature! 
it also reminds me to ask, are there docs somewhere for all this functionality?

satellite15: Not yet, the task is on the board and will get them done when ready to remove the feature flag

PR #3193, opened 2020-03-24
Make ES gateway timeouts configurable

jaym:
S3 operations get really slow with many backups / daily indexes.
We've seen a delete exceed the hard coded default of 5 mins, and
cause gateway timeouts.

This changes the default to 10 mins and makes the timeouts configurable


api/config/esgateway/config_request.proto
```diff
@@ -68,6 +68,8 @@ message ConfigRequest {
 					google.protobuf.Int32Value ssl_verify_depth = 15;
 					google.protobuf.Int32Value server_names_hash_bucket_size = 16;
 					google.protobuf.StringValue client_body_buffer_size = 17;
+					google.protobuf.Int32Value proxy_read_timeout = 18;
+					google.protobuf.Int32Value proxy_timeout_timeout = 19;
```
ryancragun:
```suggestion
					google.protobuf.Int32Value proxy_send_timeout = 19;
```

components/automate-es-gateway/habitat/default.toml
```diff
@@ -48,3 +48,5 @@ tcp_nopush = "on"
 tcp_nodelay = "on"
 ssl_verify_depth = 2
 proxy_connect_timeout = 5
+proxy_read_timeout = 600
+proxy_send_timeout = 600
```
stevendanna: I know this is annoying, but I think for hab config items that are also user-configurable, we should probably set the default in the config_request.go file as well for consistency with the rest of the configuration?

PR #3194, opened 2020-03-24
ensure region is always set for aws api integration

vjeffrey:
### :nut_and_bolt: Description: What code changed, and why?
this issue was reported in slack

a user found that all controls for the aws profile had results like "region value is not set/missing". they also noticed the env value was being set to gcp-api

this change fixes the env region to be the correct "aws-api" and ensures that the region gets set when calling the awscreds function

### :+1: Definition of Done
aws api scans working

### :athletic_shoe: How to Build and Test the Change
add a node integration for aws (https://github.com/chef/automate/blob/master/components/compliance-service/scripts/create-pg-data.sh#L144)
run a scan job on it, see the error about missing region
rebuild compliance and nodemanager
run scan again, no region error, just normal results

### :white_check_mark: Checklist

- [x] I have read the [CONTRIBUTING document](https://github.com/chef/automate/blob/master/CONTRIBUTING.md).
- [x] Tests added/updated?
- [n/a] Docs added/updated?
- [x] All commits have been signed-off for the [Developer Certification of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

### :camera: Screenshots, if applicable
Screen Shot 2020-03-24 at 13 57 20

PR #3199, opened 2020-03-24
ensure the region is set for credentialless node integrations

vjeffrey:

follow up to https://github.com/chef/automate/pull/3194

that one fixed the case for aws api integrations that use credentials.
it unfortunately missed the case for when we read creds from cloud env

### DIFF
--- a/components/compliance-service/inspec-agent/resolver/resolver.go
+++ b/components/compliance-service/inspec-agent/resolver/resolver.go
@@ -339,7 +339,16 @@ func (r *Resolver) handleAwsApiNodesSingleNode(ctx context.Context, m *manager.N
 		awsCreds = managers.GetAWSCreds(secret)
 	}
 
-	nodeInfo := nodeInfo{
+	nodeInfo, tc, secrets := assembleAwsApiNodeInfo(node, m, awsCreds)
+	inspecJob, err := assembleJob(job, nodeInfo, []*inspec.Secrets{&secrets}, tc)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error assembling job. aborting scan for node %s", node.Name)
+	}
+	return []*types.InspecJob{inspecJob}, nil
+}
+
+func assembleAwsApiNodeInfo(node *nodes.Node, m *manager.NodeManager, awsCreds awsec2.AwsCreds) (nodeInfo, inspec.TargetBaseConfig, inspec.Secrets) {
+	nodeDetails := nodeInfo{
 		UUID:           node.Id,
 		CloudID:        m.AccountId,
 		Name:           node.Name,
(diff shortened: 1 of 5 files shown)


components/compliance-service/inspec-agent/resolver/resolver.go
```diff
@@ -339,29 +339,38 @@ func (r *Resolver) handleAwsApiNodesSingleNode(ctx context.Context, m *manager.N
...
+func assembleAwsApiNodeInfo(node *nodes.Node, m *manager.NodeManager, awsCreds awsec2.AwsCreds) (nodeInfo, inspec.TargetBaseConfig, inspec.Secrets) {
+	nodeDetails := nodeInfo{
 		UUID:           node.Id,
 		CloudID:        m.AccountId,
 		Name:           node.Name,
 		Environment:    "aws-api",
 		ManagerID:      m.Id,
 		CloudAccountID: m.AccountId,
 	}
+	region := awsCreds.Region
```
vjeffrey: line 359 to 367 here are the real changes.

i wanted to add a unit test to i pulled this bit of node assembling logic out

vjeffrey: (if no credential id is provided, which is the case for the "read creds from my cloud env" option, awsCreds.Region is blank

stevendanna:
Here is the only thing I'm not 100% understanding about these fixes:

When we inspected the database entry for the relevant in the  `node_managers_nodes` the region was clearly listed. Is the region that is store there just unused? Is it intended to be used?

vjeffrey:
> When we inspected the database entry for the relevant in the node_managers_nodes the region was clearly listed. Is the region that is store there just unused? Is it intended to be used?

my initial thought was, well. the region is set on the nodemanager level, not the node, so thats why.
but i think we're saving the region to the node based on the nodemanager setting, so that should technically work - we should be able to grab the node.Region like we grab the node.Name -- if the function we're getting that has that info/is grabbing that val from the db. im not sure yet though, i'll play with it

vjeffrey:
k, step one: when we add a nodemanager, we end up here: https://github.com/chef/automate/blob/master/components/nodemanager-service/pgdb/node-managers-cloud.go#L112
which is incorrectly hardcoding the region to be us east 1, ignoring the user's desired region target (this is important for fed customers, bc they need to specify govcloud region)

vjeffrey:
i pushed a commit that checks the value of the node's region and takes it if the nodemanager one isnt specified
now im going to see about not hardcoding the region value when creating the node

vjeffrey:
ok so i removed the hardcoded region and replaced with the value from the aws creds for when we're creating a node. locally i created another integration and saved us-west-2 as the region. so my old integration still used us-east-1, my new one used us-west-2
```
:27Z" level=info msg="region being used for aws scan: us-east-1"
[2020-03-25T01:03:27+00:00] secrets-service.default(O): time="2020-03-25T01:03:27Z" level=info msg="read secret with : da34d660-2581-49a4-a99f-71409560a5a9"
[2020-03-25T01:03:27+00:00] compliance-service.default(O): time="2020-03-25T01:03:27Z" level=info msg="region being used for aws scan: us-west-2"
```

vjeffrey:
which is cool -- and good for integrations created going forward, but we still need to check the value of the awsCreds.Region and prioritize that for govcloud customers who already created integrations, or they would end up getting us-east-1.
this does mean that we should always have the region from the aws creds or the node itself, but i still left the extra empty region check there just in case because yikes

vjeffrey:
rebased on master cause CI looked wonky

vjeffrey:
```
Mar 25 16:28:45 a2-local-inplace-upgrade-dev.cd.chef.co hab[1177]: compliance-service.default(O): time="2020-03-25T16:28:45Z" level=info msg="region being used for aws scan: us-east-1"
Mar 25 16:28:46 a2-local-inplace-upgrade-dev.cd.chef.co hab[1177]: secrets-service.default(O): time="2020-03-25T16:28:46Z" level=info msg="read secret with : 69f97e40-d3a3-41ca-8a8f-994d6ecafd8c"
Mar 25 16:28:46 a2-local-inplace-upgrade-dev.cd.chef.co hab[1177]: compliance-service.default(O): time="2020-03-25T16:28:46Z" level=info msg="region being used for aws scan: us-gov-west-1"
Mar 25 16:28:46 a2-local-inplace-upgrade-dev.cd.chef.co hab[1177]: secrets-service.default(O): time="2020-03-25T16:28:46Z" level=info msg="read secret with : 5a30e8b3-682d-4860-9b18-a5d71d3e63ba"
Mar 25 16:28:46 a2-local-inplace-upgrade-dev.cd.chef.co hab[1177]: compliance-service.default(O): time="2020-03-25T16:28:46Z" level=info msg="region being used for aws scan: us-east-1"
```

Issue #2545, opened 2019-12-18
ui: Waivers - Profile list reflects the new Waived Status of a profile 


jonong1972:
## User Story

As an Automate User, I will use this work when I want to see which profiles have been waived.

## Problem Statement
In the rare case, if a profile has all waived controls within it, we need to show that the profile itself has been waived.

## Definition of Done
1. A user can see if a Profile is waived based on the new status that the API returns for a profile
2. A user can filter by status of Waived

## Mock Ups
![V3 - Compliance - Report - Profiles](https://user-images.githubusercontent.com/1316587/71296588-0574c900-2335-11ea-9219-692b1c910724.png)

vjeffrey:
this work involves: 
- adding the "waived" icon for profiles with a status of waived:
![Screen Shot 2020-02-26 at 05.36.24.png](https://images.zenhubusercontent.com/5cfebc28d99ac71fbfbdcb73/6e9fe230-df5e-43cc-b373-3e0575afae73) 
- ensuring the logic for showing status of a profile accounts for waived profiles
- add status filter buttons to the reporting profiles list page, as shown above. api response already has the info:
![Screen Shot 2020-02-26 at 05.35.28.png](https://images.zenhubusercontent.com/5cfebc28d99ac71fbfbdcb73/9031cd29-0031-41c5-83ab-285c649c27aa)
```
curl -s --insecure -H "api-token: $ACCEPTANCE_TOKEN" https://a2-local-inplace-upgrade-acceptance.cd.chef.co/api/v0/compliance/reporting/profiles -d '{"filters":[{"type":"status","values":["passed"]}]}'
{"profiles":[{"name":"apache-baseline","title":"DevSec Apache Baseline","id":"41a02784bfea15592ba2748d55927d8d1f9da205816ef18d3bb2ebe4c5ce18a8","version":"2.0.0","status":"passed"}],"counts":{"total":1,"failed":0,"skipped":0,"passed":1,"waived":0}}
```

vjeffrey:
load data by using the `load_compliance_reports` command from your studio dev env

vjeffrey:
status filters blocked by https://github.com/chef/automate/issues/3178

Issue #3178, opened 2020-03-23
compliance reporting profiles counts should NOT change when filtering by status


vjeffrey:
## Describe the bug
When calling the compliance/reporting/profiles API, counts are returned. These counts will be used for the filter buttons in the UI.
It appears that these counts are being modified when a status filter is included in the request.
```
"counts":{"total":6,"failed":4,"skipped":2,"passed":0,"waived":0}}
                                                                                                                           
curl -s --insecure -H "api-token: $ACCEPTANCE_TOKEN" https://a2-local-inplace-upgrade-acceptance.cd.chef.co/api/v0/compliance/reporting/profiles -d '{"filters":[{"type":"status","values":["passed"]}]}'
{"profiles":[],"counts":{"total":0,"failed":0,"skipped":0,"passed":0,"waived":0}}
```

## To Reproduce
- call the compliance reporting profiles api. note counts.
- call the same api with a status filter. note difference in counts.

## Expected behavior
The counts returned with the API should be unaffected by a status filter. 
In other words, if i call the API with no filters, I should get a full list of profiles and the counts.
If I call the API with a status filter, I should get a list of profiles that have the same status as the requested one, but the counts should remain unchanged.
This will allow us to use backend filtering for the status filters on that page.

Issue #3177, opened 2020-03-23
compliance reporting nodes counts should NOT change when filtering by status


vjeffrey:
## Describe the bug
When calling the compliance/reporting/nodes API, counts are returned. These counts are used for the filter buttons in the UI.
It appears that these counts are being modified when a status filter is included in the request.
```
no filters: "total":5,"total_passed":0,"total_failed":5,"total_skipped":0,"total_waived":0}

with filter: 
curl -H "api-token: $ACCEPTANCE_TOKEN" -s https://a2-local-inplace-upgrade-acceptance.cd.chef.co/api/v0/compliance/reporting/nodes/search -d '{"filters": [{"type": "status", "values": ["passed"] } ]}'
{"nodes":[],"total":0,"total_passed":0,"total_failed":0,"total_skipped":0,"total_waived":0}
```

## To Reproduce
- call the compliance reporting nodes api. note counts.
- call the same api with a status filter. note difference in counts.

## Expected behavior
The counts returned with the API should be unaffected by a status filter. 
In other words, if i call the API with no filters, I should get a full list of nodes and the counts.
If I call the API with a status filter, I should get a list of nodes that have the same status as the requested one, but the counts should remain unchanged.
This will allow us to use backend filtering for the status filters on that page.

vjeffrey:
related: https://github.com/chef/automate/issues/3178

PR #3201, opened 2020-03-25
Automate-2545 Profile List Waivers

tarablack01:
As an Automate User, I will use this work when I want to see which profiles have been waived.

In the rare case, if a profile has all waived controls within it, we need to show that the profile itself has been waived.

https://github.com/chef/automate/issues/2545

### :+1: Definition of Done
1. A user can see if a Profile is waived based on the new status that the API returns for a profile
2. A user can filter by status of Waived

load data by using the `load_compliance_reports` command from your studio dev env

Screen Shot 2020-03-25 at 10 02 30 AM
Screen Shot 2020-03-25 at 10 02 37 AM

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-profiles/reporting-profiles.component.ts
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-profiles/reporting-profiles.component.ts
@@ -23,6 +23,7 @@ export class ReportingProfilesComponent implements OnInit, OnDestroy {
   layerOneData: any = {};
   layerTwoData: any = {};
   control: any = {};
+  profileFilterStatus = 'all';
 
   scanResultsPane = 0;
   openControls = {};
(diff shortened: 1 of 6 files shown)


vjeffrey:
note: this pr introduces the status filters with ui filtering b/c the backend is not quite ready. once the backend is ready (https://github.com/chef/automate/issues/3178, https://github.com/chef/automate/issues/3177) we will be updating both the nodes reporting and profiles reporting pages to filter via api call

components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-profiles/reporting-profiles.component.ts
```diff
@@ -23,6 +23,7 @@ export class ReportingProfilesComponent implements OnInit, OnDestroy {
   layerOneData: any = {};
   layerTwoData: any = {};
   control: any = {};
+  profileFilterStatus = 'all';
```
SEAjamieD: I'm not sure this is of use to you yet until its added in (or if you decide to add it earlier) - just added strong typing that looks like it would support this here: https://github.com/chef/automate/commit/dbb3374e2427320d0241aea7bdf553e33c4f4da2#diff-44b3f8b3bfe76490fb6f906b379bc1a4

tarablack01: Added type :+1::skin-tone-3:

tarablack01: @SEAjamieD Had to revert, it complained that the type was added with a literal string.

Issue #2544, opened 2019-12-18
ui: Waivers - Control list has hover state for waived details


jonong1972:
## User Story

As an Automate User, I will use this work when I want to more details about the waived controls.

## Problem Statement

- We need a way to show the Reason why a waiver was set, the (if applicable) the expiration date, and (if ran) the status of nodes that were waived.
- Sometimes, there may be multiple waivers used with the same control and we need a way to reflect that.

## Definition of Done

1. A user can go to the Control List view, hover over the waived icon on any control that was waived and view the:
a. Expiration Date
b. Reason for being waived
c. [conditional] if it was ran, then show the nodes status in the hover dialogue, but do not show that control as passed or failed or skipped on the control row.

##  Mock Ups
This shows a control that had 2 waivers on it that both ran.
![Screen Shot 2019-12-20 at 2 25 25 PM](https://user-images.githubusercontent.com/1316587/71296517-bd55a680-2334-11ea-857e-caf5990427e3.png)

This shows a control that had 2 waivers on it that did not run.
![Screen Shot 2019-12-20 at 2 26 12 PM](https://user-images.githubusercontent.com/1316587/71296520-c21a5a80-2334-11ea-8940-94364a626f78.png)

vjeffrey:
load data with `load_compliance_reports`

for any node with more than 0 for the waived count under control_summary, apply a hover with details about the waived control
this work involves:
- on hover over waived icon, show data about the waiver applied to the control and the status of the waived control if it was executed.
_data for waived control that was **not** executed will have "yes" for the value of "waived_str":_
<img width="933" alt="Screen Shot 2020-03-12 at 20 08 06" src="https://user-images.githubusercontent.com/10341541/76582932-5d871b80-649d-11ea-882e-341560ce057c.png">

_data for waived control that was executed will have "yes_run" for the value of "waived_str":_
<img width="909" alt="Screen Shot 2020-03-12 at 20 08 24" src="https://user-images.githubusercontent.com/10341541/76582938-637cfc80-649d-11ea-832e-3e31ca75ca03.png">

Issue #2543, opened 2019-12-18
ui: Waivers - Control List displays waived controls


jonong1972:
## Use Case
As an Automate User, I will use this work when I want to see if there are any waived controls.

## Problem

- We need a way to show waivers on the Compliance Report Control List View.  
- We need to show the count of nodes on a particular control that have been waived.
- We need a way to filter by controls that were waived on nodes.

## Definition of Done

1. A user can see the count of nodes that have been waived on any control within the control list
2. A user can filter by "Waived Controls" and the list will display only controls that waived nodes.
3. The "Waived Controls" filter button should display the count of waived controls and should be clickable.
5. Total Controls count should be equal to the sum of Failed + Passed + Skipped + Waived counts

## Mock Up
![V3 - Compliance - Report - Controls](https://user-images.githubusercontent.com/1316587/71104042-09a3b980-2170-11ea-945a-6469faaf34b3.png)

vjeffrey:
from the `api/v0/compliance/reporting/controls` API:

```
{"control_items":[
{"id":"pro2-con1","title":"Profile 2 - Control 1", "profile":{"name":"","title":"My Profile 2 title","id":"447542ecfb8a8800ed0146039da3af8fed047f575f6037cfba75f3b664a97ea5","version":"1.0.5","status":""},"impact":0.8,"end_time":"2020-02-18T10:18:41Z","control_summary":{"total":2,"passed":{"total":0},"skipped":{"total":0},"failed":{"total":0,"minor":0,"major":0,"critical":0},"waived":{"total":2}},"waivers":[{"waived_str":"yes_run","expiration_date":"","justification":"Sound reasoning","waiver_summary":{"total":0,"passed":{"total":0},"skipped":{"total":0},"failed":{"total":2,"minor":0,"major":0,"critical":0},"waived":{"total":2}}}]}]}
```

to load this data into the ui, run `load_compliance_reports`, and filter by profile "my profile 2 title"

mattray:
I assume this is still under development, but with build 20200310163217 if you are using waivers the control shows up under "Skipped Controls" but the count is "0"

<img width="1440" alt="Screen Shot 2020-03-19 at 5 13 16 pm" src="https://user-images.githubusercontent.com/330023/77037407-62038500-6a05-11ea-9f8d-9fbd70a95569.png">

vjeffrey:
yupp. not started yet, should be complete by end of next week

PR #3204, opened 2020-03-25
Automate 2544 2543 Control List Waived

tarablack01:
As an Automate User, I will use this work when I want to more details about the waived controls.

We need a way to show the Reason why a waiver was set, the (if applicable) the expiration date, and (if ran) the status of nodes that were waived.
Sometimes, there may be multiple waivers used with the same control and we need a way to reflect that.

https://github.com/chef/automate/issues/2544
https://github.com/chef/automate/issues/2543

### :+1: Definition of Done
- We need a way to show the Reason why a waiver was set, the (if applicable) the expiration date, and (if ran) the status of nodes that were waived.
- Sometimes, there may be multiple waivers used with the same control and we need a way to reflect that.

A user can go to the Control List view, hover over the waived icon on any control that was waived and view the:
a. Expiration Date
b. Reason for being waived
c. [conditional] if it was ran, then show the nodes status in the hover dialogue, but do not show that control as passed or failed or skipped on the control row.

if a control, under "control_summary" has more than 0 waived nodes, show the waived icon with the count as shown above

load data with load_compliance_reports

Screen Shot 2020-03-25 at 1 32 01 PM
Screen Shot 2020-03-25 at 1 32 14 PM

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-controls/reporting-controls.component.html
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-controls/reporting-controls.component.html
@@ -49,7 +49,7 @@
               <div class="control-waiver-details" *ngFor="let waiver of control.waivers">
                 <div class="waiver-details">
                   <span class="label">Expires:</span>
-                  <span class="message">{{ waiver.expiration_date | datetime:'ddd, D MMMM YYYY HH:mm:ss [UTC]' }}</span>
+                  <span class="message">{{ waiver.expiration_date | datetime: Datetime.RFC2822 }}</span>
                 </div>
                 <div class="waiver-details">
                   <span class="label">Reason:</span>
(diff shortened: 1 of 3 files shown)


components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-controls/reporting-controls.component.html
```diff
@@ -41,6 +41,39 @@
...
           </span>
+          <span *ngIf="summary.waived.total > 0" class="control-result" [ngClass]="{'waived': summary.waived.total > 0}">
+            <div [attr.id]="'waived-icon' + control.id" class="waived-icon"></div>
+            <chef-tooltip
+              [attr.for]="'waived-icon' + control.id"
+              position="left">
+              <div class="control-waiver-details" *ngFor="let waiver of control.waivers">
+                <div class="waiver-details">
+                  <span class="label">Expires:</span>
+                  <span class="message">{{ waiver.expiration_date | datetime:'ddd, D MMMM YYYY HH:mm:ss [UTC]' }}</span>
```
SEAjamieD: We actually have RFC2822 defined as a Constant DateTime helper here: https://github.com/chef/automate/blob/ce74f7cfe38018a32476cad92c479b4ac88a3b5f/components/automate-ui/src/app/helpers/datetime/datetime.ts#L6

tarablack01: @SEAjamieD I've added DateTime, thanks for finding that! :)

PR #3227, opened 2020-03-27
deployment: remove go-bindata

stevendanna:
I was going to do something else, but that something else seemed to
require that I run go-bindata. I didn't have that installed and the
`go` command line tool seemed to refuse to install it without also
upgrading it. This annoyed me enough that I've finally removed it.

We were using go-bindata in a very simplicity way. Basically, just a
way to get some strings and templates into our final binary. I've
replaced most of our uses of it by copy and pasting those files into
constants.

I changed around how some of the bind generation works such that now
we should always be able to run:

    make generate

and just have everything work.

Take it or leave it. We can't remove go-bindata as a dep since the
auth system uses it and they seem to use it in a way that actually
makes it a bit more useful.


api/config/deployment/init_config.go
```diff
@@ -139,15 +138,11 @@ func GenerateInitConfig(channel string, upgradeStrategy string, opts ...InitConf
...
 func (c InitConfig) Render() (string, error) {
-	b := assets.MustAsset("data/init-config.toml")
-	tempRaw := string(b)
-
-	var buf bytes.Buffer
-
 	temp := template.Must(template.New("init").
 		Funcs(template.FuncMap{"StringsJoin": strings.Join}).
-		Parse(tempRaw))
+		Parse(configTemplate))
```
jaym: where is configTemplate coming from

Issue #2870, opened 2020-02-13
document config mgmt export nodes endpointss


vjeffrey:
## User Story

Since the export endpoints are served via mux, the process for this is a little different. we have examples from compliance: https://github.com/chef/automate/blob/master/components/automate-chef-io/data/docs/api-static/01-reporting-export.swagger.json

## Definition of Done
Config mgmt export endpoints are documented (there are two)

PR #3231, opened 2020-03-27
Adding export API docs

lancewf:
Adding API documentation for the infra node export and node run export. Since the export endpoints are served via mux, we have to manually create the swagger docs.

#2870 

1. Follow the steps in this link to start the docs server. https://github.com/chef/automate/tree/master/components/automate-chef-io#development
1. Go to http://localhost:1313/docs/api/#operation/NodeExport
1. Go to http://localhost:1313/docs/api/#operation/ReportExport

- [ ] Tests added/updated?
- [x] Docs added/updated?

image

image


api/external/cfgmgmt/request/nodes.proto
```diff
@@ -41,3 +41,25 @@ message Runs {
   // Latest (in history) run information to return for the runs list.
   string end = 5;
 }
+
```
lancewf: These are added for reference.

components/automate-gateway/gateway/server.go
```diff
@@ -416,7 +416,7 @@ func (s *Server) startHTTPServer() error {
 
 	// custom mux route for export (ignores its request method)
 	// needed b/c gateway does not support stream; corresponds to
-	// https://github.com/chef/automate/blob/master/components/automate-gateway/api/compliance/reporting/reporting.proto
+	// https://github.com/chef/automate/blob/master/api/interservice/compliance/reporting/reporting.proto#L15
```
lancewf: This file was moved in a prior PR.

components/automate-chef-io/data/docs/api-static/05-node-export.swagger.json
```diff
@@ -0,0 +1,68 @@
+{
```
lancewf: This is the main change. Manually created the swagger doc for the streaming end points.

components/automate-gateway/gateway_mocks/mock_gateway/clients_mock.go
```diff
@@ -1,5 +1,5 @@
 // Code generated by MockGen. DO NOT EDIT.
-// Source: ../gateway/clients.go
+// Source: gateway/clients.go
```
lancewf: This changes on master also.

kagarmoe:
Hi @lancewf,
I'm running into this error when I try to see the docs on localhost:

Screen Shot 2020-03-30 at 7 40 03 PM

lancewf:
I figured out the problem so the error does not show, but I still am not seeing the end points in the docs. I am also not seeing the `GetErrors` endpoint.

kagarmoe:
Our problem is that `NodeExport` and (presumably) `ReportExport` aren't getting built in the config protos, so we can't call them with `$ref` in the `swagger.json`. 

I might be missing some background; why are we calling the response from the compliance service instead of the responses from the config-mgmt service ? (The streams are defined in `node_export.go` & `report_export.go`; how `api/interservice/cfgmgmt/response/root.proto` uses the stream isn't clear to me) 
If using the compliance service is what we want, then do we need to use them with a query?

lancewf:
@kagarmoe I did fix the problem with the `NodeExport` and `ReportExport`. I had to define them in the static file. I was using the compliance response object because it was the only one that was in the external API. I can move the compliance response to a common location to remove the confusion.

api/external/common/response.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+
+package chef.automate.api.common;
+option go_package = "github.com/chef/automate/api/external/common";
+
+message ExportData {
+	// Exported reports in JSON or CSV.
+	bytes content = 1;
+}
```
lancewf, line 9: Moved this here because both compliance and config-mgmt is using it.

kagarmoe:
It works! We need to add a few more fields to it, particularly the `required` array.  Here's an [example]( https://github.com/chef/automate/pull/2982/files?file-filters%5B%5D=.proto#diff-0fec7cf6c93014380f11ef914139844aR29-R32)
@msorens Has a good write-up on some of the complexities of how we're generating our API docs: 
https://github.com/chef/automate/pull/2982

lancewf:
Just updated this with "May be empty", "Required", "Not Required" in the descriptions.

lancewf:
@susanev I don't think we can because it is using a ref to an object.

PR #3232, opened 2020-03-27
Full Protobuf Rebuilds

danielsdeleo:

Whilst dealing with some rebase woes recently, I found that I was unable to fully rebuild our compiled protobuf code and decided to fix it. So this PR does the following:
* Add studio function to delete all the compiled `*pb*go` files
* Add studio function to bootstrap the custom `protoc-gen-*` tools used in the main protobuf builds
* Add `scripts/grpc.sh` where it was missing, and adjust `compile_all_protobuf_components` accordingly for a "weird" one
* Rebuild all of the protobufs and update ones that have changed a bit
* Update the pipeline to do a full bootstrap rebuild when it tests the protos. The pipeline can now catch issues like a `.proto` file getting removed but the corresponding compiled file remaining, or protos getting added without being correctly integrated in our build tooling.

### :+1: Definition of Done

The protobuf code can be rebootstrapped

In the studio:

* Delete all the things! `clean_all_compiled_protos` - check it out with `git status`
* `bootstrap_proto_compilers && compile_all_protobuf_components` - the repo should now be unchanged from before.

### DIFF
--- a/.studio/common
+++ b/.studio/common
@@ -34,10 +34,10 @@ DOC
 function compile_go_protobuf_component() {
   [ "x$1" == "x" ] && error "Missing component name argument; try 'describe ${FUNCNAME[0]}'" && return 1
 
-  # FIXME: this comment is wrong
-  # Note: while we remove all pb-derived files before regenerating, we're
-  # actually cheating a bit: the protoc-gen-* executables built above do partly
-  # depend on them.
+  # NOTE: To (re)build the world from scratch, automate-grpc must first be
+  # bootstrapped so that our custom protoc-gen-* commands can be used in later
+  # build steps. See `bootstrap_proto_compilers` above.
+
   local path
   if [ "$1" == "api" ]; then
     path=api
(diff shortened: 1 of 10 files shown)


.studio/common
```diff
@@ -3,6 +3,28 @@
...
 # two or more components share/require (ex. install elasticsearch)
 
+document "clean_all_compiled_protos" <<DOC
+  Delete all the files with extensions matching '*pb*go' in the repo (except in vendor/)
+DOC
+function clean_all_compiled_protos() {
+  install_if_missing core/git git
+
+  find /src/{api,components,lib} -name '*pb*go' -exec rm '{}' \; 
+  git clean -fdx
```
stevendanna: This is what I'm guessing @msorens hit.  If we leave it we should probably put a warning in the documentation.

stevendanna:
I've removed the `git clean` and also fixed up a few issues that cropped up after the rebase.  This should be g2g once the tests pass

PR #3238, opened 2020-03-29
fix weird focus outline on chef modals

### LINKED ISSUE #2371: chrome weird focus outline.
## Overview
When using Chrome, our delete modals and update project modals have weird focus outlines. This did not seem to be an issue in older versions of Chrome.

### Delete User
delete-user

### Delete Team
delete-team

### Delete Token
delete-token

### Update Projects
update-projects

susanev:
in chrome the focus of our chef modals was displaying super weird
i also noticed that we were putting focus on a non-interactive element (div)
this changes the focus to be on the first interactive element (close button)
which also resolves the weird focus outline issue

... and i forgot about the locked modals
so added a case for that, if anyone has a better way to do that please let me know

fixes #2371

### :+1: Definition of Done
no more weird focus outline
everything still works

`make update-ui-lib` from `components/automate-ui`

- [ ] Tests added/updated?
- [ ] Docs added/updated?

this is just one example, it was happening on all the chef modals

#### before
before

#### after
after

### DIFF
--- a/components/chef-ui-library/src/molecules/chef-modal/chef-modal.tsx
+++ b/components/chef-ui-library/src/molecules/chef-modal/chef-modal.tsx
@@ -118,7 +118,9 @@ export class ChefModal {
       // when Angular detects the modal is open
       // sets the focus on the close button for unlocked modals,
       // or the div for locked modals
-      const focusElement = (this.locked ? this.el.getElementsByClassName('modal').item(0) : this.el.getElementsByClassName('close').item(0).firstElementChild) as HTMLElement
+      const focusElement =
+        (this.locked ? this.el.getElementsByClassName('modal').item(0) :
+        this.el.getElementsByClassName('close').item(0).firstElementChild) as HTMLElement;
 
       const focusElementInterval = setInterval(() => {
         focusElement.focus();
--- a/components/chef-ui-library/src/molecules/chef-modal/chef-modal.tsx
+++ b/components/chef-ui-library/src/molecules/chef-modal/chef-modal.tsx
@@ -115,12 +115,15 @@ export class ChefModal {
     if (visible) {
       this.prevFocusedElement = document.activeElement as HTMLElement;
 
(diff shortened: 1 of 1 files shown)


msorens:
FYI: You can just use `make update-ui-lib` instead of `scripts/build_chef_ui_lib && npm run copy-ui-lib`.

components/chef-ui-library/src/molecules/chef-modal/chef-modal.tsx
```diff
@@ -115,13 +115,17 @@ export class ChefModal {
...
+      const focusElement =
+        (this.locked ? this.el.getElementsByClassName('modal').item(0) :
+        this.el.getElementsByClassName('close').item(0).firstElementChild) as HTMLElement;
+
+      const focusElementInterval = setInterval(() => {
+        focusElement.focus();
+        if (focusElement === document.activeElement) {
+          clearInterval(focusElementInterval);
         }
       }, 1);
```
msorens: We could do this more concisely unless there is a reason to be waiting until the element gets focus (which is what the conditional checks for) ...? So with this change sometimes the close button would get outlined and sometimes not, but the initial problem is still corrected.
```suggestion
      setTimeout(() => focusElement.focus(), 1);
```

msorens:
Don't quite see the rationale for your comment, @SEAjamieD . Using the ternary is more functional style coding which is usually preferred.

components/chef-ui-library/src/molecules/chef-modal/chef-modal.tsx
susanev: yah idk that code was already there? i assumed it was needed...

SEAjamieD:
> Don't quite see the rationale for your comment, @SEAjamieD . Using the ternary is more functional style coding which is usually preferred.

@msorens - I love ternary's.  I must have misunderstood some commentary on my code in the past that suggested not to use them.  Thanks for the note!

PR #3265, opened 2020-04-02
studio: remove dependency on chef/studio-common

stevendanna:

It appears that we use less than 350 of the 1800 lines of shell scripts
in studio common, including documentation comments.

While sharing code is often good, to me this is a rather small amount of code
to give up having a single view of the code that actually forms the development studio.

Further, changing code in studio-common tends to be harder because we have to worry
about multiple users of that repository.

Finally, by having all of the code in the repository, we can start to factor out the code a bit
to make it more easily sharable with scripts in the scripts/ directory, which will allow us to move
various functionality into scripts that won't be as impacted by environmental issues in the studio.


.studio/01_base.sh
```diff
@@ -0,0 +1,288 @@
+#!/bin/bash
```
stevendanna, line 1: All of this code is from studio-common or existing code in our studio.

.studio/00_logging.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
```
stevendanna, line 1: This code is a combination of what used to be in .studio/logging and studio-common.  The 00_ file naming convention helps make sure this is loaded first so everything else can depend on the logging functions without re-sourcing them.

.studio/secrets-service
```diff
@@ -215,7 +215,7 @@ function secrets_test_A1_migration() {
   hab svc start $PKG_IDENT
 
   # Wait for the secrets-service to come online
-  wait_for_svc_to_load $PKG_IDENT
+  wait_or_fail_for_svc_to_load $PKG_IDENT
```
stevendanna, line 218: `wait_for_svc_to_load` was actually deprecated in studio-common, so I guess very few people are calling this function.

.studio/tooling
```diff
@@ -152,36 +152,6 @@ function install_gomock() {
   install_go_tool "${go_mock_tools[@]}"
 }
 
-document "install_go_tool" <<DOC
```
lancewf, line 155: Why is this deleted? Should we keep this one and remove the one in 01_base.sh?

stevendanna: Mostly for load order so that we know when we can start depending on it.

.studio/01_base.sh
```diff
@@ -0,0 +1,288 @@
...
+#   2. Passing the name of a function will print out that functions documentation
+describe() {
+  if [[ -z "${CI:-}" ]]; then
+    echo
+    if [[ -z "$1" ]]; then
+      echo "The following functions are available for your use in this studio:"
+      echo
+      for file in /tmp/docs/*
+      do
+        fun="$(basename "$file")"
```
msorens, line 40: Cool! That (seemingly) nested quoting does correctly capture embedded spaces.

.studio/01_base.sh
```diff
@@ -0,0 +1,288 @@
...
+  # Ensure we are binlinking to the same version `hab pkg exec` would run
+  hab pkg binlink --force "$1" "$2" > /dev/null
+}
+
+document "wait_for_success" <<DOC
+  Wait for the given command to succeed.
+  @(arg:*) The command to run
+  Environment Variable:
+    TIMEOUT - The time in seconds to check until failing
+  Example 1
```
msorens, line 101: Would help to have a blank line before this "Example 1" header for readability.
Same for "Example 2" below.

.studio/01_base.sh
```diff
@@ -0,0 +1,288 @@
...
+  ---------
+  wait_for_success curl localhost:8080/status
+  Example 2: Wait for 60 seconds before timing out
+  ---------
+  TIMEOUT=60 wait_for_success curl localhost:8080/status
```
msorens, line 106: Maybe use something other than 60 in the example since that is the default. 🤷‍♂

.studio/01_base.sh
```diff
@@ -0,0 +1,288 @@
...
+    fi
+    (( COUNTER=COUNTER+1 ))
+  done
+}
+
+document "wait_or_fail_for_svc_to_load" <<DOC
+  Helper function to wait for a Habitat service (hab svc) to be loaded by the Habitat Supervisor.
+  @(arg:1) PKG_IDENT A Habitat package identifier (ex: core/redis)
+  @(arg:2) Number of seconds to wait before returning 1. (default: 60 seconds)
+  @(arg:3) Wheter or not this process runs silently
```
msorens, line 129:
```suggestion
  @(arg:3) Whether or not this process runs silently.
```

.studio/01_base.sh
```diff
@@ -0,0 +1,288 @@
...
+    fi
+
+    (( COUNTER=COUNTER+1 ))
+
+    [[ "${3:-}" != "silent" ]] && log_line "Waiting to load svc '$(yellow "$1")'. ($(yellow "$COUNTER of $SECONDS_WAITING"))";
+  done
+}
+
+document "wait_or_fail_for_port_to_listen" <<DOC
+  Wait for a port to be listening. If the port is not found in the set amount of time this helper returns 1.
```
msorens, line 150:
```suggestion
  Wait for a port to be listening. Returns 1 if the port is not found in the set amount of time.
```

.studio/01_base.sh
```diff
@@ -0,0 +1,288 @@
...
+  @(arg:1) Port to wait for to be listening
+  @(arg:2) Number of seconds to wait before returning 1. (default: 60 seconds)
+  @(arg:3) Wheter or not this process runs silently
```
msorens, line 153:
```suggestion
  @(arg:3) Whether or not this process runs silently.
```

.studio/01_base.sh
```diff
@@ -0,0 +1,288 @@
...
+      return 1
+    fi
+
+    (( COUNTER=COUNTER+1 ))
+    [[ "${3:-}" != "silent" ]] && log_line "Waiting for port '$(yellow "$1")' to be listening ($(yellow "$COUNTER of $SECONDS_WAITING"))";
+  done
+}
+
+document "wait_for_ok_response" <<DOC
+  Wait a set amount of time for the provided URL to response with status code (200).
```
msorens, line 180:
```suggestion
  Wait a set amount of time for the provided URL to respond with status code (200).
```

.studio/01_base.sh
```diff
@@ -0,0 +1,288 @@
...
+  local COUNTER=0
+  local response
+
+  while [[ $COUNTER -le "$SECONDS_TO_WAIT"  ]]; do
+    # shellcheck disable=SC1083
+    response="$(curl --write-out %{http_code} --silent --output /dev/null "$url")"
+    if [[ $response -eq $code ]]; then
+      break
+    else
+      log_line "Waiting for '$url' to response OK ($code). [Got:$response] ($(yellow "$COUNTER of $SECONDS_TO_WAIT"))"
```
msorens, line 206:
```suggestion
      log_line "Waiting for '$url' to respond OK ($code). [Got:$response] ($(yellow "$COUNTER of $SECONDS_TO_WAIT"))"
```

.studio/01_base.sh
```diff
@@ -0,0 +1,288 @@
...
+    fi
+  done
+}
+
+document "install_go_tool" <<DOC
+  Install the specified go tool(s).
+  @(arg:*) The array of packages you wish to install.
+  The default behavior is to install go tools with 'go install -v -mod=vendor'
+
+  To install tools using not included in the vendor you'll either need to
```
msorens, line 222: Not quite sure of the intent, maybe...
```suggestion
  To install tools not included in the vendor directory, you'll either need to
```

.studio/01_base.sh
```diff
@@ -0,0 +1,288 @@
...
+    -I\$GOPATH/src \\
+    --go_out=plugins=grpc:. \\
+    proto/*.proto
+  ------------------------------------------------------------------
+DOC
+compile_go_protobuf() {
+  local rc
+  proto_script="${1:-scripts/grpc.sh}"
+
+  # Verify that the script exist and it is an executable
```
msorens, line 254:
```suggestion
  # Verify that the script exists and is executable.
```

.studiorc
```diff
@@ -211,6 +212,31 @@ check_if_deployinate_started() {
 # ```
 # export HAB_STUDIO_SECRET_GITHUB_TOKEN=secret
 # ```
+document "generate_netrc_config" <<DOC
+  Create a .netrc file that can be used to authenticate against GitHub.
+  Some projects require access to private GitHub repositories. The recommended
+  pattern is for projects to use the git+https protocol in conjuction with a
```
msorens, line 218:
```suggestion
  pattern is for projects to use the git+https protocol in conjunction with a
```

PR #3267, opened 2020-04-02
Improved errors for data-feed-service API

satellite15:

Failed API calls to the data-feed-service were returning 500 in all error cases, I have updated the API to return 404 for update, get, delete operations with an id that doesn't correspond to a datafeed destination record

- [x] Tests added/updated? tested in service now dev env and with curl
- [ ] Docs added/updated? docs TBD 

### DIFF
--- a/components/data-feed-service/dao/destination.go
+++ b/components/data-feed-service/dao/destination.go
@@ -1,6 +1,7 @@
 package dao
 
 import (
+	"database/sql"
 	"strconv"
 
 	datafeed "github.com/chef/automate/api/external/data_feed"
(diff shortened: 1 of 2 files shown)


satellite15:
@vinay033 status codes have changed for some error conditions in the API. Just making you aware in case this impacts the work you are currently doing

components/data-feed-service/dao/destination.go
```diff
@@ -27,7 +28,7 @@ func addToDBDestination(inDestination *datafeed.AddDestinationRequest) *Destinat
 
 func updateToDBDestination(inDestination *datafeed.UpdateDestinationRequest) *Destination {
 	newDestination := Destination{}
-	newDestination.ID, _ = strconv.ParseInt(inDestination.Id, 10 , 64)
+	newDestination.ID, _ = strconv.ParseInt(inDestination.Id, 10, 64)
```
vjeffrey, line 31: why no err check here? is it bc we're assuming that if it's a bad value we would then end up with no update and just return the not found error? would there be any value to checking for the parseint error and returning an err if that doesn't work?  is inDestination.Id a user provided value?

satellite15: That's existing code, it's only a format change from make fmt. 

The inDestination.Id will be provided by the UI thus an existing valid id or any hand crafted external call to the API, this is likely the only case where it could be a non-integer string. The service now app will also call the API as it can configure the datafeed from the service now UI  but that again will be a valid id.  If it's not an integer I just decided to let the database error cascade back, rather than adding additional error handling

PR #3276, opened 2020-04-02
[backup] Add basic artifact repository integrity checking

ryancragun:
The shared artifact snapshot repository is used for builder artifact
backup operations. The current implementation determines which artifacts
_should_ exist in the shared repository by aggregating all existing
snapshots' required artifacts together. It compares the aggregate list
of assumed repo artifacts with the list required by the snapshot being
take. The speed of this check is desirable as it doesn't require checking
the entire repository for each backup, however, it also does ensure that
or validate that all of the assumed artifacts are in the repository.

This change introduces basic shares repository integrity validation
checking. It scans the backup artifact repository and validates any
and/or all snapshots with the actual artifacts and saves the latest
validation check results in the snapshot bucket.

`chef-automate backup integrity validate` will validate all snapshots in
the repository. A user may also choose to exclusively validate one or
more snapshots by passing backup ID's as variable arguments to the
command.

Example output:
```
chef-automate backup integrity validate 20200401000216
Backup           Last Verification  Corrupted   Missing Packages 
20200401000216   20200402212720                                  
20200401000703   20200402202118                                  
20200401202734   20200402212658                                  
20200401203021   20200402202118                                  
20200401204633   2020040220211
```

`chef-automate backup integrity show` will pretty print the current
snapshot validation metadata. Eventually it might be nice to allow this
command to take backup ID and to show all the details of a snapshot,
including which artifacts are missing if it is corrupted.

Example Output:
```
chef-automate backup integrity show  
Backup           Last Verification  Corrupted   Missing Packages 
20200401000216   20200402202044                                  
20200401000703   20200402202044                                  
20200401202734   20200402202044                                  
20200401203021   20200402202044     true        1                
20200401204633   20200402202044     true        1
```

In the future we should look to add additional commands like `repair`
which will validate the snapshot and attempt to upload any missing
artifacts from the existing builder bucket.

This change also includes a new builder scaffolding tool called
`generate real` which is a one stop shop for priming a a development
builder instance with a package.

`go run ./tools/builder-scaffolding/... generate real ./tools/builder-scaffolding/habitat/ `
will automatically build the shim package `builder-scaffolding` and
upload it to builder. It takes care of hab key, tokens, login, origins etc.

To facilitate a few functions in the new generation tool
`fileutils.CopyFile()` and `fileutils.CopyDir()` have been added.

The `ReplayableStream` stream type has been added which allows the same
stream to be repeatably read over and over. It's syntactic sugar over
existing streams so we can cache long I/O stream reads and replay it
locally many times. It was used to handle reading the remote bucket
objects once and replaying it for each snapshot that is validated.


ryancragun:
Resolves https://github.com/chef/automate/issues/2669

components/automate-cli/cmd/chef-automate/backup.go
```diff
@@ -192,6 +198,26 @@ var cancelBackupCmd = &cobra.Command{
...
+	Short: "show the shared object integrity metadata",
+	Long:  "Show the shared object integrity metadata",
+	RunE:  runBackupIntegrityShowCmd,
+	Args:  cobra.ExactArgs(0),
+}
+
+var integrityBackupValidateCmd = &cobra.Command{
+	Use:   "validate [ID IDN]",
+	Short: "validate the shared object integrity",
+	Long:  "Validate the shared object integrity. Arguments are assumed to be backup ID's that you wish to validate, if left empty it will validate all snapshots",
```
yzl, line 217: ID's -> IDs

[nit] The long usage message reads as a run-on sentence to me, so I'd probably use a semicolon instead of a comma after "validate".

components/automate-deployment/pkg/backup/artifact_streams.go
```diff
@@ -437,6 +443,123 @@ func NewCountingStream(stream ArtifactStream) CountingStream {
 	}
 }
 
+// ReplayableStream is a stream that consumes a stream to a file and allows
+// it to be replayed several times
```
yzl, line 447: "several" makes it sound like there is a limit to the number of times you can replay a stream. Is that the case?

components/automate-deployment/pkg/client/backup.go
```diff
@@ -237,6 +237,62 @@ func DeleteBackups(conTimeout, reqTimeout time.Duration, ids []*api.BackupTask)
...
+			err,
+			status.DeploymentServiceCallError,
+			"Request to cancel backup failed",
+		)
+	}
+
+	return res, err
+}
+
+// ValidateBackupIntegrity makes a gRPC request to the deployment-service and returns
```
yzl, line 267: [nit] This is the same comment as for BackupIntegrityShow() above.

components/automate-deployment/pkg/backup/artifact_streams.go
ryancragun: There's no limit, you just need to call `Reset()` in order to get a new stream.

components/automate-deployment/pkg/backup/artifact_repo.go
```diff
@@ -317,26 +330,23 @@ func (repo *ArtifactRepo) Remove(ctx context.Context, name string) error {
...
-	snapshots, _, err := repo.snapshotsRoot.List(ctx, "", false)
-	if err != nil {
-		return ErrStream(err)
-	}
+// ListAssumedArtifacts aggregates all the required artifacts from all snapshots
+// and returns a stream. It is presumed to be much cheaper to scan every snapshot
+// metadata file and aggregate the artifacts rather than list every object that
+// is currently in the repo. The downside of course is that we aren't sure that
+// such objects exist.
+func (repo *ArtifactRepo) ListAssumedArtifacts(ctx context.Context, excludedSnapshots ...string) ArtifactStream {
```
jaym, line 338: how about `ListArtifactsFromSnapshots`

components/automate-deployment/pkg/backup/artifact_repo.go
```diff
@@ -353,9 +363,215 @@ func (repo *ArtifactRepo) ListArtifacts(ctx context.Context, excludedSnapshots .
 	if lastMergedStream == nil {
 		return EmptyStream()
 	}
+
 	return lastMergedStream
 }
 
+// ListArtifacts returns a stream of the actual object names from the bucket.
+func (repo *ArtifactRepo) ListArtifacts(ctx context.Context) ArtifactStream {
```
jaym, line 371: maybe `ListArtifactsInRepo`

components/automate-deployment/pkg/backup/artifact_repo.go
```diff
@@ -432,9 +650,65 @@ func (repo *ArtifactRepo) openSnapshotFile(ctx context.Context, name string) (Ar
...
 
+func (repo *ArtifactRepo) openSnapshotFile(ctx context.Context, name string) (ArtifactStream, string, error) {
+	return repo.openGzipFile(ctx, toSnapshotName(name))
+}
+
+// ReadSnapshotIntegrityMetadata reads the snapshot integrity metadata file
+func (repo *ArtifactRepo) ReadSnapshotIntegrityMetadata(ctx context.Context) (*ArtifactRepoIntegrityMetadata, error) {
+	metadata := NewArtifactRepoIntegrityMetadata()
+
+	metadataStream, _, err := repo.openGzipFile(ctx, SnapshotIntegrityMetadataFileName)
```
jaym, line 661: I dont think artifact stream is the right thing to use here. The artifact streams are line oriented files that are used to deal with a large list of things. This should be reading the json.gz file. I'm sort of not clear why this even works as it would strip newlines and corrupt the gzip?

jaym: It seems to be because LoggingStream writes line oriented files

components/automate-deployment/pkg/backup/artifact_repo.go
```diff
@@ -353,9 +363,215 @@ func (repo *ArtifactRepo) ListArtifacts(ctx context.Context, excludedSnapshots .
...
+func NewArtifactRepoSnapshotIntegrityMetadata() ArtifactRepoSnapshotIntegrityMetadata {
+	return ArtifactRepoSnapshotIntegrityMetadata{
+		Missing: []string{},
+	}
+}
+
+// updateSnapshotIntegrityMetadata consumes a missing artifact stream for a given
+// snapshot and pdates an in memory repo integrity metadata with a new snapshot
+// integrity metadata result.
+func (repo *ArtifactRepo) updateSnapshotIntegrityMetadata(
```
jaym, line 521: this method should probably be on ArtifactRepoIntegrityMetadata?

components/automate-deployment/pkg/backup/artifact_repo.go
```diff
@@ -353,9 +363,215 @@ func (repo *ArtifactRepo) ListArtifacts(ctx context.Context, excludedSnapshots .
...
+	ctx context.Context,
```
jaym, line 522: ctx isnt used

components/automate-deployment/pkg/backup/artifact_streams.go
```diff
@@ -437,6 +443,123 @@ func NewCountingStream(stream ArtifactStream) CountingStream {
...
+// NewBucketObjectNameStream takes a bucket and returns a streams of all object
+// names for the given path
+func NewBucketObjectNameStream(ctx context.Context, bucket Bucket, pathPrefix string) (ArtifactStream, error) {
+	objects, _, err := bucket.List(ctx, pathPrefix, false)
+	if err != nil {
+		return nil, err
+	}
+	names := []string{}
+	for _, obj := range objects {
+		if filepath.Ext(obj.Name) == ".hart" {
```
jaym, line 553: why do we need to check for harts

components/automate-deployment/pkg/backup/artifact_streams.go
```diff
@@ -437,6 +443,123 @@ func NewCountingStream(stream ArtifactStream) CountingStream {
...
+			names = append(names, obj.Name)
+		}
+	}
+
+	sort.Strings(names)
```
jaym, line 558: if bucket.List isn't already sorted, i think a lot of this code is broken. I'll dig into this more

jaym: >  Amazon S3 exposes a list operation that lets you enumerate the keys contained in a bucket. Keys are selected for listing by bucket and prefix. For example, consider a bucket named "dictionary" that contains a key for every English word. You might make a call to list all the keys in that bucket that start with the letter "q". List results are always returned in UTF-8 binary order.

jaym: i think the filesystem implementation might not be sorted.

PR #3282, opened 2020-04-03
[Automate-2843] introspect get projects to improve UX

msorens:

We want to improve the user experience for the case of creating a brand new user that has not yet been placed into any teams or policies. Previously, you would immediately see one error bar like that shown, and at times, they could cascade down the page.
image

So we are applying introspection to determine if we can make the GetProjects call at all.

Points addressed:
1. **Initial login or page refresh.** The page is constructed and, during initialization, it makes a call to GetProjects to be able to determine whether there are any edits pending and thus whether the <kbd>edits pending</kbd> bar should be displayed.
2. **Periodically to keep up-to-date.** As Automate is a multi-user application, actions of one user may affect what other users do. Specifically, if one user edits a project or rule, that needs to reflect in the  <kbd>edits pending</kbd> bar not just for that user, but for all users. So every two minutes each connected browser polls with a GetProjects call.
3. **Rendering the <kbd>edits pending</kbd> bar itself.** The bar displays in varying forms for any of 3 reasons: rules have been edited, a project update has been cancelled by a user, or a project update has failed. While the second and third are still knowable for a user with no permissions, the first relies on GetProjects, so is not knowable. We decided to therefore suppress the <kbd>edits pending</kbd> bar entirely as the most reasonable UX. 
4. Display the <kbd>update in progress</kbd> bar in all cases. This does not rely on GetProjects, so is safe (and useful) to display.

NA

### :+1: Definition of Done
If the user does not have permission to call `GetProjects`, then `GetProjects` is not called (a) upon login, (b) upon page refresh, or (c) periodically (as described above). Should one navigate directly to team details, token details, or project list (which issue explicit GetProject calls), the usual error will still appear, and rightly so.

(Note should test on pages OTHER THAN team details, token details, or project list--those pages fetch the project list explicitly)

Login as admin.
```
* GET /iam/v2/projects occurs during page construction on any page. 
* Every two minutes another GET /iam/v2/projects occurs to refresh projects. 
* "Edits pending" bar appears when rules are edited, when update fails, or when update is canceled.
* "Update in progress" bar appears when project update is in progress.
```

Create a new user with no permissions. Login as this user.
```
* GET /iam/v2/projects DOES NOT occur during page construction on any page.
* Every two minutes -- nothing happens. 
* "Edits pending" bar DOES NOT appear when rules are edited, when update fails, or when update is cancelled.
* "Update in progress" bar APPEARS when project update is in progress.
```

As an admin, add that new user to the viewers team, then refresh the user's browser.
```
* GET /iam/v2/projects occurs during page construction on any page.
* Every two minutes another GET /iam/v2/projects occurs to refresh projects. 
* "Edits pending" bar appears when rules are edited, when update fails, or when update is cancelled.
* "Update in progress" bar appears when project update is in progress.
```

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/components/pending-edits-bar/pending-edits-bar.component.spec.ts
+++ b/components/automate-ui/src/app/components/pending-edits-bar/pending-edits-bar.component.spec.ts
@@ -9,8 +9,11 @@ import { NgrxStateAtom, runtimeChecks, ngrxReducers } from 'app/ngrx.reducers';
 import { ChefPipesModule } from 'app/pipes/chef-pipes.module';
 import { using } from 'app/testing/spec-helpers';
 import { FeatureFlagsService } from 'app/services/feature-flags/feature-flags.service';
-import { Project } from 'app/entities/projects/project.model';
+import { IndexedEntities } from 'app/entities/entities';
+import { UserPermEntity } from 'app/entities/userperms/userperms.entity';
+import { PermEntityState, Status } from 'app/entities/userperms/userperms.reducer';
 import { ProjectStatus } from 'app/entities/rules/rule.model';
+import { Project } from 'app/entities/projects/project.model';
 import { ProjectService } from 'app/entities/projects/project.service';
 import { ApplyRulesStatusState } from 'app/entities/projects/project.reducer';
 import {
@@ -28,60 +31,9 @@ describe('PendingEditsBarComponent', () => {
   let store: Store<NgrxStateAtom>;
 
   beforeEach(async(() => {
-
-    TestBed.configureTestingModule({
-      declarations: [
-        MockComponent({
-          selector: 'chef-toolbar',
-          template: '<ng-content></ng-content>'
-        }),
-         MockComponent({
-          selector: 'app-authorized',
-          inputs: ['allOf', 'not'],
-          template: '<ng-content></ng-content>'
-        }),
-        MockComponent({
-          selector: 'app-message-modal',
-          inputs: ['visible' ],
-          outputs: ['close' ]
-        }),
-         MockComponent({
-          selector: 'app-confirm-apply-start-modal',
-          inputs: ['visible'],
-          outputs: ['confirm', 'cancel']
-        }),
-        MockComponent({
-          selector: 'app-confirm-apply-stop-modal',
-          inputs: ['visible', 'applyRulesStatus', 'stopRulesInProgress'],
-          outputs: ['confirm', 'cancel']
-        }),
-        MockComponent({ selector: 'mat-progress-bar', inputs: ['mode', 'value', 'bufferValue'] }),
-        MockComponent({ selector: 'mat-select' }),
-        MockComponent({ selector: 'mat-option' }),
-        MockComponent({ selector: 'chef-button', inputs: ['disabled'] }),
-        MockComponent({ selector: 'chef-heading' }),
-        MockComponent({ selector: 'chef-loading-spinner' }),
-        MockComponent({ selector: 'chef-page-header' }),
-        MockComponent({ selector: 'chef-subheading' }),
-        MockComponent({ selector: 'chef-table-new' }),
-        MockComponent({ selector: 'chef-table-header' }),
-        MockComponent({ selector: 'chef-table-body' }),
-        MockComponent({ selector: 'chef-table-row' }),
-        MockComponent({ selector: 'chef-table-header-cell' }),
-        MockComponent({ selector: 'chef-table-cell' }),
-        PendingEditsBarComponent
-      ],
-      imports: [
-        ReactiveFormsModule,
-        RouterTestingModule,
-        ChefPipesModule,
-        StoreModule.forRoot(ngrxReducers, { runtimeChecks })
-      ],
-      providers: [
-        FeatureFlagsService,
-        ProjectService
-      ]
-    }).compileComponents();
+      configureWith({
+        '/iam/v2/projects': genPerm('/iam/v2/projects', true)
+      });
   }));
 
   beforeEach(() => {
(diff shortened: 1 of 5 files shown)


components/automate-ui/src/app/components/pending-edits-bar/pending-edits-bar.component.spec.ts
```diff
@@ -28,60 +31,9 @@ describe('PendingEditsBarComponent', () => {
...
-        StoreModule.forRoot(ngrxReducers, { runtimeChecks })
-      ],
-      providers: [
-        FeatureFlagsService,
-        ProjectService
-      ]
-    }).compileComponents();
+      configureWith({
+        '/iam/v2/projects': genPerm('/iam/v2/projects', true)
+      });
```
msorens, line 36: Needed to be able to apply different configurations in `TestBed.configureTestingModule` so moved it to a helper function. Notice that in this call, we are showing the user has permission to do a `GetProjects` call. This is the data used by introspection to report that.

components/automate-ui/src/app/components/pending-edits-bar/pending-edits-bar.component.spec.ts
```diff
@@ -219,6 +171,156 @@ describe('PendingEditsBarComponent', () => {
...
 
+describe('PendingEditsBarComponent--unauthorized', () => {
+  let component: PendingEditsBarComponent;
+  let fixture: ComponentFixture<PendingEditsBarComponent>;
+  let store: Store<NgrxStateAtom>;
+
+  beforeEach(async(() => {
+    configureWith({
+      '/iam/v2/projects': genPerm('/iam/v2/projects', false)
+    });
```
msorens, line 182: This new set of tests is for the unauthorized case. Notice the TestBed is configured with `false` here. Then the exact same set of tests earlier in the file (not shown in the diff here because they were not technically not changed), which caused the <kbd>edits pending</kbd> bar to be visible, will now show that the bar is hidden.

components/automate-ui/src/app/components/pending-edits-bar/pending-edits-bar.component.spec.ts
```diff
@@ -219,6 +171,156 @@ describe('PendingEditsBarComponent', () => {
...
+        genState(ApplyRulesStatusState.Running, false, true)));
+      component.updateDisplay();
+      expect(component.isBarHidden).toEqual(true);
+    });
+  });
+});
+
+
+function configureWith(perms: IndexedEntities<UserPermEntity>): void {
+  TestBed.configureTestingModule({
```
msorens, line 258: This `TestBed` configuration was moved into this helper verbatim...

components/automate-ui/src/app/components/pending-edits-bar/pending-edits-bar.component.spec.ts
```diff
@@ -219,6 +171,156 @@ describe('PendingEditsBarComponent', () => {
...
+    imports: [
+      ReactiveFormsModule,
+      RouterTestingModule,
+      ChefPipesModule,
+      StoreModule.forRoot({
+        ...ngrxReducers,
+          userperms: () => <PermEntityState>{
+            status: Status.loadingSuccess,
+            byId: perms
+          }
```
msorens, line 309: ... with this one change: allowing the `userperms` portion of the ngrx store to be configured by the passed-in `perms` parameter.

components/automate-ui/src/app/components/pending-edits-bar/pending-edits-bar.component.ts
```diff
@@ -22,28 +24,36 @@ export class PendingEditsBarComponent implements OnDestroy {
...
         this.updateDisplay();
       });
-    }
+
+    this.store.select(allPerms)
+      .pipe(takeUntil(this.isDestroyed))
+      .subscribe(perms =>
+        this.isAuthorized = get(['/iam/v2/projects', 'get'], perms)
+      );
+  }
```
msorens, line 56: Query introspection from the ngrx store.

components/automate-ui/src/app/components/pending-edits-bar/pending-edits-bar.component.ts
```diff
@@ -52,9 +62,10 @@ export class PendingEditsBarComponent implements OnDestroy {
 
   public updateDisplay() {
     this.layoutFacade.layout.userNotifications.pendingEdits =
-      this.projectsHaveStagedChanges
-      || this.updateProjectsCancelled
-      || this.updateProjectsFailed;
+      this.isAuthorized &&
```
msorens, line 65: ... and use that result here. See note under `isBarHidden` as to why it needs to be precisely here.

components/automate-ui/src/app/entities/layout/layout.facade.ts
```diff
@@ -53,15 +57,24 @@ export class LayoutFacadeService {
...
     this.updateDisplay();
-    this.store.dispatch(new GetProjects());
+
+    this.authorizedChecker = new AuthorizedChecker(fullStore);
+    this.authorizedChecker.setPermissions([{ endpoint: '/iam/v2/projects', verb: 'get' }], []);
+    this.authorizedChecker.isAuthorized$
+      .pipe(filter(identity), first())
+      .subscribe(
+        // wait for permission and, if it comes at all, fire exactly once
+        () => this.store.dispatch(new GetProjects()));
```
msorens, line 77: This GetProjects call occurs during page construction on application startup/page refresh.
Note that the `pipe` operators are crucial here.
First, `filter(identity)` translates to "wait until endpoint `/iam/v2/projects` has method `get` set to `true`. That, will never happen for a user with no permissions. And that is OK. But for a user with permission, it will happen in fairly short order. Once it does, invoke `GetProjects` and then _guarantee we never do it again_! That is handled by the `first()` call   which completes the observable, thereby effectively subscribing, grabbing a single value, and unsubscribing automatically. (Without the `first()`, if you happen to be on the ingest rules tab of a project page, it will flicker out-of-control.)

components/automate-ui/src/app/entities/projects/project.effects.ts
```diff
@@ -212,14 +214,16 @@ export class ProjectEffects {
...
     ),
-    switchMap(() => [
-      new GetProjects(),
-      new GetApplyRulesStatus()
-    ]),
+    switchMap(([[_interval, list], _status]) => {
+      return get(['/iam/v2/projects', 'get'], list) ?
+        [new GetProjects(), new GetApplyRulesStatus()] :
+        [new GetApplyRulesStatus()];
+    }),
```
msorens, line 226: This does the periodic polling of GetProjects. This uses introspection yet a third time, and if there is permission does both GetProjects and GetApplyRulesStatus, otherwise just GetApplyRulesStatus.

components/automate-ui/src/app/helpers/auth/authorized.ts
```diff
@@ -22,7 +22,7 @@ import { allPerms } from 'app/entities/userperms/userperms.selectors';
 export interface CheckObj {
   endpoint: string;
   verb: string;
-  paramList: string | string[];
+  paramList?: string | string[];
```
msorens, line 25: Should have really been this. Simplifies creating CheckObj objects.

components/automate-ui/src/app/entities/layout/layout.facade.ts
bcmdarroch: I've forgotten why this GetProjects is fired here in the first place. To populate the project filter?

Issue #3297, opened 2020-04-07
Wrong error reported for non-existent server


msorens:
## Describe the bug
Attempting to create an organization for a chef server that does not exist reports an incorrect error.
```
"error": "no org found with id \"foo\"",
```

## To Reproduce

Attempt to create an organization using a non-existent--but well-formed GUID--as a server ID:
```
$ curl -sSkH "api-token: $TOK" -X POST \
-d '{"name": "foo", "admin_user": "ADMIN_USER", "admin_key": "KEY"}' \
https://a2-dev.test/api/v0/infra/servers/b900bfa6-c699-4b34-bb4e-146901de1164/orgs?pretty
{
  "error": "no org found with id \"foo\"",
  "code": 5,
  "message": "no org found with id \"foo\"",
  "details": [
  ]
}
```
Two problems:
1. "foo" is the name, not the id
2. it should say "server not found" rather than "no org found" and include the suspect ID, for example:
```
"error": "b900bfa6-c699-4b34-bb4e-146901de1164: server not found"
```

## Expected behavior
The error should indicate the server does not exist.

vsingh-msys:
This should definitely a bug, thanks for reporting it.

vsingh-msys:
Fixes #3297 vai #3298

Issue #3294, opened 2020-04-07
Problem deleting chef server with attached orgs


msorens:
## Describe the bug
There is a problem deleting a chef server that has one or more organizations attached. I am not sure what is supposed to happen, but a fix is needed either way.

## To Reproduce

Create a server:
```
$ curl -sSkH "api-token: $TOK" -X POST \
-d '{"name": "SERVER_NAME2", "description": "SERVER_DESCRIPTION2", "fqdn": "example.com", "ip_address": "127.0.0.1"}' \
https://a2-dev.test/api/v0/infra/servers?pretty
{
  "server": {
    "id": "d31e49ce-608b-46d4-b8b5-a819d19222ad",
    "name": "SERVER_NAME2",
    "description": "SERVER_DESCRIPTION2",
    "fqdn": "example.com",
    "ip_address": "127.0.0.1",
    "orgs_count": 0
  }
}
```
Grab the `id` value from above and plug in here to add an organization:
```
$ curl -sSkH "api-token: $TOK" -X POST \
-d '{"name": "ORG_NAME2", "admin_user": "ADMIN_USER", "admin_key": "KEY"}' \
https://a2-dev.test/api/v0/infra/servers/d31e49ce-608b-46d4-b8b5-a819d19222ad/orgs?pretty
{
  "org": {
    "id": "956c1773-d5ba-471b-8e0d-ceb37603e271",
    "name": "ORG_NAME2",
    "admin_user": "ADMIN_USER",
    "admin_key": "e2ac8d74-b419-42b3-b96c-aa2f738e8f18",
    "server_id": "d31e49ce-608b-46d4-b8b5-a819d19222ad",
    "projects": [
    ]
  }
}
```
Attempt to delete the server and it fails with, at a minimum, the wrong error message:
```
$ curl -sSkH "api-token: $TOK" -X DELETE \
https://a2-dev.test/api/v0/infra/servers/d31e49ce-608b-46d4-b8b5-a819d19222ad?pretty
{
  "error": "no server found with id \"d31e49ce-608b-46d4-b8b5-a819d19222ad\"",
  "code": 5,
  "message": "no server found with id \"d31e49ce-608b-46d4-b8b5-a819d19222ad\"",
  "details": [
  ]
}
```
Delete org first:
```
$ curl -sSkH "api-token: $TOK" -X DELETE \
[...]

PR #3298, opened 2020-04-07
[infra-proxy-service] More specific foreign_key_violation  error messages.

vsingh-msys:

 - Add ErrCannotDelete & ErrForeignKeyViolation custom errors.
 - In order to deal with foreign_key_violation error for invalid & delete cascade partial string match with error object detail.
 - Earlier only error code name was used to distinguish between errors but it was returning same for both the cases like invalid foreign key and delete parent record if child exists.

 - https://github.com/chef/automate/issues/3297
 - https://github.com/chef/automate/issues/3294

### :+1: Definition of Done
 - If a user attempts to delete a chef server and if it has orgs now it raises the error:
```
"cannot delete server _<server-guid>_ because it still has organizations attached"
```
- If non-existed server ID is provided while creating the organization now it raises an error:

```
"no server found with ID _<server-guid>_"
```

Steps to build:
- Check out this PR
- `rebuild components/infra-proxy-service`

Steps to test:
 - Load sample data `infra_service_load_sample_data` or Use the following curl command to add data.

 - Create a server:
```
$ curl -sSkH "api-token: $TOK" -X POST \
-d '{"name": "SERVER_NAME", "description": "DESCRIPTION", "fqdn": "example.com", "ip_address": "127.0.0.1"}' \
https://a2-dev.test/api/v0/infra/servers?pretty
{
  "server": {
    "id": "670e1cdf-55ab-4d78-9350-dcfe27f82ca4",
    "name": "SERVER_NAME",
    "description": "DESCRIPTION",
    "fqdn": "example.com",
    "ip_address": "127.0.0.1",
    "orgs_count": 0
  }
}
```
Add an organization consider `server_id` as from above created server:
```
$ curl -sSkH "api-token: $TOK" -X POST \
-d '{"name": "ORG_NAME", "admin_user": "ADMIN_USER", "admin_key": "KEY"}' \
https://a2-dev.test/api/v0/infra/servers/670e1cdf-55ab-4d78-9350-dcfe27f82ca4/orgs?pretty
{
  "org": {
    "id": "956c1773-d5ba-471b-8e0d-ceb37603e271",
    "name": "ORG_NAME",
    "admin_user": "ADMIN_USER",
    "admin_key": "e2ac8d74-b419-42b3-b96c-aa2f738e8f18",
    "server_id": "670e1cdf-55ab-4d78-9350-dcfe27f82ca4",
    "projects": [
    ]
  }
}
```
### Fixed: Attempt to delete the server:

```
$ curl -sSkH "api-token: $TOK" -X DELETE \
https://a2-dev.test/api/v0/infra/servers/670e1cdf-55ab-4d78-9350-dcfe27f82ca4?pretty
{
  "error": "cannot delete server \"670e1cdf-55ab-4d78-9350-dcfe27f82ca4\" because it still has organizations attached",
  "code": 9,
  "message": "cannot delete server \"670e1cdf-55ab-4d78-9350-dcfe27f82ca4\" because it still has organizations attached",
  "details": [
  ]
}
```

### Fixed: Attempt to add org with non-existed server-id
```
curl -sSkH "api-token: $TOK" -X POST \
  -d '{"name": "ORG_NAME", "admin_user": "ADMIN_USER", "admin_key": "KEY"}' \
  https://a2-dev.test/api/v0/infra/servers/1d0b8c0d-a40d-4e69-83db-a597e715c231/orgs?pretty
{
  "error": "no server found with ID \"1d0b8c0d-a40d-4e69-83db-a597e715c231\"",
  "code": 5,
  "message": "no server found with ID \"1d0b8c0d-a40d-4e69-83db-a597e715c231\"",
  "details": [
  ]
}
```

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/infra-proxy-service/server/servers_test.go
+++ b/components/infra-proxy-service/server/servers_test.go
@@ -335,6 +335,47 @@ func TestServers(t *testing.T) {
 			assert.Equal(t, len(serverListBefore.Servers)-1, len(serverListAfter.Servers))
 		})
 
+		t.Run("when the server exists with orgs, raise server can not be deleted error", func(t *testing.T) {
+			secretsMock.EXPECT().Create(gomock.Any(), &newSecret, gomock.Any()).Return(secretID, nil)
+			secretsMock.EXPECT().Read(gomock.Any(), secretID, gomock.Any()).Return(&secretWithID, nil)
+			secretsMock.EXPECT().Delete(gomock.Any(), secretID, gomock.Any())
+
+			resp1, err := cl.CreateServer(ctx, &request.CreateServer{
+				Name:        "chef-infra-server1",
+				Description: "Chef infra server",
+				Fqdn:        "domain1.com",
+				IpAddress:   "10.0.0.1",
+			})
+			require.NoError(t, err)
+			require.NotNil(t, resp1)
+
+			respOrg, err := cl.CreateOrg(ctx, &request.CreateOrg{
+				Name:      "infra-org",
+				AdminUser: "admin",
+				AdminKey:  "--KEY--",
+				ServerId:  resp1.Server.Id,
+				Projects:  []string{},
+			})
+			require.NoError(t, err)
+			require.NotNil(t, respOrg)
+
+			serverListBefore, err := cl.GetServers(ctx, &request.GetServers{})
+			require.NoError(t, err)
+			assert.Equal(t, 1, len(serverListBefore.Servers))
+
+			resp, err2 := cl.DeleteServer(ctx, &request.DeleteServer{Id: resp1.Server.Id})
+			require.Error(t, err2)
+			require.Nil(t, resp)
+			grpctest.AssertCode(t, codes.FailedPrecondition, err2)
+
+			serverListAfter, err3 := cl.GetServers(ctx, &request.GetServers{})
+			require.NoError(t, err3)
+			assert.Equal(t, len(serverListBefore.Servers), len(serverListAfter.Servers))
+
+			cleanupOrg(ctx, t, cl, respOrg.Org.Id)
+			cleanupServer(ctx, t, cl, resp1.Server.Id)
+		})
+
 		t.Run("when the server ID for the server to delete is empty, raise an invalid argument error", func(t *testing.T) {
 			resp, err := cl.GetServer(ctx, &request.GetServer{
 				Id: "",
--- a/components/infra-proxy-service/server/servers_test.go
+++ b/components/infra-proxy-service/server/servers_test.go
@@ -2,6 +2,7 @@ package server_test
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	"github.com/golang/mock/gomock"
@@ -235,11 +236,13 @@ func TestServers(t *testing.T) {
 		})
 
 		t.Run("when the server does not exist, return server not found", func(t *testing.T) {
+			serverID := "97e01ea1-976e-4626-88c8-43345c5d934f"
 			resp, err := cl.GetServer(ctx, &request.GetServer{
-				Id: "97e01ea1-976e-4626-88c8-43345c5d934f",
+				Id: serverID,
 			})
 
 			require.Nil(t, resp)
+			require.Contains(t, err.Error(), fmt.Sprintf("no server found with ID \"%s\"", serverID))
 			grpctest.AssertCode(t, codes.NotFound, err)
 		})
 
@@ -366,6 +369,7 @@ func TestServers(t *testing.T) {
 			resp, err2 := cl.DeleteServer(ctx, &request.DeleteServer{Id: resp1.Server.Id})
 			require.Error(t, err2)
 			require.Nil(t, resp)
+			assert.Regexp(t, "cannot delete server.*still has organizations attached", err2.Error())
 			grpctest.AssertCode(t, codes.FailedPrecondition, err2)
 
 			serverListAfter, err3 := cl.GetServers(ctx, &request.GetServers{})
--- a/components/infra-proxy-service/service/service.go
+++ b/components/infra-proxy-service/service/service.go
@@ -1,6 +1,8 @@
 package service
 
 import (
+	"reflect"
+
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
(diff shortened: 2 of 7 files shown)


components/infra-proxy-service/service/service.go
```diff
@@ -41,13 +43,17 @@ func Start(l logger.Logger, migrationsConfig migration.Config, connFactory *secu
...
 // ParseStorageError parses common storage errors into a user-readable format.
-func ParseStorageError(err error, id interface{}, noun string) error {
+func ParseStorageError(err error, v interface{}, noun string) error {
 	if err != nil {
 		switch err {
 		case storage.ErrNotFound:
-			return status.Errorf(codes.NotFound, "no %s found with id %q", noun, id)
+			return status.Errorf(codes.NotFound, "no %s found with ID %q", noun, reflect.Indirect(reflect.ValueOf(v)).FieldByName("Id"))
+		case storage.ErrCannotDelete:
+			return status.Errorf(codes.FailedPrecondition, "cannot delete server %q because it still has organizations attached", reflect.Indirect(reflect.ValueOf(v)).FieldByName("Id"))
```
vsingh-msys, line 52: I am not 100% sure about the error code here but I thought it would be fitted for our need.

vsingh-msys: Receive the request object reference in the interface and extract the value base on the need.

components/infra-proxy-service/storage/postgres/postgres.go
```diff
@@ -59,10 +60,16 @@ func parsePQError(e *pq.Error) error {
...
 	case "foreign_key_violation":
-		// in this piece of code, a foreign key violation means the server the org
-		// should be added to doesn't exist; a "not found" seems like an OK
-		// approximation for now
-		return storage.ErrNotFound
+		// In this piece of code, a foreign key violation means the server the org
+		// should be added to doesn't exist.
+		// The only way to distinguish this violation by comparing the server_id key in Error detail.
+		// DETAIL:  Key (server_id)=(2e831dec-f25c-4fde-aef6-8f68ea366251) is not present in table "servers".
+		if strings.Contains(e.Detail, "(server_id)") {
```
vsingh-msys, line 67: I am fully sold on this comparison but I don't have a better solution as of now.

components/infra-proxy-service/server/servers_test.go
```diff
@@ -335,6 +335,47 @@ func TestServers(t *testing.T) {
...
+			require.NotNil(t, respOrg)
+
+			serverListBefore, err := cl.GetServers(ctx, &request.GetServers{})
+			require.NoError(t, err)
+			assert.Equal(t, 1, len(serverListBefore.Servers))
+
+			resp, err2 := cl.DeleteServer(ctx, &request.DeleteServer{Id: resp1.Server.Id})
+			require.Error(t, err2)
+			require.Nil(t, resp)
+			grpctest.AssertCode(t, codes.FailedPrecondition, err2)
```
msorens, line 369: Good test, but let us also check the text of the message that, after all, was the bug we are fixing. Typically I advise using a regular expression check rather than an equality check, which is too fragile for this case. Something like this both (a) checks the key pieces of data and (b) informs the reader of the test what is going on.
```
assert.Regexp(t, "cannot delete server.*still has organizations attached", err.Error())
```

components/infra-proxy-service/storage/postgres/postgres.go
msorens: That sounds reasonable, but let us tighten it up a bit, also confirming the message ("not present") and the table ("servers"), so I think a regex check like this would be preferable. Note that we want to match word boundaries (`\<` and `\>`) to make sure those are not just substrings.
```
regexp.MustCompile('\<server_id\>.*is not present.*table.*\<servers\>').MatchString(e.Detail)
```

vsingh-msys:
Thank you so much @msorens for detailed reviews.


> (1) I see you added a test for one fix but not the other, then I found the test "when the server does not exist, raise server not found error" in orgs_test.go. Since you have not changed it, that test must have been reporting a false positive; how could that test have passed previously?

- Actually, the other fix is about changing the error message, not the error code.
earlier miserably I have validated only the error code. Added assert for error message as well. https://github.com/chef/automate/pull/3298/files#diff-26345778d259836272877d139e0983a4R245


> Problem 1: Using the same non-existent server ID I attempted to delete the org I created above (attached to a different server!). This appears to have just done a GetOrg, surprisingly! (By the way, please confirm that my command is valid for deleting an org.)
{
  "org": {
    "id": "3509f164-b15e-4e3c-87dd-f31cf7d18771",
    "name": "ORG_NAME",
    "admin_user": "ADMIN_USER",
    "admin_key": "6ecf6c81-4c7b-4d80-a3eb-5b2277326d2a",
    "server_id": "041ee60f-3bb6-497f-bb94-6aba6fcdcc0b",
    "projects": [
    ]
  }
}

> Problem 2: the response was different!

> Problem 3: the error should have reported "cannot find the server".


- Sorry, I would rather call it accidentally hitting the arrow...due to UUID primary key, somehow I have neglected for few operations of orgs to validate server-id. It should have put in TODO to avoid confusion. 

however, I am working on a human-readable ID, considered these cases over there. pending to create PR.

components/infra-proxy-service/storage/postgres/postgres.go
```diff
@@ -64,7 +64,7 @@ func parsePQError(e *pq.Error) error {
 		// should be added to doesn't exist.
 		// The only way to distinguish this violation by comparing the server_id key in Error detail.
 		// DETAIL:  Key (server_id)=(2e831dec-f25c-4fde-aef6-8f68ea366251) is not present in table "servers".
-		if strings.Contains(e.Detail, "(server_id)") {
+		if regexp.MustCompile(`\(server_id\).*is not present.*table.*\"servers\"`).MatchString(e.Detail) {
```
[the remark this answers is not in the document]
vsingh-msys: Yeah, I do agree with your points, As I was verified these regexes but it does not fit our need.
1. `\<` would match the `<` in string 
2. `\b` uses to check the boundary of the words.

In order to make it more flexible might loosen the regex with escaping any char with `.`?

```
if regexp.MustCompile(`.server_id.*is not present.*table.*servers.`).MatchString(e.Detail) {

```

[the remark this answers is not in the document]
msorens: So you are confirming that `\b` matches word boundaries, right? But that is exactly what we want; I am puzzled why you are saying it would not work.

PR #3303, opened 2020-04-07
[Automate-3249] adjust clear selection in project filter

msorens:

1. Previously "Clear Selection" would clear the list in the filter, close the filter, and apply those changes, perhaps inadvertantly wiping out a lot of work for a user. Now, it just clears the filter but leaves it open and does not apply the changes. Thus, if it was done accidentally, one need only close the filter then re-open it to get back the current list of checked projects. Bonus: clearing means clearing only from those visible, i.e. those rendered by the filter!

2. Add a count to the button, maxing out at "99+".

3. Hide the button if the count is zero.

Note: Browse by commit for easy digestion.

image

NA

### :+1: Definition of Done
1. Selecting <kbd>Clear Changes</kbd> no longer auto-applies changes.
2. Button shows count of filtered and selected items.
3. Button auto-hides if no items in current filtered view or selected.

New tests that define what it does are highlighted: yellow shows testing new functionality; green shows new tests for previously existing functionality.

image

1. Rebuild automate-ui.
2. Generate a bunch of projects. From the command line `source ./scripts/auth_collections.sh`, then `authGen projects create 50 101` to create 50 projects.
3. Login in the browser.
4. Open the global projects filter and select some projects.
5. Select <kbd>Apply Changes</kbd>.
6. Re-open the global projects filter.
7. Select <kbd>Clear Changes</kbd>. Observe all changes are gone, but the underlying page does not change (assuming you had stuff assigned to projects).
8. Close the global projects filter. Observe the original projects are still indicated in the closed filter.
9. Re-open the global projects filter. Observe the original projects are still indicated in the open filter.

If you want to see the count go 97, 98, 99, 99+, you need a lot of projects.
Load the utility to do that with `source scripts/auth_collections.sh` then run e.g., `authGen projects create 200 101` to generate 200 projects.

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/modules/infra-proxy/chef-server-details/chef-server-details.component.spec.ts
+++ b/components/automate-ui/src/app/modules/infra-proxy/chef-server-details/chef-server-details.component.spec.ts
@@ -45,6 +45,7 @@ describe('ChefServerDetailsComponent', () => {
         MockComponent({ selector: 'chef-table-header-cell' }),
         MockComponent({ selector: 'chef-table-cell' }),
         MockComponent({ selector: 'a', inputs: ['routerLink'] }),
+        MockComponent({ selector: 'input', inputs: ['resetOrigin'] }),
         ChefServerDetailsComponent
       ],
       providers: [
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
@@ -35,7 +35,7 @@
           <chef-button
           tertiary
           id="projects-filter-clear-selection"
-          (click)="handleClearSelection()">Clear Selection (99+)</chef-button>
+          (click)="handleClearSelection()">Clear Selection ({{filteredSelectedCount}})</chef-button>
         </div>
 
         <div id="drop-list" class="dropdown-content">
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
@@ -35,6 +35,8 @@
           <chef-button
           tertiary
           id="projects-filter-clear-selection"
+          class="pf-clear-selection"
+          [class.active]="filteredSelectedCount !== '0'" 
           (click)="handleClearSelection()">Clear Selection ({{filteredSelectedCount}})</chef-button>
         </div>
 
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
@@ -18,7 +18,7 @@
 
         <div class="dropdown-content">
           <!-- @TODO turn this into reusable Filter Input component -->
-          <input 
+          <input
             chefInput
             type="text"
             (keyup)="handleFilterKeyUp($event.target.value)"
@@ -35,7 +35,7 @@
           <chef-button
           tertiary
           id="projects-filter-clear-selection"
-          (click)="handleClearSelection()">Clear Selection</chef-button>
+          (click)="handleClearSelection()">Clear Selection (99+)</chef-button>
         </div>
 
         <div id="drop-list" class="dropdown-content">
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
@@ -99,7 +99,7 @@
 
     .button-container { // adjust for 8px global margin on buttons
       display: flex;
-      justify-content: space-between;
+      justify-content: flex-start;
       padding: calc(14px - 8px);
       padding-bottom: 0;
     }
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
@@ -161,10 +161,11 @@
 .pf-clear-selection {
   visibility: hidden;
   opacity: 0;
-  transition: .25s;
+  transition: .2s ease-in;
 
   &.active {
     visibility: visible;
-    opacity: 1
+    opacity: 1;
+    transition: .25s ease-out;
   }
 }
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
@@ -64,7 +64,7 @@
 }
 
 .dropdown {
-  width: 295px;
+  width: 326px;
   right: -10px;
   border: 1px solid $chef-grey;
   font-size: 14px;
@@ -99,6 +99,7 @@
 
     .button-container { // adjust for 8px global margin on buttons
       display: flex;
+      justify-content: space-between;
       padding: calc(14px - 8px);
       padding-bottom: 0;
     }
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
@@ -156,3 +156,15 @@
   max-height: 365px;
   overflow-x: scroll;
 }
+
+// animation
+.pf-clear-selection {
+  visibility: hidden;
+  opacity: 0;
+  transition: .25s;
+
+  &.active {
+    visibility: visible;
+    opacity: 1
+  }
+}
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
@@ -63,11 +63,9 @@ export class ProjectsFilterDropdownComponent {
   }
 
   handleOptionChange(event, label) {
-    this.editableOptions.find(option => {
-      if (option.label === label) {
-        option.checked = event.detail; // provides the new state of the checkbox
-      }
-    });
+    // sets the new state of the specific checkbox
+    this.editableOptions
+      .find(option => option.label === label).checked = event.detail;
     this.optionsEdited = true;
   }
 
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
@@ -43,6 +43,11 @@ export class ProjectsFilterDropdownComponent {
     }
   }
 
+  get filteredSelectedCount(): string {
+    const count = this.filteredOptions.filter(option => option.checked).length;
+    return count > 99 ? '99+' : count.toString();
+  }
+
   handleFilterKeyUp(filterValue: string): void {
     this.filteredOptions = this.editableOptions
       .filter(option => option.label.toLowerCase().indexOf(filterValue.toLowerCase()) > -1);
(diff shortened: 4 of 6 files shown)


components/automate-ui/src/app/modules/infra-proxy/chef-server-details/chef-server-details.component.spec.ts
```diff
@@ -45,6 +45,7 @@ describe('ChefServerDetailsComponent', () => {
         MockComponent({ selector: 'chef-table-header-cell' }),
         MockComponent({ selector: 'chef-table-cell' }),
         MockComponent({ selector: 'a', inputs: ['routerLink'] }),
+        MockComponent({ selector: 'input', inputs: ['resetOrigin'] }),
```
msorens, line 48: PR bonus: unit test error picked up by Wallaby
(same for next file)

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
```diff
@@ -83,11 +83,8 @@ export class ProjectsFilterDropdownComponent {
   }
 
   handleClearSelection() {
```
msorens, line 85: This is the fix!

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.spec.ts
```diff
@@ -330,6 +305,44 @@ describe('ProjectsFilterDropdownComponent', () => {
     it('emits "onSelection" event with list of updated options', () => {
       expect(component.onSelection.emit).toHaveBeenCalledWith(component.editableOptions);
     });
+
+    it('emits "onOptionChange" event with list of updated options', () => {
+      expect(component.onOptionChange.emit).toHaveBeenCalledWith(component.editableOptions);
+    });
+  });
+
+  describe('handleClearSelection()', () => {
```
msorens, line 314: New test section specifically for the fix.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.spec.ts
```diff
@@ -374,3 +387,17 @@ describe('ProjectsFilterDropdownComponent', () => {
     });
   });
 });
+
+function genOptions(checkedItems: boolean[]): ProjectsFilterOption[] {
```
msorens, line 391: New factory to streamline this file.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.spec.ts
```diff
@@ -311,9 +261,34 @@ describe('ProjectsFilterDropdownComponent', () => {
     });
   });
 
+  describe('handleOptionChange() for multiple options', () => {
```
msorens, line 264: Enrich the existing tests a bit.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
bcmdarroch: I noticed the Apply Changes button stays disabled after clicking Clear Selection. Shouldn't it be enabled, in case you want to turn off all filters?

msorens: Good call!

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
```diff
@@ -35,7 +35,9 @@
           <chef-button
           tertiary
           id="projects-filter-clear-selection"
-          (click)="handleClearSelection()">Clear Selection</chef-button>
+          class="pf-clear-selection"
+          [class.active]="filteredSelectedCount !== '0'" 
+          (click)="handleClearSelection()">Clear Selection ({{filteredSelectedCount}})</chef-button>
```
msorens, line 40: Add count to the button and instrument for hiding/showing the button.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
```diff
@@ -64,7 +64,7 @@
 }
 
 .dropdown {
-  width: 295px;
+  width: 326px;
```
msorens, line 67: Accommodate the added count on the button label--width of widest value ` (99+)`

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
```diff
@@ -99,6 +99,7 @@
 
     .button-container { // adjust for 8px global margin on buttons
       display: flex;
+      justify-content: space-between;
```
msorens, line 102: The two buttons are no longer static; the width of the "Clear Selection" button can vary. Nonetheless, this keeps them pushed to their respective left/right edges, to maintain necessary alignment with other components.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
```diff
@@ -155,3 +156,16 @@
   max-height: 365px;
   overflow-x: scroll;
 }
+
+// animation
+.pf-clear-selection {
```
msorens, line 161: A bit of pizzazz to hide/display the button--thanks @SEAjamieD !

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
```diff
@@ -67,11 +68,9 @@ export class ProjectsFilterDropdownComponent {
...
 
   handleOptionChange(event, label) {
-    this.editableOptions.find(option => {
-      if (option.label === label) {
-        option.checked = event.detail; // provides the new state of the checkbox
-      }
-    });
+    // sets the new state of the specific checkbox
+    this.editableOptions
+      .find(option => option.label === label).checked = event.detail;
```
msorens, line 73: Correcting use of `find`. Above worked, but only as an unintended side effect.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.spec.ts
```diff
@@ -98,21 +101,7 @@ describe('ProjectsFilterDropdownComponent', () => {
...
-        },
-        {
-          value: 'project-2',
-          label: 'Project 2',
-          type: 'CUSTOM',
-          checked: true
-        }
-      ];
-      // need filteredOptions in order for elements to be displayed in the UI
+      component.editableOptions = genOptions([false, true]);
```
msorens, line 104: Introduce a mini-factory for streamlining.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.spec.ts
```diff
@@ -373,4 +438,263 @@ describe('ProjectsFilterDropdownComponent', () => {
       expect(nextElementSibling.focus).toHaveBeenCalled();
     });
   });
+
+  describe('filteredOptions', () => {
```
msorens, line 442: New tests for previously existing functionality.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.spec.ts
```diff
@@ -373,4 +438,263 @@ describe('ProjectsFilterDropdownComponent', () => {
...
+      ['everything', '-', 7]
+    ], function (description: string, filter: string, count: number) {
+      it(`with ${description} filter displays ${count} matching options`, () => {
+        component.handleFilterKeyUp(filter);
+        expect(component.filteredOptions.length).toEqual(count);
+      });
+    });
+  });
+
+  describe('filteredSelectedCount', () => {
```
msorens, line 478: And from here down are TDD tests for new functionality.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
susanev: would it be easy to keep the buttons 16px apart instead? so the space to the right of clear button will vary instead of the space between the buttons?

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
bcmdarroch: I'm curious how this works?

SEAjamieD: @bcmdarroch 
It's actually a pretty simple little css animation.

1. I added a base class to the "clear selection" button for easier selection --> `.pf-clear-selection`
2. I added a second dynamic class to the html - `active` that only applies when the selection count is not `0`
3. We set the CSS of the base class so that the button is invisible/not clickable by the user, but still exists in the DOM, taking up space

visibility: hidden --> this hides the elements but keeps it in the DOM
opacity: 0 --> this makes the element see through, so that we can fade it in later
transition: .2s ease-in --> shorthand
         - `.2s` the first number is the time it will take for the element to go from state A to state B (in our case.. to applying the styles in `.active`
        - `ease-in` is a predefined cubic-bezier (a timing function) that determines what the transition speed curve 
         looks like.  We actually want to use `ease-in` (or define our own curve) when an element is LEAVING the page
         - we could actually make this more specific by adding `opacity` as a third property on transition, since thats 
          really the only think we need to animate here.... currently the transition is applying to everything... maybe i'll 
          update that.

I'd like to see us design our own cubic-bezier in the future (and i think we will), but these are a solid place to start.
`ease-out` is kind of like....get there super fast but slow down at the towards the end
`ease-in` is kind of like.... start moving slowly and then crank the speed up.

4.  We add the css for when the `active` class gets applied.  When active is added, the element will fade in.
visibility: visible --> makes the element exist to the user again (thats probably poor choice of words on my part...but I imaging you get the gist)
opacity: 1 --> makes the image full opacity...or not see though
transition: .25s ease-out --> similar to above

You might notice that we are using separate timings (.2s and .25s) - not necessarily super noticeable in this application, but Google Design suggests using 195ms when leaving the page, and 220ms when entering the frame.  In general it just kind of provides a more 'human' motion.

I'm realizing now that everything I typed out makes it look more complicated than it is, but definitely encourage you to play with transitions or `@keyframes` to really have some fun.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
msorens: That was my first attempt, but it looks askew when you approach the button (i.e. hover over it).
Here it aligns-right with the form field above:
image
vs. this:
image

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
msorens: Great details, @SEAjamieD , thanks! But I have to ask: if recommendations are .195 and .220, why .2 and .25 then? 😁

SEAjamieD: 😂 fair question.  The guideline is loose - the more important piece is that the outgoing animation should be slightly faster than the incoming animation.  I felt keeping it to 2 digits read more easily.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.scss
susanev: the second image is what i want.

susanev: its more important that the beginning of the button does not move, as that can often be a frustrating experience. im unconcerned about alignment with the form input.

msorens: Changed.

SEAjamieD:
@susanev - Weiiiird... I'll dig into this but don't let it hold up a merge

SEAjamieD:
>the only thing i noticed is that the clear selection button loads before the menu, including a gif. really not worried about this though if you want to merge.

@susanev - I think I know whats going on here.  I'll prioritize for Monday next week.  BTW - What are you using to slow down the vid/browser?

PR #3312, opened 2020-04-07
[automate-3198] reduce validation

blakestier:

When creating a project, we use the same validation as other objects at the server level: character count must be 64 or less.

However, we auto-generate several roles and policies _using_ that id and additional characters. The resulting policy id can then be over the expected limit, making it impossible to _Get_ the policy bc of our validation on id length in Get requests.

With this change, `project_ids` must be 48 chars or less (allowing the other id creation without trouble).

Validation on policy Get requests (and, while I was in there, all other resource requests that didn't _create_ the id) no longer validate on length. Instead, validation confirms the required field is present and has more than whitespace characters.

It was a toss up whether to put the validation for this in the server (where we do our other proto validations, and where the problem would be caught early) or in the db (where we probably should have our more fine-grained validation code). The answer is probably both, but this change keeps it in the server, with a few moves to storage-level validation that had parity with this kind of validation.

start on a studio without this version of authz

make a project with an id longer than 50 chars
`curl -kH "api-token: $TOK" -d '{"name":"p123456789p123456789p123456789p123456789p123456789p123456789", "id":"p123456789p123456789p123456789p123456789p12342222"}' https://a2-dev.test/apis/iam/v2/projects`

attempt to fetch the associated policy (this should fail)
`curl -kH "api-token: $TOK" https://a2-dev.test/apis/iam/v2/policies/p123456789p123456789p123456789p123456789p123456789p123456789-project-editors`

`rebuild components/authz-service`

attempt to fetch the same policy (should now succeed)
`curl -kH "api-token: $TOK" https://a2-dev.test/apis/iam/v2/policies/p123456789p123456789p123456789p123456789p123456789p123456789-project-editors`

attempt to create a project with an id longer than 48 chars (should now fail)
` curl -kH "api-token: $TOK" -d '{"name":"p123456789p123456789p123456789p123456789p123456789p123456789", "id":"p123456789p123456789p123456789p123456789p12342222"}' https://a2-dev.test/apis/iam/v2/projects
{"error":"invalid CreateProjectReq.Id: value does not match regex pattern \"^[a-z0-9-_]{1,48}$\"","code":3,"message":"invalid CreateProjectReq.Id: value does not match regex pattern \"^[a-z0-9-_]{1,48}$\"","details":[]}`

Also, any other requests (GetRule, GetPolicy, etc) now fail with InvalidArg if the required ids are blank or absent.

- [ ] Tests added/updated?
- [ ] Docs added/updated?


api/interservice/authz/v2/project.proto
```diff
@@ -16,15 +16,15 @@ message Project {
 
 message CreateProjectReq {
     string name = 1;
-    string id = 2 [(validate.rules).string.pattern = "^[a-z0-9-_]{1,64}$"];
+    string id = 2 [(validate.rules).string.pattern = "^[a-z0-9-_]{1,48}$"];
```
blakestier, line 19: where it all began

components/automate-cli/pkg/diagnostics/integration/helpers.go
```diff
@@ -13,7 +13,7 @@ import (
 // TimestampName returns a name with a timestamp
 func TimestampName() string {
 	ts := time.Now()
-	return "integration-diagnostic-" + ts.Format("20060102150405")
+	return "diagnostic-" + ts.Format("20060102150405")
```
blakestier, line 16: didn't fit the new char limit

inspec/a2-api-integration/controls/iam/iam_apis.rb
```diff
@@ -948,7 +948,7 @@
       end
 
       after(:all) do
-        resp = automate_api_request("/apis/iam/v2/projects/#{CUSTOM_RULE[:project_id]}/rules/#{CUSTOM_RULE[:id]}", http_method: 'DELETE')
+        resp = automate_api_request("/apis/iam/v2/projects/#{CUSTOM_RULE_1[:project_id]}/rules/#{CUSTOM_RULE_1[:id]}", http_method: 'DELETE')
```
blakestier, line 951: driveby fix

api/interservice/authz/v2/policy_test.go
```diff
@@ -134,31 +134,11 @@ func TestValidationCreatePolicy(t *testing.T) {
 			Statements: []*v2.Statement{&validStatement},
 		},
 		// projects
-		"whitespace projects list": &v2.CreatePolicyReq{
```
bcmdarroch, line 137: these are all pushed into server?

components/authz-service/server/v2/project_test.go
```diff
@@ -189,8 +189,8 @@ func TestGetProject(t *testing.T) {
 			grpctest.AssertCode(t, codes.InvalidArgument, err)
 			assert.Nil(t, resp)
 		}},
-		{"if the project id is invalid, returns 'invalid argument'", func(t *testing.T) {
-			resp, err := cl.GetProject(ctx, &api.GetProjectReq{Id: "no spaces allowed"})
```
bcmdarroch, line 193: so these `no space allowed` would return not found right? Could we add tests that verify that behavior?

components/authz-service/server/v2/validation.go
```diff
@@ -0,0 +1,47 @@
...
+func confirmRequiredIDandName(id, name, resourceName string) error {
+	err := confirmRequiredField(id, "id", resourceName)
+	if err != nil {
+		return err
+	}
+	err = confirmRequiredField(name, "name", resourceName)
+	return err
+}
+
+func confirmRequiredField(field, fieldName, resourceName string) error {
```
bcmdarroch, line 22: nice DRY refactor 🎉

components/authz-service/storage/v2/rule.go
```diff
@@ -52,24 +52,10 @@ func NewRule(id string, projectID string, name string,
...
-		return errors.New(
-			"a rule name is required and must contain at least one non-whitespace character")
-	}
-	if emptyOrWhitespaceOnlyRE.MatchString(projectID) {
-		return errors.New("a rule id must have an associated project_id")
-	}
+func validateConditionsForType(ruleType RuleType, conditions []Condition) error {
 	if len(conditions) == 0 {
-		return errors.New("a rule must contain at least one condition")
+		return errors.New("at least one condition is required")
```
bcmdarroch, line 57: should this move up to the server code, since that where the other required fields are being checked?

components/authz-service/server/v2/policy_test.go
```diff
@@ -1174,7 +1174,7 @@ func TestReplacePolicyMembers(t *testing.T) {
 
 			require.Nil(t, pol)
 			grpctest.AssertCode(t, codes.InvalidArgument, err)
-			assert.Regexp(t, "invalid ReplacePolicyMembersReq.Id: value does not match regex pattern", err.Error())
+			assert.Regexp(t, "rpc error: code = InvalidArgument desc = a policy id is required and must contain at least one non-whitespace character", err.Error())
```
msorens, line 1177: No need for `assert.Regexp` if you are matching exactly... however, that is rather fragile. Would we, for example, want to fail, if the generator produced "code=InvalidArgument" instead of "code = InvalidArgument" ? Consider matching just the key content:
```suggestion
			assert.Regexp(t, "InvalidArgument.*policy id is required.*must contain at least one non-whitespace character", err.Error())
```

components/authz-service/server/v2/policy_test.go
```diff
@@ -1602,6 +1602,51 @@ func TestCreateRole(t *testing.T) {
...
+				Actions:  []string{"infra:some:action"},
+				Projects: []string{},
+			}
+
+			resp, err := cl.CreateRole(ctx, &req)
+			require.Nil(t, resp)
+
+			assert.Equal(t, 0, store.ItemCount())
+			grpctest.AssertCode(t, codes.InvalidArgument, err)
+			assert.Regexp(t, "a role name is required and must contain", err.Error())
```
msorens, line 1633: Hmm... in contrast to the above, this seems a rather unusual truncation. Would suggest be consistent with above.

components/authz-service/server/v2/project.go
```diff
@@ -118,9 +123,13 @@ func (s *ProjectState) GetProject(ctx context.Context,
...
 func (s *ProjectState) CreateProject(ctx context.Context,
 	req *api.CreateProjectReq) (*api.CreateProjectResp, error) {
+	err := confirmRequiredField(req.Name, "name", "project")
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, err.Error())
+	}
 	p, err := storage.NewProject(req.Id, req.Name, storage.Custom, storage.NoRules)
 	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument,
+		return nil, status.Errorf(codes.Internal,
```
msorens, line 132: Good eye!

components/authz-service/server/v2/rules_test.go
```diff
@@ -74,9 +74,9 @@ func TestCreateRule(t *testing.T) {
 			grpctest.AssertCode(t, codes.InvalidArgument, err)
 			assert.Nil(t, resp)
 		}},
-		{"if the passed rule project id is invalid, returns 'invalid argument'", func(t *testing.T) {
+		{"if the passed rule project id is empty spaces, returns 'invalid argument'", func(t *testing.T) {
```
msorens, line 77: Consistency with previous 2 files suggest "is whitespace":
```suggestion
		{"if the passed rule project id is whitespace, returns 'invalid argument'", func(t *testing.T) {
```

components/authz-service/server/v2/rules_test.go
```diff
@@ -227,9 +227,9 @@ func TestUpdateRule(t *testing.T) {
 			grpctest.AssertCode(t, codes.InvalidArgument, err)
 			assert.Nil(t, resp)
 		}},
-		{"if the rule id is invalid, returns 'invalid argument'", func(t *testing.T) {
+		{"if the rule id is all whitespace, returns 'invalid argument'", func(t *testing.T) {
```
msorens, line 230:
```suggestion
		{"if the rule id is whitespace, returns 'invalid argument'", func(t *testing.T) {
```

components/authz-service/storage/v2/role.go
```diff
@@ -67,22 +55,3 @@ func NewUpdateRole(id string, name string, actions []string, projects []string)
...
-	if emptyOrWhitespaceOnlyRE.MatchString(id) {
-		return errors.New(
-			"a role id is required and must contain at least one non-whitespace character")
-	}
-	if emptyOrWhitespaceOnlyRE.MatchString(name) {
-		return errors.New(
-			"a role name is required and must contain at least one non-whitespace character")
-	}
-	if len(actions) == 0 {
-		return errors.New("a role must contain at least one action")
```
msorens, line 81: I see this check is covered by the proto file so OK to remove it here. (Just wondering if we are happy with that error message vs. this one, thinking about our move away from proto-validation. 🤔 ).

components/authz-service/storage/v2/role.go
```diff
@@ -67,22 +55,3 @@ func NewUpdateRole(id string, name string, actions []string, projects []string)
...
-	}
-	err := ValidateProjects(projects)
```
msorens, line 83: Is this check covered elsewhere?

components/authz-service/server/v2/policy.go
```diff
@@ -362,11 +396,20 @@ func (s *policyServer) RemovePolicyMembers(ctx context.Context,
 func (s *policyServer) CreateRole(
 	ctx context.Context,
 	req *api.CreateRoleReq) (*api.Role, error) {
-	storageRole, err := storage.NewRole(req.Id, req.Name, storage.Custom, req.Actions, req.Projects)
+	err := confirmRequiredField(req.Name, "name", "role")
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, err.Error())
+	}
+	err = validateProjects(req.Projects)
```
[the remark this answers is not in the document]
blakestier: Yes-- once it's called within the primary server call itself (where the code is processed/returned), the other time in a helper, which just bubbles the error. That said, I can try to make that less confusing!

components/authz-service/storage/v2/role.go
blakestier: 🤔

components/authz-service/storage/v2/role.go
blakestier: Yep-- back in the server

api/interservice/authz/v2/policy_test.go
blakestier: Several that were about validation are removed as they are now the equivalent of a "not found"-- instead of duplicating more invalid things that would result in "not found", I removed them.

components/authz-service/storage/v2/role.go
blakestier: Oh the unassigned! Now it is :D

components/authz-service/server/v2/project_test.go
blakestier: since we're not validating at this level, an invalid name is the same as a not found. there are other tests for not found names, but they seemed functionally equivalent. What do think?

components/authz-service/server/v2/policy.go
```diff
@@ -94,6 +94,10 @@ func (s *policyServer) CreatePolicy(
 	req *api.CreatePolicyReq) (*api.Policy, error) {
 
 	// API requests always create custom policies.
+	err := validateRequiredFieldsAndProjectsForPolicy(req.Id, req.Name, req.Projects)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, err.Error())
```
srenatus, line 99: [nit] There's nothing to format here, so we could use [`status.Error`](https://pkg.go.dev/google.golang.org/grpc/status?tab=doc#Error):
```suggestion
		return nil, status.Error(codes.InvalidArgument, err.Error())
```

(The same applies below.)

components/authz-service/server/v2/project.go
```diff
@@ -151,10 +160,15 @@ func (s *ProjectState) CreateProject(ctx context.Context,
 
 func (s *ProjectState) UpdateProject(ctx context.Context,
 	req *api.UpdateProjectReq) (*api.UpdateProjectResp, error) {
+
+	err := confirmRequiredIDandName(req.Id, req.Name, "project")
```
srenatus, line 164: [nit] It looks weird, but the canonical casing here would be this:
```suggestion
	err := confirmRequiredIDAndName(req.Id, req.Name, "project")
```

...but I'm down to turning a blind eye 🙈

components/authz-service/server/v2/validation.go
```diff
@@ -0,0 +1,75 @@
...
+
+var emptyOrWhitespaceOnlyRE = regexp.MustCompile(`^\s*$`)
+
+func confirmRequiredIDandName(id, name, resourceName string) error {
+	err := confirmRequiredID(id, resourceName)
+	if err != nil {
+		return err
+	}
+	err = confirmRequiredName(name, resourceName)
+	return err
```
srenatus, line 19: [nit] 😉 
```suggestion
	return confirmRequiredName(name, resourceName)
```

components/authz-service/server/v2/validation.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func confirmRequiredField(field, fieldName, resourceName string) error {
+	if emptyOrWhitespaceOnlyRE.MatchString(field) {
+		e := fmt.Sprintf("a %s %s is required and must contain at least one non-whitespace character", resourceName, fieldName)
+		return errors.New(e)
```
srenatus, line 25: [nit]
```suggestion
		return fmt.Errorf("a %s %s is required and must contain at least one non-whitespace character", resourceName, fieldName)
```

Below, we're using [`errors.Errorf`](https://github.com/chef/automate/pull/3312/files#diff-80445eaad60cf019d0dd4c54dc9a488dR69), that would work, too.

components/authz-service/server/v2/validation.go
```diff
@@ -0,0 +1,75 @@
...
+func validateRequiredFieldsAndProjects(id string, name string, projectIDs []string, resourceName string) error {
+	err := confirmRequiredIDandName(id, name, resourceName)
+	if err != nil {
+		return err
+	}
+	err = validateProjects(projectIDs)
+	if err != nil {
+		return err
+	}
+	return nil
```
srenatus, line 51: [nit] I believe this is equivalent to
```go
return validateProjects(projectIDs)
```

blakestier: lol

components/authz-service/server/v2/project.go
blakestier: this does seem like an opportunity conceding that the "right" way is not always the most readable, huh

srenatus: Yeah probably. Also naming is hard 😉

msorens:
FYI: Created issue #3371 to mirror this strategy change to teams and authn services.

PR #3334, opened 2020-04-10
Replace stencil chef-table with angular chef-table

SEAjamieD:
This branch swaps out all of our stencil tables with pure angular components.  The look and feel of the tables are maintained.

Accomplishes: https://github.com/chef/automate/issues/2061

### :+1: Definition of Done
There are no more traces of the following stencils in the codebase 
`chef-table`, 
`chef-thead`,
`chef-tr`, 
`chef-th`, 
`chef-tbody`, 
`chef-td`, 
`chef-tfoot`

The application looks exactly the same.
note: `chef-table-new` has been renamed to `chef-table`

cd into `components/automate-ui`
make ui-update-lib
make serve

All tables should look the same as they did previously, but should use new angular only markup (no stencil components)

- [x] Tests added/updated?
- [ ] Docs added/updated?


susanev:
curious, do we need to continue to call it chef-table-new if we ware taking the old one away?

SEAjamieD:
> curious, do we need to continue to call it chef-table-new if we ware taking the old one away?

Yeah - as soon as I pull the stencil, I'm going to return all the `chef-table-new` back into `chef-table`.  I may do the same to the other related elements as well.

components/automate-ui/src/app/components/table/table-cell/table-cell.component.scss
```diff
@@ -9,11 +9,17 @@
   border-bottom: 1px solid $chef-grey;
   border-top: 1px solid $chef-grey;
   overflow: hidden;
-  text-overflow: ellipsis;
+  // text-overflow: ellipsis; // TODO: apply this locally
```
SEAjamieD, line 12: @susanev - Apoligies, realized I need to loop through this piece with you again

susanev: i think we should stick with wrapping for now

components/automate-ui/src/app/components/table/table-body/table-body.component.html
```diff
@@ -1 +1,5 @@
-<ng-content select="chef-table-row"></ng-content>
+<ng-content>
+  <!-- This should primarily house <chef-table-row> elements.
+  it is left open so that it may also hold other elements when necessary
+  such as a modal as seen in credentials-list.html -->
```
msorens, line 4: Great comment!

components/automate-ui/src/app/components/table/table-cell/table-cell.component.scss
msorens, line 12: Delete the comment...?

components/automate-ui/src/app/components/table/table.component.ts
```diff
@@ -1,7 +1,7 @@
 import { Component, HostBinding } from '@angular/core';
 
 @Component({
-  selector: 'chef-table-new',
+  selector: 'chef-table',
```
msorens, line 4: 🎉

components/automate-ui/src/app/pages/+compliance/+scanner/containers/job-scans-list/job-scans-list.component.scss
```diff
@@ -5,12 +5,19 @@
...
+  chef-table-cell {
+    display: flex;
+    justify-content: center;
+
+    &:last-child {
+      flex-flow: row nowrap;
+      justify-content: flex-end;
+      align-items: center;
 
-  chef-td:last-child {
```
msorens, line 12: As close as I could mark it: Need to delete the similar line from src/app/pages/+compliance/+scanner/containers/scanner/scanner.component.scss::L45
```
::ng-deep chef-td:last-child {
```

components/automate-ui/src/app/modules/infra-proxy/chef-servers-list/chef-servers-list.component.html
```diff
@@ -38,21 +38,21 @@
           </chef-table-header>
           <chef-table-body>
             <chef-table-row *ngFor= "let server of sortedChefServers$ | async">
-              <chef-table-cell>
+              <chef-table-cell class="align-flex-center">
```
msorens, line 41: Of the 160 instances of `<chef-table-cell>` only 24 of them use `class="align-flex-center"`. Can it really be true that only 15% of these columns need to be centered? That does not feel right. What is special about these 24?

components/automate-ui/src/app/modules/policy/details/policy-details.component.html
```diff
@@ -43,12 +43,12 @@
...
           </div>
         </chef-toolbar>
-        <chef-table-new *ngIf="members.length > 0">
+        <chef-table *ngIf="members.length > 0">
           <chef-table-header>
             <chef-table-row>
               <chef-table-header-cell>ID</chef-table-header-cell>
               <chef-table-header-cell>Type</chef-table-header-cell>
-              <chef-table-header-cell></chef-table-header-cell>
+              <chef-table-header-cell class="three-dot-column"></chef-table-header-cell>
```
msorens, line 51: 👍

SEAjamieD:
Thanks @susanev!

> * table headers have def grown in size, but i think you already knew this. were you gonna come back later to fix that?
> * this one tables header didn't grow, not sure if that means anything: compliance/reports/controls
Thanks for catching this.  Part of it was padding.  That's been fixed now.

I'm working on the others now.  This is happening because of the toggle buttons inside the header.  What I noticed though (and this is more for dev than UX) is that in half the app, we're using hand-built toggles, and in the other half, we're using a component we built named `chef-toggle-button`


> * the column widths seems strange here, compared to the rest: settings/roles
I just double checked and the widths are exactly the same as master.  Happy to move them around if you like

components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-node/reporting-node.component.scss
```diff
@@ -227,12 +227,12 @@ chef-tooltip[for='skipped-count'] {
...
-  margin: 2em 35px;
+  padding: 2em 35px;
 
-  chef-th,
-  chef-td {
+  chef-table-header-cell,
+  chef-table-cell {
     &:first-child {
-      flex-grow: 5;
+      flex-basis: 500%;
```
msorens, line 235: Perhaps a comment for this extraordinary value

components/automate-ui/src/app/pages/data-feed/data-feed.component.spec.ts
```diff
@@ -99,8 +99,8 @@ describe('DatafeedComponent', () => {
     });
 
     it('shows expected item count', () => {
-      const tableBody = getElementByCss(`${listId}`);
-      expect(tableBody.children.length).toBe(destinations.length + 1);
+      const tableBody = getElementByCss(`${listId} chef-table-body`);
+      expect(tableBody.children.length).toBe(destinations.length);
```
msorens, line 103: Can you explain this discrepancy? (Why would the test result be different?)

susanev:
>I just double checked and the widths are exactly the same as master. Happy to move them around if you like

oh yah, just seemed different than the rest, i think the other tables split space evenly, auth team can do that in a different pr though if its out of scope for this work

components/chef-ui-library/package-lock.json
```diff
@@ -5290,7 +5290,7 @@
         },
         "readable-stream": {
           "version": "1.0.34",
```
msorens, line 5292: These downgrades from https to http are unexpected. Suggest revert these changes here and start a fresh issue. Some relevant links (the second one has a suggested way to clean it up):
https://github.com/npm/npm/issues/20719
https://npm.community/t/npm-install-changes-package-lock-json-resolve-from-http-to-https/6899

SEAjamieD:
@susanev - All the header and row sizes should be uniform height now, as long as they don't have wrapping content.  Thanks again!

components/automate-ui/src/app/modules/infra-proxy/chef-servers-list/chef-servers-list.component.html
SEAjamieD: Good catch - initially I didn't use when the table doesn't appear to ever be in a scenario where they would have text wrapping applied.  However, probably better to add this to all chef-table cells at the top level and then apply locally when the change is needed.

components/automate-ui/src/app/pages/data-feed/data-feed.component.spec.ts
SEAjamieD: In the previous HTML markup, the table markup was semantically off - it was missing a `table-head` wrapping, so the list items inside the table included the header row - which required the test to need one extra list item, even though that item was actually a header row.

With the new HTML markup, the table now has a `chef-table-head` wrapping the header row.  A `chef-table-body` is holding the _actual_ table body contents.  So the number of list items should match only the number of items in the list.

components/chef-ui-library/package-lock.json
SEAjamieD: thanks for those fixes - appears to have worked.  Are you suggesting I open an issue with npm?

msorens: No, I was suggesting to fix it on our side separately from this PR... but if the linked fix already cured it for you, nothing more to do. 😁

SEAjamieD:
Thanks for reviewing!  I'm throwing a DO-NOT-MERGE on it for a few hours til I can poll on semantics, then i'll merge accordingly.

PR #3351, opened 2020-04-14
fix IAM v2 migration of policies with single-term resource

bcmdarroch:

A customer ran into an issue that broke authz-service when they tried to migrate a policy with the resource `cfgmgmt` to v2. This PR fixes that bug and ensures we skip migrating any v1 policies with single term resources that don't directly correspond to an existing API's policy annotations. Those policies were not enforcing any permissions in v1 anyway.

The exceptions to this skip-single-term-resource rule are:
- [nodes](https://github.com/chef/automate/blob/master/api/external/nodes/nodes.proto#L53)
- [events](https://github.com/chef/automate/blob/master/components/automate-gateway/api/event_feed/event_feed.proto#L21)
- [license](https://github.com/chef/automate/blob/master/components/automate-gateway/api/license/license.proto#L18)
- [nodemanagers](https://github.com/chef/automate/blob/master/api/external/nodes/manager/manager.proto#L38)
- [service_groups](https://github.com/chef/automate/blob/master/api/external/applications/applications.proto#L39)

The resources above match an existing v1 policy annotation API, so they do enforce permissions on v1 and should continue to do so on v2.

### :+1: Definition of Done

- [x] IAM v1->v2 migration does not fail on migrating v1 policies with single-term resources

TODO

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/authz-service/storage/postgres/migration/legacy/iam.go
+++ b/components/authz-service/storage/postgres/migration/legacy/iam.go
@@ -378,16 +378,22 @@ func legacyPolicyFromV1(pol *v1Policy) (*v2Policy, error) {
 func customPolicyFromV1(pol *v1Policy) (*v2Policy, error) {
 	name := fmt.Sprintf("%s (custom)", pol.ID.String())
 
+	resource, err := convertV1Resource(pol.Resource)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not derive v2 resource")
+	}
+	// in the case of most policies with single term resources (i.e. a container policy
+	// like 'cfgmgmt'), there is no v2 equivalent resource
+	// so the policy should be skipped
+	if resource == "" {
+		return nil, nil
+	}
+
 	// TODO: If we encounter an unknown action can we just be less permissive with a warning?
 	// AKA just use []string{"*"} instead of failing the migration?
 	action, err := convertV1Action(pol.Action, pol.Resource)
 	if err != nil {
 		return nil, errors.Wrap(err, "could not derive v2 action")
-	}
-
-	resource, err := convertV1Resource(pol.Resource)
-	if err != nil {
-		return nil, errors.Wrap(err, "could not derive v2 resource")
 	}
 
 	// Note: v1 only had (custom) allow policies
@@ -426,6 +432,11 @@ func convertV1Resource(resource string) (string, error) {
 
 	if len(terms) == 1 && terms[0] == "*" {
 		return "*", nil
+	}
+
+	singleTermsToMigrate := []string{"nodes", "events", "license", "nodemanagers", "service_groups"}
+	if len(terms) == 1 && !includes(singleTermsToMigrate, terms[0]) {
+		return "", nil
 	}
 
 	switch terms[0] {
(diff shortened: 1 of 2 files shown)


components/authz-service/storage/postgres/migration/legacy/iam.go
```diff
@@ -501,6 +512,11 @@ func convertV1Resource(resource string) (string, error) {
 }
 
 func convertV1Cfgmgmt(terms []string) (string, error) {
+	if len(terms) == 1 {
+		// skip?
```
tylercloke, line 516: Can we remove the question mark?

components/authz-service/storage/postgres/migration/legacy/iam.go
```diff
@@ -428,6 +434,11 @@ func convertV1Resource(resource string) (string, error) {
...
 	}
 
+	singleTermsToMigrate := []string{"nodes", "events", "license", "nodemanagers", "service_groups"}
+	if len(terms) == 1 && !includes(singleTermsToMigrate, terms[0]) {
+		return "", nil
+	}
+
 	switch terms[0] {
 	case "service_info":
 		switch terms[1] {
```
tylercloke, line 444: Did we test all these switch cases that use `terms[1]` with no guard as well?

components/authz-service/storage/postgres/migration/legacy/iam.go
```diff
@@ -501,6 +512,11 @@ func convertV1Resource(resource string) (string, error) {
 }
 
 func convertV1Cfgmgmt(terms []string) (string, error) {
+	if len(terms) == 1 {
+		// skip?
+		return "", nil
+	}
+
```
tylercloke, line 519: This line in convertV1Actions looks like it might need a length guard as well.

https://github.com/chef/automate/pull/3351/files#diff-586ab1c0cc234c002424a036a86fbf85R593

components/authz-service/storage/postgres/migration/legacy/iam.go
bcmdarroch: Yes, all single term resource policies are now tested [in this test](https://github.com/chef/automate/pull/3351/files/833641eded1ff2bd358ad97fcf3577325128ae85#diff-b05308e7d958693621e9a536ed26554dR234).

Lines 438-441 provide the guard for any single-term resources. We only let through resources like `nodes` and `license`, which are used in v1 policy annotations.

components/authz-service/storage/postgres/migration/legacy/iam.go
bcmdarroch: We won't hit this case any longer because I moved [the `convertResource` function ahead of the `convertAction`](https://github.com/chef/automate/pull/3351/files#diff-586ab1c0cc234c002424a036a86fbf85L388). So if there's a single resource, like `ingest`, the `convertResource `function returns "" and then the function exits early, skipping the policy.

components/authz-service/storage/postgres/migration/legacy/iam.go
```diff
@@ -638,3 +654,12 @@ func convertV1Action(action string, resource string) ([]string, error) {
 		return nil, fmt.Errorf("could not parse V1 action: %s", action)
 	}
 }
+
+func includes(terms []string, termToCheck string) bool {
```
blakestier, line 657: Minor potential refactor: I don't see you using this anywhere else, which makes me wonder if we could rename it and push the `singleTermsToMigrate` into the function. Then we could call something like `shouldSkipTermMigration(term)`.

NBD tho since we wanna 🚢

PR #3359, opened 2020-04-14
Rename s/description/name in token db and server code

tylercloke:

Also make id a mandatory field on creation

```
# rebuild components/authn-service/ && rebuild components/automate-gateway/ && rebuild components/automate-deployment/
```

Make some tokens via the UI, CLI, and API. If you leave out ID, you'll get an error:

```
curl -XPOST -d '{"name":"asdf","projects":null,"active":true}' https://a2-dev.test/apis/iam/v2/tokens
{"error":"invalid CreateTokenReq.Id: value does not match regex pattern \"^[a-z0-9-_]{1,64}$\"","code":3,"message":"invalid CreateTokenReq.Id: value does not match regex pattern \"^[a-z0-9-_]{1,64}$\"","details":[]}
```

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/automate-deployment/pkg/server/api_token.go
```diff
@@ -63,8 +64,9 @@ func generateAdminToken(ctx context.Context,
 	authzV2Client := authz_v2.NewPoliciesClient(authzConnection)
 
 	response, err := authnClient.CreateToken(ctx, &authn.CreateTokenReq{
```
srenatus, line 66: 💭 Was this call idempotent before? If so, is it still? I don't know from the top of my head where this method is called, but it might be good to double-check.

components/automate-gateway/handler/iam/v2/tokens/tokens.go
```diff
@@ -126,12 +126,11 @@ func (s *Server) ResetAllTokenProjects(
...
 }
 
-// Maps the IAMV2 token (name, id) to original token model (id, description).
-// Value returned from internal auth service
+// Maps internal type to gateway type
 func convert(token *authn.Token) *pb_common.Token {
 	return &pb_common.Token{
 		Id:        token.Id,
-		Name:      token.Description,
+		Name:      token.Name,
```
srenatus, line 133: Yay

srenatus:
Nice commit history. Makes me think you have internalized that it's _squashed-and-merged_ anyways in the end... 😉

components/authn-service/tokens/pg/sql/README.md
```diff
@@ -14,3 +14,4 @@
 - [`12_add_filter_function_for_token_projects.up.sql`](12_add_filter_function_for_token_projects.up.sql)
 - [`13_add_db_id_column_to_tokens.up.sql`](13_add_db_id_column_to_tokens.up.sql)
 - [`14_drop_all_projects.up.sql`](14_drop_all_projects.up.sql)
+- [`15_v2_update.up.sql`](15_v2_update.up.sql)
```
msorens, line 17: Just wondering if this should be more specific... There are other files in the list here that specify a column being renamed. 🤷

components/automate-deployment/pkg/server/api_token_test.go
```diff
@@ -134,7 +134,7 @@ func TestGenerateAdminToken(t *testing.T) {
 		}
 
 		req := &api.GenerateAdminTokenRequest{
-			Description: testDescription,
+			Description: testName,
```
bcmdarroch, line 137: Let's rename this Description field too

components/automate-deployment/pkg/server/api_token.go
tylercloke: Naw, before we were just using authz-service to generate the UUID for us.

components/automate-deployment/pkg/server/api_token_test.go
tylercloke: Lemme see how big of a change that'll end up being. If big I'll do it in a follow up.

tylercloke: Looks like it's only used in one place, I'll fix it up here.

Issue #3060, opened 2020-03-09
remove all iam v1 and iam v2 references


susanev:
When we had to distinguish between v1 and v2 implementations, we labeled them with either v1 or v2. Now that only v2 exists, we can remove those distinctions.

This looks like finding places where we have named an import `v2_authz` or something, and changing it to just `authz`, for example.

We should also remove the `v2` directories authz-service and teams-service, and rename any `v2` package

PR #3361, opened 2020-04-15
[automate-3112] update teams table to v2 format

bcmdarroch:

We've hit the point where everything in teams-service but the storage layer respected the v2 format of teams. Storage teams still followed v1 format:
```
id: UUI                # autogen UUID
name: string           # user-created, unique
description: string    # user-created, anything goes
projects: []string     # list of project IDs
user_ids: []string     # list of user GUIDs
```

Now it follows the v2 format:
```
id: string             # user-created, unique alphanumeric ID
name: string           # user-created, anything goes
created_at: time.Time
updated_at: time.Time
projects: []string
```

follow-up: https://github.com/chef/automate/pull/3414

### :+1: Definition of Done
- [x] v1 team Name is now team ID in v2 world
- [x] v1 team Description is now team Name in v2 world
- [x] v1 team UUID is gone

- no functional changes, just need tests to pass

- [x] Tests added/updated?
- [ ] Docs added/updated?


components/teams-service/server/v2/server.go
```diff
@@ -22,12 +24,24 @@ func NewServer(service *service.Service) *Server {
 	return &Server{service: service}
 }
 
+// GetVersion returns the version of Teams GRPC API
```
srenatus, line 27: I wonder if we make use of this outside of the smoke tests... 🤔

components/teams-service/server/v2/server_test.go
```diff
@@ -63,6 +64,13 @@ func runAllServerTests(
 	t.Helper()
 	defer close()
 
+	t.Run("GetVersion", func(t *testing.T) {
+		resp, err := cl.GetVersion(context.Background(), &version.VersionInfoRequest{})
+		require.NoError(t, err)
+		// Version is injected via linker flags that we don't set during tests
```
srenatus, line 70: [nit] we could set it though, couldn't we? It's a package-level string variable IIRC.

components/teams-service/storage/postgres/postgres.go
```diff
@@ -16,20 +16,10 @@ import (
...
 var emptyOrWhitespaceOnlyRE = regexp.MustCompile(`^\s*$`)
 
-// WARNING
-// TODO (tc): The storage interface is still using V1 verbiage, so
-// name is really the ID in V2 terms. We'll refactor at GA when V1 is removed.
-// Also, project filtering has only been implemented for storage functions
-// that are used by the V2 server.
-
-// TODOs
-// - CREATE EXTENSION "uuid-ossp"; (currently done via Makefile for tests)
```
srenatus, line 31: Hah! There's a line in the `Makefile` to remove that'll bring this PR to `-2000` 🎉

components/teams-service/storage/postgres/migration/sql/11_convert_table_to_v2.up.sql
```diff
@@ -0,0 +1,29 @@
...
+ALTER TABLE teams DROP COLUMN id;
+
+-- under IAM v1, team names were unique identifiers
+-- under IAM v2, IDs are the unique human-readable identifiers, 
+-- though it is the db_id that is used as the identifier among db tables
+ALTER TABLE teams RENAME COLUMN name TO id;
+
+--- rename uniqueness constraint
+ALTER TABLE teams DROP CONSTRAINT teams_name_key;
+ALTER TABLE teams ADD CONSTRAINT teams_id_key UNIQUE (id);
```
srenatus, line 12: 💭 there's absolutely no possibility that a customer would have non-unique IDs, right? (Since this would kill upgrading if there was..)

components/teams-service/server/v2/server.go
bcmdarroch: we don't atm 🤷‍♀️

components/teams-service/server/v2/server_test.go
bcmdarroch: I'm not sure how do set it. Maybe we could pair on it real quick?

components/teams-service/storage/postgres/migration/sql/11_convert_table_to_v2.up.sql
bcmdarroch: we've been storing all IDs under `name` which had a [unique constraint](https://github.com/chef/automate/blob/master/components/teams-service/storage/postgres/migration/sql/01_create-pgstore-table.up.sql#L3) since its creation so I feel pretty confident. 

The main issue customers may run into is that we allowed spaces and non-alphanumeric customers in v1 names. We decided to handle those on a case by case basis rather than try to write a crazy migration

components/teams-service/server/v2/server.go
```diff
@@ -131,28 +141,24 @@ func (s *Server) AddTeamMembers(ctx context.Context,
...
 	if err != nil && err != storage.ErrConflict {
-		return nil, service.ParseStorageError(err, teamUUID, "user")
+		return nil, service.ParseStorageError(err, req.Id, "user")
 	}
 
 	// TODO (tc): Get the updated set of user membership for the team since
 	// that is not what is returned from AddUsers for some reason
 	// (can refactor on V1 deprecation).
-	userIDs, err := s.service.Storage.GetUserIDsForTeam(ctx, teamUUID)
+	userIDs, err := s.service.Storage.GetUserIDsForTeam(ctx, req.Id)
```
bcmdarroch, line 159: going to address the comment here in a follow-up PR

components/teams-service/server/v2/server.go
```diff
@@ -164,28 +170,24 @@ func (s *Server) AddTeamMembers(ctx context.Context,
...
 	req *teams.RemoveTeamMembersReq) (*teams.RemoveTeamMembersResp, error) {
 
-	// TODO (tc): The storage interface is still using V1 verbiage, so
-	// name is really the ID in V2 terms. We'll refactor at GA when V1 is removed.
-	//
-	// Fetch the UUID since that's what the membership table uses.
-	// Will fail the request here if the user in question is filtered by projects.
-	teamForUUID, err := s.service.Storage.GetTeamByName(ctx, req.Id)
+	// verify team exists and isn't filtered out by the project filter
+	_, err := s.service.Storage.GetTeam(ctx, req.Id)
```
bcmdarroch, line 174: again, hoping to minimize the storage calls from server level in a follow-up PR

components/teams-service/server/v2/server_test.go
```diff
@@ -1763,22 +1776,214 @@ func runAllServerTests(
...
 		})
 	})
-}
 
-func cleanupTeam(ctx context.Context, t *testing.T, cl teams.TeamsClient, teamID string) {
-	t.Helper()
-	deleteReq := teams.DeleteTeamReq{Id: teamID}
-	_, err := cl.DeleteTeam(ctx, &deleteReq)
-	require.NoError(t, err)
+	t.Run("PurgeUserMembership", func(t *testing.T) {
```
bcmdarroch, line 1783: this was ported from the v1 server tests

components/teams-service/storage/conformance/conformance_test.go
```diff
@@ -243,117 +241,90 @@ func testPurgeUserMembership(ctx context.Context, t *testing.T, s storage.Storag
...
+	_, err = s.DeleteTeam(ctx, id)
 	require.NoError(t, err)
 
 	teams, err = s.GetTeams(ctx)
 	require.NoError(t, err, "failed to read back teams")
 	require.Equal(t, len(storage.NonDeletableTeams), len(teams))
 }
 
-func testEditTeamConflict(ctx context.Context, t *testing.T, s storage.Storage) {
-	name := "Test Team"
```
bcmdarroch, line 268: I decided this was a v1-specific case so I removed it

components/teams-service/storage/conformance/conformance_test.go
```diff
@@ -130,18 +128,18 @@ func TestStorage(t *testing.T) {
 }
 
 func testGetTeamByName(ctx context.Context, t *testing.T, s storage.Storage) {
-	teamName := "Test Team"
-
-	team, err := s.StoreTeam(ctx, teamName, "description")
+	id := "test-team"
```
bcmdarroch, line 131: mostly just changing some variables to be valid IDs and slight rearranging here

components/teams-service/storage/memstore/memstore.go
```diff
@@ -8,26 +8,19 @@ import (
...
 
-// WARNING
-// TODO (tc): The storage interface is still using V1 verbiage, so
-// name is really the ID in V2 terms. We'll refactor at GA when V1 is removed.
-
 type memstore struct {
-	teams       map[uuid.UUID]storage.Team
-	teamsV2     map[string]storage.Team
-	usersByTeam map[uuid.UUID][]string
+	teams       map[string]storage.Team
```
bcmdarroch, line 14: decided to save ripping out memstore for another time

components/teams-service/storage/storage.go
```diff
@@ -6,32 +6,21 @@ import (
...
-	EditTeam(context.Context, Team) (Team, error)
 	GetTeams(context.Context) ([]Team, error)
-	GetTeam(context.Context, uuid.UUID) (Team, error)
-	StoreTeam(ctx context.Context, name string, description string) (Team, error)
-	StoreTeamWithProjects(ctx context.Context, name string, description string, projects []string) (Team, error)
-	RemoveUsers(ctx context.Context, id uuid.UUID, users []string) (Team, error)
-	GetTeamByName(ctx context.Context, teamName string) (Team, error)
-	PurgeUserMembership(ctx context.Context, userID string) (teamsUpdated []uuid.UUID, err error)
-	AddUsers(ctx context.Context, teamID uuid.UUID, users []string) (Team, error)
+	StoreTeam(ctx context.Context, id string, name string, projects []string) (Team, error)
```
bcmdarroch, line 14: `StoreTeam` was the v1 function, but I decided to delete it and then rename `StoreTeamWithProjects` to `StoreTeam` to preserve shorter function names. (same with GetTeamByName, DeleteTeamByName, EditTeamByName)

components/teams-service/storage/conformance/conformance_test.go
```diff
@@ -152,88 +150,88 @@ func testGetTeamByName(ctx context.Context, t *testing.T, s storage.Storage) {
...
 
 	assert.ElementsMatch(t, ids, expectedIDs) // Comparing the objects wholly will fail due to updated_at :(
 }
 
 func testPurgeUserMembership(ctx context.Context, t *testing.T, s storage.Storage) {
-	name := "Team1 With User"
-	team, err := s.StoreTeam(ctx, name, name+" Description")
-	require.NoError(t, err, "setup: failed to create test team")
+	id := "team-1-with-user"
+	team, err := s.StoreTeam(ctx, id, id+" Name", []string{})
```
msorens, line 191: Not really significant, but several of the variables (team, team2, teamWithout) are assigned multiple times before being examined, so all the earlier (unused) assignments could be removed (i.e. just use `_`).

components/teams-service/storage/conformance/conformance_test.go
```diff
@@ -405,20 +378,21 @@ func testRemoveUsers(ctx context.Context, t *testing.T, s storage.Storage) {
...
 
 func testRemoveUsersNotFound(ctx context.Context, t *testing.T, s storage.Storage) {
-	_, err := s.RemoveUsers(ctx, uuid.Must(uuid.NewV4()), []string{"user-1"})
+	_, err := s.RemoveUsers(ctx, "not-found", []string{"user-1"})
 	assert.Equal(t, storage.ErrNotFound, err)
 }
 
 func testGetTeam(ctx context.Context, t *testing.T, s storage.Storage) {
-	team, err := s.StoreTeam(ctx, "name", "description")
+	team, err := s.StoreTeam(ctx, "id", "name", []string{})
```
msorens, line 395: Consider adding a named constant to make all these tests a wee bit clearer:
```suggestion
	team, err := s.StoreTeam(ctx, "id", "name", emptyProjectsList)
```

components/teams-service/storage/storage.go
```diff
@@ -42,12 +31,11 @@ type Resetter interface {
...
-	Description string
-	CreatedAt   time.Time
-	UpdatedAt   time.Time
-	Projects    []string
+	ID        string
+	Name      string
+	CreatedAt time.Time
+	UpdatedAt time.Time
+	Projects  []string
 }
```
msorens, line 39: `IsTeamDeletable` lower down in the file needs an edit, too.
```
// IsTeamDeletable returns whether or not an input team is deletable by id.
func IsTeamDeletable(teamID string) bool {
	return !stringutils.SliceContains(NonDeletableTeams, teamID)
}
```

components/teams-service/server/v2/server.go
srenatus: I think we should be using the health protocol for health checks 😅 

This endpoint is pretty useless, given that we never shipped updates to services in isolation (only complete A2 versions). **But** 🤷‍♂️ the endpoint doesn't hurt.

srenatus: https://github.com/chef/automate/blob/31442fe17fe52e537809ca6bf9ea21eb729f636a/components/teams-service/server/server.go#L34 this one

components/teams-service/server/v2/server_test.go
srenatus: We can, however, should you want to follow the breadcrumbs:

- We inject the version into the linker args here https://github.com/chef/automate/blob/31442fe17fe52e537809ca6bf9ea21eb729f636a/components/teams-service/Makefile#L9
- That includes the path of the string constant that's set, i.e. `-X ${LIBRARY_PATH}/version.Version=${BUILD_TIME}`, which, [using `LIBRARY_PATH` from `Makefile.common_go`](https://github.com/chef/automate/blob/31442fe17fe52e537809ca6bf9ea21eb729f636a/Makefile.common_go#L9) resolves to `github.com/chef/automate/lib//version.Version`.
- That defined here: https://github.com/chef/automate/blob/31442fe17fe52e537809ca6bf9ea21eb729f636a/lib/version/version.go#L10

With that, I'd try to set it to something by importing `github.com/chef/automate/lib/version`, and then setting `version.Version = "foobar"`.

components/teams-service/server/v2/server.go
```diff
@@ -3,9 +3,11 @@ package v2
 import (
```
srenatus, line 3: Do we want to keep the `v2` in this package's name? I suppose we could now move the code to `server/server.go`?

components/teams-service/storage/postgres/migration/sql/11_convert_table_to_v2.up.sql
```diff
@@ -0,0 +1,29 @@
...
+    RETURNS INTEGER
+    AS $$
+    SELECT
+        db_id
+    FROM
+        teams
+    WHERE
+        id = _id;
+$$
+LANGUAGE SQL;
```
srenatus, line 29: I was wondering for a moment if this table could have other tables reference its `id` col, but since teams-services doesn't have that many tables, I don't think that it's the case.

components/teams-service/server/v2/server.go
srenatus: Ah, looks like we're tracking this in #3060.

components/teams-service/storage/postgres/migration/sql/11_convert_table_to_v2.up.sql
bcmdarroch: yeah we've already refactored the user association table to use the db_id and that's unchanged (yay for that refactor!)

components/teams-service/storage/storage.go
bcmdarroch: eep good catch! led me to fix the `AdminTeamName` constant too

PR #3363, opened 2020-04-15
Jamie/2254 bug filter persistence

SEAjamieD:

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-node/reporting-node.component.html
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-node/reporting-node.component.html
@@ -2,7 +2,7 @@
   <div class="container">
     <main>
       <chef-breadcrumbs>
-        <chef-breadcrumb [link]="['/compliance/reports/nodes']">Reports</chef-breadcrumb>
+        <chef-breadcrumb [routerLink]="['/compliance/reports/nodes']" [queryParams]='returnParams'>Reports</chef-breadcrumb>
         {{ activeReport.node_name }}
       </chef-breadcrumbs>
 
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-node/reporting-node.component.ts
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-node/reporting-node.component.ts
@@ -43,7 +43,7 @@ export class ReportingNodeComponent implements OnInit, OnDestroy {
     this.layoutFacade.ShowPageLoading(true);
     const id: string = this.route.snapshot.params['id'];
     const reportQuery = this.reportQuery.getReportQuery();
-    this.returnParams = this.formatReturnUrl(reportQuery);
+    this.returnParams = this.formatReturnparams(reportQuery);
     reportQuery.filters = reportQuery.filters.concat([{type: {name: 'node_id'}, value: {id}}]);
 
     this.statsService.getReports(reportQuery, {sort: 'latest_report.end_time', order: 'DESC'})
@@ -183,7 +183,7 @@ export class ReportingNodeComponent implements OnInit, OnDestroy {
     return moment(timestamp).fromNow();
   }
 
-  private formatReturnUrl(reportQuery): any { // return type is a url
+  private formatReturnparams(reportQuery): any { // return type is a url
     const structuredFilters = {};
     reportQuery.filters.map(filter => {
       return structuredFilters[filter.type.name] = filter.value.text;
@@ -194,6 +194,7 @@ export class ReportingNodeComponent implements OnInit, OnDestroy {
 
     const today = new Date();
     const isToday = moment(today).diff(endDate, 'days') === 0;
+
     if (isEmpty(structuredFilters) && isToday ) {
       return {};
     } else {
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-node/reporting-node.component.ts
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-node/reporting-node.component.ts
@@ -2,12 +2,11 @@ import { Component, OnInit, OnDestroy } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
 import { StatsService } from '../../shared/reporting/stats.service';
 import { Subject } from 'rxjs';
-import { ReportQueryService, ReportQuery } from '../../shared/reporting/report-query.service';
+import { ReportQueryService } from '../../shared/reporting/report-query.service';
 import * as moment from 'moment';
 import { DateTime } from 'app/helpers/datetime/datetime';
 import { LayoutFacadeService, Sidebar } from 'app/entities/layout/layout.facade';
 import { takeUntil } from 'rxjs/operators';
-import { isEmpty } from 'lodash';
 
 @Component({
   selector: 'app-reporting-node',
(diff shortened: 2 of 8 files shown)


components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-node/reporting-node.component.ts
```diff
@@ -30,7 +32,7 @@ export class ReportingNodeComponent implements OnInit, OnDestroy {
   constructor(
     private route: ActivatedRoute,
     private statsService: StatsService,
-    private reportQuery: ReportQueryService,
+    private reportQueryService: ReportQueryService,
```
SEAjamieD, line 35: I renamed this service in a few files from reportQuery to reportQueryService to be more reflective of what it is.  There is also a ReportQuery object and type - having reportQuery sometimes be the service and sometimes be an object was getting confusing.

components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-node/reporting-node.component.ts
```diff
@@ -39,15 +41,16 @@ export class ReportingNodeComponent implements OnInit, OnDestroy {
     this.layoutFacade.showSidebar(Sidebar.Compliance);
     this.reportLoading = true;
     this.layoutFacade.ShowPageLoading(true);
+    this.returnParams = this.reportQueryService.formatReturnParams();
```
SEAjamieD, line 44: onInit - we call the service right away so that we can populate the breadcrumb with the return value immediately.

components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-node/reporting-node.component.html
```diff
@@ -2,7 +2,7 @@
   <div class="container">
     <main>
       <chef-breadcrumbs>
-        <chef-breadcrumb [link]="['/compliance/reports/nodes']">Reports</chef-breadcrumb>
+        <chef-breadcrumb [routerLink]="['/compliance/reports/nodes']" [queryParams]='returnParams'>Reports</chef-breadcrumb>
```
SEAjamieD, line 5: Using routerlink, I added the queryParams attribute so that if we have filters, we can apply them dynamically, and if we don't - no harm the link will just go back to the overview page without filters.

components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.ts
```diff
@@ -26,38 +26,41 @@ import {
...
 import { FilterC } from './types';
 
+
 @Component({
   templateUrl: './reporting.component.html',
   styleUrls: ['./reporting.component.scss'],
   providers: [SuggestionsService]
 })
+
 export class ReportingComponent implements OnInit, OnDestroy {
```
SEAjamieD, line 43: I wanted stronger typing in for my `ReturnParams` interface, which required me to reach out a little further and add some stronger types to the filters overall.

components/automate-ui/src/app/pages/+compliance/shared/reporting/report-query.service.spec.ts
```diff
@@ -48,4 +48,12 @@ describe('ReportQueryService', () => {
...
     });
   });
+
+  describe('formatReturnParams', () => {
+    it('returns empty object when there are no filters and date is today ', () => {
+      service.formatReturnParams();
+      // this is just testing an empty report Query
+      expect(service.formatReturnParams()).toEqual({});
+    });
+  });
```
SEAjamieD, line 58: I could use some help figuring out how to Mock a Report Query response so that I can make this test better.  Right now its just testing that an empty call comes back empty, which doesn't really do much.

components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.ts
```diff
@@ -390,7 +393,7 @@ export class ReportingComponent implements OnInit, OnDestroy {
   onFiltersClear(_event) {
     const queryParams = {...this.route.snapshot.queryParams};
 
-    const filteredParams = pickBy((_value, key) => {
+    const filteredParams = pickBy((_value, key: ReportingFilterTypes) => {
```
SEAjamieD, line 396: I think I did this piece right, but i'm not positive.  Linter didn't like me passing a "string" as they type, and needed it to be specifically ReportingFilterTypes which is the enum where the Keys actually originate.

components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.ts
```diff
@@ -460,7 +463,7 @@ export class ReportingComponent implements OnInit, OnDestroy {
 
     reportQuery.filters = urlFilters.filter(
       (urlParm: Chicklet) => this.isTypeControlTag(urlParm.type)
-        || this.allowedURLFilterTypes.indexOf(urlParm.type) >= 0)
+        || this.allowedURLFilterTypes.indexOf(urlParm.type as ReportingFilterTypes) >= 0)
```
SEAjamieD, line 466: Same here with urlParm.type actually being part of the Enum `ReportingFilterTypes`

components/automate-ui/src/app/pages/+compliance/shared/reporting/report-query.service.ts
```diff
@@ -89,4 +119,33 @@ export class ReportQueryService {
...
+
+    // With no filters when the date it today, we don't want
+    // to add any parameters to the URL.
+    if (isEmpty(structuredFilters) && isToday) {
+      return {};
+    } else {
+      return {
+        end_time: endDate,
+        ...structuredFilters
+      };
```
msorens, line 148: I cannot infer anything about possible values for `structuredFilters` so better to be safe by spreading first, to guarantee our `endDate` value is used.
```suggestion
      return { ...structuredFilters, end_time: endDate };
```
(If `structuredFilters` is really `ReportingFilterTypes`, then definitely need the change.)

SEAjamieD:
> im getting some weird behaviour trying this out locally, where when i try to back out with breadcrumbs link i end up with 0 nodes or a spinner

wow good catch!  fixed now :)

This was happening because when I was populating the query parameter variables, there is the option to use `id` or `text` from the queryReport.  most of the times the text and id are the same, but sometimes they are different and the url parameters _actually_ rely on the `id`.  Since I was using text..._most_ of the time it worked.  Glad it was caught now!

components/automate-ui/src/app/pages/+compliance/shared/reporting/report-query.service.ts
SEAjamieD: thanks!  I updated this now and added stronger typing to StructuredFilters

components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.ts
```diff
@@ -390,7 +393,7 @@ export class ReportingComponent implements OnInit, OnDestroy {
   onFiltersClear(_event) {
     const queryParams = {...this.route.snapshot.queryParams};
 
-    const filteredParams = pickBy((_value, key) => {
+    const filteredParams = pickBy((_value, key: ReportingFilterTypes) => {
 
       // Only handles control_tag filters
       if (key.includes('control_tag')) {
```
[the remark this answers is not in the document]
SEAjamieD: Solid suggestion, I've cleaned this function up - but it needed to be done a little bit differently.  I've made a mesh of your suggestion and the previous code, because there are actually two `control_tag` type values, and the clearAll function won't work properly if it is included because control tags rely on second level filtering.  I've updated the note a bit as well.

vjeffrey:
hmm. i still seem to be having trouble:
```
load_compliance_reports
filter by a control from controls list
go to profiles tab
see details of a profile
go back using breadcrumbs
0 everything (looks to be using name for request where it should use id)
```
same deal for a standalone profile filter and a node filter

i tried deleting and repulling your branch locally to see if maybe something went wrong; no luck :/

components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-node/reporting-node.component.html
tarablack01: @SEAjamieD did you test what @msorens mentioned about the [queryParams] not being applied to the page?

SEAjamieD: Yeah - I believe he was referring to using `link` which was applying the query params as comma separated.  So needed to stick with `routerLink`

Issue #2892, opened 2020-02-18
Event Feed lists "policys" instead of "policies" in guitar-string mouse-overs


mattray:
![policys](https://user-images.githubusercontent.com/330023/74704166-72d4a580-5263-11ea-8ee8-750d22e00e8e.png)
Build: 20200131232134

SEAjamieD:
I'll wrap this into whatever ticket I do next

SEAjamieD:
note to self: this is happening in event-feed-guitar-strings and types.ts

PR #3392, opened 2020-04-16
Patch for policies value and added test

SEAjamieD:
In the event feed guitar strings on the event feed dashboard, the word 'policy' was getting pluralized to 'policys' - this corrects pluralization to 'policies'

fixes https://github.com/chef/automate/issues/2892

### :+1: Definition of Done
Tooltips no longer show the work "policys" when they are plural, and now reflect the correct. "policies"

build component/automate-ui-devproxy && start_all_services
make serve

in hab studio run the following a few times:
`cat components/ingest-service/examples/actions/policy_update.json | jq --arg date "$(date +%FT%TZ -d '1 day ago')" '.recorded_at = $date' | send_chef_data_raw lb`

navigate to https://a2-dev.test/dashboards/event-feed
Hover over the guitar string icons in the following photo until you see the tooltip pop up.
Screen Shot 2020-04-16 at 12 07 59 PM

Wherever there has been an update to a policy, we should the name reflected as `policy` if singular and `policies` if plural.

- [x] Tests added/updated?
- [ ] Docs added/updated?

after:
Screen Shot 2020-04-16 at 12 04 24 PM
Screen Shot 2020-04-16 at 12 04 31 PM

### DIFF
--- a/components/automate-ui/src/app/page-components/event-feed-guitar-strings/event-feed-guitar-strings.component.spec.ts
+++ b/components/automate-ui/src/app/page-components/event-feed-guitar-strings/event-feed-guitar-strings.component.spec.ts
@@ -106,7 +106,7 @@ describe('EventFeedGuitarStringsComponent', () => {
     const text = component.getTooltipText(item, item.isMultiple(), guitarString);
 
     expect(text).toBe('<strong>5 nodes</strong>, <strong>9 cookbooks</strong>,' +
-      ' and <strong>7 roles</strong> were created.');
+      ' <strong>7 roles</strong>, and <strong>3 policies</strong> were created.');
   });
 
   describe('GuitarStringDataContainer', () => {
@@ -393,7 +393,8 @@ describe('EventFeedGuitarStringsComponent', () => {
       types = [
         {'name': 'node', 'count': 5},
         {'name': 'cookbook', 'count': 9},
-        {'name': 'role', 'count': 7}
+        {'name': 'role', 'count': 7},
+        {'name': 'policy', 'count': 3}
       ];
     } else {
       types = [
--- a/components/automate-ui/src/app/types/types.ts
+++ b/components/automate-ui/src/app/types/types.ts
@@ -1,4 +1,4 @@
-import { merge, endsWith, replace } from 'lodash';
+import { merge } from 'lodash';
 import * as moment from 'moment';
 
 export type Status = 'success' | 'failure' | 'missing' | 'skipped';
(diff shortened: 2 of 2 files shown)


SEAjamieD:
@vjeffrey / @chef/ux-team 

We've got a few options for this fix.  The first is something like what I added.  Its a very simple patch and will work fine for now, but has the possibility of returning incorrect pluralizations again in the future.

Second is copying the pluralize function from https://github.com/chef/automate/blob/master/components/automate-ui/src/app/page-components/event-feed-table/event-feed-table.component.ts#L123 - this will take a little bit longer because there are typescript complaints to work through re: RexExp vs String.  It is a little more hearty will definitely last longer.

the Third option, which I feel is the "right" thing to do, Is to take the larger function `getEventTypeLabel()` which is nearly identical out of the event feed and the types files and move them to a top level Event Feed file, export them, and use them in both the locations they are currently being used.  This is the one that will take the longest - but I'd argue is the right way to do it.

I guess that leaves a 4th option - which is more time respective, which is to make a new ticket to refactor this in the future.

thoughts?

vjeffrey:
i think i would vote for option 4 + option 3, in other words, get this in, make an issue to do option 3 -- id love to hear other thoughts tho!

SEAjamieD:
@vjeffrey - Thanks for adding that!  I had written down all the instructions to test and never clicked save 😂

msorens:
How about one more option? 😁 Use our general-purpose `pluralize` pipe (components/automate-ui/src/app/pipes/pluralize.pipe.ts). I suspect you discounted this @SEAjamieD because you are in TypeScript rather than in a template. But pipes can easily be used on the TypeScript side, too.

See:
1. [Is it possible to use a pipe in the code?](https://stackoverflow.com/questions/35159079/is-it-possible-to-use-a-pipe-in-the-code)
2. Example in our code base: https://github.com/chef/automate/blob/59410910c85ee2cb32ecaafb10d713103baecfbc/components/automate-ui/src/app/modules/desktop/check-in-time-series/check-in-time-series.component.ts#L37

SEAjamieD:
> How about one more option? 😁 Use our general-purpose `pluralize` pipe (components/automate-ui/src/app/pipes/pluralize.pipe.ts). I suspect you discounted this @SEAjamieD because you are in TypeScript rather than in a template. But pipes can easily be used on the TypeScript side, too.

That's _exactly_ what I did!  I had no idea!  Thanks for the info - i'll give that a try first 👍

PR #3403, opened 2020-04-17
automate-cli/Makefile: attempt to fix CLI docs issue around removed commands

srenatus:
This should fix stale CLI docs of removed commands sticking around.

The removed yamls aren't included in this PR, merging it should show us
if it works, or if there's state somewhere else that needs to be reset.

Running this in the dev env with `env CHEF_DEV_ENVIRONMENT=false make docs`
yields the following deleted files:

    deleted:    components/automate-chef-io/data/docs/cli_chef_automate/chef-automate_uninstall.yaml
    deleted:    components/automate-chef-io/data/docs/cli_chef_automate/commands/chef-automate_airgap_bundle_unpack.yaml
    deleted:    components/automate-chef-io/data/docs/cli_chef_automate/commands/chef-automate_debug.yaml
    deleted:    components/automate-chef-io/data/docs/cli_chef_automate/commands/chef-automate_debug_get-version.yaml
    deleted:    components/automate-chef-io/data/docs/cli_chef_automate/commands/chef-automate_debug_profile.yaml
    deleted:    components/automate-chef-io/data/docs/cli_chef_automate/commands/chef-automate_debug_set-log-level.yaml
    deleted:    components/automate-chef-io/data/docs/cli_chef_automate/commands/chef-automate_debug_trace.yaml
    deleted:    components/automate-chef-io/data/docs/cli_chef_automate/commands/chef-automate_diagnostics.yaml
    deleted:    components/automate-chef-io/data/docs/cli_chef_automate/commands/chef-automate_iam_reset-to-v1.yaml
    deleted:    components/automate-chef-io/data/docs/cli_chef_automate/commands/chef-automate_iam_upgrade-to-v2.yaml
    deleted:    components/automate-chef-io/data/docs/cli_chef_automate/commands/chef-automate_node-inventory.yaml

PR #3405, opened 2020-04-17
automate-chef-io: remove stale CLI doc data files

srenatus:
The proper fix would be #3403, but it seems some data files aren't rebuilt.

Figuring out why that is would be interesting, but this fixes the issue
at hand.

Re-running `env CHEF_DEV_ENVIRONMENT=false make docs` in `components/automate-cli` (in the studio) *won't* rebuild those files.


srenatus:
Wrong thing. Right thing: #3403.

PR #3411, opened 2020-04-17
Add log_disconnects and client_min_messages to postgres config

jaym:
Being able to set these is part of some hardening requirements for
a customer.
Also, `log_disconnects` is useful for making sure we're properly
reusing postgres connections.

We should release this when we consume the next core plans refresh


components/automate-postgresql/habitat/default.toml
```diff
@@ -15,6 +15,8 @@ checkpoint_timeout = "5mins"
 checkpoint_completion_target = 0.5
 md5_auth_cidr_addresses = ["0.0.0.0/0", "::0/0"]
 ssl_ciphers = "HIGH:+3DES:!aNULL:!eNULL:!EXPORT"
+log_disconnects = "off"
+client_min_messages = "NOTICE"
```
stevendanna, line 19: Should we put these in config_request.go as well?

PR #3414, opened 2020-04-17
[automate-3112] refactor team member APIs in storage layer

bcmdarroch:

doin' this TODO:
```
// TODO (tc): Get the updated set of user membership for the team since	
// that is not what is returned from AddUsers for some reason	
// (can refactor on V1 deprecation).
```

follows: https://github.com/chef/automate/pull/3361

### :+1: Definition of Done

- [x] AddUsers and RemoveUsers storage functions always return the final updated list of user IDs, instead of the Team object
- [x] AddUsers and RemoveUsers server functions only need to make one storage call instead of multiple

- `rebuild components/teams-service`
- add and remove users from the API or UI. everything should work as usual

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/teams-service/server/v2/server.go
+++ b/components/teams-service/server/v2/server.go
@@ -142,27 +142,13 @@ func (s *Server) AddTeamMembers(ctx context.Context,
 		return nil, status.Error(codes.InvalidArgument, "missing user IDs")
 	}
 
-	// verify team exists and isn't filtered out by the project filter
-	_, err := s.service.Storage.GetTeam(ctx, req.Id)
-	if err != nil {
-		return nil, service.ParseStorageError(err, req.Id, "team")
-	}
-
-	_, err = s.service.Storage.AddUsers(ctx, req.Id, req.UserIds)
+	updatedUserIDs, err := s.service.Storage.AddUsers(ctx, req.Id, req.UserIds)
 	if err != nil && err != storage.ErrConflict {
 		return nil, service.ParseStorageError(err, req.Id, "user")
 	}
 
-	// TODO (tc): Get the updated set of user membership for the team since
-	// that is not what is returned from AddUsers for some reason
-	// (can refactor on V1 deprecation).
-	userIDs, err := s.service.Storage.GetUserIDsForTeam(ctx, req.Id)
-	if err != nil {
-		return nil, service.ParseStorageError(err, req.Id, "team")
-	}
-
 	return &teams.AddTeamMembersResp{
-		UserIds: userIDs,
+		UserIds: updatedUserIDs,
 	}, nil
 }
 
@@ -170,27 +156,13 @@ func (s *Server) AddTeamMembers(ctx context.Context,
 func (s *Server) RemoveTeamMembers(ctx context.Context,
 	req *teams.RemoveTeamMembersReq) (*teams.RemoveTeamMembersResp, error) {
 
-	// verify team exists and isn't filtered out by the project filter
-	_, err := s.service.Storage.GetTeam(ctx, req.Id)
-	if err != nil {
-		return nil, service.ParseStorageError(err, req.Id, "team")
-	}
-
-	_, err = s.service.Storage.RemoveUsers(ctx, req.Id, req.UserIds)
-	if err != nil {
-		return nil, service.ParseStorageError(err, req.Id, "team")
-	}
-
-	// TODO (tc): Get the updated set of user membership for the team since
-	// that is not what is returned from RemoveUsers for some reason
-	// (can refactor on V1 deprecation).
-	userIDs, err := s.service.Storage.GetUserIDsForTeam(ctx, req.Id)
+	updatedUserIDs, err := s.service.Storage.RemoveUsers(ctx, req.Id, req.UserIds)
 	if err != nil {
 		return nil, service.ParseStorageError(err, req.Id, "team")
 	}
 
 	return &teams.RemoveTeamMembersResp{
-		UserIds: userIDs,
+		UserIds: updatedUserIDs,
 	}, nil
 }
 
--- a/components/teams-service/server/v2/server_test.go
+++ b/components/teams-service/server/v2/server_test.go
@@ -1007,9 +1007,10 @@ func runAllServerTests(
 					resp, err := cl.CreateTeam(ctx, req)
 					require.NoError(t, err)
 
+					targetMemberID := "88f13b6b-b20b-4335-9fd6-2c09edf45cf9"
 					resp1, err := cl.AddTeamMembers(insertProjectsIntoNewContext([]string{constants.UnassignedProjectID}), &teams.AddTeamMembersReq{
 						Id:      req.Id,
-						UserIds: []string{"88f13b6b-b20b-4335-9fd6-2c09edf45cf9"},
+						UserIds: []string{targetMemberID},
 					})
 					require.NoError(t, err)
 					require.Equal(t, 1, len(resp1.UserIds))
@@ -1026,7 +1027,7 @@ func runAllServerTests(
 					require.NotNil(t, resp2)
 					assert.Equal(t, len(addReq.UserIds)+1, len(resp2.UserIds))
 
-					expectedUsers := append(addReq.UserIds, "88f13b6b-b20b-4335-9fd6-2c09edf45cf9")
+					expectedUsers := append(addReq.UserIds, targetMemberID)
 					assert.ElementsMatch(t, expectedUsers, resp2.UserIds)
 
 					cleanupTeam(t, cl, resp.Team.Id)
--- a/components/teams-service/server/v2/server_test.go
+++ b/components/teams-service/server/v2/server_test.go
@@ -983,7 +983,58 @@ func runAllServerTests(
 			}
 		})
 
-		t.Run("fails to add user with NotFound when a project filter that excludes the team", func(t *testing.T) {
+		t.Run("successfully adds user to team with existing users", func(t *testing.T) {
+			tests := []struct {
+				users []string
+				desc  string
+			}{
+				{[]string{"6ed95714-9466-463b-80da-0513ecb42a08"}, "single user"},
+				{[]string{
+					"299ea25b-62d4-4660-965a-e25870298792",
+					"d1f642c8-8907-4e8b-a9a0-b998a44dc4bf",
+				}, "multiple users"},
+			}
+			for _, test := range tests {
+				t.Run("when provided valid team and "+test.desc, func(t *testing.T) {
+					ctx := context.Background()
+
+					// arrange
+					req := &teams.CreateTeamReq{
+						Name:     "Gotta Catch Em All",
+						Id:       "corgis-inc",
+						Projects: []string{},
+					}
+					resp, err := cl.CreateTeam(ctx, req)
+					require.NoError(t, err)
+
+					resp1, err := cl.AddTeamMembers(insertProjectsIntoNewContext([]string{constants.UnassignedProjectID}), &teams.AddTeamMembersReq{
+						Id:      req.Id,
+						UserIds: []string{"88f13b6b-b20b-4335-9fd6-2c09edf45cf9"},
+					})
+					require.NoError(t, err)
+					require.Equal(t, 1, len(resp1.UserIds))
+
+					addReq := &teams.AddTeamMembersReq{
+						Id:      req.Id,
+						UserIds: test.users,
+					}
+					// act
+					resp2, err := cl.AddTeamMembers(insertProjectsIntoNewContext([]string{constants.UnassignedProjectID}), addReq)
+
+					// assert
+					require.NoError(t, err)
+					require.NotNil(t, resp2)
+					assert.Equal(t, len(addReq.UserIds)+1, len(resp2.UserIds))
+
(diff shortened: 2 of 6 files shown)


components/teams-service/server/v2/server.go
```diff
@@ -142,55 +142,27 @@ func (s *Server) AddTeamMembers(ctx context.Context,
...
 	}
 
-	// verify team exists and isn't filtered out by the project filter
-	_, err := s.service.Storage.GetTeam(ctx, req.Id)
-	if err != nil {
-		return nil, service.ParseStorageError(err, req.Id, "team")
-	}
-
-	_, err = s.service.Storage.AddUsers(ctx, req.Id, req.UserIds)
+	updatedUserIDs, err := s.service.Storage.AddUsers(ctx, req.Id, req.UserIds)
```
bcmdarroch, line 145: the goal of this PR: simplify this server function to one single storage call that does all the work: 
- checking team existence within the project filter
- adding (or removing) the given user IDs
- returning the final updated list of user IDs

components/teams-service/server/v2/server_test.go
```diff
@@ -983,7 +983,58 @@ func runAllServerTests(
 			}
 		})
 
-		t.Run("fails to add user with NotFound when a project filter that excludes the team", func(t *testing.T) {
+		t.Run("successfully adds user to team with existing users", func(t *testing.T) {
```
bcmdarroch, line 986: realized we were missing this test scenario. all the other tests add users to teams with no users

components/teams-service/storage/conformance/conformance_test.go
```diff
@@ -172,20 +171,19 @@ func testGetTeamsForUser(ctx context.Context, t *testing.T, s storage.Storage) {
...
 	require.NoError(t, err, "setup: failed to add users to team-2-with-user")
 	_, err = s.AddUsers(ctx, idWithout, []string{"two", "three"})
 	require.NoError(t, err, "setup: failed to add users to test team-without-user")
 
 	returnedTeams, err := s.GetTeamsForUser(ctx, "user-for-team-fetch")
 	require.NoError(t, err)
-	expectedFetchedTeams := []storage.Team{team, team2}
-	ids := []string{returnedTeams[0].ID, returnedTeams[1].ID}
-	expectedIDs := []string{expectedFetchedTeams[0].ID, expectedFetchedTeams[1].ID}
+	expectedFetchedTeamIDs := []string{id, id2}
```
bcmdarroch, line 183: since we no longer have team GUIDs to worry about and we can just use the user-created ID, had the opportunity to simplify some test code

components/teams-service/storage/postgres/postgres.go
```diff
@@ -258,42 +258,44 @@ func (p *postgres) GetTeams(ctx context.Context) ([]storage.Team, error) {
...
-		WHERE user_id = ANY($1) AND team_db_id = team_db_id($2)`,
-		pq.Array(userIDs), id)
-	if err != nil {
-		return storage.Team{}, p.processError(err)
-	}
-	num, err := res.RowsAffected()
+	var dbID int
+	var teamProjects []string
+	// ensure the team exists and isn't filtered out by the project filter
+	err = p.db.QueryRowContext(ctx,
```
bcmdarroch, line 273: could be a candidate for a helper function since I used this across AddUsers, RemoveUsers, and GetTeamForUsers

components/teams-service/storage/postgres/postgres.go
```diff
@@ -362,24 +365,62 @@ func (p *postgres) PurgeUserMembership(ctx context.Context, userID string) ([]st
...
-				WHERE id=$1
-			RETURNING team_db_id
-		)
-		UPDATE teams SET updated_at=NOW()
-		WHERE db_id in (SELECT DISTINCT * FROM moved_row_ids)
-		RETURNING id, name, projects, created_at, updated_at;`, id, pq.Array(userIDs)).Scan(
-		&t.ID, &t.Name, pq.Array(&t.Projects), &t.CreatedAt, &t.UpdatedAt)
+	 			FROM teams
+				WHERE db_id=$1
+				ON CONFLICT ON CONSTRAINT teams_users_pkey
```
bcmdarroch, line 397: these lines, helpfully added by @vsingh-msys, handle the case when the user already exists on the team

components/teams-service/storage/postgres/postgres.go
```diff
@@ -362,24 +365,62 @@ func (p *postgres) PurgeUserMembership(ctx context.Context, userID string) ([]st
...
+	projectsFilter, err := ProjectsListFromContext(ctx)
+	if err != nil {
+		return nil, p.processError(err)
+	}
+
+	var dbID int
+	var teamProjects []string
+	// ensure the team exists and isn't filtered out by the project filter
+	err = p.db.QueryRowContext(ctx,
+		`SELECT db_id, projects FROM teams
```
bcmdarroch, line 384: This function now completes the following queries in this order:
within transaction:
- SELECT the team with the projects filter to verify that it exists within the filter
- INSERT INTO user IDs in the team_user_association table, doing nothing on conflict
- UPDATE the team's `updated_at` field

after that transaction completes, we fetch the user IDs from the team.

Open to feedback on making this more concise!

components/teams-service/storage/postgres/postgres.go
tylercloke: We should start and do all of these related queries in a transaction so things don't change from under us.

components/teams-service/storage/postgres/postgres.go
tylercloke: Same transaction comment for this function.

components/teams-service/storage/postgres/postgres.go
bcmdarroch: Oh good point, I missed adding the transaction here

components/teams-service/storage/postgres/postgres.go
bcmdarroch: Right now I've got the first three queries in a transaction, but I had to move the SELECTing of the final user IDs outside of the transaction since it wasn't picking up the changes to the user table.

components/teams-service/storage/postgres/postgres.go
```diff
@@ -362,24 +365,62 @@ func (p *postgres) PurgeUserMembership(ctx context.Context, userID string) ([]st
...
+		WHERE id = $1 AND projects_match(projects, $2::TEXT[])`,
+		id, pq.Array(projectsFilter)).
+		Scan(&dbID, pq.Array(&teamProjects)) // TODO is there a way to get these projects without returning them?
```
bcmdarroch, line 387: oh yeah I had a question if there was another way to write this query that gets the team's projects for the projects_match but doesn't return the projects (since we just need them for that check)

components/teams-service/storage/postgres/postgres.go
```diff
@@ -390,11 +431,28 @@ func (p *postgres) AddUsers(ctx context.Context,
...
+	projectsFilter, err := ProjectsListFromContext(ctx)
+	if err != nil {
+		return nil, p.processError(err)
+	}
+
+	var dbID int
+	var teamProjects []string
+	// ensure the team exists and isn't filtered out by the project filter
+	err = p.db.QueryRowContext(ctx,
+		`SELECT db_id, projects FROM teams
```
bcmdarroch, line 443: There must be a way to combine these  two SELECTs 🤔

components/teams-service/storage/postgres/postgres.go
msorens: I exercised the raw SQL inside a transaction in an interactive session and it definitely picks up the changes if the final query is inside the transaction (as it is supposed to). When run from Go code, as you observed, it is not. That tells us that the Go interface is the culprit, and indeed it is: you were not asking for data from the transaction.
Change the line above from this:
```
p.db.QueryRowContext(ctx,
```
to this:
```
tx.QueryRowContext(ctx,
```

components/teams-service/storage/postgres/postgres.go
```diff
@@ -258,42 +258,56 @@ func (p *postgres) GetTeams(ctx context.Context) ([]storage.Team, error) {
 // server code before we get here if the team in question was filtered, but if we call this
```
srenatus, line 258: Is this comment still accurate? (Two lines up)
> NOTE: We aren't filtering on projects here [...]

components/teams-service/storage/postgres/postgres.go
```diff
@@ -258,42 +258,56 @@ func (p *postgres) GetTeams(ctx context.Context) ([]storage.Team, error) {
...
 	}
 
 	err = tx.Commit()
 	if err != nil {
-		return storage.Team{}, p.processError(err)
+		return nil, p.processError(err)
 	}
-	return team, nil
+
+	// now that the transaction is complete and the team's users are updated,
```
srenatus, line 300: Why do we do this _after_ the transaction? If we ran it inside, would the response differ?

components/teams-service/storage/postgres/postgres.go
```diff
@@ -362,24 +377,64 @@ func (p *postgres) PurgeUserMembership(ctx context.Context, userID string) ([]st
 // from a different context we should apply project filtering in this function as well.
```
srenatus, line 377: Oh look, the same comment, isn't it? Let's update it, too.

components/teams-service/storage/postgres/postgres.go
srenatus: Clarity might be subjective, I find (1.) clearer. Even more so if we had a `team_projects(id)` helper, maybe:
```
SELECT db_id FROM teams WHERE id = $1 AND
  projects_match(team_projects(id), $2::TEXT[]);
```
But I can't shake the feeling that this could be simpler.

Doesn't this do the trick?
```
SELECT db_id FROM teams
WHERE id = $1 AND projects_match(projects, $2::TEXT[])
```
-- it's your query without returning the `projects` (from the `SELECT`) -- after all, `projects` is a column on `teams`, so this should be valid.

components/teams-service/storage/postgres/postgres.go
```diff
@@ -362,24 +377,64 @@ func (p *postgres) PurgeUserMembership(ctx context.Context, userID string) ([]st
...
+				ON CONFLICT ON CONSTRAINT teams_users_pkey
+				DO NOTHING
+				`, dbID, pq.Array(userIDs))
 	if err != nil {
-		return storage.Team{}, p.processError(err)
+		return nil, p.processError(err)
 	}
-	return t, nil
+
+	_, err = tx.ExecContext(ctx, `UPDATE teams SET updated_at=NOW()
```
srenatus, line 416: Could be a follow up issue to do these via triggers? 💭Or did we decide not to do that? Can't remember right now.

components/teams-service/storage/postgres/postgres.go
srenatus: Do we want to use a transaction? We're using the the result from query 1 in query 2, and without a tx the state of the world (or database table) could have changed in between... I think

components/teams-service/storage/postgres/postgres.go
bcmdarroch: oh I thought I'd tried this and the tests failed 🤔 but it worked so must have been mislead by something else! 
Thanks! :)

components/teams-service/storage/conformance/conformance_test.go
```diff
@@ -322,16 +320,13 @@ func testAddUsers(ctx context.Context, t *testing.T, s storage.Storage) {
...
+	updatedUserIDs, err := s.AddUsers(ctx, id, userIDs)
+	assert.NotNil(t, updatedUserIDs)
 
-	userIDs, err := s.GetUserIDsForTeam(ctx, id)
-	require.NoError(t, err)
-
-	assert.Equal(t, len(users), len(userIDs), "number of users added does not match")
-	for _, user := range users {
-		assert.Contains(t, userIDs, user, "user %q not added", user)
+	assert.Equal(t, len(userIDs), len(updatedUserIDs), "number of users added does not match")
```
blakestier, line 327: do we have a test that adds users, then adds them again to confirm we get the *new* list back?

bcmdarroch: we do have a test that makes sure we don't duplicate user IDs if we add the same user twice? I'm not sure if that's the same case you mean

components/teams-service/storage/postgres/postgres.go
bcmdarroch: woops missed that

components/teams-service/storage/postgres/postgres.go
bcmdarroch: It was failing when I originally had it inside the transaction so I moved it out. But I just moved it back in and it still passes, so I must have misunderstood the original test failure.

components/teams-service/storage/postgres/postgres.go
bcmdarroch: Can't remember off the top of my head but I can add the card

bcmdarroch: https://github.com/chef/automate/issues/3471

components/teams-service/storage/postgres/postgres.go
bcmdarroch: decided to try my hand at an `INNER JOIN` and got it working ✨

PR #3416, opened 2020-04-17
Only HTTPS on API

kagarmoe:

Removes `http` from the schema.

Screen Shot 2020-04-17 at 3 55 22 PM


susanev:
youll need to rebase i think, #3415 should fix that fail

PR #3419, opened 2020-04-18
Alternate versioning strategy

kagarmoe:

We need to get the versioning right in the APIs.
This PR:

Removes `basePath` from 00-meta.json
Does a brute-force search and replace to get `apis\iam\v2`

Next Steps: pre-pend the remaining endpoints

I'm surprised that it even built and I expect that I changed things that shouldn't be changed. Even though I expect this won't work, it is probably the approximate state we need to get to.

Screen Shot 2020-04-17 at 6 02 16 PM


msorens:
I really appreciate your trying to tackle this thorny issue!

One big problem though: the gateway uses the paths in the proto file to map to the remote procedure calls. It takes whatever is defined for the HTTP method key and appends that to this: e.g. "https://a2-dev.test/apis/" for IAM calls:
```
option (google.api.http) = {
      get: "/apis/iam/v2/teams"
    };
```
That means the complete URL is now--notice the double `apis` term--this:
```
https://a2-dev.test/apis/apis/iam/v2/teams
```
And that breaks everything IAM. 🤷 In other words, we cannot change the proto paths unless we adjust the generated URLs in tandem.

That is where the load balancer acts on the `/apis` portion... 
https://github.com/chef/automate/blob/57577936c94209bb787d8e6e7012b76ec108c79b/components/automate-load-balancer/habitat/config/nginx.conf#L198

Once through the load balancer, here is where the gateway then acts on it:
https://github.com/chef/automate/blob/c8f110926945f0363759effd949e3cab1218fd62/components/automate-gateway/gateway/server.go#L346

What we need to find, though, is where the first `/apis` fragment is generated; not sure where that is.

components/automate-gateway/api/iam/v2/introspect.pb.policy.go
```diff
@@ -9,13 +9,13 @@ import (
 )
 
 func init() {
-	policy.MapMethodTo("/chef.automate.api.iam.v2.Authorization/IntrospectAll", "iam:introspect", "iam:introspect:getAll", "GET", "/iam/v2/introspect", func(unexpandedResource string, input interface{}) string {
+	policy.MapMethodTo("/chef.automate.api.iam.v2.Authorization/IntrospectAll", "iam:introspect", "iam:introspect:getAll", "GET", "/apis/iam/v2/introspect", func(unexpandedResource string, input interface{}) string {
```
srenatus, line 12: 💡 This is another, uhm, side-effect I hadn't kept in mind. The API paths are used for introspection. I'm not sure if this will _just continue to work_ or if there's extra hoops to jump through. @msorens What do you think?

srenatus:
Looking into this some more, it turns out that **for `/apis`, we're already sending the prefix along:
1. The load-balancer send its along (via the absence of a trailing `/` in `proxy_pass https://automate-gateway;`: https://github.com/chef/automate/blob/31442fe17fe52e537809ca6bf9ea21eb729f636a/components/automate-load-balancer/habitat/config/nginx.conf#L198-L203
2. There's code in `automate-gateway` to make the _unprefixed paths_ (from the proto annotations) work with the _prefixed requests_, here: https://github.com/chef/automate/blob/31442fe17fe52e537809ca6bf9ea21eb729f636a/components/automate-gateway/gateway/server.go#L346

So, removing (2.) should make the fixed `/apis` annotations (or rather, the code generated therefrom) match what the server's `versionedMux` receives.

For the remaining endpoints, we're removing both the prefixes `/api/v0` and `/api/v1` (I'm not sure if anyone uses that) in the load-balancer, and handle them as-is in the gateway. I suppose if we changed them to a similar treatment as happens with `/apis`, stuff should work out.

As mentioned in my inline comment, I'm not 100% certain what'll happen to introspection, but let's find out.

srenatus:
Jinx! I should have reloaded this page. Turns out @msorens had already collected all the evidence 😅

msorens:
Evidence, yes, but I did not quite grok the actual fix. Thanks, @srenatus !

components/automate-gateway/api/iam/v2/introspect.pb.policy.go
msorens: Good catch! I am thinking this will probably fail but will have to take a closer look.

srenatus:
I'll adopt this WIP. I'm using a new branch, though, it'll be `sr+kg/api-versions`.

The laundry list is
- [ ] fix verify
  - [ ] `[unit] automate-gateway` has an introspection failure, I'll see if fixing this is equivalent to the required introspection changes
- [ ] fix verify_private

kagarmoe:
Thanks @srenatus !

susanev:
see #3474

Issue #3394, opened 2020-04-16
ssm scans not working


vjeffrey:
## User Story

token retrieval for our ssm scans is not currently working;
```
rpc error: code = PermissionDenied desc = Projects cannot be modified because the [] subject is missing iam:projects:assign on this project: ["(unassigned)"]
```

PR #3421, opened 2020-04-18
 fix ssm jobs

vjeffrey:
we now have to add a token to an ingest policy after creation for that token to be a valid ingest token; this change makes that happen

https://github.com/chef/automate/issues/3394

### :+1: Definition of Done
tokens created for ssm jobs work

to truly test this, you need to bring up the dev env in aws and create an ec2 node integration, then try scanning some ssm nodes

the other way to test locally is to call the runssmjob function from the regular scan job path, log the token, and then ensure the token works for sending in a report and fetching a profile

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/compliance-service/compliance.go
+++ b/components/compliance-service/compliance.go
@@ -13,7 +13,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/chef/automate/api/external/secrets"
-	auth "github.com/chef/automate/api/interservice/authn"
+	authn "github.com/chef/automate/api/interservice/authn"
 	iam_v2 "github.com/chef/automate/api/interservice/authz/v2"
 	"github.com/chef/automate/api/interservice/compliance/ingest/ingest"
 	"github.com/chef/automate/api/interservice/compliance/jobs"
@@ -474,7 +474,7 @@ func setupDataLifecyclePurgeInterface(ctx context.Context, connFactory *secureco
 func setup(ctx context.Context, connFactory *secureconn.Factory, conf config.Compliance,
 	esr relaxting.ES2Backend, db *pgdb.DB, cerealManager *cereal.Manager) error {
 	var err error
-	var conn, mgrConn, secretsConn, authConn *grpc.ClientConn
+	var conn, mgrConn, secretsConn, authnConn, authzConn *grpc.ClientConn
 	timeoutCtx, cancel := context.WithTimeout(ctx, 15*time.Second)
 	defer cancel()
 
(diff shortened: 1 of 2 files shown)


components/compliance-service/compliance.go
```diff
@@ -554,22 +554,36 @@ func setup(ctx context.Context, connFactory *secureconn.Factory, conf config.Com
...
-		authClient := auth.NewTokensMgmtClient(authConn)
-		if authClient == nil {
+		authnClient := authn.NewTokensMgmtClient(authnConn)
+		if authnClient == nil {
 			logrus.Errorf("serveGrpc got nil for NewTokensMgmtClient: %s", err)
 			return err
 		}
+		// get authz connection
+		authzConn, err = connFactory.DialContext(timeoutCtx, "authz-service", fmt.Sprintf("%s:%d", conf.Authz.HostBind, conf.Authz.Port),
+			grpc.WithBlock())
```
vjeffrey, line 571: we already have the authz host/port info bc we use it during project tagging and updating

components/compliance-service/inspec-agent/remote/remote.go
```diff
@@ -29,16 +33,28 @@ func RunSSMJob(ctx context.Context, ssmJob *types.InspecJob) *inspec.Error {
 	logrus.Debugf("Running ssm job: %+v", ssmJob)
 	// 1. CREATE TOKEN
 	// this is needed to hand over to inspec so it can report to automate
-	if RemoteJobInfo.TokensMgmtClient == nil {
+	if RemoteJobInfo.TokensMgmtClient == nil || RemoteJobInfo.PoliciesClient == nil {
```
vjeffrey, line 36: make sure we fail if we dont have the policies client

components/compliance-service/inspec-agent/remote/remote.go
```diff
@@ -29,16 +33,28 @@ func RunSSMJob(ctx context.Context, ssmJob *types.InspecJob) *inspec.Error {
...
+		Name:   tokenID,
 		Active: true,
 	})
 	if err != nil {
-		logrus.Errorf("unable to create auth token for reporting to automate: %s", err.Error())
+		logrus.Errorf("unable to create auth token for reporting to automate; aborting job %s: %s", ssmJob.JobName, err.Error())
+		return translateToInspecErr(err)
+	}
+	// add token to ingest policy
+	_, err = RemoteJobInfo.PoliciesClient.AddPolicyMembers(ctx, &v2.AddPolicyMembersReq{
```
vjeffrey, line 52: this is the real new stuff/the fix -- adding the token to the ingest policy

components/compliance-service/inspec-agent/remote/remote.go
```diff
@@ -29,16 +33,28 @@ func RunSSMJob(ctx context.Context, ssmJob *types.InspecJob) *inspec.Error {
 	logrus.Debugf("Running ssm job: %+v", ssmJob)
 	// 1. CREATE TOKEN
 	// this is needed to hand over to inspec so it can report to automate
-	if RemoteJobInfo.TokensMgmtClient == nil {
+	if RemoteJobInfo.TokensMgmtClient == nil || RemoteJobInfo.PoliciesClient == nil {
 		logrus.Error("unable to create auth token")
 		return translateToInspecErr(fmt.Errorf("unable to connect to auth client: aborting job run for job %+v", ssmJob))
 	}
+	tokenID := fmt.Sprintf("inspec-to-automate-scanjob-%s", ssmJob.JobID)
+	ctx = auth_context.NewOutgoingContext(auth_context.NewContext(ctx, []string{"tls:service:deployment-service:internal"}, []string{}, "", ""))
```
msorens, line 41: You are inadvertently masquerading as the deployment service here. That may be OK but there is a check for that in the gateway that is used for a specific purpose: https://github.com/chef/automate/blob/c8f110926945f0363759effd949e3cab1218fd62/components/automate-gateway/gateway/middleware/auth_interceptor.go#L246-L250

Probably better to use your own designator, perhaps:
```suggestion
	ctx = auth_context.NewOutgoingContext(auth_context.NewContext(ctx, []string{"tls:service:compliance-service:internal"}, []string{}, "", ""))
```

vjeffrey: ah yes! thank you!

PR #3082, opened 2020-03-11
Bump automate UI from Angular 8 -> Angular 9

msorens:
### :nut_and_bolt: Description: What code changed, and why?

This PR bumps our Angular framework from v8 to v9, current at the time of writing. Furthermore, this PR also includes freshening almost every other 3rd party package used by automate-ui, with just a few exceptions noted below. 
The steps for this global version update are detailed below.

#### Migrate Angular

This did two things:

1. Update the various `@angular` packages in package.json.
2. Automatically apply migration patches to our code base.

Commands:
`ng update @angular/core @angular/cli`
`ng update @angular/material`
`ng update @nguniversal/express-engine` (because `npm outdated` now revealed this was needed)

Reference: https://update.angular.io/#8.2:9.0l2

#### Update all other npm packages to current (or constrained) version

This was done with a combination of `npm install`, `npm update`, and `npm audit fix`.
The remaining items are all explicitly constrained:

```
$ npm outdated
Package                         Current      Wanted       Latest  Location
@nguniversal/express-engine  9.0.0-rc.2  9.0.0-rc.2        9.0.2  automate-ui
@types/jasmine                   3.3.10      3.3.10        3.5.9  automate-ui
@types/node                    12.12.29    12.12.29       13.9.0  automate-ui
diff2html                        2.12.2      2.12.2        3.1.3  automate-ui
immutable                         3.8.2       3.8.2  4.0.0-rc.12  automate-ui
rrule                             2.4.1       2.4.1        2.6.4  automate-ui
tslint                           5.20.1      5.20.1        6.1.0  automate-ui
typescript                        3.7.5       3.7.5        3.8.3  automate-ui
```

Reasons for these constraints are detailed in the README.md file.

#### Patch code incompatibilities due to all of the above

The few necessary patches, included in various commits herein, have achieved these milestones:

 - [x] build (`ng build`)
 - [x] unit tests (via wallaby in VSCode)
 - [x] unit tests (`make unit` via command-line)
 - [x] e2e tests (protractor via `make e2e`)
 - [x] lint (`make lint`)
 - [x] launch application (`make serve`)
 - [x] application runs with no errors in browser
 - [x] application runs with no warnings in browser

There are some [breaking changes](https://angular.io/guide/updating-to-version-9#new-breaking-changes) in Angular 9, but it does not appear that that broke anything for us. 🤞 

### :chains: Related Resources
Last Angular Update (v6 -> v8): #1707
Update node (v10 -> v12): #2579 
[...]

PR #3438, opened 2020-04-20
Move desktop dashboard to top-level route

scottopherson:
- Moved `/dashboards/desktop` to `/desktop`. Desktop dashboard will be a top-level route with its own link in the top-level navbar.
- Removed the sidebar in desktop dashboard view. Desktop dashboard will be a full-width view.

Before
Screen Shot 2020-04-20 at 12 22 50 PM

After
Screen Shot 2020-04-20 at 12 21 32 PM

### DIFF
--- a/components/automate-ui/src/app/modules/desktop/dashboard/dashboard.component.ts
+++ b/components/automate-ui/src/app/modules/desktop/dashboard/dashboard.component.ts
@@ -4,6 +4,7 @@ import { Store } from '@ngrx/store';
 import { NgrxStateAtom } from 'app/ngrx.reducers';
 import { map, filter } from 'rxjs/operators';
 import { last, reverse } from 'lodash/fp';
+import { LayoutFacadeService } from 'app/entities/layout/layout.facade';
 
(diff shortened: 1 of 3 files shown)


components/automate-ui/src/app/modules/desktop/dashboard/dashboard.component.ts
```diff
@@ -151,6 +153,8 @@ export class DashboardComponent implements OnInit {
       this.store.select(unknownDesktopDurationCounts).pipe(
       map(counts => counts.updated)
     );
+
+    setTimeout(() => this.layoutFacade.hideSidebar());
```
scottopherson, line 157: Used the same `setTimeout` workaround from https://github.com/chef/automate/pull/3082/commits/9eb7496997d9e77ba642b65be9db6f5f9f4568c4 to avoid:

Screen Shot 2020-04-20 at 12 30 24 PM

Need to investigate the implementation and usage of `layoutFacade` to find a way a better way around this particular issue.

Issue #3413, opened 2020-04-17
global project filter apply changes button issues


susanev:
## overview
the apply changes button should only be enabled when there are changes to apply. currently, if you make a change and then undo it the apply changes button stays enabled even though there are no new changes to apply.

## repro steps

1. create some projects
1. notice the disabled apply changes button in the global projects filter
1. check some boxes in the global projects filter
1. notice the apply changes button becomes enabled
1. uncheck the boxes you checked (or use the clear selection button)
1. notice the apply changes button does not go back to its original disabled state

## definition of done
if you have no new changes to apply then the apply changes button should be disabled

PR #3441, opened 2020-04-21
Global Projects Filter: Disable Apply Changes when No changes are present

SEAjamieD:
Currently when using the global projects filter, if a user makes selections, and then reverses those selections, they are still allowed to apply changes, even though no changes are present.

This branch disables the `apply changes` button when there are not "new" changes present.

Fixes: https://github.com/chef/automate/issues/3413

### :+1: Definition of Done
Users see a disabled `apply changes` button until they have made true changes to the filter selections.

make serve

In the UI, create a handful of projects at https://a2-dev.test/settings/projects
in the top right of any page, click on the global projects dropdown
check some of the projects, see the `apply changes` button activate
uncheck those projects, see the `apply changes` button deactivate

check some of the projects, see the `apply changes` button activate
click `clear selection` button, see the `apply changes` button deactivate

Check some of the projects again and click to apply
After the selection has been applied
check some of the projects, see the `apply changes` button activate
uncheck those projects, see the `apply changes` button deactivate

- [x] Tests added/updated?
- [ ] Docs added/updated?

Project selection

### DIFF
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
@@ -21,7 +21,8 @@
           <input
             chefInput
             type="text"
-            (keyup)="handleFilterKeyUp($event.target.value)"
+            #filter
+            (keyup)="handleFilterKeyUp(filter.value)"
             placeholder="Filter projects..." />
         </div>
 
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
@@ -1,5 +1,5 @@
 import { Component, EventEmitter, Input, Output } from '@angular/core';
-import { cloneDeep } from 'lodash/fp';
+import { cloneDeep, isEqual } from 'lodash/fp';
 import { ProjectsFilterOption } from 'app/services/projects-filter/projects-filter.reducer';
 import { ProjectConstants } from 'app/entities/projects/project.model';
 
@@ -32,8 +32,8 @@ export class ProjectsFilterDropdownComponent {
   // Filtered options is merely a copy of the editable options
   // so they can be filtered while maintaining the actual options.
   filteredOptions: ProjectsFilterOption[] = [];
-  // InitialFilters holds a copy of the initial state of filtered Options
-  initialFilters: ProjectsFilterOption[] = [];
+  // InitialFilters holds a boolean copy of the initial checked/unchecked state of filtered Options
+  initialFilters: boolean[] = [];
 
   optionsEdited = false;
 
@@ -44,7 +44,7 @@ export class ProjectsFilterDropdownComponent {
       this.filteredOptions = this.editableOptions = cloneDeep(this.options);
       // Keep a reference to the filteredOptions in initialFilters
       // to check deep equality when a user unchecks/checks same button
-      this.initialFilters = cloneDeep(this.filteredOptions);
+      this.initialFilters = this.filteredOptions.map(option => option.checked);
     }
   }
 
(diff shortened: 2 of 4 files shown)


components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
```diff
@@ -109,4 +116,12 @@ export class ProjectsFilterDropdownComponent {
       (element as HTMLElement).focus();
     }
   }
+
+  checkInitialEquality(): void {
+    // Check for deep equality against new changes, disabling
+    // apply button if changes are not new
+    JSON.stringify(this.initialFilters) === JSON.stringify(this.editableOptions)
```
msorens, line 123: This is an expensive comparison, likely too expensive, given that this method will likely be called every digest cycle. Two factors contribute to the expense: (a) converting each object to a string, and (b) including all properties of each `ProjectsFilterOption` when we only care about the `checked` property.

Two alternatives to consider.

**(1) Tighten the data.** 

First, rather than a copy of the entire array, take a "slice" of just the `checked` property:
```
this.initialFilters = this.filteredOptions.map(o => o.checked);
```
Then use the lodash/fp `isEqual` operator rather than `stringify` to be more efficient.

**(2) Track the changes as they occur.**

This is more in keeping with the Angular style but is just a tad more work to set up. That is, starting with `this.initialFilters` being a boolean array initialized to all false, when the user checks or unchecks an item, toggle the corresponding array element of `this.initialFilters`, too. Then the `disabled` property of the button would just need to watch `!this.initialFilters.some(identity)`, i.e., the Apply button should be enabled if-and-only-if any element of `this.initialFilters` is true.

msorens:
Sorry for being so verbose... sort of got carried away. 😁

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
SEAjamieD: Thanks for the suggestion.  I ended up going with option one because it was easier to follow.

Ultimately i'm using `isEqual`, instead of `isEqualWith` and a customizer function because the I couldn't quite get the customizer to operate in the fashion I expected it to.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.spec.ts
```diff
@@ -283,6 +283,12 @@ describe('ProjectsFilterDropdownComponent', () => {
       component.handleOptionChange({ detail: true }, 'Project 1');
       expect(component.optionsEdited).toEqual(true);
     });
+
+    it('checks for equality against the initial filters applied', () => {
```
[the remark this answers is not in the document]
SEAjamieD: Thanks for the advice @msorens - I've added a whole slew of tests against this now.  The filtered version was a bit more difficult to do in a repeatable manner so I instead chose to roll a few by hand.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.spec.ts
```diff
@@ -174,6 +174,277 @@ describe('ProjectsFilterDropdownComponent', () => {
...
+          (_o, index) => component.handleOptionChange({ detail: false }, `Project ${index + 1}`));
+        expect(component.optionsEdited).toBe(true);
+
+        component.editableOptions.forEach(
+          (_o, index) => component.handleOptionChange({ detail: true }, `Project ${index + 1}`));
+        expect(component.optionsEdited).toBe(false);
+      });
+    });
+
+    describe('when options have filters applied', () => {
```
scottopherson, line 403: :+1: I would agree with the behavior described in this test block. Changed options should enable the Apply button regardless if they are filtered afterwards.

Good feedback @msorens ensuring we specify behavior here.

SEAjamieD: Thanks @scottopherson - I appreciate the feedback!

SEAjamieD:
@msorens - out of the three typescript issues in this file, I was able to resolve one of them.  The other two need the same fix, but I can't seem to decipher how to resolve.  In order to not hold this up, I've decided to loop all the typescript updates needed for files i've touched into a single chore for myself here: https://github.com/chef/automate/issues/3472 where I will follow up with next week.

components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.html
```diff
@@ -21,7 +21,8 @@
           <input
             chefInput
             type="text"
-            (keyup)="handleFilterKeyUp($event.target.value)"
+            #filter
+            (keyup)="handleFilterKeyUp(filter.value)"
```
SEAjamieD, line 25: I updated this as a result of the documentation found here: https://angular.io/guide/user-input#!#type-the--*event*

which suggests that passing $event is dubious and can be simplified so that we are only getting what we need, using a reference to the element.

msorens: Nice--thank you!

PR #3442, opened 2020-04-21
configurable authz pg connections

bcmdarroch:

Aiming to help fix a customer issue where compliance/infra fall over when `ListRulesForProject` times out. 

implements a MaxConnections config option for authz [similar to how it's done in authn](https://github.com/chef/automate/blob/master/components/authn-service/tokens/pg/pg.go#L24)

### :+1: Definition of Done
- can set the max number of pg connections authz-service allows

save this as `test.toml`
```
[auth_z]
  [auth_z.v1]
    [auth_z.v1.sys]
      [auth_z.v1.sys.storage]
        max_connections = 5
```

this command should succeed:
```
chef-automate config patch test.toml
```

- [x] Tests added/updated?
- [ ] Docs added/updated?


components/authz-service/storage/postgres/postgres.go
```diff
@@ -37,12 +37,16 @@ func New(ctx context.Context, migConf migration.Config, dataMigConf datamigratio
...
-func initPostgresDB(ctx context.Context, pgURL string) (*sql.DB, error) {
+func initPostgresDB(ctx context.Context, pgURL string, maxConnections int32) (*sql.DB, error) {
 	d, err := db.PGOpenContext(ctx, pgURL)
 	if err != nil {
 		return nil, err
 	}
 
+	if maxConnections > 0 {
+		d.SetMaxOpenConns(maxConnections)
+	}
```
stevendanna, line 48: I'm wonder if we might get a bit more predictable performance by also bumping SetMaxIdleConns.

components/authz-service/storage/postgres/postgres.go
```diff
@@ -37,12 +37,16 @@ func New(ctx context.Context, migConf migration.Config, dataMigConf datamigratio
...
 
-func initPostgresDB(ctx context.Context, pgURL string) (*sql.DB, error) {
+func initPostgresDB(ctx context.Context, pgURL string, maxConnections int32) (*sql.DB, error) {
 	d, err := db.PGOpenContext(ctx, pgURL)
 	if err != nil {
 		return nil, err
 	}
 
+	if maxConnections > 0 {
+		d.SetMaxOpenConns(maxConnections)
```
stevendanna, line 47: I think you'll need to convert this to an int here or at some point in the call chain

```suggestion
		d.SetMaxOpenConns(int(maxConnections))
```

components/authz-service/storage/postgres/postgres.go
jaym: what do you think about having them be the same

stevendanna: For a small number of connections, I would advocate having them be the same.

components/authz-service/storage/postgres/postgres.go
bcmdarroch: we're setting it to `int` here: https://github.com/chef/automate/pull/3442/files#diff-7b11e67859dabc4eb026fda620cc2815R23

PR #3443, opened 2020-04-21
add compliance service to the universal access policy

vjeffrey:

i had changed the value to compliance service based on some pr
feedback but it seems that it does not work when compliance is referenced:
```
["tls:service:compliance-service:internal"] subject is missing iam:projects:assign
on this project: ["(unassigned)"]
```

my bad, i should have retested after making the change; i didnt realize compliance would be unauthorized for this?

https://github.com/chef/automate/pull/3421

### DIFF
--- a/components/authz-service/server/v2/system.go
+++ b/components/authz-service/server/v2/system.go
@@ -12,10 +12,14 @@ func SystemPolicies() []*storage.Policy {
 		ID:   constants.UniversalAccessPolicyID,
 		Type: storage.System,
 		// used by automate-cli for API requests through the gateway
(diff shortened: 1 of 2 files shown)


jaym:
this doesn't seem right. maybe we need to add a default policy for compliance service to be able to do this?

tylercloke:
@jaym we can make a new system policy only giving compliance service access to parts of the system. curious what your concern is with adding more general access for that service though.

tylercloke:
@jaym oh sorry saw you were commenting on a previous commit :doh:

PR #3447, opened 2020-04-21
adding proxy host logic

danielcbright:
Signed-off-by: Daniel Bright <danielcbright@gmail.com>

I updated the `automate-es-gateway` logic to allow for the setting of the `proxy_set_header Host` value to be one of the users choosing. This will allow for external ES to work when using an external cluster that uses hostname based routing (think most Nginx load balancers, Elastic Cloud, and probably more). It sets the default to `$http_host` if it is not defined, which is the current behaviour.

To enable the setting you would add this to a `config.toml` during deploy or via a config patch:

```
[esgateway.v1.sys.ngx.http]
proxy_set_header_host = "yourawesomeexternaleshostname:1234"
```

The `esgateway` api config was also updated to reflect the changes.

### :+1: Definition of Done

You can make the above changes, and they are reflected in the `automate-es-gateway`'s final, rendered `nginx.conf` in the `/hab/svc/automate-es-gateway/config/nginx.conf` file.

1. `start_all_services`
2. create a custom `patch.toml` with the above settings
3. patch the chef-automate install
4. verify that the `/hab/svc/automate-es-gateway/config/nginx.conf` reflects the changes

- [ ] Tests added/updated?
- [x] Docs added/updated?


components/automate-chef-io/content/docs/install.md
```diff
@@ -156,8 +156,14 @@ Add the following to your config.toml:
 # [global.v1.external.elasticsearch.ssl]
 #  Specify either a root_cert or a root_cert_file
 #  root_cert = """$(cat </path/to/cert_file.crt>)"""
-#  root_cert_file = "</path/to/cert/file>"
 #  server_name = "<elasticsearch server name>"
+
+# Uncomment and fill out if using external elasticsearch that uses hostname-based routing/load balancing
```
susanev, line 161: does this need to be added to the config docs page too?

danielcbright: I can certainly go ahead and make the additions there too!

components/automate-es-gateway/habitat/config/nginx.conf
```diff
@@ -72,7 +72,7 @@ http {
     client_body_buffer_size {{ cfg.ngx.http.client_body_buffer_size }};
     client_max_body_size {{cfg.ngx.http.client_max_body_size}};
 
-    proxy_set_header Host $http_host; # http_host preserves port # if present
+    proxy_set_header Host {{cfg.ngx.http.proxy_set_header_host}}; # http_host preserves port # if present
```
jaym, line 75: [nit] I would prefer for this case to to have the default value for `proxy_set_header_host` be null and do an if else based on that. For example:
```
{{#if cfg.ngx.http.proxy_set_header_host }}
proxy_set_header Host {{cfg.ngx.http.proxy_set_header_host}};
{{ else }}
proxy_set_header Host $http_host;
{{/if}}
```

components/automate-chef-io/content/docs/install.md
danielcbright: on second thought, I see that there is no documentation for the external ES, I think that's more for if you're using the "built-in" ES cluster vs outsourcing it

kavanista:
👍

components/automate-es-gateway/habitat/config/nginx.conf
danielcbright: will do!

components/automate-chef-io/content/docs/install.md
```diff
@@ -156,8 +156,14 @@ Add the following to your config.toml:
...
+# [esgateway.v1.sys.ngx.http]
+#  proxy_set_header_host = "yourawesomeexternaleshostname:1234"
```
mjingle, line 163:
```suggestion
#  proxy_set_header_host = "<your external es hostname>:1234"
```

mjingle, line 163: Is the "1234" supposed to be an example port number? Or the exact port to be used?

components/automate-chef-io/content/docs/install.md
```diff
@@ -156,8 +156,14 @@ Add the following to your config.toml:
...
+
+# Uncomment and add if you need to change the ssl_verify_depth for the root cert bundle
```
mjingle, line 165:
```suggestion
# Uncomment and add to change the ssl_verify_depth for the root cert bundle
```

components/automate-chef-io/content/docs/install.md
danielcbright: the `1234` is an example port number, it can really be anything, with some Cloud Providers, the port number is 443, or 80, or 92xx or something similar

danielcbright:
I've never used the suggested edit commit before, so apparently those two that I merged didn't have the DCO, how can I go back and add my Signed-off-by to those commits?

susanev:
@danielcbright in the future you can add `Signed-off-by: susanev <susan.ra.evans@gmail.com>` (but with whoever suggested info) directly when you choose to commit that in the ui as the commit message.

as far as dealing with it now, i think ur gonna have to amend the commit from the command line if there's no new changes to commit. but only the last commit really needs one to pass that dco check.

danielcbright:
@susanev @mjingle ok I've rebased/signed my last 2 commits, should be good now! thanks for the help

PR #3451, opened 2020-04-21
Bumps release version of builder to week of 4/20

eeyun:
New builder release is live, updating version pins to bring you the latest

Signed-off-by: Ian Henry <ianjhenry00@gmail.com>


jaym:
looks like we're running a newer version now

PR #3452, opened 2020-04-21
Fix automate-workflow-ctl habitat plan paths

btm:
I'm unsure what changed, but these won't build for me locally without
more specific paths.

### DIFF
--- a/components/automate-workflow-ctl/habitat/plan.sh
+++ b/components/automate-workflow-ctl/habitat/plan.sh
@@ -52,10 +52,10 @@ do_build() {
 }
 
 do_install() {
-  cp -R lib/* "$pkg_prefix/lib"
-  install -m 0755 "bin/${CTL_BIN_NAME}" "$pkg_prefix/bin/${CTL_BIN_NAME}"
-  install -m 0644 Gemfile.lock "$pkg_prefix/Gemfile.lock"
-  install -m 0644 Gemfile "$pkg_prefix/Gemfile"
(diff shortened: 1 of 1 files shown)


btm:
Later instead of the pattern of `automate/components/automate-workflow-ctl$ hab pkg build habitat` I started using `automate$ hab studio enter` followed by `cd automate/components/automate-workflow-ctl ; build` and that made the difference for some packages, but this doesn't seem wrong.

stevendanna:
My guess is `hab pkg build ./` would have worked there.

> I wonder if using the HAB_CACHE_SRC_PATH like that is safe. It could have stale files, if we keep copying into it. Not sure if our builds are sufficiently isolated (i.e. their HAB_CACHE_SRC_PATH is nuked)...

I'm not sure I am understanding your concern here.  Based on my reading of this change, if this is a problem then it is a problem with the existing code as well.

stevendanna:
My guess is `hab pkg build ./` would have worked there.

> I wonder if using the HAB_CACHE_SRC_PATH like that is safe. It could have stale files, if we keep copying into it. Not sure if our builds are sufficiently isolated (i.e. their HAB_CACHE_SRC_PATH is nuked)...

I'm not sure I am understanding your concern here.  Based on my reading of this change, if this is a problem then it is a problem with the existing code as well.

srenatus:
> Based on my reading of this change, if this is a problem then it is a problem with the existing code as well.

Yup, my thought wasn't really related to this change, sorry.

PR #3457, opened 2020-04-22
[automate-3060] Remove all unnecessary v2 references from authz-service

tylercloke:

This removes all unnecessary v2 references from authz *except for the interservice GRPC API*. Wasn't sure how we wanted to handle that and we can do it in a follow up PR if necessary.

The commit trail is pretty descriptive for what actually changed in this PR so please read the messages. There should be zero functional changes in this PR (mostly renames and some restructuring).  

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/authz-service/server/subject_purge_test.go
```diff
@@ -39,7 +39,7 @@ func TestCommonSubjectPurgeService(t *testing.T) {
 
 	properties.Property("calls both with subject, returns output", prop.ForAll(
 		func(sub string, v2Pols []string) bool {
-			s, err := v2.NewSubjectPurgeServer(ctx, l,
+			s, err := server.NewSubjectPurgeServer(ctx, l,
```
bcmdarroch, line 42: we don't need this as a separate server any longer right? let's get a card for merging it into the policy server

components/authz-service/testhelpers/testhelpers.go
```diff
@@ -99,7 +97,7 @@ func SetupProjectsAndRulesWithDB(t *testing.T) (
 	// TODO(sr): refactor our constructors. Having to maintain the middleware in
 	// three places is tedious and error-prone.
 	serv := connFactory.NewServer(grpc.UnaryInterceptor(
-		grpc_server.InputValidationInterceptor(),
+		server.InputValidationInterceptor(),
```
bcmdarroch, line 100: is this in the same package now?

components/authz-service/storage/postgres/policy_test.go
```diff
@@ -0,0 +1,2146 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/chef/automate/components/authz-service/constants"
+	"github.com/chef/automate/components/authz-service/storage"
+	"github.com/chef/automate/components/authz-service/testhelpers"
+	"github.com/chef/automate/lib/projectassignment"
+)
+
+func TestGetPolicy(t *testing.T) {
```
bcmdarroch, line 18: :happycry:

components/authz-service/server/server.go
```diff
@@ -122,16 +121,16 @@ func NewGRPCServer(ctx context.Context,
...
 
 	// register all services
 	health.RegisterHealthServer(g, health.NewService())
-	api_v2.RegisterPoliciesServer(g, v2PolServer)
-	api_v2.RegisterProjectsServer(g, v2ProjectsServer)
-	api_v2.RegisterAuthorizationServer(g, v2AuthzServer)
+	api_v2.RegisterPoliciesServer(g, polServer)
+	api_v2.RegisterProjectsServer(g, projectsServer)
+	api_v2.RegisterAuthorizationServer(g, authzServer)
 	common.RegisterSubjectPurgeServer(g, subjectPurgeServer)
```
srenatus, line 134: I'd second @bcmdarroch's coment about eventually merging the subject purge stuff into `PoliciesServer`

components/authz-service/server/subject_purge_test.go
tylercloke: Trueeeee! Good catch.

components/authz-service/testhelpers/testhelpers.go
tylercloke: Yeah `server/v2` was flattened into `server` which is is where our `InputValidationInterceptor` lived. It previously had this interesting input label:

```
grpc_server "github.com/chef/automate/components/authz-service/server"
```

tylercloke:
@msorens 

1,3,4. Will do! 
2. Those data migrations were specific to when v1 vs v2 were a thing, so I don't think changing that readme name makes sense.

components/authz-service/server/server.go
tylercloke: I'll do that in a follow up because getting rid of `common` will involve touching the other services.

PR #3444, opened 2020-04-21
Part of #3332: fix diagnostics left overs -- projects and rules

srenatus:
### :nut_and_bolt: Description: What code changed, and why?

We had a bug or two in our diagnostics code that caused projects and rules to stick around after a successful run. Please see the first commit for the fix. The rest of the PR is fixing other occurrences, and minor issues.

⚠️ This doesn't fix these left-overs:
- Token `This token was generated by the chef-automate diagnostic tool. It has admin level access on the entire Automate API.`
- Policy `admin policy for token 1735392e-b4cf-42eb-bf58-611b17e28241`

They are created using a different code-flow and hence their removal can be tackled separately.

### :chains: Related Resources

- #3332 
- Fixes #3445 

### :+1: Definition of Done

Run diagnostics, see the project **not** left over.

### :athletic_shoe: How to Build and Test the Change

- `rebuild components/automate-cli`
- `chef-automate diagnostics run iam`
- observe existing projects

### :white_check_mark: Checklist

- [ ] I have read the [CONTRIBUTING document](https://github.com/chef/automate/blob/master/CONTRIBUTING.md).
- [ ] Tests added/updated?
- [ ] Docs added/updated?
- [ ] All commits have been signed-off for the [Developer Certification of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

PR #3458, opened 2020-04-22
[automate-3332] clean up admin token generated for diagnostics

bcmdarroch:

The diagnostic test runner generates an admin token so it can exercise the APIs. We haven't been cleaning it up. Now we do!

https://github.com/chef/automate/pull/3444

in studio:
```
build components/automate-cli && build components/automate-deployment/
start_all_services
```
Verify by curl or the UI that there is no token with the name `This token was generated by the chef-automate diagnostic tool. It has admin level access on the entire Automate API.`

Run: `chef-automate diagnostics run`

After the diagnostics finish, there should still be no token with the name `This token was ...`

There should also be no extra policies generated. We now add the generated token to the existing admin policy. When the token's deleted, it's also removed as a member of the policy.

- [x] Tests added/updated?
- [ ] Docs added/updated?


components/automate-cli/pkg/diagnostics/context.go
```diff
@@ -340,9 +343,22 @@ func (c *testContext) adminToken() (string, error) {
...
 
 	return c.Globals.CachedToken, nil
 }
+
+func (c *testContext) CleanupAdminToken() error {
+	_, err := c.DoLBRequest(
+		fmt.Sprintf("/apis/iam/v2/tokens/%s", c.Globals.TokenID),
+		lbrequest.WithMethod("DELETE"),
+	)
+	if err != nil {
```
srenatus, line 359: Will this catch a non-200 response?

srenatus:
While we're at it, would you mind replacing the _name_ of the generated token with something that's not a description? I know it'll be deleted, but still, please? ;)

api/interservice/deployment/automate_deployment.proto
```diff
@@ -124,6 +124,7 @@ message GenerateAdminTokenRequest {
 
 message GenerateAdminTokenResponse {
 	string api_token = 1;
```
msorens, line 126: Could we change this to `token_value` or is that causing a lot of undue work? Every time I will see this I will be wondering, "Wait, there's an ID of a token next, so this is... a whole token? no, so what then...?"

components/automate-deployment/pkg/server/api_token.go
```diff
@@ -73,21 +72,12 @@ func generateAdminToken(ctx context.Context,
...
-		Name: "admin policy for token " + tokenID,
-		Statements: []*authz_v2.Statement{
-			{
-				Effect:    authz_v2.Statement_ALLOW,
-				Resources: []string{"*"},
-				Actions:   []string{"*"},
-				Projects:  []string{"*"},
-			},
-		},
+	_, err = authzV2Client.AddPolicyMembers(ctx, &authz_v2.AddPolicyMembersReq{
```
msorens, line 75: 🎉

api/interservice/deployment/automate_deployment.proto
bcmdarroch: it's only being used in one place as far as I can tell so I can make that change

components/automate-cli/pkg/diagnostics/context.go
bcmdarroch: Not sure 🤔 I'll double check

bcmdarroch: Added some extra handling for non 200/403/500 responses (DoLBRequest only catches 403 and 500s 💡)

components/automate-cli/pkg/diagnostics/context.go
```diff
@@ -334,15 +337,43 @@ func (c *testContext) adminToken() (string, error) {
...
+		return errors.Wrap(err, "Failed to clean up the diagnostics admin token")
+	}
+	defer resp.Body.Close()
+
+	// as long as the delete was successful or the token is not found
+	// we consider the cleanup a success
+	if resp.StatusCode == 200 || resp.StatusCode == 404 {
```
srenatus, line 365: I'm not sure I like the 404 here. Besides being the case for nonexistent tokens, it's also likely what we'd get for a changed URL path.

What's the reason for keeping it? I don't suppose the token is ever not there when this runs, is it?

bcmdarroch: yeah that's a good point. i spose if we lost the admin token at any point during the test run we would run into an error sooner, so we can assume it exists by the time we fall here.

PR #3464, opened 2020-04-22
Don't Fatal log when compliance/ingest cant get authz rules

jaym:
Getting authz rules can fail in normal operation. The services now
handle this more gracefully.

- Expired messages are immediately removed 
- Fetching authz rules now uses a 5 second context instead of the
  context attached to the message being processed. This will prevent
  a backlog from triggering a failure when geting authz rules
- When we fail to get authz rules, we log an error instead of crashing
- If we fail to get the authz rules, we start dropping messages.

### DIFF
--- a/components/compliance-service/ingest/pipeline/processor/reports_project_tagger.go
+++ b/components/compliance-service/ingest/pipeline/processor/reports_project_tagger.go
@@ -2,13 +2,17 @@ package processor
 
 import (
 	"context"
+	"time"
 
 	iam_v2 "github.com/chef/automate/api/interservice/authz/v2"
 	"github.com/chef/automate/components/compliance-service/ingest/pipeline/message"
 	"github.com/chef/automate/components/compliance-service/reporting/relaxting"
 	"github.com/chef/automate/lib/stringutils"
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
+
+const maxDropOnError = 128
 
 // BundleReportProjectTagger - Build a project tagger processor for InSpec reports
 func BundleReportProjectTagger(authzClient iam_v2.ProjectsClient) message.CompliancePipe {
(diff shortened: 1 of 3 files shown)


lancewf:
> Fetching authz rules now uses a 5 second context instead of the
> context attached to the message being processed. This will prevent
> a backlog from triggering a failure when geting authz rules

Oh, I have never thought about this case. So, if a message has been in the queue for over 60 seconds nginx would cancel the context. Then when that context is used to request the authz rules it would fail. Then crashing the ingest-service.

Nice catch.

components/compliance-service/ingest/pipeline/processor/reports_project_tagger.go
```diff
@@ -29,16 +33,32 @@ func reportProjectTagger(in <-chan message.Compliance, authzClient iam_v2.Projec
...
 				logrus.WithFields(logrus.Fields{
 					"message_id": msg.Report.ReportUuid,
-					"bundleSize": bundleSize,
+					"bundleSize": nextBundleSize,
 				}).Debug("BundleProjectTagging - Update Project rules")
-				projectRulesCollection = getProjectRulesFromAuthz(msg.Ctx, authzClient)
+				var err error
+				projectRulesCollection, err = getProjectRulesFromAuthz(authzClient)
+				if err != nil {
+					dropComplianceMessages(in, err, nextNumToDrop-1)
```
lancewf, line 53: This section keeps falling more and more messages until all the authz request works. I don't understand why the number of requests dropped increases. Why don't we drop all the messages that are currently in the queue?

lancewf: We could also loop through all the messages and drop only the ones that have a context that has timed out. Then try the auth request again.

jaym: the idea is that we don't want to drop the entire queue if we see a blip from authz. Instead, we will ease into dropping large amounts of messages. So the first one, we only drop the message that is being processed, then we drop ourselves + another, then ourselves + 2*other, then ourself + 2^n*other.

lancewf: I think the first dropped message will not call `m.FinishProcessingCompliance(err)`, because dropComplianceMessages gets called with numToDrop = 0.

components/compliance-service/ingest/pipeline/processor/reports_project_tagger.go
```diff
@@ -29,16 +33,32 @@ func reportProjectTagger(in <-chan message.Compliance, authzClient iam_v2.Projec
...
+					nextNumToDrop *= 2
+					if nextNumToDrop > maxDropOnError {
+						nextNumToDrop = maxDropOnError
+					}
+					continue
```
lancewf, line 58: I think you need `msq.FinishProcessingCompliance(err)` before this line.

components/compliance-service/ingest/pipeline/processor/reports_project_tagger.go
lancewf: So, you are increasing the number of dropped messages for every time the authz call fails, but why not increase the number of messages drop based on the load of the system? So, if there is one message coming in every 5 seconds then this would drop just one message at a time. If there were 100 messages every 5 seconds it would start to get backed up and then exponentially start dropping more and more. The problem is at first, the queue might fill up when there is a heavy load.

jaym: good catch, i'll add the finish processing call.

As far as shedding load, there are a number of things that can be done. The reason I don't want to drop the entire queue is because i'm not convinced there isn't a problem with making calls to authz. Retrying once may get us back on track. There's another PR out that tries to deal with queues filling up that I'm working on.

PR #3466, opened 2020-04-22
chore: bump OPA to 0.19.1

srenatus:
🔧chore dep bump. Let's see how this works out in CI.

ℹ️ I've added comments to `go.mod`, trying to fix the version of `github.com/golang/protobuf` so it wouldn't be bumped to 1.4.0 (which grpc-gateway can't work with so far, I believe).


go.mod
```diff
@@ -39,37 +38,31 @@ require (
...
 	github.com/go-delve/delve v1.3.1
 	github.com/go-gorp/gorp v2.0.1-0.20180410155428-6032c66e0f5f+incompatible
-	github.com/gobwas/glob v0.2.3 // indirect
 	github.com/gocarina/gocsv v0.0.0-20170928100509-7099e67763c2
 	github.com/gofrs/uuid v2.1.0+incompatible
 	github.com/golang-migrate/migrate v3.5.4+incompatible
 	github.com/golang/mock v1.4.3
-	github.com/golang/protobuf v1.3.5
-	github.com/gorilla/context v1.1.1 // indirect
+	github.com/golang/protobuf v1.4.0
```
srenatus, line 45: And still, here it is :(

srenatus: No wait, it's **not** updated in `vendor/`. So this could have worked. I'm still a little uncertain about all the other bumps this did, though.

srenatus:
Hmm shall we just merge it? 🤔

PR #3468, opened 2020-04-22
api docs: fix examples to use correct quoting

vjeffrey:

fix the quoting on the examples so that we use single quotes on outside and double inside


vjeffrey:
Screen Shot 2020-04-22 at 11 01 12

vjeffrey:
gonna try to force push again to get past the expeditor hang

Issue #3401, opened 2020-04-17
Fix API docs URL paths


srenatus:
Previously, our proto files' http annotations had **not** been carrying the complete path -- they were omitting the prefix (`/api/v0`). Additonally, introducing a versioning scheme that hasn't caught on, all new IAM endpoints are under `/apis/iam/v2`, which also doesn't show in the proto files. As a result, our API docs URL paths are wrong.

The work spiked in #3419 showed that it's a better idea to put the *correct path* into the proto files, so we don't have to worry about mismatches.

To finish this, we need to
- [X] adapt the IAM protos
- [X] adapt the IAM introspection requests
- [x] adapt the other protos
- [x] adapt their introspection requests
- [x] fix routing
- [ ] deal with the test fallout 

For more notes, please see #3474.

<details><summary>previous issue body</summary>

The IAM endpoints use a versioning scheme that hasn't caught on. As a result, their API prefix (`/apis/iam/...`) doesn't appear properly in the generated API docs.

To keep the present prefix (for compatibility), but also to make the docs **correct**, we could investigate adding another set of endpoints, either

1. by fiddling with the serve mux in automate-gateway, or
2. by adding a rewrite to automate-load-balancer.

Then, we need to ask ourselves how we want to test the added endpoints. Do we want to treat both sets of endpoints as proper ones; or keep our tests as is (testing `/apis/iam`), or rewrite them to test `/api/v0` etc.
</details>

susanev:
as a note, this will only fix part of the path that is broken in the api docs
currently, all the api docs use `https://automate.chef.io/api` with no version

susanev:
this should probably not be linked to #3418 because that merge will not fix this issue

srenatus:
I think the work started (or drafted) in #3419 is the right way to go. :)

PR #3474, opened 2020-04-23
Fix API docs' paths

srenatus:

To make the API docs match the paths used in the system, this change makes all proto files use their complete paths, i.e., `/api/v0/foo/bar`, not `/foo/bar`. The rest of the PR is fixing the fallout.

#### Notes

- [x] authz introspection makes use of the proto's paths, and they'll change now
  - there's two ways to fix this: massage the paths, or make it work with the fixed paths
  - earlier, we have made exceptions from API compatibility for introspection, and I believe it's warranted to ignore it here, too; if working with the fixed paths is simpler in any way (which I think it'll be)
- [x] for the other paths, `/api/v0`, we'll need to pick either `/api/v0` or `/api/v1`, as the primary path -- the one used in the proto annotations that we process
  - there's an additional path setting, I think, I'll try to see if we can leverage that, or if it hinders doc generation -- **update** I've decided against that, since updating the protos programmatically (read: something better than sed) is blocked. So, I'll make the proto paths say `/api/v0` and use the load-balancer or the gateway to massage `/api/v1`s to `/api/v0`.
  - [x] for introspection, only the primary path will matter, so maybe we'll have to rewrite some paths in the process (UI or backend) -- **still needs fixing**
- [x] fix remaining test fallout
- [x] cleanup the git history, this should not contain changes in the automate-chef-io api docs data directory

- #3419 👈 this incorporates that work (rebased!)
- #3401 👈 this is the issue

### :+1: Definition of Done

- our docs have the correct API docs
- introspection from the browser still works

- To see that the UI still behaves, rebuild:
  - automate-load-balancer
  - automate-gateway
  - automate-ui

To view the docs:
  - go to `components/automate-chef-io`
  - run `make sync_swagger_files` to regenerate needed files
  - run `make serve` to start the docs server
  - browse to `http://localhost:1313/docs/api`

For the public API, nothing has changed. Note the passing tests 😉 

- [x] Tests added/updated?
- [x] Docs added/updated?


srenatus:
My intention is to first get this into a working state with only the iam paths fixed, and then address the rest.

components/automate-gateway/api/iam/v2/policy.proto
```diff
@@ -63,7 +63,7 @@ service Policies {
 			}
 		};
     option (google.api.http) = {
-      post: "/iam/v2/policies"
+      post: "/apis/apis/iam/v2/policies"
```
srenatus, line 66: 🤦‍♂️

srenatus:
⚠️ I'm marking this RFR, but it's not. I just want to see the conflicts vs master.

srenatus:
Starting the week with a rebase...

.studio/automate-gateway
```diff
@@ -23,7 +23,7 @@ function gateway_integration() {
   # the relevant services up and running:
```
msorens, line 23: In .studio/license-control-service, need the same fix on this line:
https://github.com/chef/automate/blob/cba671834c2277e6dd978b957552c48da009cb75/.studio/license-control-service#L23

srenatus:
@susanev the docs thing is a faux pas in the reset commit, I should have not included the static stuff. Will fix in a bit.

I'll look into the other bugs next, thanks!

components/automate-ui/src/app/entities/userperms/README.md
```diff
@@ -45,8 +45,8 @@ Example:
 ```json
 {
   "endpoints": {
-    "/iam/v2/tokens": { "delete": false, "get": true, "post": true, "put": false },
-    "/iam/v2/policies": { "delete": false, "get": true, "post": true, "put": false },
+    "/apis/iam/v2/tokens": { "delete": false, "get": true, "post": true, "put": false },
+    "/apis/iam/v2/policies": { "delete": false, "get": true, "post": true, "put": false },
     "/cfgmgmt/stats/run_counts": { "delete": false, "get": true, "post": false, "put": false },
     "/cfgmgmt/suggestions": { "delete": false, "get": true, "post": false, "put": false },
     "/cfgmgmt/version": { "delete": false, "get": true, "post": false, "put": false }
```
msorens, line 52: Also need these:
```suggestion
    "/api/v0/cfgmgmt/stats/run_counts": { "delete": false, "get": true, "post": false, "put": false },
    "/api/v0/cfgmgmt/suggestions": { "delete": false, "get": true, "post": false, "put": false },
    "/api/v0/cfgmgmt/version": { "delete": false, "get": true, "post": false, "put": false }
```

susanev:
im still seeing issues with the api doc paths
doc-path-2

doc-path-1

srenatus:
@susanev I can't seem to reproduce it -- when doing the docs steps that @msorens has added to the PR preamble:

image

image

Maybe the two of you could figure this out together? Otherwise, I'd be of course happy to look into this with you tomorrow (your) morning. :)

susanev:
oh yah sorry, i havent tried again since yall let me know i had to do those steps. will try again later.

vjeffrey:
im having trouble reaching an "available" profile's details -- it redirects me to event feed instead. this is the type of url it should be trying to access: https://a2-local-inplace-upgrade-dev.cd.chef.co/compliance/compliance-profiles/profile-details?name=cis-aix-5.3-6.1-level2&version=1.1.0-5

susanev:
these ones are missing the api/v0
node-export
compliance-export

PR #3475, opened 2020-04-23
tools: add prettifier for protos, run it

srenatus:
This is preliminary work to programmatically update all our proto files' API paths.

### Open questions
- Why is it eating all comments? It shouldn't.
- Do we have a preferred indentation?
- Does anyone have strong objections to this?

👉Available printer options can be found here: https://pkg.go.dev/github.com/jhump/protoreflect/desc/protoprint?tab=doc#Printer


api/external/cfgmgmt/cfgmgmt.proto
```diff
@@ -1,353 +1,282 @@
...
-  Authorization Action:
-  ```
-  infra:nodes:list
-  ```
-  */
-  rpc GetErrors(cfgmgmt.request.Errors) returns (cfgmgmt.response.Errors) {
-    option (google.api.http).get = "/cfgmgmt/errors";
-    option (chef.automate.api.iam.policy).resource = "infra:nodes";
-    option (chef.automate.api.iam.policy).action = "infra:nodes:list";
+  // GetNodes
```
stevendanna, line 22: ⚠️ Our documentation processing scripts appear to depend on the using the `/*` commenting style, so this comment style change is actually a bit of a breaking change.  There is an option in the pretty-printer to prefer the `/*` style, but I fear that might then pull in comments we didn't intend to publish into the docs.

srenatus:
💥 As @ssd noticed, this work is blocked by the way #2982 was done: it depends on the way comments are used. Arguably, syncing comments from annotations is something that would best be done in a similar to manner to this PR's tool -- parse the proto, use the rich descriptors to fix stuff, (pretty) print them.

**But**, this has already been going on for long enough, so I'll leave it at that, to revisit at a later point. 😞

PR #3477, opened 2020-04-23
Apply backpressure for ingest-service ingestion

jaym:
The ways backpressure is applied:
- If the queue is full, the message is immediately rejected
- If the client goes away, the message is dropped from the pipeline

We can also configure the buffer size. By default, its configured to 200. It can changed with:

```toml
[ingest.v1.sys.service]
message_buffer_size = 200
```


components/ingest-service/cmd/ingest-service/commands/serve.go
```diff
@@ -91,8 +91,8 @@ func readCliParams() *serveropts.Opts {
 		PurgeActionsAfterDays:         int32(viper.GetInt("actions-days")),
 		ChefIngestServerConfig: serveropts.ChefIngestServerConfig{
 			MaxNumberOfBundledActionMsgs: viper.GetInt("max-number-of-bundled-action-msgs"),
+			MessageBufferSize:            viper.GetInt("message-buffer-size"),
 			ChefIngestRunPipelineConfig: serveropts.ChefIngestRunPipelineConfig{
-				MaxNumberOfBundledMsgs:   viper.GetInt("max-number-of-bundled-run-msgs"),
```
jaym, line 95: shouldn't be deleted

components/ingest-service/pipeline/message/chef_run.go
```diff
@@ -14,6 +14,14 @@ import (
 
 type ChefRunPipe func(<-chan ChefRun) <-chan ChefRun
 
+func PropogateChefRun(out chan<- ChefRun, msg *ChefRun) {
```
stevendanna, line 17: We don't need to do it in this PR, but this might be a nice place for a small interface for Messages so we can share some functions across these pipelines.

jaym: i tried this last week and failed. Gave up and decided it wouldn't be possible without generics. but if you have ideas, please share

stevendanna: I meant something like this: https://github.com/chef/automate/commit/02d5ccb372e142ec9ff9682b005ca99e34243b25#diff-7056bfe5063a802fbe9dc89de9191f26R5-R8. But I admittedly didn't look at it past a surface level.  Was there something specific you hit?

jaym: i think thats what i was trying to do. but you cant have a function that takes a `chan Finishable` and give it a `chan ChefRun`.

stevendanna: The cure is worse than the disease I think. Carry on.

PR #3491, opened 2020-04-24
Wrap notifications to pin at top

SEAjamieD:
This branch rebuilds the notification component out of stencil and into automate as an angular component.  Using an angular component allowed me to animate it as well. 

The ribbons retain their same look, feel, and containment, but now they animate in and out of the page.  There is also a small UX patch that sticks them to the top of their respective window when scrolling.

### :+1: Definition of Done
notifications for things like "new api token created" are visible even when the user is deep down the list.

tests are added that test against behavior.

make serve

visit various pieces of automate that have the ability to cause a notification
example: 
visit: https://a2-dev.test/settings/projects
create several projects so that you have a decent list
scroll down deep - delete a project - view the notification stuck on top of the scrollable window

- [x] Tests added/updated?
- [ ] Docs added/updated?

Chef Automate (2)


SEAjamieD:
@susanev - Figured I'd pause here til you have a look.  The animations are super duper easy to change now so can switch up or remove or do whatever

SEAjamieD:
weird - my updates in the explanation about the PR got reverted...fixing now

SEAjamieD:
@susanev We can really do any kind of animation you'd like now!  I'll put something like you're talking about together and then we can link up and fine tune it.

>if a second success notification happens before the first one has dismissed, could we dismiss the first one then?

I'll make a second ticket for this.  While working on this ticket I noticed that we have different ways of showing notifications across automate.

There is a main notification center, which retains a spot in the UI across all of automate.  Then in several other files, we have notification areas that and are only present in their respective files, and the notification is triggered in that file and not through the global notification center.  I think the latter files should be rewired first so that all are using the the global notification center.  It looks to me like the latter was how Automate did things in the past, and they still need to be updated.

SEAjamieD:
@susan - altered the animation, but please feel free to pick apart, change, whatever you want, all of it is pretty easily configurable.

components/automate-ui/src/app/components/notification/notification.component.ts
```diff
@@ -0,0 +1,52 @@
...
+      style({
+        opacity: 1,
+        transform: 'translateY(0%) scale(1)'
+      }))
+  ]),
+  transition(':leave', [
+    style({ opacity: 1 }),
+    animate('195ms ease-in', style({ opacity: 0 }))
+  ])
+]);
```
SEAjamieD, line 21: This piece is controlling the animations of the notification ribbons.  Each of these properties are cheap, using translate and scale, neither of which cause a reflow.

components/automate-ui/src/app/components/notifications/notifications.component.html
```diff
@@ -1,9 +1,12 @@
-<div class="notifications">
-  <ng-container *ngFor="let notification of notifications">
-    <chef-notification *ngIf="notification.type !== 'license'"
-      [type]="notification.type"
-      [timeout]="notification.timeout" (dismissed)="handleNotificationDismissal(notification.id)">
-      {{ notification.message }}
-    </chef-notification>
+<div class="notifications"> 
+  <!-- @todo - custom reverse pipe -->
+  <ng-container *ngFor="let notification of notifications.slice().reverse()">
```
SEAjamieD, line 3: @susanev - for UX, i reversed the list because I thought it looked better having items added to the bottom of the list, but I don't know if you have newest at the top for a reason.

for devs - If UX keeps the list reversed, I'll create a custom pipe for this.  If not...then theres no pipe to be made.

susanev: yah latest should be at the top, with others shifting down.

components/automate-ui/src/styles.scss
```diff
@@ -17,6 +17,12 @@
   }
 }
 
+.sticky-notifications {
+  position: sticky;
+  top: 0;
+  z-index: 199;
+}
+
```
SEAjamieD, line 25: This keeps all notifications _except_ the license notification pinned to the top on scroll.

e2e/cypress/integration/ui/settings/team_mgmt.spec.ts
```diff
@@ -105,8 +105,7 @@ describe('team management', () => {
 
       cy.get('[data-cy=save-button]').click();
       cy.get('app-team-management chef-modal').should('not.be.visible');
-      cy.get('#main-content-wrapper').scrollTo('top');
```
SEAjamieD, line 108: now that notifications do not push content down, many of these tests were failing because the page is no longer long enough to be scrolled.  Several of these `scrollTo` have been removed to remedy the problem.

components/automate-ui/src/app/components/notification/notification.component.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+        expect(component.isInfinite).toBe(isInfinite);
+      });
+    });
+
+    using([
+      [Type.info, 1, false],
+      [Type.info, 2, false],
+      [Type.error, 3, false]
+    ], function (type: Type, timeout: number, isInfinite: boolean) {
+      it('notification should close when infinite', fakeAsync(() => {
```
msorens, line 93: This comment applies to all your data-driven tests here:
Remember that you need to make the test names unique. That simply means one or more parameters from each test case needs to be embedded in the test name string. Often, that involves adding another test parameter that I typically call `description` that describes what the test does.

(Why? When a test fails, it will say _"notification should close when infinite" failed_. So which of the three tests here failed? Second, some test runners generate a report of the tests run; seeing 8 occurrences of undifferentiated test "xyz" in the results list is not very useful.)

components/automate-ui/src/app/components/notification/notification.component.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+        flush();
+      }));
+    });
+
+    using([
+      [Type.info, 0, true],
+      [Type.info, 0, true],
+      [Type.error, 0, true]
+    ], function (type: Type, timeout: number, isInfinite: boolean) {
+      it('notification does not close when infinite', fakeAsync(() => {
```
msorens, line 112: Also, on test naming: consider this test and the prior test together. As they exist now:
```
notification should close when infinite
notification does not close when infinite
```
Pretty sure one of those is wrong. 😁 

Also, in a data-driven test you should not have any parameters that are invariant, so both `timeout` and `isInfinite` should be local variables, not test parameters.

components/automate-ui/src/app/components/notification/notification.component.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+        expect(component.handleClose).toHaveBeenCalled();
+
+        fixture.destroy();
+        flush();
+      }));
+    });
+
+    using([
+      [Type.info, 0, true],
+      [Type.info, 0, true],
```
msorens, line 109: Duplicate.

components/automate-ui/src/app/components/notification/notification.component.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+      [Type.info, 2, false],
+      [Type.error, 3, false]
+    ], function (type: Type, timeout: number, isInfinite: boolean) {
+      it('notification should close when infinite', fakeAsync(() => {
+        spyOn(component, 'handleClose');
+        genNotification(type, timeout);
+        component.ngAfterViewInit();
+
+        expect(component.isInfinite).toBe(isInfinite);
+        tick((timeout * 1000) + 1);
```
msorens, line 99: I am not well-versed on the `fakeAsync`, but I believe these `tick` calls are using real-time, which is a major red flag. That slows down the unit tests by a huge factor; any given test file should run in milliseconds, not seconds. (Sometimes time-related testing is where you do need to test implementation. 😁 ) Found one article that (rather teresely) indicates that `flush` is sufficient and `tick` may not be needed: https://alligator.io/angular/testing-async-fakeasync/

components/automate-ui/src/app/modules/license/license-notifications/license-notifications.component.html
```diff
@@ -1,13 +1,14 @@
 <ng-container *ngFor="let notification of notifications">
-  <chef-notification
+  <app-chef-notification
```
msorens, line 2: I think it is more consistent to use this:
```suggestion
  <app-notification
```
... because the wrapper is `<app-notifications>`, not `<app-chef-notifications>`

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
```diff
@@ -1,9 +1,11 @@
-<chef-notification
-  *ngIf="(serviceGroupsStatus$ | async) === 'loadingFailure'"
-  [type]="'error'"
-  [timeout]="10">
-  {{ (serviceGroupsError$ | async).status }}: {{ (serviceGroupsError$ | async).error.message }}
-</chef-notification>
+<div class="sticky-notifications">
```
msorens, line 1: Seems like you are using this wrapper a lot... almost feels like that should be a parameter of `<app-chef-notification>` instead.

Upon further reading... you have a comment on the CSS indicating that keeps everyone except the license notification sticky... which immediately suggests to invert that. That is, the default behavior should be sticky; it is the exception where you should have to do something explicit (and thereby making maintenance much less fragile).

components/automate-ui/src/app/pages/+compliance/+profile/+profile-details/profile-details.component.ts
```diff
@@ -25,6 +25,9 @@ export class ProfileDetailsComponent implements OnInit, OnDestroy {
   deleteErrorVisible = false;
   downloadErrorVisible = false;
   installErrorVisible = false;
+  downloadFailureNotificationContent = 'Failed to download profile.Please try again later.';
+  deleteFailureNotificationContent = 'Failed to delete profile. Please try again later.';
+  installFailureNotificationContent = 'Failed to install profile. Please try again later.';
```
msorens, line 30: I do not see the reason for extracting these into locals. They are referenced in exactly one spot, the template, and better to have literal text in the template itself unless there is a reason not to.
(Same comment applies in other files.)

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
SEAjamieD: Having the sticky part contained would be ideal, though not accomplishable in the current state.  

The notification component is only responsible for the component itself.  It makes no distinction as to where it is placed on the page and is only a single ribbon.  On top of that, we want the ribbon to lay over content, and not take up space.  This makes its sticky positioning the responsibility of the container holding it, which sometimes holds multiple notifications at once.

The best way I can think of (for now) to maintain some consistency, is have the sticky container be a global style.  I'll add a note for its use.  In the near future, I hope to refactor our app notifications to use the global notifications container thats already in place and used by about half the notifications. but that will have to be a larger ticket for another time.

I'm thinking my comment wasn't clear enough about the license notification.  The clearer reason I wrote that note, is that i'm actually just _not using_ the sticky class on the license notification container (as opposed to the class not working the same way), so that it doesn't cover any of the top nav   Though UX may change that.

SEAjamieD: double checking and highlighting this led me to a different small issue.  Glad you mentioned it!

components/automate-ui/src/app/components/notification/notification.component.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+    using([
+      [Type.info, 10],
+      [Type.info, 7],
+      [Type.info, 4],
+      [Type.info, 2],
+      [Type.info, 0],
+      [Type.error, 43],
+      [Type.error, 13],
+      [Type.error, 3],
+      [Type.error, 23]
```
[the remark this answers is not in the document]
SEAjamieD: Thanks for linking that article.  Those two pieces alone provide a better way for me to think about what to toss in the test funnel.

components/automate-ui/src/app/components/notification/notification.component.spec.ts
SEAjamieD: This was a great quick read -thank you!

components/automate-ui/src/app/components/notification/notification.component.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+  describe('After a notification is visible', () => {
+
+    using([
+      ['is not infinite', Type.info, 10, false],
+      ['is infinite', Type.info, 0, true],
+      ['is not infinite', Type.error, 43, false],
+      ['is infinite', Type.error, 0, true],
+      ['is infinite', Type.error, -1, true]
+    ], function (description: string, type: Type, timeLimit: number, expected: boolean) {
+      it(`${description} when type is ${type} and timeout is ${timeLimit}`, () => {
```
msorens, line 73: Was not quite sure how to interpret what this test is testing--here I just plugged in a set of values:
```
After a notification is visible,
is not infinite when type is Type.info and timeout is 10
```
Seems like you might mean:
```
After a notification appears,
it is scheduled to close in 10 seconds when type is Type.info
```
or
```
After a notification appears,
it remains permanently open when type is Type.info and timeout is -1
```
but those seem to be sufficiently covered in the next 2 sets of tests. 🤷 

(And minor point, but making `this.infinite` public for the sole reason that a test needs it is a little red flag, too.)

SEAjamieD: Good observation - I'll clean that up.

Yeah - I was a little unsure on what to do regarding `this.isinfinite`.  I read that If a variable or function are private, its the responsibility of the test of the function that contains it - but felt safer to test it specifically.  I can pull that out.

SEAjamieD:
Applying the do not merge tag so that UX can review.  There's no rush to get this in.

components/automate-ui/src/app/components/notification/notification.component.spec.ts
msorens: BTW, if removing the test, then `isinfinite` can be restricted down to a local variable within the function that is using it.

PR #3496, opened 2020-04-24
Document that profiles need a particular version format to be valid

sean-horn:
As hinted at by https://github.com/chef/compliance-profiles#versioning-of-profiles,
the versioning of Inspec profiles is very particular. Outside of the expected semver format, 
A2 will not accept the profile upload completely and so cannot serve the profile version

Signed-off-by: Sean Horn <sean_horn@chef.io>

### DIFF
--- a/components/automate-chef-io/content/docs/profiles.md
+++ b/components/automate-chef-io/content/docs/profiles.md
@@ -43,6 +43,10 @@ Uploads use either the `.tar.gz` or zip archive file formats.
 
 All profiles are stored in PostgreSQL, and are covered by backup and restore functionality.
 
+{{% info %}}
+All profiles should have a valid version in their inspec.yml. Valid versions consist of at least three elements of the form x.y.z, for example "1.2.0", and optionally can have one more element, like "1.2.0-20". Profile versions with two elements, for example "1.2" will silently fail to install when uploaded through the UI, and will be visible in the UI, but cannot be retrieved by Inspec runs.
+{{% /info %}}
+
 ### Updating profiles
 
 New releases of profiles are shipped with the product when available.


components/automate-chef-io/content/docs/profiles.md
```diff
@@ -43,6 +43,10 @@ Uploads use either the `.tar.gz` or zip archive file formats.
 
 All profiles are stored in PostgreSQL, and are covered by backup and restore functionality.
 
+{{% info %}}
+All profiles should have a valid version in their inspec.yml. Valid versions consist of at least three elements of the form x.y.z, for example "1.2.0", and optionally can have one more element, like "1.2.0-20". Profile versions with two elements, for example "1.2" will silently fail to install when uploaded through the UI, and will be visible in the UI, but cannot be retrieved by Inspec runs.
```
kagarmoe, line 47: I'm not sure I follow--is this note an instruction to update the `inspec.yml` file or for how to upload the profile in the UI and from the command line?

rwc: @kagarmoe This is an instruction/warning/formatting requirement. The version must be in the format x.y.z versus a.b, or else the profile will not be able to be consumed.

susanev:
@vjeffrey is there any way to fix this so it doesn't fail silently in the ui?

vjeffrey:
> is there any way to fix this so it doesn't fail silently in the ui?

ill take a look today - i would imagine it would fail inspec check but apparently not?  ill play around with it today

susanev:
im gonna add a do-not-merge label to give us some time to figure out if we can fix the root cause instead

vjeffrey:
i uploaded a profile to my dev env that had '2.0' as the version number.
i ran a scan job with that profile, and everything worked.
then i ran audit cb referencing the same profile, against acceptance, and it worked.. https://a2-local-inplace-upgrade-acceptance.cd.chef.co/compliance/reports/profiles

susanev:
@sean-horn @rwc are yalls repro steps different than @vjeffrey's?

sean-horn:
@susanev @vjeffrey I watched it happen, on-site, with a customer. Their audit cookbook could not retrieve the 2 digit versioned profiles. Their A2 release and audit cookbook were probably older. Is there a reason to not document it this way? Its complete, supposed to work for every version and we have customers that will continue to use the old stuff

susanev:
if we can figure out what build it was fixed in, we could add a note that it only works that way for that build and newer. currently, docs standard is to document latest build, @chef/docs-team yall have thoughts on this?

kagarmoe:
Hi! @chef/docs-team can't take on this work until June. Let's turn it into an issue until you're able to reproduce it and we have a solution. Thanks!

kagarmoe:
I can't check out this branch.

susanev:
the issue is already fixed in latest automate. i believe @sean-horn concern is that its not documented for older versions of automate. if we could get some note added that explains that it might meet the needs of users on older versions.

kagarmoe:
The general rule is to document the current release. Do both two and three digit version formats work now?

susanev:
@kagarmoe yes thats correct, in latest versions both formats work.

susanev:
i think the fork was deleted, so imma start a new one

susanev:
now in #4004

PR #3502, opened 2020-04-24
[automate-3498] add validate lib

bcmdarroch:

This PR aims to continue @blakestier's work improving our server validation by moving her validation code from authz-service into a shared library. 

Once we have a shared validation library, we can refactor authn, teams, and local-user services (and hopefully others!) to utilize it in their servers.

https://github.com/chef/automate/pull/3312

### :+1: Definition of Done
- validation works as before, tests pass

- [x] Tests added/updated?
- [ ] Docs added/updated?


components/authz-service/server/policy.go
```diff
@@ -167,7 +168,7 @@ func (s *policyServer) GetPolicy(
 	ctx context.Context,
 	req *api.GetPolicyReq) (*api.Policy, error) {
 
-	err := confirmRequiredID(req.Id, "policy")
+	err := validate.RequiredID(req.Id, "policy")
```
srenatus, line 171: Just a thought -- If we defined this as
```go
type hasID interface {
  GetId() string
}

func RequiredID(x hasID, y string) error {
    // ...
}
```

...we could here call `validate.RequiredID(req, "policy")`. Now, that doens't make much of a difference, but we could then also do the same for ID and name, like

```go
func RequiredIDandName(x interface { hasID, hasName }, y string) error
```

that would let us do `validate.RequiredIDandName(req, "thing")`  instead of `validate.RequiredIDandName(req.Id, req.Name, "thing")`.

What do you think?

We could also try to make us of field masks if we want to go protobuf-all-in. 🤔 

----

Taking things further, we could wire this into a services middleware, and write something like "with everything that has an ID, that ID is required" via
```go
if x, ok:= req.(HasID); ok {
  if err := validate.RequiredID(x, "???"); err != nil {  /* ... */ }
}
```
but we'd not know what to put into `???` easily, I suppose...

lib/validate/validate.go
```diff
@@ -0,0 +1,64 @@
...
+func RequiredFieldsAndProjects(id string, name string, projectIDs []string, resourceName string) error {
+	err := RequiredIDandName(id, name, resourceName)
+	if err != nil {
+		return err
+	}
+	return RequiredProjects(projectIDs)
+}
+
+// RequiredProjects verifies that the projects are not empty and do not include '(unassigned)'
+func RequiredProjects(projects []string) error {
```
blakestier, line 51: this is the only one that seems authz specific rather than extractable. . . 🤔 That's probably fine though

bcmdarroch: I think any service that creates resources with a `projects` field needs this check.

components/authz-service/server/project.go
```diff
@@ -811,35 +811,42 @@ func fromAPIType(t api.ProjectRuleTypes) (storage.RuleType, error) {
 	}
 }
 
-func (s *ProjectState) prepareStorageRule(inID, projectID, name string,
-	inType api.ProjectRuleTypes, inConditions []*api.Condition) (*storage.Rule, error) {
+type apiRule interface {
```
srenatus, line 814: Q: Couldn't we re-use the protobuf-generated type here?

srenatus:
We can also consider using these definitions in _status details_: https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto#L164-L180 see @johanbrandhorst's post here https://jbrandhorst.com/post/grpc-errors/

components/authz-service/server/project.go
bcmdarroch: I wanted a type that fit both the Create and Update request types.

lib/validate/validate.go
blakestier: I guess, but it's not the domain of those services, if that makes sense. You're totally right that there is utility in them checking, but it's not really their responsibility to do so as much as it is authz's

bcmdarroch:
Created this follow-up task: https://github.com/chef/automate/issues/3554

lib/validate/validate.go
bcmdarroch: I see what you mean. So we could make this explicitly authz's responsibility by pushing this validation into the `validateProjects` call to authz.

.studiorc
```diff
@@ -154,7 +154,7 @@ fi
 
 document "start_all_services" <<DOC
   Simple wrapper for 'start_deployment_service' and 'chef-automate dev deployinate'.
-  Also applies a license (if present), creates IAM v2 dev users, and upgrades IAM to v2.
+  Also applies a license (if present) and creates IAM v2 dev users.
```
bcmdarroch, line 157: unrelated fix

msorens: How about delete the "v2" though.

components/authz-service/server/policy.go
```diff
@@ -762,3 +764,32 @@ func (s *policyServer) logPolicies(policies []*storage.Policy) {
 	}
 	s.log.WithFields(kv).Info("Policy definition")
 }
+
+// requiredFieldsAndProjects verifies that all the required inputs are not empty
+func requiredFieldsAndProjects(obj interface {
```
msorens, line 769: Projects is just a field, too, right? Could this just be `requiredFields` ?

components/authz-service/server/policy.go
```diff
@@ -762,3 +764,32 @@ func (s *policyServer) logPolicies(policies []*storage.Policy) {
...
+// requiredProjects verifies that the projects are not empty and do not include '(unassigned)'
+func requiredProjects(projects []string) error {
+	for _, project := range projects {
+		if validate.EmptyOrWhitespaceOnlyRE.MatchString(project) {
+			e := fmt.Sprintf("projects must contain at least one non-whitespace character")
+			return errors.New(e)
+		}
+		if project == constants.UnassignedProjectID {
+			return errors.Errorf("%q cannot explicitly be set. "+
+				"If you wish to create an object in %q, you should pass no projects on creation.",
```
msorens, line 790: "create an object in (unassigned)" sounds awkward. Perhaps:
"(unassigned) cannot explicitly be set; supply an empty projects array instead."

components/authz-service/server/policy.go
```diff
@@ -762,3 +764,32 @@ func (s *policyServer) logPolicies(policies []*storage.Policy) {
...
+}, resourceName string) error {
+	err := validate.RequiredIDandName(obj, resourceName)
+	if err != nil {
+		return err
+	}
+	return requiredProjects(obj.GetProjects())
+}
+
+// requiredProjects verifies that the projects are not empty and do not include '(unassigned)'
+func requiredProjects(projects []string) error {
```
msorens, line 782: So I saw the conversation about moving this out of the lib into authz... then how to teams and tokens do project verification?

bcmdarroch: Right now they're covered by this validation: https://github.com/chef/automate/blob/master/api/interservice/authz/v2/authz.proto#L94-L98

That works for now. But for consistency, we'll need to add this block of project validation into the ValidateProjectAssignment server code as well.

blakestier: We don't need to move it-- there's a case that this is the "authz validation lib" and it's meant to be used to validate yer things that use projects 🤷 . I was just thinking out loud 😬

components/authz-service/server/project.go
```diff
@@ -405,7 +406,7 @@ func (s *ProjectState) ListRulesForAllProjects(ctx context.Context,
 }
 
 func (s *ProjectState) CreateRule(ctx context.Context, req *api.CreateRuleReq) (*api.CreateRuleResp, error) {
-	r, err := s.prepareStorageRule(req.Id, req.ProjectId, req.Name, req.Type, req.Conditions)
+	r, err := s.prepareStorageRule(req)
```
blakestier, line 409: tenor-197748060

reducing those params 🎉

PR #3511, opened 2020-04-27
[Automate-3385] Policy dropdown for tokens

msorens:

This PR provides a valuable exercise in applying [SOLID](https://en.wikipedia.org/wiki/SOLID) principles.

As shown, the ProjectsDropdownComponent is used directly in 3 places: CreateObjectModalComponent, TeamDetailsComponent, and ApiTokenDetailsComponent,
while the CreateObjectModal in turn is used in 3 places: TeamManagement, ApiTokenList, and ProjectList. The newly introduced pieces are in blue.

image

#### Single Responsibility Principle (SRP)

Previously, each of those components did some of the work that should have been encapsulated within the ProjectsDropdownComponent itself, namely:
(a) Generating the list of assignable projects and passing it through to the ProjectsDropdownComponent.  However, that list just came from the ngrx/store, so it is just as accessible to the ProjectsDropdownComponent to do itself, with the added benefit of (1) encapsulating that responsibility where it should be and (2) eliminating multiple occurrences of repeated code.
(b) Enabling and disabling the ProjectsDropdownComponent, based on the list of assignable projects. It is the ProjectsDropdownComponent itself that knows that list, so it can manage that itself.
(c) Reacting to each check or uncheck action within the dropdown. Clearly, that is the responsibility of the ProjectsDropdownComponent, not the calling components; they only need to know when the user has finished within the dropdown, identified by when the dropdown closes.

The above led to eliminating three of the four ProjectsDropdownComponent parameters and adding two new ones (with the net result of one less parameter total).

#### Open-Closed Principle (OCP)

The original ProjectsDropdownComponent could have been modified to handle policies as well as projects by tweaking the code with some conditionals and a few extra functions.  But a better approach was to genericize the component so it does not care if it is handling policies or projects.  This new component (in blue) is ResourceDropdownComponent. ProjectsDropdownComponent, which knows about projects, has now become a subclass of ResourceDropdownComponent. A second subclass now knows about policies. In most languages, subclassing always deals with first-class objects. But Angular is unique in that you could do that (as with ProjectsDropdownComponent), but you can also do "inline" subclasses in a template. It made sense to do that because (a) it is only needed once, inside CreateObjectModal, and (b) it does not require any custom code, just setting parameters.  In the diagram, you will observe the two subclasses indicated by the dotted lines.

#### Liskov substitution principle (LSP)

To make the above sub-classing work, ResourceDropdownComponent needed to be able to handle a common data type. This relied on the fact that even though policies and projects are quite different objects, they share the commonalities needed: `name` (for display) and `id` (for processing), plus the extension required for the dropdowns, the `checked` property.  So the ProjectsDropdownComponent manages a list of projects (extended with `checked`) while the CreateObjectModalComponent manages a list of policies (also extended with the `checked` property).  Thanks to covariance, we can have a parameter `resources` (an array of `ResourceChecked` in ResourceDropdownComponent) that is a receptacle for either `ProjectChecked` or `PolicyChecked` inputs provided by its subclasses. That is, it is OK to have more properties than `ResourceChecked` requires; those just sit on the sidelines while we work with the required properties.

To be done:
1. Permissions around the dropdown.
2. Organizing by chef managed and custom within the dropdown.
3. UX work: need to notify user that policies, unlike projects, are not part of the token, and you cannot go into the details to edit them or even see them.

NA

### :+1: Definition of Done
The dialog opens upon selecting <kbd>Create Token</kbd> on Tokens page; teams and projects dialogs (using the same modal) are unchanged.

image

(1) Opening the modal shows all available policies (currently sorted by name; that will be enhanced in a follow-up).
(2) Entering a string that occurs in some policy names, filters the list to that set of names.
(3) Entering a string that does not occur shows the message indicated.
image

Selecting one policy mirrors that policy name in the dropdown label; selecting more than one indicates that number. 
image

Upon selecting <kbd>Create Token</kbd> from the modal, notifications occur for creating the token plus for each amended policy.
image
(Note that I modified the AddMember notification to include the policy id; even with just a single policy, it enhances the meaning.)

Finally, note that if you go in to edit a token or team and change the contents of the projects dropdown, it will enable the Save button. If you then change it back to its original state, it will disable the Save button.

1. Rebuild automate-ui
2. Go to Settings >> Tokens
3. Add a new token to see the **_policy_** dropdown

Important: to confirm that the **_project_** dropdown still works, visit the tokens and teams list pages to create them, plus the tokens and teams details pages to update them.

- [x] Tests added/updated?
- [ ] Docs added/updated?


components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
```diff
@@ -49,11 +49,18 @@ <h2 slot="title">Create {{ objectNoun | titlecase }}</h2>
             Also create owner, editor, and viewer policies for this project.
           </chef-checkbox>
         </div>
-        <div class="dropdown-margin" *ngIf="objectNoun != 'project'">
+        <div class="dropdown-margin" *ngIf="objectNoun === 'token'">
+          <app-resource-dropdown
```
msorens, line 53: The new dropdown for policies; it only shows up on the tokens page. (Still need to add permissions guard around it in follow-up PR.)
It is really just this block of HTML and a couple lines of code in the code-behind.

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
```diff
@@ -49,11 +49,18 @@ <h2 slot="title">Create {{ objectNoun | titlecase }}</h2>
...
+          </app-resource-dropdown>
+        </div>
+        <div class="dropdown-margin" *ngIf="objectNoun !== 'project'">
           <app-projects-dropdown
-            [projects]="projects"
-            (onProjectChecked)="onProjectChecked($event)"
-            [disabled]="dropdownDisabled()"
+            (onDropdownClosing)="onProjectDropdownClosing($event)"
             [projectsUpdated]="projectsUpdatedEvent">
           </app-projects-dropdown>
```
msorens, line 65: Simplified footprint of the projects dropdown, as described in the preamble.

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.spec.ts
```diff
@@ -63,22 +69,7 @@ describe('CreateObjectModalComponent', () => {
     });
   });
 
-  it('ngOnChanges initializes checked status to false for all projects ', () => {
```
msorens, line 66: Removed redundant test.

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.spec.ts
```diff
@@ -63,22 +69,7 @@ describe('CreateObjectModalComponent', () => {
...
-      { assignableProjects: new SimpleChange({}, assignableProjects, true) });
-    expect(Object.values(component.projects).length).toEqual(assignableProjects.length);
-    assignableProjects.forEach(p => {
-      const proj = component.projects[p.id];
-      expect(proj).toEqual({ ...p, checked: false });
-    });
-  });
-
-  it('ngOnChanges resets checked status to false for all projects ', () => {
+  it('upon opening, checked status of all projects is set to false', () => {
```
msorens, line 72: Changed existing tests to test behavior instead of implementation.

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
```diff
@@ -22,28 +25,37 @@ export class CreateObjectModalComponent implements OnInit, OnDestroy, OnChanges
...
     .subscribe((isConflict: boolean) => {
       this.conflictError = isConflict;
       // Open the ID input on conflict so user can resolve it.
       this.modifyID = isConflict;
     });
+
+    this.store.select(allPolicies)
+      .pipe(takeUntil(this.isDestroyed))
+      // OK to leave `checked` undefined--will be initialized upon `visible`
+      .subscribe(policies => this.policies = policies as PolicyChecked[]);
```
msorens, line 58: Populate the policies list for the dropdown.

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
```diff
@@ -52,37 +64,35 @@ export class CreateObjectModalComponent implements OnInit, OnDestroy, OnChanges
...
+        this.updatePolicyCheckbox(true); // default to checked upon opening
       }
     }
   }
 
-  onProjectChecked(project: ProjectChecked): void {
-    this.projects[project.id].checked = project.checked;
-    const projectsSelected = Object.keys(this.projects).filter(id => this.projects[id].checked);
+  onProjectDropdownClosing(projectsSelected: string[]): void {
     this.createForm.controls.projects.setValue(projectsSelected);
```
msorens, line 84: Moved micro-managing into the dropdown where it belongs. We only have to set the value once, when the dropdown closes.

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.html
```diff
@@ -1,51 +1,9 @@
-<label>
```
msorens, line 1: All of the deleted text is now the resource-dropdown component; this component subclasses that.

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.scss
```diff
@@ -1,116 +0,0 @@
-@import "~styles/variables";
```
msorens, line 1: Again, all moved to new parent class.

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.spec.ts
```diff
@@ -20,45 +25,11 @@ describe('ProjectsDropdownComponent', () => {
...
     component = fixture.componentInstance;
-    component.projectsUpdated = new EventEmitter();
     fixture.detectChanges();
   });
 
   it('should create', () => {
     expect(component).toBeTruthy();
   });
 
-  describe('dropdown', () => {
```
msorens, line 31: Tests moved, too. 😁

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
```diff
@@ -1,139 +1,80 @@
-import { Component, EventEmitter, Input, Output, OnChanges, OnInit, SimpleChanges } from '@angular/core';
```
msorens, line 1: Better **not** to look at the diff on this file--all the old text now comprises the parent class. The contents of this file is all new, having just the little bit needed for the subclass.

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
```diff
@@ -1,139 +1,80 @@
...
-    if (this.active) {
-      this.active = false;
-    }
-  }
+  constructor(private store: Store<NgrxStateAtom>) { }
 
-  handleFilterKeyUp(): void {
-    this.filteredProjects = this.filterProjects(this.filterValue);
+  ngOnInit(): void {
+    this.store.select(assignableProjects)
```
msorens, line 40: This block of code to generate the projects list from the ngrx store was previously duplicated by every caller; now it is encapsulated here, in the component itself.

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
```diff
@@ -1,139 +1,80 @@
...
-    } else if (event.key === 'ArrowDown') {
-      nextElement = <HTMLElement>targetElement.nextElementSibling;
-    }
-
-    if (nextElement == null) {
-      return;
-    } else {
-      nextElement.focus();
+  ngOnChanges(changes: SimpleChanges): void {
+    if (changes.checkedProjectIDs && changes.checkedProjectIDs.currentValue && this.projects) {
```
msorens, line 62: There are two sets of inputs to be concerned with: the set of all existing projects and the set of projects previously associated with the token. The block above generates the projects list from the ngrx state representation of allowable projects. Here we factor in the set of projects the token had previously selected (if any). Think of `onChanges` like an Observable for `@Input`s to the component.

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.html
```diff
@@ -0,0 +1,42 @@
+<label>{{ objectNounPlural | capitalize }}</label>
```
msorens, line 1: This was copied verbatim from the old projects-dropdown component, then just replaced "projects" with `objectNounPlural` to genericize it.

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+    it('displays a list of checkbox options', () => {
+      const options = Array.from(fixture.nativeElement.querySelectorAll('chef-checkbox'));
+      expect(options.length).toEqual(component.filteredResources.length);
+      options.forEach((option: HTMLInputElement, index: number) => {
+        const { name, checked } = component.filteredResources[index];
+        expect(option.textContent).toEqual(name);
+        expect(option.checked).toEqual(checked);
+      });
+    });
+
```
msorens, line 59: Yes, more tests coming soon!

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
```diff
@@ -0,0 +1,132 @@
...
+  // This input is needed only when re-displaying the resource dropdown
+  // for *additional* resources, as with the create-object-modal-component.
+  // Consumers that use it only for a single resource, e.g., team-details.component, should omit.
+  @Input() resourcesUpdated: EventEmitter<boolean>;
+
+  // (Optional) Label to use with no resources selected; defaults to given value if not provided.
+  @Input() noneSelectedLabel = 'None';
+
+  // Plural display name of resource (e.g. "policies", "projects").
+  @Input() objectNounPlural = 'MISSING REQUIRED PARAMETER';
```
msorens, line 36: To emphasize this is a required parameter, give it a very visibly wrong value indicative of what is needed.

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
```diff
@@ -0,0 +1,132 @@
+import { Component, EventEmitter, Input, Output, OnChanges, OnInit, SimpleChanges } from '@angular/core';
```
msorens, line 1: This file, which was previously projects-dropdown, is virtually unchanged, except for changing "projects" to "resources".

components/automate-ui/src/app/entities/policies/policy.effects.ts
```diff
@@ -125,11 +125,11 @@ export class PolicyEffects {
   @Effect()
   addPolicyMembersSuccess$ = this.actions$.pipe(
     ofType<AddPolicyMembersSuccess>(PolicyActionTypes.ADD_MEMBERS_SUCCESS),
-    map(({ payload: { members_added: added } }) => {
+    map(({ payload: { id, members_added: added } }) => {
       const addedStr = added.length === 1 ? added[0].displayName : `${added.length} members`;
       return new CreateNotification({
           type: Type.info,
-          message: `Added ${addedStr}.`
+          message: `Added ${addedStr} to "${id}" policy.`
```
msorens, line 132: A vital UX change; was rather odd-looking on screen without this context.

components/automate-ui/src/app/entities/projects/project.model.ts
```diff
@@ -9,9 +10,25 @@ export interface Project {
...
+}
+
 export class ProjectConstants {
   static readonly UNASSIGNED_PROJECT_ID = '(unassigned)';
   static readonly UNASSIGNED_PROJECT_LABEL = '(unassigned)';
   static readonly ALL_PROJECTS_LABEL = 'All projects';
   static readonly MULTIPLE_PROJECTS_LABEL = 'Multiple projects';
 }
+
+export function noProjectsUpdated(
```
msorens, line 28: DRYed up this function used in a couple other places.

components/automate-ui/src/app/modules/token/token-list/api-token-list.component.ts
```diff
@@ -150,6 +138,14 @@ export class ApiTokenListComponent implements OnInit, OnDestroy {
     };
     this.store.dispatch(new CreateToken(tok));
 
+    const member = stringToMember(`token:${tok.id}`);
+    (this.createTokenForm.controls.policies.value as string[])
+      .forEach(id =>
+        this.store.dispatch(new AddPolicyMembers(<PolicyMembersMgmtPayload>{
+          id,
+          members: [member]
+        }))
```
msorens, line 147: This is all that was needed to add the token to each selected policy.

components/automate-ui/src/app/entities/projects/project.model.ts
```diff
@@ -9,9 +9,6 @@ export interface Project {
   skip_policies?: boolean;
 }
 
-// Extend the project model with the checked field.
```
bcmdarroch: I liked that this comment specified that this is for UI state

components/automate-ui/src/app/modules/token/token-details/api-token-details.component.ts
```diff
@@ -130,9 +130,10 @@ export class ApiTokenDetailsComponent implements OnInit, OnDestroy {
     return <FormControl>this.updateForm.controls.name;
   }
 
-  // updates whether the project was checked or unchecked
-  onProjectChecked(project: ProjectChecked): void {
-    this.projects[project.id].checked = project.checked;
+  onProjectModalClosing(selectedProjects: string[]): void {
```
bcmdarroch: what's the project modal on the token details page?

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
```diff
@@ -52,37 +64,35 @@ export class CreateObjectModalComponent implements OnInit, OnDestroy, OnChanges
...
+
+      Object.values(this.projects).forEach(p => p.checked = false); // reset projects
       this.projectsUpdatedEvent.emit();
+
+      Object.values(this.policies).forEach(p => p.checked = false); // reset policies
+      this.policiesUpdatedEvent.emit();
+
       if (this.createProjectModal) {
-        this.updatePolicyCheckbox(true); // always set to checked upon opening
+        this.updatePolicyCheckbox(true); // default to checked upon opening
```
bcmdarroch, line 78: trying to remember why we would need to check if it's a project modal is making me want to refactor this modal too 😅

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.html
```diff
@@ -0,0 +1,42 @@
...
+<div class="dropdown-wrap">
+  <button class="dropdown-button" aria-label="Select resources"
```
bcmdarroch, line 3: will that aria-label change to the correct resource name?

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
msorens: Good observation; I saw that too, but too much scope creep for here.

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.html
msorens: Fixed

Issue #3277, opened 2020-04-02
ui: weird/inaccurate filtering results on compliance node details


vjeffrey:
## User Story
There have been a couple reports from customers about, for example, landing on the compliance reporting node detail page and seeing 4 results in the list where there should have been 6 according to the counts:
![image](https://user-images.githubusercontent.com/10341541/78310580-9c901600-750b-11ea-91ec-4e8c23aea1a7.png)

I haven't been able to reproduce this with the failures, but have seen it with waived controls:
<img width="1215" alt="Screen Shot 2020-04-02 at 17 54 37" src="https://user-images.githubusercontent.com/10341541/78310385-0e1b9480-750b-11ea-9af5-f5df9422a1be.png">

## Definition of Done
filter lists match the expected counts/report

## Demo Script / Repro Steps
`load_compliance_reports`
view node details on one of the compliance nodes
filter by waived controls

SEAjamieD:
@vjeffrey - After digging around for a while in this, I think there may be some value in pairing on this with someone from API (superteam?). From what I can tell, the counts for each status, aren't being generated in the front end.

I located a couple different functions in some Go files that might help? but i'm not saavy enough in Go yet to debug, or really even just print out some statements.

looking at these helper methods:
https://github.com/chef/automate/blob/c98adce639629da5f33cc9031ab7acb522e48e7e/components/compliance-service/ingest/events/compliance/helper_methods.go#L40-L93

Or possibly in reports.go?
https://github.com/chef/automate/blob/master/components/compliance-service/reporting/relaxting/reports.go#L993-L998

Thoughts/opinions on this?

susanev:
are we sure its the count that's wrong and not the list?

SEAjamieD:
I went through the API response manually, and based on the response I'm seeing, the list is correct.

I looked at the response programmatically and printed as well:

<img width="703" alt="Screen Shot 2020-04-20 at 3 03 37 PM" src="https://user-images.githubusercontent.com/16737484/79804185-5d89ff00-8318-11ea-9195-34be1dc35a6b.png">


<img width="266" alt="Screen Shot 2020-04-20 at 3 04 31 PM" src="https://user-images.githubusercontent.com/16737484/79804200-65e23a00-8318-11ea-9b99-66ef719138fd.png">

vjeffrey:
ya -- you're right -- this is a backend issue - thank you!

rickmarry:
waived controls take come in two varieties:
[...]

PR #3523, opened 2020-04-28
Add new function for checking if waived.

SEAjamieD:
The filters for compliance scans were incorrectly checking for waived values.  With guidance from Superteam, this branch corrects the check for waived values, so that the number matches the amount of reports displayed in all column.

fixes https://github.com/chef/automate/issues/3277

### :+1: Definition of Done
Counts on the top of each of the compliance reports columns matches the number or reports shown.

build component/automate-ui-devproxy && start_all_services
make serve

run `load_compliance_reports` in hab studio
navigate to https://a2-dev.test/compliance/reports/nodes
click on one of the nodes that has a lot of runs to see the detail view
filter by waived controls
the number on the top of waived controls column should match how many are in the filtered display

- [ ] Tests added/updated?
- [ ] Docs added/updated?

Screen Shot 2020-04-28 at 9 37 34 AM

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
+++ b/components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
@@ -15,6 +15,8 @@ export interface DateRange {
   end: Date | moment.Moment;
 }
 
+export type ControlStatus = 'passed' | 'failed' | 'waived' | 'skipped';
+
 @Injectable()
 export class StatsService {
   constructor(
@@ -215,14 +217,19 @@ export class StatsService {
   /* TODO Auth team: Helper functions that are only not private because
      they have unit testing. Should we delete their tests and make
      them private? */
-  getControlStatus(control): string {
+  getControlStatus(control): ControlStatus {
     const statuses = control.results.map(r => r.status);
-    const waived = control.waiver_data && control.waiver_data.skipped_due_to_waiver;
+    const waived = this.checkIfWaived(control.waived_str);
     if (waived) { return 'waived'; }
     if (statuses.every(s => s === 'passed')) { return 'passed'; }
     if (statuses.every(s => s === 'skipped')) { return 'skipped'; }
     return 'failed';
   }
+
+  private checkIfWaived(waivedStatus: string): boolean {
+    return waivedStatus === 'yes_run' || waivedStatus === 'yes';
+  }
+
 
   addDateRange(filters, dateRange) {
     if (filters) {
--- a/components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
+++ b/components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
@@ -218,9 +218,15 @@ export class StatsService {
      they have unit testing. Should we delete their tests and make
      them private? */
   getControlStatus(control): ControlStatus {
-    const statuses = control.results.map(r => r.status);
+    // In some cases, nodes don't have results data.  In these
+    // cases we want to display those nodes as skipped
+    const hasResults = control.results.length > 0;
+    if (!hasResults) { return 'skipped'; }
+
     const waived = this.checkIfWaived(control.waived_str);
     if (waived) { return 'waived'; }
+
+    const statuses = control.results.map(r => r.status);
     if (statuses.every(s => s === 'passed')) { return 'passed'; }
     if (statuses.every(s => s === 'skipped')) { return 'skipped'; }
     return 'failed';
--- a/components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
+++ b/components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
@@ -218,18 +218,16 @@ export class StatsService {
      they have unit testing. Should we delete their tests and make
      them private? */
   getControlStatus(control): ControlStatus {
-    // In some cases, nodes don't have results data.  In these
-    // cases we want to display those nodes as passed
-    const hasResults = control.results.length > 0;
-    if (!hasResults) { return 'passed'; }
-
     const waived = this.checkIfWaived(control.waived_str);
     if (waived) { return 'waived'; }
 
     const statuses = control.results.map(r => r.status);
     if (statuses.every(s => s === 'passed')) { return 'passed'; }
+    if (statuses.every(s => s === 'failed')) { return 'failed'; }
     if (statuses.every(s => s === 'skipped')) { return 'skipped'; }
-    return 'failed';
+    // In some cases, nodes don't have results data.  In these
+    // cases we want to display those nodes as passed
+    return 'passed';
   }
 
   private checkIfWaived(waivedStatus: string): boolean {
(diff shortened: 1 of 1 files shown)


components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
```diff
@@ -215,15 +217,20 @@ export class StatsService {
...
-    const waived = control.waiver_data && control.waiver_data.skipped_due_to_waiver;
+    const waived = this.checkIfWaived(control.waived_str);
     if (waived) { return 'waived'; }
     if (statuses.every(s => s === 'passed')) { return 'passed'; }
     if (statuses.every(s => s === 'skipped')) { return 'skipped'; }
     return 'failed';
   }
 
+  private checkIfWaived(waivedStatus: string): boolean {
+    return waivedStatus === 'yes_run' || waivedStatus === 'yes';
```
rickmarry, line 230: nice!

SEAjamieD, line 230: Ultimately the error was coming from we needed to be checking against if `control.waived_str` is equal to `yes` or `yes_run`.  If it is, this means its waived.
`yes` means that the item was waived, but chosen to not run
`yes_run` means that the item was waived, but still run.

SEAjamieD:
Thanks @vjeffrey @rickmarry  - will merge as soon as buildkite is back

components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
```diff
@@ -215,15 +217,20 @@ export class StatsService {
...
      them private? */
-  getControlStatus(control): string {
+  getControlStatus(control): ControlStatus {
     const statuses = control.results.map(r => r.status);
-    const waived = control.waiver_data && control.waiver_data.skipped_due_to_waiver;
+    const waived = this.checkIfWaived(control.waived_str);
     if (waived) { return 'waived'; }
     if (statuses.every(s => s === 'passed')) { return 'passed'; }
     if (statuses.every(s => s === 'skipped')) { return 'skipped'; }
     return 'failed';
```
vjeffrey, line 226: oooh one thing i just noticed -- some of the other reports we got were about the failure numbers not matching up -- bc the control didnt have any tests, for example. so if a control didnt have tests, it wouldnt have results -- and then it would get "failed" assigned to it, right?  

if any failed = failed
if pass + skip = pass
if all skip = skip
if no results...i guess this could be skip.

vjeffrey: we can do that in a separate change if ppl wanna hold since that changes the logic a bit more

SEAjamieD: the way its set up now, yeah the default is failed.
sounds like default should be skipped, instead?

SEAjamieD: makes no difference to me - can't merge right now anyways with the license issue.  Let me know what Superteam would like to see and i'll scope it out 👍

vjeffrey: ok, ya, if you can work out the logic for the failed/pass/skip/waived as described above, that would be awesome. we’ll also make sure the logic in the backend (for the counts) matches this.

vjeffrey: ok, ya, if you can work out the logic for the failed/pass/skip/waived as described above, that would be awesome. we’ll also make sure the logic in the backend (for the counts) matches this.

vjeffrey: ah - actually - no tests == passed.

so any failed == failed
all skip == skip
default pass

SEAjamieD: @vjeffrey - Its hard to test since i'm not sure where nodes have missing tests - but if I understand what you're asking for, I think adding a guard clause is a pragmatic approach.

components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
```diff
@@ -215,15 +217,26 @@ export class StatsService {
...
      they have unit testing. Should we delete their tests and make
      them private? */
-  getControlStatus(control): string {
-    const statuses = control.results.map(r => r.status);
-    const waived = control.waiver_data && control.waiver_data.skipped_due_to_waiver;
+  getControlStatus(control): ControlStatus {
+    // In some cases, nodes don't have results data.  In these
+    // cases we want to display those nodes as skipped
+    const hasResults = control.results.length > 0;
+    if (!hasResults) { return 'skipped'; }
```
SEAjamieD, line 224: @vjeffrey - said guard clause.   just checking if that results array has anything in it.  if its empty - its skipped.

vjeffrey: ah thank you -- can we change the skipped to passed real quick? i double checked the backend and when no results we actually say passed (which is kinda weird to me but i guess we should get consistent before changing)

SEAjamieD: @vjeffrey - no prob! updated to `passed`

SEAjamieD: ahh jk - default is passed....updating now

vjeffrey: thank you!

components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
```diff
@@ -219,9 +219,9 @@ export class StatsService {
      them private? */
   getControlStatus(control): ControlStatus {
     // In some cases, nodes don't have results data.  In these
```
vjeffrey: nitpick (on the comment): "nodes" -> "controls"

a profile is executed against a node
a profile is a collection of controls
a control is a collection of tests
each test has a result

components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
SEAjamieD: ok - default is set to passing now, also added a line checking for failed

components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
SEAjamieD: 😅 Comment updated.  Thanks for explaining the reasoning!

components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
```diff
@@ -215,15 +217,24 @@ export class StatsService {
...
-  getControlStatus(control): string {
-    const statuses = control.results.map(r => r.status);
-    const waived = control.waiver_data && control.waiver_data.skipped_due_to_waiver;
+  getControlStatus(control): ControlStatus {
+    const waived = this.checkIfWaived(control.waived_str);
     if (waived) { return 'waived'; }
+
+    const statuses = control.results.map(r => r.status);
     if (statuses.every(s => s === 'passed')) { return 'passed'; }
+    if (statuses.every(s => s === 'failed')) { return 'failed'; }
```
vjeffrey, line 226: statuses.any ?

any failure == failed

SEAjamieD: yikes - i need to slow down.  I changed to find, so that if it finds a match, its failure.

PR #3311, opened 2020-04-07
[hab] update pins for core plan refresh

ryancragun:
:warning: DO NOT MERGE until the refresh-2020q1-008 hab channel has been promoted to stable :warning: 

This change updates the hardcoded pins in Chef Automate to use versions in the refresh-2020q1-008 core plan refresh channel. It has been tested against refresh-2020q1-008 channel:
* https://buildkite.com/chef-oss/chef-automate-master-verify/builds/11382
* https://buildkite.com/chef/chef-automate-master-verify-private/builds/11761

## How to rebuild the universe
* [x] refresh-2020q1-008 has been promoted to the stable channel
* [ ] Rebuild and merge this PR (or force merge if you have godmode)
* [ ] Create a new master automate build job and use the master branch. https://buildkite.com/chef/chef-automate-master-habitat-build
  We use expeditors `smart_build` option so we'll need to trigger a build job manually to rebuild all the things. If we don't rebuild everything our airgap bundles will include multiple universes of packages.
* [ ] After the manual habitat/build job has completed and the artifacts have been promoted to the dev channel, create a new PR that updates the `automate-elasticsearch` and `automate-postgresql` pins in `.expeditor/create-manifest.rb` to the packages built in the manual habitat build.
* [ ] Merge the new PR
* [ ] :beers: 

Signed-off-by: Ryan Cragun <ryan@chef.io>

PR #3524, opened 2020-04-28
Add license_scout exceptions

jaym:
These were not being enforced before because the plans were
not using standard license string formats. The plans were
updated in the recent core plans refresh


jaym:
replaced by #3311

PR #3527, opened 2020-04-29
Only dispatch for policies during token creation if some specified

msorens:

This fixes a browser console error "cannot read property 'forEach' of null" if no policies are selected when creating a token.

NA

### :+1: Definition of Done
No browser console error

Rebuild automate-ui
Go to Settings >> Tokens >> Create Token
Attempt to create a token without specifying any policies to attach to.

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/modules/token/token-list/api-token-list.component.ts
+++ b/components/automate-ui/src/app/modules/token/token-list/api-token-list.component.ts
@@ -139,13 +139,15 @@ export class ApiTokenListComponent implements OnInit, OnDestroy {
     this.store.dispatch(new CreateToken(tok));
 
     const member = stringToMember(`token:${tok.id}`);
-    (this.createTokenForm.controls.policies.value as string[])
-      .forEach(id =>
+    const policies: string[] = this.createTokenForm.controls.policies.value;
+    if (policies) {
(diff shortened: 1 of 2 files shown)


components/automate-ui/src/app/modules/token/token-list/api-token-list.component.ts
```diff
@@ -139,13 +139,15 @@ export class ApiTokenListComponent implements OnInit, OnDestroy {
     this.store.dispatch(new CreateToken(tok));
 
     const member = stringToMember(`token:${tok.id}`);
-    (this.createTokenForm.controls.policies.value as string[])
```
srenatus, line 142: > cannot read property 'forEach' of null

I generally like `.forEach` just doing nothing for an empty array. What do you think, couldn't we do something like
```ts
(this.createTokenForm.controls.policies.value as string[] || []).forEach( id => /* ... */ );
```

It's a matter of code style, though, and what we deem more readible.

msorens: 👍 I do like that... but will add that in a follow-up. I do not want to delay promoting to acceptance with anything non-blocking.

PR #3528, opened 2020-04-29
automate-ui/teams/add-users: feed payload into success action

### LINKED ISSUE #3521: notification message wrong when adding a user to a team
## overview
if you add 1 user to a team, the notification gives you back the total number of users on a team. for example, if i have a team of 10 users. then add 1 more user the notification says `Added 11 users.`.

## what it should be
the notification message should represent the action you just took, so if i add x users to a team with n users already on the team the message should say `Added <x> users.`

srenatus:

Assuming that the UI knows the current state of team membership, this
change will fix the wrong notification:

For a team with n users, having added another k users, the notification
would have read

    "Added k+n users to team."

Now, we'll just say

    "Added k users to team."

Note that this may or may not be true: if another actor (using the UI or
the API) had changed team memberships at the same time, it might now be
accurate. But since this problem always exists, I think it's also
acceptible here.

- #3521 

### :+1: Definition of Done

✅ When adding one user to a team of two the notification reads "Added 1 user." not "Added 3 users."

- UI-only change: rebuild automate-ui from the branch,
- go to a teams page, add a user, observe the notification

image

### DIFF
--- a/components/automate-ui/src/app/entities/teams/team.effects.ts
+++ b/components/automate-ui/src/app/entities/teams/team.effects.ts
@@ -189,7 +189,7 @@ export class TeamEffects {
   addTeamUsers$ = this.actions$.pipe(ofType<AddTeamUsers>(TeamActionTypes.ADD_USERS),
     mergeMap(({ payload }: AddTeamUsers) =>
       this.requests.addTeamUsers(payload).pipe(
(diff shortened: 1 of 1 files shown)


srenatus:
Trivial, will merge when green.

PR #3529, opened 2020-04-29
automate-ui/telemetry: don't send projects header

### LINKED ISSUE #3130: Telemetry UI console log error
## Describe the bug
When filtering with a project with the global filter on the client runs page ( https://a2-dev.test/infrastructure/client-runs) and error is displayed in the console log. 

## To Reproduce
1. Add an infra node with `send_chef_run_failure_example`
1. Create a project with a rule that matches the org = "chef_delivery"
1. Apply the rule. 
1. Go to the client runs page https://a2-dev.test/infrastructure/client-runs
1. Filter the nodes by the new project created. 
1. When the project filter is applied the error should appear. 

## Expected behavior
No errors in the console

## Screenshots
Below is the error seen in the console.
image

- OS: ubuntu 18.04
- Browser chrome
- Automate Build Number 20200317181619 - this was master at the time

srenatus:

Now, we don't have a bunch of exceptions related to that header and CORS
in the browser console, and the requests flow again:

image

However, it gets us a new warning:

image

This is less bad. So let's iterate from here 😉 

- Fixes #3130 

### :+1: Definition of Done

✅ No console exception
✅ Telemetry stuff is fired way

- Rebuild UI, open browser with console open. Notice `api_keys`, `events`, and `t` requests succeeding.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/services/http/http-client-auth.interceptor.spec.ts
+++ b/components/automate-ui/src/app/services/http/http-client-auth.interceptor.spec.ts
@@ -151,7 +151,7 @@ describe('HttpClientAuthInterceptor', () => {
           const options = { params: { unfiltered: String(setting) } };
           httpClient.get('/endpoint', options).subscribe(done);
 
-          const httpRequest = httpMock.expectOne(`/endpoint?unfiltered=${setting}`);
+          const httpRequest = httpMock.expectOne('/endpoint');
           httpRequest.flush('response');
 
           const headerKeys = httpRequest.request.headers.keys();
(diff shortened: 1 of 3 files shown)


components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
```diff
@@ -31,17 +31,17 @@ export class HttpClientAuthInterceptor implements HttpInterceptor {
...
-    // after we've upgraded to angular 7.2+ (for this issue:
-    // https://github.com/angular/angular/issues/18812).
-    // const params = request.params.delete('unfiltered');
+    const params = request.params.delete('unfiltered');
     if (this.projects && filtered) {
       headers = headers.set('projects', this.projects);
     }
     return next
-      .handle(request.clone({ headers })).pipe(
+      .handle(request.clone({ headers, params })).pipe(
```
lancewf, line 44: Do you have to pass the `params` because they are updated? Where before they were part of the `request` object and were copied in the clone?

srenatus: ❓I'll just admit it, I have no idea. Monkey see, monkey do. This could be _not the right thing_, I'm afraid. What do you think?

srenatus:
@lancewf I think the high number of requests sent to segment is alright (at least I think I remember that being the case earlier) -- the radio silence we had before was a bug. I do wonder if anyone cares on the receiving end there: it must have been odd to have *no telemetry data* from a certain release on...?

srenatus:
@lancewf thanks for the review btw :)

components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
lancewf: That was my guess. That parameter is called "update" https://angular.io/api/common/http/HttpRequest.

srenatus: Thanks for the docs. I think that's what happens, yeah. :bulb:

msorens: To answer your original question @lancewf : _yes_.
Here is what I observed examining live code:
The `clone` operation clones everything: body, header, params, and others. If you examine the data structures in live code, when we effectively copy the headers (`request.headers.set`) or the params (`request.params.delete`), it is not just a simple object with those fields. It sets the `update` field you noticed, to the delta, essentially instructions on what to change in the object when that is passed to the `clone` method.

components/automate-ui/src/app/services/http/http-client-auth.interceptor.spec.ts
```diff
@@ -151,7 +151,7 @@ describe('HttpClientAuthInterceptor', () => {
           const options = { params: { unfiltered: String(setting) } };
           httpClient.get('/endpoint', options).subscribe(done);
 
-          const httpRequest = httpMock.expectOne(`/endpoint?unfiltered=${setting}`);
+          const httpRequest = httpMock.expectOne('/endpoint');
```
msorens, line 154: 🎉

susanev:
https://github.com/chef/automate/pull/3547 will fix that failure

srenatus:
will merge when green

PR #3556, opened 2020-04-30
External data stores lead to install inconsistency

sean-horn:
Creates problems with no benefit for new users.
This capability was intended for A2 HA cluster or some other engineered TBD product.
An installation installed by Chef Software, Inc. automation can be understood, will have been created to a certain spec, and is known to work. Other avenues lead to uncertainty about where the responsibility for support and troubleshooting lies.

Signed-off-by: Sean Horn <sean_horn@chef.io>

### DIFF
--- a/components/automate-chef-io/content/docs/install.md
+++ b/components/automate-chef-io/content/docs/install.md
@@ -32,11 +32,7 @@ sudo ./chef-automate init-config
 
 You can customize your FQDN, login name, and other values, by changing the values in the `config.toml` in your editor.
 
-If you have requirements around data size and/or redundancy, see [Configuring External
-Data Stores]({{< relref "#configuring-external-data-stores" >}}) for information on
-configuring Chef Automate to use an externally-deployed PostgreSQL database cluster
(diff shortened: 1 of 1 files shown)


ryancragun:
Deleting documentation about supported features doesn't seem like the correct avenue here. People have deployed Automate with external data stores and we support it for non-HA installs, therefore yanking the documentation rug out from under them doesn't seem that delightful. It seems like the core concern is that new users believe that they need use these features when they don't, so perhaps we need to add more clarity to the documentation that explains that these are advanced features that you should not use by default. What do you think?

danielcbright:
This doesn't seem like the best way to handle this, there are multiple customers already using this approach, and more to come since it's a documented feature. Perhaps the better avenue is to fully document and test the "big 3" cloud providers (as I've already started doing), and provide guidance on how to do this, with a clear line of delineation for where support begins and ends.

kagarmoe:
Let's handle this by creating an issue. We can return to the documentation once the underlying problem gets addressed.

PR #3561, opened 2020-04-30
Add legacy token to ingest policy

tylercloke:

Add legacy token into ingest policy if one has been configured on authn startup. If GRPC request fails, log an actionable error.

```
# cat test.toml
[auth_n.v1.sys.service]
a1_data_collector_token = "mytesttoken"
# chef-automate config patch test.toml
```

Verify that token exists but it is not in the ingest policy.

```
```

Verify that token is now in ingest policy.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/authn-service/server/server.go
+++ b/components/authn-service/server/server.go
@@ -26,7 +26,7 @@ import (
 )
 
 const (
-	IngestAccessPolicyID = "ingest-access"
+	IngestPolicyID = "ingest-access"
 )
 
 // Config holds the server's configuration options.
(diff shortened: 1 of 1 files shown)


components/authn-service/server/server.go
```diff
@@ -24,6 +25,10 @@ import (
 	"github.com/chef/automate/lib/version"
 )
 
+const (
+	IngestAccessPolicyID = "ingest-access"
```
msorens, line 29: Let's be consistent with authz-service:
```suggestion
	IngestPolicyID = "ingest-access"
```

components/authn-service/server/server.go
```diff
@@ -177,6 +166,35 @@ func newServer(ctx context.Context, c Config) (*Server, error) {
...
+			tokenID = existingID
+		}
+
+		_, err := policiesClient.AddPolicyMembers(ctx, &authz.AddPolicyMembersReq{
+			Id:      IngestAccessPolicyID,
+			Members: []string{fmt.Sprintf("token:%s", tokenID)},
+		})
+		if err != nil {
+			s.logger.Warn(errors.Wrap(err, "there was an error granting the legacy data collector token ingest access").Error())
+			s.logger.Warn(fmt.Sprintf("please manually add token with ID %s to the policy with ID %s", tokenID, IngestAccessPolicyID))
```
msorens, line 194: Will be slightly clearer:
```suggestion
			s.logger.Warn(fmt.Sprintf("please manually add token with ID %q to the policy with ID %q", tokenID, IngestAccessPolicyID))
```

srenatus: Don't we have a Warnf? (or is this the zap logger...?) :thinking:

tylercloke: Zap :wee:

PR #3562, opened 2020-05-01
[Automate-3537] Sequence addPolicyMembers from createToken

msorens:

The initial pass introducing adding a token to policies during token creation made the two actions concurrent: creating the token and adding it to policies. That strategy only works well, though, if the token creation succeeds. If it fails due to an id conflict--a token of the same name already exists--then it is inappropriate for the policy assignment to still be done.

This PR sequences those so that the policy assignment is initiated iff the token creation is successful.

Let us also consider the flip side--what if token creation succeeds but the policy assignment fails? Should the two operations be done transactionally so that both must succeed otherwise neither does? I would suggest no, it is OK to create the token regardless, and if the policy assignments failed they could be retried through other routes. (Partly because that seems reasonable to me and partly because the implementation of this is straightforward; hence, this PR. 😁 )

PR #3511 (Introduce policy dropdown for tokens)

### :+1: Definition of Done
Policy assignments are not attempted if token creation fails for any reason.

1. Rebuild automate-ui.
2. Navigate to Settings >> Tokens
3. Use <kbd>Create Token</kbd> to create an initial token, t1; neither projects nor policies matters.
4. Again open up the token creation modal with Use <kbd>Create Token</kbd>. Enter the same id for the token and this time be sure to select one or more policies from the dropdown.
5. Upon pressing <kbd>Create</kbd> you should see an in-modal error notification and on the underlying page **_no_** notification for policies.
6. Change the id to something unique so the token creation will succeed. This time observe that the modal closes, the first notification indicates token created, and then there is an additional success notification for each policy you specified.

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/modules/token/token-list/api-token-list.component.spec.ts
+++ b/components/automate-ui/src/app/modules/token/token-list/api-token-list.component.spec.ts
@@ -1,12 +1,15 @@
+import { HttpErrorResponse } from '@angular/common/http';
+import { ReactiveFormsModule } from '@angular/forms';
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 import { RouterTestingModule } from '@angular/router/testing';
-import { ReactiveFormsModule } from '@angular/forms';
 import { StoreModule, Store } from '@ngrx/store';
 import { MockComponent } from 'ng2-mock-component';
 
 import { runtimeChecks, ngrxReducers, NgrxStateAtom } from 'app/ngrx.reducers';
 import { ChefPipesModule } from 'app/pipes/chef-pipes.module';
 import { FeatureFlagsService } from 'app/services/feature-flags/feature-flags.service';
+import { CreateTokenSuccess, CreateTokenFailure } from 'app/entities/api-tokens/api-token.actions';
+import { ApiToken } from 'app/entities/api-tokens/api-token.model';
 import { ApiTokenListComponent } from './api-token-list.component';
 
 describe('ApiTokenListComponent', () => {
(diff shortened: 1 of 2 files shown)


components/automate-ui/src/app/modules/token/token-list/api-token-list.component.ts
```diff
@@ -87,6 +87,7 @@ export class ApiTokenListComponent implements OnInit, OnDestroy {
       .subscribe(state => {
         this.creatingToken = false;
         if (state === EntityStatus.loadingSuccess) {
+          this.assignPolicies();
```
msorens, line 90: Previously, this call (inline) to dispatch AddPolicyMembers was done at the same time the CreateToken action was dispatched. By moving it here, we now wait for a successful notification of token creation before starting the policy additions.

components/automate-ui/src/app/modules/token/token-list/api-token-list.component.spec.ts
```diff
@@ -90,18 +93,48 @@ describe('ApiTokenListComponent', () => {
...
+      component.createModalVisible = true;
       spyOn(store, 'dispatch').and.callThrough();
       component.createTokenForm.controls.policies.setValue(null);
       component.createToken();
       expect(store.dispatch).toHaveBeenCalledTimes(1);
+
+      store.dispatch(new CreateTokenSuccess({ 'id': 't1', 'projects': [] } as ApiToken));
+      fixture.detectChanges();
+
+      const setupCallsToDispatch = 2;
```
msorens, line 105: That's one from calling `createToken` plus one explicitly called on L102 above.

components/automate-ui/src/app/modules/token/token-list/api-token-list.component.ts
```diff
@@ -132,22 +133,20 @@ export class ApiTokenListComponent implements OnInit, OnDestroy {
...
-        this.store.dispatch(new AddPolicyMembers(<PolicyMembersMgmtPayload>{
-          id,
-          members: [member]
-        }))
-      );
-    }
+  private assignPolicies(): void {
+    const member = stringToMember(`token:${this.createTokenForm.controls.id.value}`);
+    const policies: string[] = this.createTokenForm.controls.policies.value || [];
+    policies.forEach(id => this.store.dispatch(new AddPolicyMembers(<PolicyMembersMgmtPayload>{
```
srenatus, line 146: ❔ What does the cast give us here? (Sorry, I can see it's moved code, just wondering.)

components/automate-ui/src/app/modules/token/token-list/api-token-list.component.ts
```diff
@@ -132,22 +133,20 @@ export class ApiTokenListComponent implements OnInit, OnDestroy {
...
-    if (policies) {
-      policies.forEach(id =>
-        this.store.dispatch(new AddPolicyMembers(<PolicyMembersMgmtPayload>{
-          id,
-          members: [member]
-        }))
-      );
-    }
+  private assignPolicies(): void {
+    const member = stringToMember(`token:${this.createTokenForm.controls.id.value}`);
```
srenatus, line 144: Would this look cleaner if the id was passed? `assignPolicies(tokenID: string, policyIDs: string[]): void`? What do you think?

components/automate-ui/src/app/modules/token/token-list/api-token-list.component.ts
msorens: Actually, it is unneeded here--good eye!

components/automate-ui/src/app/modules/token/token-list/api-token-list.component.ts
msorens: I agree with you: on one hand, it shows a bit more... intent. But on the other hand (which is why I chose to do it this way), I prefer my subscriptions (the call site for this) to be more "high-level" in a sense, typically just dispatching something or setting a property to open/close something, then let the helper functions have the nitty-gritty details.

PR #3565, opened 2020-05-01
[automate-2699] deflake team add users

blakestier:

Deflaking some teams add users tests.

Refreshing the store is causing the modal to continue to pop up, so hiding the modal "fixed" the flakiness. Seems like there should be another way to ensure that all the things are present without the storage changes, but none were obvious to me. 

We were also passing a malformed request to delete the users after the fact.

These tests passed CI on 4 separate builds . . . is that enough?

Run the tests?

### DIFF
--- a/e2e/cypress/integration/ui/settings/team_add_users.spec.ts
+++ b/e2e/cypress/integration/ui/settings/team_add_users.spec.ts
@@ -52,6 +52,7 @@ describe('team add users', () => {
     // the whole app has to reload, slowing down the test and causing timeouts
     cy.visit(`/settings/teams/${teamID}/add-users`);
     cy.wait(['@getTeam', '@getTeamUsers', '@getUsers']);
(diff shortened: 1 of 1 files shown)


e2e/cypress/integration/ui/settings/team_add_users.spec.ts
```diff
@@ -52,6 +50,7 @@ describe('team add users', () => {
     // the whole app has to reload, slowing down the test and causing timeouts
```
msorens, line 50: BTW, did you happen to try what this TODO comment suggested?

blakestier: @msorens I did, without much luck. If only done in the before, it seemed that the store wasn't properly updated. I also considered clicking the "Add Users" button every time, although that seemed slightly out of scope (is that then a "Team Details" test? 🤔).

PR #3578, opened 2020-05-04
Limit the number of inflight datacollector requests

jaym:
The config:
```toml
[gateway.v1.sys.data_collector.limiter]
# Setting disable to true will allow an unbounded number of
# data collector requests to remain inflight concurrently.
disable = false
# max_inflight_requests will set the maximum number of
# concurrent inflight data collector requests. By default,
# this value is 60 * runtime.CpuCount()
max_inflight_requests = 100
```


components/automate-gateway/gateway/server.go
```diff
@@ -160,6 +163,10 @@ func (s *Server) Start() error {
 	return s.startSignalHandler()
 }
 
+func (s *Server) loadDatacollectorLimiter() {
+	s.datacollectorLimiter = limiter.NewInflightRequestLimiter("collector-requests", 800)
```
stevendanna, line 167: I wonder if we want this to be something like 100*CpuCount

stevendanna:
I meant to comment but not approve yet.

components/automate-gateway/gateway/server.go
jaym: maybe. the number is difficult to pick and the knobs that are now on ingest-service and compliance-service make it even harder.

components/automate-gateway/pkg/limiter/limiter.go
```diff
@@ -0,0 +1,90 @@
...
+	max = l.maxRequests
+	if cur < max {
+		l.inflight++
+		taken = true
+	}
+	l.lock.Unlock()
+
+	if !taken {
+		err := status.Errorf(codes.ResourceExhausted,
+			"Resource limit exceeded for %s: cur=%d max=%d", l.name, cur, max)
```
srenatus, line 55: 💭 Would cur and max ever have values that aren't equal?

components/automate-gateway/pkg/limiter/limiter.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+type noopRequestLimiterTicket struct {
+}
+
+func NewNoopRequestLimiter() Limiter {
+	return noopRequestLimiter{}
+}
+
+func (l noopRequestLimiter) TakeTicket() (Ticket, error) {
```
srenatus, line 84: [nit]
```suggestion
func (noopRequestLimiter) TakeTicket() (Ticket, error) {
```

components/ingest-service/pipeline/chef_run.go
```diff
@@ -35,6 +35,8 @@ func NewChefRunPipeline(client backend.Client, authzClient authz.ProjectsClient,
 		counter int64 = 0
 	)
 
+	log.Infof("Message buffer size = %d", messageBufferSize)
```
srenatus, line 38: Is this intentional? Looks slightly unrelated 🤔

components/automate-gateway/pkg/limiter/limiter.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"errors"
+	"sync"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+var ErrLimitExceeded = errors.New("Resource limit exceeded")
```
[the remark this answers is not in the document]
srenatus: Related -- it's a bit odd this general purpose thing returns grpc statuses. But it's not a big deal.

components/automate-gateway/gateway/datacollector.go
```diff
@@ -58,6 +58,14 @@ func (s *Server) dataCollectorHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	ticket, err := s.datacollectorLimiter.TakeTicket()
+	if err != nil {
+		s.logger.Error(err.Error())
+		http.Error(w, err.Error(), http.StatusServiceUnavailable)
```
srenatus, line 64: Should we follow the mapping that grpc-gateway uses for the endpoints it provides? I believe that would mean 429 Too Many Requests,  https://github.com/googleapis/googleapis/blob/1b5a8d2bbb69c5a04db26bd546d2888e609c6bab/google/rpc/code.proto#L104-L108

Also, it would perhaps be more defensive to check if the returned error really is the one where there's no tickets left. At least from the interface, it could be anything that's going wrong.

jaym: I dont think 429 is the correct one in this case as it isn't a limit applied to an individual client. The server is overloaded. My returning a 5xx, we tell the clients to come back. I'm not sure 429 would do that.

Ill fix up the errors to return the generic thing

components/automate-gateway/pkg/limiter/limiter.go
jaym: only if there are bugs

srenatus: I'd prefer to assume their presence. But feel free to ignore me if you feel differently.

jaym: we don't need to log both, but we should log at least one as it makes debugging easier

jaym:
@srenatus do you know which ones specifically? I can take a look

srenatus:
components/ingest-service/test/ the stuff tested here, /events/chef/run etc

jaym:
i cant find any reference to those in the docs. double checking with the data collection peoples

components/automate-gateway/gateway/datacollector.go
```diff
@@ -58,6 +59,18 @@ func (s *Server) dataCollectorHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	ticket, err := s.dataCollectorLimiter.TakeTicket()
+	if err != nil {
+		if errors.Is(err, limiter.ErrLimitExceeded) {
```
srenatus, line 64: Thanks for not ignoring me! Also, cool, `errors.Is` :+1:

components/automate-gateway/pkg/limiter/limiter.go
```diff
@@ -0,0 +1,88 @@
...
+	cur = l.inflight
+	max = l.maxRequests
+	if cur < max {
+		l.inflight++
+		taken = true
+	}
+	l.lock.Unlock()
+
+	if !taken {
+		err := fmt.Errorf("resource=%s cur=%d max=%d: %w",
```
srenatus, line 52: Now I'm confused. Is `errors.Is(err, ErrLimitExceeded)` _true_? But I'm also tired. Will have another look in the morning.

jaym: the tests below say yes?

stevendanna: I believe it will be. The `%w` formatter does more than just take the error's message and keeps track of the wrapped errors. errors.Is will return true even for deeply nested versions of this error.

srenatus: Yeah, I hadn't claimed it wasn't true, I just didn't see how. Thanks @stevendanna, that's indeed the case:

> If the format specifier includes a %w verb with an error operand, **the returned error will implement an Unwrap method returning the operand.** It is invalid to include more than one %w verb or to supply it with an operand that does not implement the error interface. The %w verb is otherwise a synonym for %v.

https://golang.org/pkg/fmt/#Errorf

Nice. We could look into automatically refactoring all our error wraps...

PR #3583, opened 2020-05-04
Comment out another flaky sidebar test

tylercloke:


tylercloke:
@msorens if this passes i'll do that in a follow up PR so that we don't have to wait on CI to unblock the pipeline

msorens:
Probably no need for a follow-up comment fix, @tylercloke ; I will likely touch that in my latest card-in-progress.

Issue #3515, opened 2020-04-28
one node view: issues


vjeffrey:
## User Story

we've created a basic (hidden) view in automate, at `automate-url/nodes`, that surfaces the data provided by the `api/v0/nodes` endpoint.

this issue is to track any issues we see with that data:
- platform version is sometimes doubled up
<img width="433" alt="Screen Shot 2020-04-27 at 19 39 27" src="https://user-images.githubusercontent.com/10341541/80437449-d3fda280-88be-11ea-9305-d6591e143f2e.png">

- manager field is blank for nodes that are attached to a chef server - we could fill this field in with the value of the chef server

- environment value should be added to the tags sent to nodemanager so we can show it 

- nodes marked as terminated (from aws-ec2, for example) are not auto-deleted. should we auto-delete them?

- nodes attached to a manager (aws-ec2, for example) are not auto-deleted when deleting the manager, should we?

- should add the created at field to nodemanager

vjeffrey:
wip https://github.com/chef/automate/tree/vj/nodes-fixes

PR #3596, opened 2020-05-05
nodemanager nodes: fix platform and tags info, add state to possible sort options

vjeffrey:
a few issues:
- platform version was showing up twice in the nodemanager nodes: this is bc we were mistakenly sending the full platform field + the version, instead of either just sending the full field or jsut sending the family and version. i switched to used family for now, for consistency with the compliance side of things, but we may want to come back and just have everything send the full field? not sure yet. should discuss how customers would want to filter that data.
- tags: added environment value as a tag to the node, and also modified the compliance bits to include chef tags in the tags information sent to nodemanager
- state: we did not previously allow sorting by state; we should!

### :+1: Definition of Done
platform information doesn't double up. tags are all sent over, including env. can filter by state.

go_update_component compliance-service
same for ingest-service
and nodemanager-service
send in some reports, e.g. load_compliance_reports
check the data returned by `api/v0/nodes`, accessible at automate-url/nodes in the ui

- [x] Tests added/updated?
- [n/a] Docs added/updated?

### DIFF
--- a/components/ingest-service/backend/chef_client_run.go
+++ b/components/ingest-service/backend/chef_client_run.go
@@ -250,7 +250,7 @@ func (ccr *ChefClientRun) initializeNodeInfo() (sharedNodeInfo NodeInfo, err err
 		UptimeSeconds:         int64(uptimeSeconds),
 		OrganizationName:      ccr.OrganizationName,
 		Environment:           ccr.NodePayload.ChefEnvironment,
-		Platform:              getPlatformWithVersion(ccr.NodePayload),
+		Platform:              ccr.PlatformWithVersion(),
 		PlatformFamily:        EmptyStringIfNil(ccr.NodePayload.Automatic["platform_family"]),
 		PlatformVersion:       EmptyStringIfNil(ccr.NodePayload.Automatic["platform_version"]),
 		Source:                ccr.Source,
@@ -353,6 +353,10 @@ func (ccr *ChefClientRun) CloudProvider() string {
 	cloud := extractMapOrEmpty(ccr.NodePayload.Automatic, "cloud")
 
 	return EmptyStringIfNil(cloud["provider"])
+}
+
+func (ccr *ChefClientRun) Platform() string {
+	return EmptyStringIfNil(ccr.NodePayload.Automatic["platform"])
 }
 
 // ChefVersion Returns a chef version string retrieved from automatic attributes or an empty string if it is not present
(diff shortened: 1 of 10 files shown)


components/nodemanager-service/pgdb/nodes.go
```diff
@@ -166,6 +166,7 @@ var nodesSortFields = map[string]string{
 	"status":           "LOWER(n.status)",
 	"manager":          "LOWER(n.manager)",
 	"last_contact":     "n.last_contact",
+	"state":            "n.source_state",
```
vjeffrey, line 169: leaving as draft til i get a test written for this

components/ingest-service/pipeline/publisher/node_manager_publisher.go
```diff
@@ -75,11 +75,14 @@ func gatherInfoForNode(node backend.Node) (*manager.NodeMetadata, error) {
...
 	}
+	if node.Environment != "" {
+		tags = append(tags, &common.Kv{Key: "environment", Value: node.Environment})
+	}
 
 	return &manager.NodeMetadata{
 		Uuid:            node.EntityUuid,
 		Name:            node.NodeName,
-		PlatformName:    node.Platform,
+		PlatformName:    node.PlatformFamily,
```
vjeffrey, line 85: maybe we want to eventually go to just sending the platform full field?  for now, i switched it to using family for consistency with compliance side of things, since it's sending name/version as two sep fields

components/nodemanager-service/pgdb/nodes.go
vjeffrey: done!

components/ingest-service/pipeline/publisher/node_manager_publisher.go
lancewf: For our example CCR https://github.com/chef/automate/blob/82772a8c0a82da983bb9b329fb31716447d66609/components/ingest-service/examples/converge-success-report.json#L1391 The platform_family "rhel" is different than the platform "redhat").

lancewf: I think I have a solution :thinking:

lancewf: Here is one solution https://github.com/chef/automate/pull/3612

vjeffrey:
ill get this rebased

components/ingest-service/backend/chef_client_run.go
```diff
@@ -355,6 +355,10 @@ func (ccr *ChefClientRun) CloudProvider() string {
 	return EmptyStringIfNil(cloud["provider"])
 }
 
+func (ccr *ChefClientRun) Platform() string {
```
vjeffrey, line 358: oh i think this is from the rebase and no longer needed?

vjeffrey: oh nm we call it below

PR #3599, opened 2020-05-05
On project list fetch, remove any projects that no longer exist

tylercloke:

If a project got deleted by someone besides the active UI user, they could persist in localstorage. Now when we fetch the project list on app start, intersect localstorage with that list so there are no zombies 👻.

* Create project `test` in the UI.
* Check it in projects dropdown.
* See that it shows up in local storage.
* Delete the project via the API:
```curl -H "api-token: $TOK" -XDELETE  https://localhost/apis/iam/v2/projects/test --insecure```
* Refresh the app and see that it no longer exists in localstorage and that the dropdown goes away since now there is only the default project.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/services/projects-filter/projects-filter.effects.ts
+++ b/components/automate-ui/src/app/services/projects-filter/projects-filter.effects.ts
@@ -63,13 +63,13 @@ export class ProjectsFilterEffects {
         const fetched = this.convertResponse(resp.projects);
         let restored = this.projectsFilter.restoreOptions() || [];
 
-        // throw out any localstoreage keys that no longer exist.
-        // intersectionBy choses results from the first array so
+        // throw out any LocalStorage keys that no longer exist.
+        // intersectionBy chooses results from the first array so
         // restored's checked results are preserved.
         restored = intersectionBy('value', restored, fetched);
 
         // don't reload the page, but persist to localstorage
-        this.projectsFilter.updateLocalstorage(restored);
+        this.projectsFilter.updateLocalStorage(restored);
 
         return new LoadOptionsSuccess({
           fetched,
--- a/components/automate-ui/src/app/services/projects-filter/projects-filter.effects.ts
+++ b/components/automate-ui/src/app/services/projects-filter/projects-filter.effects.ts
@@ -3,6 +3,8 @@ import { HttpErrorResponse } from '@angular/common/http';
 import { Actions, Effect, ofType } from '@ngrx/effects';
 import { interval as observableInterval, of as observableOf, Observable } from 'rxjs';
 import { catchError, mergeMap, map, tap } from 'rxjs/operators';
+import { intersectionBy } from 'lodash/fp';
+
 
 import { Project } from 'app/entities/projects/project.model';
 import { ProjectsFilterOption } from './projects-filter.reducer';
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: components/automate-ui/src/app/services/projects-filter/projects-filter.effects.ts [hunk 1]
```diff
@@ -57,11 +59,20 @@ export class ProjectsFilterEffects {
...
   private loadOptionsAction$(): () => Observable<ProjectsFilterActions> {
     return () => this.requests.fetchOptions().pipe(
-      map((fetched: AuthorizedProjectsResponse) => {
-        const converted = this.convertResponse(fetched.projects);
-        const restored = this.projectsFilter.restoreOptions() || [];
+      map((resp: AuthorizedProjectsResponse) => {
+        const fetched = this.convertResponse(resp.projects);
+        let restored = this.projectsFilter.restoreOptions() || [];
+
+        // throw out any localstoreage keys that no longer exist.
```
bcmdarroch, line 66:
```suggestion
        // throw out any localstorage keys that no longer exist.
```

components/automate-ui/src/app/services/projects-filter/projects-filter.service.ts
```diff
@@ -43,8 +43,12 @@ export class ProjectsFilterService {
     this.store.dispatch(new UpdateSelectionCount(options));
   }
 
-  storeOptions(options: ProjectsFilterOption[]) {
+  updateLocalstorage(options: ProjectsFilterOption[]) {
     localStorage.setItem(STORE_OPTIONS_KEY, JSON.stringify(options));
+  }
+
+  storeOptions(options: ProjectsFilterOption[]) {
```
bcmdarroch, line 50: what does this change do?

components/automate-ui/src/app/services/projects-filter/projects-filter.service.ts
```diff
@@ -43,8 +43,12 @@ export class ProjectsFilterService {
     this.store.dispatch(new UpdateSelectionCount(options));
   }
 
-  storeOptions(options: ProjectsFilterOption[]) {
+  updateLocalstorage(options: ProjectsFilterOption[]) {
```
msorens, line 46:
```suggestion
  updateLocalStorage(options: ProjectsFilterOption[]) {
```

components/automate-ui/src/app/services/projects-filter/projects-filter.effects.ts
```diff
@@ -57,11 +59,20 @@ export class ProjectsFilterEffects {
...
+        // intersectionBy choses results from the first array so
```
msorens, line 67:
```suggestion
        // throw out any localStorage keys that no longer exist.
        // intersectionBy chooses results from the first array so
```

msorens:
@bcmdarroch Yes, we do refresh every two minutes. The hard refresh is needed only if you are impatient. 😁

components/automate-ui/src/app/services/projects-filter/projects-filter.effects.ts
tylercloke: :facepalm:

PR #3600, opened 2020-05-06
add a1 legacy token integration test

srenatus:
Adding the test now since I've just been messing with this for acceptance testing.

- [x] see test pass correctly (the logs have a permission granted line, so the 400 is the ingestion handler)
- [x] move into nightlies 🌃

### DIFF
--- a/.expeditor/nightly.pipeline.yml
+++ b/.expeditor/nightly.pipeline.yml
@@ -96,6 +96,15 @@ steps:
           single-use: true
           privileged: true
 
+  - label: "[integration] a1 legacy token"
+    command:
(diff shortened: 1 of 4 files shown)


srenatus:
@stevendanna absolutely! thanks for the hint

srenatus:
I'll move it once I'm satisfied with the test script.

srenatus:
@chef/automate-deployment-team What do think about https://github.com/chef/automate/pull/3600/commits/5bcd4125d3785180f82a35a94654c165401bcc51? I can drop it, it just seemed like it belonged there.

PR #3612, opened 2020-05-06
Passing the base platform name to the nodemanager

lancewf:
Before:
before_nodes

After:
after_nodes

### DIFF
--- a/components/ingest-service/backend/chef_client_run.go
+++ b/components/ingest-service/backend/chef_client_run.go
@@ -250,7 +250,7 @@ func (ccr *ChefClientRun) initializeNodeInfo() (sharedNodeInfo NodeInfo, err err
 		UptimeSeconds:    int64(uptimeSeconds),
 		OrganizationName: ccr.OrganizationName,
 		Environment:      ccr.NodePayload.ChefEnvironment,
-		Platform:         getPlatformWithVersion(ccr.NodePayload),
+		Platform:         ccr.PlatformWithVersion(),
 		PlatformFamily:   EmptyStringIfNil(ccr.NodePayload.Automatic["platform_family"]),
 		PlatformVersion:  EmptyStringIfNil(ccr.NodePayload.Automatic["platform_version"]),
 		// Tags                 ccr. TODO: Is this compliance tags?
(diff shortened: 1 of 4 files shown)


components/ingest-service/backend/chef_client_run.go
```diff
@@ -250,7 +250,7 @@ func (ccr *ChefClientRun) initializeNodeInfo() (sharedNodeInfo NodeInfo, err err
 		UptimeSeconds:    int64(uptimeSeconds),
 		OrganizationName: ccr.OrganizationName,
 		Environment:      ccr.NodePayload.ChefEnvironment,
-		Platform:         getPlatformWithVersion(ccr.NodePayload),
+		Platform:         ccr.PlatformWithVersion(),
```
lancewf, line 253: This is cleaning up the way we grab the platform.

components/ingest-service/pipeline/message/chef_run.go
```diff
@@ -30,6 +30,7 @@ type ChefRun struct {
 	NodeRun          backend.Run
 	NodeAttribute    backend.NodeAttribute
 	BulkableRequests []elastic.BulkableRequest
+	Platform         string
```
lancewf, line 33: Added the raw platform to the message that is passed along in the pipeline.

components/ingest-service/pipeline/processor/transmogrify_run.go
```diff
@@ -80,6 +80,9 @@ func ChefRunTransmogrify(in <-chan message.ChefRun, out chan<- message.ChefRun,
 				continue
 			}
 
+			// Needed for the nodemanager because the platform field is combined with the version.
+			msg.Platform = ccr.Platform()
```
lancewf, line 84: This is the only place we have access to the raw platform. So, we set the pipeline msg platform here.

components/ingest-service/pipeline/publisher/node_manager_publisher.go
```diff
@@ -82,7 +84,7 @@ func gatherInfoForNode(node backend.Node) (*manager.NodeMetadata, error) {
 	return &manager.NodeMetadata{
 		Uuid:            node.EntityUuid,
 		Name:            node.NodeName,
-		PlatformName:    node.PlatformFamily,
+		PlatformName:    msg.Platform,
```
lancewf, line 87: Then at this point, we can set the platform from the pipeline message.

PR #3606, opened 2020-05-06
[automate-gateway] Enable /status endpoint via the load balancer

stevendanna:
Signed-off-by: Steven Danna <steve@chef.io>

PR #3614, opened 2020-05-07
automate-{gateway,load-balancer} integration: add test for /api/v0/status

srenatus:
Left-over from #3606 (broken in #3474): add integration test coverage for the status endpoint

There are so many places where these could be put -- I've chosen the Bash place.


srenatus:
Trivial, merging.

PR #3617, opened 2020-05-07
[Automate-1044] Permission-based routing for top nav bar

msorens:

This PR applies the permission-based routing scheme (previously in use just for the sidebar navigation panel) to the top navigation bar.

1. Adds an introspection guard to the one unguarded element, `Dashboards`. Now, each of the top navigation choices will appear only when the user has permission to access it.
2. Changes from static routing to dynamic (permission-based routing): upon navigating to the root level URL (e.g. `https://a2-dev.test/` or `https://a2-local-inplace-upgrade-acceptance.cd.chef.co/`), previously one landed just on the first listed element, `Dashboards` -- whether or not having had permission to access it. Now, the set of top-level navigation choices are examined in turn; the user lands on the first one that they have permission to access, and the URL in the browser bar auto-adjusts to that route.
3. Should the user have **_no_** permission for any of the top-level navigation choices, they are presented a new, "no permission" screen and the URL remains at the root level URL.

OK, but what is really cool about this pull request?
This is almost a "no-code" implementation. There is one new component, a `TopNavLandingComponent` that subclasses the `LandingComponent` by providing one variable (a set of routes and permissions) and a one-line function (`SetPageCoverage`) that tells the parent to hide the sidebar when there are no permissions. This was possible because of the recently redesigned sidebar and landing component architectures (thanks @tarablack01 for the sidebar!).

Note that permission-based routing **_only_** occurs when navigating to the root level URL (e.g. `https://a2-dev.test/`); it does not occur if you happen to know a sub-URL under there and you enter that directly (e.g., `https://a2-dev.test/dashboards/event-feed`). Thus, it is still useful for individual pages to be able to show a page-level "no permissions here", which several pages do:
image

* [Convert SettingsLandingComponent to reusable landing component](https://github.com/chef/a2/pull/5039)
* [Settings landing page for permission-based routing](https://github.com/chef/a2/pull/4642)

### :+1: Definition of Done
See next section.

Rebuild automate-ui.
Create a new user, e.g. `bob`.
Login as `bob` and observe that the URL shows the top-level domain in the address bar (e.g. `https://a2-dev.test/`) and the new "no permissions" page.
(Note that the same styles used on the pre-existing page shown above were applied to this page for consistency.)
image

Next, create a policy granting `bob` selective permissions to see tokens (under `Settings`) and nodes (under `Infrastructure`):
```
curl -sSkH "api-token: $TOK" -X POST \
-d '{ "name": "perm routing test", "id": "perm-routing-test", "members": [ "user:local:bob" ], "statements": [ { "effect": "ALLOW", "actions": [ "iam:tokens:*","infra:nodes:list" ], "projects": [ "*" ] } ], "projects": [] }' \
$TARGET_HOST/apis/iam/v2/policies?pretty
```
Refresh bob's browser page. You will now see just those two choices available on the top navigation bar, and you will land on the first one, with the address bar reflecting the newly-routed page (`https://a2-dev.test/infrastructure/client-runs`).
image

Next, open up bob's permissions further, by, e.g. adding to a project owner policy and you will see all available top nav bar choices (or you could just log in as an admin to see the same):
image

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-chef-io/content/docs/event-feed.md
+++ b/components/automate-chef-io/content/docs/event-feed.md
@@ -12,6 +12,8 @@ toc = true
 Use the event feed for actionable insights and operational visibility.
 The guitar strings and improved query language help you drill into infrastructure & compliance automation events quickly to isolate errors.
 
+Permission for the `event:*` action is required to interact with the **Event Feed** page. In order to search on this page the `infra:nodes:list` action is also required.
+
 ## Event Guitar Strings
 
 ![Guitar Strings](/images/docs/guitar-strings.png)
@@ -36,7 +38,7 @@ In the event feed, events of the same type by the same user are grouped. The eve
 
 ## Filtering events
 
-To filter the event feed and event timeline by event type, Chef Infra Server, or Chef Organization, use the search bar. Available event type filters are clients, cookbooks, data bags, environments, nodes, policyfiles, profiles, roles, and scan jobs.
+To filter the event feed and event timeline by event type, Chef Infra Server, or Chef Organization, use the search bar. The search bar is only shown to folks who also have permission on the `infra:nodes:list` action. Available event type filters are clients, cookbooks, data bags, environments, nodes, policyfiles, profiles, roles, and scan jobs.
 To filter, select Event Type, Chef Infra Server or Chef Organization in the search bar and start typing the name. You cannot filter compliance events -- profiles and scan jobs -- by organization or Chef Infra Server. Compliance events are not visible when either of these filters are applied.
 
 ![Event feed filters by server](/images/docs/event-feed-filters-servers.png)
--- a/components/automate-chef-io/content/docs/event-feed.md
+++ b/components/automate-chef-io/content/docs/event-feed.md
@@ -12,7 +12,7 @@ toc = true
 Use the event feed for actionable insights and operational visibility.
 The guitar strings and improved query language help you drill into infrastructure & compliance automation events quickly to isolate errors.
 
-Permission for the `event:*` action is required to interact with the **Event Feed** page. In order to search on this page the `infra:nodes:list` action is also required.
+Users require permission for the `event:*` action to view and interact with *Event Feed*. Filter and search ability in *Event Feed* requires user permission for the `infra:nodes:list` action.
 
 ## Event Guitar Strings
 
@@ -38,7 +38,8 @@ In the event feed, events of the same type by the same user are grouped. The eve
 
 ## Filtering events
 
-To filter the event feed and event timeline by event type, Chef Infra Server, or Chef Organization, use the search bar. The search bar is only shown to folks who also have permission on the `infra:nodes:list` action. Available event type filters are clients, cookbooks, data bags, environments, nodes, policyfiles, profiles, roles, and scan jobs.
+To filter the event feed and event timeline by Event Type, Chef Infra Server, or Chef Organization, use the search bar. The search bar appears to users with permission for the `infra:nodes:list` action. Available event type filters are clients, cookbooks, data bags, environments, nodes, policyfiles, profiles, roles, and scan jobs.
+
 To filter, select Event Type, Chef Infra Server or Chef Organization in the search bar and start typing the name. You cannot filter compliance events -- profiles and scan jobs -- by organization or Chef Infra Server. Compliance events are not visible when either of these filters are applied.
 
 ![Event feed filters by server](/images/docs/event-feed-filters-servers.png)
--- a/components/automate-ui/Makefile
+++ b/components/automate-ui/Makefile
@@ -1,3 +1,5 @@
+PAGE_DIR := pages
+PAGE_FLAGS := --module app
 PAGE_COMPONENT_DIR := page-components
 PAGE_COMPONENT_FLAGS := --module app
 COMPONENT_PREFIX := chef
@@ -60,11 +62,24 @@ license_scout:
 	cd /tmp/license_scout && bundle install --without=development --quiet && echo "license_scout version" && git rev-parse HEAD
 	/tmp/license_scout/bin/license_scout
 
+# These are shortcuts for generating new components in various subdirs.
+# Example:
+#    $ make page/TopNavLanding
+#    ng "generate" "component" "pages/TopNavLanding" "--module" "app"
+#    CREATE src/app/pages/top-nav-landing/top-nav-landing.component.scss (0 bytes)
+#    CREATE src/app/pages/top-nav-landing/top-nav-landing.component.html (30 bytes)
+#    CREATE src/app/pages/top-nav-landing/top-nav-landing.component.spec.ts (679 bytes)
+#    CREATE src/app/pages/top-nav-landing/top-nav-landing.component.ts (310 bytes)
+#    UPDATE src/app/app.module.ts (14252 bytes)
+
 demo/%:
 	$(NG_CMD) generate component $(DEMO_DIR)/demo-$(@F) $(DEMO_FLAGS)
 
 component/%:
 	$(NG_CMD) generate component $(COMPONENT_DIR)/$(@F) $(COMPONENT_FLAGS)
+
+page/%:
+	$(NG_CMD) generate component $(PAGE_DIR)/$(@F) $(PAGE_FLAGS)
 
 page-component/%:
 	$(NG_CMD) generate component $(PAGE_COMPONENT_DIR)/$(@F) $(PAGE_COMPONENT_FLAGS)
--- a/components/automate-ui/e2e/app.e2e-spec.ts
+++ b/components/automate-ui/e2e/app.e2e-spec.ts
@@ -7,7 +7,7 @@ describe('App', () => {
     browser.waitForAngularEnabled(false);
     browser.get('/');
 
-    expect(browser.getCurrentUrl()).toMatch(/http:\/\/localhost:\d+\//);
+    expect(browser.getCurrentUrl()).toMatch(/http:\/\/localhost:\d+\/$/);
   });
 });
 
--- a/components/automate-ui/e2e/app.e2e-spec.ts
+++ b/components/automate-ui/e2e/app.e2e-spec.ts
@@ -3,11 +3,11 @@ import { fakeServer } from './helpers/fake_server';
 import { enableWelcomeModal, disableWelcomeModal } from './helpers/welcome_modal';
 
 describe('App', () => {
-  it('lands on the event feed page', () => {
+  it('starts on the top-level domain (before permission-based routing occurs)', () => {
     browser.waitForAngularEnabled(false);
     browser.get('/');
 
-    expect(browser.getCurrentUrl()).toMatch(/\/event-feed$/);
+    expect(browser.getCurrentUrl()).toMatch(/http:\/\/localhost:\d+\//);
   });
 });
 
@@ -74,7 +74,8 @@ xdescribe('Welcome Modal', () => {
   });
 });
 
-describe('Main Navigation', () => {
+// TODO: all unavailable due to authz introspection guards
+xdescribe('Main Navigation', () => {
   beforeAll(disableWelcomeModal);
 
   xit('redirects to Client Runs', () => {
@@ -83,14 +84,12 @@ describe('Main Navigation', () => {
     expect(browser.getCurrentUrl()).toMatch(/infrastructure\/client-runs/);
   });
 
-  it('redirects to Event Feed', () => {
+  xit('redirects to Event Feed', () => {
     browser.waitForAngularEnabled(false);
     element(by.linkText('Dashboards')).click();
     expect(browser.getCurrentUrl()).toMatch(/dashboards\/event-feed/);
   });
 
-  // TODO: this now depends on authz introspection -- it's hidden in the e2e
-  // situation, it seems
   xit('redirects to Admin', () => {
     browser.waitForAngularEnabled(false);
     element(by.linkText('Admin')).click();
--- a/components/automate-ui/src/app/app-routing.module.ts
+++ b/components/automate-ui/src/app/app-routing.module.ts
@@ -55,6 +55,7 @@ import {
 
 // Other
 import { SettingsLandingComponent } from './pages/settings-landing/settings-landing.component';
+import { TopNavLandingComponent } from './pages/top-nav-landing/top-nav-landing.component';
 
 const routes: Routes = [
   {
@@ -63,8 +64,8 @@ const routes: Routes = [
     canActivate: [ChefSessionService],
     children: [{
       path: '',
-      redirectTo: 'dashboards/event-feed',
-      pathMatch: 'full'
+      pathMatch: 'full',
+      component: TopNavLandingComponent
     },
     {
       path: 'settings',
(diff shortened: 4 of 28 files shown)


components/automate-ui/Makefile
```diff
@@ -60,12 +62,25 @@ license_scout:
 	cd /tmp/license_scout && bundle install --without=development --quiet && echo "license_scout version" && git rev-parse HEAD
 	/tmp/license_scout/bin/license_scout
 
+# These are shortcuts for generating new components in various subdirs.
```
msorens, line 65: PR bonus: Added a new convenience helper in the Makefile--and gave a concrete example so folks will know how to use these! 😁

components/automate-ui/e2e/app.e2e-spec.ts
```diff
@@ -74,7 +74,8 @@ xdescribe('Welcome Modal', () => {
   });
 });
 
-describe('Main Navigation', () => {
+// TODO: all unavailable due to authz introspection guards
+xdescribe('Main Navigation', () => {
```
msorens, line 78: Probably covered enough with cypress so I did not think it was worth digging into these.

components/automate-ui/src/app/app-routing.module.ts
```diff
@@ -63,8 +64,8 @@ const routes: Routes = [
     canActivate: [ChefSessionService],
     children: [{
       path: '',
-      redirectTo: 'dashboards/event-feed',
-      pathMatch: 'full'
+      pathMatch: 'full',
+      component: TopNavLandingComponent
```
msorens, line 68: Change from static to dynamic routing, using the new component.

components/automate-ui/src/app/components/landing/landing.component.ts
```diff
@@ -44,9 +48,14 @@ export class LandingComponent implements OnInit {
...
-  private authComponent: AuthorizedComponent;
-  private index = 0;
+  /**
+   * A callback function to perform some actions in the event that no landing page was found.
+   * If no callback is provided, the caller guarantees one will always be found.
+   * (That should be the case for sidebars because we will never attempt to
+   * materialize a sidebar unless at least one item is permissible.)
+   * Thus, this defaults to a no-op function.
+   */
+  @Input() onNotFound: () => void = () => { };
```
msorens, line 58: Not nearly as impressive as it looks. It is just a standard variable: `<name>: <type> = <initial value>`, but the type is a function `() => void` so the initial value is too `() => { }`.

components/automate-ui/src/app/components/landing/landing.component.ts
```diff
@@ -69,6 +78,8 @@ export class LandingComponent implements OnInit {
       this.router.navigate(this.routeToSegmentList(this.route), { replaceUrl: true });
     } else if (++this.index < this.routePerms.length) {
       this.setPerms();
+    } else {
+        this.onNotFound();
```
msorens, line 82: The loop above this checks each permission in turn. If we run out of permissions to check we will eventually get here. Because the optional `onNotFound` input defaults to a no-op function, no additional `if` is needed here to see if the user supplied it or not.

components/automate-ui/src/app/entities/layout/layout.facade.ts
```diff
@@ -113,6 +113,13 @@ export class LayoutFacadeService {
     this.layout.userNotifications.display = false;
   }
 
+  showFullPagePlusTopBar(): void {
+    this.contentHeight = '100vh';
+    this.layout.header.display = true;
+    this.layout.sidebar.display = false;
```
msorens, line 119: Almost no work to hide the sidebar; just flip this switch.

components/automate-ui/src/app/page-components/navbar/navbar.component.html
```diff
@@ -5,28 +5,49 @@
...
       </div>
       <div class="navigation-menu">
-        <div role="menuitem" class="nav-link"><a routerLink="/dashboards/event-feed" routerLinkActive="active" tabindex="0">Dashboards</a></div>
+        <app-authorized [allOf]="[
+            ['/api/v0/eventfeed', 'get'],
+            ['/api/v0/eventstrings', 'get'],
+            ['/api/v0/event_type_counts', 'get'],
+            ['/api/v0/event_task_counts', 'get'],
+            ['/api/v0/cfgmgmt/suggestions', 'get']]">
+          <div role="menuitem" class="nav-link"><a routerLink="/dashboards/event-feed" routerLinkActive="active" tabindex="0">Dashboards</a></div>
```
msorens, line 14: The only substantive change in this file is adding `<app-authorized>` to this element.
All the remaining changes are reformatting one-per-line for clarity.

components/automate-ui/src/app/pages/client-runs/client-runs.component.html
```diff
@@ -1,11 +1,11 @@
 <div class="content-container">
   <div *ngIf="permissionDenied$ | async" class="permission-denied-banner">
-    <p>You do not currently have permission to view this page.</p>
-    <p>Please contact your administrator for permission.</p>
+    <p>You do not currently have permission to view this page.
+    Please contact your administrator for permission.</p>
```
msorens, line 4: Tweaks needed due to my changes in styles.scss.

components/automate-ui/src/app/pages/top-nav-landing/top-nav-landing.component.html
```diff
@@ -0,0 +1,14 @@
+<app-landing [routePerms]="routeList" [onNotFound]="SetPageCoverage"></app-landing>
```
msorens, line 1: "Call" the super-class to do permission-based routing.
The rest of the content of the page is what displays when there are no permissions--because we have not routed anywhere else.

components/automate-ui/src/app/pages/top-nav-landing/top-nav-landing.component.ts
```diff
@@ -0,0 +1,65 @@
...
+        ['/apis/iam/v2/policies', 'get', ''],
+        ['/apis/iam/v2/roles', 'get', ''],
+        ['/apis/iam/v2/projects', 'get', '']
+      ], route: '/settings'
+    }
+  ];
+
+  constructor(public layoutFacade: LayoutFacadeService) { }
+
+  SetPageCoverage = () => this.layoutFacade.showFullPagePlusTopBar();
```
msorens, line 64: We invoke this when the permission-based routing fell through without success--in this case, all we want to add is to hide the sidebar.

susanev:
@msorens i sent you a message on slack, but please match the designs that i did for the page when a user has no permissions. i only know of the event feed having the overlay, and ideally, we could just do what all the other pages do, display blank (except for heading and subheading) along with an authorization failure notification on the event feed as well. that is until we have a specific you are not authorized page to send them to.

susanev:
i guess im now seeing this overlay on client runs, yah i really was hoping to get rid of this overlay.

components/automate-ui/src/app/components/landing/landing.component.ts
SEAjamieD: Why is `onNotFound` necessary when there's no callback?  How come you can't just let it pass through without being acted on?

msorens: `onNotFound` is the callback--but it is an optional parameter. It is set to do nothing by default but callers may override that by providing a value, such as is done here:
https://github.com/chef/automate/blob/07172b329eb8950b6c9b8f82833ac5de803b5b4e/components/automate-ui/src/app/pages/top-nav-landing/top-nav-landing.component.html#L1

msorens:
@susanev Regarding using your color in an orderly fashion:
Your design indicates #212334, which is used in just 3 things so far within automate-ui and it is used as a literal, not as a named color. 

See if this makes sense: That color seems very close to #26293A, which exists as a named base color (`$charcoal') and that in turn as a chef color (`$chef-darkest`). Now `$chef-darkest` is used in only two places. So how about if we get rid of that, by changing `$charcoal` to be your specified #212334, then I can replace the literal uses of it with `$chef-darkest`? Let's chat if this is not clear. 😁

susanev:
@msorens sorry, this issue was written a very long time ago when we were still using hsla and so stuff is a bit off. the correct hex is #222435 which is the variable `chef-primary-dark: $ebony-clay;`

components/automate-ui/e2e/app.e2e-spec.ts
bcmdarroch: Since we appear to be skipping all these tests why don't we delete the file?

components/automate-ui/src/app/pages/top-nav-landing/top-nav-landing.component.ts
```diff
@@ -0,0 +1,65 @@
...
+import { LayoutFacadeService } from 'app/entities/layout/layout.facade';
+
+@Component({
+  selector: 'app-top-nav-landing',
+  templateUrl: './top-nav-landing.component.html',
+  styleUrls: []
+})
+export class TopNavLandingComponent {
+
+  // order must match layout in navbar.component.html
```
bcmdarroch, line 12: is there a possible refactor so that we only save these checks in one place?

msorens: Great question! For the sidebars, yes, I am going to do that. For the top nav bar, we will be able to do that when the top bar gets migrated into the sidebar.

components/automate-ui/src/app/components/landing/landing.component.ts
SEAjamieD: thanks!

components/automate-ui/e2e/app.e2e-spec.ts
```diff
@@ -3,11 +3,11 @@ import { fakeServer } from './helpers/fake_server';
 import { enableWelcomeModal, disableWelcomeModal } from './helpers/welcome_modal';
 
 describe('App', () => {
-  it('lands on the event feed page', () => {
+  it('starts on the top-level domain (before permission-based routing occurs)', () => {
     browser.waitForAngularEnabled(false);
     browser.get('/');
 
-    expect(browser.getCurrentUrl()).toMatch(/\/event-feed$/);
+    expect(browser.getCurrentUrl()).toMatch(/http:\/\/localhost:\d+\//);
```
srenatus, line 10: I take it we don't know the port. However, can we somehow say where this ends? I.e. add a `$` anchor?

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
```diff
@@ -254,17 +254,15 @@ export class LayoutSidebarService {
               this.setVisibleValuesToObservables(menuItem);
             });
           }
-          menuGroup.visible$ = menuGroup.visible$ ? menuGroup.visible$ : new BehaviorSubject(true);
+          menuGroup.visible$ = menuGroup.visible$ || new BehaviorSubject(true);
```
srenatus, line 257: I don't follow. If `menuGroup.visible$` is `true`, this will set it to `true`. If it's `false`, this will also set it to `true`. What am I missing?

components/automate-ui/src/app/pages/client-runs/client-runs.component.ts
```diff
@@ -164,7 +164,7 @@ export class ClientRunsComponent implements OnInit, OnDestroy {
   nodeCounts$: Observable<NodeCount>;
 
   // Does the request have the correct permissions
-  permissionDenied$: Observable<boolean>;
+  permissionDenied$: Observable<boolean>; // not currently used
```
srenatus, line 167: Why keep it then?

components/automate-ui/src/app/pages/top-nav-landing/top-nav-landing.component.ts
```diff
@@ -0,0 +1,65 @@
...
+  selector: 'app-top-nav-landing',
+  templateUrl: './top-nav-landing.component.html',
+  styleUrls: ['./top-nav-landing.component.scss']
+})
+export class TopNavLandingComponent {
+
+  // order must match layout in navbar.component.html
+  public routeList: RoutePerms[] = [
+    {
+      allOfCheck: [
```
srenatus, line 15: Since we're doing microoptimizations, can we ditch the third element that always is `''`?

msorens:
After reasonably trying a policy with just `event:*`, @susanev found she was getting the "no perms page". The reason was that the permissions I encoded include those but also require `infra:nodes:list` for the suggestions API (/api/v0/cfgmgmt/suggestions). After discussion, we want to just remove /api/v0/cfgmgmt/suggestions from the required permissions, and adjust the page to wrap the search bar in `<app-authorized>` so it only appears if permissible. Then `event:*` will suffice to access the page.

components/automate-ui/src/app/pages/top-nav-landing/top-nav-landing.component.scss
```diff
@@ -0,0 +1,20 @@
...
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  background: rgba($chef-white, 0.85);
+  z-index: 190;
+  padding-top: 77px;
+  text-align: center;
+  font-weight: bold;
```
susanev, line 13:
```suggestion
  font-weight: normal;
```

components/automate-chef-io/content/docs/event-feed.md
```diff
@@ -12,6 +12,8 @@ toc = true
 Use the event feed for actionable insights and operational visibility.
 The guitar strings and improved query language help you drill into infrastructure & compliance automation events quickly to isolate errors.
 
+Permission for the `event:*` action is required to interact with the **Event Feed** page. In order to search on this page the `infra:nodes:list` action is also required.
```
mjingle, line 15: @msorens @susanev Is the `event:*` action required to _interact_ or to _view_ tthe Event Feed page?

susanev: yes

mjingle: So then, the answer would be both "view and interact."

components/automate-ui/src/app/entities/layout/layout-sidebar.service.ts
msorens: `menuGroup.visible$` is an `Observable<boolean>`, not `boolean`. This is setting the observable to a "fixed" observable of 'true' in the event that the observable is not defined. (It is only defined for things where it needs to be able to vary (is it IAM v2?, is it a feature flag?, and that ilk).

components/automate-ui/src/app/pages/client-runs/client-runs.component.ts
msorens: Sounded like it might be coming back soon, though rendered slightly differently. 🤷

components/automate-ui/src/app/pages/top-nav-landing/top-nav-landing.component.ts
msorens: Good point; done.

components/automate-chef-io/content/docs/event-feed.md
susanev: i guess. i think this is a little different than editor or viewer though, i dont know if that matters. there is nothing to edit on the event feed page.

PR #3622, opened 2020-05-08
Add compliance:reports:export to both viewer roles

tylercloke:
```
     3 | viewer            | Viewer            | chef-managed | {compliance:reports:export,infra:infraServers:list,infra:infraServers:get,secrets:*:get,secrets:*:list,infra:nodes:get,infra:nodes:list,infra:nodeManagers:get,infra:nodeManagers:list,compliance:*:get,compliance:*:list,event:*:get,event:*:list,ingest:*:get,ingest:*:list,iam:projects:list,iam:projects:get,applications:*:get,applications:*:list}
     7 | compliance-viewer | Compliance Viewer | custom       | {compliance:reports:export,compliance:*:get,compliance:*:list}
```

### DIFF
--- a/components/authz-service/storage/postgres/migration/sql/78_update_viewer_role_for_compliance.up.sql
+++ b/components/authz-service/storage/postgres/migration/sql/78_update_viewer_role_for_compliance.up.sql
@@ -1,0 +1,5 @@
+UPDATE iam_roles
+    SET actions = '{compliance:reports:export}' || actions
+    WHERE
+        (id='compliance-viewer' or id='viewer') AND
+        NOT actions @> '{compliance:reports:export}';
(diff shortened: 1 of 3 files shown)


components/authz-service/storage/postgres/migration/sql/78_update_viewer_role_for_compliance.up.sql
```diff
@@ -0,0 +1,5 @@
+UPDATE iam_roles
+    SET actions = '{compliance:reports:export}' || actions
+    WHERE
+        (id='compliance-viewer' or id='viewer') AND
+        NOT actions @> '{compliance:reports:export}';
```
msorens, line 5: Just need a carriage return at the end of the file.

blakestier:
Ok so here's the problem: if `compliance-viewer` is not a chef-managed role, we should not be touching it after we've created it. We only created it conditionally on the role _not_ already existing, so we're also potentially updating a role that a customer _created_ but we didn't know about.

susanev:
added do not merge, cause we are trying to figure out if we can change the action instead

susanev:
fixed in #3623

PR #3629, opened 2020-05-11
remove nodes with a state of terminated 

vjeffrey:
when a user adds an aws-ec2 or azure-vm manager, we query the aws/azure api occasionally to get an update on the state of the node (running, stopped, terminated)
in a world of ephemeral nodes, leaving terminated
nodes around bloats the list of nodes with a bunch of irrelevant data
about the terminated nodes. 

this change deletes the terminated nodes when polling for state

### :+1: Definition of Done
nodes with a state of terminated are deleted from the nodemanager db

rebuild nodemanager
create a node in aws
create aws-ec2 integration for same aws acct
see node in list
terminate node
after polling interval (you can change to something small in code to speed this up), note node no longer exists in nodemanager db

- [x] Tests added/updated?
- [x] Docs added/updated?


rickmarry:
If we are going to use the nodemanager db for a "one node view" page in the UI, would removing these teminated nodes create a mismatch in stats numbers between what is show in one node view vs what is reported from compliance?  We won't be deleting these nodes from the compliance reports stored in elasticsearch..

vjeffrey:
hmm, ya that's a good point rick. we could instead just filter them from the view (one backend or frontend), and then give customers the api call to be able to see all terminated nodes. @krasnow - thoughts? 
we might just want to close this pr until we get more product/ux input on this view

vjeffrey:
closing this until we can get more guidance on desired behaviour

PR #3630, opened 2020-05-11
add manager type to nodemanager process node flow

vjeffrey:
this change adds the manager type to the nodemanager process node flow, meaning
- we can assign any node which is coming through ingest service the manager "chef" since we know that node is managed by chef
- any node that is automate or aws-ec2 or azure-api etc will get the type of the processnode message, meaning that we dont have to look up the type of the manager based on the id and save an extra db call

i also added a format tags bit. we have  a troublesome tag example in the chef-load codebase used for testing, and it seems we fail to list nodes when this troublesome tag exists in the db, because we choke on the json (the tag value in question is `Dot.Comma,Big;\"Trouble`)

### :+1: Definition of Done
manager type is on process node message
nodes managed by chef are labelled as such

rebuild compliance, ingest, nodemanager
load some nodes (chef_load_nodes, load_compliance_reports, chef_load_compliance_nodes)
see the manager type for the nodes is correct!

- [x] Tests added/updated?
- [n/a] Docs added/updated?


components/ingest-service/pipeline/publisher/node_manager_publisher.go
```diff
@@ -98,6 +98,7 @@ func gatherInfoForNode(msg message.ChefRun) (*manager.NodeMetadata, error) {
 			EndTime: timestamp,
 			Status:  status,
 		},
+		ManagerType: "chef",
```
lancewf, line 101: To have the managed type of "Chef" does that include Effortless/Chef Solo Runs?

vjeffrey: right now this code is just putting "chef" on anything coming through the ingest service -- so ya those would also get "chef" as the manager

lancewf:
I have to think about this more but this test is ensuring we are only filtering on nodes with an empty Manager. Where before an empty manager meant it was a client runs node. 
https://github.com/chef/automate/blob/f8ef8ff24ff685db88d75899d22cf563982d32a1/components/nodemanager-service/tests/project_filtering_test.go#L380

lancewf:
Here is the code where we only filter nodes that have an empty manager. 
https://github.com/chef/automate/blob/099a38e60c4617ca0d54fc358bc46b11fcca8653/components/nodemanager-service/pgdb/filtering.go#L279

lancewf:
I just confirmed locally that this does break the project filtering on the nodemanager APIs for client run nodes.

vjeffrey:
oof im glad you thought of testing that!

vjeffrey:
i also need to update the ui call for scan job nodes, bc right now it is excluding nodes with empty string manager, which means all the chef nodes would show up there without further changes

components/nodemanager-service/server/nodes/server.go
```diff
@@ -316,7 +316,7 @@ func (srv *Server) validateNodeUpdate(ctx context.Context, in *nodes.Node) error
 	switch node.Manager {
 	case "aws-api", "azure-api", "gcp-api":
 		return &errorutils.InvalidError{Msg: fmt.Sprintf("invalid option. unable to update %s node", in.Manager)}
-	case "":
+	case "", "chef":
```
vjeffrey, line 319: this ensure we cannot update the name or tags of an ingested node

components/nodemanager-service/pgdb/nodes.go
```diff
@@ -690,7 +690,7 @@ func validateNodeFilters(filters []*common.Filter) error {
 func isValidNodeManagerType(item string) bool {
 	switch item {
 	// empty string covers legacy nodes (nodes created before we started associating all manual nodes with "automate" as a manager and nodes brought over from a1)
-	case "aws-ec2", "aws-api", "azure-api", "azure-vm", "automate", "":
+	case "aws-ec2", "aws-api", "azure-api", "azure-vm", "automate", "gcp-api", "chef":
```
vjeffrey, line 693: should've added gcp-api a long time ago!

components/nodemanager-service/pgdb/node-managers-process-node.go
```diff
@@ -127,20 +127,12 @@ func (db *DB) ProcessIncomingNode(node *manager.NodeMetadata) error {
 		return errors.Wrap(err, "ProcessIncomingNode unable to marshal projects data")
 	}
 
-	var mgrType string
-	if node.ManagerId != "" {
```
vjeffrey, line 131: now that the type is included in the report, we dont have to look up the type in the db

components/compliance-service/inspec-agent/runner/runner.go
```diff
@@ -521,6 +521,7 @@ func (t *InspecJobTask) reportIt(ctx context.Context, job *types.InspecJob, cont
 	report.SourceRegion = job.TargetConfig.TargetBaseConfig.Region
 	report.SourceAccountId = job.SourceAccountID
 	report.AutomateManagerId = job.ManagerID
+	report.AutomateManagerType = job.ManagerType
```
vjeffrey, line 524: this is where we stick the manager type on the report

components/compliance-service/ingest/pipeline/publisher/node_manager_publisher.go
```diff
@@ -90,6 +91,10 @@ func gatherInfoForNode(in message.Compliance) (*manager.NodeMetadata, error) {
 		tags = append(tags, &common.Kv{Key: "environment", Value: in.Report.GetEnvironment()})
 	}
 
+	mgrType := in.Report.GetAutomateManagerType()
+	if in.Report.GetSourceFqdn() != "" {
```
vjeffrey, line 95: get the type on the report, but if the sourcefqdn is not empty string then we know it's managed by chef

vjeffrey:
closing this til i can get back to it

PR #3615, opened 2020-05-07
automate-dex: bump core/dex (2.19.0 -> 2.23.0)

srenatus:
### :nut_and_bolt: Description: What code changed, and why?

This is the first step for #3611. 

⚠️ I haven't verified LDAP. We still don't do that in CI.

### :chains: Related Resources

#3611 

### :+1: Definition of Done

All still works as expected: login as local user, login using SAML, session refresh for local user.

### :athletic_shoe: How to Build and Test the Change

- deployinate, `rebuild components/automate-dex`

### :white_check_mark: Checklist

- [ ] I have read the [CONTRIBUTING document](https://github.com/chef/automate/blob/master/CONTRIBUTING.md).
- [ ] Tests added/updated?
- [ ] Docs added/updated?
- [ ] All commits have been signed-off for the [Developer Certification of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

Issue #3611, opened 2020-05-06
expose dex's allowedGroups config option


bcmdarroch:
## User Story

Dex recently added a feature that allows specifying certain SAML groups to allow. Exposing this will allow customers to reduce the number of groups passed around in requests.

## Definition of Done
- [ ] user can set config/submit a patch that specifies which groups should be allowed to authenticate and returned in dex's claims
- [ ] document it

## Demo Script / Repro Steps
```
[dex.v1.sys.connectors.saml]
groups_attr = "groups"
allowed_groups = ["IT", "ChefAdmins"]

chef-automate config patch saml.toml
```

## Reference
https://github.com/dexidp/dex/pull/1544

srenatus:
https://github.com/habitat-sh/core-plans/pull/3311 the plan is up to date ☑️

srenatus:
> Dex recently added a feature that allows specifying certain SAML groups to allow. Exposing this will allow customers to reduce the number of groups passed around in requests.

So... it just occurred to me that exposing the setting alone will **not** unbreak A2 for people with too many (SAML) groups. See [this test case](https://github.com/dexidp/dex/pull/1544/files#diff-da1bbbf416bca54596265e6dddd87d78R103-R119), and how `ident` is **not changed** in [the relevant code section](https://github.com/dexidp/dex/commit/285c1f162e19d9dd718e55dfde192ec285444a55#diff-3db4f0a48da270d2a99dd82c00574dadR425-R434).

Hence exposing this setting will allow people to restrict the set of users that can login via SAML, to those being a member of the right groups, **only**. Any groups that are included in the SAML response for the user will be included in the ID token. It's not the equivalent of the LDAP group filter. 

To really subset the incoming groups to a few, another code change to dex would be required.

srenatus:
Thinking about this more, what it gives us is *a second way* to restrict access to (SAML) users, which would be bypassing our IAM system. The same thing could be achieved by having a policy that grants access only to the chosen "allowed" SAML groups, I think.

If that assessment is correct, I'd rather not expose this config setting. At least not given the current code, non-filtering. I'll look into what it would take to filter upstream (dex).

srenatus:
- [ ] https://github.com/dexidp/dex/pull/1704
- [ ] another release with that commit in it
- [ ] adapt the dex config to include the flag together with `allowedGroups`

PR #3633, opened 2020-05-11
auth: filter SAML groups

### LINKED ISSUE #3668: LDAP login sometimes fails when there is a session for a user but no associated reference in refresh_token_table
## Describe the bug

> I have a customer running into issues with trying to login using LDAP with automate. It looks like it's accepting the credentials, but then giving an error "failed to delete refresh token: not found", then returns a 500 error to the browser. 

We've seen this bug pop up for at least four customers. The main commonality is that LDAP is configured.

While we don't understand why yet, it's apparently possible for a session to be created for an LDAP user whose user ID wasn't added to the `refresh_token table`. On subsequent logins, the user can't log in because it finds the session entry but no refresh_token entry.

## To Reproduce
We've so far been unable to reproduce the circumstances that caused the customer to not have an entry in the refresh_token table in dex.

However, you can simulate the state they're in by removing entries directly from the table.

```
chef-automate dev psql dex
delete from offline_session where user_id="some-user-id";
```

## Expected behavior
LDAP users with a session should always have an associated entry in the `refresh_token table`

We probably need some garbage collection. It'd be ideal to get a fix upstream in dex.

- Automate Build Number: 
  CLI Build: 20200504210439
  Server Build: 20200506172513

## Additional context
- relevant dex code: https://github.com/dexidp/dex/blame/f2590ee07dbe6a9cd98179ea467255fab41466f4/server/handlers.go#L821-L834
- ticket: https://getchef.zendesk.com/agent/tickets/23553

srenatus:

We want to unblock SAML users with too many groups.

ℹ️ This branch includes #3615 and ~should rebased after that one is merged.~ **replaces it**

~🚧 Waiting for `core/dex` to be bumped to 2.24.0: https://github.com/habitat-sh/core-plans/pull/3366~ ✔️ 

~⚠️ **Needs docs** ⚠️~ 

#3611
fixes #3668 

### :+1: Definition of Done

You can configure the list of groups, it'll be restricting the groups of users able to login using SAML.

- `rebuild components/automate-dex`
- `rebuild components/automate-deployment`
- `rebuild components/automate-cli`

Login using SAML, note the logs:
```
[2020-05-11T09:51:40+00:00] authz-service.default(O): time="2020-05-11T09:51:40Z" level=info msg="Projects Authorized Query" action="event:events:list" projects="[]" resource="event:events" result="[]" subject="[team:saml:Everyone team:saml:devteam team:saml:All Developers user:saml:srenatus]"
```

Apply this config:
```
[dex.v1.sys.connectors.saml]
allowed_groups = ["Everyone"]
```

```
[16][default:/src:0]# chef-automate config patch saml_groups.toml

Updating deployment configuration

Applying deployment configuration
  Started automate-dex
Success: Configuration patched
```

and notice that **upon re-signin**, the list of user groups will be only `team:saml:Everyone`:

```
[2020-05-14T11:20:32+00:00] authz-service.default(O): time="2020-05-14T11:20:32Z" level=info msg="Projects Authorized Query" action="system:license:get" projects="[]" resource="system:status" result="[*]" subject="[team:saml:Everyone user:saml:srenatus]"
```

- [X] Tests added/updated?
- [X] Docs added/updated?
- [X] All commits have been signed-off for the [Developer Certification of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).


srenatus:
I'll add a docs section for this tomorrow (if noone beats me to it).

srenatus:
@bcmdarroch good test. I've just verified that setting it to `["Everyone"]` lets me sign in, and makes me only be member of `team:saml:Everyone` (see PR preamble). I'll look into two allowedGroups now.

srenatus:
```
$ grep -iC3 group /hab/svc/automate-dex/var/etc/config.yml
      entityIssuer: https://automate.chef.co/dex/callback
      usernameAttr: email
      emailAttr: email
      groupsAttr: groups
      allowedGroups: [Everyone the other team]
      filterGroups: true
      nameIDPolicyFormat: urn:oasis:names:tc:SAML:2.0:nameid-format:persistent
```
lol so that's _one_ issue. Something wonky happens from the habitat config to dex's yaml file.

srenatus:
@bcmdarroch thinking about it, that's the same issue you've discovered: If `["Everyone", "devteam"]` ends up in the yaml as `[Everyone devteam]`, which I believe means `["Everyone devteam"]`.

srenatus:
✔️ Fixed. I didn't realise at first that `components/automate-dex/habitat/templates/config.yml` was using golang's text/template.

srenatus:
@bcmdarroch thanks again for checking so diligently. 💯

components/automate-chef-io/content/docs/saml.md
```diff
@@ -82,6 +83,14 @@ The `groups_attr` setting is optional. However, if it is not provided,
 users authenticating via SAML will not be members of any teams.
 {{% /warning %}}
 
+When `allowed_groups` is set, only users with at least on of the configured
```
susanev, line 86:
```suggestion
When `allowed_groups` is set, only users with at least one of the configured
```

components/automate-chef-io/content/docs/saml.md
```diff
@@ -82,6 +83,14 @@ The `groups_attr` setting is optional. However, if it is not provided,
...
+groups are able to sign into Chef Automate using SAML.
```
susanev, line 87:
```suggestion
groups are allowed to sign into Chef Automate using SAML.
```

idk, is that better?

api/config/dex/config_request_test.go
```diff
@@ -369,6 +370,23 @@ func TestValidate(t *testing.T) {
 		})
 	})
 
+	t.Run("allowed_groups (saml)", func(t *testing.T) {
+		valid := map[string][]string{
+			"one group":  {"admins"},
+			"two groups": {"admins", "ops"},
+			"no groups":  {}, // should this warn?
```
bcmdarroch, line 377: I think throwing an error or warning if they try to set empty groups is a good idea 👍

components/automate-chef-io/content/docs/saml.md
```diff
@@ -82,6 +83,14 @@ The `groups_attr` setting is optional. However, if it is not provided,
 users authenticating via SAML will not be members of any teams.
 {{% /warning %}}
 
+When `allowed_groups` is set, only users with at least one of the configured
+groups are allowed to sign into Chef Automate using SAML.
+Furthermore, all of the user's groups that are _not_ in the list are discarded,
+and not available to Chef Automate.
```
msorens, line 89: I am not seeing how this is accomplished in your code changes, or is it just a fact of the SAML handler that we are documenting here...?

srenatus: The relevant code change was upstream, we make use of it by changing our config.

srenatus:
@msorens the issue Brenna discovered was related to how the config is rendered. https://github.com/chef/automate/commit/95311f58507c786f725e7cd712e07a37fb16d9a7 changed that, without any need to change the applied config.

components/automate-chef-io/content/docs/saml.md
srenatus: The relevant code changes upstream are https://github.com/dexidp/dex/pull/1544 and https://github.com/dexidp/dex/pull/1704

api/config/dex/config_request_test.go
srenatus: Warnings aren't there, but we can make it error. Thinking about it, it seems worthwhile. :+1:

srenatus: OK, turns out `allowed_groups = []` is treated the same way as *unset*. So, the user cannot set this. I thus don't think we'll need an error. It cannot happen. I'll merge this 😃

srenatus:
💡 just realised I cannot merge this. It needs @chef/docs-team input on the change, I think.

PR #3647, opened 2020-05-12
Add carbon grid system

scottopherson:
This commit adds the [carbon grid](https://www.carbondesignsystem.com/guidelines/2x-grid/implementation/) scss package to the project. The grid classNames (`.bx--row`, `.bx--col`, etc) are globally available across the app's templates. Components that need access to the grid system variables and mixins need to import them from where they are stored in `@carbon/layout`:

```scss
@import "@carbon/layout/scss/layout";

:host {
  margin: $carbon--grid-gutter;

  @include carbon--breakpoint(lg) {
    margin: 0;
  }
}
```

### DIFF
--- a/components/automate-ui/package-lock.json
+++ b/components/automate-ui/package-lock.json
@@ -1861,6 +1861,25 @@
         "to-fast-properties": "^2.0.0"
       }
     },
+    "@carbon/grid": {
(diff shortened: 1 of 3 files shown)


vjeffrey:
i think buildkite is stuck -- usually a force push helps with this

Issue #3375, opened 2020-04-15
Keyboard controls not present in filter tabs


SEAjamieD:
## Describe the bug
Missing keyboard navigation in tab/card filtering based components listed in
applications/service-groups
compliance/reports/nodes
compliance/reports/profiles
settings/node-integrations/add

## To Reproduce
navigate to any/all of the URLs listed above
begin tabbing through the components, in each you will see a big chunk of the clickable elements that filter, are not keyboard accessible

## Expected behavior
When over the components outlined in red (in the screenshot) I should have a focus ring, and be able to click enter to select as filter

## Screenshots
![Screen Shot 2020-04-15 at 2.32.11 PM.png](https://images.zenhubusercontent.com/5d7aa5d5d4249c0001cab607/6a2c8829-1672-43a9-95bd-5b6a7520069d)![Screen Shot 2020-04-15 at 2.32.01 PM.png](https://images.zenhubusercontent.com/5d7aa5d5d4249c0001cab607/e35f6d6e-f0d5-4617-88b3-aaee9b404f24)

## Versions (please complete the following information):
- OS: Mac OS Mojave
- Browser:  chrome

PR #3649, opened 2020-05-12
Keyboard control updates to ChefPhatRadio and ChefStatusFilterGroup elements

SEAjamieD:
Filters in a few different areas of the site were found to not be keyboard accessible.  This branch adds keyboard control support to any elements using `chef-phat-radio` and `chef-status-filter-group`.  

These elements can be found used in 
applications/service-groups
compliance/reports/nodes
compliance/reports/profiles
settings/node-integrations/add

Fixes: https://github.com/chef/automate/issues/3375

### :+1: Definition of Done
User can keyboard tab through each of the options on the pages listed above, and can hit enter to select that option.

in hab studio build components/automate-ui-devproxy && start_all_services
in a separate terminal: cd into components/automate-ui 
run `make update-ui-lib` to update the chef-ui-library
run `make serve`

navigate to the following routes and tab through the filters.  Use enter to select that filter: 
applications/service-groups
compliance/reports/nodes
compliance/reports/profiles
settings/node-integrations/add
infrastructure/client runs/{node}/attributes, resources, run list

- [ ] Tests added/updated?
- [ ] Docs added/updated?

Example After:
Chef Automate (6)

### DIFF
--- a/components/chef-ui-library/src/molecules/chef-phat-radio/chef-phat-radio.tsx
+++ b/components/chef-ui-library/src/molecules/chef-phat-radio/chef-phat-radio.tsx
@@ -47,6 +47,14 @@ export class ChefPhatRadio {
       this.value = this.selected.value;
     }
     this.selected.selected = true;
+
+    // add keypress listeners to all the child options
+    const theseOptions = this.el.querySelectorAll(`chef-option`);
+    theseOptions
+      .forEach(option => {
+        option.addEventListener('keypress', event => { this.handleKeypress(event); });
+        option.setAttribute('tabindex', '0');
+      });
   }
 
   componentDidUpdate() {
@@ -65,6 +73,12 @@ export class ChefPhatRadio {
     const options: HTMLChefOptionElement[] = Array.from(this.el.querySelectorAll('chef-option'));
     options.forEach((opt) => opt.selected = false);
     return options;
+  }
+
+  handleKeypress = (event) => {
+    if (event.key === 'Enter') {
+      this.handleClick(event);
+    }
   }
 
 }
--- a/components/chef-ui-library/src/molecules/chef-phat-radio/chef-phat-radio.tsx
+++ b/components/chef-ui-library/src/molecules/chef-phat-radio/chef-phat-radio.tsx
@@ -50,7 +50,7 @@ export class ChefPhatRadio {
 
     // add keypress listeners to all the child options
     options.forEach(option => {
-        option.addEventListener('keypress', this.handleKeypress);
+        option.addEventListener('keypress', (event) => this.handleKeypress(event));
         option.setAttribute('tabindex', '0');
       });
   }
--- a/components/chef-ui-library/src/molecules/chef-phat-radio/chef-phat-radio.tsx
+++ b/components/chef-ui-library/src/molecules/chef-phat-radio/chef-phat-radio.tsx
@@ -49,10 +49,8 @@ export class ChefPhatRadio {
     this.selected.selected = true;
 
     // add keypress listeners to all the child options
-    const theseOptions = this.el.querySelectorAll(`chef-option`);
-    theseOptions
-      .forEach(option => {
-        option.addEventListener('keypress', event => { this.handleKeypress(event); });
+    options.forEach(option => {
+        option.addEventListener('keypress', this.handleKeypress);
         option.setAttribute('tabindex', '0');
       });
   }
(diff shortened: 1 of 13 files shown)


components/chef-ui-library/src/molecules/chef-phat-radio/chef-phat-radio.tsx
```diff
@@ -47,6 +47,14 @@ export class ChefPhatRadio {
...
     }
     this.selected.selected = true;
+
+    // add keypress listeners to all the child options
+    const theseOptions = this.el.querySelectorAll(`chef-option`);
+    theseOptions
+      .forEach(option => {
+        option.addEventListener('keypress', event => { this.handleKeypress(event); });
+        option.setAttribute('tabindex', '0');
+      });
```
SEAjamieD, line 57: In order to _not_ target all options across automate, I opted to only select the options that are inside the selected element -> in this case `chef-phat-radio`.

A tab index adds support for keyboard tabbing, and keypress event which routes the same function as the click listener.

components/chef-ui-library/src/molecules/chef-phat-radio/chef-phat-radio.tsx
```diff
@@ -67,4 +75,10 @@ export class ChefPhatRadio {
     return options;
   }
 
+  handleKeypress = (event) => {
+    if (event.key === 'Enter') {
+      this.handleClick(event);
+    }
+  }
```
SEAjamieD, line 82: Stencil wouldn't allow me to add the click listening function straight to the keypress event listener, so I had to loop it around.

SEAjamieD:
> thank you for doing this!! noticed a few more things that'd it be rad to fix too if possible
> 
> * [x]  for both sets on applications, if we can add a few pixels of padding to the top and reduce it from the margin above we can get the focus ring to not be cut off on the top
> * [x]  think we can get the ones on infrastructure > client runs > {node} > attributes too?
> 
> node-attributes-filters

@susan - updated we Resources and Run list inside that infrastructure view as well.  Found some toggles along the way that need updates but will require way more work so will make a separate card for those.

components/chef-ui-library/src/molecules/chef-phat-radio/chef-phat-radio.tsx
scottopherson: Any reason to format this way instead of
```ts
handleKeypress(event) {
   ...
}
```

components/chef-ui-library/src/molecules/chef-status-filter-group/chef-status-filter-group.tsx
```diff
@@ -69,6 +69,7 @@ export class ChefStatusFilterGroup {
 
   selected: HTMLChefOptionElement;
 
+
```
scottopherson, line 72: unneeded space

components/chef-ui-library/src/molecules/chef-phat-radio/chef-phat-radio.tsx
scottopherson: Is `theseOptions` needed? Seems like `options` would contain the same list of `chef-option` nodes, albeit as an `Array` instead of `NodeList`.

components/chef-ui-library/src/molecules/chef-phat-radio/chef-phat-radio.tsx
SEAjamieD: absolutely not - thanks for the catch!

components/chef-ui-library/src/molecules/chef-phat-radio/chef-phat-radio.tsx
```diff
@@ -47,6 +47,14 @@ export class ChefPhatRadio {
       this.value = this.selected.value;
     }
     this.selected.selected = true;
+
+    // add keypress listeners to all the child options
+    const theseOptions = this.el.querySelectorAll(`chef-option`);
+    theseOptions
+      .forEach(option => {
+        option.addEventListener('keypress', event => { this.handleKeypress(event); });
```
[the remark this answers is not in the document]
SEAjamieD: The way you suggested is how I would generally do it.  But what i've found is that in this setup, if I don't specifically pass the event, The element function does not receive it.  I was able clean this up a little bit by removing the curly braces. 

My hunch is that maybe its because this element is nested an extra level down, but I actually cant use `option.addEventListener('keypress', this.handleKeypress);`,  as it won't work.  so I've opted for `option.addEventListener('keypress', event => this.handleKeypress(event));`

components/chef-ui-library/src/molecules/chef-phat-radio/chef-phat-radio.tsx
SEAjamieD: thanks!

Issue #3533, opened 2020-04-29
create token multi-select dropdown issues


susanev:
## overview
when creating a token if a user opens the projects dropdown and then the policies dropdown the policies dropdown appears behind the projects dropdown.

## repro steps
1. navigate to settings > tokens
2. use the create token button
3. open the projects dropdown
4. open the policies dropdown
5. notice that the policies dropdown displays behind the projects dropdown

## what it should be
there's click outside functionality that is supposed to close the dropdown, so if the user opens the projects one first and then clicks outside to open the policies one the projects one should close. something is not working.

SEAjamieD:
I'm moving this back into on deck - After I realized the action that is being looked for, having the last item clicked always remain on top - it will require a little more logic and the desktop needs priority for now

susanev:
@SEAjamieD i wrote this quick and need to think more about it. i think that might not be the right approach. but either way i dont think this is ready for ur work yet, sorry about the miscommunication.

susanev:
@SEAjamieD updated...
>there's click outside functionality that is supposed to close the dropdown, so if the user opens the projects one first and then clicks outside to open the policies one the projects one should close. something is not working.

PR #3652, opened 2020-05-12
Fix logic error that prevented multi-select dropdowns from closing properly

tylercloke:

`toggleDropdown` was calling `closeDropdown`, which set `active` to false, but then `toggleDropdown` flipped the value of `active `for both the open and close case, incorrectly setting `active` back to true after `closeDropdown` was called.

Create a project then go to a multi-select dropdown (say in teams create modal).

Click on the dropdown and it will open.

Now, do any of the following and it should close properly:

* click outside the dropdown.
* click the dropdown header again.
* press enter.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
+++ b/components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
@@ -74,11 +74,10 @@ export class ResourceDropdownComponent implements OnInit, OnChanges {
     if (!this.active) { // opening
       this.filterValue = '';
       this.filteredResources = this.resourcesInOrder;
+      this.active = true;
     } else { // closing
       this.closeDropdown();
     }
-
-    this.active = !this.active;
   }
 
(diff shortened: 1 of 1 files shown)


components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
```diff
@@ -74,11 +74,10 @@ export class ResourceDropdownComponent implements OnInit, OnChanges {
     if (!this.active) { // opening
       this.filterValue = '';
       this.filteredResources = this.resourcesInOrder;
+      this.active = !this.active;
```
msorens, line 77: Because this is now used only in a single case, I think it improves clarity to do this:
```suggestion
      this.active = true;
```
(Just like there is `this.active = false;` on L95.)

susanev:
it seems that #3533 is not fixed with this, so there may still be a click outside functionality issue that can be solved in that other issue. but toggle works great now, thank you!!

PR #3663, opened 2020-05-13
Add the new inspec truncation options

alexpop:
This PR makes use of the latest inspec `reporter_backtrace_inclusion` and `reporter_message_truncation` [options](https://github.com/inspec/inspec/pull/4994) when triggering Compliance scans from Automate.

It also truncates the controls to maximum 50 (default) results per control before sending the report through ingestion. This is to prevent reports that exceed 4MB which are not allowed through grpc.
^ To tests this, you have to trigger a scan with a control that has a loop inside and generates lots of describe blocks.

### DIFF
--- a/components/compliance-service/inspec/cli.go
+++ b/components/compliance-service/inspec/cli.go
@@ -17,6 +17,9 @@ import (
 // BackendCache used for configuring inspec exec command, passed in via config flag
 var BackendCache bool
 
+// ResultMessageLimit used for configuring inspec exec command, passed in via config flag
+var ResultMessageLimit int
+
 // TmpDir is used for setting the location of the /tmp dir to be used by inspec for caching
 var TmpDir string
 
@@ -58,6 +61,8 @@ func Scan(paths []string, target *TargetConfig, timeout time.Duration, env map[s
 	// Using "json-automate" as it provides better inherited profiles report
 	target.Reporter["json-automate"] = Reporter{}
 	target.BackendCache = BackendCache
+	target.ResultIncludeBacktrace = false
+	target.ResultMessageLimit = ResultMessageLimit
 
 	env["HOME"] = os.Getenv("HOME")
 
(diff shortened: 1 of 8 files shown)


susanev:
should something be added to the automate docs about this?

alexpop:
Spot on @susanev
I have another PR that will add a third cli option and was thinking to raise a doc request for all three.

vjeffrey:
do you by any chance have a sample report/control with a bunch of results that ppl can use if they want to test?

components/compliance-service/inspec/cli.go
```diff
@@ -299,6 +304,10 @@ func getInspecError(stdOut string, stdErr string, err error, target *TargetConfi
 		}
 	}
 
+	if err == nil {
```
vjeffrey, line 307: don't we already cover this on line 321 with the return?

alexpop: if err is nil, line 321 will not be reached as the service will crash on line 311.

vjeffrey: ahh ya - 👍

alexpop:
@vjeffrey here's a control that has 102 results (1 passed, 1 skipped, 100 failed):

```ruby
control 'mix control' do
  title 'Mix control to try to get different results'
  impact 1
  describe file("/etc/group") do
    it { should be_file }
  end
  describe apache_conf('path') do
    its('setting_name') { should eq 'value' }
  end
  (1..100).each do |i|
    describe file("/etc/group#{i}") do
      it { should be_file }
    end
  end
end
```

vjeffrey:
thank you!

alexpop:
Victoria, I opened another PR that has another service config option:
https://github.com/chef/automate/pull/3697

Once it merges I'll create a PR for all options introduced recently and then make sure documentation is also in sync.

PR #3664, opened 2020-05-13
Automate-3144 API compliance reports node filters

tarablack01:
When I looked into this, I realized the UI is doing the status filtering (https://github.com/chef/automate/blob/master/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-nodes/reporting-nodes.component.ts#L64).

Filtering by status is supported by backend, and won't result in results being spread across pages, so let's just move to status filtering via api.

https://app.zenhub.com/workspaces/ui-team-5cba23a3b14fdc05970c8f87/issues/chef/automate/3144

### :+1: Definition of Done
Filtering by status is filtering via api.

hab studio enter
export DEVPROXY_URL=localhost
start_automate_ui_background
start_all_services
chef_load_nodes

Screen Shot 2020-05-13 at 11 41 20 AM
Screen Shot 2020-05-13 at 11 41 35 AM

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-nodes/reporting-nodes.component.html
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-nodes/reporting-nodes.component.html
@@ -10,7 +10,7 @@
   <chef-phat-radio
     class="nodes-list-status-filters"
     [value]="nodeFilterStatus"
-    (change)="filterNodeStatus($event, $event.target.value)">
+    (change)="statusFilter($event, $event.target.value)">
     <chef-option class="filter all" value='all'>
       <span class="filter-label">Total Nodes</span>
       <span class="filter-total">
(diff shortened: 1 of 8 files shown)


tarablack01:
@vjeffrey thanks for finding that and also requesting UX team verify the desired result. I know why this is happening as all the filters are processed through the parent reporting component. If this needs to be changed, I can do that. 😊

susanev:
good catch! summary data should not change based on the filters. the filters appear below the summary and within the tabs. so visually folks will not expect the summary data to change as they adjust the filters.

jonong1972:
Yeah this is wonky.  I think I understand the reason why the system is doing what its doing.  If I recall correctly the search was built as if a user were to search on something, that would affect the status?  For instance, show me all nodes with Control X, and then reflect status of nodes based on Control X - which then would update the Summary?  I could be wrong though.

tarablack01:
@vjeffrey @susanev @jonong1972 Fixed filtering to be only nodes.

susanev:
cool thank you!! i think this pr now addresses #3674 too
noticed that when navigating between the tabs that the filters persist, is this what you want @jonong1972?

jonong1972:
@susanev yes. would be great it would persist. But I think in order for that to happen for each respective tab we would probably have to change the query param from something like. "?status=failed" to "?node_status=failed", "?profile_status=failed"

tarablack01:
@jonong1972 @susanev the status will persist. We just need to hook up profiles with api status filtering. 😊

Issue #3660, opened 2020-05-13
update chef servers left nav icon


susanev:
## overview
update the icon in the left nav for chef servers, it should be a C in a div, similar to what was done for waivers in #3089

sachin-msys:
fixed this by PR #3667

PR #3089, opened 2020-03-12
Automate 2546 Waiver

tarablack01:
### :nut_and_bolt: Description: What code changed, and why?
As an Automate User, I will use this work when I want to drill down to the control and see which control was waived.

### :chains: Related Resources
https://github.com/chef/automate/issues/2561

### :+1: Definition of Done
A user can click on "Scan Results" on the Profile list page, and then choose a node - which will then show controls that have been waived.
Waived control will be displayed at the bottom of the control list.

### :athletic_shoe: How to Build and Test the Change
load data by using the load_compliance_reports command from your studio dev env

### :camera: Screenshots, if applicable
Screen Shot 2020-03-12 at 5 25 48 PM
Screen Shot 2020-03-12 at 5 25 56 PM

Screen Shot 2020-03-12 at 4 19 13 PM

PR #3667, opened 2020-05-13
Custom icon added for Chef Server in left side nav

sachin-msys:
Signed-off-by: sachin-msys <sbacchav@msystechnologies.com>

Custom icon added for chef server in left side nav.

#3660 
### :+1: Definition of Done
Custom icon added for Chef Servers in left side nav
1. `git fetch origin Sachin/chef-servers-left-nav-custom-icon`
2. `build components/automate-ui` 
3. Go to `Infrastructure` tab >>  `Chef Servers` its under `Chef-Server` feature flag.

chef-custome-icon-before
chef-custome-icon-after

### DIFF
--- a/components/automate-ui/src/app/components/sidebar-entry/sidebar-entry.component.html
+++ b/components/automate-ui/src/app/components/sidebar-entry/sidebar-entry.component.html
@@ -12,7 +12,9 @@
 <ng-template #sidebarEntryTemplate>
   <chef-icon class="sidebar-icon" *ngIf="icon"
     [style.transform]="'rotate(' + (iconRotation ? iconRotation : 0) + 'deg)'">{{icon}}</chef-icon>
+  <div *ngIf="customIcon === 'chef-servers'" class="chef-server-icon"></div>    
   <span class="text"><ng-content></ng-content></span>
+  <ng-content></ng-content>
   <chef-icon *ngIf="!openInNewPage" class="arrow">play_arrow</chef-icon>
   <chef-icon *ngIf="openInNewPage" class="open-in-new sidebar-icon">launch</chef-icon>
 </ng-template>
--- a/components/automate-ui/src/app/components/sidebar-entry/sidebar-entry.component.html
+++ b/components/automate-ui/src/app/components/sidebar-entry/sidebar-entry.component.html
@@ -12,7 +12,7 @@
 <ng-template #sidebarEntryTemplate>
   <chef-icon class="sidebar-icon" *ngIf="icon"
     [style.transform]="'rotate(' + (iconRotation ? iconRotation : 0) + 'deg)'">{{icon}}</chef-icon>
-  <div *ngIf="customIcon === 'chef-servers'" class="chef-servers-icon"></div>    
+  <div *ngIf="customIcon" [attr.class]="customIcon"></div>   
   <span class="text"><ng-content></ng-content></span>
   <chef-icon *ngIf="!openInNewPage" class="arrow">play_arrow</chef-icon>
   <chef-icon *ngIf="openInNewPage" class="open-in-new sidebar-icon">launch</chef-icon>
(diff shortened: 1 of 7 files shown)


components/automate-ui/src/app/components/sidebar-entry/sidebar-entry.component.html
```diff
@@ -12,6 +12,7 @@
 <ng-template #sidebarEntryTemplate>
   <chef-icon class="sidebar-icon" *ngIf="icon"
     [style.transform]="'rotate(' + (iconRotation ? iconRotation : 0) + 'deg)'">{{icon}}</chef-icon>
+  <div *ngIf="customIcon === 'chef-servers'" class="chef-servers-icon"></div>
```
msorens, line 15: Would prefer if this class did not need to know about its callers (that goes against the Open/Closed Principle -- the "O" in [SOLID](https://en.wikipedia.org/wiki/SOLID_(object-oriented_design))). Perhaps we could just do something like this:
```suggestion
  <div *ngIf="customIcon" [attr.class]="customIcon"></div>    
```

components/automate-ui/src/app/components/sidebar-entry/sidebar-entry.component.ts
```diff
@@ -13,6 +13,8 @@ export class SidebarEntryComponent {
   @Input() iconRotation: number;
   // (optional) if true, opens the link in a new page
   @Input() openInNewPage: boolean;
+  // (optional) shows custom icon
```
msorens, line 16: 👍 Good, but please add a similar line above `icon` on L9, something like:
```
// (optional) if present displays the standard icon
 @Input() icon: string;
```

Also, suggest putting the two declarations next to each other here (`icon` and `customIcon`) and please add a further comment that distinguishes the two. That is, when do I use `icon` and when do I use `customIcon` ? Like the casual reader, I actually do not know the answer as I write this. I would have to dig into the code to understand that--but let's help the reader so they do not have to. 😁 
(And a further question, coming from not yet digging into the code: do we really need both of them?)

components/automate-ui/src/app/components/sidebar-entry/sidebar-entry.component.html
sachin-msys: @msorens nice catch :1st_place_medal: , incorporated :+1:

components/automate-ui/src/app/components/sidebar-entry/sidebar-entry.component.ts
sachin-msys: @msorens yes we need both of them. `icon` is used to show material icon and `customIcon` is used when there is no respective material icon is available. I took reference from Waiver (#3089).

components/automate-ui/src/styles.scss
```diff
@@ -342,6 +342,29 @@ input {
...
+
+.chef-servers-icon:before {
+  color: $chef-darkest;
+  content: 'C';
+  font-size: 9px;
+  font-weight: bolder;
+  position: absolute;
+  right: 4px;
+  text-align: center;
+  top: -3px;
```
SEAjamieD, line 365: [nit] @susanev / @jonong1972 / @apriofrost  - not sure who this falls under.

When I spin this up - it looks off center to me and `top` should be at `-4px` maybe?

sachin-msys: custom-c-4

susanev: yah, noticed this when the waivers one was added as well. ideally id love for them to be centered, but if that isn't easy thats okay too.

SEAjamieD: @sachin-msys - there must be some browser differences in play.  I get the opposite of what you see with those settings.  🤷‍♂️   

No need for this to be a blocker.  I'll toy with it in my spare time soon.

Issue #3651, opened 2020-05-12
data lifecycle ui weirdness


susanev:
## overview
repro here: https://a2-perf-test-single-local-inplace-upgrade-dev.cd.chef.co/settings/data-lifecycle

a few weird things..
- if you change the time type to days for the missing nodes, the save button does not become enabled
- if i change the number value for missing nodes then save, then navigate to data feeds, then back to data lifecycle i see the old value and have to do a hard refresh to see the correct value
ill put this all into an issue, but we may want to try and fix this soon

didnt check other inputs, but we should prolly m ake sure they are working too

[video in internal slack](https://chefio.slack.com/archives/CBSQQBK19/p1589324881493500)

SEAjamieD:
@susan - Locally i'm unable to recreate the stuck value, but I am seeing the days not being able to be selected.  working on all of this now and will keep you updated.

susanev:
ill add a video of what i was seeing to this card in a bit too

susanev:
@stevendanna noticed that the row above the label node as missing was the one getting the value i edited. so i think something is swapped with the remove data and the label nodes as missing. gonna mark this as a high priority because this is a pretty serious bug which can lead to data loss.

SEAjamieD:
PR for save button: https://github.com/chef/automate/pull/3676

PR #3676, opened 2020-05-13
Add change origin to assist form control directive

SEAjamieD:
On the data lifecycle page, users were prevented from selecting "days" as an option in Client Runs : label node as missing, if they hadn't already altered something on the page.

This branch adds the `resetOrigin` property to the formControl inputs, as well as updates them on save, so that the user can save as many times as they want and the formControl will remember their last known value.

Fixes part 2 of: https://github.com/chef/automate/issues/3651

### :+1: Definition of Done
Users can change the values to whatever they like.  If the value is different than before, the 'Save Changes' button will activate.  If the value is the changed back to its previous value prior to saving changes, the 'Save Changes' button will deactivate.

in hab studio `build components/automate-ui-devproxy && start_all_services`
cd into components/automate-ui and run `make serve`

navigate to https://a2-dev.test/settings/data-lifecycle
unless disabled user should be able to change all values as desired and 'Save Changes' button is activated when values are different than previously set.  If they move back to the previous value prior to saving. Save Changes button is disabled again.

One thing to note, I've found that with the Number inputs, if user uses the mouse or arrow keys to alter the numbers, the formControl Directive values are not respected.  This does not prevent the user from sending any values, but we wont see the full value of the form control directive.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

After:
Chef Automate (5)

### DIFF
--- a/components/automate-ui/src/app/pages/automate-settings/automate-settings.component.html
+++ b/components/automate-ui/src/app/pages/automate-settings/automate-settings.component.html
@@ -51,7 +51,7 @@
               </chef-checkbox>
             </div>
             <p>Remove Chef Server actions after
-              <input chefInput type="number" min="0" formControlName="threshold" [resetOrigin]="shouldResetValues"
+              <input chefInput formControlName="threshold" [resetOrigin]="shouldResetValues" type="number" min="0"
                 (keydown)="preventNegatives($event.key)"/>days
             </p>
           </div>
@@ -74,7 +74,7 @@
             </div>
 
             <p>When no health check reports have been received for a service in
-              <input chefInput type="number" min="0" formControlName="threshold" [resetOrigin]="shouldResetValues"
+              <input chefInput formControlName="threshold" [resetOrigin]="shouldResetValues" type="number" min="0"
                 (keydown)="preventNegatives($event.key)" />
               <mat-form-field appearance="none">
                 <mat-select panelClass="chef-dropdown" formControlName="unit" [resetOrigin]="shouldResetValues">
@@ -101,7 +101,7 @@
               </span>
             </div>
             <p>Remove services labeled as disconnected after
-              <input chefInput type="number" min="0" formControlName="threshold" [resetOrigin]="shouldResetValues"
+              <input chefInput formControlName="threshold" [resetOrigin]="shouldResetValues" type="number" min="0"
                 (keydown)="preventNegatives($event.key)" />
               <mat-form-field appearance="none">
                 <mat-select panelClass="chef-dropdown" formControlName="unit" [resetOrigin]="shouldResetValues">
@@ -124,7 +124,7 @@
               </chef-checkbox>
             </div>
             <p>Remove data for Chef Infra Client runs after
-              <input chefInput type="number" min="0" formControlName="threshold" [resetOrigin]="shouldResetValues"
+              <input chefInput formControlName="threshold" [resetOrigin]="shouldResetValues" type="number" min="0"
                 (keydown)="preventNegatives($event.key)" />days
             </p>
           </div>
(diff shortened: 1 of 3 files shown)


components/automate-ui/src/app/pages/automate-settings/automate-settings.component.html
```diff
@@ -141,10 +141,10 @@ <h2>Client Runs</h2>
               <input chefInput 
                 type="number" 
                 min="0" 
-                formControlName="threshold" 
+                formControlName="threshold" [resetOrigin]="shouldResetValues"
                 (keydown)="preventNegatives($event.key)" />
```
msorens, line 145: Suggest rewriting some of the elements so they appear consistent.
1. This `<input>` element is the most egregious example--it puts one attribute per line. All the other ones are on a single line. 
2. Put the attributes in the same order. It just helps lessen the mental effort needed to read the file. For instance, here are the first two in the file where things are in a different order:
```
<input chefInput formControlName="threshold" [resetOrigin]="shouldResetValues" type="number" min="0" 
<input chefInput type="number" min="0" formControlName="threshold" [resetOrigin]="shouldResetValues"
```

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.ts
```diff
@@ -216,8 +219,11 @@ export class AutomateSettingsComponent implements OnInit, OnDestroy {
             this.saving = false;
           } else if (changeConfigurationSelector.status === 'loadingSuccess') {
             this.showSuccessNotification();
-            this.automateSettingsForm.markAsPristine();
             this.saving = false;
+            // After a successful Save, we flip shouldResetValues to true
+            // to update the resetOrigin values
```
msorens, line 224: Consider rewording for clarity (harping on my theme--this is more behavior, less implementation 😁 ):
```suggestion
            // After a successful save, trigger a notification to FormControlDirective
            // to consider the newly saved values as the new "original" values.
```

components/automate-ui/src/app/pages/automate-settings/automate-settings.component.ts
```diff
@@ -375,6 +381,9 @@ export class AutomateSettingsComponent implements OnInit, OnDestroy {
           break;
       }
     });
+    // AfterInit, we need to apply this for resetOrigin to reset the values of formcontrolDirective
```
msorens, line 384:
```suggestion
    // After a successful load of initial values, trigger a notification 
    // to FormControlDirective to treat them as the "original" values.
```

SEAjamieD:
> Code looks reasonable at first glance, but it is not quite working for me:
> ✔️ Upon load Save is disabled
> ✔️ Change a value and Save becomes enabled
> ❌ Change it back--Save does not disable for any number (this does work for "days", "hours", "minutes")
> ✔️ Make and save changes and Save becomes disabled

Thanks for the feedback!

For the number inputs - curious if you saw my comment above so i'm gonna post it here:

One thing to note, I've found that with the Number inputs, if user uses the mouse or arrow keys to alter the numbers, the formControl Directive values are not respected. This does not prevent the user from sending any values, but we wont see the full value of the form control directive.

Curious if these worked for you when using the number keys, as opposed to the mouse or arrow keys.  Seems like a weird behavior.

msorens:
Ah, right you are @SEAjamieD; I missed your note. It does work when typing values. Very odd. So that does put us in a better state, at least. 👍

SEAjamieD:
> Ah, right you are @SEAjamieD; I missed your note. It does work when typing values. Very odd. So that does put us in a better state, at least. 👍

This is peculiar behavior - I went a bit further and learned that this is current behavior in any form if a text input is of type `number`.  `formControlDirective` also does not establish original values if the user uses the arrow pad on their keyboard to make their number choice.  I tested this in a few places by switching the input type to `number` instead of text.  It doesn't appear we've caught this before because nearly all of our inputs are of type `text`.

While I think that it's important for an update to be made for this, I don't feel that it is worthy of holding up this branch for two reasons.

1.  This bug may currently be preventing users from reverting _back_ to a value if it happened to be a default value.
2.  This branch will no longer prevent a user from executing the save they wish.  It will only allow them to execute the save more than once.  I feel the allowing a second save is far better than preventing an initial save.

I propose opening a non-urgent bug ticket to handle the disabling of save button when the user is using a mouse or arrow pad.

Issue #3435, opened 2020-04-20
Infra proxy environment details page view


vinay033:
## User Story
we have to add the particular environment details page view.
View link: https://chef.invisionapp.com/share/EYW6R099CGH#/screens/402071330

PR #3682, opened 2020-05-14
Infra proxy environment details page UI 

vinay033:
Added environment details page UI and tabs.
https://github.com/chef/automate/issues/3435
### :+1: Definition of Done
Environment details page UI changes with the `Cookbook Constraints` and `Attributes` tabs
```
STEP 1
inside the hab studio

STEP 2
open new window
go to automate UI path

$ cd components/automate-ui
and run the command 

npm run serve:hab
```

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### :camera: Screenshots
Environment_details


components/automate-ui/src/app/entities/environments/environment-details.selectors.ts
```diff
@@ -0,0 +1,28 @@
+import { createSelector, createFeatureSelector } from '@ngrx/store';
+import { EnvironmentDetailsEntityState, environmentDetailsEntityAdapter } from './environment-details.reducer';
+import { routeParams } from 'app/route.selectors';
+import { find } from 'lodash/fp';
```
msorens, line 4: Convention is to put 3rd party in the first group, automate in the second group, and within each group from general to specific:
```suggestion
import { createSelector, createFeatureSelector } from '@ngrx/store';
import { find } from 'lodash/fp';

import { routeParams } from 'app/route.selectors';
import { EnvironmentDetailsEntityState, environmentDetailsEntityAdapter } from './environment-details.reducer';
```

components/automate-ui/src/app/entities/environments/environment.model.ts
```diff
@@ -1,3 +1,38 @@
...
+
+export class EnvironmentAttributes {
+  default_attributes: string;
+  override_attributes: string;
+  all: string;
+  constructor(resp: RespEnvironmentAttributes) {
+    this.default_attributes = resp.default_attributes &&
+      JSON.parse(resp.default_attributes) || {};
+    this.override_attributes = resp.override_attributes &&
+      JSON.parse(resp.override_attributes) || {};
```
msorens, line 27: Your line breaks actually imply different precedence than actual precedence; consider adding parentheses and adjusting as shown to disambiguate to the reader.
```suggestion
    this.default_attributes =
      (resp.default_attributes && JSON.parse(resp.default_attributes)) || {};
    this.override_attributes =
      (resp.override_attributes && JSON.parse(resp.override_attributes)) || {};
```

components/automate-ui/src/app/entities/environments/environment.model.ts
```diff
@@ -1,3 +1,38 @@
+import { merge } from 'lodash';
```
msorens, line 1: Try to use only `lodash/fp` when bringing in new functionality. (Consider `lodash` deprecated by convention.)

components/automate-ui/src/app/entities/environments/environment.model.ts
```diff
@@ -1,3 +1,38 @@
...
+  constructor(resp: RespEnvironmentAttributes) {
+    this.default_attributes = resp.default_attributes &&
+      JSON.parse(resp.default_attributes) || {};
+    this.override_attributes = resp.override_attributes &&
+      JSON.parse(resp.override_attributes) || {};
+    this.all = merge(
+      {},
+      this.default_attributes,
+      this.override_attributes);
+  }
```
msorens, line 32: Both `default_attributes` and `override_attributes` are purportedly strings and it does not make sense to use those in a `merge` operation. I suspect it is the declarations of those that are wrong, since `JSON.parse` returns an object not a string. Assuming that is the case and you need to fix the declarations, I think you can even obviate the need for the `merge` operator and simplify with:
```suggestion
    this.all = {...this.default_attributes, ...this.override_attributes};
  }
```
Bonus points for making the same fix in other places that this pattern is used. 😁

components/automate-ui/src/app/modules/infra-proxy/environment-details/environment-details.component.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+    expect(component.tabValue).toBe('attributes');
+    expect(router.navigate).toHaveBeenCalled();
+  });
+
+  describe('AttributesComponent', () => {
+    it('renders the attributes component correctly', () => {
+      expect(element.query(By.css('.jsontree_value_object'))).toBeNull();
+    });
+
+    it('fetches attributes and returns a EnvironmentAttributes object', async(() => {
```
msorens, line 130:
```suggestion
    it('fetches attributes and returns an EnvironmentAttributes object', async(() => {
```

components/automate-ui/src/app/modules/infra-proxy/environment-details/environment-details.component.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+    element = fixture.debugElement;
+    fixture.detectChanges();
+    router = TestBed.inject(Router);
+    component.attributes = new RoleAttributes(role);
+  });
+
+  it('should be created', () => {
+    expect(component).toBeTruthy();
+  });
+
```
msorens, line 108: Woohoo for tests. 🎉

components/automate-ui/src/app/entities/environments/environment.model.ts
msorens: Oh, I should have mentioned: the JavaScript spread operator (`...`) does only _shallow_ cloning. If you need deeper, then you would likely need to stay with `merge`.

components/automate-ui/src/app/modules/infra-proxy/environment-details/environment-details.component.ts
```diff
@@ -0,0 +1,139 @@
...
+export class EnvironmentDetailsComponent implements OnInit, OnDestroy {
+  public environment: Environment;
+  public cookbookVersions: CookbookVersion[];
+  public tabValue: EnvironmentTabName = 'cookbookConstraints';
+  public url: string;
+  public conflictErrorEvent = new EventEmitter<boolean>();
+  public modalType: string;
+  public serverId;
+  public orgId;
+  public name;
```
SEAjamieD, line 36: some types here would be great so we know exactly what they are.

components/automate-ui/src/app/entities/environments/environment.model.ts
vinay033: I have added this :)

components/automate-ui/src/app/modules/infra-proxy/environment-details/environment-details.component.ts
vinay033: I have added :)

components/automate-ui/src/app/entities/environments/environment.model.ts
vinay033: I tried but I am getting `Spread types may only be created from object types.` so, for now, I am staying with the `merge`.

msorens: You have missed my major concern here: there is something wrong with the types.
According to these lines:
https://github.com/chef/automate/blob/ee6ec4188a6d2b2a65ebd20964eaa726f43d74f3/components/automate-ui/src/app/entities/environments/environment.model.ts#L24-L27
...you are storing _objects_ in `default_attributes` and `override_attributes`. But those are declared as _strings_.

vinay033: I have changed to only `Object`

msorens: @vinay033 The line you changed was only one small part of the concern I was attempting to outline. As I stated above in environment.model.ts, "you are storing _objects_ in default_attributes and override_attributes. But those are declared as _strings_." 

➡️ Please review the entire file of environment.model.ts to correct that type mismatch.

Once that is done, make sure that any consumers (like L112 environment-details.component.ts that I showed as just one example) are correct as well.

Also, once you have done that, I still suggest the spread operator can then be used instead of the merge function (assuming a shallow copy is sufficient for your needs).

Thanks!

vinay033: @msorens , I checked did some R&D on this, actually, we don't know which kind of data we are receiving in API response(maybe string or maybe object) so we are not exactly sure about the type. for reference, you can check this as well. type.https://github.com/chef/automate/blob/6b3840d29126d224a816ac118666193cfb3fb131/components/automate-ui/src/app/page-components/attributes/attributes.component.ts#L66 so we need to use `string | Object` or `any`.

msorens: In this bit of code...
```
export class EnvironmentAttributes {
  default_attributes: string; // (1)
  override_attributes: string;
  all: string;
  constructor(resp: RespEnvironmentAttributes) {
    this.default_attributes = // (2)
      (resp.default_attributes && JSON.parse(resp.default_attributes)) || {};
    this.override_attributes =
      (resp.override_attributes && JSON.parse(resp.override_attributes)) || {};
```
Notice that `default_attributes` is declared as a string (1). You are then assigning an object (in all cases) to that property (2). That is the type mismatch I have been referring to.

vinay033: @msorens I have changed to `string` only to fix type mismatch.

msorens: I was going to take a look but I need instructions on how to arrive at the pages shown in your screen shots above. Please update your PR preamble with necessary details so I can exercise the code. Thanks!

msorens: By the way, looks like this issue is not localized to `environments`. Looks like it was copied from `roles`. So if you want to merge this as is and then address the issue for both roles and environments in a separate PR, that is fine. But if you choose to do so, please add a new card so this issue gets resolved as soon as we can.

vinay033: nvaigate

vinay033: roles and environment both code is added by me and I only used the pre-existing `Json tree`  component of automate UI lib, and I was mentioned the reference files. so we can fix this on a separate card.
This code is not only in this PR, `Json tree` is also using in `clients run` page ==> `attributes.component.ts` so I referred the same for now but we need to fix this throughout the automate.

components/automate-ui/src/app/entities/environments/environment.model.ts
```diff
@@ -1,3 +1,36 @@
...
+
+export interface CookbookVersion {
+  name: string;
+  version: string;
+}
+
+export class EnvironmentAttributes {
+  default_attributes: string;
+  override_attributes: string;
+  all: Object;
```
vinay033, line 20: @msorens I have changed the type of `all` to Object because only we need to default and override attributes as keys also I applied the same changes for the roles.

components/automate-ui/src/app/modules/infra-proxy/environment-details/environment-details.component.ts
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  // set selected item to selected_level
+  filter(precedence_level: string): void {
+    this.selected_level = precedence_level;
+    this.selectedAttrs = this.retrieve(precedence_level);
+  }
+
+  // retrieve attributes based on their level of precedence
+  retrieve(level: string): Object {
```
vinay033, line 112: @msorens I also changed this to `object` because our default precedence_level is `all`

components/automate-ui/src/app/entities/environments/environment.model.ts
msorens: How do I get valid infra data loaded? I have tried manually adding (but typed a random key) as well as `infra_service_load_sample_data` but always get "Could not get cookbooks: invalid org ID: private key block size invalid: private key block size invalid" when going to the org.

vjeffrey: @msorens pls see the second pinned msg in #automate-infra-views

components/automate-ui/src/app/entities/environments/environment.model.ts
msorens: Please examine this block of code--what you changed does _not_ fix it; in a way, it makes it worse.

https://github.com/chef/automate/blob/fcfad120a7077e9c2c14554af6e61252f73472b3/components/automate-ui/src/app/entities/environments/environment.model.ts#L17-L25
The expression `JSON.parse(resp.default_attributes)` returns an Object, **not** a string, so you are storing an object into a variable declared as a string. That tells me that the code is mixing up types and may get unexpected results because of these inconsistencies. 
When I run it in the browser, I saw `resp.default_attributes` come in first as an empty string, so you forced that to set `this.default_attributes` to empty string. _That is the only value where `this.default_attributes` will be set to a string._

Letting the code run further my breakpoint was hit again, this time with  `resp.default_attributes` as `'{}'`, which sets `this.default_attributes` to an empty _object_.
As far as I can see, `default_attributes` and `override_attributes` need to be declared as objects and the code should be reverted to this:
```
this.default_attributes =
      (resp.default_attributes && JSON.parse(resp.default_attributes)) || {};
this.override_attributes =
      (resp.override_attributes && JSON.parse(resp.override_attributes)) || {};
```

components/automate-ui/src/app/entities/environments/environment.model.ts
```diff
@@ -1,3 +1,41 @@
...
+}
+
+export interface RespEnvironmentAttributes {
+  default_attributes: string;
+  override_attributes: string;
+}
+
+export interface AllAttributes {
+  default_attributes: Object;
+  override_attributes: Object;
```
msorens, line 40: Prefer not to duplicate things (these are both here and infra-role.model.ts).
Suggest:
(1) delete these lines here.
(2) rename `RespRoleAttributes` to `RespAttributes`.
(3) import `RespAttributes` and `AllAttributes` from infra-role.model.ts in this file

vinay033: I have incorporated review comments. :)

PR #3684, opened 2020-05-14
ci: fix nightly a1 legacy token test

srenatus:
Verified that this works here: https://buildkite.com/chef/chef-automate-master-verify-private/builds/12380#d964062c-54bb-4f3c-bd84-f771176a672e/440-441 -- but it'll keep being run from nightlies.


srenatus:
will merge when green

PR #3690, opened 2020-05-15
automate-gateway: disallow unknown fields (with exceptions)

srenatus:

It's a trap to accept unknown fields in an API. Users may send a request, everything looks great, but really half of what they had intended didn't work.

So this change enables strict validation of request payloads for grpc-gateway-provided endpoints, with the following exception:

**If the header `Content-Type: application/json+lax` is set**, the API will ignore unknown fields.

1. For `/api/v0/ingest/events/chef/*`, this header is automatically applied.
2. All requests made by automate-ui to the backend apply this header, too.

From what I understand, (1.) is required, and inherent to our mode of operation -- we don't know all the fields we may get.

(2.) is a shortcut. I don't think the UI should need this, but test-driving this change showed me that we depend on this "feature" in enough places to split off any refactorings related to this.

None, sorry.

### :+1: Definition of Done

Everything works like before, except stricter.

- `export TOK=$(chef-automate iam token create asdf --admin)`
- create a team with too many fields:
```
[123][default:/src:0]# curl -k -H "api-token: $TOK" "https://127.0.0.1/apis/iam/v2/teams?pretty" -d '{"id":"foo", "name":"foo", "projects": [],"foo":"bar"}'
{
  "error": "unknown field \"foo\" in request.CreateTeamReq",
  "code": 3,
  "message": "unknown field \"foo\" in request.CreateTeamReq",
  "details": [
  ]
}
```
Send the same payload with the lax header:
```
[125][default:/src:130]# curl -k -H "api-token: $TOK" "https://127.0.0.1/apis/iam/v2/teams?pretty" -d '{"id":"foo", "name":"foo", "projects": [],"foo":"bar"}'  -H "content-type: application/json+lax"
{
  "error": "team with ID \"foo\" already exists",
  "code": 6,
  "message": "team with ID \"foo\" already exists",
  "details": [
  ]
}
```

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/.studio/automate-gateway
+++ b/.studio/automate-gateway
@@ -141,7 +141,7 @@ function send_chef_action_example() {
   tmp_action_json="$(jq --arg event_task "$event_task" \
                         --arg rfc_time "$rfc_time" \
                         --arg event_type "$event_type" \
-                        --arg random_key "$(head -c 100 /dev/urandom | tr -cd 'a-zA-Z')" \
+                        --arg random_key "this_is_surely_not_a_known_key" \
                         '.task = $event_task |
                          .recorded_at = $rfc_time |
                          .entity_type = $event_type |
@@ -340,7 +340,7 @@ function uniqify_chef_run_report() {
      --arg run_uuid "$run_uuid" \
      --arg node_name "$node_name" \
      --arg node_uuid "$node_uuid" \
-     --arg random_key "$(head -c 100 /dev/urandom | tr -cd 'a-zA-Z')" \
+     --arg random_key "this_is_surely_not_a_known_field" \
      '.end_time = $rfc_time |
       .start_time = $rfc_time |
       .entity_uuid = $node_uuid |
--- a/.studio/automate-gateway
+++ b/.studio/automate-gateway
@@ -138,8 +138,14 @@ function send_chef_action_example() {
   rfc_time=$(date --rfc-3339=seconds  -d "$((RANDOM % 144)) hour ago" | sed 's/ /T/' | sed 's/\+.*/Z/')
 
   # Update the 'recorded_at' time from the ChefAction Example; and set entity_type and task
-  tmp_action_json="$(jq --arg event_task "$event_task" --arg rfc_time "$rfc_time" --arg event_type "$event_type" \
-    '.task = $event_task | .recorded_at = $rfc_time | .entity_type = $event_type' < $examples_path/chef_action.json)"
+  tmp_action_json="$(jq --arg event_task "$event_task" \
+                        --arg rfc_time "$rfc_time" \
+                        --arg event_type "$event_type" \
+                        --arg random_key "$(head -c 100 /dev/urandom | tr -cd 'a-zA-Z')" \
+                        '.task = $event_task |
+                         .recorded_at = $rfc_time |
+                         .entity_type = $event_type |
+                         .[$random_key] = "foo"' < $examples_path/chef_action.json)"
 
   # note: this needs to be fed in via stdin, as passing it in as --data "$tmp_action_json"
   # will exceed argmax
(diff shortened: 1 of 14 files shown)


.studio/automate-gateway
```diff
@@ -138,8 +138,14 @@ function send_chef_action_example() {
   rfc_time=$(date --rfc-3339=seconds  -d "$((RANDOM % 144)) hour ago" | sed 's/ /T/' | sed 's/\+.*/Z/')
 
   # Update the 'recorded_at' time from the ChefAction Example; and set entity_type and task
-  tmp_action_json="$(jq --arg event_task "$event_task" --arg rfc_time "$rfc_time" --arg event_type "$event_type" \
-    '.task = $event_task | .recorded_at = $rfc_time | .entity_type = $event_type' < $examples_path/chef_action.json)"
+  tmp_action_json="$(jq --arg event_task "$event_task" \
+                        --arg rfc_time "$rfc_time" \
+                        --arg event_type "$event_type" \
+                        --arg random_key "$(head -c 100 /dev/urandom | tr -cd 'a-zA-Z')" \
```
danielsdeleo, line 144: I think I'd prefer you generate some nonsense just once and then apply it statically here. If this fails it might be difficult to reproduce if the value is not logged (or if it's hard to find the logs or the log entry). Fuzzy type testing is cool but I think you really need the additional tooling for showing and manipulating the RNG seed to get good value.

components/automate-gateway/gateway/services.go
```diff
@@ -366,11 +368,39 @@ func versionedRESTMux(grpcURI string, dopts []grpc.DialOption, toggles gwRouteFe
...
 func muxFromRegisterMap(grpcURI string, dopts []grpc.DialOption, localEndpoints map[string]registerFunc) (*runtime.ServeMux, func(), error) {
 	opts := []runtime.ServeMuxOption{
 		runtime.WithIncomingHeaderMatcher(headerMatcher),
-		runtime.WithMarshalerOption("application/json+pretty", &runtime.JSONPb{OrigName: true, EmitDefaults: true, Indent: "  "}),
-		runtime.WithMarshalerOption(runtime.MIMEWildcard, &runtime.JSONPb{OrigName: true, EmitDefaults: true}),
+		// Note(sr): pretty implies strict -- the assumption is that pretty is used by API usage from an
+		//           interactive console, so we'd want to tell them if they feed us garbage.
+		runtime.WithMarshalerOption("application/json+pretty",
+			&m{JSONPb: &runtime.JSONPb{OrigName: true, EmitDefaults: true, Indent: "  "}, unmarshaler: &jsonpb.Unmarshaler{AllowUnknownFields: false}}),
+		runtime.WithMarshalerOption("application/json+lax", &runtime.JSONPb{OrigName: true, EmitDefaults: true}),
```
danielsdeleo, line 401: Nice to have this option when you really want it 💯

components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
```diff
@@ -35,6 +35,11 @@ export class HttpClientAuthInterceptor implements HttpInterceptor {
     //           Right now, we're throwing the ID token across the internet for telemetry,
     //           where it's not needed. It would be nice to _not_ do that.
     let headers = request.headers.set('Authorization', `Bearer ${this.chefSession.id_token}`);
+    // TODO(sr): Sadly, our UI code depends on the API ignoring unknown fields in the
+    //           request payloads in many places. We should not send any of those. But
+    //           Fixing that blows the scope of my little API validation adventure, so
+    //           we take a shortcut: ask the API not to be too strict on us.
+    headers = headers.set('Content-Type', 'application/json+lax');
```
tylercloke, line 42: :(

e2e/cypress/integration/api/iam/iam_integration.spec.ts
```diff
@@ -133,7 +133,7 @@ describe('assigning projects', () => {
 
     it(`when creating ${iamResources}, successfully assigns an existing authorized project`, () => {
       cy.request({
-        headers: { 'api-token': twoAllowedProjectsTok },
+        headers: { 'api-token': twoAllowedProjectsTok, 'content-type': 'application/json+lax' },
```
bcmdarroch, line 136: do these tests fail now without the `content-type` defined?

srenatus: Unfortunately, yes. Everywhere the content-type is added is a tiny piece of tech-debt, I'm afraid.

I believe here, we're sending too many fields on purpose, to keep the code reusable for different `iamResources`? But I haven't dug too deep.

PR #3693, opened 2020-05-15
remove frontend validation for data feed urls

### LINKED ISSUE #3691: Can no longer create data feeds with IP based URLs
## Describe the bug

Since we introduced validation and testing of data feed parameters in #3476 we can no longer use IP addresses in data feed URL's and must now use FQDN's.

It's not super obvious in the UI why the validation fails since the actual connection test succeeds, and it's probably valid that we should allow IP addresses anyway.

image

## To Reproduce
- Create some kind of listening service to receive Automate data (Kafka, splunk, logstash or similar)
- Set up a data feed to SNOW
- Set the URL to point by IP to the listening service similar to `https://34.77.151.245:8080`
- Fill out the other params (UID,PW etc)
- Test the connection and note that it passes just fine
- Note the greyed out button to create the feed.

- OS: CentOS7
- Browser: Chrome 81.0.4044.138 (Official Build) (64-bit)
- Automate Build Number: 2020051320542

## Additional context
Several large customers are trialling the BETA data feed for SNOW to send data to alternate tools like Splunk, Kafka, logstash etc. They sometimes struggle to get DNS names for ephemeral machines in test environments (I know, sad in this day and age).

We are also starting to show other customers how to use this BETA feature in environments where SaaS data targets don't automagically get public DNS records.

CC @devopslifter @lhasadreams @satellite15

susanev:
folks need to be able to provide items other than standard urls

fixes #3691 

### :+1: Definition of Done
customers are able to create data feeds again

1. `rebuild components/automate-ui-devproxy`
1. navigate to settings > data feeds
1. use the create data feed button
1. provide an IP for the url, make sure it allows you to create it
1. also confirm that entering all whitespace is still disallowed, as is an empty field
1. Editing a data feed URL has the same changes

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/helpers/auth/regex.ts
+++ b/components/automate-ui/src/app/helpers/auth/regex.ts
@@ -24,7 +24,6 @@ export class Regex {
 
     // Allows valid FQDN only
     VALID_FQDN: /^(http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?[a-zA-Z0-9_]+([\-\.]{1}[a-zA-Z0-9]+)*\.[a-zA-Z]{2,5}(:[0-9]{1,5})?(\/.*)?$/,
-    VALID_DATA_FEED_URL: /^(http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?[a-zA-Z0-9_]+([\-\.]{1}[a-zA-Z0-9]+)*\.[a-zA-Z]{2,5}(:[0-9]{1,5})?(\/.*)?$/,
     // Allows valid IP Address only (ipv4)
     VALID_IP_ADDRESS: '^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$'
   };
--- a/components/automate-ui/src/app/pages/data-feed-details/data-feed-details.component.ts
+++ b/components/automate-ui/src/app/pages/data-feed-details/data-feed-details.component.ts
@@ -72,10 +72,8 @@ export class DataFeedDetailsComponent implements OnInit, OnDestroy {
     this.updateForm = this.fb.group({
       // Must stay in sync with error checks in data-feed-details.component.html
       name: ['', [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]],
-      url: ['', [Validators.required,
-        Validators.pattern(Regex.patterns.NON_BLANK),
-        Validators.pattern(Regex.patterns.VALID_FQDN)
-      ]]
+      // Note that URL here may be FQDN -or- IP!
+      url: ['', [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]]
     });
 
     this.store.pipe(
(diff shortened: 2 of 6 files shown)


components/automate-ui/src/app/pages/data-feed-details/data-feed-details.component.ts
```diff
@@ -72,10 +72,8 @@ export class DataFeedDetailsComponent implements OnInit, OnDestroy {
     this.updateForm = this.fb.group({
       // Must stay in sync with error checks in data-feed-details.component.html
       name: ['', [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]],
-      url: ['', [Validators.required,
-        Validators.pattern(Regex.patterns.NON_BLANK),
-        Validators.pattern(Regex.patterns.VALID_FQDN)
-      ]]
+      // Note that URL here may be FQDN -or- IP!
+      url: ['', [Validators.required, Validators.pattern(Regex.patterns.NON_BLANK)]]
```
SEAjamieD, line 76: I'm thinking we don't want to kill all this validation for the future.  I believe the only thing that needs to be done here is to remove `Validators.pattern(Regex.patterns.NON_BLANK)`

components/automate-ui/src/app/helpers/auth/regex.ts
```diff
@@ -24,7 +24,6 @@ export class Regex {
 
     // Allows valid FQDN only
     VALID_FQDN: /^(http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?[a-zA-Z0-9_]+([\-\.]{1}[a-zA-Z0-9]+)*\.[a-zA-Z]{2,5}(:[0-9]{1,5})?(\/.*)?$/,
-    VALID_DATA_FEED_URL: /^(http:\/\/www\.|https:\/\/www\.|http:\/\/|https:\/\/)?[a-zA-Z0-9_]+([\-\.]{1}[a-zA-Z0-9]+)*\.[a-zA-Z]{2,5}(:[0-9]{1,5})?(\/.*)?$/,
```
SEAjamieD, line 27: We don't want to kill this because the whole application may use this regex

components/automate-ui/src/app/pages/data-feed-details/data-feed-details.component.ts
SEAjamieD: i'm gonna take a quick stab at this for you to see if it's all that is needed

components/automate-ui/src/app/helpers/auth/regex.ts
SEAjamieD: thought you may have checked. already
@msorens - are we removing validation here forever?

components/automate-ui/src/app/pages/data-feed-details/data-feed-details.component.ts
SEAjamieD: oh sorry!  you already got the one I was looking at - please disregard - not the non blank 🙏

components/automate-ui/src/app/helpers/auth/regex.ts
susanev: i dont think we have any need to validate URLs like this

PR #3697, opened 2020-05-15
Added RunTimeLimit option and runner code to strip profiles metadata

alexpop:
This is the Automate equivalent change for this audit cookbook report reducing approach: https://github.com/chef-cookbooks/audit/pull/418

It will strip profile metadata from reports when the compliance-profile index already has the metadata.

### DIFF
--- a/components/compliance-service/cmd/compliance-service/cmd/run.go
+++ b/components/compliance-service/cmd/compliance-service/cmd/run.go
@@ -34,7 +34,7 @@ var conf = config.Compliance{
 		TmpDir:              os.Getenv("TMPDIR"),
 		ResultMessageLimit:  10000,
 		ControlResultsLimit: 50,
-		RunTimeLimit:        1.0,
+		RunTimeLimit:        1.0, // Value in seconds
 	},
 	ElasticSearch: config.ElasticSearch{},
 	ElasticSearchSidecar: config.ElasticSearchSidecar{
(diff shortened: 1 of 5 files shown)


components/compliance-service/inspec-agent/runner/runner.go
```diff
@@ -531,9 +538,22 @@ func (t *InspecJobTask) reportIt(ctx context.Context, job *types.InspecJob, cont
...
+	if err != nil {
+		return errors.Wrap(err, "Report profiles processing error")
+	}
+	profilesMissingMeta := make(map[string]struct{}, 0)
+	for _, profileId := range esProfilesMissingMeta {
+		profilesMissingMeta[profileId] = struct{}{}
+	}
+	stripProfilesMetadata(&report, profilesMissingMeta, RunTimeLimit)
+
+	_, err = t.ingestClient.ProcessComplianceReport(ctx, &report)
```
rickmarry, line 556: nice!

components/compliance-service/cmd/compliance-service/cmd/run.go
```diff
@@ -34,6 +34,7 @@ var conf = config.Compliance{
 		TmpDir:              os.Getenv("TMPDIR"),
 		ResultMessageLimit:  10000,
 		ControlResultsLimit: 50,
+		RunTimeLimit:        1.0,
```
rickmarry, line 37: should we have a comment on this to document the unit of time?  min? seconds?

alexpop: Good point, will put a comment and will mention it in the runFlags description.

PR #3698, opened 2020-05-15
[automate-2700] stabilize team_mgmt cypress tests

bcmdarroch:
These rascals are flaky no more 🎊 

Also fixes a minor bug with the Create Object Modal in which the projects dropdown did not reset between modal openings.

### :+1: Definition of Done
- [ ] team_mgmt cypress test suite passes locally and on buildkite consistently

[Make sure prereqs installed](https://github.com/chef/automate/tree/master/e2e#running-cypress-locally)

Opens Cypress test runner with `runFlaky=no`
```
cd e2e
npm run cypress:local
```

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
@@ -32,6 +32,7 @@ export class CreateObjectModalComponent implements OnInit, OnDestroy, OnChanges 
   @Output() createClicked = new EventEmitter<Project[]>();
 
   public projects: ProjectCheckedMap = {};
+  public checkedProjectIDs: string[] = []; // resets project dropdown btwn modal openings
   public policies: PolicyChecked[] = [];
   public modifyID = false; // Whether the edit ID form is open or not.
   public conflictError = false;
@@ -110,11 +111,13 @@ export class CreateObjectModalComponent implements OnInit, OnDestroy, OnChanges 
 
   closeEvent(): void {
     this.modifyID = false;
+    this.checkedProjectIDs = [];
     this.close.emit();
   }
 
   createObject(): void {
     this.createClicked.emit();
+    this.checkedProjectIDs = [];
   }
 
   private isNavigationKey(event: KeyboardEvent): boolean {
--- a/e2e/cypress/integration/ui/settings/team_mgmt.spec.ts
+++ b/e2e/cypress/integration/ui/settings/team_mgmt.spec.ts
@@ -117,12 +117,29 @@ describe('team management', () => {
       cy.contains(customTeamID).should('exist');
     });
 
+    it('fails to create a team with a duplicate ID', () => {
+      cy.get('[data-cy=team-create-button]').contains('Create Team').click();
+      cy.get('app-team-management chef-modal').should('exist');
+
+      cy.get('[data-cy=create-name]').type(teamName);
+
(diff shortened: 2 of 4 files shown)


components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
```diff
@@ -110,11 +111,13 @@ export class CreateObjectModalComponent implements OnInit, OnDestroy, OnChanges
 
   closeEvent(): void {
     this.modifyID = false;
+    this.checkedProjectIDs = [];
```
msorens, line 114: 🎉 FYI: This is essentially emitting a checkedProjectIDs-list-has-changed event to the `ProjectsDropdownComponent`. Even though we are always setting it to an empty list it is still treated as a change event, which is good, because that is what we need for `ProjectsDropdownComponent` to reset its list of checked items to this "new" list of empty.
(For updating a team or token (in the *DetailComponents), `checkedProjectIDs` gets set to the starting value before the update so it does actually have a use. 😁 )

e2e/cypress/integration/ui/settings/team_mgmt.spec.ts
```diff
@@ -52,14 +54,16 @@ describe('team management', () => {
...
 
   after(() => {
-    cy.cleanupIAMObjectsByIDPrefixes(cypressPrefix, ['projects']);
+    cy.cleanupIAMObjectsByIDPrefixes(cypressPrefix, ['projects', 'teams']);
   });
 
-  describe('no custom initial page state', () => {
+  describe('teams list page', () => {
+    const dropdownNameUntilEllipsisLen = 25;
+    const projectDropdown = 'app-team-management app-projects-dropdown ';
```
msorens, line 65: The trailing space here is vital but that is not at all obvious. Consider:
```suggestion
    const projectDropdown = 'app-team-management app-projects-dropdown' + ' '; // trailing space vital!
```

e2e/cypress/support/commands.ts
```diff
@@ -45,25 +45,28 @@ Cypress.Commands.add('logout', () => {
 // if there are no resulting changes to apply, it will simply click out of the filter.
```
msorens, line 45: I am guessing that you found it more reliable to alter the strategy here to always clear the filter and always apply the changes. Any idea why that is the case? It is better to let controls behave "naturally" (the way a user would use them) rather than have to force them, but I understand it may be necessary sometimes.
However, if that is the way we need to go, please update the comment on this line to match.

e2e/cypress/integration/ui/settings/team_mgmt.spec.ts
srenatus: Some ideas to make it more obvious:

- let's name the constant after what it is -- a css selector (prefix)
- let's add the space in its call sites (?)
- let's explain why it's "vital"

e2e/cypress/support/commands.ts
bcmdarroch: The way this command was originally written proved flaky in cases when the same projects were checked over and over - it would just hang after checking the boxes.

Since this helper will be used in a variety of tests with different preconditions that will be tricky to predict, I'd prefer to make sure of the `force:true` shortcut for reliability. I'd rather expand the `global_filter.spec.ts` to add tests covering specific behavior expectations where we have more control over what state the project dropdown will begin with.

PR #3568, opened 2020-05-02
Dan/prototype infra changes dashboard

danielsdeleo:
### :nut_and_bolt: Description: What code changed, and why?

Thus far, it just adds the boilerplate needed to use Postgres to config-mgmt service.

### :white_check_mark: Checklist

- [ ] I have read the [CONTRIBUTING document](https://github.com/chef/automate/blob/master/CONTRIBUTING.md).
- [ ] Tests added/updated?
- [ ] Docs added/updated?
- [ ] All commits have been signed-off for the [Developer Certification of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

Issue #3222, opened 2020-03-26
desktop offering: chef infra change tracking (changelog) in automate


vjeffrey:
## User Story

_this card is follow up to https://github.com/chef/automate/issues/3106_

3106 was a spike card.
the goal of this card is to have a working api when it is complete.
product has signed off on the approach described here: https://github.com/chef/automate/issues/3106#issuecomment-600339937

_more details to follow later_

vjeffrey:
follow up with btm, liz, jessica re: user flow for this, changes in infra

PR #3699, opened 2020-05-15
Desktop View: Change Tracking: Create and List Rollouts

danielsdeleo:

Implements create/read/list for rollout objects. Rollout objects represent the user's intent to have some set of Chef Infra nodes run a certain version of Chef Code. Rollouts contain some data describing which nodes are supposed to run the code, based on the policy name, policy group, and Chef Server URL. Rollouts also contain some data describing the code, such as the policyfile revision_id, some optional git/github info, and some optional Ci info.

Future work will associate rollouts to Chef Client runs so we can show users the status of their rollouts based on how many nodes have run the latest codes and the client run status outcomes.

* #3568
* #3222

### :+1: Definition of Done

create/read/list the rollout objects. This PR is already large so I'm cutting it off.

* build automate-gateway and config management service
* start all the services
* `config_mgmt_list_rollouts` to list the rollouts (initially you should have zero)
* `config_mgmt_create_rollout` to create a rollout
* The studio helpers are designed for piping so you can use `jq` to create more varied data


api/external/cfgmgmt/cfgmgmt.proto
```diff
@@ -399,4 +401,73 @@ service ConfigMgmt {
...
+  the latest policy revision pushed to a policy group.
+
+  Authorization Action:
+  ```
+  ingest:unifiedEvents:create
+  ```
+  */
+  rpc CreateRollout(cfgmgmt.request.CreateRollout) returns (cfgmgmt.response.Rollout) {
+    option (google.api.http) = {
+      post: "/api/beta/cfgmgmt/rollouts/create"
```
lancewf, line 418: I am glad you are using "beta" for the path.

components/config-mgmt-service/backend/postgres/schema/sql/01_noop.up.sql
```diff
@@ -0,0 +1 @@
+
```
lancewf, line 1: Why is this needed?

api/external/cfgmgmt/request/rollouts.proto
```diff
@@ -0,0 +1,48 @@
...
+  string ci_job_id = 12;
+}
+
+message Rollouts {
+  // Filters to apply to the request for the rollouts list.
+  repeated string filter = 1;
+}
+
+message RolloutById {
+  string rollout_id = 1;
```
lancewf, line 40: Why not have this be an int32?

components/config-mgmt-service/grpcserver/rollouts.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return rolloutToAPI(rollout), nil
+}
+
+func (s *CfgMgmtServer) GetRollouts(ctx context.Context, req *request.Rollouts) (*response.Rollouts, error) {
+	if len(req.GetFilter()) > 0 {
+		return nil, status.Error(codes.Unimplemented, "NOT IMPLEMENTED")
```
lancewf, line 67: This is a good idea. I will have to start using this.

.expeditor/verify_private.pipeline.yml
```diff
@@ -325,7 +325,7 @@ steps:
   - label: "config-mgmt-service"
     command:
       - . scripts/verify_setup.sh
-      - hab studio run "source scripts/verify_studio_init.sh && start_deployment_service && chef-automate dev deploy-some chef/automate-elasticsearch --with-deps && config_mgmt_integration"
+      - hab studio run "source scripts/verify_studio_init.sh && start_config_mgmt_service_deps && config_mgmt_integration"
```
lancewf, line 328: nice clean up.

components/config-mgmt-service/backend/postgres/schema/sql/01_noop.up.sql
danielsdeleo: The whitespace change is a mistake, I had been writing the "CREATE TABLE" code in this file on accident.

The reason we have a noop migration file here in the first place is that I couldn't get all the migration tooling working (with hab &etc.) without having at least one file in this directory, so I needed to make a noop migration in the last PR which added postgres to the config management service.

api/external/cfgmgmt/request/rollouts.proto
danielsdeleo: Despite what [Hyrum's law says](https://www.hyrumslaw.com) :P I am trying to make end users treat our IDs as opaque blobs. For example, if we were to ever need to change the IDs to 64 bit in the database, it would be painful if we had them as int32 in the API.

api/external/cfgmgmt/response/rollouts.proto
```diff
@@ -0,0 +1,87 @@
...
+  // If the rollout was initiated via Ci/CD or similar system, the web URL
+  // for the job that initiated the rollout.
+  string ci_job_url = 11;
+  // If the rollout was initiated by Ci/CD or similar system, the id of the job
+  // that initiated the rollout. Should include the Ci system's nickname or
+  // other identifying information users would need to associate the job ID to
+  // the Ci/CD system.
+  string ci_job_id = 12;
+
+  string id = 13;
```
jyan-chef, line 79: Can you add a description or id? Also is the something that increments forever? Can it be reset to 0 somehow?

api/external/cfgmgmt/response/rollouts.proto
```diff
@@ -0,0 +1,87 @@
...
+  string start_time = 14;
```
jyan-chef, line 80: Can you add a comment to clarify that the start time is when the code got pushed?

api/external/cfgmgmt/response/rollouts.proto
```diff
@@ -0,0 +1,87 @@
...
+  string end_time = 15;
```
jyan-chef, line 81: Can you explain how the end_time gets generated?

api/external/cfgmgmt/response/rollouts.proto
```diff
@@ -0,0 +1,87 @@
...
+  SCMWebType scm_web_type = 6;
+  // The URL used to obtain a copy of the source code repository
+  string policy_scm_url = 7;
+  // The URL used to view the source code repository via the web
+  string policy_scm_web_url = 8;
+  // The source control system's identifier for the repository version. This
+  // should be the version where the policy's lockfile was committed.
+  string policy_scm_commit = 9;
+  // A free-form description of the rollout, as given by the user.
+  string description = 10;
```
jyan-chef, line 69: Is there a char limit on how long a description can be? Can you add a comment that talks about grabbing the Git commit "comment" and also user manual entered comment if they push directly to the server?

api/external/cfgmgmt/response/rollouts.proto
```diff
@@ -0,0 +1,87 @@
...
+latest version of the Chef Infra code assigned to them and also provide the
+user insights about the code by aggregating Chef Client run results
+according to the version of Chef Infra code applied. Metadata about the code
+is stored in order to provide the user with convenient references back to
+systems they already use (such as SCM and Ci/CD systems) to manage their code.
+
+Nodes are segmented by a triple of policy name, policy group, and policy domain URL:
+* policy name generally describes what kind of system it is, e.g., a database server
+* policy group generally describes where the system fits in the user's code
+  lifecycle, e.g., "QA" or "production"
```
jyan-chef, line 28: can you include mention of "canary" here?

api/external/cfgmgmt/response/rollouts.proto
```diff
@@ -0,0 +1,87 @@
...
+  // that initiated the rollout. Should include the Ci system's nickname or
+  // other identifying information users would need to associate the job ID to
+  // the Ci/CD system.
+  string ci_job_id = 12;
+
+  string id = 13;
+  string start_time = 14;
+  string end_time = 15;
+}
+
```
jyan-chef, line 83: in your next revision, can we add the change initiator info we talked about on 6/24? (i.e. user id from chef server, git ID, email)

Issue #3472, opened 2020-04-22
TypeScript Updates for next level of enforcement to files that have been touched


SEAjamieD:
To facilitate moving to the next level of typescript checks, updates to the following files which I've made edits to in the past:

projects-filter-dropdown.component.ts

...more to add

To see all files that need updates across the application:
open tsconfig.js and flip angularComplierOptions all to true
cd into components/automate-ui
`ng-build`

SEAjamieD:
Closed in favor of tackling one file at a time while working on other tickets

PR #3700, opened 2020-05-15
Change event type of Keyboard Event to Event

SEAjamieD:
Opened Draft for feedback

error TS2345: Argument of type 'Event' is not assignable to parameter of type 'KeyboardEvent'.

Typescript is complaining about the Event type in both the `handleArrowDown` and `handleArrowUp` functions.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
+++ b/components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
@@ -99,8 +99,8 @@ export class ProjectsFilterDropdownComponent {
     this.checkInitialEquality();
   }
 
-  handleArrowUp(event: Event) {
-    event.preventDefault();
+  handleArrowUp(genericEvent: Event) {
+    const event = genericEvent as KeyboardEvent;
 
     const element = (event.target as Element).previousElementSibling;
     if (element) {
@@ -108,8 +108,8 @@ export class ProjectsFilterDropdownComponent {
     }
   }
 
-  handleArrowDown(event: Event) {
-    event.preventDefault();
+  handleArrowDown(genericEvent: Event) {
+    const event = genericEvent as KeyboardEvent;
 
     const element = (event.target as Element).nextElementSibling;
     if (element) {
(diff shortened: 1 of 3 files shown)


components/automate-ui/src/app/page-components/projects-filter-dropdown/projects-filter-dropdown.component.ts
```diff
@@ -99,7 +99,7 @@ export class ProjectsFilterDropdownComponent {
     this.checkInitialEquality();
   }
 
-  handleArrowUp(event: KeyboardEvent) {
+  handleArrowUp(event: Event) {
```
SEAjamieD, line 102: @msorens - Would you happen to have any advice on this?  Typescript is complaining about $event coming in as type `KeyboardEvent`.  Event "fixes" the problem but it feels way too broad since this is specifically a keyboard event.  
error thrown reads as:
```
error TS2345: Argument of type 'Event' is not assignable to parameter of type 'KeyboardEvent'.
```

msorens: Apologies; did not see this before. Where are you seeing the Typescript error?

msorens: Thanks for the details. Here is my suggestion:

Because we cannot cast in a template, we have to accept the argument as a generic event. So let's use that and then redirect it to what we need.
```suggestion
  handleArrowUp(genericEvent: Event) {
    const event = genericEvent as KeyboardEvent;
 ```

Additional notes:
1. For the specific functions `handleArrowUp` (and `handleArrowDown`) it is actually OK to use the signature `handleArrowUp(event: Event)` because the only properties we are using in the body of the function are, in fact, properties of the base class `Event`. That said, there is no harm in making the above change as one could argue it does signal intent a bit more clearly. Your call.

2. There are a couple occurrences of the same error though in another file (resource-dropdown.component.ts) you should remedy in this PR. The original code is this:
```
  moveFocus(event: KeyboardEvent): void {
```
And that does need the change above:
```
  moveFocus(genericEvent: Event): void {
    const event = genericEvent as KeyboardEvent;
```
because it uses properties that are specific to the `KeyboardEvent` subclass.

SEAjamieD: Thank you for this information, It makes a lot of sense and seems like a good way to move forward.  Much appreciated!

susanev:
as of 07/10/2020 this is still being worked on, just lower priority compared to some other urgent work that came in.

SEAjamieD:
Closed in favor of tackling one file at a time while working on other tickets
Closed Ticket: https://github.com/chef/automate/issues/3472

PR #3707, opened 2020-05-17
enable project filtering on nodemanager nodes

vjeffrey:

a few weeks ago we introduced the ability to use project filtering
on manually added nodes by supporting the environment tag on those nodes
and appropriately applying it on the node reports.

what we did not do at that time is also apply the project filtering
on the nodemanager nodes, which affects the list present at scan jobs/nodes
and integration/nodes.

this change ensures that we apply the project filtering to the
nodemanager nodes as well.
At this point, nodes will not be "project filtered" until they are ingested at least once.

### :+1: Definition of Done
nodemanager nodes respect project filtering

1-rebuild nodemanager, compliance
2- create a project, rule of environment: test-env
apply the rules
3-run load_scan_jobs, wait for it to complete
3-go to compliance report nodes, see 3 nodes.
filter by project, see 2 nodes.
4-go to automate-url/nodes, see 2 nodes
deselect project filter, see 3 nodes
^ this tests project tagging and filtering. to test project update and filtering, create another project with environment: test-env-local, and after applying the rules, make sure the manually added node is in the correct project by visiting compliance report nodes and automate-url/nodes

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/nodemanager-service/tests/project_filtering_test.go
+++ b/components/nodemanager-service/tests/project_filtering_test.go
@@ -184,8 +184,7 @@ func TestReadProjectFilteringIngestedNodes(t *testing.T) {
 	}
 }
 
-// All manually added nodes should be returned because they are not included in project filtering.
-func TestListProjectFilteringIngestedNodes(t *testing.T) {
+func TestListProjectFilteringAllNodes(t *testing.T) {
 	db, err := createPGDB()
 	require.NoError(t, err)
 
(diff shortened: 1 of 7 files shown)


lancewf:
I don't see three nodes when I run the `load_scan_jobs` I only see 2. The third one has this error
```
Expected InSpec STDOUT to start with '{' but found 'v'. STDOUT: vagrant@localhost's password:
Expected InSpec STDOUT to end with '}' but found ':'. STDOUT: `vagrant@localhost's password:`

Authentication failed for localhost

Errno::ENOTTY
```
This is not a problem for me because, I should be able to test this by creating local scan jobs.

vjeffrey:
ah crap ya the third one is running against the vagrant dev env so if you use the docker one that one won't work, i forgot about that

components/nodemanager-service/tests/project_filtering_test.go
```diff
@@ -9,15 +9,17 @@ import (
...
 	"github.com/stretchr/testify/require"
 
 	authzConstants "github.com/chef/automate/components/authz-service/constants"
+	"github.com/chef/automate/components/nodemanager-service/pgdb"
 	nodesserver "github.com/chef/automate/components/nodemanager-service/server/nodes"
 	"github.com/chef/automate/lib/grpc/auth_context"
 )
 
-func TestReadProjectFilteringIngestedNodes(t *testing.T) {
+func TestReadProjectFilteringNodes(t *testing.T) {
```
lancewf, line 22: Good job updating this test.

PR #3712, opened 2020-05-18
Automate 3533/multiple dropdowns

msorens:

The mechanism for auto-closing a dropdown when the user clicked anywhere outside its boundaries proved insufficient when putting two dropdowns in the same view.
This PR adds the necessary support.

#### Problem Exposition
The issue was exposed in the CreateObjectModalComponent when used for creating a token specifically, as that is where two dropdowns appear (one for policies, one for projects).

Those two dropdowns _almost_ co-existed peacefully. 
✅ If you opened either one, interacted with it, and closed it that worked.
✅ If you opened either one, and clicked elsewhere on static space in the modal, the dropdown closed.
✅ If you opened either one, and clicked even in an active input field, the dropdown closed.
❌ If however, you opened projects (1), then opened policies (2), it failed to close the projects dropdown and (rather awkwardly) put the freshly opened policies dropdown behind the projects dropdown!
image

#### Problem Description
The capability that provides the elegant close-me-if-you-click-outside is at the heart of the problem. When you click anywhere, it triggers the "check-if-outside" handler action. _This is true even if you are clicking on the dropdown itself._ 

If you actually are clicking outside the dropdown, the "check-if-outside" handler does what it should.
But if you are clicking the dropdown itself to close it, that click emits an event that first triggers a toggle handler to close the dropdown, and then triggers the "check-if-outside" handler action. The toggle handler in its original rendition knew that, if reached, we were not actually outside. It therefore suppressed that event propagation so it never went to the check-outside handler: 
https://github.com/chef/automate/blob/0278ed975f8029ad65351a12a89284b3123dfd4c/components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts#L69-L70 

Neat and simple. But that no longer works with multiple dropdowns. Suppressing the event propagation left the dropdown open--_you could not close it!_

#### Problem Resolution
I first realized that I had to allow the event propagation to proceed in order for things to work.
But that created its own problem--now, _you could not open it!_
(Why? The toggle handler toggled it open, then the "check-if-outside" handler immediately closed it. 🤷 )
The solution was that the "check-if-outside" handler had to be able to distinguish not just open and closed but also an intermediate state--opening.
That allowed opening and closing the modal to work, and open and closing _other_ modals to work.

NA

### :+1: Definition of Done
Clicking outside any open dropdown closes that dropdown.

Rebuild automate-ui.
Create at least one project (Settings >> Projects >> Create Project).
Navigate to Settings >> Tokens >> Create Token.
Open the projects dropdown by clicking on it.
Now open the policies dropdown; this should simultaneously close the projects dropdown:
image

That seems to have worked--but did it really close the projects dropdown--or just open the policies dropdown in front of it?
To prove it really closed the projects dropdown, run `git revert -n 8e73ce4`, rebuild and try again. (The only change was shifting the policy dropdown so the project dropdown remains visible.)
Refresh the page and run the trial again and you will see indeed that the other dropdown has closed:
image

Also, now you can try them in the other order, too, i.e., click (2) then click (1).

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.spec.ts
+++ b/components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.spec.ts
@@ -1,9 +1,10 @@
-import { CUSTOM_ELEMENTS_SCHEMA, EventEmitter } from '@angular/core';
+import { CUSTOM_ELEMENTS_SCHEMA, EventEmitter, SimpleChange } from '@angular/core';
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 import { FormsModule } from '@angular/forms';
 
+import { using } from 'app/testing/spec-helpers';
 import { ChefPipesModule } from 'app/pipes/chef-pipes.module';
-import { ResourceDropdownComponent } from './resource-dropdown.component';
+import { ResourceDropdownComponent, ResourceChecked } from './resource-dropdown.component';
 
 describe('ResourceDropdownComponent', () => {
   let component: ResourceDropdownComponent;
@@ -11,11 +12,11 @@ describe('ResourceDropdownComponent', () => {
 
   beforeEach(async(() => {
     TestBed.configureTestingModule({
-      declarations: [ ResourceDropdownComponent ],
-      imports: [ ChefPipesModule, FormsModule ],
-      schemas: [ CUSTOM_ELEMENTS_SCHEMA ]
+      declarations: [ResourceDropdownComponent],
+      imports: [ChefPipesModule, FormsModule],
+      schemas: [CUSTOM_ELEMENTS_SCHEMA]
     })
-    .compileComponents();
+      .compileComponents();
   }));
 
   beforeEach(() => {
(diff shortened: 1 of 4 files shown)


components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.spec.ts
```diff
@@ -48,15 +49,119 @@ describe('ResourceDropdownComponent', () => {
...
       expect(options.length).toEqual(component.filteredResources.length);
-      options.forEach((option: HTMLInputElement, index: number) => {
+      options.forEach((option, index) => {
         const { name, checked } = component.filteredResources[index];
         expect(option.textContent).toEqual(name);
         expect(option.checked).toEqual(checked);
       });
     });
 
+    it('allows customizing "none" designator', () => {
```
msorens, line 61: Not actually related to the changes in this PR, but lots of necessary tests added.

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
```diff
@@ -41,7 +41,7 @@ export class ResourceDropdownComponent implements OnInit, OnChanges {
   // filteredResources is merely a container to hold the resources
   // that can be altered
   public filteredResources: ResourceChecked[] = [];
-  public active = false;
+  public dropdownState: 'closed' | 'opening' | 'open' = 'closed';
```
msorens, line 44: Convert this from a binary to a trinary variable, as described in the preamble.

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
```diff
@@ -66,15 +66,16 @@ export class ResourceDropdownComponent implements OnInit, OnChanges {
     return ChefSorters.naturalSort(this.resources, 'name');
   }
 
-  toggleDropdown(event: MouseEvent): void {
-    event.stopPropagation();
```
msorens, line 70: Remove the suppression on event propagation--we need that to percolate when other dropdowns are present. (Again, see preamble.)

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.spec.ts
```diff
@@ -48,15 +49,119 @@ describe('ResourceDropdownComponent', () => {
...
+        [
+          genResource('other resource1', false),
+          genResource('other resource2', false),
+          genResource('target resource', true)
+        ]]
+    ], function (description: string, resources: ResourceChecked[]) {
+      it(`displays resource name for ${description}`, () => {
+        component.resources = resources;
+        component.ngOnChanges(
+          { resources: new SimpleChange([], resources, true) });
```
msorens, line 122: Oddly setting a component's `@Input` programmatically does not trigger an `OnChanges` event so we need to do that manually for the test. (`OnChanges` is triggered only when the value changes in the component's view.)

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.spec.ts
```diff
@@ -1,21 +1,22 @@
...
   let fixture: ComponentFixture<ResourceDropdownComponent>;
 
   beforeEach(async(() => {
     TestBed.configureTestingModule({
-      declarations: [ ResourceDropdownComponent ],
-      imports: [ ChefPipesModule, FormsModule ],
-      schemas: [ CUSTOM_ELEMENTS_SCHEMA ]
+      declarations: [ResourceDropdownComponent],
+      imports: [ChefPipesModule, FormsModule],
+      schemas: [CUSTOM_ELEMENTS_SCHEMA]
```
srenatus, line 17: At some point, we could introduce a bunch of `MockComponents` to get rid of this, I guess? Or has our stance towards this changed?

msorens:
@srenatus wrote:

> Would you think it might be helpful to transfer some of the comments re: why the ternary? into code comments, for future readers?

I thought I had actually done that, but I am happy to entertain specific suggestions you may have to make it more clear! 😁

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.spec.ts
msorens: Yep, should do that.

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
```diff
@@ -86,17 +87,19 @@ export class ResourceDropdownComponent implements OnInit, OnChanges {
...
+    // Arriving here, the main goal is to close this dropdown if it is open.
+    if (this.dropdownState === 'open') {
       this.closeDropdown();
     }
-    if (this.activating) {
-      this.active = true;
-      this.activating = false;
+    // Now that we have checked the above,
+    // it is safe to complete opening if the click occurred when the dropdown was closed.
+    if (this.dropdownState === 'opening') {
```
bcmdarroch, line 96: can you describe further the circumstances when the dropdown is `opening`?

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
blakestier: Are these standard js terms? Should they be constants? I've seen the string 'open' in a handful of places at this point

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
msorens: The comments in the file explain that if you read them in sequence:
```
// toggleDropdown:
// we cannot go directly to 'open' because handleClickOutside,
// firing next on the same event that arrived here, would then immediately close it.
// => so it is set to 'opening' instead of 'open'. In other words, delay going to 'open'.

// handleClickOutside()
// Arriving here, the main goal is to close this dropdown if it is open.
// check if 'open' -- and now it will not be if coming from toggleDropdown.
. . .
// Now that we have checked the above,
// it is safe to complete opening if the click occurred when the dropdown was closed.
// => so now we set to 'open'
   
If you have more questions happy to try to explain further, but perhaps live chat might be better? Up to you.

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
srenatus: It's a "string literal type", so that's as good as a constant FWIW. 
 https://mariusschulz.com/blog/string-literal-types-in-typescript has a good intro

srenatus:
@msorens I was about to squash and merge, but I didn't know if you usually massage the message. Let's get this in 🚀

PR #3717, opened 2020-05-18
Add new compliance inspec agent cli options

alexpop:
Added three inspec agent config options introduced in these automate:

- https://github.com/chef/automate/pull/3663

- https://github.com/chef/automate/pull/3697


components/compliance-service/habitat/hooks/run
```diff
@@ -56,6 +56,9 @@ CONFIG="$CONFIG --manager-manual-poll {{cfg.nodemanager.manual_polling_interval}
 CONFIG="$CONFIG --job-workers {{cfg.agent.workers}}"
 CONFIG="$CONFIG --job-buffer-size {{cfg.agent.buffer_size}}"
 CONFIG="$CONFIG --remote-inspec-version {{cfg.agent.remote_inspec_version}}"
+CONFIG="$CONFIG --result-message-limit {{cfg.agent.result_message_limit}}"
+CONFIG="$CONFIG --control-results-limit {{cfg.agent.control_results_limit}}"
+CONFIG="$CONFIG --run-time-limit {{cfg.agent.run_time_limit}}"
```
lancewf, line 61: You might need to add checks if the variable is not initialized. 

{{~#if cfg.agent.control_results_limit}}
CONFIG="$CONFIG --result-message-limit {{cfg.agent.control_results_limit}}"
{{~/if}}

I am seeing the below error when rebuilding the compliance-service
```
invalid argument "--control-results-limit" for "--result-message-limit" flag: strconv.ParseInt: parsing "--control-results-limit": invalid syntax
```

vjeffrey: would having a default val (in the default.toml) also resolve that issue?

lancewf: I think that will. It looks like there are defaults in the run.go https://github.com/chef/automate/blob/10bd0c6b59efbdbacaacd54e9b50ef49f62b4ffe/components/compliance-service/cmd/compliance-service/cmd/run.go#L35. So, if the "--result-message-limit" is not passed in it should use those defaults.

alexpop: Thanks Lance. I think this was caused because https://github.com/chef/automate/pull/3697 was not merged. It is now and I rebased this branch.

alexpop: `default.toml` fixed the cli error Lance mentioned.
Thank you Lance and Victoria!

PR #3725, opened 2020-05-18
Allow configuring non-persistent session

jaym:
It can be desirable to have users logged out when their browser closes.
This PR provides configuration to do that:

```toml
[session.v1.sys.service]
  # Setting persistent to false will disable persistent sessions.
  # Users will be logged out when their browser closes.
  persistent = false
```

Changes to the UI:
- When the application loads, we refresh our session. No requests can be made until that happens.
- Fixed a few issues with the welcome modal:
  - It was possible to use the component before it was fully initialized
  - Fixed an issue where the welcome modal could be displayed multiple times in the same session because the storage and retrieval keys were not the same


components/automate-ui/src/app/services/chef-session/chef-session.service.ts
```diff
@@ -61,36 +70,45 @@ export class ChefSessionService implements CanActivate {
...
-        const xhr = new XMLHttpRequest();
-        xhr.onreadystatechange = this.refreshSessionCallback.bind(this);
-        xhr.open('GET', '/session/refresh', true);
-        xhr.responseType = 'json';
-        xhr.setRequestHeader('Authorization', `Bearer ${this.id_token}`);
-        xhr.send(null);
-      }, minute);
+      timer(0, minute).pipe(
+        filter(_ => !this.isRefreshing),
+        mergeMap(_ => {
```
msorens, line 75: Convention for "no args" in each case:
```suggestion
        filter(() => !this.isRefreshing),
        mergeMap(() => {
```

components/automate-ui/src/app/services/chef-session/chef-session.service.ts
```diff
@@ -61,36 +70,45 @@ export class ChefSessionService implements CanActivate {
...
+          this.ingestIDToken(token);
+          this.isRefreshing = false;
+        },
+        error => {
+          this.isRefreshing = false;
+          if (error instanceof HttpErrorResponse) {
+            if (error.status == HTTP_STATUS_UNAUTHORIZED) {
+              this.logout();
+            } else {
+              console.log(`Session refresh failed: ${error.statusText}`);
```
msorens, line 90: Wondering if we want to expose these more to the user with the notification banner? 🤷

components/automate-ui/src/app/services/chef-session/chef-session.service.ts
```diff
@@ -26,29 +30,34 @@ export interface ChefSessionUser {
...
   // the application can be reinitialized multiple time during a single session.
   //// Flag to store whether or not the modal has been displayed this session.
   //// Automatically set when the modal is shown for the first time.
   MODAL_HAS_BEEN_SEEN_KEY = 'welcome-modal-seen';
 
-  constructor(private store: Store<NgrxStateAtom>) {
+  constructor(private store: Store<NgrxStateAtom>, handler: HttpBackend) {
     // In dev mode, set a generic session so we don't
     // have to round-trip to the oidc provider (dex).
+    this.tokenProvider = new ReplaySubject(1);
```
msorens, line 54: This works, but since you are always setting to 1, it is slightly cleaner to use a BehaviorSubject. It does exactly what a ReplaySubject-set-to-1 does.

jaym: I initially had BehaviorSubject, but it requires you to pass an initial value which I think i do not want

msorens: Ah, ok.

components/automate-ui/src/app/services/chef-session/chef-session.service.ts
srenatus: Isn't this going to become the new "could not verify license"? I.e. the thing that'll happen when the backend is not responding? I think we should try to verify that behaviour. If it's only logging to the console now, we should know.

components/automate-ui/src/app/services/chef-session/chef-session.service.ts
```diff
@@ -61,36 +70,45 @@ export class ChefSessionService implements CanActivate {
...
+    const httpOptions = {
+      headers: new HttpHeaders({
+        'Content-Type': 'application/json',
+        'Authorization': `Bearer ${this.id_token}`
+      })
+    };
+    return this.httpHandler.get('/session/refresh', httpOptions).pipe(
+      map(obj => {
+        return obj['id_token'];
+      }),
```
srenatus, line 110: [nit]
```suggestion
      map(obj => obj.id_token),
```

would do it, too, I think

components/automate-ui/src/app/services/chef-session/chef-session.service.ts
```diff
@@ -61,36 +70,45 @@ export class ChefSessionService implements CanActivate {
...
-        // this?
-        // NOTE 2019/06/27 (sr): this defaults to OK when using HTTP2; not worth
-        // messing with, though, this all needs to be redone.
-        console.log(`Session refresh failed: ${xhr.statusText}`);
-      }
-    }
+  private refresh(): Observable<string> {
+    const httpOptions = {
+      headers: new HttpHeaders({
+        'Content-Type': 'application/json',
```
srenatus, line 103: This content-type is correct, but I don't think we've bothered in other places. Anyways, fine.

components/automate-ui/src/app/services/chef-session/chef-session.service.ts
```diff
@@ -248,12 +271,12 @@ export class ChefSessionService implements CanActivate {
...
     localStorage.setItem(sessionKey,
-                         JSON.stringify({
-                           'uuid': 'test_subject',
-                           'username': 'testchefuser',
-                           'fullname': 'Test User',
-                           'groups': ['group1', 'group2', 'group3'],
-                           'id_token': 'test_id_token'
-                         }));
+      JSON.stringify({
+        'uuid': 'test_subject',
```
srenatus, line 275: [nit] 
```suggestion
        uuid: 'test_subject',
```

but it really doesn't matter and was like this before.

components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
```diff
@@ -49,13 +47,23 @@ export class HttpClientAuthInterceptor implements HttpInterceptor {
...
+        //           Right now, we're throwing the ID token across the internet for telemetry,
+        //           where it's not needed. It would be nice to _not_ do that.
+        headers = headers.set('Authorization', `Bearer ${id_token}`);
+        return next
+          .handle(request.clone({ headers, params })).pipe(
+            catchError((error: HttpErrorResponse) => {
+              if (error.status === 401) {
+                this.chefSession.logout();
+              }
+              return observableThrowError(error);
```
srenatus, line 64: Is this the cleanest thing we can do? IIRC, `this.chefSession.logout()` will change the location (go somepace else) directly, so may I ask for a comment indicating that? For the casual reader, there's no way to tell that _this line_ will not be reached when the status is 401.

components/automate-ui/src/app/services/chef-session/chef-session.service.ts
```diff
@@ -61,36 +70,45 @@ export class ChefSessionService implements CanActivate {
...
-        /* Using XMLHttpRequest instead of angular/http to avoid
-           dependency hell. */
-        const xhr = new XMLHttpRequest();
-        xhr.onreadystatechange = this.refreshSessionCallback.bind(this);
-        xhr.open('GET', '/session/refresh', true);
-        xhr.responseType = 'json';
-        xhr.setRequestHeader('Authorization', `Bearer ${this.id_token}`);
-        xhr.send(null);
-      }, minute);
+      timer(0, minute).pipe(
```
srenatus, line 73: OK, question for you experts: using `window.setInterval` was one of the reasons that we couldn't have our automate-ui e2e protractor tests wait for the angular state to "settle". Does the same thing apply to `timer`? (I think the workaround would have been to cram this into "angular's zone" somehow, but it's really beyond my understanding of this.)

components/automate-ui/src/app/services/chef-session/chef-session.service.ts
jaym: i feel like i tried this and something was complaining

components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
jaym: 👍 We probably should return as well? I'm not sure if changing the location stops the execution of the javascript function?

components/automate-ui/src/app/services/chef-session/chef-session.service.ts
jaym: dont know the answer to that question, but i've somehow broken the cypress tests :(

components/automate-ui/src/app/services/http/http-client-auth.interceptor.ts
srenatus: It'll go to dex, so the JS code here is unloaded. I don't know how that looks in detail, though.

components/automate-ui/src/app/services/chef-session/chef-session.service.ts
srenatus: Type issues?

SEAjamieD:
This appears to be working as advertised so far, which is exciting.  Thanks for working on this - I don't have any new comments to add to the code at this point 👍

PR #3728, opened 2020-05-18
automate-3081 ui flicker on applications tab

### LINKED ISSUE #3081: UI flicker on Applications tab
## Describe the bug

When moving between "OK" and "Disconnected" nodes, the UI moves from the loading spinner, to the _old_ data, to the _new_ data, with a visible redraw when the new data is finally shown.  Please see the following video:
 
output

## To Reproduce
- Go to Application Tab
- Select "OK" filter
- Select "Disconnected" filter.

Note I am very far from this A2 server, so high-network latency may help reproduce it.

## Expected behavior

- Smooth as butter, no flicker.

tarablack01:
When moving between "OK" and "Disconnected" nodes, the UI moves from the loading spinner, to the old data, to the new data, with a visible redraw when the new data is finally shown.

fixes #3081

### :+1: Definition of Done
Smooth as butter, no flicker.

hab studio enter
export DEVPROXY_URL=localhost
start_automate_ui_background
start_all_services

Go to Application Tab
Select "OK" filter
Select "Disconnected" filter.
Note I am very far from this A2 server, so high-network latency may help reproduce it.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

Screen Shot 2020-05-18 at 5 28 22 PM
Screen Shot 2020-05-18 at 5 28 45 PM
Screen Shot 2020-05-18 at 5 28 51 PM

### DIFF
--- a/components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
+++ b/components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
@@ -1,4 +1,4 @@
-import { map, takeUntil, withLatestFrom, distinctUntilChanged } from 'rxjs/operators';
+import { map, takeUntil, withLatestFrom, distinctUntilChanged, filter } from 'rxjs/operators';
 import { Component, OnInit, OnDestroy } from '@angular/core';
 import { Subject, Observable, combineLatest } from 'rxjs';
 import { HttpErrorResponse } from '@angular/common/http';
@@ -252,16 +252,15 @@ export class ServiceGroupsComponent implements OnInit, OnDestroy {
       this.store.select(serviceGroupsList),
       this.store.select(selectedServiceGroupHealth)
     ]).pipe(
+      filter(([sSgStatus, sgStatus, serviceGroups, sgHealth]) =>
+        sgStatus === EntityStatus.loadingSuccess
+        && sSgStatus === EntityStatus.loadingSuccess
+        && serviceGroups.length > 0
+        && sgHealth.total === 0
+      ),
       takeUntil(this.isDestroyed)
-    ).subscribe(([sSgStatus, sgStatus, serviceGroups, sgHealth]) => {
-      if (sgStatus === 'loadingSuccess' && sSgStatus === 'loadingSuccess') {
-          if (serviceGroups.length > 0) {
-              if (sgHealth.total === 0) {
-                this.onServiceGroupSelect(null, serviceGroups[0].id);
-              }
-          }
-      }
-    });
+    ).subscribe(([_sSgStatus, _sgStatus, serviceGroups, _sgHealth]) =>
+        this.onServiceGroupSelect(null, serviceGroups[0].id));
 
     this.selectedStatus$ = this.store.select(createSelector(serviceGroupsState,
       (serviceGroups) => serviceGroups.filters.status));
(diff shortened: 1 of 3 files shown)


tarablack01:
@SEAjamieD thanks for catching that bug! I'll fix it 👍🏻

components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
```diff
@@ -246,22 +246,20 @@ export class ServiceGroupsComponent implements OnInit, OnDestroy {
...
-            });
-          }
-        });
-      }
+    this.store.select(selectedServiceGroupStatus).subscribe((sSgStatus) => {
+      this.store.select(serviceGroupsStatus).subscribe((sgStatus) => {
+        if (sgStatus === 'loadingSuccess' && sSgStatus === 'loadingSuccess') {
+          this.serviceGroupsList$.subscribe((serviceGroups) => {
+            if (serviceGroups.length > 0) {
+              this.store.select(selectedServiceGroupHealth).subscribe((sgHealth) => {
```
msorens, line 254: This is a big red flag that would eventually cause memory problems: you should almost never use nested subscriptions and this one is 4 levels deep! Use a `combineLatest` instead to make it a single subscription.

Not directly related: be sure to add a `takeUntil(this.isDestroyed)` before the `subscribe` as well. Please review the rest of the file as there are a couple others missing as well.

components/automate-ui/src/app/pages/service-groups/service-groups.component.ts
```diff
@@ -246,21 +246,20 @@ export class ServiceGroupsComponent implements OnInit, OnDestroy {
...
+      if (sgStatus === 'loadingSuccess' && sSgStatus === 'loadingSuccess') {
+          if (serviceGroups.length > 0) {
+              if (sgHealth.total === 0) {
+                this.onServiceGroupSelect(null, serviceGroups[0].id);
               }
-            });
           }
-        });
       }
     });
```
msorens, line 264: With the `combineLatest` clearing up the nested subscription problem, the code is sufficient, but we can clean it up a bit further. Using the convention of applying filters instead of using conditionals, here we can substantially reduce the cyclomatic complexity by eliminating all the remaining nested conditionals.
```
      filter(([sSgStatus, sgStatus, serviceGroups, sgHealth]) =>
        sgStatus === EntityStatus.loadingSuccess
        && sSgStatus === EntityStatus.loadingSuccess
        && serviceGroups.length > 0
        && sgHealth.total === 0
      ),
      takeUntil(this.isDestroyed)
    ).subscribe(([_sSgStatus, _sgStatus, serviceGroups, _sgHealth]) =>
      this.onServiceGroupSelect(null, serviceGroups[0].id));
```

scottopherson: @msorens speaking of conventions, I've noticed a mixture of `this.isDestroyed$` and `this.isDestroyed`. I'm not sure what we'd want to prefer using (other than `isDestroyed$` seems more prevalent I think?)

Should `$` be appended to only `Observable` types, or appended to any type that can be subscribed to? (`Observable`, `Subject`, `ReplaySubject`, etc). I lean towards the latter at the moment

msorens: Great question, @scottopherson ! I have been kicking that can down the road because I thought it was just for Observables but was not really sure... A quick web search just now revealed people say Observable in one breath but then usually go on to add "... and streams". So I am happy to oblige and go with your suggestion of any type that can be subscribed to. 👍

scottopherson: @msorens ah yea, I'm seeing the "and streams" as well.
Yeah to me seems like `$` is useful for immediately identifying what can be done with a certain value (`subscribe`, `useWith$ | async`, use within `pipe` chain) and possible pitfalls that can occur with said value (memory leaks). Though seeing the types for these values in our IDEs helps with that as well 🤷

Issue #2762, opened 2020-01-30
Guard Add Users buttons on team add user modal


bcmdarroch:
## Describe the bug
Right now there is no mechanism for hiding the `Add Users` button on the team-details page when the logged-in user has no permissions.
<img width="655" alt="Screen Shot 2020-01-30 at 9 47 00 AM" src="https://user-images.githubusercontent.com/21015366/73475679-ad72ae80-4345-11ea-8f7b-27d53e5e6838.png">

This will require refactoring [the user-table](https://github.com/chef/automate/blob/master/components/automate-ui/src/app/modules/user/user-table/user-table.component.html) a bit. Right now we're passing [`overridePermissions` input to the table on team-details](https://github.com/chef/automate/blob/master/components/automate-ui/src/app/modules/team/team-details/team-details.component.html#L52-L58). This was because at the time we did not have parameterized introspection. Now we do! So let's fix this.

## To Reproduce
1. start Automate 
2. create a policy
```
echo '{
   "name": "testing",
   "id": "test1",
   "members": [
     "user:local:test"
   ],
   "statements": [
     {
       "effect": "ALLOW",
       "actions": ["iam:*:list", "iam:*:view"],
       "projects": [
         "*"
       ]
     }
   ],
   "projects": []
 }' >> pol.json

export TOK=`chef-automate iam token create ADM --admin`
curl -kH "api-token: $TOK" -X POST -d @pol.json https://localhost/apis/iam/v2beta/policies | jq
```
3. log in to the UI as admin
4. create user `test`
5. log in as user test
6. navigate to https://a2-dev.test/settings/teams/admins

## Expected behavior
If a logged-in user is allowed to see a team and its users but is not allowed to manage team membership, they will not see the `Add Users` button on the team details page.

PR #3736, opened 2020-05-19
#2762 guard add users button on team details view

srenatus:

Before, the **Add Users** button on the team details page has always been shown.

With this change, it will only be shown of the user has got sufficient privileges to actually do that.

### Definition of done

- [X] hide the add users button if they don't have permission to add users to a team
- [x] if they happen to navigate to the add users page of a team, display this message on the page: It looks like you don't have permission to add users to this team. Reach out to your administrator or contact Chef Support for help. with Chef Support being an open in new tab link to our support page
- [X] changed/added tests pass

#2762 

- rebuild automate-ui in your preferred way
- sign in as admin, go to a team's details:

Screenshot 2020-05-19 at 15 58 15

- create a user and the policy mentioned in #2762, sign in as that user, go to the team's details page again, find the button gone:

Screenshot 2020-05-19 at 15 59 20

- as `test` user, navigate to https://a2-dev.test/settings/teams/editors/add-users, notice:

image


components/automate-ui/src/app/modules/user/user-table/user-table.component.spec.ts
```diff
@@ -41,12 +31,26 @@ describe('UserTableComponent', () => {
...
-    it('should populate the permission paths', () => {
+  describe('when {get,create}PermissionsPath are not passed', () => {
+    it('should populate the permission paths using baseUrl', () => {
+      component.baseUrl = baseUrl;
+      component.ngOnInit();
       fixture.detectChanges();
-      expect(component.getPermissionsPath).toEqual([baseUrl, 'get']);
-      expect(component.createPermissionsPath).toEqual([baseUrl, 'post']);
+      expect(component.getPermissionsPath).toEqual([baseUrl, 'get']);     // TODO fails
+      expect(component.createPermissionsPath).toEqual([baseUrl, 'post']); // TODO fails
```
srenatus, line 40: I don't understand yet what I've done wrong here. The previous test setup seemed odd to me, I've tried making this more analogous to other specs. However, I've broken something in the process. Weirdly, too, the test below passes.

components/automate-ui/src/app/modules/user/user-table/user-table.component.spec.ts
```diff
@@ -1,38 +1,28 @@
...
-@Component({
-  selector: 'app-test-user-table-wrapper',
-  template: '<app-user-table [baseUrl]="baseUrl"></app-user-table>'
-})
-class TestUserTableWrapperComponent {
-  baseUrl = baseUrl;
-}
-
-describe('UserTableComponent', () => {
+fdescribe('UserTableComponent', () => {
```
srenatus, line 8: TODO: remove

srenatus:
⚠️ the "no access" page is wrong, there should not be a heading and buttons.

srenatus:
I've tried something... but now the close button (top right) is gone:

image


I suppose I could use some help here 😅

components/automate-ui/src/app/modules/user/user-table/user-table.component.spec.ts
msorens: `ngOnInit` is called when the component is initialized (in the `beforeEach` block), so setting `component.baseUrl` in your test comes too late. Several approaches to remedy.
If we expect `baseUrl` to change, then we need to accommodate it coming later--like in your test--which would be done with `ngOnChanges`. However, I think it is reasonable to say that we require `baseUrl` to be immutable--set only during instantation. (Probably should comment that!). 

More appropriate then, is to set `baseUrl` in your test _in the template_. Which, in a way, is more "real world" testing, too. Create a `TestHost`, complete with template.

A simple example: https://github.com/chef/automate/blob/0e6b4cf3b23f1ffb3bfaf3b5604ff1ddb9ca984b/components/automate-ui/src/app/components/message-modal/message-modal.component.spec.ts#L7-L22

A more elaborate example:
https://github.com/chef/automate/blob/e3ca5ae7a14539f2d097a542652fe0720cdeb9f4/components/automate-ui/src/app/components/form-control/form-control.directive.spec.ts#L63-L78

msorens:
There are several ways to resolve your "no permissions" page issue. I have pushed a commit with what is probably the simplest approach.

components/automate-ui/src/app/modules/user/user-table/user-table.component.html
```diff
@@ -1,11 +1,11 @@
-<app-authorized [allOf]="getPermissionsPath" [overrideVisible]="overridePermissionsCheck">
```
msorens, line 1: Looks like this gets rid of the last consumer of `overrideVisible` so please remove that from authorized.component.ts.

components/automate-ui/src/app/modules/team/team-add-users/team-add-users.component.ts
```diff
@@ -183,4 +187,11 @@ export class TeamAddUsersComponent implements OnInit, OnDestroy {
     this.openUserModal.emit(true);
   }
 
+  triggerFullPageMessage() {
+    // `fired` variable just minimizes browser's workload since that gets called continuously
```
srenatus, line 191: ❓ What kind of optimization is that? Do we do that in many places?

components/automate-ui/src/app/modules/user/user-table/user-table.component.spec.ts
srenatus: Ok, so the odd thing I hadn't seen before (and thoughtlessly removed) was actually the right thing. Gotcha. Thanks.

srenatus:
@msorens thanks again -- I've updated the preamble, this should be good to go now.

srenatus:
✔️ Verified "add users" request payload (no unknown fields sent) and updated the API docs example payloads.

components/automate-ui/src/app/modules/team/team-add-users/team-add-users.component.ts
msorens: No, it is something new. See my soon-to-be-posted warning in #a2-team.

susanev: the suspense!

components/automate-gateway/api/iam/v2/request/teams.proto
```diff
@@ -21,7 +21,7 @@ message CreateTeamReq {
       json_schema: {
         required: ["id", "name"]
       },
-      example: { value: '{"id": "test-id", "name": "My Test Team", "projects": ["project1", "project2"]}'}
+      example: { value: '{"id": "my-team-id", "name": "My Test Team", "projects": ["project1", "project2"]}' }
```
bcmdarroch, line 24: is this related to the issue or just a bonus fix?

components/automate-ui/src/app/modules/user/user-table/user-table.component.html
```diff
@@ -1,11 +1,11 @@
...
+<app-authorized [allOf]="getPermissionsPath">
   <ng-container *ngIf="showEmptyMessage">
     <div class="empty-state">
       <p>Add some users to get started!</p>
```
bcmdarroch, line 4: I noticed I still see this message when I'm logged in as the `test` user and I navigate to a team without any users
Screen Shot 2020-05-20 at 4 52 35 PM

susanev: nice catch, we should leverage the similar msg we display on projects. ill add that here later tonight.

components/automate-gateway/api/iam/v2/request/teams.proto
srenatus: Bonus. I've been checking the payloads to eventually remove the opt-out from our UI code. 😄

components/automate-ui/src/app/modules/user/user-table/user-table.component.html
srenatus: ℹ️ I'm looking into this now (Tuesday morning)

srenatus: ❓ I'm not sure I see what "similar msg" you've been referring to, @susanev.

When browsing a project's details as the test user, I find this, no message at all:
image

srenatus: 💡 You've meant this one:

image

I'll draft a similar message here that we can discuss later :)

srenatus: image

No, I believe this will also be shown on the users list view if there's **no local users at all**. However, I believe we can ignore that case -- there's supposed to always be a local admin at least. 🤔 I still don't like this. @msorens what do you think? How would we got about refactoring this properly?

msorens: My preference would be to fix this since it violates [OCP](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle): this class should not know anything about its consumers.  So I pushed up [a fix](https://github.com/chef/automate/pull/3736/commits/db68dff2e00b5e9894f499ee3033d42446dae85f) that let's the two consumers provide different messages.

➡️ @susanev Please review the text of both messages in that commit.

Now, that said, separately there still might be some display logic to adjust. Here is what I found what I visited each resource page when logged in as a user having no permissions then navigating directly to each page:
image
Clearly there is some work to be done there on four of those pages, but for the purposes of _this_ PR we might want to address why the Users list is blank and is not showing the message (with or without my commit).

susanev: the messages look good to me, thank you everyone for the assists!!

srenatus: @msorens good catches! Let's tackle those separately, please. ;)

srenatus: > [...] but for the purposes of this PR we might want to address why the Users list is blank and is not showing the message (with or without my commit).

👍 I'll look into it.

srenatus: @msorens can't reproduce. I've created a policy giving my user ALLOW on `*`, and DENY on `iam:users:create`, and see this:
image

msorens: @sr You exercised a different test case than the one I specified:

> when logged in as a user having no permissions then navigating directly to each page

(No strong preference whether that gets fixed in this PR, though.)

PR #3737, opened 2020-05-19
return count of nodes even if nodes result is 0

vjeffrey:

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/compliance-service/reporting/relaxting/nodes.go
```diff
@@ -175,23 +175,23 @@ func (backend *ES2Backend) GetNodes(from int32, size int32, filters map[string][
...
-		return nodes, TotalNodeCounts{
-			Total:   total,
-			Passed:  nodeSummary.Compliant,
-			Failed:  nodeSummary.Noncompliant,
-			Skipped: nodeSummary.Skipped,
-			Waived:  nodeSummary.Waived,
-		}, nil
 	}
+	//take status out of filters similar to why we remove them from suggestions.. we want to always see what's available in this context
+	delete(filters, "status")
```
vjeffrey, line 180: this section was inside the "if results greater than 0" block, so it was never getting called if nodes returned 0

vjeffrey:
Screen Shot 2020-05-19 at 08 52 40

stevendanna:
You have some detail in your PR message that isn't in your commit message.

I'd personally appreciate it if we spent a minute or two to get that information into the commit message in a form that is clear to people in the future who might not be as up-to-speed on the details of the problem.

vjeffrey:
note: this does not fully fix the issue where the ui goes to no data state on selection of that filter, it just corrects the api call to ensure that all counts are returned even if there are 0 nodes in the list.
for the ui issue, im guessing there's logic somewhere that says if 0 nodes are in the list, show empty state msg. if we change that to check if all returned counts are 0, or total is 0, that should fix things i think

components/compliance-service/api/tests/02_nodes_spec.rb
```diff
@@ -658,6 +658,19 @@
...
+    # Test filters by skipped status - expect 0 nodes, but expect counts to not be 0
+    actual_nodes = GRPC reporting, :list_nodes, Reporting::Query.new(filters: [
+      Reporting::ListFilter.new(type: 'environment', values: ['DevSec Prod beta']),
+      Reporting::ListFilter.new(type: 'status', values: ['skipped']),
+      Reporting::ListFilter.new(type: 'end_time', values: ['2018-03-04T23:59:59Z'])
+    ])
+    expected_nodes = {
+      "total" => 2,
+      "totalFailed" => 1,
+      "totalPassed" => 1
```
stevendanna, line 670: Can we explicitly put in the empty array of nodes or will that cause a problem? Without it is is kinda hard to understand that you are also asserting that the nodes array is empty.  Perhaps moving the bit in the comment above after the `-` down here would help if we can't explicitly put an empty array of nodes in this assertion.

vjeffrey: the empty list is being returned; it's just the test that doesnt show it bc of how it translates the stuff to json from grpc

vjeffrey: Screen Shot 2020-05-19 at 08 59 39

vjeffrey: ya, ill move the comment!

vjeffrey: do you have any suggestions for the future fix? maybe returning on the api response the applied filters list or something?  we need counts to not respect status filters to enable status filtering in ui, and this is a pattern we have on many apis, so if we do find it awkward i would want to fix it everywhere

vjeffrey:
i rebased and pushed the ui change to complete the fix

vjeffrey:
Screen Shot 2020-05-19 at 10 01 20

components/automate-ui/src/app/pages/+compliance/shared/reporting/report-data.service.ts
```diff
@@ -122,6 +122,13 @@ export class ReportDataService {
     return false;
   }
 
+  isZero(count: number): boolean {
+    if (count === 0) {
+      return true;
+    }
+    return false;
```
msorens, line 129: More concisely:
```suggestion
    return (count === 0);
```

vjeffrey: oh ya! good call!

PR #3739, opened 2020-05-19
single_local_all_inplace_upgrade: remove SAML flag

srenatus:
It seems like this wasn't set up in our Okta, so let's just remove it.


srenatus:
Will merge when green

PR #3748, opened 2020-05-19
Replace newlines from chef-server-admin-key.txt with \n

jeremiahsnapp:

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/terraform/test-environments/modules/chef_automate_install/templates/install_chef_automate_cli.sh.tpl
+++ b/terraform/test-environments/modules/chef_automate_install/templates/install_chef_automate_cli.sh.tpl
@@ -126,26 +126,30 @@ configure_retention() {
 
 configure_automate_infra_views() {
   if chef-automate dev grpcurl automate-gateway list | grep "chef.automate.api.infra_proxy.InfraProxy" &> /dev/null; then
-      chef_server_admin_key=$(cat "/hab/chef-server-admin-key.txt")
+      chef_server_admin_key="$(</hab/chef-server-admin-key.txt tr '\n' ':' | sed 's/:/\\n/g')"
       server_id="auto-deployed-chef-server"
       org_id="auto-deployed-chef-org"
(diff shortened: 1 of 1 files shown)


terraform/test-environments/modules/chef_automate_install/templates/install_chef_automate_cli.sh.tpl
```diff
@@ -126,26 +126,30 @@ configure_retention() {
 
 configure_automate_infra_views() {
   if chef-automate dev grpcurl automate-gateway list | grep "chef.automate.api.infra_proxy.InfraProxy" &> /dev/null; then
-      chef_server_admin_key=$(cat "/hab/chef-server-admin-key.txt")
+      chef_server_admin_key="$(</hab/chef-server-admin-key.txt tr '\n' ':' | sed 's/:/\\n/g')"
```
vjeffrey, line 129: so we translate the newline to a : and then replace all colons with \\n -- right?

vjeffrey: what is the < at the beginning for?

jeremiahsnapp: yes, that's correct

btm: @vjeffrey `cat "/hab/chef-server-admin-key.txt | FOO"` and `</hab/chef-server-admin-key.txt FOO` do the same thing. An easier way to understand the second one is to write it the "normal way" which would be `FOO < /hab/chef-server-admin-key.txt`

So this could also be
```
cat /hab/chef-server-admin-key.txt | tr '\n' ':' | sed 's/:/\\n/g'
```
or
```
tr '\n' ':' < /hab/chef-server-admin-key.txt | sed 's/:/\\n/g'
```

This is a bunch of hoop jumping to replace a real newline character with the two characters '\n'.

vjeffrey: 👍 thanks for the explanation!

PR #3762, opened 2020-05-20
[automate-1882] Optionally add project teams if checkbox selected

tylercloke:

Optionally add project teams if checkbox selected.

We should be good on length violation concerns since teams have super non-restrictive IDs atm:

https://github.com/chef/automate/blob/master/api/interservice/teams/teams.proto#L16-L20

- [ ] Tests added/updated?
- [ ] Docs added/updated?

If one of the team creation requests fail, it still creates the project:

Screen Shot 2020-05-20 at 3 19 19 PM

Resulting teams created:

Screen Shot 2020-05-20 at 3 32 07 PM

Modal style changes:

Screen Shot 2020-05-20 at 3 38 14 PM

### DIFF
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.scss
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.scss
@@ -27,6 +27,10 @@ chef-modal {
 
       .checkbox-margin {
         margin-bottom: 30px;
+
+        #add-policy-checkbox {
+          margin-bottom: 30px;
+        }
       }
 
       .dropdown-margin {
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.spec.ts
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.spec.ts
@@ -112,7 +112,7 @@ describe('CreateObjectModalComponent', () => {
     expect(store.dispatch).toHaveBeenCalledWith(new GetPolicies());
   });
 
-  it('upon opening, "also create" checkboxs are checked', () => {
+  it('upon opening, "also create" checkboxes are checked', () => {
     component.createForm = new FormBuilder().group({
       name: '',
       addPolicies: '',
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.spec.ts
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.spec.ts
@@ -112,18 +112,24 @@ describe('CreateObjectModalComponent', () => {
     expect(store.dispatch).toHaveBeenCalledWith(new GetPolicies());
   });
 
-  it('upon opening, "also create policies" checkbox is checked', () => {
+  it('upon opening, "also create" checkboxs are checked', () => {
     component.createForm = new FormBuilder().group({
       name: '',
-      addPolicies: ''
+      addPolicies: '',
+      addTeams: ''
     });
-    component.createForm.controls.addPolicies.setValue(false); // force it to start at false
+
+    // force them to start at false
+    component.createForm.controls.addPolicies.setValue(false);
+    component.createForm.controls.addTeams.setValue(false);
+
     component.createProjectModal = true;
 
     component.ngOnChanges(
       { visible: new SimpleChange(false, true, true) });
 
     expect(component.createForm.controls.addPolicies.value).toEqual(true);
+    expect(component.createForm.controls.addTeams.value).toEqual(true);
   });
 
 });
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
@@ -37,6 +37,7 @@ export class CreateObjectModalComponent implements OnInit, OnDestroy, OnChanges 
   public modifyID = false; // Whether the edit ID form is open or not.
   public conflictError = false;
   public addPolicies = true;
+  public addTeams = true;
   public projectsUpdatedEvent = new EventEmitter();
   public policiesUpdatedEvent = new EventEmitter();
   public UNASSIGNED_PROJECT_ID = ProjectConstants.UNASSIGNED_PROJECT_ID;
@@ -76,7 +77,9 @@ export class CreateObjectModalComponent implements OnInit, OnDestroy, OnChanges 
       this.policiesUpdatedEvent.emit();
 
       if (this.createProjectModal) {
-        this.updatePolicyCheckbox(true); // default to checked upon opening
+         // default to checked upon opening
+        this.updatePolicyCheckbox(true);
+        this.updateTeamCheckbox(true);
       }
     }
   }
@@ -92,6 +95,11 @@ export class CreateObjectModalComponent implements OnInit, OnDestroy, OnChanges 
   updatePolicyCheckbox(event: boolean): void {
     this.addPolicies = event;
     this.createForm.controls.addPolicies.setValue(this.addPolicies);
+  }
+
+  updateTeamCheckbox(event: boolean): void {
+    this.addTeams = event;
+    this.createForm.controls.addTeams.setValue(this.addTeams);
   }
 
   handleNameInput(event: KeyboardEvent): void {
--- a/components/automate-ui/src/app/pages/project/list/project-list.component.ts
+++ b/components/automate-ui/src/app/pages/project/list/project-list.component.ts
@@ -85,9 +85,7 @@ export class ProjectListComponent implements OnInit, OnDestroy {
     this.store.pipe(
       select(createStatus),
       takeUntil(this.isDestroyed),
-      filter(state => {
-        return this.creatingProject && state === EntityStatus.loadingSuccess;
-      }))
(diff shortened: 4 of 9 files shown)


components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -44,6 +45,12 @@ export class ProjectListComponent implements OnInit, OnDestroy {
       'RULES_APPLIED': 'Applied'
   };
 
+  readonly teamsToCreate = {
+    'project-owner': 'Project Owner',
```
bcmdarroch, line 49: these should probably all be plural

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -82,11 +90,26 @@ export class ProjectListComponent implements OnInit, OnDestroy {
...
+
         this.closeCreateModal();
 
         // This is issued periodically from projects-filter.effects.ts; we do it now
         // so the user doesn't have to wait.
         this.store.dispatch(new LoadOptions());
+
+        if (createTeamsWasChecked) {
+          for (const [teamID, teamName] of Object.entries(this.teamsToCreate)) {
+            this.store.dispatch(new CreateTeam({
```
bcmdarroch, line 106: I think to keep this component dumber, we'd want to handle this in an ngrx effect

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.scss
```diff
@@ -27,6 +27,10 @@ chef-modal {
 
       .checkbox-margin {
         margin-bottom: 30px;
+
+        #add-policy-checkbox {
+          margin-bottom: 30px;
+        }
```
srenatus, line 33: I don't get it. What does the class thing do? Is this added (to 60px), effectively?

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
```diff
@@ -94,6 +97,11 @@ export class CreateObjectModalComponent implements OnInit, OnDestroy, OnChanges
     this.createForm.controls.addPolicies.setValue(this.addPolicies);
   }
 
+  updateTeamCheckbox(event: boolean): void {
```
srenatus, line 100: [nit] `event` seems a bit oddly named, perhaps? If it's a boolean, I'd go with `checked`. But I might be missing something.

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -82,11 +90,26 @@ export class ProjectListComponent implements OnInit, OnDestroy {
       }))
       .subscribe(() => {
         this.creatingProject = false;
+
+        const createTeamsWasChecked = this.createProjectForm.controls['addTeams'].value;
```
srenatus, line 94: I don't know if and where we've settled on wrt. `this.createProjectForm.controls['addTeams'].value` vs `this.createProjectForm.controls.addTeams.value`

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -82,11 +90,26 @@ export class ProjectListComponent implements OnInit, OnDestroy {
       }))
```
srenatus, line 90: [nit] ☝️ 
```typescript
filter(state => this.creatingProject && state === EntityStatus.loadingSuccess)
```

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -55,7 +62,8 @@ export class ProjectListComponent implements OnInit, OnDestroy {
       name: ['', Validators.required],
       id: ['',
         [Validators.required, Validators.pattern(Regex.patterns.ID), Validators.maxLength(48)]],
-      addPolicies: [true]
+      addPolicies: [true],
```
srenatus, line 65: ❓ Does this not imply a default of true? I feel like we've stated that twice (with `addTeams = true` above)

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
```diff
@@ -37,6 +37,7 @@ export class CreateObjectModalComponent implements OnInit, OnDestroy, OnChanges
   public modifyID = false; // Whether the edit ID form is open or not.
   public conflictError = false;
   public addPolicies = true;
+  public addTeams = true;
```
srenatus, line 40: Hmm is `create-object-modal` still the right name? It seems quite specific to have policy- and team-creation-related code in here 🤔

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
tylercloke: I agree, once again following the policies pattern. I'll update em both.

components/automate-ui/src/app/pages/project/list/project-list.component.ts
tylercloke: Yeah, it's a bit weird. This is the constructor for the FormBuilder which exists outside of the generic add modal component and is passed in. `addTeams` in that component is just a boolean we keep in sync with the checkbox state that we then use to drive the form. The first thing that component does is change these form values to true so defaulting them here does feel pointless.

components/automate-ui/src/app/pages/project/list/project-list.component.ts
tylercloke: Yeah no idea which is considered more idiomatic. Would be a good #ui-team question.

components/automate-ui/src/app/pages/project/list/project-list.component.ts
tylercloke: Hmmm I could see going either way on this one. Wanting to create a few teams as a result of how the user used the form feels like it makes sense to trigger in here. It feels weird that creating teams would be something the projects store would do.

On the other hand, I know we wanna keep these components as simple as possible within reason.

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.scss
tylercloke: Checkbox margin contains two checkboxes. This is adding a margin below the first checkbox whereas the checkbox margin is adding a margin below the container relative to the rest of the modal.

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.spec.ts
```diff
@@ -112,18 +112,24 @@ describe('CreateObjectModalComponent', () => {
     expect(store.dispatch).toHaveBeenCalledWith(new GetPolicies());
   });
 
-  it('upon opening, "also create policies" checkbox is checked', () => {
+  it('upon opening, "also create" checkboxs are checked', () => {
```
msorens, line 115:
```suggestion
  it('upon opening, "also create" checkboxes are checked', () => {
```

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
```diff
@@ -76,7 +77,9 @@ export class CreateObjectModalComponent implements OnInit, OnDestroy, OnChanges
       this.policiesUpdatedEvent.emit();
 
       if (this.createProjectModal) {
```
msorens, line 79: Regarding special casing in this code, here's a thought that makes it slightly less of a violation of the Open/Closed Principle: don't check who the caller is, rather just let the caller specify a need:
```suggestion
      if (this.associateTeamsAndPolicies) {
```

components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -77,16 +85,29 @@ export class ProjectListComponent implements OnInit, OnDestroy {
...
         // This is issued periodically from projects-filter.effects.ts; we do it now
         // so the user doesn't have to wait.
         this.store.dispatch(new LoadOptions());
+
+        if (createTeamsWasChecked) {
+          for (const [teamID, teamName] of Object.entries(this.teamsToCreate)) {
+            this.store.dispatch(new CreateTeam({
+              id: `${projectID}-${teamID}`,
+              name: `${projectName} ${teamName}`,
+              projects: [projectID]
```
msorens, line 107: Is there a reason we cannot make the addition of teams work more similar to the addition of policies? Seems odd that the two return different types of notifications, as shown.

If that approach is not viable, though, a comment on the current implementation:
Would like to have the notifications include the team id, so we do not see the repetitive "Created a new team". That is, should say "Created team "p1-foo-project-owners", etc.

image

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
srenatus: `s/pattern we've established/hack/`  😉

srenatus: Anyways, it's fine! 😃

PR #3786, opened 2020-05-23
[documentation] Expand upon/clarify Elasticsearch heap size config

nrgetik:
This change accomplishes consistent messaging between our documentation and Elastic's recommendations around JVM heap configuration. It also provides for a more specific example which illustrates the format that the heap amount should be specified in

Signed-off-by: Thomas Gordon <tgordon@chef.io>

This was made in response to recent feedback from users/operators.

- [ ] Tests added/updated?
- [x] Docs added/updated?


stevendanna:
Do we want to change our calculation: https://github.com/chef/automate/blob/master/api/config/elasticsearch/config_request.go#L196

nrgetik:
> Do we want to change our calculation: https://github.com/chef/automate/blob/master/api/config/elasticsearch/config_request.go#L196

I don't know that the calculation needs to change. I was motivated to tweak our documentation because of the original language "up to 32GB" (this is not so cut-and-dry per https://www.elastic.co/guide/en/elasticsearch/guide/current/heap-sizing.html#compressed_oops) and the lack of a specific example. I think the calculation still establishes a safe-but-useful initial heap size, assuming the minimum memory requirement is met. I'd be interested to know what others think, though

components/automate-chef-io/content/docs/configuration.md
```diff
@@ -277,20 +277,25 @@ Uncomment and change settings as needed, and then run `chef-automate config patc
 [elasticsearch.v1.sys.runtime]
 # max_locked_memory = "unlimited"
 # es_java_opts = ""
-# NOTE: heapsize should be set to 50% of available RAM up to 32g
-# See: https://www.elastic.co/guide/en/elasticsearch/guide/current/heap-sizing.html
-# heapsize = "1g"
+# NOTE: see https://www.elastic.co/guide/en/elasticsearch/guide/current/heap-sizing.html
+# for important guidance regarding the configuration of the heap size setting
```
susanev, line 281:
```suggestion
# NOTE: [Elasticsearch Heap Sizing](https://www.elastic.co/guide/en/elasticsearch/guide/current/heap-sizing.html)
# contains important guidance regarding the configuration of the heap size setting
```

wait, idk our standard for having links in the config comments 🤷

components/automate-chef-io/content/docs/configuration.md
```diff
@@ -277,20 +277,25 @@ Uncomment and change settings as needed, and then run `chef-automate config patc
...
+# heapsize = "4g"
 ```
 
 #### Setting Elasticsearch Heap
 
-Per the Elasticsearch documentation, the [Elasticsearch heap size](https://www.elastic.co/guide/en/elasticsearch/guide/current/heap-sizing.html) should be up to 50% of the available RAM, up to 32GB. 
-To set the Elasticsearch heap size, create a TOML file that contains the partial configuration below.
-Uncomment and change settings as needed, and then run `chef-automate config patch </path/to/your-file.toml>` to deploy your change.
+The Elasticsearch heap size can, and in most cases should, be set to 50% of the available system
+memory. However, there are important caveats covered in the [Elasticsearch documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/heap-size.html)
```
susanev, line 288:
```suggestion
memory. However, there are important caveats covered in the [Elasticsearch heap size documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/heap-size.html)
```

susanev:
the private build on this is failing, my guess is its hugo related. but im sure @mjingle knows and will be able to fix it.

components/automate-chef-io/content/docs/configuration.md
nrgetik: I like the wording change, but yeah, is using `[]()` a thing here since it won't actually be rendered as a link?

kagarmoe: You would need a shortcode for the note to render correctly, in which case you're right that the link wouldn't work correctly. In this case, # NOTE will probably render as a H1, which is  not what we want.

I think we're OK dropping `# NOTE:` and using the rest of the sentence. Bonus points if you can get it into the active voice.

moutons:
> Do we want to change our calculation: https://github.com/chef/automate/blob/master/api/config/elasticsearch/config_request.go#L196

@stevendanna some time back we discussed perhaps changing the assignments in config_request.go to something like
```
EsHeapMaxGB = 28
EsHeapPortion = 0.5
```
when the system has more than 8 GB of memory, and I think that's still a good idea to keep in y'all's back pocket. It's not critical, though, as by that time a customer will probably have had a conversation with someone about tuning specific to their system.

components/automate-chef-io/content/docs/configuration.md
nrgetik: @kagarmoe My point was that this portion is all already inside `<pre><code>` in the rendered documentation, so is there a purpose to keeping `[]()` or removing `NOTE:` ? I might not understand all the places where this content ends up.

kagarmoe: Ah, I didn't see that it was inside of a code fence. The backstory is that supplying the link was a compromise & was included here for people who would just copy-paste the example and not read the whole doc. You can keep it or remove it, I don't have a strong opinion either way. I'm happy to see us update the heapsize.

kagarmoe: The code fences/<pre><code> mean that it renders exactly as we see it. @susanev may be trying to remove `see`? However, the idiom is accepted in major house style guides, meaning Microsoft & Google. This is an example of a case where trying to avoid it is unhelpful. From the docs perspective, you can reject this change.

susanev: oh no i was just trying to get link text in there but no worries if its not possible

susanev:
is this good to merge when its green?

components/automate-chef-io/content/docs/configuration.md
kagarmoe: Sadly, it won't work because of the code fences.

components/automate-chef-io/content/docs/configuration.md
```diff
@@ -290,7 +290,7 @@ which should be carefully reviewed and considered.
 
 For the purposes of an example, a system with 32GB of memory can have its Elasticsearch heap
 size set to `16g`; to do so, one would first create a TOML file that contains the partial
-configuration below, and then issue a `chef-automate config patch </path/to/your-file.toml>` to
+configuration below, uncomment and change settings as neededand, then issue a `chef-automate config patch </path/to/your-file.toml>` to
```
kagarmoe, line 293:
```suggestion
configuration below, uncomment and change settings as needed and, then run `chef-automate config patch </path/to/your-file.toml>` to
```

kagarmoe: @susanev Ended up with a merge conflict, but here's the new fix.

nrgetik: @kgarmoe Looks like this was merged already but there are still changes to be made. It was merged without your correction to `neededand`, and also, that entire statement doesn't make sense here because of the switch to a presentation of an example as described in the paragraph (this makes more sense to users). Nothing is commented out in the configuration snippet, so `uncomment and change settings as needed and` should just be removed.

susanev: ill open a follow up in a few minutes

PR #3810, opened 2020-05-27
[Automate-3439] Add sections to the policies dropdown

msorens:

To improve the UX for creating tokens, this PR enhances the generic dropdown component to accommodate multiple sections. With these changes, the projects dropdown continues to use only a single section, but the policies dropdown is now separated into chef-managed and custom sections. Furthermore, everything is sorted, except that the chef-managed `ingest-access` policy is positioned at the top of the list since that is the one that will usually be needed.

image

The video shows the full contents of the dropdown and how it behaves:

dropdown sections

And this set of unit tests (new tests highlighted) reveals the requirements that were implemented. 

- The yellow items are related to ordering. The generic dropdown component (ResourceDropdownComponent) no longer sorts its inputs, allowing its consumers (ProjectsDropdownComponent and CreateObjectModalComponent) to customize their ordering.
- The blue items cover the requirements from a behavioral perspective. The available project list is asynchronously refreshed, even when the create-token dialog is open. If that list changes while the dialog is open, experiments showed that it is rather confusing if the list in the dialog changes seemingly spontaneously. Not only would there possibly be new projects appearing or existing projects disappearing, but the user's checkmarks so far would be reset. So I opted to make the list refresh only if the create-token dialog is completely closed. Thus, most of the tests in the blue section cover behavior when (a) the project list remains constant and (b) the project list updates, confirming the behavior is the same either way.

image

#### PR Bonus
In a recent slack post, I described a performance pitfall with Angular when using function calls instead of properties to provide rendering data in a template. That is, try to always avoid calling an explicit function like this in a template:
```
There are {{ count() }} projects.
```
Also avoid "hidden" functions (aka computed properties):
```
There are {{ count }} projects.
```
... where `count` is something like `get count() { return x + y; }`. Rather, try to always use "fixed" properties:
```
There are {{ count }} projects.
```
... where `count` is a property declared like `let count: number;` or `let count = 509;`.

This PR introduces my first exploration into applying this in practice.
My recommendation, shown with examples here, is to use the "declared properties" approach rather than the "pure pipes" approach, mainly because even if you only want a custom pipe in a single file, it needs to be specified globally (in the sense of putting it in a module file).

The file resource-dropdown.component.ts had two culprits: `disabled` and `allFilteredResourceCount`.

Here is how I fixed `disabled` -- because it was a computed property, no changes were needed in the template. The trick is just to realize that its value only changes when the `@Input` named `resources` changes, so I just needed to set it in `ngOnChanges`.

image

The `allFilteredResourceCount` is also a computed property, but still provides a slightly different use case. Its value was not dependent upon some `@Input`, but dependent upon a local, `filteredResources`. So I only needed to update it when `filteredResources` changed.

image

NA

### :+1: Definition of Done
- The policies dropdown when creating a token now has sections. The chef managed section has the ingest policy at the top.
- If either section is reduced to no entries by applying a filter, the section header goes away as well.
- The projects dropdown remains with a single section, though it has no section header, so it looks exactly as it did before.
- Closing and re-opening either dropdown retains whatever was checked.
- Closing and re-opening the modal resets all items to unchecked.
- Filtering has no effect on checked items nor on the dropdown label.

After rebuilding/reloading the UI, navigate to Settings >> Tokens >> Create Token.
Use the projects and policies dropdowns, with and without filters.

If you want to exercise how asynchronous project updates affect the dropdowns, adjust projects-filter.effects.ts::POLLING_INTERVAL_IN_SECONDS to e.g. 10 seconds. Then, add or remove a project from the command line or separate browser window.

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
@@ -6,7 +6,12 @@
           <chef-form-field>
             <label>
               <span class="label">Name <span aria-hidden="true">*</span></span>
-              <input chefInput formControlName="name" type="text" (keyup)="handleNameInput($event)" id="name-input" data-cy="create-name" autocomplete="off">
+              <input chefInput formControlName="name"
+                id="name-input"
+                data-cy="create-name"
+                type="text"
+                (keyup)="handleNameInput($event)"
+                autocomplete="off">
             </label>
             <chef-error *ngIf="(createForm.get('name').hasError('required') || createForm.get('name').hasError('pattern')) && createForm.get('name').dirty">
               Name is required.
@@ -18,7 +23,12 @@
           <chef-form-field>
             <label>
               <span class="label">ID <span aria-hidden="true">*</span></span>
-              <input chefInput formControlName="id" type="text" (keyup)="handleIDInput($event)" id="id-input" data-cy="create-id" autocomplete="off"/>
+              <input chefInput formControlName="id"
+                id="id-input"
+                data-cy="create-id"
+                type="text"
+                (keyup)="handleIDInput($event)"
+                autocomplete="off"/>
             </label>
             <chef-error *ngIf="createForm.get('id').hasError('maxlength') && createForm.get('id').dirty">
               ID must be 64 characters or less.
@@ -41,7 +51,11 @@
             <span data-cy="id-label">{{ this.createForm?.value.id }}</span>
           </div>
           <chef-toolbar>
-            <chef-button tertiary (click)="modifyID = true" id="edit-button-object-modal" data-cy="edit-button">Edit ID</chef-button>
+            <chef-button tertiary
+              id="edit-button-object-modal"
+              data-cy="edit-button"
+              (click)="modifyID = true"
+            >Edit ID</chef-button>
           </chef-toolbar>
         </div>
         <div class="checkbox-margin" *ngIf="createProjectModal">
@@ -69,8 +83,11 @@
           </app-projects-dropdown>
         </div>
         <div id="button-bar">
-          <chef-button primary id="create-button-object-modal" data-cy="save-button"
-            [disabled]="!createForm?.valid || creating || conflictError" (click)="createObject()">
+          <chef-button primary
+            id="create-button-object-modal"
+            data-cy="save-button"
+            [disabled]="!createForm?.valid || creating || conflictError"
+            (click)="createObject()">
             <chef-loading-spinner *ngIf="creating"></chef-loading-spinner>
             <span *ngIf="!creating">Create {{ objectNoun | titlecase }}</span>
             <span *ngIf="creating">Creating {{ objectNoun | titlecase }}...</span>
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.spec.ts
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.spec.ts
@@ -70,37 +70,38 @@ describe('CreateObjectModalComponent', () => {
   });
 
   it('upon opening, checked status of all projects is set to false', () => {
-    component.projects = {
-      'proj1':
-        { id: 'proj1', name: 'proj1', type: 'CHEF_MANAGED', status: 'NO_RULES', checked: true },
-      'proj3':
-        { id: 'proj3', name: 'proj3', type: 'CUSTOM', status: 'EDITS_PENDING', checked: false },
-      'proj2':
-        { id: 'proj2', name: 'proj2', type: 'CUSTOM', status: 'RULES_APPLIED', checked: true }
-    };
+    component.checkedProjectIDs = ['proj1', 'proj2', 'proj3'];
 
     component.ngOnChanges(
       { visible: new SimpleChange(false, true, true) });
 
-    Object.values(component.projects).forEach(p => {
-      expect(p.checked).toBe(false);
-    });
+    // By resetting this list, that triggers ProjectsDropdownComponent to set all projects false
+    expect(component.checkedProjectIDs.length).toEqual(0);
   });
 
   it('upon opening, checked status of all policies is set to false', () => {
     component.policies = [
-        { id: 'proj1', name: 'proj1', type: 'CHEF_MANAGED', members: [], projects: [],
-          checked: true },
-        { id: 'proj3', name: 'proj3', type: 'CUSTOM', members: [], projects: [], checked: false },
-        { id: 'proj2', name: 'proj2', type: 'CUSTOM', members: [], projects: [], checked: true }
+      {
+        title: 'section1',
+        itemList: [
+          { id: 'proj1', name: 'proj1', checked: true }
+        ]
+      },
+      {
+        title: 'section2',
+        itemList: [
+          { id: 'proj1', name: 'proj1', checked: true },
+          { id: 'proj3', name: 'proj3', checked: false },
+          { id: 'proj2', name: 'proj2', checked: true }
+        ]
+      }
     ];
 
     component.ngOnChanges(
       { visible: new SimpleChange(false, true, true) });
 
-    Object.values(component.policies).forEach(p => {
-      expect(p.checked).toBe(false);
-    });
+    component.policies.forEach(p =>
+      p.itemList.forEach(i => expect(i.checked).toBe(false)));
   });
 
   it('upon opening, dispatches a call to refresh policies', () => {
--- a/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.spec.ts
+++ b/components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.spec.ts
@@ -6,7 +6,8 @@ import { MockComponent } from 'ng2-mock-component';
 
 import { ngrxReducers, runtimeChecks, NgrxStateAtom } from 'app/ngrx.reducers';
 import { using } from 'app/testing/spec-helpers';
-import { GetPolicies } from 'app/entities/policies/policy.actions';
+import { GetPolicies, GetPoliciesSuccessPayload, GetPoliciesSuccess } from 'app/entities/policies/policy.actions';
+import { Policy, IAMType } from 'app/entities/policies/policy.model';
 import { CreateObjectModalComponent } from './create-object-modal.component';
 
 describe('CreateObjectModalComponent', () => {
@@ -69,7 +70,7 @@ describe('CreateObjectModalComponent', () => {
     });
   });
 
-  it('upon opening, checked status of all projects is set to false', () => {
+  it('upon opening, checked status of all PROJECTS is set to false', () => {
     component.checkedProjectIDs = ['proj1', 'proj2', 'proj3'];
 
     component.ngOnChanges(
@@ -79,7 +80,7 @@ describe('CreateObjectModalComponent', () => {
     expect(component.checkedProjectIDs.length).toEqual(0);
   });
 
-  it('upon opening, checked status of all policies is set to false', () => {
+  it('upon opening, checked status of all POLICIES is set to false', () => {
     component.policies = [
       {
         title: 'section1',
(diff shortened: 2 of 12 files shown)


components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.html
```diff
@@ -6,7 +6,12 @@ <h2 slot="title">Create {{ objectNoun | titlecase }}</h2>
           <chef-form-field>
             <label>
               <span class="label">Name <span aria-hidden="true">*</span></span>
-              <input chefInput formControlName="name" type="text" (keyup)="handleNameInput($event)" id="name-input" data-cy="create-name" autocomplete="off">
+              <input chefInput formControlName="name"
```
msorens, line 9: PR bonus This file: line breaks for readability

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
```diff
@@ -67,12 +80,15 @@ export class CreateObjectModalComponent implements OnInit, OnDestroy, OnChanges
...
+        this.store.dispatch(new GetPolicies()); // refresh in case of updates
+      }
 
-      Object.values(this.projects).forEach(p => p.checked = false); // reset projects
+      this.checkedProjectIDs = []; // reset projects
       this.projectsUpdatedEvent.emit();
 
-      Object.values(this.policies).forEach(p => p.checked = false); // reset policies
+      this.policies.forEach(policy =>
+        policy.itemList.forEach(p => p.checked = false)); // reset policies
```
msorens, line 91: Projects and policies are different because projects are fed to ProjectsDropdownComponent that has some further handling before passing on to a child ResourceDropdownComponent. Policies, on the other hand, are fed directly to ResourceDropdownComponent so all processing is here.

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.ts
```diff
@@ -111,13 +127,11 @@ export class CreateObjectModalComponent implements OnInit, OnDestroy, OnChanges
...
   closeEvent(): void {
     this.modifyID = false;
-    this.checkedProjectIDs = [];
     this.close.emit();
   }
 
   createObject(): void {
     this.createClicked.emit();
-    this.checkedProjectIDs = [];
   }
```
msorens, line 135: Moved the resetting to the opener instead of the closers (which is where all the other resetting is).

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
```diff
@@ -40,16 +40,17 @@ export class ProjectsDropdownComponent implements OnInit, OnDestroy, OnChanges {
     this.store.select(assignableProjects)
       .pipe(takeUntil(this.isDestroyed))
       .subscribe((assignable: ProjectsFilterOption[]) => {
-        this.projects =
-          assignable.map(p => {
-            return <ResourceChecked>{
+        this.projects = [{
+          itemList: assignable.map(p => {
```
msorens, line 44: The underlying data structure was refined to contain sections needed for policies. Projects, though, do not need multiple sections, so all data goes in one.

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
```diff
@@ -100,14 +110,33 @@ export class ResourceDropdownComponent implements OnInit, OnChanges {
 
   closeDropdown(): void {
     this.dropdownState = 'closed';
-    this.onDropdownClosing.emit(
-      this.resources.filter(r => r.checked).map(r => r.id));
+    const flattenedIDs = [].concat(
+      ...this.snapshotResources.map(
+        resource => resource.itemList.filter(r => r.checked).map(r => r.id)));
+    this.onDropdownClosing.emit(flattenedIDs);
```
msorens, line 116: Returned data is still the same--a list of checked IDs. But the structure is more complicated so generating that list now takes a bit more work.

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.spec.ts
```diff
@@ -30,133 +32,356 @@ describe('ResourceDropdownComponent', () => {
...
-        expect(option.checked).toEqual(checked);
-      });
+  it('renders resources WITHOUT sorting ', () => { // thus allowing custom ordering
+    const resources = [genResourceList(
+      genResource('zz', false),
+      genResource('cc', true),
+      genResource('aa', false),
+      genResource('bb', true)
+    )];
+      component.objectNounPlural = 'policies';
```
srenatus, line 61: [nit] indentation looks off a little

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.spec.ts
```diff
@@ -30,133 +32,356 @@ describe('ResourceDropdownComponent', () => {
...
+    original: ResourceCheckedSection[], changed: ResourceCheckedSection[], firstChange) {
+    // Programmatic change to an @Input does *not* trigger ngOnChanges like in the actual view,
+    // so have to do that explicitly
+    component.resources = changed;
+    component.ngOnChanges({ resources: new SimpleChange(original, changed, firstChange) });
+  }
+
+  function genResourceList(...resources: ResourceChecked[]): ResourceCheckedSection {
+    return {
+      title: 'any',
```
srenatus, line 380: Isn't the title optional? What do we gain from specifying one here?

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
```diff
@@ -100,14 +110,33 @@ export class ResourceDropdownComponent implements OnInit, OnChanges {
 
   closeDropdown(): void {
     this.dropdownState = 'closed';
-    this.onDropdownClosing.emit(
-      this.resources.filter(r => r.checked).map(r => r.id));
+    const flattenedIDs = [].concat(
+      ...this.snapshotResources.map(
+        resource => resource.itemList.filter(r => r.checked).map(r => r.id)));
```
srenatus, line 115: Huh that's an interesting construct. What's the benefit of `[].concat(...this.snapshotResources.map(/* ... */)` over `this.snapshotResources.flatMap(/* ... */)`? (⚠️ untested)

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
```diff
@@ -100,14 +110,33 @@ export class ResourceDropdownComponent implements OnInit, OnChanges {
...
   }
 
   handleFilterKeyUp(): void {
-    this.filteredResources = this.resourcesInOrder.filter(resource =>
-      resource.id.toLowerCase().indexOf(this.filterValue.toLowerCase()) > -1
-    );
+    for (let i = 0; i < this.filteredResources.length; i++) {
+      this.filteredResources[i].itemList =
+        this.snapshotResources[i].itemList.filter(r =>
+          r.name.toLowerCase().indexOf(this.filterValue.toLowerCase()) > -1);
```
srenatus, line 123: Hmm so we're testing if it's a (case-insensitive) substring? 💡

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
```diff
@@ -100,14 +110,33 @@ export class ResourceDropdownComponent implements OnInit, OnChanges {
...
+    }
+    this.allFilteredResourcesCount = this.calculateAllFilteredResourcesCount();
+  }
+
+  private resetFilteredResources(): void {
+
+    // Not deep and not shallow!
+    // Need the individual resources to point to the same objects
+    // so that when one is checked, it is checked in BOTH structures.
+    this.filteredResources = [].concat(
```
srenatus, line 133: Same as my question above :)

srenatus: ...has it to do with the specific type of copy you want there? 🤔

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
```diff
@@ -129,13 +158,31 @@ export class ResourceDropdownComponent implements OnInit, OnChanges {
...
+
+  private calculateAllFilteredResourcesCount(): number {
+    return this.filteredResources.reduce(
+      (sum, group) => sum + group.itemList.length, 0);
+  }
+
+  private get allResourcesCount(): number {
+    return this.resources.reduce(
+      (sum, group) => sum + group.itemList.length, 0);
+  }
```
srenatus, line 180: Hmm would be be possible to define a `length()` or `count()` method on `ResourceCheckedSection[]`?

components/automate-ui/src/app/components/projects-dropdown/projects-dropdown.component.ts
```diff
@@ -59,14 +60,19 @@ export class ProjectsDropdownComponent implements OnInit, OnDestroy, OnChanges {
...
-      // Need to trigger OnChanges in ResourceDropdownComponent
-      // so we cannot just update the checked property of the existing array elements.
-      this.projects = this.projects
-        .map(p => ({
-          ...p,
-          checked: (changes.checkedProjectIDs.currentValue as string[]).includes(p.id)
-        }));
+    if (changes.checkedProjectIDs) {
+      // cannot proceed unless we have received the available projects list from the back end
+      if (this.projects) {
```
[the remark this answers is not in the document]
msorens: Glad you asked! That was quite deliberate on my part. The canonical form of `ngOnChanges` should be
```
  ngOnChanges(changes: SimpleChanges): void {
    if (changes.thing1) {
      // process change of thing1
    }
    if (changes.thing2) {
      // process change of thing2
    }
    etc...
  }
```
If the first statement in the process block is another conditional then, sure, you could combine them, but I think it is cleaner to maintain the guard on the change standing by itself.

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
msorens: matching case-insensitively, yes.

components/automate-ui/src/app/components/create-object-modal/create-object-modal.component.spec.ts
```diff
@@ -69,43 +70,45 @@ describe('CreateObjectModalComponent', () => {
...
 
     component.ngOnChanges(
       { visible: new SimpleChange(false, true, true) });
 
-    Object.values(component.policies).forEach(p => {
-      expect(p.checked).toBe(false);
-    });
+    component.policies.forEach(p =>
+      p.itemList.forEach(i => expect(i.checked).toBe(false)));
   });
```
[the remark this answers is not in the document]
msorens: Oh, yes! I think it does matter--thank you for bringing this up!
I took your idea and went a bit further with it. In case of failure, this shows the expected set of values, the actual set of values, and... an explanation, which I just realized Jasmine knows how to do. 
```
    const checkedStatusOfEveryPolicy =
      component.policies.flatMap(p => p.itemList.map(i => i.checked));
    expect(checkedStatusOfEveryPolicy)
       .withContext('some polices were not reset to false')
       .toEqual(Array(checkedStatusOfEveryPolicy.length).fill(false));
```

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.spec.ts
msorens: That causes testing failures: when comparing resources to filteredResources...
```
 expect(component.filteredResources).toEqual(resources);
```
Each item in `filteredResource` has a title set to `undefined`, but if we omit the title here on L380, the title is completely absent on each item in `resources`, so they do not match. Will add a comment.

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
msorens: None; just did not come to mind. Thanks!

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
msorens: Nothing esoteric; changed this to flatMap, too. 👍

components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.ts
msorens: I do not think you can add methods to an interface or type; would have to make it a full-fledged class, so not sure that is worth it. However, I did remove the duplicated code.

PR #3825, opened 2020-05-28
Standardize remaining "required" indicators in UI

msorens:

Quick scan of the code base revealed just these few
that were not using the accessibility-friendly
"aria-hidden" on the required marker (*).

Quoting the MDN [web docs on accessibility](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute):
> Adding aria-hidden="true" to an element removes that element and all of its children from the accessibility tree. This can improve the experience for assistive technology users by hiding:
> 
> purely decorative content, such as icons or images
> duplicated content, such as repeated text
> offscreen or collapsed content, such as menus
> 

NA

### :+1: Definition of Done
All uses of the asterisk to indicate a required field are wrapped in an element specifying `aria-hidden="true"` in the UI templates.

No visible changes.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/modules/license/license-apply/license-apply.component.html
+++ b/components/automate-ui/src/app/modules/license/license-apply/license-apply.component.html
@@ -46,7 +46,7 @@
       <div class="note">Required fields are marked with an asterisk (*).</div>
       <form class="apply" [formGroup]="applyForm" (ngSubmit)="onApplyLicenseFormSubmit()" novalidate>
         <chef-form-field>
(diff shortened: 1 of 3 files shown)


srenatus:
Seems to call for some deduplication, though. How about having a `chef-label` that has a `[required]` input, and renders accordingly?

msorens:
Good suggestions! Will commit as is and whoever has a chance can iterate on those. 😁

PR #3835, opened 2020-05-29
make sign in pages more accessible

### LINKED ISSUE #3814: sign in page a11y issues
## overview
fix some a11y issues on the sign in page. these were found using the [wave chrome extension](https://chrome.google.com/webstore/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh?hl=en-US).

## details
- [x] chef logo needs alt text of `Chef Software Logo`
- [x] needs document language
- [x] add h1 of sign in to the page thats missing a heading
- [x] add at least one page region
- [x] associate labels with form controls
- [x] fix this missed feedback #3827 (comment)

susanev:

we recently did an accessibility pass and there were a bunch on the sign-in pages, this fixes them.

fixes #3814 

### :+1: Definition of Done
stuff should work and look the same for the past part, there's one new h1 added on the fist page and one color change to the error message to make it pass the color contrast check

`rebuild components/automate-dex`
some changes to automate-ui were also added so also rebuild that

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-dex/web/templates/login.html
+++ b/components/automate-dex/web/templates/login.html
@@ -19,11 +19,11 @@
       <div>
         {{ range $c := .Connectors }}
         <form class="theme-form-row" method="post" action="{{ $c.URL }}">
-            {{ if eq $c.ID "local" }}
-            <button type="submit" class="dex-btn-text"><span class="dex-btn-icon dex-btn-icon--{{ $c.ID }}"></span>Sign in as a local user</button>
-            {{ else }}
-            <button type="submit" class="dex-btn-text"><span class="dex-btn-icon dex-btn-icon--{{ $c.ID }}"></span>Sign in with {{ $c.Name }}</button>
-            {{ end }}
+          {{ if eq $c.ID "local" }}
+          <button type="submit" class="dex-btn-text"><span class="dex-btn-icon dex-btn-icon--{{ $c.ID }}"></span>Sign in as a local user</button>
+          {{ else }}
+          <button type="submit" class="dex-btn-text"><span class="dex-btn-icon dex-btn-icon--{{ $c.ID }}"></span>Sign in with {{ $c.Name }}</button>
+          {{ end }}
         </form>
         {{ end }}
       </div>
--- a/components/automate-dex/web/templates/login.html
+++ b/components/automate-dex/web/templates/login.html
@@ -13,7 +13,7 @@
   <body>
     <main class="theme-panel">
       <div class="logo">
-        <img class="theme-navbar__logo" alt="Chef Company Logo" src="{{url .ReqPath "static/img/automate-blue-d9789f4b.svg" }}">
+        <img class="theme-navbar__logo" alt="Chef" src="{{url .ReqPath "static/img/automate-blue-d9789f4b.svg" }}">
       </div>
       <h1>User Sign In</h1>
       <div>
--- a/components/automate-dex/web/templates/login.html
+++ b/components/automate-dex/web/templates/login.html
@@ -1,7 +1,7 @@
 <!-- rework of coreos login.html-->
 <!-- rendered using https://golang.org/pkg/html/template/-->
 <!DOCTYPE html>
-<html>
+<html lang="en">
   <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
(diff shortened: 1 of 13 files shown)


susanev:
@msorens @srenatus im hoping i can pass this to one of you to finish. prior to this pr once you choose local sign in it takes you to the page and no error is presented at the top unless you enter something bad. after this pr when you use the local sign in button and it takes you to the next page the error shows right away before you have a chance to fill in the inputs. thoughts on how to fix it?

components/automate-dex/web/templates/login.html
```diff
@@ -1,7 +1,7 @@
 <!-- rework of coreos login.html-->
 <!-- rendered using https://golang.org/pkg/html/template/-->
 <!DOCTYPE html>
-<html>
+<html lang="en">
```
SEAjamieD, line 4: valid - possibly better to us `en-US`.  This declares United States English.  Open to debate.

susanev: automate-ui uses en, so i. think lets keep them consistent for now.

components/automate-dex/web/templates/login.html
```diff
@@ -10,25 +10,23 @@
...
-              <span class="dex-btn-text">Sign in as a local user</span>
-              {{ else }}
-              <span class="dex-btn-text">Sign in with {{ $c.Name }}</span>
-              {{ end }}
-          </div>
-        </a>
+  <body>
+    <main class="theme-panel">
+      <div class="logo">
+        <img class="theme-navbar__logo" alt="Chef Company Logo" src="{{url .ReqPath "static/img/automate-blue-d9789f4b.svg" }}">
```
SEAjamieD, line 16: suggest: `alt="Chef - Home"`

components/automate-dex/web/templates/password.html
```diff
@@ -11,43 +11,43 @@
...
+      <a class="back" href="javascript:history.back()">
+        <span>
+          <i class="fa fa-arrow-left" aria-hidden="true"></i>
+        </span>
+        Back
+      </a>
+      {{ end }}
+      <div class="theme-panel">
+        <div class="logo">
+          <img class="theme-navbar__logo" alt="Chef Company Logo" src="{{url .ReqPath "static/img/automate-blue-d9789f4b.svg" }}">
```
SEAjamieD, line 26: same suggestion as above

components/automate-dex/web/templates/login.html
susanev: can you say more about why? ill want to update this for automate-ui as well

SEAjamieD: ooh huge mistake here - thought that was a `link` - hadn't fired up the site yet.

I'd suggest just `alt="Chef"`

SEAjamieD: As far as the Why - I think its important to note, and you probably already know this, but for anyone else reading.. that what you have now is not wrong.

In my opinion the toughest part of a11y is that there is a lot of ambiguity in the best practices of it.  I tend to lean in favor of removing any description that isn't already known.  In this case, the Logo is sitting in an  tag, so we already know its an image.  And we know we are on the Chef software site.  

You could also argue that Chef is too common and would be misleading in that it might be a Chef of the cooking variety.

I guess all of this is to say - thats just my opinion on the fact - But I'm open to being convinced either way 🤷‍♂️ 

https://inviqa.com/blog/writing-effective-alt-text-images

components/automate-dex/web/templates/login.html
```diff
@@ -10,25 +10,23 @@
...
-  </div>
+      <h1>User Sign In</h1>
+      <div>
+        {{ range $c := .Connectors }}
+        <form class="theme-form-row" method="post" action="{{ $c.URL }}">
+            {{ if eq $c.ID "local" }}
+            <button type="submit" class="dex-btn-text"><span class="dex-btn-icon dex-btn-icon--{{ $c.ID }}"></span>Sign in as a local user</button>
+            {{ else }}
+            <button type="submit" class="dex-btn-text"><span class="dex-btn-icon dex-btn-icon--{{ $c.ID }}"></span>Sign in with {{ $c.Name }}</button>
+            {{ end }}
```
SEAjamieD, line 26: [nit]: looks like indentation may be a little deep here

components/automate-dex/web/templates/login.html
susanev: found this too: https://webaim.org/techniques/alttext/#logos

components/automate-dex/web/theme/styles.css
```diff
@@ -20,12 +20,16 @@ html, body {
...
+button {
+    display: block;
     background-color: #3864f2;
     color: #fff;
     width: 279px;
     height: 53px;
     margin: 0 auto;
+    border-radius: 4px;
+    border: 0;
+    line-height: 33px;
```
SEAjamieD, line 32: [nit]: looks like indentation is deep - linting will catch though

susanev: theres no lint on the ui pages for automate-dex...

susanev: if you by chance want to push up a commit that makes the indentation whatever the standard yall use is, that would be rad

components/automate-dex/web/templates/login.html
SEAjamieD: exactly 😂  - i like https://www.deque.com/accessible-development/ a lot as well.

susanev: oh wait this should prolly be Chef Automate now that i think about it to make it distinct from the other products.. will update again

components/automate-dex/web/theme/styles.css
SEAjamieD: yeah no prob! - i'll slack before I push

srenatus:
> prior to this pr once you choose local sign in it takes you to the page and no error is presented at the top unless you enter something bad. after this pr when you use the local sign in button and it takes you to the next page the error shows right away before you have a chance to fill in the inputs. 

That's quite odd, since there's nothing dynamic here -- it's rendered by the server, and `Invalid` shouldn't be true when you click on "local user". 🤔 I don't see the problem yet.

msorens:
@srenatus I have never looked at the dex templates before; where is the code that processes those dex templates and fills in the variables (BackLink, Invalid, etc.) ?

srenatus:
@msorens there's nothing special to it, it's using the stdlib template package. The variables are provided when the server renders the template. At any rate, the HTML change here wouldn't possibly change those values. I'll have a look later.

components/automate-dex/web/templates/login.html
```diff
@@ -10,25 +10,23 @@
...
       </div>
-      {{ end }}
-    </div>
-  </div>
+      <h1>User Sign In</h1>
+      <div>
+        {{ range $c := .Connectors }}
+        <form class="theme-form-row" method="post" action="{{ $c.URL }}">
+          {{ if eq $c.ID "local" }}
+          <button type="submit" class="dex-btn-text"><span class="dex-btn-icon dex-btn-icon--{{ $c.ID }}"></span>Sign in as a local user</button>
```
srenatus, line 23: :x: This is the problem: `button type="submit"` switches the request for `/password.html` from a GET to a POST. With GET, we'd trigger [this flow in dex](https://github.com/dexidp/dex/blob/2ca992e9b359eff69f9886d10620f255c4cff2f2/server/handlers.go#L330-L333), with `POST`, we'll trigger [that one](https://github.com/dexidp/dex/blob/2ca992e9b359eff69f9886d10620f255c4cff2f2/server/handlers.go#L363-L391).

The latter one expects that it's the entered credentials being `POST`ed.

I'll play with [formmethod](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/submit#formmethod) to see if it gets us out of this predicament.

srenatus: Looks like it might not be possible to retain the query string. I've tried various combinations, like having `<form method="get"` and put the `$c.URL` into the buttons' `formaction`, it will always be overridden.

Some random SO post on the issue: https://stackoverflow.com/a/9882750.

The problem is that the auth request ID is embedded in `$c.URL`, and we don't have access to that; so we can't use it to reconstruct the query string for the GET.

I guess we could come up with some JS to extract it, but maybe there's other a11y-conform ways to get the buttons into a shape where they both are accessible and don't need `type="submit"`? 😦

susanev:
@srenatus after you review and if it looks good, feel free to merge, thank you!

PR #3840, opened 2020-05-29
set manager to "chef" on chef managed nodes

vjeffrey:
🔩 Description: What code changed, and why?
this change adds the manager type to the nodemanager process node flow, meaning
we can assign any node which is coming through ingest service the manager "chef" since we know that node is managed by chef
any node that is automate or aws-ec2 or azure-api etc will get the type on the processnode message, meaning that we dont have to look up the type of the manager based on the id (save an extra db call)

⛓️ Related Resources
#3515

👍 Definition of Done
manager type is on process node message
nodes managed by chef are labelled as such

👟 How to Build and Test the Change
rebuild compliance, ingest, nodemanager
load some nodes (chef_load_nodes, load_compliance_reports, chef_load_compliance_nodes, load_scan_jobs)
see the manager type for the nodes is correct!


components/compliance-service/ingest/pipeline/publisher/node_manager_publisher.go
```diff
@@ -90,6 +90,10 @@ func gatherInfoForNode(in message.Compliance) (*manager.NodeMetadata, error) {
 		tags = append(tags, &common.Kv{Key: "environment", Value: in.Report.GetEnvironment()})
 	}
 
+	mgrType := in.Report.GetAutomateManagerType()
+	if in.Report.GetSourceFqdn() != "" {
```
vjeffrey, line 94: if there's a value for the chef server, we know the node is managed by chef

components/nodemanager-service/pgdb/node-managers-process-node.go
```diff
@@ -127,20 +127,12 @@ func (db *DB) ProcessIncomingNode(node *manager.NodeMetadata) error {
 		return errors.Wrap(err, "ProcessIncomingNode unable to marshal projects data")
 	}
 
-	var mgrType string
-	if node.ManagerId != "" {
```
vjeffrey, line 131: this is the db call we can now remove since we have the type

components/nodemanager-service/pgdb/node-managers.go
```diff
@@ -339,24 +339,15 @@ func validateNodeManagerFilters(filters []*common.Filter) error {
...
+				if !isValidNodeManagerType(item) {
+					return &errorutils.InvalidError{Msg: fmt.Sprintf("Invalid manager type filter: %s. manager_type must be one of the following: 'aws-ec2', 'aws-api', 'azure-api', 'chef', '', 'azure-vm', 'gcp-api' ", item)}
 				}
 			}
 		}
 	}
 	return nil
 }
 
-func isValidManagerType(managerType string) bool {
```
vjeffrey, line 351: we had two functions basically looking at the same thing and i kept getting tripped up on it. figured it's better to just use one!

components/nodemanager-service/pgdb/nodes.go
```diff
@@ -690,7 +690,7 @@ func validateNodeFilters(filters []*common.Filter) error {
 func isValidNodeManagerType(item string) bool {
 	switch item {
 	// empty string covers legacy nodes (nodes created before we started associating all manual nodes with "automate" as a manager and nodes brought over from a1)
-	case "aws-ec2", "aws-api", "azure-api", "azure-vm", "automate", "":
+	case "aws-ec2", "aws-api", "azure-api", "azure-vm", "automate", "gcp-api", "chef", "":
```
vjeffrey, line 693: woops! we didn't have gcp-api listed as valid! it has been a valid option for a long time

components/compliance-service/ingest/pipeline/publisher/node_manager_publisher.go
lancewf: If the value is "localhost" it is not managed by a chef server it is an effortless or chef solo report.

components/automate-ui/src/app/pages/+compliance/+scanner/state/scanner.effects.ts
```diff
@@ -45,7 +45,7 @@ export class ScannerEffects {
       const nodesStatusFilter = { key: 'status', values: ['reachable', 'unreachable', 'unknown'] };
       // ensures that we only collect nodes that belong to a manager
```
lancewf, line 46: This comment is stale now.

components/compliance-service/ingest/pipeline/publisher/node_manager_publisher.go
vjeffrey: whaddya think? should we distinguish between chef-server managed and just chef solo/effortless?

lancewf: I don't think we should for now. But it looks like could when and if we want to.

components/nodemanager-service/pgdb/node-managers-process-node.go
```diff
@@ -127,20 +127,12 @@ func (db *DB) ProcessIncomingNode(node *manager.NodeMetadata) error {
...
-		mgrType, err = db.SelectStr(sqlGetManagerTypeFromId, node.ManagerId)
-		if err != nil {
-			logrus.Warnf("unable to find manager for node")
```
lancewf, line 134: Should we still warn here if `node.GetManagerType()` is empty?

components/nodemanager-service/pgdb/node-managers.go
```diff
@@ -339,24 +339,15 @@ func validateNodeManagerFilters(filters []*common.Filter) error {
 		switch filter.Key {
 		case "manager_type":
 			for _, item := range filter.Values {
-				if !isValidManagerType(item) {
-					return &errorutils.InvalidError{Msg: fmt.Sprintf("Invalid manager type filter: %s. manager_type must be one of the following: 'aws-ec2', 'aws-api', 'azure-api'", item)}
+				if !isValidNodeManagerType(item) {
+					return &errorutils.InvalidError{Msg: fmt.Sprintf("Invalid manager type filter: %s. manager_type must be one of the following: 'aws-ec2', 'aws-api', 'azure-api', 'chef', '', 'azure-vm', 'gcp-api' ", item)}
```
lancewf, line 343: So, the manager type can still be empty?

e2e/cypress/integration/api/nodemanager/node_manager_type.spec.ts
```diff
@@ -0,0 +1,38 @@
+import { uuidv4 } from '../../../support/helpers';
```
lancewf, line 1: :1st_place_medal:

components/nodemanager-service/pgdb/node-managers.go
vjeffrey: no new ones should be empty, but we need to keep this around for the old nodes. they should get updated once a new run/scan comes thru but if there are no new runs/scans for the node it will be stuck in the empty manager state

e2e/cypress/integration/api/nodemanager/node_manager_type.spec.ts
vjeffrey: all your helper functions/the way you wrote the tests made it SO easy to write this one up, thank you!!

components/nodemanager-service/pgdb/node-managers-process-node.go
vjeffrey: hmm, good call ya

PR #3847, opened 2020-05-31
fix too small text a11y

### LINKED ISSUE #3844: text size too small
## overview
text size needs to be larger than 10px to pass a11y assessment. lets figure out ways to fix these.

## details
- [ ] hidden label on the apps search is too small, lets add an aria-label instead
- [ ] event feed table date/time is too small, make padding on time 4pt instead of 5pt, change event feed day and date to 11px, with line-height of 16px
- [ ] apps percentage symbol on apps page is 10px, and needs to be 11px
- [ ] event feed right panel has text that is smaller than 11px

susanev:
we had some pages with text that was smaller than 11px which causes a11y assessments to fail. this fixes it.

fixes #3844

### :+1: Definition of Done
text size is all at least 11px

- [ ] Tests added/updated?
- [ ] Docs added/updated?

##### before
event-feed-before

event-feed-right-panel-before

##### after
event-feed-after

event-feed-right-panel-after

##### before
apps-before

##### after
apps-after

##### before
client-run-node-detail-diff-before

##### after
client-run-node-detail-diff-after

### DIFF
--- a/components/automate-ui/src/app/page-components/delta-viewer/delta-viewer.component.scss
+++ b/components/automate-ui/src/app/page-components/delta-viewer/delta-viewer.component.scss
@@ -17,7 +17,7 @@
   }
 
   .d2h-tag {
-    font-size: 9px;
+    font-size: 11px;
   }
 
   .d2h-diff-table {
--- a/components/automate-ui/src/app/page-components/event-feed-table/event-feed-table.component.html
+++ b/components/automate-ui/src/app/page-components/event-feed-table/event-feed-table.component.html
@@ -47,10 +47,10 @@
           </app-event-icon>
         </div>
 
-        <p class="event-group-text">
+        <h2 class="display4">
           <strong>{{groupedEvent.eventCount}} {{getEventTypeLabel(groupedEvent)}}</strong>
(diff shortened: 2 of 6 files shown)


susanev:
missed the text to small on the event feed right panel, will add that in a bit

SEAjamieD:
is 11px something we have decided on at Chef?  From what I understand there is no hard/fast rule on font size.  Are we going off of contrast?  We have a number of font sizes less than 11px, it appears but curious of a standard so I can help identify.

susanev:
@SEAjamieD the [wave chrome extension](https://chrome.google.com/webstore/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh?hl=en-US) i was using said you cannot have text size 10px or smaller. pretty much everyone recommends at least 16px but i dont think we are going to pull that off in this pr.

SEAjamieD:
no nothing i'm concerned about, more about the correct way for me to assess what changes might need to be made going forward.  Sounds like this is a catalyst and will hopefully lead to many changes done across a lot of unrelated future PRs

susanev:
yah im hoping going forward ux design can have a standard not to use anything smaller than 11px, but still need their review on this too. and prolly some convos with the team.

Issue #3216, opened 2020-03-26
fix ui error message when a user is not authorized to create a profile


susanev:
## Overview
Currently, if a user tries to create a profile in the UI when they arent authorized to do so they get an error message telling them to try again, but it wont matter how many times they try unless they are given permissions to do it. Let's update that error message to be a little bit more useful to the user.

## Repro steps
- Sign in as `viewer` with the password `chefautomate`
- Navigate to compliance/profiles, then the available tab
- Use one of the get buttons, and notice the error
- api response is `subject ["team:local:viewers" "user:local:viewer"] is not authorized to "compliance:profiles:create" resource "compliance:profiles:viewer"`
- UI error message is: `Failed to install profile. Please try again later.`

![ugh](https://user-images.githubusercontent.com/5489125/77700392-5833d180-6f71-11ea-9074-41b678a74978.png)

## Design Details
- Instead, lets display `You are not authorized to create compliance profiles.` or similar.
- Another approach would be to not show the get buttons at all if someone doesn't have permission, using app-authorized, but i think that would slow down the page load even more

vsingh-msys:
@susanev observed similar issues while login with admin credentials as well. 

![Screenshot from 2020-03-27 03-29-33](https://user-images.githubusercontent.com/44496937/77701113-8ea63680-6fdb-11ea-98e7-a13008282a4e.png)

PR #3860, opened 2020-06-01
Updated error message when user is not authorized to create profile.

samshinde:
Signed-off-by: samshinde <ashinde@chef.io>

Fixed UI error message when a user is not authorized to create a profile

https://github.com/chef/automate/issues/3216
### :+1: Definition of Done
Updated the error message from  `Failed to install profile. Please try again later.` to  `You are not authorized to create compliance profiles.` only if we got the Unauthorized response.

- Sign in as `viewer` with the password `chefautomate`
- Navigate to compliance/profiles, then the available tab
- Use one of the get buttons, and notice the error
- api response is `subject ["team:local:viewers" "user:local:viewer"] is not authorized to "compliance:profiles:create" resource "compliance:profiles:viewer"`
- The updated UI error message is: `You are not authorized to create compliance profiles.` only if we got the Unauthorized response.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

profileErrorMessage

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.html
+++ b/components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.html
@@ -10,6 +10,13 @@
             (dismissed)="hideDownloadError()">
             Failed to install profile. Please try again later.
           </app-notification>
+          <app-notification
+            *ngIf="authErrorVisible"
+            type="error"
+            timeout="5"
+            (dismissed)="hideAuthError()">
+            You are not authorized to create compliance profiles.
+          </app-notification>
         </div>
       </div>
 
--- a/components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.ts
+++ b/components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.ts
@@ -22,6 +22,7 @@ import { AvailableProfilesService } from 'app/services/profiles/available-profil
 import { ChefSessionService } from 'app/services/chef-session/chef-session.service';
 import { find } from 'lodash';
 import { ProductDeployedService } from 'app/services/product-deployed/product-deployed.service';
+import { HttpStatus } from 'app/types/types';
 
 interface Profile {
     name: String;
@@ -64,6 +65,7 @@ export class ProfileOverviewComponent implements OnInit, OnDestroy {
 
   // notification data
   downloadErrorVisible = false;
+  authErrorVisible = false;
 
   userProfilesDataLoaded = false;
   availableProfilesDataLoaded = false;
@@ -262,13 +264,22 @@ export class ProfileOverviewComponent implements OnInit, OnDestroy {
     observableForkJoin(installProfiles).pipe(
       takeUntil(this.isDestroyed))
       .subscribe(
-        () => this.refreshProfiles(),
-        () => this.showDownloadError()
+      () => this.refreshProfiles(),
+      (error) => this.showError(error)
       );
   }
 
-  showDownloadError() {
-    this.downloadErrorVisible = true;
+  showError(error) {
+    if (error.status === HttpStatus.FORBIDDEN) {
+      this.authErrorVisible = true;
+    } else {
+      this.downloadErrorVisible = true;
+    }
+    this.refreshProfiles();
+  }
+
+  hideAuthError() {
+    this.authErrorVisible = false;
   }
 
   hideDownloadError() {
--- a/components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.spec.ts
+++ b/components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.spec.ts
@@ -267,7 +267,7 @@ describe('ProfilesOverviewComponent', () => {
       describe('when response status is an 400', () => {
         beforeEach(() => {
           spyOn(component, 'refreshProfiles');
-          spyOn(component, 'showDownloadError');
+          spyOn(component, 'showError');
           spyOn(component.availableProfilesService, 'installMarketProfile')
             .and.returnValue(throwError({ status: 400 }));
           component.getProfiles(
@@ -280,7 +280,7 @@ describe('ProfilesOverviewComponent', () => {
         });
 
         it('displays the error', () => {
-          expect(component.showDownloadError).toHaveBeenCalled();
+          expect(component.showError).toHaveBeenCalled();
         });
       });
     });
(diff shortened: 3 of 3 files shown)


components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.html
```diff
@@ -8,7 +8,7 @@
             type="error"
             timeout="5"
             (dismissed)="hideDownloadError()">
-            Failed to install profile. Please try again later.
+            You are not authorized to create compliance profiles.
```
srenatus, line 11: ❓ Is that the only way that this could go wrong? I'd think that we might inspect the actual error response and perhaps look at its code? Or use its message.

samshinde: @susanev Can you give me your thoughts on that?

susanev: oh sorry, i dont know that. well need to ask in superteam, or work with @vjeffrey

vjeffrey: yes, other things can go wrong -- ill carve out some time to look at the err msgs returned today

susanev:
im gonna be on pto so dont wait on me to get this merged.

vjeffrey:
oof, i had forgotten about this pr, sorry. adding to my todo list

components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.html
vjeffrey: so, other things can go wrong, like getting an invalid request or a 404, but those things will likely not happen via the ui call..

srenatus: Not even if another user has removed the profile using the UI in the meantime? I think we should bubble up the actual error message like we do in other places... 🤔

vjeffrey: you can delete an available profile

vjeffrey:
so the only other error msgs you could get would be invalid request or a 404, but those things will likely not happen via the ui call. 404 wouldn't happen bc users can't delete "available" profiles. invalid request shouldn't happen from the ui bc we control the request. the only other thing that could happen would be systems being down.
we can still bubble up the msg in case, do a match in the ui to the expected "not authd" msg, and then just return the orig err msg if we don't match on the auth msg..

vjeffrey:
we could match the error string on "is not authorized" substring msg and return the auth msg when we find that, otherwise just return original error msg

samshinde:
> we could match the error string on "is not authorized" substring msg and return the auth msg when we find that, otherwise just return original error msg

@vjeffrey So it means we need to keep the original error string " Failed to install profile. Please try again later." and when the owner is not admin or owner is the only viewer then we need to show this error string You are not authorized to create compliance profiles.

vjeffrey:
yaa -- so we could like, set the error msg to the "failed to install profile" thing (the orig one), and then read the error msg and if the string includes "is not authorized" we know to give the auth error msg

samshinde:
> yaa -- so we could like, set the error msg to the "failed to install profile" thing (the orig one), and then read the error msg and if the string includes "is not authorized" we know to give the auth error msg

@vjeffrey I have updated the PR with new changes.

components/automate-ui/src/app/services/profiles/available-profiles.service.ts
```diff
@@ -33,7 +33,10 @@ export class AvailableProfilesService {
     const params = new HttpParams().set('owner', this.chefSessionService.username);
     const headers = new HttpHeaders({'Content-Type': 'application/json'});
 
-    return this.httpClient.post(url, body, {params, headers});
+    return this.httpClient.post(url, body, {params, headers}).pipe(
+      catchError((err) => {
+        return throwError(err);
```
srenatus, line 38: This looks suspicious. 🤔 What happens without this? Would the caller _not_ get an error thrown?

components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.ts
```diff
@@ -261,12 +263,27 @@ export class ProfileOverviewComponent implements OnInit, OnDestroy {
...
+      },
+      (error) => this.showAuthError(error)
       );
   }
 
+  showAuthError(error) {
+    if (error.status === HttpStatus.FORBIDDEN) {
+      this.authErrorVisible = true;
+      this.refreshProfiles();
+    }
```
srenatus, line 280: If I understand this correctly, an error *not related to auth* will be ignored here?

components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.ts
```diff
@@ -261,12 +263,27 @@ export class ProfileOverviewComponent implements OnInit, OnDestroy {
     });
     observableForkJoin(installProfiles).pipe(
       takeUntil(this.isDestroyed))
-      .subscribe(
-        () => this.refreshProfiles(),
-        () => this.showDownloadError()
+      .subscribe(data => {
+        if (data) {
+          this.refreshProfiles();
+          this.showDownloadError();
```
srenatus, line 269: I'm quite certain that this isn't correct. Before, it was
```js
      .subscribe(
        () => this.refreshProfiles(),
        () => this.showDownloadError()
```
so, there's *two* callbacks given to `subscribe`, and the second one is the error callback.

Now, we're giving *two* callbacks to `subscribe`, the error callback (second one) is `(error) => this.showAuthError(error)`, and the success callback is
```js
        if (data) {
          this.refreshProfiles();
          this.showDownloadError();
       }
```
**So, we'll always be displaying a download error?**

Can you verify this code change with a successful profile download? It should *not* display that download error.

vjeffrey:
>  if it wasn't using a service (very old code), but if it was wired up with ngrx. @vjeffrey any plans to address this tech debt?

if you'd like to open a card, the ui team can totally put it on their board, sure!

components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.ts
samshinde: @srenatus  Yes you were right I have updated my changes accordingly.

components/automate-ui/src/app/pages/+compliance/+profile/+profile-overview/profile-overview.component.ts
```diff
@@ -262,11 +264,24 @@ export class ProfileOverviewComponent implements OnInit, OnDestroy {
...
+      () => this.refreshProfiles(),
+      (error) => this.showError(error)
       );
   }
 
+  showError(error) {
+    if (error.status === HttpStatus.FORBIDDEN) {
+      this.authErrorVisible = true;
+    } else {
+      this.showDownloadError();
```
srenatus, line 276: I think it would be reasonable to inline this, here, too:
```suggestion
      this.downloadErrorVisible = true;
```
and remove `showDownloadError`. What do you think?

PR #3868, opened 2020-06-03
Fix race condition where cypress test only passed if other two notifications had already timed out

tylercloke:

Needed to pass multiple to the notification dismissal cypress code since there are three notifications for this case to dismiss.

### DIFF
--- a/e2e/cypress/integration/ui/settings/project_mgmt.spec.ts
+++ b/e2e/cypress/integration/ui/settings/project_mgmt.spec.ts
@@ -161,10 +161,12 @@ describe('project management', () => {
     cy.get('[data-cy=save-button]').click();
     cy.get('app-project-list chef-modal').should('not.be.visible');
 
-    // verify success notification and then dismiss it
+    // verify success notification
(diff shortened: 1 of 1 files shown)


e2e/cypress/integration/ui/settings/project_mgmt.spec.ts
```diff
@@ -161,10 +161,12 @@ describe('project management', () => {
     cy.get('[data-cy=save-button]').click();
     cy.get('app-project-list chef-modal').should('not.be.visible');
 
-    // verify success notification and then dismiss it
+    // verify success notification
     cy.get('app-notification.info')
       .contains(`Created project ${customWithPolsProjectID} and associated policies`);
-    cy.get('app-notification.info chef-icon').click();
+
+    // dismiss all three sucess notifications
```
srenatus, line 168:
```suggestion
    // dismiss all three success notifications
```

tylercloke: My spells have not been on this week xD

PR #3876, opened 2020-06-03
Typescript type check fix

msorens:

Continuing efforts to enable new type-checking capabilities delivered with Angular 9, this fixes one set of errors reported by enabling `strictInputTypes` in `angularCompilerOptions` here:
https://github.com/chef/automate/blob/6ef4dbe7945c7a661c136fe6f58815b7d59309ad/components/automate-ui/tsconfig.json#L28

The relevant error is:
```
Error: Type 'AbstractControl' is not assignable to type 'FormGroup'.
```

Most of the changes here, though, are just a bonus: several of the files lent themselves to a nice refactor, pulling out duplicate code to a `BaseIntegrationFormComponent` base class.

The affected pages in the UI:
1. Settings >> Node Integrations >> Create Integration
2. Compliance >> Scan Jobs >> Create Scan Job

More on enabling `strictInputTypes`:
There are a variety of other things to fix, too, before this switch can be set true. Here are a couple I looked at but could not resolve at this time.

(1) Several of the remaining errors are blocked by an Angular issue. I filed [this issue](https://github.com/angular/angular/issues/37401) that led to revealing that the Angular Language Service has not kept up with the improved type checking in Ivy. They manifest as these types of errors, and all stem from the fact that the local variable instantiated by an `ngFor` reverts to type `any` even when the collection it is iterating is strongly typed!
```
src/app/pages/project/rules/project-rules.component.html:86:34
- error TS2339: Property 'key' does not exist on type 'unknown'.
[value]="option.key"

src/app/page-components/search-bar/search-bar.component.html:42:21
- error TS2339: Property 'icon' does not exist on type 'unknown'.
<div *ngIf="hint.icon">
 
src/app/pages/project/rules/project-rules.component.html:116:34
- error TS2339: Property 'controls' does not exist on type 'unknown'.
[value]="condition.controls.values.value" data-cy="rule-value"
```

(2) Another class of errors is related to [this issue](https://github.com/angular/components/issues/17495). In there, apparently bare attributes were causing problems. That issue claims to be resolved, but I think I am still seeing the same thing, as shown here:
```
src/app/pages/project/list/project-list.component.html:56:27 - error TS2322: Type 'string' is not assignable to type 'boolean'.
56           <app-authorized not [allOf]="['/apis/iam/v2/projects', 'post']">
                             ~~~
```

NA

### :+1: Definition of Done
All errors of the form `Type 'AbstractControl' is not assignable to type 'FormGroup'.` are fixed.

If you edit the tsconfig.json file and set `strictInputTypes` to true, then run `ng build` you will still get quite a few errors spewed out, but none of them will be for the AbstractControl issue.

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/automate-ui/src/app/pages/integrations/aws-form/aws-form.component.html
```diff
@@ -13,7 +13,7 @@
     <input type="checkbox" formControlName="no_creds">
   </label>
 
-  <ng-container *ngIf="showInstanceCreds()" [formGroup]="formGroup.controls.credentials">
+  <ng-container *ngIf="showInstanceCreds()" [formGroup]="credentialsFormGroup()">
```
msorens, line 16: The common issue was that the RHS is an AbstractControl and that is not compatible with the LHS, which needs a FormGroup. As far as I could tell, there is no way to do a cast within the confines of a template expression (did not like either `<FormGroup>` prefix or `as FormGroup` suffix), so had to do it in the code-behind.

components/automate-ui/src/app/pages/integrations/aws-form/aws-form.component.ts
```diff
@@ -1,19 +1,11 @@
...
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class IntegrationsAWSFormComponent {
-  @Input() formGroup: FormGroup;
-
-  showInstanceCreds() {
-    const formData = this.formGroup.value;
-    const no_creds = get('no_creds', formData);
-    return !no_creds;
-  }
```
msorens, line 18: I put all this replicated code in a common base class.

components/automate-ui/src/app/pages/integrations/gcp-form/gcp-form.component.html
```diff
@@ -1,6 +1,7 @@
 <ng-container [formGroup]="formGroup">
 
-  <ng-container [formGroup]="formGroup.controls.credentials">
+  <!-- Unlike its siblings, gcp-form is rendered unconditionally -->
```
msorens, line 3: @scottopherson Can you confirm whether this comment is correct? I added the comment based on the fact that the aws form and the azure form, essentially identical siblings to this, both include an `ngIf` but this one does not. So either it needs this comment, or it needs to have an `ngIf` added.

components/automate-ui/src/app/pages/job-add/job-add.component.html
```diff
@@ -47,15 +47,15 @@
         <ng-container [ngSwitch]="step$ | async">
           <chef-job-nodes-form
             *ngSwitchCase="Step['add-nodes']"
-            [form]="form.controls['nodesGroup']">
+            [form]="formControl('nodesGroup')">
```
SEAjamieD, line 50: I don't yet what the better way is - but when I test this out using a simple console log, we get a never-ending running function.  Similar to why we try to avoid using a function in an *ngFor.

SEAjamieD: I imagine this is similar in the above files that also use a function in the template

components/automate-ui/src/app/pages/integrations/aws-form/aws-form.component.html
SEAjamieD: I see your note here - and it relates to my comment below.  I'm curious if there may be a way to limit the function call loops

components/automate-ui/src/app/pages/job-add/job-add.component.html
msorens: Excellent--glad you noticed the function-in-the-template faux pas! I was having a problem with loading scan jobs so had not been able to test this yet. In any case, I have pushed up a fix to replace the function calls with a _pure pipe_ (the other approach to resolving that as compared to my recent PR #3810 ).

components/automate-ui/src/app/pipes/cast-formgroup.pipe.ts
```diff
@@ -0,0 +1,16 @@
+import { Pipe, PipeTransform } from '@angular/core';
+import { FormGroup } from '@angular/forms';
+
+@Pipe({
+  name: 'castFormGroup',
+  pure: true
+})
+export class CastFormGroupPipe implements PipeTransform {
```
SEAjamieD, line 8: smart work!

msorens:
This is a test

components/automate-ui/src/app/pages/integrations/aws-form/aws-form.component.ts
scottopherson: I kind of prefer simple duplication like this over adding extension classes but that's just my preference.

components/automate-ui/src/app/pages/integrations/gcp-form/gcp-form.component.html
scottopherson: This is correct afaik. Configuring credentials via UI is optional for all but GCP; it never needed `showInstanceCreds` method whenever that was added.

components/automate-ui/src/app/pipes/cast-formgroup.pipe.ts
```diff
@@ -0,0 +1,16 @@
...
+
+  constructor() {
+  }
```
scottopherson, line 11: I'd drop ctor if not needed.

components/automate-ui/src/app/pages/integrations/aws-form/aws-form.component.ts
msorens: OK, my preference is rather weak, so I will revert that.

PR #3880, opened 2020-06-05
Wallaby Configuration Update

msorens:

[Wallaby.js](https://wallabyjs.com/), the continuous unit-testing tool, has recently improved their interface so that re-syncing a configuration file with every change they commit is no longer necessary. Rather, Wallaby.js now provides _zero configuration_ (they call it [automatic configuration](https://wallabyjs.com/docs/intro/config.html#vs-code)), so that as they change and improve the system any required configuration changes are automatically picked up when the auto-updater periodically installs new versions of the Wallaby core plugin in VSCode.

OK, so that is not quite true. There is a one-time switch you need to throw in VSCode once you download the code in this PR.
Open the VSCode command palette (`View >> Command Palette` --or-- <kbd>Cmd</kbd> + <kbd>Shift</kbd> + <kbd>P</kbd>) and execute `Wallaby.js: Select Configuration`, choosing the "custom directory" option:
image
That opens a standard file dialog. Drill down to the components/automate-ui directory and choose <kbd>Select Folder</kbd>.

VSCode will remember and use that setting from then on.

You will see all the choices in the illustration above on master (before this PR merges); with this PR code you will only see the first two choices, because this PR deletes those bottom two files that are no longer needed.

If you switch back to master after exercising this PR and you see any wallaby errors, just go back to the `Wallaby.js: Select Configuration` switch and reselect the wallaby.js file that you had been using prior to this PR. Just remember that the new setting will be needed once this PR merges.

NA

### :+1: Definition of Done
Wallaby.js runs all unit tests in zero-configuration mode.

Described above.

- [ ] Tests added/updated?
- [ ] Docs added/updated?


lancewf:
I tried doing the above steps and I get the below error. 
```
[Error] Automatic Angular CLI configuration error: Cannot convert undefined or null to object 
[Error]  
[Error] We've also tried to automatically configure Wallaby.js for other frameworks 
[Error] Automatic Jest configuration error: Module jest-cli is not found in '.'. 
```
I have not upgraded Wallaby in the last few months. Do I need to get a new license and upgrade for this to work?

msorens:
@lancewf I believe that is very likely the culprit.

PR #3424, opened 2020-04-19
Treat UI unit test warnings as errors

msorens:
### :nut_and_bolt: Description: What code changed, and why?

With unit test warnings not causing build failures, it is easy to either postpone fixing them indefinitely or possibly not even notice them in the first place. This somewhat complex machination adjusts the unit test run so that not only _errors_ but _warnings_ will now also generate a non-zero exit code.

Chances are there is a flag or switch somewhere to "treat warnings as errors" but I could not find it with angular or karma or jasmine... so I went with what I knew would work, because I have used it before.

### :chains: Related Resources
#3397 Fix unit test "failures"

### :+1: Definition of Done
Unit test warnings return a non-zero exit code, forcing a build failure.

##### Case 1: No errors or warnings

Pay particular attention to the very last line of output:
```
0 ✓ (1m17s) 18:54:33 automate-ui (master) $ npm run test

> automate-ui@0.0.0 test2 /Users/msorens/code/go/src/github.com/chef/automate/components/automate-ui
> /bin/bash -c 'exec 5>&1; t1=$(ng test --watch=false --code-coverage --source-map=false --configuration=no_auth | tee /dev/fd/5; exit ${PIPESTATUS[0]}); if [ $? -ne 0 ]; then exit 1; else echo $t1 | grep 'WARN:' >/dev/null; test $? -ne 0; fi'

18 04 2020 18:58:38.425:INFO [karma-server]: Karma v4.4.1 server started at http://0.0.0.0:9876/
18 04 2020 18:58:38.427:INFO [launcher]: Launching browsers ChromeHeadless with concurrency unlimited
18 04 2020 18:58:38.607:INFO [launcher]: Starting browser ChromeHeadless
18 04 2020 18:59:03.838:INFO [HeadlessChrome 80.0.3987 (Mac OS X 10.14.6)]: Connected on socket ErbB527bDCQ59zrtAAAA with id 65774247
..............
HeadlessChrome 80.0.3987 (Mac OS X 10.14.6): Executed 14 of 1488 (skipped 1474) SUCCESS (2.774 secs / 0.482 secs)
TOTAL: 14 SUCCESS
TOTAL: 14 SUCCESS
18 04 2020 18:59:21.985:WARN [launcher]: ChromeHeadless was not killed in 2000 ms, sending SIGKILL.
0 ✓ (4m49s) 18:59:22 automate-ui (master) $ 
```
[...]

PR #3885, opened 2020-06-08
Treat UI lint warnings as errors

msorens:

With lint warnings not causing build failures, it is easy to either postpone fixing them indefinitely or possibly not even notice them in the first place. In a recent PR I massaged unit test warnings to be errors; here, I am revisiting that short script to make it parameterizable so it could perform the same conversion for linting.

One particularly nice side benefit is that the command being run is now surfaced back in package.json instead of being buried in a sub-script.

PR #3424 (Treat UI unit test warnings as errors)

### :+1: Definition of Done
Lint warnings return a non-zero exit code, forcing a build failure.

Run `make lint` or `npm run lint` in the following scenarios:

1. No errors or warnings -- this should certainly be the state of the existing code base "as is"!

2. Error present but no warnings
Here's one, showing how it renders in VSCode:
image

3. Warning present but no errors
Here's one--VSCode just says it is deprecated:
image

4. Both error and warning present

For cases (2) through (4), I have provided those embedded in this PR for your convenience. Simply run `git revert -n f60c763` to get one error and one warning in the code base. You can then run case (4). Then tweak either file manually to get cases (2) and (3).

(You may, of course, also confirm that unit test errors, warnings, and neither still function correctly since I am touching that, too, with the revised script. Examples for those are provided in PR #3424.)

- [ ] Tests added/updated?
- [ ] Docs added/updated?


components/automate-ui/package.json
```diff
@@ -15,10 +15,10 @@
     "copy-ui-lib:fonts": "cp -r ../chef-ui-library/dist/collection/assets/fonts/ src/assets/fonts",
     "build": "ng build",
     "build:prod": "npm run build -- --prod",
-    "test": "scripts/unit-tests.sh",
+    "test": "scripts/build.sh 'ng test --watch=false --code-coverage --source-map=false --configuration=no_auth' 'WARN:'",
```
msorens, line 18: Nice side effect is that the command is now surfaced back in package.json instead of buried in the script.

components/automate-ui/scripts/build.sh
```diff
@@ -1,14 +1,16 @@
...
 # Setup an unused stream to redirect to stdout in the next command.
 exec 5>&1
-# We want to capture the `ng test` output to reuse it (rather than have to re-run the command),
+# We want to capture the target command's output to reuse it (rather than have to re-run the command),
 # so sending to an unused stream and redirecting that to stdout (stream 1) where it can be captured into the `out` variable.
-# In order to check the exit code of the `ng test` itself, need to use PIPESTATUS
+# In order to check the exit code of the target command itself, need to use PIPESTATUS
 # to grab the exit code of the first pipe component rather than the final one.
-out=$(ng test --watch=false --code-coverage --source-map=false --configuration=no_auth | tee /dev/fd/5; exit "${PIPESTATUS[0]}")
+out=$(eval "$CMD_STRING | tee /dev/fd/5; exit \${PIPESTATUS[0]}")
```
srenatus, line 13: 💭 Is the `eval` really necessary? What's up with `$CMD_STRING | tee /dev/fd/5; exit ${PIPESTATUS[0]}`?

components/automate-ui/scripts/build.sh
```diff
@@ -17,7 +19,7 @@ if [ $? -ne 0 ]; then
 fi
 
 # If no errors, next check for any warnings.
-echo "$out" | grep 'WARN:' >/dev/null
+echo "$out" | grep "$WARNING_LABEL" >/dev/null
```
srenatus, line 22: [nit]
```suggestion
grep -q "$WARNING_LABEL" <<<"$out"
```
But it really doesn't matter 😃

components/automate-ui/scripts/build.sh
msorens: > Is the eval really necessary?

Yes; without it we are just giving the shell a string, not a command.

> What's up with $CMD_STRING | tee /dev/fd/5; exit ${PIPESTATUS[0]}?

That was designed for flexibility, allowing the $CMD_STRING itself to contain pipes. That had actually been needed in the code base a year or more ago when I originally wrote the code. The _current_ consumers are not passing values containing pipes, but I chose to leave the extra support mechanism in place. If you feel strongly about removing because it is not being used at the moment, I could do that, just let me know.

components/automate-ui/scripts/build.sh
msorens: I like that; thanks!

components/automate-ui/scripts/build.sh
srenatus: Thanks! I generally like to avoid eval. I guess we could put the label first, use `shift`, and do something like `out=$("$@")`.... but this works so 🤷

Issue #3688, opened 2020-05-15
Desktop View: Change Tracking Dashboard - Rollout Progress APIs


danielsdeleo:
## User Story

For the desktop view, we want to show users which nodes have run which versions of the Chef Infra code so they understand which systems have applied the policy changes that have been rolled out.

Terms:
* target node segment: a set of nodes configured with the same policy name and policy group connected to the same chef server.
* Rollout: the object on the Chef Automate representing the process of the nodes in a target node segment applying a new revision of a Chef Infra policy

The UI for this is still being developed so the shape of the API may change, but we need

* Top level list: on a target node segment basis, we want to show the counts of systems that, on their most recent Chef Run, have run the policy revision for a given rollout. Also should give the counts for previous rollouts. Should be able to filter by the 3 target node segment component values.
* "Show" view: for a given single target node segment, give the "progress" counts from the list view. Also give a breakdown of CCR success vs. failure for the CCRs broken down by rollout.

## Definition of Done

Most recent mockup is at https://chef.invisionapp.com/share/T8X4GLG2XBP#/screens/418388095_v2-Desktop_Change_Rollout-Default

The UI will not be able to be finalized for a few weeks, so we are skipping grouping behaviors for now. Also, the API will just return the raw Chef Server API--we probably need to have the backend store the type of deployment/architecture we are working with to parse the chef server API into host+org like shown.

PR #3887, opened 2020-06-08
Keep CCR status in node-state regardless of missing-ness

danielsdeleo:

This is in service of the rollouts/change tracking dashboard, where we
want to give the user success/failure statistics grouped by policyfile
versions; in this scenario we need to keep the "old" CCR-based status
around even after a node becomes missing.

Part of #3688 

* Build the ingest service
* start all services (probably only need a subset but I started them all)
* use the `send_chef_run_example` helper to send a CCR report (repeat a few times if you like)
* use the `send_chef_run_failure_example` to send a CCR failure report
* Query Elasticsearch with: `curl -Ss -k -X GET http://localhost:10144/node-state/_search -H 'Content-Type: application/json' -d '{"size": 10}' | jq . | less` 
* search with `/status`, you should see stuff like:
```
          "status": "failure",
          "chef_run_status": "failure",
```

```
          "status": "missing",
          "chef_run_status": "success",
```

What you want to see is:
* every node state document has both a `status` and `chef_run_status` field.
* if the `status` is "failure" or "success" then `chef_run_status` should have the same value
* when `status` is "missing", `chef_run_status` should still be set to success/failure (unchanged)
* if you don't have any missing nodes, you can run `send_chef_run_example` a few more times and then wait a minute or so for the mark-nodes-missing job to run.

### DIFF
--- a/components/ingest-service/backend/chef_client_run.go
+++ b/components/ingest-service/backend/chef_client_run.go
@@ -255,6 +255,7 @@ func (ccr *ChefClientRun) initializeNodeInfo() (sharedNodeInfo NodeInfo, err err
 		PlatformVersion:       EmptyStringIfNil(ccr.NodePayload.Automatic["platform_version"]),
 		Source:                ccr.Source,
 		Status:                ccr.Status,
+		ChefRunStatus:         ccr.Status,
 		TotalResourceCount:    ccr.TotalResourceCount,
 		SourceFqdn:            ccr.ChefServerFqdn,
 		ExpandedRunList:       ccr.ExpandedRunList,
--- a/components/ingest-service/backend/elastic/mappings/node_state.go
+++ b/components/ingest-service/backend/elastic/mappings/node_state.go
@@ -81,6 +81,9 @@ var nodeProps = `
 	"status": {
 		"type": "keyword"
 	},
+	"chef_run_status": {
+		"type": "keyword"
+	},
 	"deprecations": {
 		"type": "object",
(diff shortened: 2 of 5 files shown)


components/ingest-service/backend/elastic/mappings/node_state.go
```diff
@@ -81,6 +81,9 @@ var nodeProps = `
 	"status": {
 		"type": "keyword"
 	},
+	"chef_run_status": {
+		"type": "keyword"
+	},
```
danielsdeleo, line 86: In case anyone wonders, I'm expecting this to be an additive change to the node state, no migration. There could be nodes that were ingested prior to upgrading to a version of code with the `chef_run_status` that are/become missing, but these won't be relevant in the change tracking APIs anyway.

components/ingest-service/backend/chef_client_run.go
```diff
@@ -31,6 +31,7 @@ type ChefClientRun struct {
 	EndTime              time.Time            `json:"end_time"`
 	Source               string               `json:"source"`
 	Status               string               `json:"status"`
+	ChefRunStatus        string               `json:"chef_run_status"`
```
vjeffrey, line 34: can you add a comment here explaining why this exists, for our future selves? something like the title of your pr + a couple words why?

danielsdeleo: Turns out we didn't need this member of this struct anyway. But I added a comment to the other one.

Issue #3724, opened 2020-05-18
ingesting nodes that have tags with a " in them breaks the node list


vjeffrey:
## Describe the bug
Tags with `"` in them break the list nodes functionality

## To Reproduce
- create a node
- add a tag with a `"` in it
- try to load the nodes, see error

customer reported and internally reported

## Expected behavior
can read tags from nodes

vjeffrey:
if anyone with postgres chops knows how to fix this at the query level, that would be awesome https://github.com/chef/automate/blob/master/components/nodemanager-service/pgdb/nodes.go#L60

danielsdeleo:
This is a bit beyond my personal experience with postgres but I think for a quick fix you could replace `"` with `\"` via the replace function (https://www.postgresql.org/docs/9.1/functions-string.html). There may be a more elegant solution using the `row_to_json` function and removing the json string construction. https://www.postgresql.org/docs/9.3/functions-json.html

srenatus:
Let's never string concat json please 😉 Also, I think `json_build_object` is quite nice to work with, for example
```sql
  SELECT json_build_object(
    'id', r.id,
    'project_id', _project_id,
    'name', r.name,
    'type', r.type,
    'status', 'staged',
    'conditions', json_agg(
      json_build_object(
        'value', rc.value,
        'operator', rc.operator,
        'attribute', rc.attribute
      )
    )
  ) AS rule
  FROM iam_staged_project_rules AS r
  INNER JOIN iam_staged_rule_conditions AS rc
  ON rc.rule_db_id=r.db_id
  WHERE r.project_id=project_db_id(_project_id) AND
        projects_match_for_rule(_project_id, _project_filter) AND
        r.deleted=false
  GROUP BY r.id, r.name, r.type
```
Note that there's some stuff missing in the snippet, all its glory lives in `components/authz-service/storage/postgres/migration/sql/`.

Related to this, we make use of unmarshaling this into a golang struct when reading structured data from the database, like this

```go
// Rule defines an ingest rule for a project.
type Rule struct {
        ID         string      `json:"id"`
        ProjectID  string      `json:"project_id"`
        Name       string      `json:"name"`
        Type       RuleType    `json:"type"`
        Conditions []Condition `json:"conditions"`
        Deleted    bool        `json:"deleted"`
        Status     string      `json:"status"`
}

// Scan implements pq Scan interface for a Rule reference
// so we can pull them out of the database directly as the correct type.
[...]

PR #3890, opened 2020-06-09
use json_build_object instead of string_agg for pg nodes

vjeffrey:
the string agg was resulting in mixups where a double quote
inside a tag would cause the entire nodes search api call to fail
bc it couldn't parse the tag.

https://github.com/chef/automate/issues/3724

### :+1: Definition of Done
can add a node with a tag that has a double quote in it and still read the nodes list

rebuild nodemanager
navigate to scan jobs/nodes - add node
add a node with a bad tag like `Dot.Comma,Big;\"Trouble`
make sure scan jobs nodes list is still readable

- [x] Tests added/updated?
- [n/a] Docs added/updated?

### DIFF
--- a/components/compliance-service/api/tests/25_nodes_spec.rb
+++ b/components/compliance-service/api/tests/25_nodes_spec.rb
@@ -130,7 +130,7 @@ describe File.basename(__FILE__) do
       tags: [
         Common::Kv.new(key: "department", value: "marketing"),
         Common::Kv.new(key: "boss", value: "John"),
-        Common::Kv.new(key: "test", value: "Dot.Comma,Big;\"Trouble")
+        Common::Kv.new(key: "test&6^\"BAD", value: "Dot.Comma,Big;\"Trouble")
       ]
     )
     node_id2 = node2['id']
@@ -191,7 +191,7 @@ describe File.basename(__FILE__) do
               "value": "John"
             },
             {
-              "key": "test",
+              "key": "test&6^\"BAD",
               "value": "Dot.Comma,Big;\"Trouble"
             },
           ],
@@ -206,7 +206,6 @@ describe File.basename(__FILE__) do
           "id": node_id1,
           "name": "M$",
           "manager": "automate",
-          "tags": [{}],
           "status": "unknown",
           "managerIds": [
             "e69dc612-7e67-43f2-9b19-256afd385820"
@@ -253,7 +252,6 @@ describe File.basename(__FILE__) do
       "id": node_id1,
       "name": "M$",
       "manager": "automate",
-      "tags": [{}],
       "status": "unknown",
       "managerIds": [
         "e69dc612-7e67-43f2-9b19-256afd385820"
@@ -289,7 +287,7 @@ describe File.basename(__FILE__) do
           "value": "John"
         },
         {
-          "key": "test",
+          "key": "test&6^\"BAD",
           "value": "Dot.Comma,Big;\"Trouble"
         }
       ],
@@ -583,7 +581,7 @@ describe File.basename(__FILE__) do
       ),
       field: "tags"
     )
-    assert_same_elements(["department", "boss", "test"], tag_keys["fields"])
+    assert_same_elements(["department", "boss", "test&6^\"BAD"], tag_keys["fields"])
 
     names = MANAGER_GRPC manager, :search_node_fields, Manager::FieldQuery.new(
       node_manager_id: "e69dc612-7e67-43f2-9b19-256afd385820",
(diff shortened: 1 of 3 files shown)


components/nodemanager-service/pgdb/nodes.go
```diff
@@ -57,7 +57,7 @@ SELECT
   COALESCE(n.source_id, '') AS source_id,
   COALESCE(n.source_region, '') AS source_region,
   COALESCE(n.source_account_id, '') AS source_account_id,
-  COALESCE(('[' || string_agg(DISTINCT '{"key":"' || t.key || '"' || ',"value": "' || t.value || '"}', ',') || ']'), '[]') :: JSON AS tags,
+  COALESCE(json_agg(DISTINCT jsonb_build_object('key', t.key, 'value', t.value)), '[]') AS tags,
```
vjeffrey, line 60: this is the actual change

components/nodemanager-service/pgdb/nodes.go
```diff
@@ -928,7 +928,9 @@ func (db *DB) QueryManualNodesFields(ctx context.Context, filters []*common.Filt
 			results = append(results, node.Name)
 		case "tags":
 			for _, kv := range node.Tags {
-				results = append(results, kv.Key)
+				if len(kv.Key) > 0 {
```
vjeffrey, line 931: i was getting an extra empty string in the test after the pg call change, so i added this to take care of it. open to suggestions on ways to not end up with the empty string in the first place

components/compliance-service/api/tests/25_nodes_spec.rb
```diff
@@ -129,7 +129,8 @@ def cleanup
       ),
       tags: [
         Common::Kv.new(key: "department", value: "marketing"),
-        Common::Kv.new(key: "boss", value: "John")
+        Common::Kv.new(key: "boss", value: "John"),
+        Common::Kv.new(key: "test", value: "Dot.Comma,Big;\"Trouble")
```
vjeffrey, line 133: this is the troublesome tag added for testing

components/compliance-service/api/tests/25_nodes_spec.rb
```diff
@@ -201,6 +206,7 @@ def cleanup
           "id": node_id1,
           "name": "M$",
           "manager": "automate",
+          "tags": [{}],
```
vjeffrey, line 209: the switch to json_build_object with json_agg leaves us with an empty array with an empty object inside. this isn't really problematic, but im open to suggestions on how to make it not happen..

components/nodemanager-service/pgdb/nodes.go
lancewf: When you create a new node without any tags a tag with empty key and value is created. The tag can not be deleted. 
image

lancewf: I found one solution. Add the below to line 256.
```
	if len(tags) == 1 && tags[0].Key == "" && tags[0].Value == "" {
		newNode.Tags = []*common.Kv{}
	} else {
		newNode.Tags = tags
	}
```
But I am still looking to see why the empty object is being created.

lancewf: For a node with no tags, we are getting back `[{\"key\": null, \"value\": null}]` from the database.

lancewf: I think I figured it out!
https://stackoverflow.com/questions/24155190/postgresql-left-join-json-agg-ignore-remove-null. 
We use `  COALESCE(json_agg(DISTINCT jsonb_build_object('key', t.key, 'value', t.value)) FILTER (WHERE t.id IS NOT NULL), '[]') AS tags,` to remove the null in the array.

components/nodemanager-service/pgdb/nodes.go
lancewf, line 60:
```suggestion
    COALESCE(json_agg(DISTINCT jsonb_build_object('key', t.key, 'value', t.value)) FILTER (WHERE t.id IS NOT NULL), '[]') AS tags,
```

components/compliance-service/api/tests/25_nodes_spec.rb
lancewf: This should be fixed with the below SQL change.

components/nodemanager-service/pgdb/nodes.go
vjeffrey: nice! i’ll give it a try when i’m done with lunch! thanks!

vjeffrey: this worked nicely; thanks! i also updated the key to have a wonky string as discussed in standup

Issue #3745, opened 2020-05-19
[NON-RELIANT]: Question Mark Hover Actions


SEAjamieD:
There is a hover action that will need to be created for the Question Marks on each heading.

![Screen Shot 2020-05-19 at 1.33.22 PM.png](https://images.zenhubusercontent.com/5d7aa5d5d4249c0001cab607/0bf94f64-6640-47ee-a2a4-6a57fba6efe6)

![Screen Shot 2020-05-19 at 1.21.30 PM.png](https://images.zenhubusercontent.com/5d7aa5d5d4249c0001cab607/6ab97558-ef04-4feb-9c70-12407095a08c)

Default Mockup: https://chef.invisionapp.com/d/main#/console/19726898/412590750/preview
Find specific tooltips in Tooltips and dropdown section: https://chef.invisionapp.com/d/main#/projects/prototypes/19726898

apriofrost:
Attached the mockup screenshot here:
![image](https://user-images.githubusercontent.com/9401459/83922084-9b11e280-a734-11ea-8c9b-af4706102c8a.png)

The spec can be inspected here in Sketch Cloud here:  https://www.sketch.com/s/133f1f0f-a3e8-4d09-bcc6-f00385057bad/a/kPJQdl

PR #3894, opened 2020-06-09
Jamie/3745 desktop question mark tooltips

SEAjamieD:
Adding tooltips for information about what each section is for on Desktop

I also added the scaffolding necessary for each of the 4 components involved.  I was unable to find a good way to unit test hover actions, so I ended up NOT adding tests for these tooltips.  I think this is okay as we have tests already for chef-tooltip, and the text is not dynamic anyways, so theres no real special/unique test to be done for these.

Adds: https://github.com/chef/automate/issues/3745

### :+1: Definition of Done
Tooltips are visible when hovering over the information icon, matching the messaging from the spec.  Current tooltip styling remains.

in hab: `build components/automate-ui-devproxy`
in automate-ui: `make serve`

navigate to https://a2-dev.test/desktop
- look for the 4 info icons, one in each section (Daily Check-in, Check-in History, Time since last check-in, Top errors).  They look like an 'i' with a circle around it
- hover over each one of the info icons and view the tooltip

- [ ] Tests added/updated?
- [x] Docs added/updated?

After:
Screen Shot 2020-06-09 at 11 27 27 AM

### DIFF
--- a/components/automate-ui/src/app/modules/desktop/check-in-time-series/check-in-time-series.component.html
+++ b/components/automate-ui/src/app/modules/desktop/check-in-time-series/check-in-time-series.component.html
@@ -1,7 +1,13 @@
 <div class="content-container">
   <div class="header">
     <div class="heading">
-      <h2 class="heading-title">Check-in History</h2>
+      <div class="heading-title-container">
+        <h2 class="heading-title">Check-in History</h2>
+        <chef-icon id="checkin-history-info" style='font-size: 24px'>info_outline</chef-icon>
+        <chef-tooltip for='checkin-history-info'>
+          <p class="tooltip-text">Percentage of checked-in to total desktops at 24 hour intervals.</p>
+        </chef-tooltip>
+      </div>
       <h6 class="heading-time">Updated {{ lastUpdatedMessage }}</h6>
     </div>
     <select (change)="dateChanged($event.target.value)">
--- a/components/automate-ui/src/app/modules/desktop/check-in-time-series/check-in-time-series.component.html
+++ b/components/automate-ui/src/app/modules/desktop/check-in-time-series/check-in-time-series.component.html
@@ -3,7 +3,7 @@
     <div class="heading">
       <div class="heading-title-container">
         <h2 class="heading-title">Check-in History</h2>
-        <chef-icon id="checkin-history-info" style='font-size: 24px'>info_outline</chef-icon>
+        <chef-icon id="checkin-history-info">info_outline</chef-icon>
         <chef-tooltip for='checkin-history-info'>
           <p class="tooltip-text">Percentage of checked-in to total desktops at 24 hour intervals.</p>
         </chef-tooltip>
--- a/components/automate-ui/src/app/modules/desktop/check-in-time-series/check-in-time-series.component.scss
+++ b/components/automate-ui/src/app/modules/desktop/check-in-time-series/check-in-time-series.component.scss
@@ -1,22 +1,5 @@
 @import "~styles/variables";
 
-.heading-title {
-  color: var(--chef-primary-dark);
-  margin-top: 0;
-  font-weight: inherit;
-  font-style: inherit;
-  font-size: 18px;
-}
-
-.heading-time {
-  color: var(--chef-primary-dark);
-  margin-top: 0;
-  margin-bottom: 12px;
-  font-weight: inherit;
-  font-style: inherit;
-  font-size: 10px;
-}
-
 .header {
   display: flex;
   align-items: flex-start;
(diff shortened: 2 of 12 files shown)


components/automate-ui/src/app/modules/desktop/check-in-time-series/check-in-time-series.component.html
```diff
@@ -1,7 +1,13 @@
 <div class="content-container">
   <div class="header">
     <div class="heading">
-      <h2 class="heading-title">Check-in History</h2>
+      <div class="heading-title-container">
```
SEAjamieD, line 4: I added a container around the title because I prefer to use flex for alignment purposes.  You will see this in all the html files

components/automate-ui/src/app/modules/desktop/daily-check-in/daily-check-in.component.spec.ts
```diff
@@ -0,0 +1,26 @@
+import { CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
```
SEAjamieD, line 1: I added spec files and the scaffolding to start testing for all the new desktop components I touched, but ended up not adding tests for the tooltips as I couldn't find a good way to trigger hover in unit tests.  When I thought about it more, this seems ok, because the tooltip text is not dynamic, and chef-tooltip already has tests for displaying the tip text.

components/automate-ui/src/app/modules/desktop/check-in-time-series/check-in-time-series.component.scss
```diff
@@ -27,6 +10,47 @@
...
+    }
+
+    chef-icon {
+      margin-left: 4px;
+      font-size: 24px;
+      cursor: help;
+    }
+
+    chef-tooltip {
+      width: 288px;
```
scottopherson, line 36: How was 288px chosen?

components/automate-ui/src/app/modules/desktop/daily-check-in/daily-check-in.component.spec.ts
scottopherson: Does the 'hover' action need to be unit tested here? Should be sufficient to just check for the presence of the 'info' icon and that it has a tooltip presently associated with it.

"Does `#icon` element exist?"
"Does tooltip `[for="icon"]` element exist?"

components/automate-ui/src/app/modules/desktop/check-in-time-series/check-in-time-series.component.scss
SEAjamieD: it was the width from invision files, which made the text layout match those from the mock up.

SEAjamieD:
>Does the 'hover' action need to be unit tested here? Should be sufficient to just check for the presence of the 'info' icon and that it has a tooltip presently associated with it.
"Does #icon element exist?"
"Does tooltip [for="icon"] element exist?"

@scottopherson - Checking for existence of the icon element is a good idea - it seems counter to the idea of testing behavior though.  Thoughts?  Existence tests seem better than no tests though.

components/automate-ui/src/app/modules/desktop/daily-check-in/daily-check-in.component.spec.ts
SEAjamieD: Checking for existence of the icon element is a good idea - it seems counter to the idea of testing behavior though. Thoughts? Existence tests seem better than no tests though.

scottopherson: Tradeoffs. You can test the hover behavior here but are you going to test every tooltip hover interaction everywhere? It'd be thorough and could possibly catch some unforeseen things but would be really slow.

We should test hover behavior in `chef-tooltip`'s test suite but everywhere else it should be sufficient enough to say that if I use the element as documented, it should work as documented. If you're doing something unique with the tooltip interaction or styling on a particular page, sure, add extra tests there but otherwise you should be fine.

For example, we don't usually write tests to ensure that `input` elements with a `disabled` attribute cannot not have their `value` changed. We usually just check that the `disabled` attribute is present and assume the element will behave correctly.

SEAjamieD: appreciate the input, i'm gonna add some of existence tests here before merging 👍

msorens: My notions are very much inline with @scottopherson's, a unit test should be testing things relevant to the current "unit"--generally the current component--and you assume other components perform as advertised.

The one thing I might add is that you can expand the thinking so rather than "existence tests" you could think of them as "value tests" to perhaps be a bit more palatable: _Does attribute `x` have value `y`?_  The case of a "raw" attribute--like `disabled`--is the degenerate case, where the value is just implicitly true.

SEAjamieD: Thank you! I appreciate the responses 👍

Issue #3800, opened 2020-05-27
improve error logging in inspec runner


vjeffrey:
## User Story

would be nice to log more of the errors in the inspec runner code to make it easier to debug what is going on when things aren't working!

## Definition of Done
improved error logging in: https://github.com/chef/automate/blob/master/components/compliance-service/cmd/inspec_runner/inspec_runner.go

teknofire:
Specifically, there are several errors that can be returned from `changeToUser` and the error message tends to be a generic OS error like `operation not permitted` and it's not clear if that happened during user lookup, or the UID/GID change or even somewhere else.

It would help with troubleshooting if we can add more debug messages about what it's trying to do: "Found nobody with UID/GID 12345/12345", "Changing GID to 12345", etc...

vjeffrey:
just noticed this ticket as well https://github.com/chef/automate/issues/674

Issue #674, opened 2019-06-25
compliance-service inspec_runner missing setuid check


jeremymv2:
# User Story
As an Operator, I would like to know when/if the compliance-service inspec_runner file permission is changed, rendering the service unable to upload new profiles.

We discovered the setuid [permissions](https://github.com/chef/automate/blob/master/components/compliance-service/habitat/plan.sh#L92) missing at a customer site; possibly due to an install anomaly or a security hardening script. When the `inspec_runner` has no setuid permission, it cannot upload new profiles, resulting in an error similar to this which is somewhat hard to debug:

```
rpc error: code = InvalidArgument desc = Check InSpec check failed for /hab/svc/compliance-service/var/tmp/inspec-upload292792092.tar.gz with message: exit status 1
2019/06/24 10:04:55 operation not permitted
```

If the compliance service were to routinely check the setuid permission, or were to report a more informative error message it would help Operators reduce the time to fix.

PR #3909, opened 2020-06-10
more logging in inspec_runner code to aid in troubleshooting

vjeffrey:

adding more logging to inspec runner code

https://github.com/chef/automate/issues/674
https://github.com/chef/automate/issues/3800

### :+1: Definition of Done
more logs to help with troubleshooting

### DIFF
--- a/components/compliance-service/cmd/inspec_runner/inspec_runner.go
+++ b/components/compliance-service/cmd/inspec_runner/inspec_runner.go
@@ -11,6 +11,7 @@ import (
 	"syscall"
 
 	"github.com/chef/automate/components/compliance-service/cmd/inspec_runner/platform"
+	"github.com/sirupsen/logrus"
 )
 
 // Set at build time via linker flags.
@@ -31,6 +32,7 @@ func main() {
 	args := append([]string{cmd}, os.Args[1:]...)
 
 	if err := syscall.Exec(EXECUTABLE_PATH, args, os.Environ()); err != nil {
+		logrus.Errorf("inspec_runner unable to complete with executable path: %s, args: %v, env: %s", EXECUTABLE_PATH, args, os.Environ())
 		log.Fatal(err)
 	}
 }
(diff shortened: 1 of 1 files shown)


components/compliance-service/cmd/inspec_runner/inspec_runner.go
```diff
@@ -40,26 +42,31 @@ func main() {
 func changeToUser(username string) error {
 	user, err := user.Lookup(username)
 	if err != nil {
+		logrus.Errorf("inspec_runner unable to look up user %s: %v", username, err)
```
vjeffrey, line 45: i considered only adding a log statement at the function caller above instead of one per possible error in this function, but given that we're doing this bc it's very hard to debug and that function should be throwing a fatal log on error, i decided to lean towards more verbose/specific error msgs

components/compliance-service/cmd/inspec_runner/inspec_runner.go
```diff
@@ -11,6 +11,7 @@ import (
 	"syscall"
 
 	"github.com/chef/automate/components/compliance-service/cmd/inspec_runner/platform"
+	"github.com/sirupsen/logrus"
```
stevendanna, line 14: I know it is a pain, but I think we should try to avoid most external dependencies in this tool.  We would like this stub to have as little code as possible at runtime.

components/compliance-service/cmd/inspec_runner/inspec_runner.go
stevendanna, line 45: Since this function returns an error, let's wrap the errors and just return them as they are already logged by the caller.

stevendanna: I think we can still have a specific error message here by wrapping the error before we return it.  The caller calls log.Fatal on the error so any information we put into the wrapped error should get logged.

PR #3925, opened 2020-06-11
Add id and for attributes to tooltip and alter styling

SEAjamieD:
Adjust so that tooltips render
Chef Automate (10)

### DIFF
--- a/components/automate-ui/src/app/modules/nodes/nodes-list/nodes-list.component.html
+++ b/components/automate-ui/src/app/modules/nodes/nodes-list/nodes-list.component.html
@@ -56,7 +56,7 @@
               <chef-td (click)="filterFor('manager_type', node.manager)">{{ node.manager }}</chef-td>
               <chef-td (click)="filterFor('last_contact', node.last_contact)">{{ node.last_contact }}</chef-td>
               <chef-td>
-                <chef-icon class="status-icon" [ngClass]="node.scan_data.status"
+                <chef-icon [attr.id]="node.name" class="status-icon--pointer" [ngClass]="node.scan_data.status"
                 (click)="filterFor('last_scan_status', node.scan_data.status)">
                   {{ statusIcon(node.scan_data.status) }}
                 </chef-icon>
@@ -64,9 +64,10 @@
                 (click)="filterFor('last_scan_penultimate_status', node.scan_data.penultimate_status)">
                   {{ statusIcon(node.scan_data.penultimate_status) }}
                 </chef-icon> 
-                <chef-tooltip
-                [ngClass]="['run-info-tooltip', node.scan_data.status]"
-                interactable>
+                <chef-tooltip 
+                  [attr.for]="node.name"
+                  [ngClass]="['run-info-tooltip', node.scan_data.status]"
+                  interactable>
                   <p class="run-info-tooltip-info">
                     <span class="run-info-tooltip-text">{{ node.scan_data.status | titlecase }}</span>
                     <span class="run-info-tooltip-date">{{ node.scan_data.end_time | datetime: DateTime.RFC2822 }}</span>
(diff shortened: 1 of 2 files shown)


components/automate-ui/src/app/modules/nodes/nodes-list/nodes-list.component.html
```diff
@@ -56,17 +56,18 @@
               <chef-td (click)="filterFor('manager_type', node.manager)">{{ node.manager }}</chef-td>
               <chef-td (click)="filterFor('last_contact', node.last_contact)">{{ node.last_contact }}</chef-td>
               <chef-td>
-                <chef-icon class="status-icon" [ngClass]="node.scan_data.status"
+                <chef-icon [attr.id]="node.name" class="status-icon--pointer" [ngClass]="node.scan_data.status"
```
SEAjamieD, line 59: in order for chef-tooltips to show, the "target" or thing you hover over needs to have an `id` that matches the `for` attribute on the chef-tooltip.

SEAjamieD: i also added a small pointer utility class that you can choose to do away with if you want - just giving the cursor a pointer when hovering the icon

components/automate-ui/src/app/modules/nodes/nodes-list/nodes-list.component.scss
```diff
@@ -25,13 +25,17 @@ chef-icon.SKIPPED {
...
+  }
+}
+
 .run-info-tooltip {
   border-left: $spacing-03 solid;
   border-right: $spacing-03 solid transparent;
   padding-right: $spacing-03;
-  bottom: 100%;
-  left: 50%;
-  margin-left: -60px;
```
SEAjamieD, line 34: tooltip placement is generated by the `chef-tooltip` component, so adding our own placement on top of that will throw it off.

components/automate-ui/src/app/modules/nodes/nodes-list/nodes-list.component.scss
```diff
@@ -44,10 +48,6 @@ chef-icon.SKIPPED {
   &.UKNOWN, &.SKIPPED {
     border-left-color: $status-unknown;
   }
-  &.hydrated { 
-    visibility: visible;
-    opacity: 1;
-  }
```
SEAjamieD, line 50: no need to manually hydrate.  I haven't seen an instance so far where this was needed manually but I don't know much about it.

SEAjamieD:
I've only added to the first Icon of each set in Last Scans

components/automate-ui/src/app/modules/nodes/nodes-list/nodes-list.component.scss
vjeffrey: ya this was me trying hard to force it to show up

components/automate-ui/src/app/modules/nodes/nodes-list/nodes-list.component.scss
vjeffrey: that makes sense!

components/automate-ui/src/app/modules/nodes/nodes-list/nodes-list.component.html
vjeffrey: thank you!! that's the 🔑 -- thank you so much!

do we have a place where our components are documented - like in the code or anything? i didn't actually take the time to look (i should have) but wondering for my future self.

SEAjamieD: cd into components/chef-ui-library
npm install
npm start
navigate to localhost:3333

There is pretty solid documentation for each component there!

PR #3931, opened 2020-06-12
Include Services Data in Platform-wide Data Lifecycle API

danielsdeleo:

Integrate the applications-service's data lifecycle settings APIs with the new Automate-wide data lifecycle APIs. With this in place we can remove the grey-out on the configuration section for these options in the settings page.

build applications-service and gateway.

##### Using the new features in the existing (separate) API:
```
curl -X POST -fsS -k -H "api-token: ${T}" https://localhost:2000/api/v0/retention/service_groups/delete_disconnected_services/config  -d '
{                                                                                                                                                          
  "threshold": "5m",
  "running": true,
  "recurrence": "FREQ=SECONDLY;DTSTART=20200612T182142Z;INTERVAL=68"
}'
```
Then `gateway_get /api/v0/retention/service_groups/delete_disconnected_services/config | jq .`

And

```
curl -X POST -fsS -k -H "api-token: ${T}" https://localhost:2000/api/v0/retention/service_groups/disconnected_services/config  -d '
{                                                                                                                                                          
  "threshold": "5m",
  "running": true,
  "recurrence": "FREQ=SECONDLY;DTSTART=20200612T182142Z;INTERVAL=68"
}'
```
Then `gateway_get /api/v0/retention/service_groups/disconnected_services/config | jq .`

##### Errors/Validation:

```
curl -X POST  -k -H "api-token: ${T}" https://localhost:2000/api/v0/retention/service_groups/disconnected_services/config  -d '
{                                                                                                                                                          
  "threshold": "5m",
  "running": true,
  "recurrence": "FREQ=SECONDLY;DTSTART=20200612T182166Z;INTERVAL=68"
}'
```

```
curl -X POST  -k -H "api-token: ${T}" https://localhost:2000/api/v0/retention/service_groups/delete_disconnected_services/config  -d '
{                                                                                                                                                          
  "threshold": "5m",
  "running": true,
  "recurrence": "FREQ=SECONDLY;DTSTART=20200612T182166Z;INTERVAL=68"
}'
```

##### Configuring Services Data Lifecycle in the standardized API

Get the JSON payload from the docs page (components/automate-chef-io/content/docs/data-lifecycle.md ) - it's huge so I'm not gonna paste it here. Put it in a file named `dlc.json`. Then do:
```
export TOKEN=$(get_admin_token)
# upload the config:
curl -k -s -H "api-token: $TOKEN" -X PUT --data "@dlc.json" https://a2-dev.test/api/v0/data-lifecycle/config | jq .
# fetch it back 
curl -k -s -H "api-token: $TOKEN" https://a2-dev.test/api/v0/data-lifecycle/status | jq .
# fetch just the services:
curl -k -s -H "api-token: $TOKEN" https://a2-dev.test/api/v0/data-lifecycle/services/status | jq .
```

Edit the file to e.g., set recurrence interval to a new value and/or change the threshold, then show it again to confirm it changed.


api/external/applications/applications.proto
```diff
@@ -274,7 +278,6 @@ service ApplicationsService {
 
   Changes the configuration for the task that marks services as disconnected after
   'threshold'. Threshold is a string that follows Elasticsearch's date math expressions.
-  This job cannot be disabled, and therefore no information about running is accepted.
```
danielsdeleo, line 277: The example requests should show all the options. Also need some wordage to encourage users to not actually configure the recurrence or disable.

api/external/applications/applications.proto
```diff
@@ -469,27 +531,77 @@ message ServicesStatsRes {
...
+// changed to make the various data lifecycle APIs consistent with each other.
+// Thus, there is a `running` field which will disable the job if set to false.
+// That is not recommended.
 message PeriodicMandatoryJobConfig {
   // The `threshold` setting used by periodic jobs for evaluating services.
   // Threshold is a string that follows Elasticsearch's date math expressions. For more information, see the simpledatemath package under `lib/`.
   string threshold = 1;
+  // Enable/disable the job. `false` is disabled, `true` is enabled. It is not
+  // recommended to disable this job.
+  google.protobuf.BoolValue running = 2;
```
danielsdeleo, line 564: I needed to use a wrapper here to preserve API compatibility. Existing users might have code/docs/scripts to send a JSON payload like

```json
'{"threshold": "15m" }'
```
If we used a plain boolean for the new element, it would be `false` whenever the user didn't include it in the JSON payload. Thus, the `running` setting would be false and disable the job. We _could_ have flipped the setting to be "disabled" (so a `false` value would make the job enabled), but then it would be different than the other job which would be ugly. 

Using a wrapper lets us distinguish between "not passed" and "false" and has the same JSON representation so it was the best option in the end.

PR #3932, opened 2020-06-12
experimental one node view: various fixes

vjeffrey:

- set default sort on page load to be last contact, desc
- added tooltip (with help from @SEAjamieD, thank you!) to the infra and inspec run status hover
- fixed up the filtering logic in the ui code to handle daterange filters
- added "missing" as valid node state filter value

### :+1: Definition of Done
nodes page is sorted by last contact, desc on initial page load
tooltip on status that links to client run/inspec scan details
handle timerange filters in the ui code

rebuild nodemanager-service and ui
load_compliance_reports
go to a2-dev.test/nodes
check the sorting (last contact, desc)
try to filter by a time in last contact, expect it to return all nodes that had a last contact on that day
hover over a run or scan status and see the tooltip, follow the link to client run or inspec scan details

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/modules/nodes/nodes-list/nodes-list.component.scss
+++ b/components/automate-ui/src/app/modules/nodes/nodes-list/nodes-list.component.scss
@@ -4,7 +4,7 @@
 chef-td {
   font-size: 80%;
 
-  &.pointer {
+  &.help-pointer {
     cursor: help;
   }
 }
--- a/components/automate-ui/src/app/modules/nodes/nodes-list/nodes-list.component.scss
+++ b/components/automate-ui/src/app/modules/nodes/nodes-list/nodes-list.component.scss
@@ -66,8 +66,13 @@ chef-icon.SKIPPED {
     color: inherit;
   }
 }
+.run-info-tooltip-text {
+  display: block;
+  font-weight: 700;
+  padding-bottom: 3px;
+}
 
-.run-info-tooltip.missing {
+.run-info-tooltip.UNKNOWN {
   .run-info-tooltip-date,
   .run-info-tooltip-link {
     display: none;
--- a/components/automate-ui/src/app/modules/nodes/nodes-list/nodes-list.component.scss
+++ b/components/automate-ui/src/app/modules/nodes/nodes-list/nodes-list.component.scss
@@ -66,6 +66,7 @@ chef-icon.SKIPPED {
     color: inherit;
   }
 }
+
 .run-info-tooltip-text {
   display: block;
   font-weight: 700;
--- a/components/automate-ui/src/app/modules/nodes/nodes-list/nodes-list.component.scss
+++ b/components/automate-ui/src/app/modules/nodes/nodes-list/nodes-list.component.scss
@@ -3,6 +3,10 @@
 
 chef-td {
   font-size: 80%;
+
+  &.pointer {
+    cursor: help;
+  }
 }
 
 chef-tr {
@@ -23,13 +27,6 @@ chef-icon.UNKNOWN {
 
 chef-icon.SKIPPED {
   color: $chef-dark-grey;
-}
-
-.status-icon {
-
-  &--pointer {
-    cursor: pointer;
-  }
 }
 
 .run-info-tooltip {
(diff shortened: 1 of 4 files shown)


components/automate-ui/src/app/modules/nodes/nodes-list/nodes-list.component.ts
```diff
@@ -79,7 +80,19 @@ export class NodesListComponent implements OnInit, OnDestroy {
...
-    this.nodesListFilters.push({key: type, values: [item]});
+    if (type === 'last_contact') {
+      const baseTime = item.split('T');
+      // for a last contact filter, we send in two dates - beg of day and end of day
+      // so we split on T to get the date and then just add the times ourselves
+      this.nodesListFilters.push({key: type,
+        values: [
+          baseTime[0] + 'T00:00:00Z',
+          baseTime[0] + 'T23:59:59Z'
+        ]
```
msorens, line 91: Better to deal with objects semantically rather than raw string manipulations.
```suggestion
      this.nodesListFilters.push({key: type,
        values: [
          moment(item).startOf('day').toISOString(),
          moment(item).endOf('day').toISOString()
        ]
```

components/nodemanager-service/pgdb/nodes.go
```diff
@@ -665,7 +665,7 @@ func validateNodeFilters(filters []*common.Filter) error {
 		case "state":
 			for _, item := range filter.Values {
 				if !isValidState(item) {
-					return &errorutils.InvalidError{Msg: fmt.Sprintf("Invalid state filter: %s. state must be one of the following: 'RUNNING', 'STOPPED', 'TERMINATED'", item)}
+					return &errorutils.InvalidError{Msg: fmt.Sprintf("Invalid state filter: %s. state must be one of the following: 'RUNNING', 'STOPPED', 'TERMINATED', 'MISSING'", item)}
```
msorens, line 668:
```suggestion
					return &errorutils.InvalidError{Msg: fmt.Sprintf("Invalid state filter: %s. State must be one of the following: 'RUNNING', 'STOPPED', 'TERMINATED', 'MISSING'", item)}
```

SEAjamieD:
Question on tooltip usage.  Posted a video to show more clearly.

tooltip question


1.  There are currently only tooltips attached to the first status icon of each column, Was that intentional?  Sometimes there are two different status symbols in the same column.

2.  Inside the tooltip - this might be just the fake data, I noticed that when there are two opposite status icons i.e. one column holds `failed` and `passed`, the tooltip says `previous status: {some status}`.  The "some status" is looking forward, so maybe they are reversed?  I realize that this very well could be the "last status" that we are not seeing displayed, but in every case it _always_ matches the status to its right, so wanted to check.

screenshot for clarity on number 2: this is hovering over a `failed status`
Screen Shot 2020-06-16 at 8 54 42 AM

vjeffrey:
there's nothing to link to for the previous (penultimate_ status), so that's why we just say the status there in the tooltip (for now, this should be uxd further down the road) -- the order of the icons in the html is status, penultimate status - so i think what you're showing in the screenshot is correct but you were expecting the latest to be the rightmost icon?

vjeffrey:
i think it would be cool to apply the tooltip to the whole cell -- i suppose that's just a matter of adding the attr.id to the other icon?

SEAjamieD:
> i think it would be cool to apply the tooltip to the whole cell -- i suppose that's just a matter of adding the attr.id to the other icon?

@vjeffrey - yeah it should just be a matter of moving the id over - I can do that right now, should only take a sec.

The last status being shown makes sense to me then.

vjeffrey:
if you don't mind - sure, thank you!!

SEAjamieD:
@vjeffrey pushed up that change 👍

components/automate-ui/src/app/modules/nodes/nodes-list/nodes-list.component.scss
```diff
@@ -24,6 +29,50 @@ chef-icon.SKIPPED {
   color: $chef-dark-grey;
 }
 
-chef-tooltip {
-  width: 400px;
+.run-info-tooltip {
+  border-left: $spacing-03 solid;
+  border-right: $spacing-03 solid transparent;
+  padding-right: $spacing-03;
```
SEAjamieD, line 35: I think you can just delete these - looks like together these cancel eachother out

PR #3944, opened 2020-06-17
Fix error banner appearing for non-local users on login

msorens:
#### How to reproduce:
Connect to a box with SAML (LDAP boxes would exhibit the problem, too).
Login with SAML (LDAP). Note that the dev environment is configured to use SAML so you could do this on your local machine.

Assuming you have appropriate permission to fetch users, previously you would have seen this error:
image

Without permission from a policy, you would still see the same root error ("Could not get logged in user") but it would be greyed out and indicate no permissions:
image

#### What was failing
ngrx dispatch to the `SetUserSelfID` action in ChefSessionService.

#### What changed to make it fail
PR 3725, merged 14 days ago, changed the flow slightly as to where `tryInitializeSession` was being called. The timing used to be such that the `SetUserSelfID` dispatch was not being run and with PR 3725 it was.

#### The fix
Looking into what `SetUserSelfID` was doing:
(a) set the ID into the ngrx store
(b) call GetUser on the back-end
As far as I could tell, (b) was unnecessary; that call will be done as needed on other pages.

### :+1: Definition of Done
No error banner for non-local users.

If you login with permissions, you get the standard logged in view with no error banner.

If you login with no permissions you simply get the "no permissions" page but no error banner:
image

Logged in as Admin, go to Settings >> Policies >> Administrator >> Members >> Add Members >> Add Member Expression and enter `user:saml:*`.
Log out then log back in with your SAML id.

NB: This error would only occur between your logging in and arriving on your first page. It would not occur with just page refresh.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/entities/users/userself.effects.ts
+++ b/components/automate-ui/src/app/entities/users/userself.effects.ts
@@ -60,11 +60,6 @@ export class UserSelfEffects {
     }));
 
   @Effect()
-  setUserSelfId$ = this.actions$.pipe(
-    ofType(UserSelfActionTypes.SET_ID),
-    map(() => new GetUserSelf()));
-
-  @Effect()
   updatePasswordSelf$ = this.actions$.pipe(
     ofType<UpdatePasswordSelf>(UserSelfActionTypes.UPDATE_PASSWORD_SELF),
     mergeMap((action: UpdatePasswordSelf) =>
--- a/components/automate-ui/src/app/pages/signin/signin.component.spec.ts
+++ b/components/automate-ui/src/app/pages/signin/signin.component.spec.ts
@@ -156,7 +156,7 @@ describe('SigninComponent', () => {
         'kilgore@kilgore.trout',
         valid_id_token,
         ['authors'],
-        false);
+        true);
     });
   });
 
--- a/components/automate-ui/src/app/pages/signin/signin.component.ts
+++ b/components/automate-ui/src/app/pages/signin/signin.component.ts
@@ -48,8 +48,6 @@ export class SigninComponent implements OnInit {
   }
 
   setSession(): void {
-    const isLocalUser = this.id.federated_claims &&
-      this.id.federated_claims.connector_id === 'local';
     this.session.setSession(
       this.id.sub,
       this.id.name,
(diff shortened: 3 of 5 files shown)


jaym:
do you understand why setDefaultSession is being called? USE_DEFAULT_SESSION should be set to false from what I understand

msorens:
@jaym Yes, my further findings agree with yours. I think this is a red herring and am now looking at this bit from chef-session.service.ts as suspect:
```
    if (!this.user || username !== this.user.username) {
      this.store.dispatch(new SetUserSelfID({ id: username }));
    }
```

msorens:
As far as recent code changes, the call to `ingestIDToken` is of interest, because that was changed recently, and inside there is a call to `setSession`.

components/automate-ui/src/app/entities/users/userself.effects.ts
```diff
@@ -59,11 +59,6 @@ export class UserSelfEffects {
       });
     }));
 
-  @Effect()
-  setUserSelfId$ = this.actions$.pipe(
-    ofType(UserSelfActionTypes.SET_ID),
-    map(() => new GetUserSelf()));
```
msorens, line 65: This is the fix--see the commits for details.

components/automate-ui/src/app/pages/signin/signin.component.ts
```diff
@@ -48,8 +48,6 @@ export class SigninComponent implements OnInit {
   }
 
   setSession(): void {
-    const isLocalUser = this.id.federated_claims &&
-      this.id.federated_claims.connector_id === 'local';
```
msorens, line 52: Consolidation of this in the next two files is unrelated, rather a preventative measure that might have caused problems in the future.

components/automate-ui/src/app/pages/signin/signin.component.spec.ts
```diff
@@ -156,7 +156,7 @@ describe('SigninComponent', () => {
         'kilgore@kilgore.trout',
         valid_id_token,
         ['authors'],
-        false);
+        true);
```
tylercloke, line 159: Do we even need this anymore?

stevenoneill-jmfamily:
Thanks for fixing this one! Appreciate it!

Issue #667, opened 2019-06-24
"suggest" field values for `api/v0/nodes` endpoint


vjeffrey:
# User Story

This is supportive work for https://github.com/chef/automate/issues/468.

In order to give users the best experience when using the searchbar on this view, we will need to be able to "suggest" field values.  
The suggestions should be "filtered suggestions" (i.e. apply the filters when getting suggestions, so that we don't offer values that contradict eachother).
More information here: https://github.com/chef/automate/issues/468#issuecomment-504259238
We already have the ability to "suggest" values given a key (https://github.com/chef/automate/blob/master/components/nodemanager-service/pgdb/nodes.go#L890) - this work should only involve expanding this to the list of fields needed in the searchbar.

# Definition of Done
Can "suggest" field values for all fields we support filtering on, i.e. https://github.com/chef/automate/blob/master/components/automate-chef-io/content/docs/nodes.md#filtering-nodes

PR #3945, opened 2020-06-17
add GetNodeSuggestions function

vjeffrey:
this is not used by any API/other function yet, but will be.
it only supports suggestions for columns that are in the nodes table.
it respects filters, it does not apply sorting to the returned list.

https://github.com/chef/automate/issues/667

### :+1: Definition of Done
basic function for returning suggestions on nodes table column fields

rebuild nodemanager, run the db tests! (`make test-integration-db`)

- [x] Tests added/updated?
- [n/a] Docs added/updated?

### DIFF
--- a/components/nodemanager-service/pgdb/nodes.go
+++ b/components/nodemanager-service/pgdb/nodes.go
@@ -31,6 +31,13 @@ const (
 	NodeManagersTableAbbrev = "nm"
 )
 
+const sqlGetNodeFields = `
+SELECT DISTINCT %s 
+from nodes n
+%s %s
+%s;
+`
+
 const selectSecretIdsFromNodes = `
 SELECT secret_id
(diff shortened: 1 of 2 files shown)


components/nodemanager-service/pgdb/nodes.go
```diff
@@ -31,6 +31,13 @@ const (
 	NodeManagersTableAbbrev = "nm"
 )
 
+const sqlGetNodeFields = `
+SELECT DISTINCT %s 
+from nodes n
+%s %s
```
vjeffrey, line 37: i'm reusing the filtering functions we already have in place for GetNodes so followed the same pattern here. 
its basically select distinct {field} from nodes n {where blabla} {having blabla} {and/where field is not empty string}

PR #3955, opened 2020-06-22
Satellite15/datfeed service config

satellite15:

Updating the datafeed config_request to expose properties for patching.
Also adding the external fqdn of automate to be used in the node section of the datafeed.

compile_all_protobuf_components 
build components/data-feed-service
build components/automate-cli
build components/automate-gateway
build components/automate-load-balancer
build components/automate-deployment

Deploy the changes and test by patching 

chef-automate config patch patch.toml

example patch:

```
[data_feed_service]
  [data_feed_service.v1.sys.service]
    feed_interval = "3m"
    asset_page_size = 50
    reports_page_size = 100
    node_batch_size = 50
    updated_nodes_only = true
    disable_cidr_filter = true
    cidr_filter = "0.0.0.0/0"
  [data_feed_service.v1.sys.log]
    level = "debug"
```

- [x] Tests added/updated? Patching tested manually and fqdn verified in service now 
- [ ] Docs added/updated? NA


api/config/data_feed/config_request.proto
```diff
@@ -11,10 +11,13 @@ import "google/protobuf/wrappers.proto";
 message ConfigRequest {
 	option (chef.automate.api.service_config) = {name: "data-feed-service"};
 
-	V1 v1 = 1;
+    reserved 1 to 2;
+
+	V1 v1 = 3;
```
jaym, line 16: If you do this, people will lose any configuration they've set on your service.

api/config/data_feed/config_request.proto
```diff
@@ -11,10 +11,13 @@ import "google/protobuf/wrappers.proto";
...
 
 	message V1 {
 		System sys = 1;
+        Service svc = 2;
```
jaym, line 20: I don't think you need this?

api/config/data_feed/config_request.proto
satellite15: both these lines?

```
    reserved 1 to 2;

	V1 v1 = 3;
```

api/config/data_feed/config_request.proto
satellite15: cool, thanks. Seen it being used in other config_requests.

api/config/data_feed/config_request.proto
stevendanna: If you change the field id for `v1`, it means previously saved protos will no longer populate the v1 field.  You can remove `reserved 1 to 2;` and keep `V1 v1 = 1;` as before.

PR #3958, opened 2020-06-22
Use latest available chef-load in our pipeline

jaym:


terraform/test-environments/modules/chef_load_instance/variables.tf
```diff
@@ -52,7 +52,7 @@ variable "chef_environment" {
 #
 
 variable "chef_load_channel" {
-  default     = "stable"
+  default     = "unstable"
```
lancewf, line 55: Yeah, this makes sense because we normally test chef load with our own origins.

PR #3959, opened 2020-06-22
Automate-3576 right panel + bottom project banner issues

tarablack01:
when a project update is in progress it takes up a 54 px at the bottom of the screen. We've got a few right panel issues with this panel.

https://app.zenhub.com/workspaces/ui-team-5cba23a3b14fdc05970c8f87/issues/chef/automate/3576

### :+1: Definition of Done
hides information, we should either add a scroll bar or some extra space to the bottom of those panels so folks can access all of the information.

hab studio enter
export DEVPROXY_URL=localhost
start_automate_ui_background
start_all_services
load compliance data
start a project update
navigate to compliance > reports > nodes > some node > scan history
notice you cannot scroll all the way to the bottom

- [ ] Tests added/updated?
- [ ] Docs added/updated?

Screen Shot 2020-06-22 at 12 12 18 PM
Screen Shot 2020-06-22 at 12 12 30 PM

### DIFF
--- a/components/automate-ui/src/app/ui.component.html
+++ b/components/automate-ui/src/app/ui.component.html
@@ -24,7 +24,10 @@
   </div>
 
   <!-- App Content -->
-  <div id="app-content-wrapper" [attr.hasNotifications]="layoutFacade.hasGlobalNotifications() ? true : null" [ngStyle]="layoutFacade.getContentStyle()">
+  <div id="app-content-wrapper"
+    [attr.hasNotifications]="layoutFacade.hasGlobalNotifications() ? true : null"
+    [attr.hasUserNotifications]="layoutFacade.hasUserNotifications() ? true : null"
+    [ngStyle]="layoutFacade.getContentStyle()">
     <div id="main-content-wrapper">
       <div id="notifications-wrapper" class="sticky-notifications">
         <app-chef-notifications></app-chef-notifications>
--- a/components/automate-ui/src/styles.scss
+++ b/components/automate-ui/src/styles.scss
@@ -13,6 +13,7 @@
 
   &[hasUserNotifications] {
     chef-side-panel {
+      // Adjust height when global notifications are visible.
       max-height: calc(100% - 120px);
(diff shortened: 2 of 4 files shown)


components/automate-ui/src/app/ui.component.html
```diff
@@ -24,7 +24,10 @@
   </div>
 
   <!-- App Content -->
-  <div id="app-content-wrapper" [attr.hasNotifications]="layoutFacade.hasGlobalNotifications() ? true : null" [ngStyle]="layoutFacade.getContentStyle()">
+  <div id="app-content-wrapper"
+    [attr.hasNotifications]="layoutFacade.hasGlobalNotifications() ? true : null"
+    [attr.hasUserNotifications]="layoutFacade.hasUserNotifications() ? true : null"
```
msorens, line 29: Do these have to evaluate to null on false or would this change be sufficient?
```suggestion
    [attr.hasNotifications]="layoutFacade.hasGlobalNotifications()"
    [attr.hasUserNotifications]="layoutFacade.hasUserNotifications()"
```

components/automate-ui/src/styles.scss
```diff
@@ -11,11 +11,23 @@
     top: 68px;
   }
 
+  &[hasUserNotifications] {
+    chef-side-panel {
+      max-height: calc(100% - 120px);
```
msorens, line 16: Could you add a comment that indicates what the 120px is?
Same thing for the 162px below. Thx!

components/automate-ui/src/app/ui.component.html
tarablack01: @msorens If you want to remove the attribute, you need to pass null 😕
https://stackoverflow.com/questions/36745734/how-to-add-conditional-attribute-in-angular-2

msorens: Kind of guessed that, but was not sure off the top of my head. Thanks for confirming. 👍

PR #3968, opened 2020-06-23
Specify the on-prem Builder API endpoint

yzl:
Make it explicit that the on-prem Builder API endpoint to use with the Habitat command-line tools is `/bldr/v1`. 

- [ ] Tests added/updated?
- [x] Docs added/updated?

### DIFF
--- a/components/automate-chef-io/content/docs/on-prem-builder.md
+++ b/components/automate-chef-io/content/docs/on-prem-builder.md
@@ -54,7 +54,7 @@ Chef Automate and Chef Habitat Builder require:
 * `systemd` as the init system
 * `useradd`
 * `curl` or `wget`
-* The shell that starts Automate should have a max open files setting of at least 65535
+* The shell that starts Chef Automate should have a max open files setting of at least 65535
 * Run the installation and bootstrapping procedures as the superuser or use `sudo` at the start of each command.
 
 ### Unsupported Topologies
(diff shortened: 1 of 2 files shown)


components/automate-chef-io/content/docs/on-prem-builder.md
```diff
@@ -231,6 +231,17 @@ Select your Gravatar icon on the top right corner of the Chef Habitat Builder we
 
 Once you are signed in to the Chef Habitat Builder UI, select the **New Origin** button and enter `core` as the name for the origin.
 
+## Access Chef Habitat Builder On-prem With Habitat Command-Line Tools
```
yzl, line 234: I do not love this heading; if you change it, make sure to change the corresponding link as well.

integration/tests/upgrade_current_master_habdev.sh
```diff
@@ -15,6 +15,19 @@ Environment=CHEF_DEV_ENVIRONMENT=true
 EOF
 }
 
+do_deploy() {
+    #shellcheck disable=SC2154
+    chef-automate deploy config.toml \
+        --product automate \
+        --product automate-dev \
```
jaym, line 22: this looks wrong. we shouldn't need to specify both. also, this test should not need automate-dev

jaym: https://github.com/chef/automate/pull/3935 is the fix

yzl: @jaym This is a direct cut and paste from the product-dev test; does it need to be fixed there as well?

I'll remove this and rebase with https://github.com/chef/automate/pull/3935.

jaym: that one just needs to specify automate-dev and not both

components/automate-chef-io/content/docs/on-prem-builder.md
```diff
@@ -231,6 +231,17 @@ Select your Gravatar icon on the top right corner of the Chef Habitat Builder we
...
+
+Use the URL `https://{{< example_fqdn "automate" >}}/bldr/v1` (note the `v1`) when
```
mjingle, line 236: @yzl Is this URL technically always the Builder API endpoint of the installation? Asking since you're referencing this section with the link text "Builder API endpoint of your installation" in line 343.

yzl: @mjingle If someone is using on-prem Builder deployed with Automate, then the Builder API endpoint is `https://{{< example_fqdn "automate" >}}/bldr/v1`.

If someone is using pre-Automate on-prem Builder or public Builder, then the endpoint is something like `https://example.com/bldr` without the v1.

components/automate-chef-io/content/docs/on-prem-builder.md
mjingle: I don't love this heading either, but I don't have anything better to offer.

PR #3985, opened 2020-06-25
Fixing Event Feed chef server filter bug

lancewf:

Fixing event feed chef-server filtering. The problem was that the UI was using the term 'chef_server' but the only valid terms were ''source_fqdn' or 'service_hostname'. The solution was to add 'chef_server' as an alias. 

### :+1: Definition of Done
Filtering the event feed with a chef-server works. 

1. `build components/config-mgmt-service && start_all_servcies`
1. `chef_load_nodes 100`
1. `chef_load_actions 100`
1. Go to https://a2-dev.test/dashboards/event-feed and add a chef-server filter. 

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/config-mgmt-service/params/parameters.go
+++ b/components/config-mgmt-service/params/parameters.go
@@ -137,7 +137,7 @@ func ConvertParamToActionBackend(parameter string) string {
 	switch parameter {
 	case "organization":
 		return backend.ActionOrganization
-	case "source_fqdn":
+	case "source_fqdn", "chef_server":
 		return backend.ActionSourceFQDN
 	case "event-type":
 		return backend.EntityTypeName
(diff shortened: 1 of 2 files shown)


components/config-mgmt-service/params/parameters.go
```diff
@@ -137,7 +137,7 @@ func ConvertParamToActionBackend(parameter string) string {
 	switch parameter {
 	case "organization":
 		return backend.ActionOrganization
-	case "source_fqdn":
+	case "source_fqdn", "chef_server":
```
vjeffrey, line 140: im curious -- why not also include "service_hostname" here? is that not another alias?

lancewf: The "service_hostname" term is the real value that we use to filter on. The `backend.ActionSourceFQDN` equals "service_hostname". 
If we do want to add that term as part of our API we could add it here. That term seemed to me to be internal use only. Maybe I should remove it from the unit test :thinking:

vjeffrey: 👍  i just didn't know what the term was

PR #3989, opened 2020-06-25
api docs: add docs for event feed

vjeffrey:

Adding api docs for event feed bits.

n/a

### :+1: Definition of Done
event feed api calls are documented

```
make serve
http://localhost:1313/docs/api/#operation/GetEventFeed
```

- [ ] Tests added/updated?
- [x] Docs added/updated?

### DIFF
--- a/components/automate-gateway/api/event_feed/request/event.proto
+++ b/components/automate-gateway/api/event_feed/request/event.proto
@@ -4,18 +4,29 @@ package chef.automate.api.event_feed.request;
 option go_package = "github.com/chef/automate/components/automate-gateway/api/event_feed/request";
 
 message EventFilter {
+  // Filters to be applied to the request.
   repeated string filter = 1;
+  // Earliest events to return.
   int64 start = 2;
+  // Latest events to return.
   int64 end = 3;
+  // Count of events to return per page.
   int32 page_size = 4;
+  // Used for pagination, to request results in ascending order.
   int64 after = 5;
+  // Used for pagination, to request results in descending order.
   int64 before = 6;
+  // Used for pagination, to bookmark the place from which to grab the next set of results.
   string cursor = 7;
+  // Used to group similar events into one.
   bool collapse = 8;
 }
 
 message EventCountsFilter {
+  // Filters to be applied to the request.
   repeated string filter = 1;
+  // Earliest events to return.
   int64 start = 2;
+  // Latest events to return.
   int64 end = 3;
 }
--- a/components/automate-gateway/api/event_feed/request/event.proto
+++ b/components/automate-gateway/api/event_feed/request/event.proto
@@ -16,9 +16,9 @@ message EventFilter {
   int64 after = 5;
   // Used for pagination, to request results in descending order.
   int64 before = 6;
-  // Used for pagination, to bookmark the place from which to grab the next set of results.
+  // Used for pagination, to bookmark next set of results.
   string cursor = 7;
-  // Used to group similar events into one.
+  // Used to group similar events together.
   bool collapse = 8;
 }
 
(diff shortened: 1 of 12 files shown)


components/automate-gateway/api/event_feed/request/event.proto
```diff
@@ -4,9 +4,13 @@ package chef.automate.api.event_feed.request;
...
 message EventFilter {
+  // Filters to be applied to the request.
   repeated string filter = 1;
+  // Earliest events to return.
   int64 start = 2;
+  // Latest events to return.
   int64 end = 3;
+  // Count of events to return per page.
   int32 page_size = 4;
   int64 after = 5;
```
vjeffrey, line 15: i don't know what after, before, cursor, or collapse mean

components/automate-gateway/api/event_feed/response/event.proto
```diff
@@ -11,12 +11,18 @@ message Events {
...
+  // Start time of the event.
   google.protobuf.Timestamp start_time = 3;
   string entity_name = 4;
+  // Requestor type on the event record.
   string requestor_type = 5;
+  // Requestor name on the event record.
   string requestor_name = 6;
+  // Hostname from which the record was gathered.
   string service_hostname = 7;
   string start_id = 8;
```
vjeffrey, line 27: start id? what is this?

components/automate-gateway/api/event_feed/request/event.proto
lancewf: Ok, these fields are confusing because we are not using pagination. When requesting pages after the first one we need to know the last event's ID and date. The last event's ID goes in the "cursor" field and the last event's date goes in the "before" field.  For the event feed page, we only want events in descending order so we will only use the "before" field.  If an API user wanted to get pages in ascending order they would use the "after" field when requesting pages after the first one.

lancewf: The "collapse" field is if we want to group similar messages like below. 
image

vjeffrey: thank you!

components/automate-gateway/api/event_feed/response/event.proto
lancewf: The "start_id" and "end_id" are only different when it is a grouped/collapsed event. For grouped events, we need to know the start date, end date, start ID, and end ID to request all the events in the grouped to show in the right side panel (shown below).

This is used similarly to the above when requesting events. 

image

components/automate-gateway/api/event_feed/request/event.proto
```diff
@@ -4,18 +4,29 @@ package chef.automate.api.event_feed.request;
...
   int64 end = 3;
+  // Count of events to return per page.
   int32 page_size = 4;
+  // Used for pagination, to request results in ascending order.
   int64 after = 5;
+  // Used for pagination, to request results in descending order.
   int64 before = 6;
+  // Used for pagination, to bookmark the place from which to grab the next set of results.
   string cursor = 7;
+  // Used to group similar events into one.
```
susanev, line 21:
```suggestion
  // Used to group similar events together.
```

components/automate-gateway/api/event_feed/request/event.proto
```diff
@@ -4,18 +4,29 @@ package chef.automate.api.event_feed.request;
...
   int64 start = 2;
+  // Latest events to return.
   int64 end = 3;
+  // Count of events to return per page.
   int32 page_size = 4;
+  // Used for pagination, to request results in ascending order.
   int64 after = 5;
+  // Used for pagination, to request results in descending order.
   int64 before = 6;
+  // Used for pagination, to bookmark the place from which to grab the next set of results.
```
susanev, line 19:
```suggestion
  // Used for pagination, to bookmark the next set of results.
```

PR #4006, opened 2020-06-30
Provide a recovery message on failed attempts to run a restore.

yzl:
The intention of the recovery message for a failed restore is to help the user get to a known state from which the restore can proceed correctly. See https://github.com/chef/customer-bugs/issues/187.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-cli/cmd/chef-automate/backup.go
+++ b/components/automate-cli/cmd/chef-automate/backup.go
@@ -28,6 +28,16 @@ import (
 	"github.com/chef/automate/lib/io/fileutils"
 )
 
+const restoreRecovery = `Check the logs (journalctl -u chef-automate) for errors related
+to the failed restore.
+
+Before retrying the restore process, remove the file %s.
+
+Under normal circumstances, restoring over an existing Chef Automate installation where
+some services are down will work correctly. If such a restore does not work correctly, ensure that
+either all services are up (chef-automate restart-services)
+or all services are down (chef-automate stop) before retrying the restore.`
+
 var backupCmdFlags = struct {
 	noProgress     bool
 	requestTimeout int64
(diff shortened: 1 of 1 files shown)


components/automate-cli/cmd/chef-automate/backup.go
```diff
@@ -28,6 +28,15 @@ import (
...
 )
 
+const restoreRecovery = `Check the logs (journalctl -f -u chef-automate) for errors related
+to the failed restore.
+
+Before retrying the restore process, remove the file %s.
+
+If you are restoring over an existing Chef Automate installation, also ensure that
+installation is in a known state such that either all services are up (chef-automate restart-services)
+or services are down (chef-automate stop) before retrying the restore.`
```
susanev, line 38:
```suggestion
Before restoring over an existing Chef Automate installation all services need to be either running or down. The restore will not work if it is run while services are in an unknown state. To restart services use chef-automate restart-services. To stop services use chef-automate stop
```

idk if thats any better?

yzl: @susanev Under normal circumstances, restore will work even if some of the services are down. In the case of the issue that motivated adding the recovery message, the restart was necessary because of weird state further down the stack. I'll think about if/how to rephrase.

components/automate-cli/cmd/chef-automate/backup.go
```diff
@@ -28,6 +28,16 @@ import (
 	"github.com/chef/automate/lib/io/fileutils"
 )
 
+const restoreRecovery = `Check the logs (journalctl -f -u chef-automate) for errors related
```
yzl, line 31: @chef/automate-deployment-team I could use input/feedback on what it is we want to say in the recovery message. This would have gotten the customer who filed the original ticket to a place where they could restore, but I'm not sure about it as general advice.

ryancragun: I think the only thing that might be confusing is the `-f`. If we're following we'll probably only see new log lines and not what potentially went wrong.

Issue #4015, opened 2020-07-01
compliance date picker woes


susanev:
## overview
the data picker on the compliance page has issues when set in Australian eastern time, and prolly others. also seems to be a problem on the event feed.

## repro steps
- around the utc day change over set your system time to Australian eastern time
- go to the compliance reports page, and use the date picker
- notice when you pick a date it sets to the previous date

## expected behaviour
- the date you select is shown in the selector and that data is shown

SEAjamieD:
Can we confirm that - as of today, we would like the date picker to display UTC date?

jonong1972:
@SEAjamieD Yes. We would like the date picker to display UTC date.

PR #1624, opened 2019-09-16
Fixing time zone problems

lancewf:
### :nut_and_bolt: Description: What code changed, and why?
Fixing a bug where the compliance reporting page would not allow the user in UTC+1 view the current day. The problem was the reportQuery state was being changed unintentionally. There was also some sections of code that when converting from moment to date was switching the current hour because of time zones. 

### :+1: Definition of Done

When in timezone UTC+1 the current date can be selected. 

### :athletic_shoe: How to Build and Test the Change
1. Switch your local computer to London time (UTC+1)
1. `build components/automate-ui && start_all_services`
1. Open the https://a2-dev.test/compliance/reports/overview
1. Ensure today's UTC date is selected. 
1. Ensure the date can be switched to yesterday back to today. 

### :white_check_mark: Checklist

- [x] Tests added/updated?
- [ ] Docs added/updated?

### :camera: Screenshots, if applicable

PR #4030, opened 2020-07-02
Jamie/aus utc bug

SEAjamieD:
The date picker on compliance reports page is displaying the previous date to customers in Australia time zone (and possibly others).  We would like the date picker to always display in UTC.

Addresses: https://github.com/chef/automate/issues/4015
Need to be careful that this does not undo all of work in https://github.com/chef/automate/pull/1624

### :+1: Definition of Done
The Date picker is always displaying UTC time.

make serve

navigate to https://a2-dev.test/compliance/reports/overview
By default, the date should match up with the current UTC Date
Select a date in the future 
  - the url parameters should match the selected date.
  - the date selector should display the date that was clicked on.
  - the reports should display from the date that was clicked on.
Repeat this flow several times for several different dates.

Change your timezone to locations in Japan, Australia, US, Europe, Russia, etc.
Repeat steps from above in each location.

** to change your timezone in chrome dev tools click on the three dot menu in devtools > more tools > sensors

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.spec.ts
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.spec.ts
@@ -80,14 +80,19 @@ describe('ReportingComponent', () => {
   });
 
   using([
-    ['01', '01', '23', 'an hour before end of day'],
-    ['04', '22', '00', 'exactly end of day'],
-    ['06', '30', '01', 'an hour after end of day']
+    ['01', '01', '23', 'an hour before end of day']
+    // ['04', '22', '00', 'exactly end of day'],
+    // ['06', '30', '01', 'an hour after end of day']
   ], function (month: string, day: string, hour: string, description: string) {
-    it(`displays the date correctly in UTC timestandard for ${description}`, () => {
-      const originalDate = moment.utc(`2019${month}${day}-${hour}01`, 'YYYYMMDD-HHMM');
+    xit(`displays the date correctly in UTC timestandard for ${description}`, () => {
+      const originalDate = moment.utc(`2019-${month}-${day}-${hour}00`, 'YYYY-MM-DDZ');
+      const start = moment.utc(originalDate).subtract(6, 'days');
+
       component.reportQuery.setState({
-        endDate: originalDate
+        startDate: start,
+        endDate: originalDate,
+        interval: 0,
+        filters: []
       } as ReportQuery);
       fixture.detectChanges();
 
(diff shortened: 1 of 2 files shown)


components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.spec.ts
```diff
@@ -84,9 +84,9 @@ describe('ReportingComponent', () => {
     ['00', 'exactly end of day'],
     ['01', 'an hour after end of day']
   ], function (hour: string, description: string) {
-    it(`maps hour correctly without timezone for ${description}`, () => {
+    it(`maps hour correctly to UTC timestandard for ${description}`, () => {
```
msorens, line 87: Right now this is a phantom test: we have not demonstrated that the fix does anything useful, simply because the test passed before the fix and the test _without change_ still passes after the fix. What we need to do is revise the set of tests so that:
(a) some of the tests fail with the original code under test, and
(b) all of the tests pass with the fix in place.
The set of equivalence classes already present are looking at the boundaries of the end-of-the-day but are oblivious to UTC boundaries. (Indeed, if you examine `dateBefore` it shows `isUTC` is false, while in the production code `convertMomentToDate` will be called with `isUTC` set true.)
Ultimately: Should the hour be coming back the same as the input, or should it be adjusted by the UTC offset?

Furthermore, this set of tests is only examining the hour.  what we are more concerned with in this PR is the month and day. We should have tests, for example, that emulate the Australian situation, coming in with e.g. July 2 and getting July 1 returned (and a nice boundary value of coming in with July 1 and getting June 30 returned).

And a minor point: `dateBefore` and `dateAfter` lead one to think (incorrectly) that the dates are sequenced. I would suggest `originalDate` and `resultDate` or some such.

SEAjamieD: @msorens - This is still a phantom test but i'm trying to wrap my head around the best way to do this.

The reason we were seeing a day behind is interesting but here's a small example: https://stackoverflow.com/questions/17855842/moment-js-utc-gives-wrong-date

Simply, When the new date was being created, it was done in local time which sets the hours to 00 and then gives us a day behind when formatted.

I'd appreciate if you might help me figure out the best way to test this, because the issue doesn't present itself until its actually formatted, so the tests will pass for the current output, and for the fix output.  I thought I would just use the DatetimePipe in the test, but it didn't like that for some reason.

SEAjamieD:
@susanev - This branch still needs updated tests - but I'm ok with _testing_ it in the wild as long as we can get tests done before its actually released.

PR #4046, opened 2020-07-08
Status fixes and support for status failed in ingestion

alexpop:
This PR adds support in the ingestion & reporting parts of Automate for InSpec reports that contained profiles with a `failed` status. The related InSpec change is here: https://github.com/inspec/inspec/pull/5128

A new report has been added in order to integration-test this change:
`../test_data/audit_reports/2018-04-02_ubuntu(4)-alpha-myskippy(s)-myfaily(f)-apache(f)-linux(p)-failed.json`

I made some improvements to the compliance-service Makefile as running it on Linux was not as delightful as it was on OSX.

It's hard to end-to-end test this at the moment with the audit cookbook since the InSpec change hasn't merge yet.
Alternatively, you can bring up an Automate server from this branch, rebuild the automate gateway, compliance service and automate-ui. Then ingest `2018-04-02_ubuntu(4)-alpha-myskippy(s)-myfaily(f)-apache(f)-linux(p)-failed.json` via `curl` or `test_data/audit_reports/send_to_data_collector.sh`. 

I've just done that and captured the following screenshots after the above report has been ingested. What you are basically looking at is a report for an InSpec run with 4 profiles:
 * 1 (myskippy) that has **skipped** without any control being run `use-case supported by InSpec`
 * 2 (myfaily, apache-baseline) that have **failed** without any control being run `use-case currently being implemented in InSpec`
 * 1 (linux-baseline) that has **passed** with one **passed** control executed `use-case supported by InSpec`

Screenshot 2020-07-08 at 15 11 31

^^^ I’ve put the failed message in those right side `Profiles` tooltips, but I think we should make that more visible when someone lands on the node/scan-history page where a profile has raised a runtime exception. The idea is to make a runtime error easier to notice and in some cases, the error can be extremely verbose and won't fit nicely in a tooltip.

Screenshot 2020-07-08 at 15 11 10
Screenshot 2020-07-08 at 15 11 01


components/compliance-service/ingest/pipeline/processor/summary.go
```diff
@@ -63,7 +75,22 @@ func ComplianceShared(in <-chan message.Compliance) <-chan message.Compliance {
 			}
 			msg.Shared.PerProfileSums = perProfileSums
 			msg.Shared.AllProfileSums = totalSum
-			msg.Shared.Status = compliance.ReportComplianceStatus(totalSum)
+
+			if failedProfiles > 0 {
+				// If at least one of the profiles in the report comes in with "failed" status, we ignore the totalSum aggregation which might have all
```
vjeffrey, line 80: thanks for including the comments here and all over, these will be super useful when our future selves try to figure out all this stuff

components/compliance-service/ingest/ingestic/ingestic.go
```diff
@@ -267,6 +274,12 @@ func (backend *ESClient) InsertInspecReport(ctx context.Context, id string, endT
 
 func (backend *ESClient) InsertInspecProfile(ctx context.Context, data *relaxting.ESInspecProfile) error {
 	mapping := mappings.ComplianceProfiles
+	if len(data.Controls) == 0 {
```
rickmarry, line 277: is it an all or zero thing?  in other words, if we're supposed to have 10 controls, would it be possible to have less than that in a report, yet still more than 0?

components/compliance-service/ingest/ingestic/mappings/comp-rep-date.go
```diff
@@ -408,6 +408,9 @@ var ComplianceRepDate = Mapping{
             "skip_message": {
               "type": "keyword"
             },
+            "status_message": {
```
rickmarry, line 411: we're going to need to bump the compliance index version in mappings.go

vjeffrey: wait what? i thought this was a new field only and as such doesn’t require a migration? we need to talk to support if we’re doing a migration..

components/compliance-service/ingest/ingestic/ingestic.go
alexpop: I've never seen that before. When we get controls from inspec, we get all of them.
Even this situation with 0 controls it's only happening on profiles with run status of skipped/failed.

components/compliance-service/ingest/ingestic/mappings/comp-rep-date.go
alexpop: > by not specifying, status_message will default to type "text"

We are specifying it in the mapping here and the structs. Earlier Automate versions will get this `status_message` but without data collector extraction and existance in the go structs, the field will not reach ElasticSearch.

PR #4052, opened 2020-07-08
Add product versions to release notes

kagarmoe:

Adds a section for Chef product versions to the release notes

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/.expeditor/publish-release-notes.sh
+++ b/.expeditor/publish-release-notes.sh
@@ -33,8 +33,8 @@ pushd ./automate.wiki
 ## Chef Product Versions
 
 This release depends on:
-  Chef Habitat version
-  Chef InSpec version
+- Chef Habitat version
+- Chef InSpec version
 
 EOH
 
--- a/.expeditor/publish-release-notes.sh
+++ b/.expeditor/publish-release-notes.sh
@@ -35,9 +35,8 @@ pushd ./automate.wiki
 This release uses:
 - Chef Habitat version
 - Chef Habitat Builder version
-- Chef Infra Client version
 - Chef Infra Server version
-- Chef InSpec version 
+- Chef InSpec version
 
 EOH
 
--- a/.expeditor/publish-release-notes.sh
+++ b/.expeditor/publish-release-notes.sh
@@ -30,6 +30,12 @@ pushd ./automate.wiki
 ## Backward Incompatibilities
 -
 
+## Chef Product Versions
+
+This release depends on:
+  Chef Habitat version
+  Chef InSpec version
+
 EOH
 
   # Push changes back up to GitHub
--- a/.expeditor/publish-release-notes.sh
+++ b/.expeditor/publish-release-notes.sh
@@ -32,9 +32,12 @@ pushd ./automate.wiki
 
 ## Chef Product Versions
 
-This release depends on:
+This release uses:
 - Chef Habitat version
-- Chef InSpec version
+- Chef Habitat Builder version
+- Chef Infra Client version
+- Chef Infra Server version
+- Chef InSpec version 
 
 EOH
 


.expeditor/publish-release-notes.sh
```diff
@@ -30,6 +30,12 @@ pushd ./automate.wiki
 ## Backward Incompatibilities
 -
 
+## Chef Product Versions
+
+This release depends on:
```
susanev, line 35: is uses a better word here? im not sure
do we want the bldr version too?

susanev: oh theres also chef infra server right?

mjingle: @susanev I'm happy to add those two. Anything else before I add another commit?

susanev: well hold on, i kinda want @vjeffrey or someone else to confirm for me @ryancragun thoughts?

.expeditor/publish-release-notes.sh
```diff
@@ -30,6 +30,15 @@ pushd ./automate.wiki
 ## Backward Incompatibilities
 -
 
+## Chef Product Versions
+
+This release uses:
+- Chef Habitat version
+- Chef Habitat Builder version
+- Chef Infra Client version
```
vjeffrey, line 38: infra client version shouldn't be included.

inspec, builder, habitat, and infra server make sense to me

@chef/automate-deployment-team whaddyall think?

.expeditor/publish-release-notes.sh
```diff
@@ -30,6 +30,14 @@ pushd ./automate.wiki
 ## Backward Incompatibilities
 -
 
+## Chef Product Versions
+
+This release uses:
```
yzl, line 35: suggestions only:
1. "This Automate release"
2. Provide a link to the manifest.

susanev: +1 in linking to the manifest

mjingle: What's the manifest again?

kagarmoe:
What are we trying to achieve by linking to the manifest? What specific need does it solve and are the release notes the best place to solve it?

.expeditor/publish-release-notes.sh
mjingle: Okay, I found what you were talking about - https://github.com/chef/automate/blob/master/products.meta - which is updated by https://github.com/chef/automate/blob/master/.expeditor/create-manifest.rb

kagarmoe: What are we trying to achieve by linking to the manifest? What specific need does it solve and are the release notes the best place to solve it?

yzl: @kagarmoe The manifest provides a set structure for users to probe a release in an automated way to see if, e.g. there's a new version of the specific product they care about. It's the same information that the list of product versions provides, but people can parse json to get at the data instead of scraping text.

kagarmoe: I can see what the manifest contains, but I'm not sure that the use you're proposing is a common enough information need for linking out to it from the release notes. I see it as potentially more confusing than enlightening for many users. 

With the addition of links out to PRs and also to a Changelog, we're already transforming the information context of the release notes. I'm not sure that the information added by the manifest is equally clear and actionable.

We could publish the manifest in other ways. For example, we could automatically ingest it as a page in the "Reference" section, so it would update with every release.

yzl: Meh. We can see if the other changes help people find what they need.

kagarmoe: OK.

stevendanna: Just for clarity for those following along. The manifest linked above isn't _quite_ what was being referred to. Rather, I believe yzl and susan were likely referring to the per-release version manifests.  Here is an example of one:

https://packages.chef.io/manifests/current/automate/20200529133631.json

yzl: Yes, the manifest that Steven points to is the one I meant.

kagarmoe: Yes, but could you define what information need that linking to the manifest solves for users? Which users want this information? Under what circumstances? And what actions--if any--should a user take based on this information?

mjingle: @yzl Would you be so kind as to file an issue on implementing manifest so we can investigate this further? Also, please include information that addresses @kagarmoe's above questions. I appreciate you bringing this up, but this discussion shouldn't block the merging of this PR.

yzl: It tells the user:

1. that the manifest exists
2. that there is a programmatic, low-granularity way for them to see if new version of the service is different from the version of the service they already have
3. more detail about what comprises a package, which is often the kind of thing change control boards want to know. (I think they'd actually want more detail 

The release manifests are already published, and users can find out about them via various channels (the Slacks, CFT, and so forth); I thought it would be convenient to link to them in the release notes, that is all.

What is potentially more useful for security teams and CCBs to know about is the license scout output that says what all the dependencies and versions that comprise a new package are, e.g. https://packages.chef.io/licenses/automate/20190820163418.json

stevendanna: In addition to the "security team" user that @yzl mentioned, I can think of a couple of others:

- Sysadmin/Operators:  Often we use release notes when things have gone wrong and we are trying to work backwards from a set of error messages and google results to understand some problem. For example, the administrator upgrades and suddenly finds that something isn't working.  The sys-admin tracks this down to their monitoring software being incompatible with the version of Elasticsearch that we are using.  They then want to go through the release notes to find where we did minor version bumps of Elasticsearch to see if this new problem is because of a change in their monitoring system or a change in our Elasticsearch version. (This is an example, but this is the kind of thing sys-admins do a half-dozen times a week across the many systems they manage).

- Chef Solutions Engineers and Customer who use Airgap: You need the manifest to create an airgap bundle.  While you can remember the URL based on the version number, sometimes it is nicer to be able to copy and paste the link.

mjingle: @yzl and @stevendanna Thank you for these answers! If you would like, you can comment further on this Issue https://github.com/chef/automate/issues/4067 I'm looking forward to working on this.

Issue #3639, opened 2020-05-11
desktop: places to update to real data


vjeffrey:
## User Story
@scottopherson has done some AMAZING work to make the desktop view all pretty and nice!

there are a few views/features within this view that still need to be updated with real data.
here is a list of those,

**desktop detail view**
all details for that table
 -API:  GET: https://a2-dev.test/api/v0/cfgmgmt/nodes?filter=node_id:<node-id>

the status graph
 -API:  GET: https://a2-dev.test/api/v0/cfgmgmt/node_runs_daily_status_time_series?node_id=<node_id>&days_ago=3

the download button
 -API: POST: https://a2-dev.test/api/v0/cfgmgmt/reports/export 
UI example: https://github.com/chef/automate/blob/fbe5d3654a8d429e2ed3e5170724e3c94a96032e/components/automate-ui/src/app/services/node-details/node-runs.service.ts#L62

the link to infra run
 -navigate to https://a2-dev.test/infrastructure/client-runs/node-id/runs/run-id

**desktop list view**
filter buttons/counts
 -API: GET https://a2-dev.test/api/v0/cfgmgmt/nodes_field_value_counts?terms=platform&terms=status
PR with examples: https://github.com/chef/automate/pull/3360

download button
 -API: POST: https://a2-dev.test/api/v0/cfgmgmt/nodes/export
UI example: https://github.com/chef/automate/blob/d99e44ac7227d2b0d5f58ce33576c62efce11bc7/components/automate-ui/src/app/entities/client-runs/client-runs.requests.ts#L33
attributes dropdown
 -is this static, @apriofrost ?

**dash view**
filtering when selecting "time since last checked in" item
The "start" and "end" parameters filter on the checked-in date time. 
 -API: https://a2-dev.test/api/v0/cfgmgmt/nodes?pagination.page=1&pagination.size=100&start=2020-05-03T21:48:57Z&end=2020-05-06T21:48:57Z

we also need to add the filters/sorting/pagination in the url

**_everyone please feel free to add to this issue as needed!_**

SEAjamieD:
Gonna start working on the Dash portion of this: added an issue just for tracking purposes.  https://github.com/chef/automate/issues/3741

scottopherson:
@SEAjamieD what's your plan? That date filtering is not really part of the dash overview; that's filtering of the desktop list which is an area I'm currently focusing on.

SEAjamieD:
[...]

PR #4058, opened 2020-07-09
Fill out remaining dashboard desktop detail info

scottopherson:

This commit replaces the remaining hardcoded data in the dashboard desktop detail view with real data fetched from the cfgmgmt API.

https://github.com/chef/automate/issues/3639

### :+1: Definition of Done

There is no longer any hardcoded data in the dashboard desktop detail view.

- Start all services and UI
- Have some data to look at in the dashboard:
```
# From hab studio
for _ in {1..50}; do m=$((RANDOM % 4)); n=$((RANDOM % 4)); generate_chef_run_failure_example | jq --arg msg "Error $n occurred" --arg type "Chef::ExampleError$m" '.error.message = $msg | .error.class = $type' | send_chef_data_raw; done
```
- Navigate to `a2-dev.test/desktop` and select any item on the dashboard to bring up the desktop list view
- Select a desktop from the desktop list to bring up the desktop detail view


scottopherson:
Occasionally, a node will be returned with some missing data attributes (usually if the node has a lot of failed/missing runs), but `timezone` seems to always return undefined 🤔

PR #4061, opened 2020-07-13
Fix ontop backups for builder

jaym:
If you were to take a backup of builder, then on another machine, install automate + builder, and restore the backup, it would fail because we'd change the minio secret. Minio doesn't like this because it encrypts config with that secret now. We don't actually care about that config, so this PR detects that case and blows away the config

### DIFF
--- a/components/automate-minio/habitat/hooks/run
+++ b/components/automate-minio/habitat/hooks/run
@@ -2,9 +2,6 @@
 
 exec 2>&1
 
-chmod 0600 "{{pkg.svc_config_path}}/secrets"
-source "{{pkg.svc_config_path}}/secrets"
-
 mkdir -p "{{pkg.svc_data_path}}/depot"
 
 # Minio requires TLS certs to be in a 'certs' subdirectory of the config-dir.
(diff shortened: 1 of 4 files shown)


components/automate-minio/habitat/hooks/run
```diff
@@ -2,9 +2,6 @@
 
 exec 2>&1
 
-chmod 0600 "{{pkg.svc_config_path}}/secrets"
```
ryancragun, line 5: This implies that we'll no longer allow people to set credentials in the config. It's probably best to let the credential resolvers look for AWS creds in the standard location but I'm not sure if this is going to break people or not.

jaym: those values were never set. The file was copied and pasted from the backup-gateway config:
```
[94][default:/src:0]# cat /hab/svc/automate-minio/config/secrets 

[95][default:/src:0]# 

```

jaym:
```
{{~#if cfg.gateway.backup.s3.credentials.access_key}}
export AWS_ACCESS_KEY_ID="{{cfg.gateway.backup.s3.credentials.access_key}}"
{{~/if}}
{{~#if cfg.gateway.backup.s3.credentials.secret_key}}
export AWS_SECRET_ACCESS_KEY="{{cfg.gateway.backup.s3.credentials.secret_key}}"
{{~/if}}
{{~#if cfg.gateway.backup.s3.credentials.session_token}}
export AWS_SESSION_TOKEN="{{cfg.gateway.backup.s3.credentials.session_token}}"
{{~/if}}
```

Thats the contents of the deleted file. It's impossible to set those values as this is not the backup gateway

ryancragun: :facepalm: I thought I was looking at the bgw

PR #4083, opened 2020-07-17
Enable desktop list column hide/show selection

scottopherson:

This commit wires up the insights-attribute-dropdown to the desktop list view to enable the column hide/show functionality.

d

https://github.com/chef/automate/issues/3639

### :+1: Definition of Done

When changes to the attributes dropdown are applied, the desktop table columns react accordingly. 

- Start all services and UI
- Have some data to look at in the dashboard:
```
# From hab studio
for _ in {1..50}; do m=$((RANDOM % 4)); n=$((RANDOM % 4)); generate_chef_run_failure_example | jq --arg msg "Error $n occurred" --arg type "Chef::ExampleError$m" '.error.message = $msg | .error.class = $type' | send_chef_data_raw; done
```
- Navigate to `a2-dev.test/desktop` and select any item on the dashboard to bring up the desktop list view
- Interact with the attributes dropdown in desktop list view


components/automate-ui/src/app/page-components/insight-attributes-dropdown/insight-attributes-dropdown.model.ts
```diff
@@ -1,53 +0,0 @@
-export interface FilterOption {
```
scottopherson, line 1: Moved these types into desktop.model since there was overlap.

components/automate-ui/src/app/page-components/insight-attributes-dropdown/insight-attributes-dropdown.component.ts
```diff
@@ -1,87 +1,76 @@
-import { Component, OnInit, Input, Output, EventEmitter } from '@angular/core';
```
scottopherson, line 1: In order to allow state to be pushed into it from the reducer, I needed to refactor this component so that less state was being tracked/manipulated within it. Refactored bits are similar to projects-filter-dropdown.

components/automate-ui/src/app/entities/desktop/desktop.requests.ts
```diff
@@ -77,6 +77,20 @@ export interface RespDesktop {
   deprecations_count: number;
   chef_version: string;
   domain: string;
```
scottopherson, line 79: Need to include more data fields now that you can choose to view them in desktop list.

PR #4085, opened 2020-07-17
Disable xpack in deprecated ES backup test

jaym:
We strip xpack from ES. We cannot connect to an ES cluster
with xpack enabled, which is the deprecated way of way of
using external ES.


jaym:
we don't need this anymore

PR #4090, opened 2020-07-17
Event feed project update

lancewf:
Adding project update to the event-feed. This will only update the events from Chef Server Actions. This adds authz-service as a dependence. 

#444 

1. `build components/event-feed-service && build components/automate-deployment && build components/authz-service && start_all_services`
1. Add three chef-server events with different values organization values ("org1", "org2", and "org3"). Do not assign any projects. 
1. `generate_random_chef_server_event | jq '.data.chef_organization = "org1" | del(.Projects)' | send_event`
1. `generate_random_chef_server_event | jq '.data.chef_organization = "org2" | del(.Projects)' | send_event`
1. `generate_random_chef_server_event | jq '.data.chef_organization = "org3" | del(.Projects)' | send_event`
1. Create a project named "project1" with a rule of type "event" that matches "chef organization" with "org1"
1. Create a project named "project2" with a rule of type "event" that matches "chef organization" with "org2"
1. Run the project update
1. After the update is complete, open the event feed page. https://a2-dev.test/dashboards/event-feed.
1. Filter on project1 and ensure only one event is displayed.
1. Filter on project2 and ensure only one event is displayed.
1. Remove all project filters and ensure all three events are displayed.  

- [x] Tests added/updated?
- [ ] Docs added/updated?


components/authz-service/server/project_update_manager.go
```diff
@@ -48,6 +48,7 @@ var SerializedProjectUpdateDomainServices = []string{
 	"compliance",
 	"ingest",
 	"nodemanager",
+	"feed",
```
lancewf, line 51: Adds the event feed to one of the services that are called from authz during a project update.

components/event-feed-service/pkg/grpc/grpc.go
```diff
@@ -96,6 +98,38 @@ func Spawn(c *config.EventFeed, connFactory *secureconn.Factory) error {
...
+	err = project_update_lib.RegisterSerialTaskExecutors(projectUpdateManager, "feed", feedStore, authzProjectsClient)
+	if err != nil {
+		return err
+	}
+
+	if err := projectUpdateManager.Start(context.Background()); err != nil {
+		return err
+	}
+	defer projectUpdateManager.Stop() // nolint: errcheck
+
```
lancewf, line 132: Adding authz and the cereal project updater.

components/event-feed-service/pkg/integration_test/project_update_test.go
```diff
@@ -0,0 +1,414 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/chef/automate/api/interservice/authz"
+	"github.com/chef/automate/components/event-feed-service/pkg/feed"
+	"github.com/chef/automate/components/event-feed-service/pkg/persistence"
+	event "github.com/chef/automate/components/event-service/config"
+)
+
+func TestProjectUpdatePainlessElasticsearchScript(t *testing.T) {
```
lancewf, line 17: Most of these test came from here https://github.com/chef/automate/blob/4d90b019189761ea8fe0e833a4b0b95ff862c3df/components/ingest-service/integration_test/project_update_actions_test.go#L16

components/event-feed-service/pkg/persistence/project_update.go
```diff
@@ -0,0 +1,215 @@
...
+	return projectUpdateStatus, nil
+}
+
+func (efs ElasticFeedStore) CancelProjectUpdateTask(ctx context.Context,
+	projectUpdateID string, id project_update_lib.SerializedProjectUpdateTaskID) error {
+	logrus.Debug("running CancelProjectUpdateTask")
+	return efs.JobCancel(ctx, string(id))
+}
+
+func (efs ElasticFeedStore) updateEventProjectTags(
```
lancewf, line 115: This function does the project update. The code in the script variable is sent to elasticsearch and run on each event.

components/event-feed-service/habitat/config/config.toml
```diff
@@ -33,3 +33,8 @@ default_purge_after_days = 7
 {{~#eachAlive bind.cereal-service.members as |member|}}
 address = "{{member.sys.ip}}:{{member.cfg.port}}"
 {{~/eachAlive}}
+
+[authz]
+{{~#eachAlive bind.authz-service.members as |member|}}
```
danielsdeleo, line 38: We could fix this separately but I think this would not be valid if we actually had more than one item in `authz-service.members`. IIRC, TOML doesn't allow you to redefine keys, so something like this would be an error:

```
address = "member1:port"
address = "member2:port"
```

Then again we probably won't have more than one item in there any time soon so no rush. But the fix is to check for `@last` like this https://github.com/chef/automate/blob/e8684cce1a1d02e95ea9429896f9793d56a37b53/components/applications-service/habitat/config/config.toml#L13-L17

lancewf: Thank you for pointing that out. I update all the services in this config file.

Issue #3987, opened 2020-06-25
Create Rollout Metadata Collector CLI


danielsdeleo:
## User Story

In order to have contextual information related to policy change rollouts in Automate, we want to collect metadata associated with an infra code change from the client and submit this data to automate.o

## Definition of Done

* Collect all possible attributes of a rollout, as defined here: https://github.com/chef/automate/blob/master/api/external/cfgmgmt/response/rollouts.proto#L38
* It should have reasonable behavior whether or not the user checks their lock file into git.
* may need a separate callback/hook behavior for promotions(?)

Issue #2873, opened 2020-02-14
Desktop Offering View in Automate


nfisher-chef:

susanev:
>How do we guarantee this user only sees the Desktop Offering Dashboard view?

we can create a new access management role for this

kmacgugan:
- Need to store more than 24 hours worth of data

This is already true excepting the ohai data, which we only store the latest data for. Will we be needing historical ohai data or just the current ohai data? Ohai data can be very large and managing historical data introduces new technical challenges. We store the rest of the chef-client run reports going back to configurable data retention settings.

nfisher-chef:
> * Need to store more than 24 hours worth of data
> 
> This is already true excepting the ohai data, which we only store the latest data for. Will we be needing historical ohai data or just the current ohai data? Ohai data can be very large and managing historical data introduces new technical challenges. We store the rest of the chef-client run reports going back to configurable data retention settings.

yes, we will need historical Ohai data for certain values. Does that help? We don't need ALL of the data, but some.

kmacgugan:
Yes, it helps clarify, do we know what those keys/values are? One of the challenges we've had with ohai data is that it is largely dynamic (including the keys! sometimes it uses values as keys!) so it poses a challenge to mapping in ES.

nfisher-chef:
We are doing some discovery, so we will definitely know before we stop implementation! Data mapping should be done during our UX work.

danielsdeleo:
### Product Decision: Top Errors API/UI ###

The UI mockup shows a "Top Errors" panel. We had a discussion of how this should work over time, in the context of several use-case scenarios and what is the most feasible to implement today. 

We agreed that for an initial implementation, we will aggregate the top errors from the node state index, with a hard-coded filter to include only errors from the past 7 days.

In other words, we are counting the errors on a node-wise basis where the most recent Chef Client run for that error ended in failure, and the most recent Chef Client run occurred within the past 7 day time period.

This design does not support trend information or historical analysis use-cases. Work to support those will be scheduled for later.

PR #4109, opened 2020-07-21
Dan/rollouts from actions

danielsdeleo:

Create rollouts from Chef Server actions messages alongside the external metadata API. The preferred method for creating rollouts will be the client-side metadata collector, which provides more useful data about the code being rolled out, but we also wish to create rollouts from Chef Server action messages for the following cases:
* Ensure that we create at least a minimal record of new code rollouts for cases where the metadata collector isn't properly configured, the user takes out of band action on a system where there metadata collector isn't configured, etc.
* Give a preview of the feature to users who are deciding whether to adopt.

This behavior is implemented by forwarding policy update actions from the ingest service to the config management service, which creates rollout records for the incoming actions if they do not exist. More specifically:
* Add a bind at the hab and deployment service level so ingest can make a connection to config management
* Add a publisher in the ingest service actions pipeline to send the messages to config management
* Make an API endpoint for config management to accept policy update action data
* Insert a rollout in the rollouts table when a policy update action is received. In case of a conflict (i.e., we already have a record of this rollout), do nothing.
* Modify the API that the metadata collector uses to create rollouts, so that it does an upsert. The "happy path" will be that we get a metadata API call and an actions message for each new rollout, and we cannot control which one comes first. If the actions message comes first, we will have a "bare bones" record of the rollout in the database, and we need to add the extra metadata to that row. We use postgres' built-in upsert to help avoid concurrency issues.

#3987 #2873

* `build components/automate-deployment/ && build components/ingest-service/ && build components/config-mgmt-service/`
* I think the dev channel is behind master because of a pipeline issue. Because we changed deployment service and rebuilt from master, this means you need to `build components/event-feed-service/` :| 
* `start_all_services` and `start_chef_server` and `bootstrap_chef_user_data`
* `hab pkg install chef/chef-dk -bf` to get the `chef` CLI in your PATH
* Get a policyfile installed so you can push it to a chef server (you'll have to accept the Chef license)`pushd /tmp/ && git clone https://github.com/danielsdeleo/policyfile-jenkins-demo.git && cd policyfile-jenkins-demo && chef install`
* `chef push -c /tmp/.client.rb foo` to upload the policy to the Chef Server
* `config_mgmt_list_rollouts` to see the rollouts


components/config-mgmt-service/grpcserver/actions_handler.go
```diff
@@ -0,0 +1,53 @@
...
+	"google.golang.org/grpc/status"
+
+	"github.com/chef/automate/api/interservice/cfgmgmt/request"
+	"github.com/chef/automate/api/interservice/cfgmgmt/response"
+	"github.com/chef/automate/components/config-mgmt-service/backend/postgres"
+)
+
+func (s *CfgMgmtServer) HandlePolicyUpdateAction(ctx context.Context, req *request.PolicyUpdateAction) (*response.PolicyUpdateAction, error) {
+	// FIXME remove this or make it debug w/e
+	log.Infof("got HandlePolicyUpdateAction call %+v", *req)
```
lancewf, line 19: You have a note to remove this line

components/ingest-service/pipeline/publisher/config_mgmt_publisher.go
```diff
@@ -0,0 +1,92 @@
...
+	start := time.Now()
+
+	req := &request.PolicyUpdateAction{
+		PolicyName:         msg.InternalChefAction.EntityName,
+		PolicyGroup:        msg.InternalChefAction.ParentName,
+		PolicyRevisionId:   msg.InternalChefAction.RevisionId,
+		ChefServerFqdn:     msg.InternalChefAction.ServiceHostname,
+		ChefServerOrgname:  msg.InternalChefAction.OrganizationName,
+		ChefServerUsername: msg.InternalChefAction.RequestorName,
+		PolicyfileContent:  msg.InternalChefAction.Data,
```
lancewf, line 58: I will just parse out all the cookbooks from the "Data" field and add a new field to the PolicyUpdateAction.

components/ingest-service/pipeline/publisher/config_mgmt_publisher.go
```diff
@@ -0,0 +1,92 @@
...
+
+	logCtx := log.WithFields(log.Fields{
+		"message_id":  msg.ID,
+		"buffer_size": len(out),
+		"dur":         time.Since(start),
+	})
+
+	if err != nil {
+		logCtx.WithError(err).Error("Failed to publish actions message to config management service")
+		return
```
lancewf, line 74: If it fails to publish the action to the config-mgmt-service we should fail the pipeline with `msg.FinishProcessing(err)`

danielsdeleo: 👍 will do, I wasn't sure if we wanted to try a best effort to pass it to the next publisher

components/config-mgmt-service/grpcserver/actions_handler.go
danielsdeleo: haha, I looked for TODOs before I pushed

PR #4110, opened 2020-07-21
Wrap os/user to fall back to getent

### LINKED ISSUE #4105: Automate deploy fails when the linux hab user comes from external auth instead of local accounts
## Describe the bug

Automate is unable to complete the deployment step if `hab` the hab user is coming from an external auth like LDAP in Linux instead of as a local account in `/etc/passwd`.

Initially, we saw errors from `useradd` complaining that the user already completed and a workaround was added #4078 to get past this.  

After the workaround, we now see the following error when running `CHEF_AUTOMATE_SKIP_HAB_USER=true chef-automate deploy`

```
Jul 20 17:57:26 HOSTNAME.com hab[30310]: deployment-service.default(O): failed to start server: failed to initialize secret store: user lookup: user: unknown user hab
```

The underlying problem appears to be caused by GoLang not using libc to query for user and group UID/GIDs and instead only pulling them from `/etc/passwd` and `/etc/group`. This will cause the deploy to fail as it can't find the `hab` user and attempts to create it via `useradd`. Which will then fail as it already exists in LDAP, also any other code that attempts to fetch the `hab` UID/GID` will have problems as it won't get a valid ID.

## To Reproduce

1. configure Linux to use external LDAP user accounts (via /etc/nsswitch.conf)
2. create a `hab` user in LDAP
3. Attempt to install Automate `chef-automate deploy`, latest current version will throw an error 
    
    ```
    Creating Habitat user and group
    Error: setup failed: creating hab user failed: "useradd: user 'hab' already exists\n": exit status 9  
    DeployError: Unable to install, configure and start the
    ```

4. If using a version from the dev channel `CHEF_AUTOMATE_SKIP_HAB_USER=true chef-automate deploy` to skip the `useradd` error and get the error about the `secret store` and `hab`.

    ```
    Jul 20 17:57:26 HOSTNAME.com hab[30310]: deployment-service.default(O): failed to start server: failed to initialize secret store: user lookup: user: unknown user hab
    ```

## Expected behavior
Automate will complete the deployment successfully.

- OS: RHEL 7
- Browser: N/A
- Automate Build Number: latest current and dev builds

jaym:
Partial fix for #4105

If the hab user and groups and not local to the system, automate is unable to start. A good portion of the problem is that we use golang's `os/user` library to get UIDs and GIDs, which just looks at `/etc/passwd` and `/etc/group` when compiled without CGO. The bulk of this PR works around that. Instead of using `os/user`, we create out own user library which first tries `os/user`, the falls back to shelling out to `getent`. We can't shell out to any getent though, we need to find a system one as its going to know how to deal with the systems `/etc/nsswitch.conf`.

Unfortunately, this does not fix everything. The habitat supervisor links against a glibc that isn't going to lookup the nss modules (`libnss_*`) on the system. That means just because your system tools say a user/group exist, doesn't mean that hab will agree. To get around this, for now, we need to lay down a systemd unit overlay:
```
cat > /etc/systemd/system/chef-automate.service.d/custom.conf <<EOF
[Service]
Environment=LD_PRELOAD=/lib64/libnss_ldap.so
EOF
```
Doing this also makes all the places we depend on `core/coreutils` work. I think our main use-case would be `chown`.
Alas, we use busybox in a bunch of places to get chown, the `LD_PRELOAD` wont help.

Things that will happen in another PR:
- Stop using busybox and perhaps create a `chown` that can benefit from our `getent` hack in golang. Its also unclear we need to be using chown in a lot of places that we use it. automate-elasticsearch and automate-es-gateway currently complain, but automate-postgresql, automate-deployment, automate-cs-nginx, automate-ui, and automate-build-api-proxy need to be audited.
- We need changes to the supervisor so it will work without `LD_PRELOAD`


integration/tests/ldap_hab_user.sh
```diff
@@ -0,0 +1,85 @@
+#!/bin/bash
+
+#shellcheck disable=SC2034
+test_name="ldap_hab_user"
+test_backup_restore=true
+
+do_setup() {
```
stevendanna, line 7: Would it be helpful to add some assertion to this test that the hab user definitely doesn't exist in /etc/passwd?

integration/tests/ldap_hab_user.sh
```diff
@@ -0,0 +1,85 @@
...
+loginShell: /sbin/nologin
+EOF
+
+    authconfig --enableldap --enableldapauth --ldapserver=127.0.0.1 --ldapbasedn="dc=example,dc=local" --enablemkhomedir --update
+    systemctl restart nslcd
+
+    mkdir /etc/systemd/system/chef-automate.service.d
+    cat > /etc/systemd/system/chef-automate.service.d/custom.conf <<EOF
+[Service]
+Environment=LD_PRELOAD=/lib64/libnss_ldap.so
```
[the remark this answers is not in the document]
stevendanna: This was addressed here: https://github.com/chef/automate/pull/4115/commits/54a7fdee9375f7eb8c4e607a7e735aeaf5853db6

integration/tests/ldap_hab_user.sh
jaym: i put it in the other pr

Issue #4107, opened 2020-07-21
UI: Table Formatting breaking when data exceeds 3 characters


jonong1972:
## Describe the bug
https://a2-perf-test-single-local-inplace-upgrade-acceptance.cd.chef.co/compliance/reports/profiles/ac3735092676fe76843e312cbc917ccad958ead6bbf227a109c6e848d8533fc1

When viewing test results on the Profile Report page in Compliance, if the number is 5 or 6 characters long, the formatting will show the number vertically.

## To Reproduce
- Go to an Automate instance with more than 10,000 nodes being scanned.
- go to the Compliance page > Profile Tab
- Click on any profile
- On the Profile Report detail page view the test results for the Controls

## Expected behavior
The number should not display vertically

## Screenshots
<img width="1550" alt="Screen Shot 2020-07-21 at 11 21 07 AM" src="https://user-images.githubusercontent.com/1316587/88091758-5c00ea80-cb44-11ea-828a-829cfe4c9c4d.png">

## Versions (please complete the following information):
- OS: MacOS
- Browser Chrome
- Automate Build Number 20200714225434

PR #4113, opened 2020-07-22
Accomodate more space for numbers

SEAjamieD:
Numbers were stacking vertically due to insufficient space in the compliance reports profile table, this branch addresses this problem by accommodating for more space.

Fixes: https://github.com/chef/automate/issues/4107

### :+1: Definition of Done
Large numbers are accomodated for in compliance profile table

make serve

navigate to: https://a2-dev.test/compliance/reports/profiles
note: if you do not have any compliance profiles - run `load_compliance_reports` in hab studio.
click on an individual profile
you may need to manually plug in large numbers in order to see that the numbers fit.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

numbers

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-profile/reporting-profile.component.html
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-profile/reporting-profile.component.html
@@ -48,7 +48,7 @@
           <chef-thead>
             <chef-tr>
               <chef-th>{{ controls.length }} Controls</chef-th>
-              <chef-th>Impact</chef-th>
+              <chef-th class="impact-status">Impact</chef-th>
               <chef-th>Test Results</chef-th>
               <chef-th></chef-th>
             </chef-tr>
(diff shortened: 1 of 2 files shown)


SEAjamieD:
@jonong1972 - I attempted to address what you discovered in this, but it should be noted that this will still cause the numbers to break/collapse at a certain browser width, as there just isn't enough space.  

There are a few ways I can think of to address this, but it would be appreciated if I could get some input from you/UX on what you'd like to see happen.  Since there are 4 numbers in total, possibly stacking the numbers 2 on top, 2 on bottom as smaller screens?

jonong1972:
@SEAjamieD thanks. That's a lot better than what it was. As for the numbers still breaking collapsing when the view port is reduced, that may be a bigger problem - the media queries and a better responsive design to accommodate mobile/tablets/etc.

Ideally for the number I probably wouldn't want to wrap. But that is me. I want to talk to the UX team to get consensus on behaviors like that. 

But for now, I think this is better. We may need to revisit in long run.

SEAjamieD:
@jonong1972 - I agree.  To clarify, I mean that there are 4 _results_, and maybe we could stack them, 2 x 2, not 4 numbers.

Anyways - Yeah with some input we can reshape that better for small screens and I'll plan to merge this as a stopgap.  Thanks!

jonong1972:
@SEAjamieD oh yeah, stacking would be sweet. :D

PR #4115, opened 2020-07-23
Be explicit about dependencies on coreutils

### LINKED ISSUE #4105: Automate deploy fails when the linux hab user comes from external auth instead of local accounts

jaym:
Be explicit about dependencies on coreutils

We want to depend on coreutils as it will be able correctly look up UIDs
and GIDs when it comes to LDAP.

Don't use busybox

Follow up to #4110
Fixes #4105

### DIFF
--- a/components/automate-elasticsearch/habitat/plan.sh
+++ b/components/automate-elasticsearch/habitat/plan.sh
@@ -15,8 +15,7 @@ pkg_build_deps=(
   core/patchelf
 )
 pkg_deps=(
-  core/coreutils-static
-  core/busybox-static
+  core/coreutils
   core/glibc
   core/zlib
 
--- a/components/automate-builder-api-proxy/habitat/plan.sh
+++ b/components/automate-builder-api-proxy/habitat/plan.sh
@@ -11,6 +11,7 @@ pkg_license=("Chef-MLSA")
 pkg_upstream_url="https://www.chef.io/automate"
 pkg_svc_user="root"
 pkg_deps=(
+  core/coreutils
   core/bash
   chef/mlsa
   "${local_platform_tools_origin:-chef}/automate-platform-tools"
--- a/components/automate-cs-nginx/habitat/plan.sh
+++ b/components/automate-cs-nginx/habitat/plan.sh
@@ -9,6 +9,7 @@ pkg_license=('Chef-MLSA')
 # WARNING: Version managed by .expeditor/update_chef_server.sh
 pkg_version="13.0.47"
 pkg_deps=(
+  core/coreutils
   chef/mlsa
   # TODO 2020-05-12: PIN PIN PIN
   #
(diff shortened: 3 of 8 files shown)


components/automate-elasticsearch/habitat/plan.sh
```diff
@@ -15,8 +15,7 @@ pkg_build_deps=(
   core/patchelf
 )
 pkg_deps=(
-  core/coreutils-static
-  core/busybox-static
```
stevendanna, line 19: Do we know why busybox was here at all? Just copy/pasta?

jaym: i think i copy-pasta'd it last week

lib/user/lookup.go
```diff
@@ -0,0 +1,237 @@
...
+			}
+		}
+	}
+
+	return getentLookupProvider{
+		getentPath: getentPath,
+	}
+}
+
+func (g getentLookupProvider) Lookup(username string) (*User, error) {
```
stevendanna, line 151: not required at all for this PR, but from here down in this file there are a number of opportunities to pull out some well named constants and potentially extract some common code.  But I think it is fine as-is as well.

jaym: that was the hope, but the error type got in the way

PR #4123, opened 2020-07-24
Prepare to lint the protos with buf

danielsdeleo:

`buf`, https://buf.build/ is, among other things, a linter for protobufs. We want to adopt a linter for protobufs in A2 to keep things aligned with the upcoming cloud stuff in terms of API shapes, etc. 

`buf` is also pretty opinionated which I guess you might expect from a linter and it doesn't seem to like the way we do things at the moment. In particular, `buf` is pretty strict on how you specify your import paths. `buf` wants you to have this:
* For any import path, there is only one file. You cannot have multiple files with the same (relative path) name and use precedence rules to pick a winner.
* All of your Proto files have one import path. You cannot have e.g. a file `foo/bar/baz.proto` and then set your import paths to `foo/, foo/bar` such that the file could be found by both `bar/baz.proto` AND `bar.proto`.

The way that `buf` enforces these things is by placing constraints on the `roots` configuration within the build config, which is documented here: https://buf.build/docs/build-configuration#root-requirements In particular, you cannot specify your import paths such that one is a subdir of another one. 

One approach to dealing with these requirements is to make all of your imports relative to your repo root. This option would be nice, but it doesn't work if you have vendored stuff. For example, we import `google/api/annotations.proto` all over the place. It would be less disruptive if we could just change these imports to something like `import "vendor/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis/google/api/annotations.proto"` This doesn't work though, because that file needs to import `google/protobuf/descriptor.proto` at that path and not `vendor/github.com/golang/protobuf/protoc-gen-go/descriptor/descriptor.proto` So the only workable option for us is to pick a few roots in our repo and make our import paths relative to those.

* install `buf`. I will make a standardized way to install it in the future, but for now you have to do it manually. Instructions are here: https://buf.build/docs/installation -- note that if you try to use `go get` installation within the A2 repo, it will fail because we have a too old version of some dependency.
* Run `buf check lint` -- there are a LOT of failures. Which I will deal with in the future. But for now, the improvement is that `buf` actually runs the checks instead of failing with an error like `api/external/applications/applications.proto:11:8:google/api/annotations.proto: does not exist`


scripts/changed_components.rb
```diff
@@ -87,6 +89,11 @@ def get_hab_scaffolding(source_dir)
   end
 end
 
+FORCE_SKIP_PACKAGES.each do |skip_pkg|
+  changed_packages.delete(skip_pkg)
+  changed_packages_reasons.delete(skip_pkg)
+end
```
danielsdeleo, line 95: This change is excluding workflow from changed_components. This is good because [workflow is EOL December 31, 2020](https://blog.chef.io/new-chef-automate-functionality-and-end-of-life-impacts/) and there is some SCSS import issue that blocks it from building now. We _probably_ won't need to build it again. But if we do then this will probably confuse someone. I'm open to a better way to do it.

Issue #3261, opened 2020-04-01
client-runs node detail Is it possible to make run_id as optional?


vsingh-msys:
While exploring the possibility of making `run_id` as optional for client-runs node detail page found a couple of improvement area

- Manually navigating to `https://a2-dev.test/infrastructure/client-runs/NODE_ID/missing-runs` it is saying node has no client run data hold on but it has.

![client-runs-missing-runs](https://user-images.githubusercontent.com/44496937/78150303-bd496480-7454-11ea-801c-7e33ab9f0514.png)

- `Client-runs` breadcrumbs should navigate to the client-runs page but it seems it gets redirected to root page.

![client-runs-missing-runs](https://user-images.githubusercontent.com/44496937/78151824-a0ae2c00-7456-11ea-8ad6-60f76f42154e.png)

#### A possible improvement would be:

- Navigate to client node detail page only with `node-id` by making `run-id` as optional
ie. `https://a2-dev.test/infrastructure/client-runs/NODE_ID` redirect it to:

   a) if the node has run history fetch the `last_run` detail as it's stored in `nodemanager`'s nodes table.

   b) if node does not have run history redirect to missing-runs page.

Note: This is what it needed for mapping automate infra views node to client-runs node.

lancewf:
We could change the https://github.com/chef/automate/blob/9af5125bbf2ecfcc1841b296ef37424575f24539/components/automate-ui/src/app/services/node-details/node-details-resolver.service.ts#L9 `NodeDetailsResolverService` to find the latest run if one is not passed to it. This line https://github.com/chef/automate/blob/e0df48ceb8776085b432e2eeeb11351ad4f8d714/components/automate-ui/src/app/app-routing.module.ts#L229 would also have to be updated to have the run_id optional. 

The user is directed to the https://a2-dev.test/infrastructure/client-runs/NODE_ID/missing-runs page when there are no run data for the node from the NodeDetailsResolverService.

vsingh-msys:
@lancewf thanks for the insights will look and revert you back accordingly!

PR #4125, opened 2020-07-24
Client runs node detail with optional run-id

sachin-msys:
Navigate to client node detail page only with node-id by making run-id as optional, added route for node-id and its resolver service, now in route resolver service we have find out `run-id` of specific `node-id` if run-id exists then redirect to node details route if not then redirect to missing run page.

Fixes: #3261 
### :+1: Definition of Done
1. Route created  for `node-id` with optional `run-id`
2. Breadcrumb redirection issue fixed in missing run page
make serve
Navigate to Infrastructure >>> Client Runs >>> now add existing node-id in route ( a2-dev.test/infrastructure/client-runs/`node-id` )
expectations if `run-id` exists for node then it will redirect to node-details else it will redirect to missing-runs route.

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/services/node-details/node-norunid-resolver.service.spec.ts
+++ b/components/automate-ui/src/app/services/node-details/node-norunid-resolver.service.spec.ts
@@ -1,12 +1,12 @@
+import { Router, ActivatedRouteSnapshot } from '@angular/router';
 import { TestBed } from '@angular/core/testing';
 
+import { NodeRun } from 'app/types/types';
 import { NodeNoRunIdResolverService } from './node-norunid-resolver.service';
 import { NodeRunsService } from './node-runs.service';
-import { Router } from '@angular/router';
-import { NodeRun } from '../../types/types';
 
 class MockRouter {
-  constructor() {}
+  constructor() { }
   navigate(_route: Array<string>) {
   }
 }
@@ -14,7 +14,7 @@ class MockRouter {
 class MockNodeRunsService {
   response: Promise<NodeRun> = Promise.resolve<NodeRun>(null);
 
-  constructor() {}
+  constructor() { }
 
   public getNodeRunsByID(_nodeId: string) {
     return this.response;
@@ -27,17 +27,16 @@ class MockNodeRunsService {
 
 class MockRoute {
   paramMap = {
-    get() {}
+    get() { }
   };
 }
 
 describe('NodeDetailsResolverService', () => {
-  let route;
+  let route: ActivatedRouteSnapshot;
   let service: NodeNoRunIdResolverService;
   const router = new MockRouter();
-  let nodeRunsService: MockNodeRunsService;
 
-  nodeRunsService = new MockNodeRunsService();
+  const nodeRunsService = new MockNodeRunsService();
 
   beforeEach(() => {
     TestBed.configureTestingModule({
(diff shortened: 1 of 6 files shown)


components/automate-ui/src/app/services/node-details/node-norunid-resolver.service.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+    service = TestBed.inject(NodeNoRunIdResolverService);
+    route = new MockRoute();
+  });
+
+  describe('null nodeRun is returned', () => {
+    beforeEach(() => {
+      nodeRunsService.setResponse(Promise.resolve<NodeRun>(null));
+    });
+
+    it('should redirect to missing runs page', (done) => {
```
msorens, line 59: We do not really need a `describe` block containing a single `it` here. For one thing, it breaks up the test name in two pieces so it is less clear what is going on. Also, it requires more code. Consider this more direct form:
```suggestion
  it('redirects to missing runs page when null nodeRun is returned', () => {
    nodeRunsService.setResponse(Promise.resolve<NodeRun>(null));
```
Same comment applies to the remaining tests in this file--just a single `it` for each would suffice with something like these test names:
* redirects to client runs page when non-404 error is returned
* redirects to missing runs page when 404 error is returned

components/automate-ui/src/app/services/node-details/node-norunid-resolver.service.ts
```diff
@@ -0,0 +1,42 @@
+import { Injectable } from '@angular/core';
+import { Router, Resolve, RouterStateSnapshot,
+         ActivatedRouteSnapshot } from '@angular/router';
+
+import { NodeRun } from '../../types/types';
```
msorens, line 5:
```suggestion
import { NodeRun } from 'app/types/types';
```

components/automate-ui/src/app/services/node-details/node-runs.service.ts
```diff
@@ -59,6 +59,21 @@ export class NodeRunsService {
       });
   }
 
+  getNodeRunsByID(nodeId: string) {
```
msorens, line 62: Please use strong types on this function and all the others in this file:
```suggestion
  getNodeRunsByID(nodeId: string): Promise<AbridgedNodeRun[]> {
```
Besides the obvious benefits right here, it also gives types in the consumer NodeNoRunIdResolverService so one can understand the data there much better.

components/automate-ui/src/app/services/node-details/node-norunid-resolver.service.ts
```diff
@@ -0,0 +1,42 @@
...
+  constructor(private service: NodeRunsService, private router: Router) {}
+
+  resolve(route: ActivatedRouteSnapshot, _state: RouterStateSnapshot): Promise<NodeRun> {
+    const nodeId = route.paramMap.get('node-id');
+
+    return this.service.getNodeRunsByID( nodeId ).
+      then(nodeRuns => {
+        // if the node runs is found we select the latest run-id.
+        // Go to node runs detail page
+        if (nodeRuns && Object.keys(nodeRuns[0]).length !== 0) {
```
msorens, line 19: Once I annotated `getNodeRunsByID` with the correct type, I uncovered some issues here. I think you are trying to check if the first node run is not a null object.  Should first be checking that a first node run even exists--that's the `nodeRuns.length` check here. Then, we could check if `nodeRuns[0] !== null` but I think it is more clear to just directly check for what we care about, the `runId`. So we end up with this:
```suggestion
        if (nodeRuns && nodeRuns.length && nodeRuns[0].runId) {
```
But can we really get a nodeRuns array returned with null objects in it? I have no idea on that point, but if not it could be simplified to:
```
        if (nodeRuns && nodeRuns.length) {
```

PR #4129, opened 2020-07-24
ci: move upload-files from an expeditor bash action to a buildkite pipeline

stevendanna:
Our upload-files.sh script is suddenly 6+ times slower when run via
expeditor bash actions then it was previously. We think this is
because of slower disk speeds in the new expeditor setup.

Moving this script to buildkite gets around this limitation and comes
with a few other benefits:

- streaming logs during the script execution
- ability to easily upload debugging data as buildkite artifcats
- ability to easily parallelize the process if it becomes slower

We now trigger the post-promote pipeline after the
promote_habitat_packages pipeline. The fastly cache purge, deploy/dev,
and deploy/automate-chef-io pipelines then trigger on the succefful
competion of the post-promote pipeline.

I've moved the code coverage step into this pipeline just to cut down
on the number of pipelines we need to look at in buildkite. Also, it
didn't appear to me that the code-coverage job needed to be in the
same concurrency group as it shouldn't affect the shared deploy
infrastructure at all so I moved it to its own concurrency group.


.expeditor/config.yml
```diff
@@ -151,14 +151,16 @@ subscriptions:
   - workload: buildkite_hab_build_group_published:{{agent_id}}:*
     actions:
       - built_in:promote_habitat_packages
-      - bash:.expeditor/upload-files.sh:
+      - trigger_pipeline:post-promote:
+          post_commit: true
+  - workload: buildkite_build_passed:{{agent_id}}:post-promote:*
+    actions:
+      - purge_packages_chef_io_fastly:dev/automate/latest:
```
stevendanna, line 158: I think we want to purge the cache before deploying to dev to make sure that dev picks up the latest manifest.

.expeditor/config.yml
```diff
@@ -31,6 +31,11 @@ pipelines:
   - nightly:
       description: Nightly tests against master
       definition: .expeditor/nightly.pipeline.yml
+  - post-promote:
+      description: Generate and Upload of release artifacts
```
stevendanna, line 35:
```suggestion
      description: Generate and Upload release artifacts
```

PR #4135, opened 2020-07-27
ci: remove pin of docker images

stevendanna:
These pins were added a while ago after seeing a number of pipeline
breakages from docker image updates. We haven't seen such a breakage
in some time but what we have seen is docker image update PRs sit for
a considerable amount of time.

Let's remove the pin. If we see instability in the future, we can
address it then.


susanev:
@stevendanna it looks like dco is stuck on this

Issue #2887, opened 2020-02-15
step through setting up saml + azure ad and add to automate docs


susanev:
## Overview

step through setting up saml + azure ad and add to automate docs

[how to setup saml with azure ad](https://github.com/chef-cft/chef-examples/blob/master/examples/A2SamlWithAzureAD.md). Let's try it out and confirm that it works and then add it to the [supported list](https://automate.chef.io/docs/saml/#supported-identity-management-systems) along with any information needed for others to set it up.

@mjingle needs to review and edit before merge.

jmassardo:
You'll need to add `name_id_policy_format = "urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress"` to the saml config for IAMv2 to work properly.

srenatus:
Please note my signing cert concerns in the other github issue.

susanev:
we cannot document this in the automate docs, because even though it will work at the start it will break as soon as azure rotates the keys.

[internal slack thread](https://chefio.slack.com/archives/C61F9HHKK/p1594218269274200)

susanev:
[chef support documentation](https://getchef.zendesk.com/hc/en-us/articles/360045977191)

susanev:
its never ending fun

srenatus:
the magic words are _non-gallery app_ 💡

PR #4136, opened 2020-07-27
docs/content/saml: add Azure AD section

srenatus:

Some of our customers have been using this for a while now. I've verified that the instructions work, and added it, together with an emphasis on manual signing certificate management, to the docs.

#2887 

### :+1: Definition of Done

There's docs describing how to use Azure AD's SAML with A2.

- `cd docs; make serve` and `open http://localhost:1313/docs/saml/` given that you've gotten all the docs tools installed.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-chef-io/content/docs/saml.md
+++ b/components/automate-chef-io/content/docs/saml.md
@@ -49,23 +49,26 @@ Users who sign in via SAML will have a session time of 24 hours before needing t
 - Ping
 - Tivoli Federated Identity Manager
 
-In the following section, you can find further details about specific supported identity management systems.
-
 ### Azure AD
 
-For using Chef Automate's SAML integration with Azure AD, please note that **signing key rotation** is
-not supported to be done automatically. The signing certificate used with Azure for SAML sign-ins
-has to be **managed manually**.
+Using Azure AD as an SAML IdP requires specific configuration for both Azure AD and Chef Automate.
 
-After setting up your application as a _"non-gallery application"_, configure its SAML sign-on method.
-For both _Identifier (Entity ID)_ and _Reply URL (Assertion Consumer Service URL)_, enter
-`https://{{< example_fqdn "automate" >}}/dex/callback`.
+{{< info >}}
+The signing certificate used for Chef Automate's SAML integration with Azure AD requires manual management.
+Signing key rotation is not done automatically.
+{{< /info >}}
 
-The default claims can be used, see below for their counterparts in the Chef Automate configuration.
-Download the _Certificate (Base64)_, and take note of the _Login URL_.
+First, set up your configuration within Azure AD.
+Set up your application as a _"non-gallery application"_, and then configure its SAML log-in method.
+Enter `https://{{< example_fqdn "automate" >}}/dex/callback` as the value for both _Identifier (Entity ID)_ and _Reply URL (Assertion Consumer Service URL)_,.
 
-The minimal configuration snippet then looks like this (replace `<<something>>` with the values
-noted above):
+You may use the default claims provided by Azure AD.
+Remember to edit the Chef Automate configuration in `config.toml` to reflect this claims information.
+
+Download the _Certificate (Base64)_ and take note of the _Login URL_ for use in the Chef Automate configuration.
+
+After configuring Azure AD, edit your Chef Automate `config.toml` configuration file to reflect the values entered in the Azure AD interface.
+The minimal configuration snippet in `config.toml` will look similar to:
 
 ```toml
 [dex.v1.sys.connectors.saml]
@@ -78,7 +81,13 @@ username_attr = "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddr
 entity_issuer = "https://{{< example_fqdn "automate" >}}/dex/callback"
 ```
 
-See below for further options.
+where
+
+- `ca_contents` contains the value of the _Certificate (Base64)_
+- `sso_url` contains the value of _Login URL_
+- `entity_issuer`  contains the value of _Identifier (Entity ID)_
+
+See the SAML Configuration Settings section below for further configuration options.
 
 ## SAML Configuration Settings
 
--- a/components/automate-chef-io/content/docs/saml.md
+++ b/components/automate-chef-io/content/docs/saml.md
@@ -58,7 +58,7 @@ The signing certificate used for Chef Automate's SAML integration with Azure AD 
 Signing key rotation is not done automatically.
 {{< /info >}}
 
-In Azure AD, add Chef Automate as a _"non-gallery application"_, and then configure its SAML log-in method.
+In Azure AD, add Chef Automate as a _"non-gallery application"_, and then configure its SAML sign-in method.
 [The Azure AD documentation](https://docs.microsoft.com/en-us/azure/active-directory/manage-apps/configure-saml-single-sign-on) provides a detailed guide.
 Enter `https://{{< example_fqdn "automate" >}}/dex/callback` as the value for both _Identifier (Entity ID)_ and _Reply URL (Assertion Consumer Service URL)_.
 
--- a/components/automate-chef-io/content/docs/saml.md
+++ b/components/automate-chef-io/content/docs/saml.md
@@ -59,7 +59,7 @@ Signing key rotation is not done automatically.
 {{< /info >}}
 
 First you need to add Chef Automate as a _"non-gallery application"_, and then configure its SAML log-in method.
-[The Azure AD documentation](https://docs.microsoft.com/en-us/azure/active-directory/manage-apps/configure-single-sign-on-non-gallery-applications)
+[The Azure AD documentation](https://docs.microsoft.com/en-us/azure/active-directory/manage-apps/configure-saml-single-sign-on)
 provides a detailed guide.
 Enter `https://{{< example_fqdn "automate" >}}/dex/callback` as the value for both _Identifier (Entity ID)_ and _Reply URL (Assertion Consumer Service URL)_,.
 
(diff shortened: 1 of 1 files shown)


components/automate-chef-io/content/docs/saml.md
```diff
@@ -42,12 +42,44 @@ Users who sign in via SAML will have a session time of 24 hours before needing t
...
 - Ping
 - Tivoli Federated Identity Manager
 
+In the following section, you can find further details about specific supported identity management systems.
+
+### Azure AD
+
+For using Chef Automate's SAML integration with Azure AD, please note that **signing key rotation** is
+not supported to be done automatically. The signing certificate used with Azure for SAML sign-ins
+has to be **managed manually**.
```
stevendanna, line 58: Maybe something like: 

```suggestion
When using Chef Automate's SAML integration with Azure AD, please note that Chef Automate does not support automatic **signing key rotation**. The signing certificate used with Azure for SAML sign-ins has to be **managed manually**.
```

components/automate-chef-io/content/docs/saml.md
```diff
@@ -42,12 +42,44 @@ Users who sign in via SAML will have a session time of 24 hours before needing t
...
+
+After setting up your application as a _"non-gallery application"_, configure its SAML sign-on method.
+For both _Identifier (Entity ID)_ and _Reply URL (Assertion Consumer Service URL)_, enter
+`https://{{< example_fqdn "automate" >}}/dex/callback`.
```
stevendanna, line 62: Do we want to guide them to _where_ they should be "setting up your application" or is this obvious to someone using Azure AD?

components/automate-chef-io/content/docs/saml.md
srenatus: Taken, thanks!

components/automate-chef-io/content/docs/saml.md
srenatus: I'm not sure. What do you think? 👇 

Screenshot 2020-07-27 at 15 30 39

Here, we select "View" below the middle tile, and get to

Screenshot 2020-07-27 at 15 32 09

...where clicking on "Enterprise application" will get us to

Screenshot 2020-07-27 at 15 32 15

where we find the "non-gallery application".

I'm sure there's other ways. I think I'd like to avoid adding too much of that here, since then our docs are tied to a UI in Azure; which may change at some point.

components/automate-chef-io/content/docs/saml.md
```diff
@@ -42,12 +42,42 @@ Users who sign in via SAML will have a session time of 24 hours before needing t
...
+
+The minimal configuration snippet then looks like this (replace `<<something>>` with the values
+noted above):
+
+```toml
+[dex.v1.sys.connectors.saml]
+ca_contents="""
+<<Certificate (Base64)>>
+"""
+sso_url = "<< Login URL >>"
```
msorens, line 73: Should be consistent: either spaces just inside the angle brackets (L73) or not (L71).

components/automate-chef-io/content/docs/saml.md
mjingle: Would this piece of Azure documentation be relevant and a possible link to include? https://docs.microsoft.com/en-us/azure/active-directory/manage-apps/add-application-portal#add-an-app-to-your-azure-ad-tenant

components/automate-chef-io/content/docs/saml.md
```diff
@@ -42,12 +42,44 @@ Users who sign in via SAML will have a session time of 24 hours before needing t
...
 
+In the following section, you can find further details about specific supported identity management systems.
+
+### Azure AD
+
+For using Chef Automate's SAML integration with Azure AD, please note that **signing key rotation** is
+not supported to be done automatically. The signing certificate used with Azure for SAML sign-ins
+has to be **managed manually**.
+
+After setting up your application as a _"non-gallery application"_, configure its SAML sign-on method.
```
mjingle, line 60: Are the fields in _italics_ supposed to be the relevant fields for the Azure AD set-up (and not directly for the Chef Automate interface)?

components/automate-chef-io/content/docs/saml.md
```diff
@@ -42,12 +42,44 @@ Users who sign in via SAML will have a session time of 24 hours before needing t
...
+For both _Identifier (Entity ID)_ and _Reply URL (Assertion Consumer Service URL)_, enter
+`https://{{< example_fqdn "automate" >}}/dex/callback`.
+
+The default claims can be used, see below for their counterparts in the Chef Automate configuration.
+Download the _Certificate (Base64)_, and take note of the _Login URL_.
+
+The minimal configuration snippet then looks like this (replace `<<something>>` with the values
```
mjingle, line 67: Is this the minimal configuration snippet for Chef Automate? If so, what file does this example refer to?

components/automate-chef-io/content/docs/saml.md
srenatus: Yeah! Is there another markup you'd like to have instead?

components/automate-chef-io/content/docs/saml.md
srenatus: It's the minimal SAML-related config to make this work, when patching it into the running settings.

components/automate-chef-io/content/docs/saml.md
srenatus: Hmm I don't know about that specific link, it seems to be a little misleading: it's describing the github integration, which is easier to set up.

Do you prefer having some official docs to link to? I can try digging up a more fitting link.

mjingle: I know we won't find specific docs referring to Chef, but it would be good to link official docs that covers the general aspects of Azure's "adding an application".  That way, our Chef Automate documentation can provide the needed details rather than having to recreate the wheel with detailed documentation that may change.

srenatus: Yeah we're on the same page here. I'll go look for a generic docs page that we can reference. :)

components/automate-chef-io/content/docs/saml.md
mjingle: Italics are fine. I just wanted to make sure that I was interpreting that those italics were for fields/labels since I can't reference them for myself. Thanks!

components/automate-chef-io/content/docs/saml.md
```diff
@@ -42,12 +42,53 @@ Users who sign in via SAML will have a session time of 24 hours before needing t
...
+Using Azure AD as an SAML IdP requires specific configuration for both Azure AD and Chef Automate.
+
+{{< info >}}
+The signing certificate used for Chef Automate's SAML integration with Azure AD requires manual management.
+Signing key rotation is not done automatically.
+{{< /info >}}
+
+First, set up your configuration within Azure AD.
+Set up your application as a _"non-gallery application"_, and then configure its SAML log-in method.
+Enter `https://{{< example_fqdn "automate" >}}/dex/callback` as the value for both _Identifier (Entity ID)_ and _Reply URL (Assertion Consumer Service URL)_,.
```
mjingle, line 63: @srenatus Ideally, I would like to have a bulleted list with the Azure AD field & its expected value listed. Since I assume these on line 63 are the only two required fields and respective value, I wasn't going to change the format here.

components/automate-chef-io/content/docs/saml.md
```diff
@@ -42,12 +42,53 @@ Users who sign in via SAML will have a session time of 24 hours before needing t
...
+### Azure AD
+
+Using Azure AD as an SAML IdP requires specific configuration for both Azure AD and Chef Automate.
+
+{{< info >}}
+The signing certificate used for Chef Automate's SAML integration with Azure AD requires manual management.
+Signing key rotation is not done automatically.
+{{< /info >}}
+
+First, set up your configuration within Azure AD.
```
mjingle, line 61: @srenatus I would think we could put your official documentation link towards the beginning on either line 61 or 62? Either line may have to be reworded. I just realized both sentences begin with "set up."

components/automate-chef-io/content/docs/saml.md
```diff
@@ -49,23 +49,26 @@ Users who sign in via SAML will have a session time of 24 hours before needing t
...
+{{< info >}}
+The signing certificate used for Chef Automate's SAML integration with Azure AD requires manual management.
+Signing key rotation is not done automatically.
+{{< /info >}}
 
-After setting up your application as a _"non-gallery application"_, configure its SAML sign-on method.
-For both _Identifier (Entity ID)_ and _Reply URL (Assertion Consumer Service URL)_, enter
-`https://{{< example_fqdn "automate" >}}/dex/callback`.
+First, set up your configuration within Azure AD.
+Set up your application as a _"non-gallery application"_, and then configure its SAML log-in method.
```
srenatus, line 62: I believe we've been trying to standardize on "sign-in" instead of "log-in" -- am I on the wrong track here?

components/automate-chef-io/content/docs/saml.md
srenatus: From playing with it, these are sufficient. Besides managing the certificate.

The original CFT docs (linked in https://github.com/chef/automate/issues/2887) included renaming the claims, but I've found that to be not necessary.

components/automate-chef-io/content/docs/saml.md
srenatus: Linked something :) Not sure about the wording there, though -- please have a look.

components/automate-chef-io/content/docs/saml.md
jmassardo: The only thing I would add would to include `name_id_policy_format = "urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress"` in the saml.toml. If I remember correctly, I had to do this to get Automate to use the email address as a login instead of the user UUID from AAD.

srenatus: Hrm I've tried it and it worked without. So, I'd rather keep this as simple as can be...? 🤔

jmassardo: Cool. I set mine lab up several months ago so it's very possible some other thing got fixed between now and then. I'm good leaving it as simple as possible.

PR #4139, opened 2020-07-27
modify sugg api to accept type_key field to be explicit about control tag key we are searching on

vjeffrey:
When retrieving suggestions for control tag values, we use the control tag keys included in the filters to determine the control tag key for which we are trying to find values 
as noted by @rickmarry in https://github.com/chef/automate/pull/4134, this leads to some ambiguity about which control tag key to associate with the value search
we decided to add a new field to the api so we can be explicit.
this pr adds the new field to the api, modifies the ui code to use that field instead of adding the control tag key to the filters, and modifies the backend code to understand that new field.

~**NOTE: this change as-is is technically a breaking change because the old way of specifying the control tag key in the filters when doing a suggestions search for the values is no longer supported. we should figure out if that's ok or not and change as needed.**~ we modified things so this is no longer not-backwards-compat

https://github.com/chef/automate/pull/4134

### :+1: Definition of Done
new field in api to specify control tag key when doing sugg search for values

rebuild ui, automate-gateway, compliance
load_compliance_reports
use the searchbar to search for nodes with control tag keys/vals

- [ ] Tests added/updated?
- [ ] Docs added/updated?


vjeffrey:
@chef/ui-team i may have broken some rules about what we import where with this change. idk. this was the only way i could get things working. happy to talk about alternative solutions!

components/automate-ui/src/app/pages/+compliance/shared/reporting/suggestions.service.ts
```diff
@@ -15,10 +15,12 @@ export class SuggestionsService {
     private statsService: StatsService
   ) {}
 
+  selectedControlTagKey: string;
+
   getSuggestions(type: string, text: string, reportQuery: ReportQuery): Observable<any> {
     const url = `${CC_API_URL}/reporting/suggestions`;
     const formatted = this.statsService.formatFilters(reportQuery);
-    const body = {type, text, filters: formatted};
+    const body = {type, text, 'type_key': this.selectedControlTagKey, filters: formatted};
```
SEAjamieD, line 23:
```suggestion
        const body = {
          type,
          text,
          type_key: this.selectedControlTagKey,
          filters: formatted
        };
```

They've slipped through in some spots, but it was pointed out to me that the preferred route is not using the quotations

components/automate-ui/src/app/types/types.ts
```diff
@@ -64,6 +64,7 @@ export class ComplianceNode implements ComplianceData {
 export interface Chicklet {
   text: string;
   type: string;
+  type_key?: any;
```
SEAjamieD, line 67:
```suggestion
  type_key?: string;
```

I think all the type_keys are always strings, from what i've noticed

vjeffrey:
just pushed a commit for the backwards compat stuff

PR #4140, opened 2020-07-27
Compliance UI make use of root status_message

alexpop:
This is MVP UI change for this backend improvement: https://github.com/chef/automate/pull/4131

@jonong1972 is working into some designs to bring this together with the failure `status_message` fields when profiles inside a report fail.

To test, rebuild `components/automate-ui` and ingest a report, like the one from the tests of https://github.com/chef/automate/pull/4131 

I left the boxes with the zero control counts to further highlight that this report is empty/abnormal and needs investigation.

Screenshot of the error div here:
Screenshot 2020-07-27 at 23 05 26

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-node/reporting-node.component.html
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-node/reporting-node.component.html
@@ -169,7 +169,16 @@
           </chef-option>
         </chef-phat-radio>
 
-        <chef-table class="controls-table">
+        <div class="metadata-container err-container">
+          <div class="metadata failed err-failed-container">
+            <span class="metadata-title err-color">
+              <chef-icon id="err-icon-id">{{ statusIcon('failed') }}</chef-icon>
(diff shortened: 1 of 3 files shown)


components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-node/reporting-node.component.html
```diff
@@ -169,7 +169,16 @@ <h2 class="metadata-group-heading">Metadata</h2>
           </chef-option>
         </chef-phat-radio>
 
-        <chef-table class="controls-table">
+        <div class="metadata-container err-container">
```
lancewf, line 172: I think this should only be shown when there is an error. I am seeing this when a node does not fail.

image

alexpop: You are right @lancewf, I had the ngif to make the div only show, but it vanished in my css struggles. I pushed another commit to bring the ngif back. Tested it with a failed node and a padded one and works as expected now. Thank you

PR #4146, opened 2020-07-28
ci: use subscription to promote workload

stevendanna:
The old way has been deprecated. This is the new way.

### DIFF
--- a/.expeditor/config.yml
+++ b/.expeditor/config.yml
@@ -58,43 +58,7 @@ pipelines:
       env:
         - CHANNEL: dev
 
-# These actions are taken when `/expeditor promote` is run from Slack
 promote:
-  action:
-    - bash:.expeditor/promote-hab-pkgs-and-cli.sh:
-        post_commit: true
-    - bash:.expeditor/publish-release-notes.sh:
-        post_commit: true
-        only_if_conditions:
-          - value_one: "{{target_channel}}"
-            operator: equals
-            value_two: current
-    - purge_packages_chef_io_fastly:{{target_channel}}/automate/latest:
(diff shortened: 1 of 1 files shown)


.expeditor/config.yml
```diff
@@ -106,6 +70,33 @@ staging_areas:
...
+  - workload: project_promoted:{{agent_id}}:dev:*
+    actions:
+    - bash:.expeditor/promote-hab-pkgs-and-cli.sh:
+        post_commit: true
+    - purge_packages_chef_io_fastly:{{target_channel}}/automate/latest:
+        post_commit: true
+    - bash:.expeditor/announce-acceptance.sh:
+        post_commit: true
+    - trigger_pipeline:deploy/acceptance
+    - trigger_pipeline:deploy/automate-chef-io
```
stevendanna, line 84: I believe this was running on all promotions before.  Is this really what we want? Do we need to deploy automate.chef.io on dev->acceptance promotions?

.expeditor/config.yml
```diff
@@ -106,6 +70,33 @@ staging_areas:
       workload: pull_request_merged:chef/automate:master:*
 
 subscriptions:
+  # dev -> acceptance promotion
+  # These actions are taken when `/expeditor promote` is run from Slack
+  - workload: project_promoted:{{agent_id}}:dev:*
```
msorens, line 75: Just to confirm:
With the old actions, the channel was the end-state (e.g. only_if == "acceptance"), but now with subscriptions, the channel is the start-state...?

stevendanna: Yup

PR #4147, opened 2020-07-28
Dan/upgrade protoc

danielsdeleo:

Upgrades the protobuf compiler libraries to the latest. If this is workable then we will be able to install `buf` via `go get` which is my primary goal. But it's also nice to not be pinned to old things! That said, with this change we get a LOT of nags like so:

```
WARNING: Package "github.com/golang/protobuf/protoc-gen-go/generator" is deprecated.
	A future release of golang/protobuf will delete this package,
	which has long been excluded from the compatibility promise.
```

The references to this package are from: 

```
components/automate-grpc/protoc-gen-policy/policy/policy.go:12: "github.com/golang/protobuf/protoc-gen-go/generator"
vendor/github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway/descriptor/types.go:8:    gogen "github.com/golang/protobuf/protoc-gen-go/generator"
vendor/github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway/gengateway/template.go:11:        generator2 "github.com/golang/protobuf/protoc-gen-go/generator"
vendor/github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger/genswagger/template.go:22:     gogen "github.com/golang/protobuf/protoc-gen-go/generator"
vendor/github.com/lyft/protoc-gen-star/lang/go/name.go:8:       "github.com/golang/protobuf/protoc-gen-go/generator"
vendor/github.com/lyft/protoc-gen-star/lang/go/name.go:89:// See: https://godoc.org/github.com/golang/protobuf/protoc-gen-go/generator#CamelCase
```

So it will take a bit more work to get rid of that. Can we live with the spam for now?

This is when we pinned it in the first place: https://github.com/chef/automate/pull/3466 

### :+1: Definition of Done

`go get -u   github.com/bufbuild/buf/cmd/buf   github.com/bufbuild/buf/cmd/protoc-gen-buf-check-breaking   github.com/bufbuild/buf/cmd/protoc-gen-buf-check-lint` does not fail.


components/automate-grpc/scripts/bootstrap_protoc.sh
```diff
@@ -13,10 +13,10 @@ set -e
 
 
 printf 'GEN: %s\n' components/automate-grpc/protoc-gen-a2-config/api/a2conf/*.proto
-protoc -I /src --go_out=logtostderr=true,paths=source_relative:/src \
```
danielsdeleo, line 16: `logtostderr` went away as an option. I think it was in this commit: https://github.com/golang/protobuf/commit/aaf9c3b54ff2d3e40f85540b0c98b0d2f5fbaf34#diff-8c603013608023320d5242916c4ea03b

It's tricky to follow because they pointed the CLI at a new implementation that didn't support all the same flags instead of modifying the existing one. In any case, if you look at the godoc for the new implementation (e.g., https://godoc.org/google.golang.org/protobuf/compiler/protogen#Options.Run) all the functions say they log errors and whatnot to stderr. So tl;dr the thing we are asking for is now the default behavior and the flag to enable it is gone.

api/external/secrets/secrets_test.go
```diff
@@ -189,9 +189,9 @@ func TestValidateSecretFailures(t *testing.T) {
 	for _, test := range exampleValidationFailures {
 		err := test.in.Validate()
 		if err == nil {
-			t.Errorf("Sprintf(%q, validateSecret()) => nil, want %q", test.in, test.out)
+			t.Errorf("Sprintf(%q, validateSecret()) => nil, want %q", &test.in, test.out)
 		} else if err.Error() != test.out {
-			t.Errorf("Sprintf(%q, validateSecret()) => %q, want %q", test.in, err.Error(), test.out)
+			t.Errorf("Sprintf(%q, validateSecret()) => %q, want %q", &test.in, err.Error(), test.out)
```
danielsdeleo, line 194: the `String()` method has a pointer receiver, dunno why it worked before and not now

components/authz-service/server/rules_property_test.go
```diff
@@ -815,7 +816,7 @@ func reportRules(t *testing.T, rules []api.CreateRuleReq) {
 	valueBytes := 0
 	for _, rule := range rules {
 		condCount += len(rule.Conditions)
-		ruleBytes += rule.XXX_Size()
```
danielsdeleo, line 818: the `XXX_foo` methods are gone, though we should not have been using them anyway

components/automate-deployment/pkg/preflight/checks.go
```diff
@@ -403,7 +403,11 @@ func requiredPortsForConfig(skipShared bool, config *dc.AutomateConfig) ([]int,
 	portsToCheck = append(portsToCheck, 9631, 9638)
 	acValue := reflect.ValueOf(*config)
 	for i := 0; i < acValue.NumField(); i++ {
-		if v, ok := acValue.Field(i).Interface().(a2conf.A2ServiceConfig); ok {
+		f := acValue.Field(i)
+		if !f.CanInterface() {
```
danielsdeleo, line 407: this skips private struct fields. with the older proto code there weren't any, but now there are.

components/automate-gateway/handler/license.go
```diff
@@ -75,10 +75,13 @@ func (t *LicenseServer) GetStatus(ctx context.Context,
 	}
 
 	resp := license.GetStatusResp{
-		LicenseId:      lcResp.LicenseId,
-		ConfiguredAt:   lcResp.ConfiguredAt,
-		CustomerName:   lcResp.CustomerName,
-		LicensedPeriod: (*license.GetStatusResp_DateRange)(lcResp.LicensedPeriod),
```
danielsdeleo, line 81: these direct casts don't work any longer because you can't cast if your struct has private members (unless both are defined in the same package)

components/automate-grpc/scripts/bootstrap_protoc.sh
```diff
@@ -14,11 +14,11 @@ set -e
 pushd components
 
 printf 'GEN: %s\n' components/automate-grpc/protoc-gen-a2-config/api/a2conf/*.proto
-protoc -I /src/components --go_out=logtostderr=true,paths=source_relative:/src/components \
+protoc -I /src/components --go_out=paths=source_relative:/src/components \
   automate-grpc/protoc-gen-a2-config/api/a2conf/*.proto
 
 printf 'GEN: %s\n' components/automate-grpc/protoc-gen-policy/iam/*.proto
-protoc -I /src/components --go_out=logtostderr=true,paths=source_relative:/src/components \
```
danielsdeleo, line 21: the new golang proto compiler only logs to stderr. It does not recognize this option.

components/compliance-service/ingest/events/compliance/helper_methods.go
```diff
@@ -41,7 +40,7 @@ func ProfileControlSummary(profile *inspec_api.Profile) *reportingTypes.NodeCont
 	summary := reportingTypes.NodeControlSummary{}
 	for _, control := range profile.Controls {
 		summary.Total++
-		if control.WaiverData != nil && !reflect.DeepEqual(*control.WaiverData, inspec_api.WaiverData{}) && !strings.HasPrefix(control.WaiverData.Message, "Waiver expired") {
+		if control.WaiverData != nil && !pcmp.DeepEqual(*control.WaiverData, inspec_api.WaiverData{}) && !strings.HasPrefix(control.WaiverData.Message, "Waiver expired") {
```
danielsdeleo, line 43: see lib/pcmp for the implementation but basically it should be the same as `reflect.DeepEqual` but ignores the internal/private protobuf fields (that used to be `XXX_foo`)

lib/grpc/grpctest/uptodate.go
```diff
@@ -100,7 +100,17 @@ func AssertAllProtoMethodsAnnotated(t *testing.T, file string, vsnIdentifier str
...
+	// Matching string examples:
+	//   A simpler method.GetOptions is like:
+	//
+	// [chef.automate.api.iam.policy]:{resource:"system:config" action:"system:telemetryConfig:get"} [google.api.http]:{get:"/api/v0/telemetry/config"}
+	//
+	//   A more complicated one is like:
+	//
+	// [chef.automate.api.iam.policy]:{resource:"iam:policies:{id}" action:"iam:policies:update"} [google.api.http]:{put:"/apis/iam/v2/policies/{id}" body:"*"} [grpc.gateway.protoc_gen_swagger.options.openapiv2_operation]:{tags:"policies" extensions:{key:"x-code-samples" value:{list_value:{values:{struct_value:{fields:{key:"lang" value:{string_value:"JSON"}} fields:{key:"source" value:{string_value:"{\"name\": \"My Updated Viewer Policy\", \"members\": [\"user:ldap:newuser\", \"team:ldap:newteam\"], \"statements\": [{\"role\": \"viewer\",\"projects\":[\"project1\", \"project2\"], \"effect\": \"ALLOW\"},{\"role\": \"qa\",\"projects\": [\"acceptanceProject\"], \"effect\": \"ALLOW\"}],\"projects\": []}"}}}}}}}}
+	//
+	// NOTE that in the older version of the golang protobuf toolchain, angle brackets (`<>`) were used instead of curly braces (`{}`)
```
danielsdeleo, line 112: this is the reason for the change here.

go.mod
```diff
@@ -20,7 +20,7 @@ require (
 	github.com/briandowns/spinner v0.0.0-20180123222039-b3ed21e9d3b2
 	github.com/buger/goterm v0.0.0-20180307092342-c9def0117b24
 	github.com/buger/jsonparser v0.0.0-20180808090653-f4dd9f5a6b44
-	github.com/chef/toml v0.3.1-0.20180511201931-e53972c43816
+	github.com/chef/toml v0.3.1-0.20200730001027-920c30b33b5d
```
danielsdeleo, line 23: see https://github.com/chef/toml/pull/4

PR #4149, opened 2020-07-28
Removes option to add teams for new project

blakestier:

Because teams is a separate service that is dependent on auth,
tying these created teams to policies in the most readable,
maintainable way requires significant effort (using cereal or merging teams into authz), so we are
removing the option for now.

Enter the studio and bring up the UI-- Navigate to "Settings" in the UI, then "Projects". Click "Create Project":

-- there should not be an option to automatically create teams
-- when project is created, there should be no associated teams

### DIFF
--- a/components/automate-ui/src/app/pages/project/list/project-list.component.ts
+++ b/components/automate-ui/src/app/pages/project/list/project-list.component.ts
@@ -18,7 +18,6 @@ import {
   allProjects, getAllStatus, createStatus, createError
 } from 'app/entities/projects/project.selectors';
 import { GetProjects, CreateProject, DeleteProject, ProjectPayload  } from 'app/entities/projects/project.actions';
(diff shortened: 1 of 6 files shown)


components/automate-ui/src/app/pages/project/list/project-list.component.ts
```diff
@@ -45,12 +44,6 @@ export class ProjectListComponent implements OnInit, OnDestroy {
...
 
-  readonly teamsToCreate = {
-    'project-owners': 'Project Owners',
-    'editors': 'Editors',
-    'viewers': 'Viewers'
-  };
-
   constructor(
     private layoutFacade: LayoutFacadeService,
     private store: Store<NgrxStateAtom>,
```
msorens, line 49: Just below here you need to delete this line:
```
addTeams: [true]
```

PR #4161, opened 2020-07-30
Add bookshelf keypair and bifrost superuser id to bootstrap bundle

jaym:
- Chef server hands and presigned URLs in the sandbox API, and thus all the automate frontends need to agree on the access key/secret key pair for bookshelf.
- The ha chef server tests now load balance between the multiple frontends instead of running the tests on each frontend
- Pulled the bifrost superuser id into the bootstrap bundle. I think the tests require the frontends agree on this, although nothing should be broken without it outside the tests.


stevendanna:
> think the tests require the frontends agree on this, although nothing should be broken without it outside the tests.

Thanks for this clarification, was a bit confused why this would matter

PR #4170, opened 2020-08-03
Switch inspec config passing to new cli option

alexpop:
Changing how CLI config options are passed based on recent InSpec change.

https://github.com/inspec/inspec/issues/5172#issuecomment-667981882

This change is for Automate triggered scan jobs.


stevendanna:
Is this fixing something that is currently broken on the main branch? If so, is there a test we need to add? If not, should this be paired with an inspec version bump?

Schwad:
@stevendanna 

For background: `--json-config` has been deprecated since InSpec 3.6 for `--config`.

Alex uncovered a bug when trying to pipe json-style config to `--json-config` where it didn't catch the truncation option. 

This bug does not happen with the supported `--config` option. As I understand it this PR updates that.

Issue #3983, opened 2020-06-25
control menu not growing in width


susanev:
## overview
the control menus should grow in width so all labels fit, right now labels are being cut off with an ellipsis
![image](https://user-images.githubusercontent.com/5489125/85767336-c3a25200-b6cc-11ea-8381-96985c90acce.png)

SEAjamieD:
@susanev Any Idea if there is a specific page or browser this is coming from?  I'm not able to reproduce this yet. 

I've checked settings>projects, but i've also checked several of the other control menus.

msorens:
@SEAjamieD 
Here is a reproducible scenario.
Visit https://a2-local-inplace-upgrade-acceptance.cd.chef.co/settings/projects and open the control menu on any row.
 - Problem occurs only on the Projects page.
 - Problem does **not** occur on any other resource page under Settings.
 - Problem occurs in Chrome.
 - Problem does **not** occur in Firefox.
 - Problem occurs equally whether browser is full screen or resized to be narrow horizontally.

And here is the kicker:
 - Problem occurs when logged in as a SAML user.
 - Problem does **not** occur when logged in as a local user.

msorens:
Hmm... seems to be more convoluted than that. I went through various choices again, and this time when logged in as a SAML user the issue was not present! 🤔

msorens:
Now I am seeing it logged in as a local user (admin, Chrome) or Settings > Policies > <policy details>
<img width="1092" alt="image" src="https://user-images.githubusercontent.com/6817500/87726983-ea582380-c774-11ea-93b4-f1310b34d5b9.png">
And in the same session, now again on the original Settings > Projects page:
<img width="1088" alt="image" src="https://user-images.githubusercontent.com/6817500/87727036-0c51a600-c775-11ea-9044-f9453807752f.png">

PR #4178, opened 2020-08-04
Prevent ellipsis-ing of control menu text

SEAjamieD:
The control menu text in project files was getting ellipsed seemingly randomly.  @susanev hunted down the cause, and it turned out a css selector was leaking from the api-tokens page.

The fix in this branch encapsulates those selectors so they no longer leak.

Fixes: https://github.com/chef/automate/issues/3983

### :+1: Definition of Done
Text inside control menus is not longer being cut off and ellipsed.

make serve

on a clean branch to repro the issue
- navigate to api tokens
- create a token
- open the token control menu
- navigate to projects
- create a project
- open that menu and it will have the `...`
hard refresh, open projects again notice no `...`
then repeat above and it should show the `...` again

on this branch
Run through the repro sets and you should no longer see this happen.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/modules/token/token-list/api-token-list.component.scss
+++ b/components/automate-ui/src/app/modules/token/token-list/api-token-list.component.scss
@@ -8,14 +8,14 @@
     right: 24px;
     width: 54px;
   }
-  
+
   ::ng-deep .mat-select-panel-wrap {
     margin-left: -75px;
   }
 
   .empty-state {
     text-align: center;
-  
+
     p {
       font-size: 18px;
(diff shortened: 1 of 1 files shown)


msorens:
I just have to point out that it is not an ellipse, but an ellipsis, so the correct title of this PR should include "ellipsis-ing". 😁

SEAjamieD:
> I just have to point out that it is not an ellipse, but an ellipsis, so the correct title of this PR should include "ellipsis-ing". 😁

😂  - touché

PR #4182, opened 2020-08-05
Lock pg migration dirty migration fixer

danielsdeleo:

I made an alternative to this PR in #4183

A customer with 4 FE systems found that migrations that are not written defensively will fail. After investigating the relevant code, I found that:
* The migration library we use imposes locks on writes but not reads. In particular, the `Up()` and `Force()` methods we use are guarded by a lock, but the `Version` function we use to check for dirty state is not.
* The migration method sets the version to the desired version and the dirty flag to true, then runs the migration logic, then sets dirty to false.
* Our code calls `Version` to get the value of the `dirty` flag, and if it's set to true, then we use `Force` to set the migration version back by one (and reset the dirty flag to false).

Thus, if two or more copies of the migrator are running concurrently, you can have:
* The first one sees a clean database, skips the cleanup step, starts the migration, sets dirty to true
* the second one sees a dirty database when it calls version
* the first one finishes migrating
* the second one force sets the migration level down one and re-runs the migration that had just succeeded
* the second one will fail if it's not defensively coded.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/lib/db/migrator/migrator.go
+++ b/lib/db/migrator/migrator.go
@@ -1,9 +1,13 @@
 package migrator
 
 import (
+	"crypto/sha256"
+	"fmt"
 	"net/url"
+	"strconv"
 
 	"github.com/golang-migrate/migrate"
+	"github.com/golang-migrate/migrate/database"
 	_ "github.com/golang-migrate/migrate/database/postgres" // make driver available
 	_ "github.com/golang-migrate/migrate/source/file"       // make source available
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


lib/db/migrator/migrator.go
```diff
@@ -33,26 +37,42 @@ func MigrateWithMigrationsTable(pgURL, migrationsPath, migrationsTable string, l
...
 	}
+
 	m, err := migrate.New(addScheme(migrationsPath), purl)
 	if err != nil {
 		return errors.Wrap(err, "init migrator")
 	}
 	m.Log = migrationLog{Logger: l, verbose: verbose}
+
+	// Plan 1:
+	// get a database driver from the migration package and wrap the version checky stuff in a lock
```
danielsdeleo, line 48: Need to fixup these comments

lib/db/migrator/migrator.go
```diff
@@ -33,26 +37,42 @@ func MigrateWithMigrationsTable(pgURL, migrationsPath, migrationsTable string, l
...
+	databaseDrv, err := database.Open(purl)
```
danielsdeleo, line 49: Here we're getting a second database driver, different from the one owned by the `m` struct. We could instead create the `m` migrator with the `migrate.NewWithDatabaseInstance` function https://github.com/golang-migrate/migrate/blob/31c22870d306d0facce2b70c291b458c007b327e/migrate.go#L121 But that would not work with the code as written, as the driver won't let you Lock if you are already locked. https://github.com/golang-migrate/migrate/blob/31c22870d306d0facce2b70c291b458c007b327e/database/postgres/postgres.go#L157-L176

danielsdeleo:
I'm gonna close this in favor of #4183 -- I think adding another layer of locks could work as a solution, but I'd want to implement it differently than this. There's too much nuance here about what happens if you acquire a lock you already have, which locks are global vs. specific to an object, etc.

Issue #4169, opened 2020-08-02
sql lacking idempotency clauses causes failure loop in ha


vjeffrey:
## Describe the bug
customer with automate ha ran into a circular failure when upgrading to the latest release. we determined the cause to be a couple missing `if not exists` 

https://github.com/chef/automate/blob/master/components/config-mgmt-service/backend/postgres/schema/sql/04_rollouts_unique_exclude_constraint.up.sql#L31

https://github.com/chef/automate/blob/master/components/config-mgmt-service/backend/postgres/schema/sql/04_rollouts_unique_exclude_constraint.up.sql#L55
 
## Versions (please complete the following information):
- Automate Build Number: 20200728181447

PR #4187, opened 2020-08-05
Test for Idempotent Schema Migration

danielsdeleo:

In https://github.com/chef/automate/pull/4183 we fixed an error in the migrator that was re-running migrations in multiple front-end cluster topologies, but we still try to forcibly redo migrations if they fail and leave a dirty schema state. So it's best if our migrations are able to re-run. This PR adds a function to our migrator library which drops all the data in the database, then runs a service's migrations all the way through twice. In addition to checking that migrations are idempotent, this approach makes it convenient to develop new migrations, you can keep re-running the integration tests and any bad state created by mistakes gets cleaned up.

I only modified the two services I know well in this pass, it should be pretty easy to add to others if we like.

https://github.com/chef/automate/pull/4183
https://github.com/chef/automate/pull/4169

The fix is in the tests mostly :) but you can run them with:

```
start_deployment_service
chef-automate dev deploy-some chef/config-mgmt-service --with-deps
chef-automate dev deploy-some chef/applications-service --with-deps
```

```
config_mgmt_integration
applications_integration
```

You can delete some of the `IF NOT EXISTS` things to make it fail.

### DIFF
--- a/components/config-mgmt-service/backend/postgres/schema/sql/02_add_rollouts.up.sql
+++ b/components/config-mgmt-service/backend/postgres/schema/sql/02_add_rollouts.up.sql
@@ -18,10 +18,10 @@ CREATE TABLE IF NOT EXISTS rollouts (
   end_time                  TIMESTAMP
 );
 
-CREATE INDEX policy_name_idx        ON rollouts (policy_name);
-CREATE INDEX policy_node_group_idx  ON rollouts (policy_node_group);
-CREATE INDEX policy_revision_id_idx ON rollouts (policy_revision_id);
-CREATE INDEX policy_domain_url_idx  ON rollouts (policy_domain_url);
+CREATE INDEX IF NOT EXISTS policy_name_idx        ON rollouts (policy_name);
+CREATE INDEX IF NOT EXISTS policy_node_group_idx  ON rollouts (policy_node_group);
+CREATE INDEX IF NOT EXISTS policy_revision_id_idx ON rollouts (policy_revision_id);
+CREATE INDEX IF NOT EXISTS policy_domain_url_idx  ON rollouts (policy_domain_url);
 
 
 -- We call the policy name, policy node group, and policy domain URL the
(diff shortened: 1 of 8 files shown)


components/config-mgmt-service/backend/postgres/schema/sql/02_add_rollouts.up.sql
```diff
@@ -18,10 +18,10 @@ CREATE TABLE IF NOT EXISTS rollouts (
...
 );
 
-CREATE INDEX policy_name_idx        ON rollouts (policy_name);
-CREATE INDEX policy_node_group_idx  ON rollouts (policy_node_group);
-CREATE INDEX policy_revision_id_idx ON rollouts (policy_revision_id);
-CREATE INDEX policy_domain_url_idx  ON rollouts (policy_domain_url);
+CREATE INDEX IF NOT EXISTS policy_name_idx        ON rollouts (policy_name);
+CREATE INDEX IF NOT EXISTS policy_node_group_idx  ON rollouts (policy_node_group);
+CREATE INDEX IF NOT EXISTS policy_revision_id_idx ON rollouts (policy_revision_id);
+CREATE INDEX IF NOT EXISTS policy_domain_url_idx  ON rollouts (policy_domain_url);
```
lancewf, line 24: Are some of these changes are from #4183? Is this because the PR has not been rebased?

lib/db/migrator/migrator.go
```diff
@@ -67,6 +67,41 @@ func MigrateWithMigrationsTable(pgURL, migrationsPath, migrationsTable string, l
...
 }
 
+// DestructiveMigrateForTests will:
+// * Drop the database to give you a clean slate
+// * Run all your migrations
+// * Forcibly run all the migrations again to verify that they are idempotent.
+//
+// Obviously you don't want this for production, but you should use it instead
+// of the plain Migrate function in your tests if you can.
+func DestructiveMigrateForTests(pgURL, migrationsPath string, l logger.Logger, verbose bool) error {
```
lancewf, line 77: This is nice. Thank you.

components/config-mgmt-service/backend/postgres/schema/sql/02_add_rollouts.up.sql
danielsdeleo: These are in the right place :) If you look at the commits-view, you can see that I 1) added the destructive double migrator thing, 2) changed config-mgmt to use it and fixed the problems it uncovered, 3) added the destructo migrator to applications service (which didn't have any errors)

PR #4197, opened 2020-08-07
enable configurable content-type header for data-feed-service

satellite15:

There is no standard mime-type for line separated json. The datafeed sets the Content-Type header as application/json.
The most appropriate value for the header would be application/x-ndjson as the datafeed messages adheres to the x-ndjson spec. However to ensure the content-type is supported by any 3rd party integration this update makes it configurable. The default value will remain application/json until the servicenow app is certified to accept application/x-ndjson. The recommendation will be to use application/-x-ndjson, application/json or text/plain in that order ith 3rd party integrations.

compile_all_protobuf_components

build the following components
automate-deployment
automate-gateway
automate-cli
data-feed-service

and deploy

Boostrap nodes or run compliance scans and capture the output in ELK, ServiceNow or requestbin.

Tests have been carried out with ELK, ServiceNow and requestbin.

- [x] Tests added/updated?
- [ ] Docs added/updated? Doc ticket raised to complete docs before feature flag removed


components/data-feed-service/service/data-feed-aggregate-task_test.go
```diff
@@ -0,0 +1,41 @@
...
+	ipAddress = "192.168.10.10"
+	nodeId    = "e5a3d952-5dfc-3215-bfe4-541eb0b5d345"
+)
+
+func TestIsIpaddressTrue(t *testing.T) {
+	isIPAddress := isIPAddress(ipAddress)
+	if !isIPAddress {
+		t.Log("Expected true got false")
+		t.Fail()
+	}
```
srenatus, line 18: If we wanted to use https://github.com/stretchr/testify here, we could make this
```suggestion
	assert.True(isIPAddress(ipAddress))
```
If we want to avoid the dependency (but really, it's used all over Automate already), we could do
```suggestion
	if !isIPAddress(ipAddress) {
		t.Error("expected true got false")
	}
```
[since](https://golang.org/pkg/testing/#T.Error) _"Error is equivalent to Log followed by Fail"_

components/data-feed-service/service/data-feed-aggregate-task_test.go
```diff
@@ -0,0 +1,41 @@
...
+	var filters []string
+	nodeData, err := aggregateTask.safeGetNodeData(context.Background(), filters)
+	if nodeData == nil || len(nodeData) != 0 {
+		t.Logf("Expected empty nodeData, got %v", nodeData)
+		t.Fail()
+	}
+	if err != nil {
+		t.Logf("Expected nil, got %v", err)
+		t.Fail()
+	}
```
srenatus, line 40: With the package mentioned above (its `require` variant), we could make this say
```suggestion
	nodeData, err := aggregateTask.safeGetNodeData(context.Background(), filters)
	require.NoError(t, err)
	require.NotNil(t, nodeData)
	require.NotEmpty(t, nodeData)
```

satellite15: Nice, I'll start using that next time I add some more tests.

Issue #4203, opened 2020-08-07
Lint Automate's Protos with `buf`


danielsdeleo:
## User Story

In order to provide a good experience to API-using customers, we want to standardize our protobuf conventions and avoid breaking changes in our proto-derived APIs. We want to adopt a proto linter to help achieve this, and `buf` is our top choice.

## Definition of Done

* `buf` lints our protos on each Ci run and fails the build if style/correctness violations are detected.
* Old code is excepted from the checks when we cannot safely change it.

## Demo Script / Repro Steps

* Add new content to a .proto file which violates the style and/or correctness rules
* run `buf` locally and it fails
* make a pull request and it goes red

PR #4214, opened 2020-08-11
Add Buf Linter to Go Modules

danielsdeleo:

Adds the buf linter to our go modules and lets you install the binary into the studio. This is the next step to running the linter on our protos in Ci.

#4203 

* Enter the studio
* `install_protoc_toolchain`
* `buf check lint`

There are a lot of errors/violations/whatevers when you run the linter, I will be getting to those in a future patch.


.studio/common
```diff
@@ -31,8 +31,23 @@ function bootstrap_proto_compilers() {
...
+document "install_protoc_toolchain" <<DOC
+  Install all the protobuf plugins, linters, etc. we use to compile and vet our protobufs.
+DOC
+function install_protoc_toolchain() {
+  local proto_go_tools=(
+    github.com/chef/automate/components/automate-grpc/protoc-gen-policy
+    github.com/chef/automate/components/automate-grpc/protoc-gen-grpc-mock # generate mock grpc server implementations
+    github.com/chef/automate/components/automate-grpc/protoc-gen-a2-config
+    github.com/envoyproxy/protoc-gen-validate
+    github.com/ckaznocha/protoc-gen-lint
```
srenatus, line 43: I believe we can ditch this one if we're hooking up buf.

It's called here: https://github.com/chef/automate/blob/c9a5ab05e76c4fc4e91f0342caa5a16c1a431fb1/components/automate-gateway/scripts/grpc.sh#L86-L99

and we can remove the `nolint` comments from here, I suppose:
```console
$ git grep lint | grep proto:
api/config/gateway/config_request.proto:3:// nolint: 'V1' and 'S3' doesn't look like CamelCase to the linter. Curious.
components/automate-gateway/api/notifications/notifications.proto:3:// nolint: keeping this as-is for backwards-compat reasons (for now)
```

api/interservice/authn/authenticate.pb.validate.go
```diff
@@ -33,6 +33,9 @@ var (
 	_ = ptypes.DynamicAny{}
 )
 
+// define the regex for a UUID once up-front
+var _authenticate_uuidPattern = regexp.MustCompile("^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$")
+
```
srenatus, line 38: Huh this means we must have also upgraded our deps...? Not a bad thing. It means we could use the `uuid` well-known string type in a few places, maybe.

danielsdeleo: Yeah, installing `buf` via go modules requires pretty new versions of all the grpc/proto things, we should be completely up to date now 🎉

.studio/common
danielsdeleo: 👍 thanks for digging in, I'll put this on my list for when I start actually running buf against the codebase

PR #4216, opened 2020-08-11
chef-server: Upgrade to latest current build

stevendanna:
This upgrades us to the latest unstable packages built from master.
To do so, it:

- Adds server_name_indication SSL configuration and our custom `{verify, verify_ca}`to avoid the issue
  outlined in chef/automate#2002.

- Updates other SSL related configuration to be in compliance with our security checks.

- Adapts our various wrapper scripts to changes in the upstream
  packages around ruby+bundler. Chef Server is currently in the
  process of trying to upgrade many of its deps, so more churn here is
  likely.

- Sets the new retry_on_conn_closed option for oc_httpc to true, which
  hopefully works around some recent ibrowse changes.


components/automate-cs-oc-erchef/habitat/config/sys.config
```diff
@@ -126,12 +138,14 @@
             {cull_interval, {1, min}},
             {max_age, {70, sec}},
             {max_connection_duration, {70, sec}},
+            {retry_on_conn_closed, true},
```
PrajaktaPurohit, line 141: ohh nice! you remembered to get this!

Issue #4201, opened 2020-08-07
Locked for only one ingest-service migration


lancewf:
## User Story

Because of Automate HA frontend, we can have multiple of the same service running elasticsearch migrations at the same time. This task is to create the first cereal elasticsearch migration to ensure only one service is running the migration. 

## Definition of Done
When multiple ingest services are running under Automate HA frontend, only one can run an elasticsearch migration. 

## Demo Script / Repro Steps
1. Start Automate HA frontend with two frontends
1. Update both ingest-service to run the chef actions migration
1. Ensure only one runs the migration and the other one waits until it is complete.

PR #4219, opened 2020-08-11
Ingest service Elasticsearch migration lock

lancewf:

For ingest-service elasticsearch migrations, we added PostgreSQL lock to prevent Automate HA frontend with multiple ingest-services from running the same migration at the same time.

#4201 

1. `start_all_services`
1. While automate is starting, change this line from "false" to "true". https://github.com/chef/automate/blob/c0a74e9df9db459d2d5b2fa9f8b74ca1d0c2386d/components/ingest-service/migration/migration.go#L200
1. Add the below code to here https://github.com/chef/automate/blob/c0a74e9df9db459d2d5b2fa9f8b74ca1d0c2386d/components/ingest-service/migration/migration.go#L153
```
	log.Info("Starting migration")
	time.Sleep(time.Minute)
	log.Info("Finished Migration")
```
1. Setup tmux with 
```
cat > /root/.profile <<'endmsg'
if [ -f /tmp/tmux ]; then
  source /src/.studiorc
fi

touch /tmp/tmux
endmsg
```
1. Install tmux with `hab pkg install -b core/tmux`
1. Start a tmux session with `tmux`
1. Create a new terminal with `CTR-b "` Switch between the terminal with `CTR-b` down or up arrows.
1. In one of the terminals (A) run `chef-automate dev psql chef_ingest_service`
1. In the other (B) run `sl | grep "ingest-service.default" &`
1. Then run `rebuild components/ingest-service` in terminal B
1. To watch the postgresql lock in terminal A run `SELECT COUNT(*) FROM pg_locks WHERE locktype = 'advisory' AND objid =77;`
1. When you see the "Starting migration" in terminal B there is a lock in postgresql.
1. After seeing "Finished Migration" the lock should be removed. 
1. Run the `rebuild components/ingest-service` another time and when the migration is running (right after "Starting migration" is logged) run `kill_running_service ingest-service` and ensure the postgresql lock is removed. 
1. Simulate another ingest-service is running a migration with adding the lock manually with `SELECT pg_advisory_lock(77);` in terminal A. 
1. Then in terminal B run `rebuild components/ingest-service` and ensure you see this message in the logs "HA Frontend: migration is currently running on another service". 

- [ ] Tests added/updated?
- [ ] Docs added/updated?

image

### DIFF
--- a/components/ingest-service/grpc/grpc.go
+++ b/components/ingest-service/grpc/grpc.go
@@ -28,6 +28,7 @@ import (
 	"github.com/chef/automate/lib/cereal"
 	"github.com/chef/automate/lib/cereal/postgres"
 	"github.com/chef/automate/lib/datalifecycle/purge"
+	libdb "github.com/chef/automate/lib/db"
 	"github.com/chef/automate/lib/grpc/secureconn"
 	platform_config "github.com/chef/automate/lib/platform/config"
 )
@@ -55,18 +56,27 @@ func Spawn(opts *serveropts.Opts) error {
 		return err
 	}
 
-	var (
-		grpcServer = opts.ConnFactory.NewServer()
-		migrator   = migration.New(client)
-		uri        = fmt.Sprintf("%s:%d", opts.Host, opts.Port)
-	)
-
+	uri := fmt.Sprintf("%s:%d", opts.Host, opts.Port)
 	log.WithField("uri", uri).Info("Starting gRPC Server")
 	conn, err := net.Listen("tcp", uri)
 	if err != nil {
 		log.WithError(err).Error("TCP listen failed")
 		return err
 	}
+
+	grpcServer := opts.ConnFactory.NewServer()
+
+	pgURL, err := pgURL(opts.PGURL, opts.PGDatabase)
+	if err != nil {
+		log.WithError(err).Fatal("could not get PG URL")
+		return err
+	}
+
+	dbMap, err := libdb.PGOpen(pgURL)
+	if err != nil {
+		return errors.Wrapf(err, "Failed to open database with uri: %s", pgURL)
+	}
+	migrator := migration.New(client, dbMap)
 
 	// Register Status Server
 	ingestStatus := server.NewIngestStatus(client, migrator)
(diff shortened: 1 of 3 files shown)


components/ingest-service/grpc/grpc.go
```diff
@@ -83,19 +93,29 @@ func Spawn(opts *serveropts.Opts) error {
...
-			return err
+	migrationRunning, err := migrator.MigrationRunning()
+	if err != nil {
+		log.WithError(err).Error("Failed checking if a migration is running")
+		return err
+	}
+
+	if !migrationRunning {
+		if migrationNeeded {
+			err = migrator.Start()
```
danielsdeleo, line 104: There could be a small gap of time between calling `migrator.MigrationRunning()` and calling `migrator.Start()`, and another front-end could claim the lock during that time. The documentation (https://www.postgresql.org/docs/9.6/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS) is not as explicit as I would prefer but it says that `pg_advisory_lock` will block/wait until the lock becomes available. 

One thing you could do instead is to use `pg_try_advisory_lock()` which returns a boolean (false if you didn't get the lock). Then your program flow here might be like

```go
// make a specific error for when the migration lock is taken, change `(ms *Status) lock()` to return this when appropriate
var MigrationInProgress = fmt.Errorf("another process or system has the migration advisory lock 77")

if migrationNeeded {
  err = migrator.Start()
  if err != nil && err == migration. MigrationInProgress {
   // you might want to log the number you're using for the lock in case something gets stuck and a human wants to manually unstick it
    log.Info("another process or system is already running the migration, skipping it...")
  }
```

lancewf: Thank you for pointing that out. I was thinking that it would fail [here](https://github.com/chef/automate/pull/4219/files#diff-0d1121023e5ce205c1be3cc1b5ff83b9R74) if another service tried to start the migration at the same time. That would kill the service and restart it. I like your idea of recovering from the error though. I will update it.

lancewf: Oh, I thought the `pg_advisory_lock` would fail if it could not get the lock. Good call.

lancewf: @danielsdeleo I changed this to use the `pg_try_advisory_lock`command, but the service will restart if the try fails. I tested this out and I did not see a problem.

Issue #3708, opened 2020-05-17
notification create consistency improvements


susanev:
## overview
some consistency improvements for the notifications create modal

## design details
[logo and icon image files](https://drive.google.com/drive/folders/1PBRUILJ_AqOCDhXByRNgKaSZGC4xIHNl?usp=sharing)

#### slack and webhook
- Name *
- Webhook Type *
- Failure Type *
- Webhook URL *

#### service now
- use an animation to expand the form a bit, if they choose service now as the webhook type, to reveal the two additional fields. reach out to @SEAjamieD if you need help deciding on an animation
- Name *
- Webhook Type *
- Failure Type *
- Webhook URL *
- ServiceNow Username *
- Service Now Password *
- help text should be using `font-weight: 300;` this also needs to be updated on the data feeds create form

### screenshots
![notifications-create](https://user-images.githubusercontent.com/5489125/82158144-e70aef00-983a-11ea-98f8-fd496187e6bd.png)

![notifications-create-service-now](https://user-images.githubusercontent.com/5489125/82158220-54b71b00-983b-11ea-9068-289cbba70427.png)

PR #4226, opened 2020-08-12
Create notification form and delete notification consistency changes

vinay033:
consistency improvements for the create notification, same as the rest of automate.

1. create notification using the modal popup
2. test the notification webhook URL with success/failure notification on same modal popup
3. delete a notification, same as the rest of automate

https://github.com/chef/automate/issues/3708
### :+1: Definition of Done
I have added consistency changes for the `create notification` and `delete notification` with UI changes.
visit the Settings >> Notifications page
 create notification - if you have no data then add notification entries using the `create notification` button.
a modal popup is open and you can create the notification >> added notification will display on the notification page 
 delete notification - click to any notification left side three-dot menu `...`  where you can see the delete notification link to test delete the notification.
```
STEP 1
inside the hab studio

STEP 2
open new window
go to automate UI path

$ cd components/automate-ui
and run the command 
```

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### :camera: Screenshots
1. create modal popup 
notification1
notification2
notification3
2. test URL with success notificaion
notification4
3. test URL with a failure notification
notification5
4. create notification success notification
notification6
5. delete notification confirmation modal
notification7
6. delete notification success notification
notification8


components/automate-ui/src/app/app-routing.module.ts
```diff
@@ -176,14 +175,6 @@ const routes: Routes = [
...
               component: NotificationsComponent
             },
-            {
-              path: 'form',
-              component: NotificationFormComponent
-            },
-            {
-              path: 'form/:id',
-              component: NotificationFormComponent
-            },
```
msorens, line 186: Why are these no longer needed?

components/automate-ui/src/app/pages/create-notification-modal/create-notification-modal.component.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  urlPresent() {
+    return this.createForm.controls.targetUrl.value !== '' ? true : false;
+  }
+
+  getAlertTypeKeys() {
+    return this.notificationRule.getAlertTypeKeys();
+  }
+
+  getTagetTypeKeys() {
```
msorens, line 71:
```suggestion
  getTargetTypeKeys() {
```

components/automate-ui/src/app/entities/notification_rules/notification_rule.effects.ts
```diff
@@ -45,7 +54,63 @@ export class NotificationRuleEffects {
...
+    }));
+
+  @Effect()
+  deleteNotificationRuleFailure$ = this.actions$.pipe(
+    ofType(NotificationRuleActionTypes.DELETE_FAILURE),
+    map(({ payload }: DeleteNotificationRuleFailure) => {
+      const msg = payload.error.error;
+      return new CreateNotification({
+        type: Type.error,
+        message: `Could not delete rule: ${msg || payload.error}`
```
msorens, line 113: To be consistent 😉 let's use this (like L84):
```suggestion
        message: `Could not delete rule: ${payload.error.error || payload.error}`
```

components/automate-ui/src/app/entities/notification_rules/notification_rule.reducer.ts
```diff
@@ -33,43 +41,113 @@ export function notificationRuleEntityReducer(
...
 
     case NotificationRuleActionTypes.GET_ALL_SUCCESS:
       return pipe(
-        set(GET_ALL_STATUS, EntityStatus.loadingSuccess))
-        (notificationRuleEntityAdapter.setAll(action.payload, state));
+        set(GET_ALL_STATUS,
+          EntityStatus.loadingSuccess)
+        )
+        (notificationRuleEntityAdapter.setAll(action.payload, state)
+      ) as NotificationRuleEntityState;
```
msorens, line 56: Can simplify this expression--no need for the pipe:
```
      return set(GET_ALL_STATUS,
        EntityStatus.loadingSuccess,
        notificationRuleEntityAdapter.setAll(action.payload, state));
```

components/automate-ui/src/app/entities/notification_rules/notification_rule.reducer.ts
```diff
@@ -33,43 +41,113 @@ export function notificationRuleEntityReducer(
...
 
     case NotificationRuleActionTypes.GET_ALL_FAILURE:
-      return set(GET_ALL_STATUS, EntityStatus.loadingFailure, state);
+      return set(
+        GET_ALL_STATUS,
+        EntityStatus.loadingFailure,
+        state
+      ) as NotificationRuleEntityState;
```
msorens, line 63: A thing I noticed recently in the code base: we actually do not need type casts on set calls in reducers, only on pipe calls. So you can remove that on every `set` call in this file like this.
```suggestion
      );
```
(Not suggesting to update other files just now, but it is cleaner going forward to do this on files you touch.)

components/automate-ui/src/app/entities/notification_rules/notification_rule.reducer.ts
```diff
@@ -33,43 +41,113 @@ export function notificationRuleEntityReducer(
...
 
     case NotificationRuleActionTypes.UPDATE_SUCCESS:
       return set(UPDATE_STATUS, EntityStatus.loadingSuccess,
         notificationRuleEntityAdapter.updateOne({
           id: action.payload.id,
           changes: action.payload
-        }, state));
+        }, state)
+      ) as NotificationRuleEntityState;
```
msorens, line 122: Looks like you added the typecast here, but prefer to remove the unneeded ones, not add more.

components/automate-ui/src/app/ngrx.reducers.ts
```diff
@@ -29,13 +29,13 @@ import * as integrationsDetail from './pages/integrations/detail/integrations-de
...
 import * as license from './entities/license/license.reducer';
 import * as layout from './entities/layout/layout.reducer';
-import * as notificationEntity from './entities/notifications/notification.reducer';
 import * as credential from './entities/credentials/credential.reducer';
 import * as jobAdd from './pages/job-add/job-add.reducer';
 import * as jobEdit from './pages/job-edit/job-edit.reducer';
 import * as jobEntity from './entities/jobs/job.reducer';
 import * as jobList from './pages/job-list/job-list.reducer';
 import * as manager from './entities/managers/manager.reducer';
+import * as notificationEntity from './entities/notifications/notification.reducer';
```
msorens, line 38: If you are sorting, there are a few other lines out of place (license, layout, credential, orgEntity).

components/automate-ui/src/app/pages/create-notification-modal/create-notification-modal.component.html
```diff
@@ -0,0 +1,126 @@
...
+          <chef-error *ngIf="conflictError">
+            Name "{{createForm.get('name').value}}" already exists.
+          </chef-error>
+          <span class="i-desc">Don't worry, notification names can be changed later.</span>
+          </chef-form-field>
+        </div>
+        <div class="margin">
+          <chef-form-field>
+            <label>
+              <span class="label">Webhook type<span aria-hidden="true">*</span></span>
```
msorens, line 24:
```suggestion
              <span class="label">Webhook type <span aria-hidden="true">*</span></span>
```

components/automate-ui/src/app/pages/create-notification-modal/create-notification-modal.component.html
```diff
@@ -0,0 +1,126 @@
+<chef-modal [visible]="visible" (closeModal)="closeEvent()">
+  <h2 slot="title">Create Notification </h2>
```
msorens, line 2:
```suggestion
  <h2 slot="title">Create Notification</h2>
```

components/automate-ui/src/app/pages/create-notification-modal/create-notification-modal.component.html
```diff
@@ -0,0 +1,126 @@
...
+                  {{ notificationRule.TargetTypeLabels[key] }}
+                </mat-option>
+              </mat-select>
+            </label>
+          </chef-form-field>
+        </div>
+        <div class="margin">
+          <chef-form-field>
+            <label>
+              <span class="label">Failure type  <span aria-hidden="true">*</span></span>
```
msorens, line 41:
```suggestion
              <span class="label">Failure type <span aria-hidden="true">*</span></span>
```

components/automate-ui/src/app/pages/create-notification-modal/create-notification-modal.component.html
```diff
@@ -0,0 +1,126 @@
...
+    <div class="flex-container">
+      <form [formGroup]="createForm">
+        <div class="margin">
+          <chef-form-field>
+            <label>
+              <span class="label">Name <span aria-hidden="true">*</span></span>
+              <input chefInput formControlName="name" type="text" id="name-input" (keyup)="handleInput($event)" autocomplete="off"/>
+            </label>
+            <chef-error
+            *ngIf="(createForm.get('name').hasError('required') || createForm.get('name').hasError('pattern')) && createForm.get('name').dirty">
```
msorens, line 12: Should indent here and in other similar ones.
```suggestion
              *ngIf="(createForm.get('name').hasError('required') || createForm.get('name').hasError('pattern')) && createForm.get('name').dirty">
```

components/automate-ui/src/app/pages/create-notification-modal/create-notification-modal.component.html
```diff
@@ -0,0 +1,126 @@
...
+          <div class="margin">
+            <chef-form-field>
+              <label>
+                <span class="label">ServiceNow Username <span aria-hidden="true">*</span></span>
+                <input chefInput formControlName="username" type="text" id="username-input" (keyup)="handleInput($event)" autocomplete="off"/>
+              </label>
+              <chef-error *ngIf="createForm.get('username').hasError('required') && createForm.get('username').dirty">
+                Username is required.
+              </chef-error>
+              <span class="i-desc">Encrypted, and only used to access the notification.</span>
```
msorens, line 85: I assume this message should be the same as that on L97?
```suggestion
              <span class="i-desc">Encrypted, and only used to access the notification feed.</span>
```

components/automate-ui/src/app/pages/create-notification-modal/create-notification-modal.component.html
```diff
@@ -0,0 +1,126 @@
...
+              </chef-error>
+              <span class="i-desc">Encrypted, and only used to access the notification feed.</span>
+            </chef-form-field>
+          </div>
+        </div>
+
+        <div id="button-bar">
+          <div class="url-failed" *ngIf="hookStatus === urlState.Failure">
+            <chef-icon class="url-failed-icon">warning</chef-icon>
+            <span>Unable to connect, check URL, username and password.</span>
```
msorens, line 105: Grammar fix:
```suggestion
            <span>Unable to connect: check URL, username and password.</span>
```

components/automate-ui/src/app/pages/create-notification-modal/create-notification-modal.component.ts
```diff
@@ -0,0 +1,103 @@
...
+  createNotification(): void {
+    this.createClicked.emit();
+  }
+
+  sendTest() {
+    this.sendTestClicked.emit();
+  }
+
+  urlPresent() {
+    return this.createForm.controls.targetUrl.value !== '' ? true : false;
```
msorens, line 64:
```suggestion
    return this.createForm.controls.targetUrl.value !== '';
```

components/automate-ui/src/app/pages/create-notification-modal/create-notification-modal.component.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+      username: ['', null],
+      password: ['', null]
+    });
+    component.conflictErrorEvent = new EventEmitter();
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
```
msorens, line 67: Please add tests of the component's code.
(If this is replacing notification-form.component.spec.ts, are any of those tests able to be migrated over here?)

components/automate-ui/src/app/pages/notifications/notifications.component.html
```diff
@@ -5,6 +5,21 @@
...
+        [notificationRule]="notificationObj"
+        [createForm]="createNotificationForm"
+        [targetKeysValue]="targetKeys"
+        [alertTypeKeysValue]="alertTypeKeys"
+        (close)="closeCreateModal()"
+        [conflictErrorEvent]="conflictErrorEvent"
+        (createClicked)="createNotification()"
+        (sendTestClicked)="sendTestForNotification()"
+        (targetTypeChanged)="updateTargetType()">
+      </app-create-notification-modal>
```
msorens, line 22: This is a red flag--there are way too many inputs here. If you recall, some time ago I suggested the approach of making modals more self-contained. They should manage their own state rather than having the parent do it, when possible. Please take a look and see if you can reduce the surface area.

components/automate-ui/src/app/pages/notifications/notifications.component.ts
```diff
@@ -189,12 +293,52 @@ export class NotificationsComponent implements OnInit {
...
+    this.hookStatus = status;
+  }
+
   // TODO - this was common in all three uses, but I'm not sure this is the best
   // way to do it - do we really need to refresh after the server confirms the action
   // successful? Seem it should be possible to update the local model with the
   // changes and have that trigger view updates?
   refreshRules() {
-    this.rules$ = this.service.fetchRules();
+    // this.rules$ = this.service.fetchRules();
```
msorens, line 341: Delete

susanev:
@vinay033 thank you for doing this work, i found 2 things id like fixed
- [ ] there is extra space at the bottom of the create modal that should not be there, the modal should auto expand in height to account for the notification. the space below the buttons should match whatever is in the other create modals
create-modal-extra-space

- [ ] it should not allow you to use the create button if there is an invalid url provided, right now it lets you press create and then gives you an error notification

components/automate-ui/src/app/app-routing.module.ts
vinay033: yes, we don't need this, because now we can create notification using the modal pop and for edit and show our consistency changes are already merged.

components/automate-ui/src/app/ngrx.reducers.ts
vinay033: actually I only use sort for my code but I will arrange others as well.

components/automate-ui/src/app/pages/notifications/notifications.component.html
vinay033: every input required, because we can also avoid using function calls on HTML so we are using the input variable instead of function.

components/automate-ui/src/app/pages/create-notification-modal/create-notification-modal.component.spec.ts
vinay033: I have have covered in notification.component.ts

components/automate-ui/src/app/pages/notifications/notifications.component.html
msorens: @vinay033 You are not quite understanding my suggestion. Took me a bit to dig up my reference example, but please take a look at https://github.com/chef/automate/pull/2704, where I first introduced the **self-contained modal** pattern.

Specifically, what I am suggesting is to move all the logic relevant to _creating_ a notification into `CreateNotificationModalComponent` rather than have it sit in the notifications list page (`NotificationsComponent`), which need only know about _listing_ notifications. Analogously, in the PR I referenced, you will see the logic relevant to creating a user has been move into `CreateUserModalComponent`, because the consumer of that component, `UserManagementComponent`, needs only to know about listing users, not about creating them.

So following the reference PR, you would add an input that is a signal to the modal to open. You can then get rid of the `[visible]` input above; the component manages that property itself.
https://github.com/chef/automate/blob/22e5d7d8dc4c954a6beb0db2c9d238b2b02b8be1/components/automate-ui/src/app/modules/team/team-add-users/team-add-users.component.ts#L186-L188

Then, you move the logic to do the creation as well, which means you can get rid of the `[creating]` input; that will be managed within the modal. Also, testing the notification is done inside the modal (the <kbd>Test Notification</kbd> button), so move that logic in there and that gets rid of the `[sending]` input. 

In a similar fashion, I believe you can remove most if not all of the other inputs. As to what happens when the modal closes, you should not even need to send back any output from `CreateNotificationModalComponent` to `NotificationsComponent`--the subscriptions on the parent page should automatically load the new notification into the list.

vjeffrey:
can we merge this?

msorens:
Prefer not; feedback has not been addressed.

vinay033:
> Prefer not; feedback has not been addressed.

work in progress for this and will update this PR soon.

components/automate-ui/src/app/pages/notifications/notifications.component.html
vinay033: sorry for the late reply, I followed the PR which you were suggested and I have updated the code part only I need to change in test cases, as soon as I will update the PR.

vinay033: @msorens I have updated the PR, please check it once.

components/automate-ui/src/app/pages/notifications/notifications.component.ts
```diff
@@ -39,13 +46,18 @@ export class NotificationsComponent implements OnInit {
...
   // This is exposed here to allow the component HTML access to ServiceActionType
   serviceActionType = ServiceActionType;
+
   public notificationToDelete: NotificationRule;
   public deleteModalVisible = false;
+  public hookStatus = UrlTestState.Inactive;
+  public notificationObj = new NotificationRule('', '', null, '', null, '', false);
+  private isDestroyed = new Subject<boolean>();
+
+  public openNotificationModal = new EventEmitter<boolean>();
```
msorens, line 56: I realize you adapted the existing model code for this (👍 ), but I think we can improve it slightly. We do not need a boolean at all; we just need the event.
Make this a `void` (or `null` also would work here):
```suggestion
  public openNotificationModal = new EventEmitter<void>();
```
Then change the emit below from this
```
this.openNotificationModal.emit(true);
```
to this
```
this.openNotificationModal.emit();
```
And similarly need a change in create-notification-modal.component.ts from `<boolean>` to `<void>`.

components/automate-ui/src/app/pages/notifications/notifications.component.ts
```diff
@@ -70,6 +83,15 @@ export class NotificationsComponent implements OnInit {
     this.toggleSort('name');
   }
 
+  ngOnDestroy(): void {
+    this.isDestroyed.next(true);
```
msorens, line 87: This is not hooked up--need to add `takeUntil(this.isDestroyed)` on subscriptions in this file.

components/automate-ui/src/app/pages/notifications/notifications.component.ts
```diff
@@ -194,7 +215,6 @@ export class NotificationsComponent implements OnInit {
   // successful? Seem it should be possible to update the local model with the
   // changes and have that trigger view updates?
   refreshRules() {
-    this.rules$ = this.service.fetchRules();
     this.rules$.subscribe(rules => {
```
msorens, line 218: Since you are in here, this is a problem that needs addressing. Subscriptions should almost always be in ngOnInit only. This will result in multiple subscriptions to the same thing--that is a bad thing. You should be able to move this `this.rules$.subscribe` block into `ngOnInit` and get rid of `refreshRules` completely and everything should still work the same.

components/automate-ui/src/app/pages/create-notification-modal/create-notification-modal.component.ts
```diff
@@ -0,0 +1,219 @@
...
+      username: [[]],
+      password: [[]]
+    });
+
+    this.targetKeys = this.notificationRule.getTargetTypeKeys();
+    this.alertTypeKeys = this.notificationRule.getAlertTypeKeys();
+  }
+
+  ngOnInit() {
+    this.openEvent.pipe(takeUntil(this.isDestroyed))
```
msorens, line 83: You have everything nicely implemented with the `takeUntil` -- but `this.isDestroyed` never fires because you do not implement `OnDestroy` in the class.

msorens (commented):
Really excellent work implementing the self-contained modal!
You are almost there; found just a few other things that do need correcting.

vinay033:
@msorens Thank you for the review, I have addressed all review comments, please have a look.

components/automate-ui/src/app/pages/create-notification-modal/create-notification-modal.component.ts
```diff
@@ -41,7 +42,7 @@ enum UrlTestState {
   templateUrl: './create-notification-modal.component.html',
   styleUrls: ['./create-notification-modal.component.scss']
 })
-export class CreateNotificationModalComponent implements OnInit {
+export class CreateNotificationModalComponent implements OnInit, OnDestroy {
   @Input() openEvent: EventEmitter<boolean>;
```
msorens, line 46: Just need to update this to match the earlier change:
```suggestion
  @Input() openEvent: EventEmitter<void>;
```

msorens (approved):
Looks fine, except for one bit of bookkeeping. I am going to ask if @susanev can do a re-review due to all the changes, and then we should be good to go.

components/automate-ui/src/app/pages/create-notification-modal/create-notification-modal.component.ts
vinay033: I have added these changes and updated the PR.

PR #4230, opened 2020-08-12
Modify Internal Protos to Conform to the Linter

danielsdeleo:

Modify the internal protobuf messages to conform to the linter. This way we can have the fewest exceptions to the lint rules which will reduce the feeling that the linter is inconsistent.

### :+1: Definition of Done

The code is supposed to work the same , at least from the point of view of an external user. Just we now follow the protobuf style guide a little better.


srenatus:
@msorens the rules are layed out here: https://buf.build/docs/style-guide including some rationale ("Why?"), too.

Note that some linters are disabled, see `buf.yaml` in the toplevel of this repo.

PR #4246, opened 2020-08-20
docs: add Azure AD note on SAML NameID setting and transformations

srenatus:

This tries to add a useful hint to avoid people stumbling over this, as has happened in a recent customer issue.

I'm finding it difficult to describe the Azure UI succinctly; the dropdown that matters is this one:

### DIFF
--- a/components/docs-chef-io/content/automate/saml.md
+++ b/components/docs-chef-io/content/automate/saml.md
@@ -92,17 +92,13 @@ where:
 See the [SAML Configuration Settings]({{< relref "saml.md#saml-configuration-settings" >}}) for further configuration options.
 
 {{< warning >}}
-Azure AD lets you chose the _NameID_ field, and optionally apply _transformations_
-to it. The setting is only respected if the value selected in "Choose name identifier
-format" matches the `name_id_policy_format` configured in Chef Automate.
+Azure AD lets you chose the _NameID_ field, and optionally apply _transformations_ to it.
+The SAML configuration setting is only respected if the selected value in "Choose name identifier format" in Azure AD matches the `name_id_policy_format` configuration in Chef Automate.
 
-For example, if you want to select a specific user attribute to be used as NameID,
-say, `user.email`, and you want to apply a transformation to it, these values need
-to up:
-1. In "Choose name identifier format", select "Email address", and configure the
-   attribute, and eventually transformation, you want to use for that.
-2. In Chef Automate's configuration, `name_id_policy_format` needs to be set to
-   `urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress`.
+For example, when selecting the specific user attribute `user.email` for NameID in Azure AD:
+
+1. In Azure AD for "Choose name identifier format", select "EmailAddress", and configure the attribute and (optional) transformation.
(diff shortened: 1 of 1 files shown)


components/automate-chef-io/content/docs/saml.md
```diff
@@ -89,6 +89,12 @@ where:
 
 See the SAML Configuration Settings below for further configuration options.
 
+{{% warning %}}
+Azure AD lets you chose the _NameID_ field, and optionally apply _transformations_
+to it. The setting is only respected if the value selected in "Choose name identifier
```
yzl, line 94: [suggestion only] I would be explicit about which setting (I'm assuming it's _transformations_).

susanev: also if ur worried about being able to describe it correctly with words, we could put a screenshot here to help maybe?

srenatus: I think it's about every setting. I.e. if you configure the NameID field to be `user.petname`, and you don't change the name id format from "persistent" (to "email"), and A2 is configured to request name id policy format "email", azure ad's AuthnResponse will contain some sort of default for the "email" name id policy, presumably `user.email`.

Only if you _also_ change the drop down to "email" will A2 end up getting `user.petname` in the relevant NameID field.

However, the same applies when a transformation is applied (say, `ToLowercase(user.petname)`).

I have yet to verify this detail, however 🙄  -- would you have an idea how to change the text to make this more apparent?

srenatus: > also if ur worried about being able to describe it correctly with words, we could put a screenshot here to help maybe?

@susanev I'm happy to supply a screenshot, I'm just hesitant to blow this one detail out of proportion... what do you think?

yzl: @srenatus Off the top of my head, the one idea I have is to state which things have to line up in order for stuff to work with A2, then say something like "That means that if you apply a transformation, stuff has to line up like X," and give a concrete example.

srenatus:
image

☝️ I don't really know what went wrong here ☹️

components/docs-chef-io/content/automate/saml.md
```diff
@@ -89,7 +89,17 @@ where:
...
-See the SAML Configuration Settings below for further configuration options.
+See the [SAML Configuration Settings]({{< relref "saml.md#saml-configuration-settings" >}}) for further configuration options.
+
+{{< warning >}}
+Azure AD lets you chose the _NameID_ field, and optionally apply _transformations_ to it.
+The SAML configuration setting is only respected if the selected value in "Choose name identifier format" in Azure AD matches the `name_id_policy_format` configuration in Chef Automate.
+
+For example, when selecting the specific user attribute `user.email` for NameID in Azure AD:
+
+1. In Azure AD for "Choose name identifier format", select "EmailAddress", and configure the attribute and (optional) transformation.
```
mjingle, line 100: @srenatus Are there more details about configuring the attribute and transformation to add for this example?

mjingle: With discussion, this question has been answered! We're keeping this warning simple.

PR #4108, opened 2020-07-21
Sending chef-server actions as events to the event-feed-service

lancewf:
### :nut_and_bolt: Description: What code changed, and why?
Migrating all the chef-server actions from the config-mgmt-service's "actions-*" indexes to the event-feed-service. This places all the events in one service. This allows other services to easily add events to the event feed. This simplifies the event feed service to not split requests between two services. 

### :bug: Bug fixed
Project update and project tagging on ingestion for Chef actions were using the incorrect field. It was using the "remote_hostname" but it should have been using the "service_hostname". These are both defined here https://github.com/chef-boneyard/chef-rfc/blob/master/rfc077-mode-agnostic-data-collection.md. 
This PR fixes the problem and re-ingests the action which fixes any incorrect project tags. 

Another bug was found that prevented the event-feed-service's events from filtering correctly on event types. components/event-feed-service/pkg/feed/feed.go

### :chains: Related Resources

#444 

### :athletic_shoe: How to Build and Test the Change
1. `start_all_services`
1. Add some actions with `chef_load_actions 25`
1. To ensure the actions are in the old indexes run `es_list_indices`. Then ensure there are indexes that start with "actions-".
1. Open the event feed UI page take a snapshot or remember how it looks.  
1. Update to the PR changes with: `rebuild components/event-feed-service && rebuild components/config-mgmt-service && rebuild components/ingest-service && rebuild components/automate-gateway && rebuild components/automate-deployment/`
1. To ensure the migration worked correctly, open the event feed page, and ensure the events added above have all the correct information.
1. Run `es_list_indices` again to ensure all the "action-" indexes are removed. 

### :white_check_mark: Checklist

- [x] I have read the [CONTRIBUTING document](https://github.com/chef/automate/blob/master/CONTRIBUTING.md).
- [x] Tests added/updated?
- [ ] Docs added/updated?
- [x] All commits have been signed-off for the [Developer Certification of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

PR #4249, opened 2020-08-20
Local User events

lancewf:

Adding create, update, and delete events for Automate local users.

https://github.com/chef/automate/pull/4108

### :+1: Definition of Done
Events are created when creating, deleting, and updating Automate's local users. 

1. `build components/automate-gateway && start_all_services`
1. Start the local UI or `rebuild componets/automate-ui`
1. Go to https://a2-dev.test/settings/users and create a new user. 
1. Update that users display name
1. Delete the user. 
1. Go to https://a2-dev.test/dashboards/event-feed and ensure there are three events for each action. 

- [ ] Tests added/updated?
- [ ] Docs added/updated?

Screen Shot 2020-08-20 at 1 19 10 PM


components/automate-gateway/gateway/services.go
```diff
@@ -208,7 +208,7 @@ func (s *Server) RegisterGRPCServices(grpcServer *grpc.Server) error {
 	if err != nil {
 		return errors.Wrap(err, "create client for users mgmt service")
 	}
-	pb_iam.RegisterUsersServer(grpcServer, handler_users.NewServer(usersMgmtClient))
+	pb_iam.RegisterUsersServer(grpcServer, handler_users.NewServer(usersMgmtClient, eventFeedClient))
```
lancewf, line 211: The event feed client is used to send the event.

components/automate-gateway/handler/iam/v2/users/users.go
```diff
@@ -39,6 +47,11 @@ func (p *Server) CreateUser(
 		return nil, err
 	}
 
+	_, err = p.eventFeedClient.HandleEvent(ctx, createUserCreateEvent(resp))
+	if err != nil {
+		logrus.Errorf("Problem sending event %v", err.Error())
+	}
```
lancewf, line 53: Sending the create user event.

components/automate-gateway/handler/iam/v2/users/users.go
```diff
@@ -63,6 +76,12 @@ func (p *Server) DeleteUser(
 	if err != nil {
 		return nil, err
 	}
+
+	_, err = p.eventFeedClient.HandleEvent(ctx, createUserDeleteEvent(req))
+	if err != nil {
+		logrus.Errorf("Problem sending event %v", err.Error())
+	}
```
lancewf, line 83: Sending the delete user event.

components/automate-gateway/handler/iam/v2/users/users.go
```diff
@@ -89,6 +108,11 @@ func (p *Server) UpdateUser(
 		return nil, err
 	}
 
+	_, err = p.eventFeedClient.HandleEvent(ctx, createUserUpdateEvent(resp))
+	if err != nil {
+		logrus.Errorf("Problem sending event %v", err.Error())
+	}
+
```
lancewf, line 115: Sending the update user event.

components/automate-ui/src/app/helpers/event-feed/event-helper.ts
```diff
@@ -0,0 +1,88 @@
+
+import { endsWith, replace } from 'lodash/fp';
```
lancewf, line 2: This is a refactor to have the logic of getting the correct icon for an event type and event type display name in one place. This is also adding the new event type of "local_user".

srenatus (approved):
Code looks good to me.

I'm wondering a little about having local users there, but not local teams.

Also, it may be possible that someone is not allowed to view all users, but see the event feed? They'd then get to know stuff they shouldn't. I'm not certain how realistic this concern is, though.

susanev:
by default, only admins are allowed to see users, and all other users see the event feed so i think the situation @srenatus describes will be common.

lancewf:
I don't see a problem adding events for local teams also. 

@susanev Does that mean we don't want events for local users then?

susanev:
@lancewf 🤷 im not sure, is there anyway to make them only appear if for example the user has permission on `iam:users:list` or similar?

lancewf:
@susanev I don't know an easy way to filter them on those permissions.

tylercloke:
Natalie and the team thinks information around local users should be admin only. Since event feed is all or nothing permissions wise, this is probably not a good thing to add for now if we want that info to be admin only. Long-term, we could consider making event feed function different to filter based on some AuthZ constraints or we could have an admin event feed endpoint that has events like these that we want to be restricted.

susanev:
to clarify one point there, that it should match with the users permissions not necessarily be admin only.

lancewf:
Thank you for considering this. I will close this PR for now.

PR #4270, opened 2020-08-26
Adjust RAM settings for Elasticsearch when changing machines docs

mattray:
If you happen to restore to a machine with less memory, your restore will fail unless you adjust Elasticsearch's settings

Signed-off-by: Matt Ray <github@mattray.dev>

### DIFF
--- a/components/docs-chef-io/content/automate/restore.md
+++ b/components/docs-chef-io/content/automate/restore.md
@@ -52,6 +52,13 @@ Ensure access for the backup type used:
          # """
     ```
 
+1. To restore a backup to a machine with less memory than the original system, adjust for the appropriate lower memory settings by adding the following to the configuration file:
+
+    ```toml
+    [elasticsearch.v1.sys.runtime]
+      heapsize = 4096m
+    ```
+
 ## Restore From a Filesystem Backup
 
 Meet the required [prerequisites]({{< ref "restore.md#prerequisites" >}}) before beginning your restore process.
@@ -193,11 +200,6 @@ Set the log level to `debug` before re-running a failed restore to output debug 
 chef-automate debug set-log-level deployment-service debug
 ```
 
-If you are restoring to a machine with less memory than the original system, you my need to add the following to your config patch (adjusted for your appropriate memory settings):
-```
-[elasticsearch.v1.sys.runtime]
-heapsize = 4096m
-```
 ## References
 
 See the [`chef-automate backup restore` command reference]({{< ref "cli_chef_automate/#chef-automate-backup-restore" >}}).
--- a/components/docs-chef-io/content/automate/restore.md
+++ b/components/docs-chef-io/content/automate/restore.md
@@ -193,6 +193,11 @@ Set the log level to `debug` before re-running a failed restore to output debug 
 chef-automate debug set-log-level deployment-service debug
 ```
 
+If you are restoring to a machine with less memory than the original system, you my need to add the following to your config patch (adjusted for your appropriate memory settings):
+```
+[elasticsearch.v1.sys.runtime]
+heapsize = 4096m
+```
 ## References
 
 See the [`chef-automate backup restore` command reference]({{< ref "cli_chef_automate/#chef-automate-backup-restore" >}}).
--- a/components/docs-chef-io/content/automate/restore.md
+++ b/components/docs-chef-io/content/automate/restore.md
@@ -52,12 +52,13 @@ Ensure access for the backup type used:
          # """
     ```
 
-1. To restore a backup to a machine with less memory than the original system, adjust for the appropriate lower memory settings by adding the following to the configuration file:
+1. To restore a backup to a machine with less memory than the original system, adjust for the appropriate lower memory settings by creating a `patch.toml` file that specifies the heapsize, and providing the file at restore time:
 
-    ```toml
-    [elasticsearch.v1.sys.runtime]
-      heapsize = 4096m
-    ```
+     ```toml
+       [elasticsearch.v1.sys.runtime]
+         heapsize = "4096m"
+         # Use "m" for megabytes and "g" for gigabytes
+     ```
 
 ## Restore From a Filesystem Backup
 


stevendanna (approved):
This seems reasonable to me. 

One the code side we could clamp the user-provided value to some maximum value that makes sense for the machine?

components/docs-chef-io/content/automate/restore.md
```diff
@@ -193,6 +193,11 @@ Set the log level to `debug` before re-running a failed restore to output debug
 chef-automate debug set-log-level deployment-service debug
 ```
 
+If you are restoring to a machine with less memory than the original system, you my need to add the following to your config patch (adjusted for your appropriate memory settings):
```
yzl, line 196: my -> may

yzl, line 196:
```suggestion
If you are restoring to a machine with less memory than the original system, you may need to add the following to your config patch (adjusted for your appropriate memory settings):
```
"my" in the original change should be "may"

yzl: Would it be worth putting this in the Prerequisites section on the Restore page, rather than in the Troubleshooting section?

mattray: I'm happy to have it recorded anywhere, as long as it's documented.

yzl: @mattray This shouldn't block merge; I'm happy getting this in as-is and then moving the text later.

components/docs-chef-io/content/automate/restore.md
```diff
@@ -52,6 +52,13 @@ Ensure access for the backup type used:
          # """
     ```
 
+1. To restore a backup to a machine with less memory than the original system, adjust for the appropriate lower memory settings by adding the following to the configuration file:
```
mjingle, line 55: @mattray @yzl I incorporated this content into the Restore Prerequisites section as y'all previously discussed. Does this information need to follow the "create a `patch.toml` file instructions on line 33 rather than a direct change to the `config.toml` file? a la
> create a `patch.toml` file that specifies the lower memory amount and provide it at restore time:

yzl: @mjingle Thanks for doing this. What I would suggest is rewriting the item that starts "To restore a backup to a host with a different fully-qualified domain name..." to something like "Patch the configuration at restore time if the restore host has a different FQDN and/or less memory than the host on which the backup was taken," and then provide the config for FQDN and the config for memory as sub-items.

components/docs-chef-io/content/automate/restore.md
```diff
@@ -52,6 +52,13 @@ Ensure access for the backup type used:
...
+
+    ```toml
+    [elasticsearch.v1.sys.runtime]
+      heapsize = 4096m
```
yzl, line 59: I think you need to specify the units for the heapsize.

mjingle: @mattray Do you have more information on heapsize units used?

components/docs-chef-io/content/automate/restore.md
mjingle: No worries, @yzl. So in my edits, I want to keep each piece of information about FQDN or about lower memory in its separate numbered point for clarity. Also, I'm trying to keep the beginning sentence structure for line 55 the same as the other numbered sentences in this section, so folks can compare the sentences and figure out what applies to their situation in a quicker way. 

What I want to confirm is whether adding the following -

>  [elasticsearch.v1.sys.runtime]
>      heapsize = 4096m

- should be done 1) by editing the configuration file OR 2) by creating and applying a `patch.toml` file (like it is for FQDN).

yzl: @mjingle They should do 2.

mjingle: @yzl That's what I suspected! Thank you for confirming. I'll be sure to update accordingly then.

components/docs-chef-io/content/automate/restore.md
yzl: @mjingle @mattray O sorry -- I should have looked more closely. The `m` is showing up in red for me both in the diff view, and I thought that meant it had been deleted. It's also showing up red in the [full file view](https://github.com/chef/automate/pull/4270/files?short_path=27930b7#diff-27930b7e654e98bda569c8f108979683) so I'm not sure what's going on.

mjingle: No worries. Usually if a character shows up with a red text background, it is caused by the error-highlighting feature of GitHub's text editor/viewer.

yzl: @mjingle Aha -- I think it needs to be `heapsize ="4096m"` with double quotes.

mjingle: @yzl 😄  Great eye! Thanks! I'll make that change also. 

@mattray If you would like, we can also include info on heapsize units. :)

mattray: I'm having a hard time tracking down an authoritative source but it's all over the elasticsearch and JVM documentation. Putting "4096m" will be megabytes and "4g" will be gigabytes.

mjingle: @mattray Understood. Thank you!

components/docs-chef-io/content/automate/restore.md
```diff
@@ -52,6 +52,14 @@ Ensure access for the backup type used:
          # """
     ```
 
+1. To restore a backup to a machine with less memory than the original system, adjust for the appropriate lower memory settings by creating a `patch.toml` file that specifies the heapsize, and providing the file at restore time:
```
yzl, line 55: We might want to say something about "creating a patch.toml or adding to an existing one", lest someone think they must (or can) submit a toml file for every config change they want to make prior to restore.

PR #4271, opened 2020-08-26
data-feed-service: move from port 14001 to 10501

stevendanna:
Any port over 13000 is problematic because of our recommended
ephemeral port range for our A2 HA customers:

```
cat /proc/sys/net/ipv4/ip_local_port_range
10300	65024
```

Thus, any port over 10300 might be in use as an ephemeral port and not
available at service start-up or service restart.

Port 10501 was the first port returned by

```
chef-automate dev ports free
```


stevendanna:
cc @thomascate

thomascate:
Perfect, this solves the issue we had with it colliding with ephemeral ports used by other services.

PR #4272, opened 2020-08-26
ci: fix ha cluster test failures

stevendanna:
The argument parsing in Habitat has changed such that the command we
were using to create the database user actually displayed the
help (because of our use of the `-h` flag intended to be passed to
`psql` but now parsed by hab).  The help output returns an exit code
of 0, meaning that the user went uncreated without an error being
returned by our setup script.


srenatus (approved):
yay

Issue #3916, opened 2020-06-11
Long names in dropdown components should be wrapped


msorens:
## Describe the bug
Long values in dropdowns should be truncated with an ellipsis, and with the full value in a tooltip, instead of just running to the edge. As shown, this will occur wherever the dropdown is used, both in a details page and in a create modal.

![image](https://user-images.githubusercontent.com/6817500/84346344-a0bc5d80-ab64-11ea-9719-b1bfe8197602.png)

Note that the fix should apply to the parent dropdown component, ResourceDropdownComponent, so that all dropdowns are corrected.

## To Reproduce
 - Navigate to Settings >> Projects >> Create Project.
 - Create a project with a long name.
 - Navigate to Settings >> Teams >> Create Team
 - Open the Projects dropdown.

## Expected behavior
Long name should wrap ~appear properly truncated with an ellipsis~.

susanev:
these should wrap not truncate

PR #4275, opened 2020-08-27
Changes to fix long names in drop-down components should be wrapped

vinay033:

Long values in dropdowns should be wrapped, this will occur wherever the dropdown is used, both in a details page and in a create modal.

https://github.com/chef/automate/issues/3916
visit the Settings >> projects/policies
create project or policy - if you have no data then add a project or policy entries with the long name using the create modal.
check the wrapped value in a dropdown - for check this you can go to  Settings >> API tokens then click to `create API token` 
a modal popup will be open and you can see the dropdown values also you can check any API token details page to see a dropdown with long value.
```
STEP 1
inside the hab studio

STEP 2
open new window
go to automate UI path

$ cd components/automate-ui
and run the command 
```

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### :camera: Screenshots
dropdown1
dropdown2

### DIFF
--- a/components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.scss
+++ b/components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.scss
@@ -124,5 +124,5 @@ chef-dropdown {
 }
 
 ::ng-deep chef-checkbox .label-wrap {
-  overflow-wrap: anywhere;
+  word-break: break-all;
 }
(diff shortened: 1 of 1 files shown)


components/automate-ui/src/app/components/resource-dropdown/resource-dropdown.component.scss
```diff
@@ -122,3 +122,7 @@ chef-dropdown {
   font-size: 12px;
   line-height: 1.6;
 }
+
+::ng-deep chef-checkbox .label-wrap {
+  overflow-wrap: anywhere;
```
scottopherson, line 127: Safari still doesn't [fully support](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap#Browser_compatibility) `overflow-wrap`



So we'll probably need to settle for: `word-break: break-all`

vinay033: I have fixed this.

PR #4277, opened 2020-08-27
Move errorutils to the API

danielsdeleo:

Our compiled golang protos are not usable outside of automate because all of automate is one golang module with broken dependencies (dex won't install for some reason related to go modules). Fixing the dependency would still leave a situation where external consumers of the API would need to install all of Automate's deps just to get some structs. Making the external API a separate module will result in much better behavior.

In order for this approach to deliver the desired benefit, the external API must have no dependencies on the rest of Automate. You can run `git grep -h -o 'chef/automate\S\+' | grep -v chef/automate/api | sort | uniq` from the api/external directory to find the golang imports we need to fix. Prior to this PR, we get the following:

```
chef/automate/api/interservice/data_lifecycle
chef/automate/components/automate-gateway/api/iam/v2/policy
chef/automate/components/automate-grpc/protoc-gen-policy/iam
chef/automate/lib/errorutils
```

This PR moves `chef/automate/lib/errorutils` into the external API to make the dependency go in the direction we want (i.e., automate components dep on the API but not the reverse).

#4273 #4278

### :+1: Definition of Done

Code should work the same.  If you run `git grep -h -o 'chef/automate\S\+' -- api/external | grep -v chef/automate/api/external | sort | uniq` from the repo root, `chef/automate/lib/errorutils` does not appear.


stevendanna (approved):
Thanks!

PR #4278, opened 2020-08-27
Remove proto dep external->interservice

danielsdeleo:

Our compiled golang protos are not usable outside of automate because all of automate is one golang module with broken dependencies (dex won't install for some reason related to go modules). Fixing the dependency would still leave a situation where external consumers of the API would need to install all of Automate's deps just to get some structs. Making the external API a separate module will result in much better behavior.

In order for this approach to deliver the desired benefit, the external API must have no dependencies on the rest of Automate. You can run `git grep -h -o 'chef/automate\S\+' | grep -v chef/automate/api | sort | uniq` from the api/external directory to find the golang imports we need to fix. Prior to this PR, we get the following:

```
chef/automate/api/interservice/data_lifecycle
chef/automate/components/automate-gateway/api/iam/v2/policy
chef/automate/components/automate-grpc/protoc-gen-policy/iam
chef/automate/lib/errorutils
```

In this PR, I moved some message definitions from the inter service data lifecycle proto to the external API proto. This is a breaking change at the generated code level. In the go code for example, the structs are now in a different package. That said, the code as it is prior to this PR is "pre-broken" as it uses code with no compatibility guarantees (api/interservice protos) inside code that is supposed to have compatibility guarantees (api/external). Compatibility should be maintained at the protobuf wire level and the protobuf-as-JSON level, which, combined with the current brokenness of using our compiled golang protos, makes it unlikely we will break any customers. 

#4273 #4277 

### :+1: Definition of Done

Code should work the same, except for the compatibility caveat described above. If you run `it grep -h -o 'chef/automate\S\+' -- api/external | grep -v chef/automate/api/external | sort | uniq` from the repo root, you no longer see the data lifecycle interservice protos in the output.

- [ ] Tests added/updated?
- [ ] Docs added/updated?


stevendanna:
> dex won't install for some reason related to go modules

If the fundamental issue is still:

https://github.com/dexidp/dex/issues/1578

Then we might consider the pinning strategy outlined in the comments there rather than our replace directive. I'm not sure if that would resolve the dependency issue or not though.

srenatus:
Was that not resolved with https://github.com/dexidp/dex/pull/1750? I haven't kept close track, but I thought so... 🤔

stevendanna:
@srenatus Aha! I missed that.  At the very least it is worth trying to get rid of our replace directive

srenatus (approved):

PR #4284, opened 2020-08-31
fix esIndex computation for exactly one fully year selection

rickmarry:
There was an edge case in our esIndex computation where, if we had exactly one full year (nothing more, nothing less), we were computing the esIndices to be queried to be an empty string, which would yield not search results.

this pr fixes that.

### :+1: Definition of Done
All original tests pass and new test that tests one full year also passes

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/compliance-service/reporting/relaxting/indices_test.go
+++ b/components/compliance-service/reporting/relaxting/indices_test.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/chef/automate/components/compliance-service/reporting/relaxting"
+	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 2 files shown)


components/compliance-service/reporting/relaxting/indices_test.go
```diff
@@ -352,3 +353,17 @@ func TestIndexDatesNonLeapYear(t *testing.T) {
...
+
+func TestIndexDatesSingleFullYearRightUpToFinal59Seconds(t *testing.T) {
+	// Let debug logs show up when running tests
+	logrus.SetLevel(logrus.DebugLevel)
+	startTime := "2018-01-01T00:00:00Z"
+	endTime := "2018-12-31T23:59:59Z"
+	esIndex, err := relaxting.IndexDates(relaxting.CompDailySumIndexPrefix, startTime, endTime)
+
+	if assert.NoError(t, err) {
+		assert.Equal(t, fmt.Sprintf("%[1]s2018.01*,%[1]s2018.02*,%[1]s2018.03*,%[1]s2018.04*,%[1]s2018.05*,"+
```
alexpop, line 365: Thank you for the fix Rick!
Should this return just `"%s2018.*"` ?

rickmarry: I could have but because of the way that i enter here with this new condition, it returns the year in the form of months.. i can look deepr and get it to return the whole year instead as you mention but functionally, it would make no diff, so i didn't spend any more time

alexpop: Ok, we can leave it like this for now. 👍

alexpop (approved):
🚀

PR #4287, opened 2020-09-01
Move gateway IAM API to external API in the the tree

danielsdeleo:

In prior PRs (linked below) we moved code that is depended upon by the external API but is outside of that portion of the tree, into api/external, with the goal of having no dependencies from api/external to outside parts of Automate. This will allow us to split api/external into a separate golang module, which outside consumers can use to integrate with Automate's API while only needing to install minimal dependencies.

This PR moves the contents of `automate-gateway/api/iam` into api/external, which should be the last part we need to move.

#4273 #4277 #4278

### :+1: Definition of Done

After this PR the command ` git grep -h -o 'chef/automate\S\+' -- api/external | grep -v chef/automate/api/external ` returns no results, indicating there are no dependencies from api/external to outside code in automate.


srenatus (approved):
Thank you

PR #4293, opened 2020-09-02
Updating elastic search to 7.9.0

lancewf:

Looking into updating Elasticsearch to 7.9.0

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-elasticsearch/habitat/plan.sh
+++ b/components/automate-elasticsearch/habitat/plan.sh
@@ -4,12 +4,12 @@
 pkg_name="automate-elasticsearch"
 pkg_description="Wrapper package for core/elasticsearch"
 pkg_origin="chef"
-pkg_version="6.8.3"
+pkg_version="7.9.0"
 pkg_maintainer="Chef Software Inc. <support@chef.io>"
 pkg_license=("Chef-MLSA")
 pkg_upstream_url="https://www.chef.io/automate"
-pkg_source="https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-${pkg_version}.tar.gz"
-pkg_shasum=824078e421c9f7e5ab9c875e4019d9ebfe3ada99db286b54dec090f97d1cbe25
+pkg_source="https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-${pkg_version}-linux-x86_64.tar.gz"
+pkg_shasum=9f8ce7308cd0d0c691b0bd5b43716cf58e40f2cd54a7f4c36e9188aca199a379
 
 pkg_build_deps=(
   core/patchelf
@@ -53,7 +53,7 @@ do_build() {
 
 do_install() {
   cd "$HAB_CACHE_SRC_PATH/elasticsearch-${pkg_version}"
-  install -vDm644 README.textile "${pkg_prefix}/README.textile"
+  install -vDm644 README.asciidoc "${pkg_prefix}/README.asciidoc"
   install -vDm644 LICENSE.txt "${pkg_prefix}/LICENSE.txt"
   install -vDm644 NOTICE.txt "${pkg_prefix}/NOTICE.txt"
 
@@ -66,13 +66,10 @@ do_install() {
   # mkdir -p "$pkg_prefix/es/config"
   # install -vDm644 config/jvm.options "$pkg_prefix/es/config/jvm.options"
 
-  # Delete unused binaries to save space
-  rm "${pkg_prefix}/es/bin/"*.bat "${pkg_prefix}/es/bin/"*.exe
-
   LD_RUN_PATH=$LD_RUN_PATH:${pkg_prefix}/es/modules/x-pack-ml/platform/linux-x86_64/lib
   export LD_RUN_PATH
 
-  for bin in autoconfig autodetect categorize controller normalize; do
+  for bin in data_frame_analyzer autodetect categorize controller normalize; do
     build_line "patch ${pkg_prefix}/es/modules/x-pack-ml/platform/linux-x86_64/bin/${bin}"
     patchelf --interpreter "$(pkg_path_for glibc)/lib/ld-linux-x86-64.so.2" --set-rpath "${LD_RUN_PATH}" \
       "${pkg_prefix}/es/modules/x-pack-ml/platform/linux-x86_64/bin/${bin}"


lancewf:
[2020-09-02T17:44:21+00:00] automate-elasticsearch.default(O): [2020-09-02T17:44:21,438][WARN ][o.e.d.c.s.Settings       ] [solo] [node.data] setting was deprecated in Elasticsearch and will be removed in a future release! See the breaking changes documentation for the next major version.
[2020-09-02T17:44:21+00:00] automate-elasticsearch.default(O): [2020-09-02T17:44:21,446][WARN ][o.e.d.c.s.Settings       ] [solo] [node.max_local_storage_nodes] setting was deprecated in Elasticsearch and will be removed in a future release! See the breaking changes documentation for the next major version.
[2020-09-02T17:44:21+00:00] automate-elasticsearch.default(O): [2020-09-02T17:44:21,460][INFO ][o.e.e.NodeEnvironment    ] [solo] using [2] data paths, mounts [[/ (overlay)]], net usable_space [42.1gb], net total_space [58.4gb], types [overlay]
[2020-09-02T17:44:21+00:00] automate-elasticsearch.default(O): [2020-09-02T17:44:21,460][INFO ][o.e.e.NodeEnvironment    ] [solo] heap size [1.9gb], compressed ordinary object pointers [true]
[2020-09-02T17:44:21+00:00] automate-elasticsearch.default(O): [2020-09-02T17:44:21,462][WARN ][o.e.d.c.s.Settings       ] [solo] [node.master] setting was deprecated in Elasticsearch and will be removed in a future release! See the breaking changes documentation for the next major version.
[2020-09-02T17:44:21+00:00] automate-elasticsearch.default(O): [2020-09-02T17:44:21,551][INFO ][o.e.n.Node               ] [solo] node name [solo], node ID [oJzoYfZQQzajtfgcX6EEeg], cluster name [chef-insights]
[2020-09-02T17:44:23+00:00] automate-elasticsearch.default(O): [2020-09-02T17:44:23,778][WARN ][o.e.d.c.s.Settings       ] [solo] [discovery.zen.ping.unicast.hosts] setting was deprecated in Elasticsearch and will be removed in a future release! See the breaking changes documentation for the next major version.
[2020-09-02T17:44:23+00:00] automate-elasticsearch.default(O): [2020-09-02T17:44:23,825][WARN ][o.e.d.c.s.Settings       ] [solo] [discovery.zen.minimum_master_nodes] setting was deprecated in Elasticsearch and will be removed in a future release! See the breaking changes documentation for the next major version.
[2020-09-02T17:44:23+00:00] automate-elasticsearch.default(O): [2020-09-02T17:44:23,826][WARN ][o.e.d.c.s.Settings       ] [solo] [gateway.expected_nodes] setting was deprecated in Elasticsearch and will be removed in a future release! See the breaking changes documentation for the next major version.
[2020-09-02T17:44:23+00:00] automate-elasticsearch.default(O): [2020-09-02T17:44:23,827][WARN ][o.e.d.c.s.Settings       ] [solo] [discovery.zen.fd.ping_timeout] setting was deprecated in Elasticsearch and will be removed in a future release! See the breaking changes documentation for the next major version.
[2020-09-02T17:44:23+00:00] automate-elasticsearch.default(O): [2020-09-02T17:44:23,828][WARN ][o.e.d.c.s.Settings       ] [solo] [gateway.expected_master_nodes] setting was deprecated in Elasticsearch and will be removed in a future release! See the breaking changes documentation for the next major version.

lancewf:
Event-feed-service has a problem - [2020-09-02T17:46:49+00:00] event-feed-service.default(O): time="2020-09-02T17:46:49Z" level=error msg="initializing elasticsearch: creating index eventfeed-2-feeds: elastic: Error 400 (Bad Request): The mapping definition cannot be nested under a type [_doc] unless include_type_name is set to true. [type=illegal_argument_exception]"

components/automate-cs-oc-erchef/cmd/init-es/init-es.go
```diff
@@ -10,8 +10,8 @@ import (
 	"os"
 	"time"
 
+	elastic "github.com/olivere/elastic/v7"
```
lancewf, line 13: This needs tested

components/automate-elasticsearch/habitat/config/elasticsearch.yml
```diff
@@ -1,5 +1,7 @@
 cluster.name: {{cfg.cluster.name}}
 
+cluster.initial_master_nodes: {{cfg.node.name}}
```
lancewf, line 3: Needed to start elasticsearch. A master will not be chosen if this is not set.

components/compliance-service/ingest/ingestic/ingestic.go
```diff
@@ -103,7 +103,7 @@ func (backend *ESClient) createStore(ctx context.Context, indexName string, mapp
 
 // ProfileExists returns true if profile exists already in ES.. false if not
 func (backend *ESClient) ProfileExists(hash string) (bool, error) {
-	idsQuery := elastic.NewIdsQuery(mappings.DocType)
+	idsQuery := elastic.NewIdsQuery()
```
lancewf, line 106: The elastic.NewIdsQuery function needs to be tested without a type in them.

components/event-feed-service/pkg/persistence/mappings.go
```diff
@@ -159,12 +158,10 @@ var feedProps = `
 // Feeds mapping used to create the feeds-[version] index
 var Feeds = Mapping{
 	Index:      IndexNameFeeds,
-	Type:       DocType,
 	Timeseries: false,
 	Properties: fmt.Sprintf(`{ %s }`, feedProps),
 	Mapping: fmt.Sprintf(`
 			{
-				"template":"`+IndexNameFeeds+`",
```
thomascate, line 167: In ES 7.X Elasticsearch changed from 5 shards to 1 shard as the default. This doesn't effect standalone, but could cause some issues on clustered deploys.

To get the previous behavior, we'll need to add "shards": 5, to all of these templates.

lancewf: Thank you, I will add that.

components/compliance-service/reporting/relaxting/profiles.go
```diff
@@ -263,7 +262,7 @@ func (backend *ES2Backend) GetProfile(hash string) (reportingapi.Profile, error)
 		return profile, errors.Wrap(err, "GetProfile, cannot connect to ElasticSearch")
 	}
 
-	idsQuery := elastic.NewIdsQuery(mappings.DocType)
+	idsQuery := elastic.NewIdsQuery()
```
lancewf, line 265: Need to ensure this change works correctly

components/compliance-service/reporting/relaxting/util.go
```diff
@@ -66,7 +66,7 @@ func (backend *ES2Backend) ES2Client() (*elastic.Client, error) {
 	//this is now a singleton as per best practice as outlined in the comment section of the elastic.NewClient
 	once.Do(func() {
 		esClient, err = elastic.NewClient(
-			elastic.SetHttpClient(backend.getHttpClient()),
+			// elastic.SetHttpClient(backend.getHttpClient()),
```
lancewf, line 69: Is this needed? What does this do?

components/event-feed-service/pkg/persistence/feed-elastic.go
```diff
@@ -313,7 +313,7 @@ func newRangeQuery(start string, end string, fieldTime string) (*olivere.RangeQu
 	var ok = false
 
 	rangeQuery := olivere.NewRangeQuery(fieldTime).
-		Format("yyyy-MM-dd||yyyy-MM-dd-HH:mm:ss||yyyy-MM-dd'T'HH:mm:ssZ")
+		Format("yyyy-MM-dd||yyyy-MM-dd-HH:mm:ss||yyyy-MM-dd'T'HH:mm:ssX||yyyy-MM-dd'T'HH:mm:ssXXX")
```
lancewf, line 316: Time formats with timezones need updated https://www.elastic.co/guide/en/elasticsearch/reference/current/migrate-to-java-time.html#java-time-migration-incompatible-date-formats.

"yyyy-MM-dd'T'HH:mm:ssZ" need to be changed to "yyyy-MM-dd'T'HH:mm:ssX" for times like "2017-09-01-07:58:06Z". But for time like "2010-01-01T01:02:03-07:00" we need to add more X like "yyyy-MM-dd'T'HH:mm:ssXXX"

lancewf:
License scout
```

+--------+---------------------------------------------------------------------------+------------+--------------+
--
  | \| Type   \| Dependency                                                                \| License(s) \| Results      \|
  | +--------+---------------------------------------------------------------------------+------------+--------------+
  | \| golang \| github.com/golang/groupcache (v0.0.0-20190702054246-869f871628b6)         \|            \| Missing      \|
  | \|        \| github.com/smartystreets/go-aws-auth (v0.0.0-20180515143844-0c1422d1fdb9) \|            \| Undetermined \|
  | \|        \| github.com/smartystreets/gunit (v1.4.2)                                   \|            \| Undetermined \|
  | +--------+---------------------------------------------------------------------------+------------+--------------+

```

components/ingest-service/test/chef_action.js
```diff
@@ -58,7 +58,7 @@ describe("/events/chef/action", function () {
           return helpers.ActionRefresh().then(function(response) {
             return chakram.get(actionsUrl()).then(function(response) {
               expect(response).to.have.status(200);
-              expect(response.body.hits.total).to.equal(0);
+              expect(response.body.hits.total.value).to.equal(0);
```
lancewf, line 61: Elasticsearch 7 response changed to return an object for the total field with the value inside.

components/ingest-service/test/liveness.js
```diff
@@ -27,6 +27,7 @@ describe("/events/chef/liveness", function () {
   });
 
   describe("posting example message", function () {
+    before(function () { return helpers.ESCleanupDocuments() });
```
lancewf, line 30: I am not sure why this is needed now, but without this line there is a node left in the node-state index with the id "test-id".

thomascate:
A new setting in 7.X that may cause issues for our users is cluster.max_shards_per_node, this defaults to 1,000. In standalone we have been creating 5 shards per index. This setting would lock index creation at 200 days of node history. If they're using compliance and logging Chef Actions as well, that drops to 50 days.

We can increase this default, though a preflight upgrade check that polls _cat/shards and reports when it returns greater than 1k shards per node would be super helpful.

lancewf:
@thomascate Thank you for bringing this up. I will bring this up to the team today. Do you see any problem with setting the cluster.max_shards_per_node to say 8,000 or more allow over a year of indexes?

thomascate:
The only issue is that once shards gets that large performance usually tanks. So it might be worth adding a tunable health check to es-sidecar to check how many shards they have, with a default alert at 1,000.

lancewf:
@thomascate That is a good idea to add the tunable health check to es-sidecar. We are going to have to update the `cluster.max_shards_per_node` field to a large value because some compliance users are not deleting any historical reports. 

I will create an issue to add this health check.

components/compliance-service/reporting/relaxting/suggestions.go
```diff
@@ -539,7 +539,7 @@ func (backend ES2Backend) getControlSuggestions(ctx context.Context, client *ela
 	var innerQuery elastic.Query
 	if len(text) >= 2 {
 		innerBoolQuery := elastic.NewBoolQuery()
-		innerBoolQuery.Must(elastic.NewMatchQuery(fmt.Sprintf("%s.engram", target), text).Operator("or"))
+		innerBoolQuery.Must(elastic.NewMatchQuery(fmt.Sprintf("%s.engram", target), text).Operator("and"))
```
lancewf, line 542: This change does make the tests pass, but I am not sure if it is correct. The problem is with ES7 the "or" is not filtering enough values.

lancewf:
With the Elasticsearch 7 upgrade, we have found a problem with the ListSuggestion RPC in the compliance-service only when requesting type “controls”. We are seeing that the controls returned are not the closest match to the text.  The specific problem is the filter aggregation has a “or” match that is behaving differently than ES6. In ES6 the “or” match removed the controls that were not a close match. 
We also found an existing problem with the ES query that the controls found are not sorted by the closest match. The controls are sorted by the number of controls with the same name. 

Below is a link to the ES7 and ES6 raw queries and results. 
https://gist.github.com/lancewf/be7bb66bce4c57f22871bc377b4db4e1

Below is a link to the go code that creates the query. 
https://github.com/chef/automate/blob/master/components/compliance-service/reporting/relaxting/suggestions.go#L542

lancewf:
### Circuit-breaker Elasticsearch 7
After looking into the new circuit-breaker Elasticsearch 7 feature, I don't recommend any different configuration than the default. This feature is preventing heap overflow exceptions. I have been able to cause this exception in the hab studio environment by running `chef_load_compliance_nodes 100` and then trying any call that pulls data from Elasticsearch. By waiting for about 30 seconds, the exception will go away. I think this is because the Java garbage collection has run which reduces the amount of running memory. This temporary exception will notify the client that their system needs more resources without leaving Elasticsearch in a bad state.  This new error will have to be provided to support so that they will more easily recognize the problem.  
```
{
  "error" : {
    "root_cause" : [
      {
        "type" : "circuit_breaking_exception",
        "reason" : "[parent] Data too large, data for [<http_request>] would be [1582945680/1.4gb], which is larger than the limit of [1478832947/1.3gb], real usage: [1582945680/1.4gb], new bytes reserved: [0/0b], usages [request=0/0b, fielddata=0/0b, in_flight_requests=0/0b, model_inference=0/0b, accounting=1601192/1.5mb]",
        "bytes_wanted" : 1582945680,
        "bytes_limit" : 1478832947,
        "durability" : "PERMANENT"
      }
    ],
    "type" : "circuit_breaking_exception",
    "reason" : "[parent] Data too large, data for [<http_request>] would be [1582945680/1.4gb], which is larger than the limit of [1478832947/1.3gb], real usage: [1582945680/1.4gb], new bytes reserved: [0/0b], usages [request=0/0b, fielddata=0/0b, in_flight_requests=0/0b, model_inference=0/0b, accounting=1601192/1.5mb]",
    "bytes_wanted" : 1582945680,
    "bytes_limit" : 1478832947,
    "durability" : "PERMANENT"
  },
  "status" : 429
}
```

lancewf:
The below terms are Elasticsearch configs that have been deprecated. Below is the warning messing from the logs. 
"setting was deprecated in Elasticsearch and will be removed in a future release! See the breaking changes documentation for the next major version."
- node.data
- node.max_local_storage_nodes
- node.master
- discovery.zen.ping.unicast.hosts
- discovery.zen.minimum_master_nodes
- gateway.expected_nodes
- discovery.zen.fd.ping_timeout
- gateway.expected_master_nodes

lancewf:
I am seeing the below log when starting the automate-elasticsearch service. 
```
[2020-09-25T16:16:39+00:00] automate-elasticsearch.default(O): [2020-09-25T16:16:39,810][WARN ][c.a.a.p.i.BasicProfileConfigFileLoader] [automate] Unable to load config file null
[2020-09-25T16:16:39+00:00] automate-elasticsearch.default(O): java.security.AccessControlException: access denied ("java.io.FilePermission" "/.aws/config" "read")
```
I am not sure why elasticsearch is trying to get the AWS creds.

lancewf:
I tested upgrading to ES7 from the latest dev build on a performance server and it took less than 2m30.487s. The performance box is based on the https://github.com/chef/a2-loadtest setup. Automate was running on an m4.xlarge with 4gb Elasticsearch heap size. With a separate chef load server sending 20,000 nodes with updates every 15 minutes. This equals 1,333 CCRs every minute.

lancewf:
End of Life for the current version of Elasticsearch 6.8 is November 20th, 2020 https://www.elastic.co/support/eol

lancewf:
Confirmed that ES6 indexes are writable under ES7. Therefore we do not need to do a reindex of all indexes.

PR #4298, opened 2020-09-02
Make external API a separate module

danielsdeleo:

Make the api/external directory a separate golang module. When using golang modules, when you depend on any package within a module, you depend on that module and therefore also all of that module's dependencies. Currently all of Automate is one big module which is nice for our internal dependency management but for external users it means you have to take on a lot of deps even if you only intend to use a few structs from api/external directory.

At present the goal is to support a project that is integrating some Chef Workstation functionality with Chef Automate, but this would be helpful to anyone who wishes to write code using Automate's API in go.

- [ ] Tests added/updated?
- [ ] Docs added/updated?


stevendanna:
I wonder if this is a time to re-look at our vendoring in this repository.  Having to re-vendor our own code anytime we make a change to it feels like it will trip people up regularly.

srenatus:
@stevendanna you're suggesting that we stop vendoring, and rely completely on go modules instead? It has worked fine for the other project, I'd say. 👍

.bldr.toml
```diff
@@ -439,6 +439,7 @@ paths = [
   "vendor/github.com/grpc-ecosystem/grpc-gateway/*",
   "vendor/github.com/grpc-ecosystem/grpc-opentracing/*",
   "vendor/github.com/hashicorp/hcl/*",
+  "vendor/github.com/josharian/intern/*",
```
srenatus, line 442: 🤔 I wonder how this got dragged in

stevendanna:
>  you're suggesting that we stop vendoring, and rely completely on go modules instead?

I think we should consider it, yes.  We started vendoring when using `dep` which was substantially slower without vendoring.

In my views, if we stop vendoring, the cons are:

- Updating dependencies no longer provides an easy way to review what is changing in those dependencies.
- We'll likely have to give up on rebuilding all affected components when a dependency changes. This means occasionally we could have build-breaks caused by a previous dependency update. (we might be able to do something here with changes to our existing tooling).
- More network resources required during the build (likely can be mitigated with some caching)

The pros are:

- Fewer situations where people need to regenerate bldr.toml.
- Likely makes it easier to split out more modules over time if we care to.
- Less work when updating a dependency

srenatus:
> Updating dependencies no longer provides an easy way to review what is changing in those dependencies.

Yeah, it's still in the `go.{sum,mod}` files, but you won't see the code right away.

I agree with your pros and cons there, thanks for summing this up. :)

danielsdeleo:
Update: in #4314 we have removed vendoring, so now this change is mostly transparent for an Automate developer. I checked that you can update a proto in api/external, compile it, and then use it right away (no need to revendor or git commit or anything).  Updating deps that appear in both modules may be confusing, and I'm not sure what I could do about that (or rather, I would like to have some practical experience of the problem before attempting to fix). So I think this is ready to go now.

go.mod
```diff
@@ -21,6 +23,7 @@ require (
 	github.com/bufbuild/buf v0.20.5
 	github.com/buger/goterm v0.0.0-20180307092342-c9def0117b24
 	github.com/buger/jsonparser v0.0.0-20180808090653-f4dd9f5a6b44
+	github.com/chef/automate/api/external v0.0.0-00010101000000-000000000000
```
lancewf, line 26: Does this line need updated when changing the external protobufs?

lancewf: Is it going to be a two PR process to add an RPC to the external protobufs? One to check in the changes to the external protobufs then the second to update this line and add the code to the automate-gateway.

danielsdeleo: Nope, that's why we ripped out the vendoring. The `replace github.com/chef/automate/api/external => ./api/external/` later in this file makes imports with `github.com/chef/automate/api/external` look in the local dir, and since we are not vendoring, that's enough to make it work. I explicitly tested by making changes to the applications service protos, compiling them and then referring to the updated stuff in the application service code. There's no need to commit or revendor or anything like that.

lancewf: I am trying the same kind of test that you mentioned. I added a new RPC to the external config-mgmt files. Then ran `compile_all_protobuf_components_nolint`. Then when building automate-gateway `build components/automate-gateway` got the below errors. 
```
y
# github.com/chef/automate/components/automate-gateway/handler
handler/cfgmgmt.go:140:66: undefined: "github.com/chef/automate/api/external/cfgmgmt/request".Errors2
handler/cfgmgmt.go:140:84: undefined: "github.com/chef/automate/api/external/cfgmgmt/response".Errors2
handler/cfgmgmt.go:146:10: undefined: "github.com/chef/automate/api/external/cfgmgmt/response".Errors2
   automate-gateway: Build time: 1m3s
   automate-gateway: Exiting on error
```
Below are the changes I made. 
https://github.com/chef/automate/commit/40dba9516b30081a6938e31b300b5febca272cfb

danielsdeleo: @lancewf Your branch is based on an older version of this PR before I rebased on top of the latest mainline (with the novendor stuff merged). If I cherry pick your commit on top of the latest, there's one small bug in the `return` statement for the `GetErrors2` method, after fixing that, it builds correctly :)

lancewf: Ok thanks, I will try that out.

lancewf: That pulling the latest from this branch did fix the problem I was seeing, Sorry about that. I should have tried to update the code first.

lancewf (approved):
Everything worked the same for me.

PR #4304, opened 2020-09-03
[Auth-4301] Add permission check for CreateRule

msorens:
Add checks for project permissions when creating a rule.

NA

### :+1: Definition of Done

 - Admin should still be able to create rules for any project

With user bob being a `proj1 Project Owner` and a `proj2 ProjectViewer`, plus a member of the `proj-update` policy (see below), then:
 - bob should be able to create rules for proj1 
 - bob should not be able to create rules for proj 2

In hab studio `rebuild components/authz-service`.

Create projects proj1 and proj2, with policies.

Create an admin token.

Create a token t1 and make it a member of `proj1 Project Owners` and `proj2 Project Viewers`.

Setup a policy with `iam:projects:update` for proj1 only and include token t1 as a member--note that this needs the admin token:
```
curl -sSkH "api-token: $ADMINTOK" -d '{ "name": "proj-update", "id": "proj-update", "members": [ 'token:t1' ], "statements": [ { "effect": "ALLOW", "actions": [ "iam:projects:update" ], "projects": [ "proj1" ] } ], "projects": [] }' $TARGET_HOST/apis/iam/v2/policies?pretty
```

Attempt to add a rule to proj1 using the T1 token:
```
curl -sSkH "api-token: $T1TOK" -d "$(jq -n '{ id: "rule8", name: "my foo rule", type: "NODE", project_id: "foo-project", conditions: [{ operator: "EQUALS", attribute: "CHEF_SERVER", values: ["The dogs"]}]}')"  $TARGET_HOST/apis/iam/v2/projects/proj2/rules?pretty 
```

Should now get this error
```
{
  "error": "error creating rule with ID \"rule8\": rpc error: code = NotFound desc = project with ID \"proj2\" not found",
  "code": 13,
  "message": "error creating rule with ID \"rule8\": rpc error: code = NotFound desc = project with ID \"proj2\" not found",
  "details": [
  ]
}
```

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/authz-service/storage/postgres/rule.go
+++ b/components/authz-service/storage/postgres/rule.go
@@ -9,6 +9,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/chef/automate/components/authz-service/storage"
+	"github.com/chef/automate/lib/stringutils"
 )
 
 // These must match what SQL function query_rule_table_associations returns.
@@ -18,25 +19,17 @@ const (
 )
 
 func (p *pg) CreateRule(ctx context.Context, rule *storage.Rule) (*storage.Rule, error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 
 	projectsFilter, err := projectsListFromContext(ctx)
 	if err != nil {
 		return nil, err
 	}
-	found := len(projectsFilter) == 0 // empty means all projects allowed
-	if !found {
-		for _, proj := range projectsFilter {
-			if proj == rule.ProjectID {
-				found = true
-			}
-		}
-	}
-	if !found {
+	allProjectsAlowed := len(projectsFilter) == 0
+	if !allProjectsAlowed && !stringutils.SliceContains(projectsFilter, rule.ProjectID) {
 		return nil, fmt.Errorf("project with ID %q not found", rule.ProjectID)
 	}
-
-	ctx, cancel := context.WithCancel(ctx)
-	defer cancel()
 
 	tx, err := p.db.BeginTx(ctx, nil)
 	if err != nil {
--- a/components/authz-service/storage/postgres/rule.go
+++ b/components/authz-service/storage/postgres/rule.go
@@ -3,6 +3,7 @@ package postgres
 import (
 	"context"
 	"database/sql"
+	"fmt"
 
 	"github.com/lib/pq"
 	"github.com/pkg/errors"
@@ -17,6 +18,21 @@ const (
 )
 
 func (p *pg) CreateRule(ctx context.Context, rule *storage.Rule) (*storage.Rule, error) {
+
+	projectsFilter, err := projectsListFromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+	found := false
+	for _, proj := range projectsFilter {
+		if proj == rule.ProjectID {
+			found = true
+		}
+	}
+	if !found {
+		return nil, fmt.Errorf("project with ID %q not found", rule.ProjectID)
+	}
+
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
--- a/components/authz-service/storage/postgres/rule.go
+++ b/components/authz-service/storage/postgres/rule.go
@@ -23,10 +23,12 @@ func (p *pg) CreateRule(ctx context.Context, rule *storage.Rule) (*storage.Rule,
 	if err != nil {
 		return nil, err
 	}
-	found := false
-	for _, proj := range projectsFilter {
-		if proj == rule.ProjectID {
-			found = true
+	found := len(projectsFilter) == 0 // empty means all projects allowed
+	if !found {
+		for _, proj := range projectsFilter {
+			if proj == rule.ProjectID {
+				found = true
+			}
 		}
 	}
 	if !found {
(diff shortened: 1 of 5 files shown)


inspec/a2-api-integration/controls/iam/iam_apis.rb
```diff
@@ -862,7 +862,8 @@ def staged(rule)
         it "creates a new rule" do
           resp = automate_api_request("/apis/iam/v2/projects/#{custom_project_id}/rules",
             http_method: 'POST',
-            request_body: CUSTOM_RULE.to_json
+            request_body: CUSTOM_RULE.to_json,
+            request_headers: { 'projects': [ custom_project_id ] }.to_json
```
blakestier, line 866: i think 'projects' doesn't need to be a string?

components/authz-service/storage/postgres/rule.go
```diff
@@ -17,6 +18,23 @@ const (
...
+				found = true
+			}
+		}
+	}
+	if !found {
+		return nil, fmt.Errorf("project with ID %q not found", rule.ProjectID)
+	}
+
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
```
srenatus, line 39: [nit] This block should be the first thing happening in this function, I think.

components/authz-service/storage/postgres/rule.go
```diff
@@ -17,6 +18,23 @@ const (
...
+	if err != nil {
+		return nil, err
+	}
+	found := len(projectsFilter) == 0 // empty means all projects allowed
+	if !found {
+		for _, proj := range projectsFilter {
+			if proj == rule.ProjectID {
+				found = true
+			}
+		}
```
srenatus, line 32: [nit] This could use [`stringutils. SliceContains`](https://github.com/chef/automate/blob/83e0e071be9f124f2fba5a4915e7109c5d7da5cd/lib/stringutils/slice.go#L7):
```suggestion
		found = stringutils.SliceContains(projectsFilter, rule.ProjectID)
```

components/authz-service/storage/postgres/rule.go
```diff
@@ -17,6 +18,23 @@ const (
...
+	}
+	if !found {
+		return nil, fmt.Errorf("project with ID %q not found", rule.ProjectID)
+	}
```
srenatus, line 36: Re: previous nit, instead of overwriting `found`, I'd suggest
```suggestion
	if !found && !stringutils.SliceContains(projectsFilter, rule.ProjectID) {
		return nil, fmt.Errorf("project with ID %q not found", rule.ProjectID)
	}
```

srenatus: 🤔 maybe there's a better name than `found`, here, too? `allProjectsAllowed := len(...) == 0` maybe?

srenatus (approved):
Looks good to me. Some comments on the code style inline 👇 🙃

components/authz-service/storage/postgres/rule.go
```diff
@@ -17,6 +18,23 @@ const (
 )
 
 func (p *pg) CreateRule(ctx context.Context, rule *storage.Rule) (*storage.Rule, error) {
+
+	projectsFilter, err := projectsListFromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+	found := len(projectsFilter) == 0 // empty means all projects allowed
```
blakestier, line 26: perhaps we could lose the comment if we rename this var something like `allProjectsAllowed` ?

components/authz-service/storage/postgres/rule.go
```diff
@@ -17,6 +18,23 @@ const (
...
+	if !found {
+		for _, proj := range projectsFilter {
+			if proj == rule.ProjectID {
+				found = true
```
blakestier, line 30: likewise, `found` here could be `projectAllowed` or `hasPermsOnProj`

components/authz-service/storage/postgres/rule.go
blakestier: 😎

components/authz-service/storage/postgres/rule.go
blakestier: lol I've just created a bunch of comments redundant to @srenatus 😂

components/authz-service/storage/postgres/rule_test.go
```diff
@@ -131,6 +131,23 @@ func TestCreateRule(t *testing.T) {
 			assertCount(t, 3, db.QueryRow(
 				`SELECT count(*) FROM iam_staged_rule_conditions WHERE rule_db_id=(SELECT r.db_id FROM iam_staged_project_rules r WHERE r.id=$1)`, ruleID))
 		},
+		"when the project filter does not match, return ErrNotFound": func(t *testing.T) {
```
blakestier, line 134: I wonder, given the confusion yesterday, if we should also have  "if no projects allowed" and "if all projects allowed" test cases?

msorens: Added test for "all". For the "no projects allowed case", it would never get here--that is caught earlier with this error:
```
"unauthorized: members [\"token:t1\"] cannot perform action \"iam:projects:update\" on resource \"iam:projects:proj2\" filtered by projects []",
```

tylercloke (approved):

Issue #4027, opened 2020-07-02
allow user to get data from last 24 hrs (relative time)


vjeffrey:
## User Story
problem: users only see data per utc day, leading to empty dashboard right after we cross utc day time.

related: https://github.com/chef/automate/issues/4015

writeup: https://docs.google.com/document/d/1V8qCAazLoEf2Oz1UTzWXAueMqny1MOXN4tXLecFAA5k/edit#heading=h.jez6nwd19bf5

right now it seems like we're leaning towards doing some work for the user to see the last 24hrs (relative) data. this may include writing to extra indices. let's write up the options, discuss, and fix things. also last 24 hrs might not be the right solution?

vjeffrey:
https://chef.aha.io/epics/SH-E-164

vjeffrey:
doc and api docs changes wip: https://github.com/chef/automate/compare/vj/24hr-change-doc

PR #4310, opened 2020-09-04
Compliance APIs default to 24h without end_time

alexpop:

Compliance backend change to default Automate compliance APIs to showing data for the last 24 hourst as opposed to current UTC day. More details here: https://github.com/chef/automate/issues/4027

If no `start_time` & `end_time` is specified in the filters, we default to the last 24 hours window. For data in the past, `end_time` is always needed now, even for API calls that before didn't require it, like filtering for a `job_id`, suggestions or getting a report by just providing the report id. All POST API calls now that take `filters` will only search the last 24h worth of data unless `end_time` is specified.

It includes some dynamic reports that are being ingested with an end_time relative to current time. This allows us to test the 24h window with the existing tests:

```
$ ls -1 components/compliance-service/test_data/audit_reports/NOW_*
components/compliance-service/test_data/audit_reports/NOW_MINUS_1440_MINUS_0005_ubuntu(0)-alpha-failed.json
components/compliance-service/test_data/audit_reports/NOW_MINUS_1440_PLUS_0001_ubuntu(1)-alpha-myprofile(s)-skipped.json
components/compliance-service/test_data/audit_reports/NOW_MINUS_1440_PLUS_0300_osx(2)-omega-pro1(p)-failed.json
components/compliance-service/test_data/audit_reports/NOW_MINUS_1440_PLUS_0360_osx(2)-omega-pro1(p)-passed.json
components/compliance-service/test_data/audit_reports/NOW_MINUS_1440_PLUS_1380_osx(2)-omega-pro1(p)-passed.json
components/compliance-service/test_data/audit_reports/NOW_MINUS_1440_PLUS_1410_ubuntu(0)-alpha-failed.json
```

- [x] Tests added/updated?
- [x] Docs added/updated?


alexpop:
I managed to all the test to pass, except for the ones in `components/compliance-service/integration_test`

Other than those, there are two performance things to try to do before merging: 

1. In this change here, pull the time of the job from the database and pass it along so we can do the ES query with that as `end_time`, instead of querying all indices: https://github.com/chef/automate/pull/4310/commits/e35a474f211552a6e37d78d9004bd93629867034

2. Look for a way to do a Refresh "false" as we had it before. I change it to true, as the two update calls were overlapping and one was canceling the other. Rick recommended doing a single update call or try to exclude each others based on ID filtering:
https://github.com/chef/automate/blob/ap-rm/24hours/components/compliance-service/ingest/ingestic/ingestic.go#L332

rickmarry:
>     Rick recommended doing a single update call or try to exclude each others based on ID filtering:

I think that we can gather up the run_id for the daily_latest today's index and the last_24 for today and yesterday index for this node_id, once we have those run_uuids, we can set them to false for both flags and do so with refresh set to false.. this will keep us from having the "last update wins" problem that we are seeing with those independent updates

rickmarry (approved):
thank you @alexpop.. this is a great improvement!

alexpop:
While testing the latest UI change pushed by @scottopherson, I noticed an ingestion bug to reports older than 24h. This is due to the `daily_latest` flag staying true for reports of the same node in a UTC day. Investigating a fix now.

PR #4321, opened 2020-09-08
Dont ssl check cert validity period for internal CA

danielsdeleo:

The SSL Tester is now asserting that certs should be valid for less than 398 days. The constraints of our internal CA are much different than for normal web use, so we are disabling this check for the internal CA.


stevendanna (approved):
Thanks for jumping on this

PR #4314, opened 2020-09-04
Stop Vendoring Go Dependencies

danielsdeleo:
### :nut_and_bolt: Description: What code changed, and why?

In #4298 I am making the external API a separate go module. Go modules vendoring doesn't make any special distinction for code in the same repo, which means that to propagate a change to the API, a developer must compile the updated protos and then revendor the dependencies to update the vendored copy. In response to this problem we can:

* Accept the need to revendor when updating the external API and let it be slow. It's slow because there's a lot of IO to the shared filesystem. It's also confusing that you have to update the local dependency. The latter problem can be somewhat mitigated by updating our studio tooling to compile the protos, but as you can see in this PR there's a lot of places where protos are built in a different way.
* Stop vendoring. The tradoffs for automate in general are:
  - (con) Updating dependencies no longer provides an easy way to review what is changing in those dependencies.
  - (con) We'll likely have to give up on rebuilding all affected components when a dependency changes. This means occasionally we could have build-breaks caused by a previous dependency update. (we might be able to do something here with changes to our existing tooling).
  - (con) More network resources required during the build (likely can be mitigated with some caching)
  - (pro) Fewer situations where people need to regenerate bldr.toml.
  - (pro) Likely makes it easier to split out more modules over time if we care to.
  - (pro) Less work when updating a dependency
* Attempt to create tooling to update the vendor dir quickly after compiling new protos. This likely involves doing unsupported things with the internals of the go module system.

This PR demonstrates removing the vendoring. A summary of the changes:
* Delete the vendor dir
* Update our revendor tooling to vendor external protobufs in a protovendor dir (it's renamed because the presence of a `vendor` directory may cause go tooling to behave in vendor mode). Modify protobuf build scripts to match
* Update docs, readmes, etc. that reference vendoring or give instructions for using `dep` 🙀 
* Update other scripts that assume the presence of the vendor dir.

### :white_check_mark: Checklist

- [ ] I have read the [CONTRIBUTING document](https://github.com/chef/automate/blob/master/CONTRIBUTING.md).
- [ ] Tests added/updated?
- [ ] Docs added/updated?
- [ ] All commits have been signed-off for the [Developer Certification of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

PR #4330, opened 2020-09-13
Developer-docs: Add a bit more info on module maintenance

msorens:

With the recent change from vendor mode to module mode, thought it would be useful to have just a bit more guidance for the casual developer.

PR #4314 Stop Vendoring Go Dependencies

### :+1: Definition of Done
More docs -- yay!

NA

- [ ] Tests added/updated?
- [x] Docs added/updated?

### DIFF
--- a/dev-docs/development-overview.md
+++ b/dev-docs/development-overview.md
@@ -5,6 +5,35 @@
 We're using Go's modules now. For an introduction, see:
 https://blog.golang.org/using-go-modules
 
+### How to add a new dependency
+
+When you want to use a new module in your code, just add an `import` statement for the module and use it:
+
+```go
+import "rsc.io/quote"
+. . .
+func Hello() string {
+    return quote.Hello()
+}
+```
+
+The next time you build, the module will be automatically downloaded and your go.mod/go.sum files updated.
+Be sure to commit the go.mod/go.sum changes.
+
+You can check that the imported module is in place with, for example:
+
+```shell
+go list -m rsc.io/q...
+```
+
+### How to remove a dependency
+
+It is important to be a good steward and clean up after yourself when you remove a dependency.
+Using `go build` can easily tell when something is missing and needs to be added, but not when it can safely be removed.
+The `go mod tidy` command is provided to do this extra work, but you must run it explicitly after you have removed uses of the no-longer-needed dependency from your code.
+If you run `go list -m all` before and after the tidy command, you should see that it has removed the unused dependency from both Go's cache and your go.mod/go.sum files.
+Be sure to commit the go.mod/go.sum changes.
+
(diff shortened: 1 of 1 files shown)


dev-docs/development-overview.md
```diff
@@ -5,6 +5,35 @@
...
+
+```shell
+go list -m rsc.io/q...
+```
+
+### How to remove a dependency
+
+It is important to be a good steward and clean up after yourself when you remove a dependency.
+Using `go build` can easily tell when something is missing and needs to be added, but not when it can safely be removed.
+The `go mod tidy` command is provided to do this extra work, but you must run it explicitly after you have removed uses of the no-longer-needed dependency from your code.
```
danielsdeleo, line 33: My previous work left the `scripts/revendor` script in place, I think it’s the only “right” way to remove a dep at the moment.

There is a bit of a weird situation where we are still vendoring protos and still using `go get` to fetch the repos with those protos, but those repos don’t have any go files in them. The process of fetching the proto repos adds the modules to the `go.sum` file, but `go mod tidy` will remove them. So you have to revendor the protos or else there’s churn in the `go.sum`, which I *think* will fail the repo health check. The revendor script does all of these things in the desired order.

I’ll try to find a way to clean this up. Perhaps making the protovendor directory a separate go module would do it.

msorens: @danielsdeleo Do you want to also update this PR, or put it on hold until you have a chance to do the work you mention?

danielsdeleo: #4346 should clean things up

danielsdeleo (approved):
Yay docs! 📚 💥 

Let's get #4346 in first so as not to lead anyone astray.

dev-docs/development-overview.md
msorens: Cool; it sounds like this doc file needs updating a bit due to #4346, though I am not clear on precisely what the changes should be. @danielsdeleo could you push a commit at least with the basics...?

danielsdeleo: @msorens #4346 makes what you wrote here correct :) 

Longer explanation: prior to #4346, you needed to run `scripts/revendor.sh` to remove a go dependency because our protobuf and golang dependency management were intertwined, which meant we used go tools to fetch some repos with no go files in them. Fetching those repos with `go get` would add them to the the `.mod` and `.sum` files, but running `mod tidy` would remove them because they didn't have any go files (and thus nothing did/could import them). Now protobuf dependency management is contained to the `protovendor` part of the tree (which is its own go module now), so the standard go tooling can be used. That said, you can use the revendor script if you want.

srenatus (approved):

PR #4336, opened 2020-09-14
Regenerate the dev certs

jaym:
They're expired, this should get the tests to pass


msorens:
Could you include instructions in the PR preamble above about how to do this for next time? Thanks!

dev/certs/event-gateway-sni-one.crt
```diff
@@ -1,25 +0,0 @@
------BEGIN CERTIFICATE-----
```
stevendanna, line 1: Are these not needed anymore?

stevendanna:
In general:

```
go run /src/tools/dev-cert-gen/main.go --regen-ca
go run /src/tools/dev-cert-gen/main.go --all --force
```

should do it.  Although it looks like we have a few certs that this tool doesn't know about.

stevendanna:
Renegerated the extra certs with:

```
 go run /src/tools/dev-cert-gen/main.go event-gateway-sni-one
 go run /src/tools/dev-cert-gen/main.go event-gateway-sni-two
```

dev/certs/event-gateway-sni-one.crt
jaym: i just emptied out the directory and reran the generate commands. maybe they dont generate everything

PR #4337, opened 2020-09-14
Fix applications migrations to run correctly twice

danielsdeleo:

In 6f8b8332d62cc158471ec55ec2d212a79cbdd177 we modified the applications service integration tests to run the service's schema migrations twice. This is intended to ensure we can rerun a migration if it's stopped by a transient failure. Though I tested it at the time (perhaps I was not testing what I thought I was testing?), I am now seeing new failures that I did not see originally (and have not appeared in Ci).

### :+1: Definition of Done

You can run the following without seeing a failure in the schema migration: 

```
start_applications_service
applications_integration
```

In the studio (no need to build): 
```
start_applications_service
applications_integration
```

### DIFF
--- a/components/applications-service/pkg/storage/postgres/schema/sql/11_add_denormalized_services_table.up.sql
+++ b/components/applications-service/pkg/storage/postgres/schema/sql/11_add_denormalized_services_table.up.sql
@@ -1,3 +1,4 @@
+DROP TABLE IF EXISTS service_full;
 CREATE TABLE IF NOT EXISTS service_full (
   id                        SERIAL    PRIMARY KEY,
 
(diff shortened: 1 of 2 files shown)


components/applications-service/pkg/storage/postgres/schema/sql/11_add_denormalized_services_table.up.sql
```diff
@@ -1,3 +1,4 @@
+DROP TABLE IF EXISTS service_full;
```
danielsdeleo, line 1: In migration 15, we drop the column `health_check_message`, but it is referenced in migration 12. Thus `CREATE TABLE IF NOT EXISTS` is not adequate, we need to drop the table.

Longer explanation
- (1st time) migration 11 runs, creates the table including the `health_check_message` column
- (1st time) migration 12 runs, the column is there
- (1st time) migration 15 removes the column
- (2nd time) migration 11 runs, doesn't create the table because it's already there, but its the one with `health_check_message` already removed
- (2nd time) migration 12 runs and tries to reference the `health_check_message` column which isn't there.

Dropping the table should be fine, in a production situation, the table should not exist or at worst we are immediately trying to recover after some intermittent failure halted this migration. In either case there shouldn't be data in the table.

lancewf (approved):

PR #4338, opened 2020-09-14
Make services listen on localhost

jaym:
I've moved all the internal services to listen only on localhost. There was the hope that one day we'd run this in some sort of clustered way using the hab ring, but that's likely not going to happen given that the ha implementation just treats each automate installation as a stateless frontend.

Things to look at:
- I made secureconn pass the no proxy option for dialing. This actually came about because automate-gateway was failing the proxy tests because I didn't correctly fix the health check script. However, the logs showed that the unix domain socket we use was trying to go through the proxy. This is a bug that currently exists on master, you'll see it in the logs for the proxy tests if you look.
- postgres will only listen on localhost
- I didn't change elasticsearch and event gateway


components/compliance-service/habitat/hooks/run
```diff
@@ -92,7 +86,7 @@ CONFIG="$CONFIG --message-buffer-size {{cfg.service.message_buffer_size}}"
 {{~#eachAlive bind.authn-service.members as |authn|}}
 {{~#if @last}}
 addNoProxy {{authn.sys.ip}}
```
stevendanna, line 88: We can likely remove a lot of these `addNoProxy` lines here and in other configuration files as well.

jaym: ya, i've been inconsistent about it. will need to go delete those

components/automate-dex/habitat/templates/config.yml
```diff
@@ -203,7 +203,9 @@ connectors:
       {{- if cfg.service.external_fqdn}}
       redirectURI: https://{{cfg.service.external_fqdn}}/dex/callback
       {{- else}}
-      redirectURI: http://{{cfg.service.host}}:{{cfg.service.port}}/dex/callback
+      # This is likely not the behavior you want. You should set external
+      # fqdn
```
stevendanna, line 207: This comment has me nervous because it seems like a message to the user but the user is unlikely to read the comments here :D.

stevendanna:
💯 Overall I think this is going to be a great improvement for a lot of our customers.

components/automate-dex/habitat/templates/config.yml
jaym: i'm about 51% sure nobody will ever make it here

stevenoneill:
Looking forward to this one!

stevendanna (approved):
Great work, thanks for powering through this.

yzl (approved):

PR #4339, opened 2020-09-14
Connected Services Filter (API only)

danielsdeleo:

[Apps tab] A customer has requested to be able to filter on only connected services. They want to see if all supervisors that are expected to upgrade have indeed upgraded. Their estate includes systems distributed over sometimes unreliable WAN links, which can become disconnected and not expected to update. Therefore it will be helpful to filter out these disconnected systems to show only the connected ones.

This PR implements filtering by `status:connected` when listing services.

UI Card: #4343

### :+1: Definition of Done

You can filter for only connected services in the Automate API

```
hab pkg install -bf core/curl
hab pkg install -bf core/jq-static

build components/applications-service
start_all_services
```

At first all the services are connected, so nothing is filtered out. You can see the connected services with something like (adjust the jq query as you like):  

```
curl -fsS -k -H "api-token: ${api_token}" "https://localhost/api/v0/applications/services?filter=status:connected"  | jq '.services | .[] | .id, .disconnected '
```

After 5 minutes or so, all the services should be disconnected and you won't see them in the above query. At this point, if you re-run `applications_populate_database` you will have a mix of connected and disconnected, which you can see by flipping the filter from `status:connected` to `status:disconnected`.

### DIFF
--- a/components/applications-service/integration_test/services_test.go
+++ b/components/applications-service/integration_test/services_test.go
@@ -232,6 +232,13 @@ func TestGetServicesWithDisconnectedServices(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Len(t, res.Services, 0)
 
+	reqConnectedOnly := &applications.ServicesReq{
+		Filter: []string{"status:connected"},
+	}
+	connectedOnlyRes, err := suite.ApplicationsServer.GetServices(ctx, reqConnectedOnly)
+	assert.NoError(t, err)
+	assert.Len(t, connectedOnlyRes.Services, 6)
+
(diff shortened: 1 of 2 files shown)


kagarmoe:
@danielsdeleo Where would we document this--in https://github.com/chef/automate/blob/master/api/external/applications/applications.proto?

danielsdeleo:
@kagarmoe this is an extension of the status filtering, which is currently here: https://github.com/chef/automate/blob/8535c9a3980a65cfb071e39ed7e6e4f2b8f62d89/api/external/applications/applications.proto#L446

vjeffrey (approved):

PR #4346, opened 2020-09-15
Separate proto and golang dep management

danielsdeleo:

After removing go module vendoring, we still have some special behavior in scripts where we use the go tooling to fetch proto-only repos for vendoring. The process of fetching the vendored protos adds the vendored repo to the go.sum file, but running `go mod tidy` removes it from the go.sum file because there are no go files in the repo and thus no go imports pointing to it.

In this PR I am separating our golang dependency management and protobuf dependency management. This will allow developers to manage go dependencies using the standard go tooling commands and processes. Our protobuf dependency management is still custom and requires using our scripts, which is expected because the protobuf ecosystem doesn't provide standardized tooling.

This change is achieved by:
* making the `protovendor` dir a separate module. This hides it from the view of any go dependency management operations that happen in the outer repo.
* track the protovendor deps in the protovendor's go.mod/go.sum
* update our scripting to match the changes.

#4330
#4314

### :+1: Definition of Done

You can run `go mod tidy` from the root of the repo and no files are changed.

Run `go mod tidy` from the root of the repo, and nothing is changed. Running `scripts/revendor.sh` from the root of the repo succeeds.


lancewf (approved):
Thank you for updating this. Everything worked for me.

PR #4349, opened 2020-09-16
Use Hugo asset pipeline for API docs

kagarmoe:

Using the Hugo asset pipeline could help with 
`Error: Error building site: failed to render pages: render of "page" failed: "/chef-web-docs/_vendor/github.com/chef/automate/components/docs-chef-io/layouts/_default/data-api.html:14:35": execute of template failed: template: _default/data-api.html:14:35: executing "_default/data-api.html" at <"js/vendor/redoc.standalone.js">: nil pointer evaluating resource.Resource.Permalink
make: *** [serve] Error 255`

Consideration:
Slow. Possibly even slower than the original. We may need to call it earlier and minifiy, or something similar.
https://gohugo.io/hugo-pipes/

### DIFF
--- a/components/docs-chef-io/layouts/_default/data-api.html
+++ b/components/docs-chef-io/layouts/_default/data-api.html
@@ -11,6 +11,7 @@
   <body>
     <a class="return" href="/automate/">Return to Docs</a>
     <ReDoc spec-url="/automate-api-docs/all-apis.swagger.json" scroll-y-offset="34"></ReDoc>
-    <script src="{{ (resources.Get "js/vendor/redoc.standalone.js").Permalink }}"></script>
+    {{ $redocJs := resources.Get "js/vendor/redoc.standalone.js" }}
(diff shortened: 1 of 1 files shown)


danielsdeleo (approved):
As far as I can tell, the generation of the API docs content should be totally unchanged, as we are feeding the content from `automate-api-docs/all-apis.swagger.json` into redoc in the same way before and after this pull request.

IanMadd:
Yeah, I don't think this is really going to change anything. We should probably add `type="text/javascript"` to the existing `script` tag, but I think Hugo's going to process this the same way that it already does. You might just have a newer version of Hugo that doesn't play well with the existing site. A [Hugo update](https://gohugo.io/news/0.74.0-relnotes/) changed the way it processes Javascript, which may have broken older functionality. I should spend some time looking into updating our site to match the newest Hugo version.

IanMadd (approved):
tenor-1500184

Issue #4333, opened 2020-09-14
automate scan results only displays one result at a time


vjeffrey:
## User Story
if you open a result in scan results sidebar, and then open another one, you'll see the displayed result is the last one clicked for both of the opened results

original issue: https://github.com/chef/customer-bugs/issues/243
slack link to video: https://chefio.slack.com/archives/CBSQQBK19/p1599707246151600

PR #4357, opened 2020-09-21
Fix incorrect reporting-profile sidebar expansion content

scottopherson:

This commit fixes https://github.com/chef/automate/issues/4333

The scan results expansion items were not being correctly mapped to the nodes they belong to.

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/shared/reporting/scan-results.service.ts
+++ b/components/automate-ui/src/app/pages/+compliance/shared/reporting/scan-results.service.ts
@@ -1,6 +1,10 @@
 import { Injectable } from '@angular/core';
 import { BehaviorSubject, Observable } from 'rxjs';
 
+interface ControlsByNode {
+  [nodeId: string]: any;
+}
+
 @Injectable()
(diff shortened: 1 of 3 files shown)


vjeffrey (approved):

Shadae (approved):
:shipit:

components/automate-ui/src/app/pages/+compliance/shared/reporting/scan-results.service.ts
```diff
@@ -1,6 +1,10 @@
 import { Injectable } from '@angular/core';
 import { BehaviorSubject, Observable } from 'rxjs';
 
+interface ControlsByNode {
+  [nodeId: string]: any;
```
msorens, line 5: So the controls do not have any definite type? I do see a lot of `any` values here and below. 🤷

scottopherson: Yes, this entire area of the code really needs more types.

susanev (approved):

Issue #4263, opened 2020-08-24
add "last 24 hrs" selection to compliance reporting searchbar


vjeffrey:
## User Story
we are modifying the backend logic to support showing the last 24 hrs of data when no end time has been provided.  for the ui portion of things, the plan is to default to the "last 24 hrs" selection, allowing a user to select a specific date if desired.

more details in the original card, https://github.com/chef/automate/issues/4027

API specifics: 
- when requesting last 24 hrs, do NOT provide end date
- when requesting specific date, send date the same way we do today

## Definition of Done
ui has option to select 24 hrs instead of just specific days

alexpop:
@scottopherson something else I want to highlight: For data in the past, `end_time` is always needed now, even for API calls that before didn't need it, like filtering for a `job_id`, suggestions or getting a report by just providing the report id. All these calls now will only search the last 24h worth of data unless end_time is specified.

PR #4358, opened 2020-09-21
Add option to select last 24h compliance report

scottopherson:
WIP: depends on https://github.com/chef/automate/pull/4310

This commit adds the option to view the last 24h of compliance data. Compliance API endpoints were [recently updated](https://github.com/chef/automate/pull/4310) to return last 24h window of data if `start_time` and `end_time` are excluded from requests.

https://github.com/chef/automate/issues/4263
https://github.com/chef/automate/pull/4310

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.html
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.html
@@ -66,10 +66,24 @@
       <chef-icon>filter_list</chef-icon>
       <span>{{ filters.length }}</span>
     </chef-button>
-    <chef-button class="calendar-btn" secondary (click)="toggleCalendar()">
-      <chef-icon>date_range</chef-icon>
-      <span>{{ date | datetime: CHEF_SHORT_DATE }}</span>
+    <chef-button class="calendar-menu-btn" secondary (click)="toggleCalendarMenu()">
+      <span *ngIf="last24h">Last 24 hours</span>
+      <ng-container *ngIf="!last24h">
+        <chef-icon>date_range</chef-icon>
+        <span>{{ date | datetime: CHEF_SHORT_DATE }}</span>
+      </ng-container>
     </chef-button>
+    <chef-dropdown class="calendar-menu-dropdown" [attr.visible]="calendarMenuVisible">
+      <chef-click-outside omit="calendar-menu-btn" (clickOutside)="hideCalendarMenu()">
+        <chef-button class="select-last-btn" secondary (click)="handleSelectLast24()">
+          Last 24 hours
+        </chef-button>
+        <chef-button class="calendar-btn" secondary (click)="showCalendar()">
+          <chef-icon>date_range</chef-icon>
+          <span>{{ date | datetime: CHEF_SHORT_DATE }}</span>
+        </chef-button>
+      </chef-click-outside>
+    </chef-dropdown>
     <chef-dropdown class="calendar-dropdown" [attr.visible]="calendarVisible">
       <chef-click-outside omit="calendar-btn" (clickOutside)="hideCalendar()">
         <chef-calendar
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.ts
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.ts
@@ -313,11 +313,15 @@ export class ReportingComponent implements OnInit, OnDestroy {
     const queryParams = {...this.route.snapshot.queryParams};
     const endDate = moment.utc(event.detail);
 
-    if (moment().utc().format('YYYY-MM-DD') === moment(endDate).format('YYYY-MM-DD')) {
-      delete queryParams['end_time'];
-    } else {
-      queryParams['end_time'] = moment(endDate).format('YYYY-MM-DD');
-    }
+    queryParams['end_time'] = moment(endDate).format('YYYY-MM-DD');
+
+    this.router.navigate([], {queryParams});
+  }
+
+  onLast24Selected() {
+    const queryParams = {...this.route.snapshot.queryParams};
+
+    delete queryParams['end_time'];
 
     this.router.navigate([], {queryParams});
   }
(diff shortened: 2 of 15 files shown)


components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.html
```diff
@@ -66,11 +66,25 @@
...
+        <span>{{ date | datetime: CHEF_SHORT_DATE }}</span>
+      </ng-container>
     </chef-button>
-    <chef-dropdown [attr.visible]="calendarVisible">
+    <chef-dropdown class="calendar-menu-dropdown" [attr.visible]="calendarMenuVisible">
+      <chef-click-outside omit="calendar-menu-btn" (clickOutside)="hideCalendarMenu()">
+        <chef-button class="select-last-btn" secondary (click)="handleSelectLast24()">
+          Last 24 hours
+        </chef-button>
+        <chef-button class="calendar-btn" secondary (click)="showCalendar()">
```
msorens, line 81: This should not be working as written:
```suggestion
        <chef-button class="calendar-btn" secondary (click)="showCalendarMenu()">
```

components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.ts
```diff
@@ -517,4 +522,8 @@ export class ReportingComponent implements OnInit, OnDestroy {
   convertMomentToDate(m: moment.Moment): Date {
     return new Date(Date.UTC(m.year(), m.month(), m.date()));
   }
+
+  isLast24h(urlFilters: Chicklet[]): boolean {
```
msorens, line 526:
```suggestion
  private isLast24h(urlFilters: Chicklet[]): boolean {
```

components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.ts
```diff
@@ -313,11 +313,15 @@ export class ReportingComponent implements OnInit, OnDestroy {
     const queryParams = {...this.route.snapshot.queryParams};
     const endDate = moment.utc(event.detail);
 
-    if (moment().utc().format('YYYY-MM-DD') === moment(endDate).format('YYYY-MM-DD')) {
-      delete queryParams['end_time'];
```
msorens, line 317: No longer need to remove the end_time in this case...?

components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
```diff
@@ -258,7 +258,7 @@ export class StatsService {
     return filters;
   }
 
-  formatFilters(reportQuery: ReportQuery) {
+  formatFilters(reportQuery: ReportQuery, requestsLast24h = true) {
     const apiFilters = reportQuery.filters.reduce((formatted, filter) => {
         let type = filter['type']['name'];
         let value = filter['value']['id'] || filter['value']['text'];
```
msorens, line 264: Consider dot notation to make it less noisy here.

components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
```diff
@@ -313,6 +313,12 @@ export class StatsService {
       return formatted;
     }, []);
 
+    // If last 24 hour interval is selected, exclude start_time and end_time
+    // from requests that return data from 24h search index.
+    if (reportQuery.last24h && requestsLast24h) {
```
msorens, line 318: Looks like `requestsLast24h` is the request here; then what is `reportQuery.last24h` ? (Why does it need both?)

msorens (commented):
Looks good; just a few questions/comments.
FYI: My review is just from code-reading; have not exercised the code.

alexpop:
Hey Scott, looking great overall. While doing some testing I noticed two things I want to bring up.

**1.** When I go to Compliance > Nodes > click on a node > Scan History
I see only reports from the last 24h. Looking at the API call to `/compliance/reporting/reports`, it doesn't have `start_time` and `end_time` filters. This results in only the last 24 hour reports to be listed for the node. See below screenshot:

scan_history_24h

**2.** When I go to Compliance > Nodes > click on a node > REFRESH PAGE > Scan History
I see reports for more than 24h. Looking at the API call to `/compliance/reporting/reports`, it has `start_time` and `end_time` filters spanning the last 10 days. See below screenshot:

scan_history_10_days

Do we only want 10 days worth of reports to show up? @jonong1972 @kekaichinose

alexpop:
The date picker goes under the controls label here:

calendar

alexpop:
Also, we noticed in standup today while reviewing the change that the range displayed below the trend graph:

calendar_range

In this case ^ `Thu, 17 Sep 2020 - Sun, 27 Sep 2020` doesn't match the actual range of the bucket displayed (18 Sept - 28 Sept).
Would be great to get the two ranges aligned as well.
We also talked about displaying **UTC** somewhere in there to make it clear that the vertical buckets are UTC dates. Will let @jonong1972 expand on that.

components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.ts
scottopherson: Nope. Empty end_time query param is now used for 'last 24hr' instead of 'current day'.

components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.html
scottopherson: I might be misunderstanding this suggestion.

Clicking the existing datetime button in the searchbar now calls `showCalendarMenu()` which opens a new 'menu' dropdown  where you can select `last 24 hours` or select the date for a specific date in the calendar. Selecting the date option in this new menu calls the existing `showCalendar()` method to display the calendar picker as before.

Actually, it seems I didn't need to add `showCalendarMenu()` since only `toggleCalendarMenu()` and `hideCalendarMenu()` are currently being used.

msorens: Could be my reporting error; I thought when I checked there was no `showCalendar` function.

components/automate-ui/src/app/pages/+compliance/shared/reporting/stats.service.ts
scottopherson: I'll save this suggestion for a separate cleanup commit since this will increase the diff on this function quite a bit.

alexpop (approved):

Issue #3673, opened 2020-05-13
Compliance report: Pagination component not being updated when the status filter is applied.


susanev:
## overview
There is a dependency on #3664 

Prior to #3664, if you applied a status filter the nodes, profiles, or controls would still be distributed across all of their pages before the filter was applied. 

#3664 improves it so that all objects get moved to the beginning of the pages, but all the empty pages are left around and if a user navigates to one of the empty pages they get stuck and have to do a hard refresh.

if a user applies a status filter, pagination buttons should only be shown if there are objects on them.

## repro
[internal slack video](https://chefio.slack.com/archives/CBSQQBK19/p1589418413012200)

- add at least 101 compliance nodes, `chef_load_compliance_nodes`
- navigate to compliance > reports > nodes (or profiles or controls)
- apply a status filter
- navigate to the second page via the pagination component
- notice the blank pages where you get stuck

## definition of done
when a user applies a status filter, pagination is recalculated and all results are shown on the first pages and no pagination buttons are shown with no data on them and users arent able to get stuck

PR #4363, opened 2020-09-22
Automate-3673 Compliance reports pagination pages not correct

tarablack01:
There is a dependency on #3664

Prior to #3664, if you applied a status filter the nodes, profiles, or controls would still be distributed across all of their pages before the filter was applied.

#3664 improves it so that all objects get moved to the beginning of the pages, but all the empty pages are left around and if a user navigates to one of the empty pages they get stuck and have to do a hard refresh.

if a user applies a status filter, pagination buttons should only be shown if there are objects on them.

https://github.com/chef/automate/issues/3673

### :+1: Definition of Done
Pagination shows correct page numbers

add at least 101 compliance nodes, chef_load_compliance_nodes
navigate to compliance > reports > nodes (or profiles or controls)
apply a status filter
navigate to the second page via the pagination component
notice the blank pages where you get stuck

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-nodes/reporting-nodes.component.html
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-nodes/reporting-nodes.component.html
@@ -116,7 +116,7 @@
   <app-page-picker
     class="reporting-nodes-paging"
     [perPage]="reportData.nodesListParams.perPage"
-    [total]="reportData.nodesList.total"
+    [total]="getTotalNodes()"
     [page]="reportData.nodesListParams.page"
     (pageChanged)="onNodesListPageChanged($event)">
   </app-page-picker>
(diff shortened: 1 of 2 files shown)


components/automate-ui/src/app/pages/+compliance/+reporting/+reporting-nodes/reporting-nodes.component.html
```diff
@@ -116,7 +116,7 @@
   <app-page-picker
     class="reporting-nodes-paging"
     [perPage]="reportData.nodesListParams.perPage"
-    [total]="reportData.nodesList.total"
+    [total]="getTotalNodes()"
```
msorens, line 119: So here's the thing... Whenever we use a **data** function like this (as opposed to an **action** function like a click handler), Angular cannot know if its value has changed so it needs to call it. Repeatedly. I have not spent time fully characterizing how often, but in some instances I have seen it looping endlessly for no apparent reason. Since that time I have been espousing a ban on data functions in templates unless it is absolutely unavoidable. 

There are two alternative approaches: "declared properties" and "pure pipes". See my extensive notes on these in the bottom half of the preamble to https://github.com/chef/automate/pull/3810. That PR also shows examples of the "declared properties" approach, which is more appropriate in this case.

Issue #4329, opened 2020-09-10
Session gets expired if 401 error raised from chef server rest API in automate infra view modules tabs


sachin-msys:
## Describe the bug
Session gets expired if 401 error raised from chef server rest API navigating to infra proxy modules tabs which are  Roles, Environments, Data Bags, Clients, PolicyFiles.

## To Reproduce
- Call one of chef server rest API which gives 401 error and navigate to any tabs except Cookbooks, Details, and Reset Admin Key on organization details page. 

## Expected behavior
For module like automate infra view we are communicating with chef servers API's that is external API's in that case we don't want session gets expired on getting 401 error, instead of that we need to show user a error notification so that user can take a preventive action for error.

## Screenshots
![401-error](https://user-images.githubusercontent.com/54940695/92713921-270d5900-f379-11ea-8bac-9a062d750945.png)

PR #4369, opened 2020-09-23
Handle 401 error in infra modules tab and show empty state UI 

sachin-msys:
1. In Automate infra view module we are communicating with chef servers API's that is external API's in that case we don't want session gets expired on getting 401 error, instead of that we need to show user a error notification so that user can take a preventive action for error so we have introduced a `skip-interceptor` header for external API's to ignore chef session logout loop in auth interceptor service on getting 401 error from server.

2. Introduced a UI component to show empty state for all infra modules tabs when the admin key is invalid, before empty state UI we were showing empty list table. As a UX improvement we have introduced a UI for empty state and redirection link for Reset Admin Key tab.

3. Append `$` on `isDestroyed` to visually indicate that it's an observable / subscribable through out the infra module as a PR bonus.

Fixes: #4329 
### :+1: Definition of Done
1. Session does not gets expired on getting 401 from chef servers API's in infra proxy Tabs.
2. Now on getting 401 error now user can see empty state UI with reset admin key redirection link instead of empty data tables.
1. Go to 'Infrastructure' from top nav.
2. Select 'Chef Servers' option from side nav
3. Now to add test data go to [adding test data](https://github.com/chef/automate/blob/master/dev-docs/adding-data/adding_test_data.md#adding-data-to-infra-views) file and follow the 3rd step and run the given cmd in `hab studio`.
4. Now refresh application you will see the `Server` list get populated with data.
5. Click on `Server list` item >> `Organisation list` item now you see a tab structure
6. Now click on `Reset Admin Key` and update the key with random invalid key.

- Before on getting 401 error from chef servers API's

session

- After on getting 401 error from chef servers API's

cookbook-empty-state
role-empty-state
environments-empty-state
databags-empty-state
clients-empty-state
policyfiles-empty-state


msorens:
(1) Currently your instructions to test this say just this: "Call infra proxy API which gives 401 error." Need further details there. How does one call the infra proxy API? And how does one do so with both a successful call (200) and a failed call (401) ?

(2) I appreciate the initiative of the "PR bonus". In the future, however, best to reserve that for fairly small additions to a PR. In this case, most of the PR is that bonus so it would have been better to make it a separate PR. (Not suggesting you need to change it now; just keep in mind for the future.)

susanev (approved):

sachin-msys:
> (1) Currently your instructions to test this say just this: "Call infra proxy API which gives 401 error." Need further details there. How does one call the infra proxy API? And how does one do so with both a successful call (200) and a failed call (401) ?
> 
> (2) I appreciate the initiative of the "PR bonus". In the future, however, best to reserve that for fairly small additions to a PR. In this case, most of the PR is that bonus so it would have been better to make it a separate PR. (Not suggesting you need to change it now; just keep in mind for the future.)

@msorens 1. Updated the steps to test this code changes please have a look.
2. I know I went to far with the flow :) will keep this in mind for the future for sure, Thanks. +1:

components/automate-ui/src/app/modules/infra-proxy/data-bags-list/data-bags-list.component.ts
```diff
@@ -42,12 +45,24 @@ export class DataBagsListComponent implements OnInit {
...
+        this.dataBags = allDataBagsState;
+        this.dataBagsListLoading = false;
+      } else if (getDataBagsSt === EntityStatus.loadingFailure) {
+        this.dataBagsListLoading = false;
+        this.authFailure = true;
+      }
     });
   }
+
+  ResetKeyTabRedirection(resetLink: boolean) {
```
scottopherson, line 60: We follow the typical JS convention of having function and method names as `camelCase`.  Class names and decorators do use `PascalCase`.

components/automate-ui/src/app/modules/infra-proxy/empty-state/empty-state.component.html
```diff
@@ -0,0 +1,6 @@
+<chef-toolbar>
+    <div class="empty-state">
+      <p>Unable to show {{ moduleTitle }} because admin key is invalid.</p>
+      <chef-button primary (click)="ResetKeyTabRedirection()">Reset Admin Key</chef-button>
+    </div>
```
scottopherson, line 5: Consistent 2-space tabbing

components/automate-ui/src/app/modules/infra-proxy/empty-state/empty-state.component.ts
```diff
@@ -0,0 +1,19 @@
...
+
+
+
+export class EmptyStateComponent {
+    @Input() moduleTitle: string;
+    @Output() resetKeyRedirection = new EventEmitter<boolean>();
+
+    ResetKeyTabRedirection() {
+        this.resetKeyRedirection.emit(true);
+    }
```
scottopherson, line 18: Consistent 2-space tabbing, camelCase, and there's some extra newlines above and below the `@Component` decorator.

scottopherson (approved):
Functionality and code organization looks good. Just a couple notes on code style and formatting. 👍

msorens:
@sachin-msys Thinking about your additional adding of the `$` to `isDestroyed$`, and reviewing the entire code base, I would like to suggest you do not do that in this PR. If we want to do that, it should be done to the entire code base for consistency, not just a handful of files. Right now, the code base uses `isDestroyed` about 97% of the time and `isDestroyed$` about 3%. Your change here would update that ratio to perhaps 80% vs 20%. Though it may not be ideal, I think it is better to minimize changes and we can later just "correct" that 3%.

sachin-msys:
> @sachin-msys Thinking about your additional adding of the `$` to `isDestroyed$`, and reviewing the entire code base, I would like to suggest you do not do that in this PR. If we want to do that, it should be done to the entire code base for consistency, not just a handful of files. Right now, the code base uses `isDestroyed` about 97% of the time and `isDestroyed$` about 3%. Your change here would update that ratio to perhaps 80% vs 20%. Though it may not be ideal, I think it is better to minimize changes and we can later just "correct" that 3%.

@msorens Thanks for the suggestion, reverted the `isDestroyed$`.

PR #4373, opened 2020-09-24
terraform: remove noexec on /tmp in hardened environment

stevendanna:
The hardened AMI we use now mounts /tmp as noexec. This is a common
hardening configuration and one that we already test extensively in
our integration tests.

Unfortunately, a good portion of our terraform setup does not appear
to work with /tmp as noexec.

This remounts /tmp, hopefully allowing the rest of our terraform setup
to work.


srenatus (approved):
LGTM

PR #4375, opened 2020-09-24
terraform: exit 0 on non-hardened environments

stevendanna:

### DIFF
--- a/terraform/test-environments/modules/chef_automate_install/main.tf
+++ b/terraform/test-environments/modules/chef_automate_install/main.tf
@@ -113,6 +113,7 @@ EOF
   provisioner "remote-exec" {
     inline = [
       "${var.hardened_security} && sudo /home/${var.ssh_username}/soften_mounts.sh",
+      "exit 0",
     ]
   }
 }


terraform/test-environments/modules/chef_automate_install/main.tf
```diff
@@ -113,6 +113,7 @@ EOF
   provisioner "remote-exec" {
     inline = [
       "${var.hardened_security} && sudo /home/${var.ssh_username}/soften_mounts.sh",
```
stevendanna, line 115: We may want set -e here I suppose

stevendanna: Let's not for now

srenatus (approved):
LGTM.
I wonder if we could also have an expression there? https://www.terraform.io/docs/configuration/expressions.html not a good link, but what I mean is if the var is true, render the sudo, if not, empty string or "exit 0". Less shell to deal with I suppose.

stevendanna:
@srenatus I like that suggestion, I left it as-is since it matches a pattern used elsewhere.  Perhaps we can follow up by fixing them all at once.

PR #4380, opened 2020-09-24
terraform: new CIS AMIs don't pass CIS profile

stevendanna:


ryancragun (approved):
Should we skip the failing controls in a2-hardend-security instead?

stevendanna:
@ryancragun Good call, updated

PR #4385, opened 2020-09-30
Bump elasticsearch and postgres packages

jaym:
The all the things to localhost change requires updated config for
elasticsearch and postgres. This brings in those changes

### DIFF
--- a/.expeditor/create-manifest.rb
+++ b/.expeditor/create-manifest.rb
@@ -299,8 +299,8 @@ pins = {
   # This will pin the services that we package and start with the deployment service,
   # while still allowing the clients of these databases to upgrade their client
   # libraries if any fixes are shipped there.
-  "automate-postgresql"    => {"origin" => "chef", "name" => "automate-postgresql",    "version" => "9.6.11", "release" => "20200723182209"},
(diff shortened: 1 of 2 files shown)


.expeditor/create-manifest.rb
```diff
@@ -300,7 +300,7 @@ def parse_products_meta_file(path)
   # while still allowing the clients of these databases to upgrade their client
   # libraries if any fixes are shipped there.
   "automate-postgresql"    => {"origin" => "chef", "name" => "automate-postgresql",    "version" => "9.6.11", "release" => "20200723182209"},
```
stevendanna, line 302: Do we neeed to do postgresql as well?

stevendanna (approved):
Thanks!

.expeditor/create-manifest.rb
jaym: let me double check

jaym: the package was modified, so i brought it in

stevendanna (approved):
Double Approved!

yzl (approved):

PR #4386, opened 2020-10-01
automate-dex: make local user signin disappear via config

srenatus:
This would be a noop if there always were local users -- regardless of
configuring LDAP or SAML, you'd always have more than one method.

If we want to allow disabling local users, we'll have to enable this, because
if you had ONLY SAML, you couldn't sign out -- you'd immediately be sent back
to your SAML IdP; and if your session was still valid, you'd even go straight
back to being signed into A2.

To not have users click "sign in with X" if they only have LDAP, or only local
users, enabled, but have them go straight to the username/password field, we
set alwaysShowLoginScreen to true only if SAML is in the mix.

If you have LDAP+SAML, or local+SAML, you'll see the selection anyways.

ℹ️ I've added the extra commits to
- add a new configurable
- use it to disable local users in dex *if* there's another connector set up. Otherwise, it would fail to start yelling about "no connectors set up" (rightly so!)

- deployinate, `rebuild components/automate-dex`, sign in and sign out with different SAML/non-SAML combinations
- you should see no "sign in with email" button if there's only local users, but should go straight to username/password
- same for "only LDAP/MSAD"
- as long as there's more than one connector; or ONLY SAML (not configurable right now), you'd see the selection.

ℹ️ To play with disabling the local users, use this config snippet,
```toml
[dex.v1.sys.connectors]
disable_local_users = true
```
and rebuild `automate-dex`, `automate-cli` and `automate-deployment`. Then try to sign in, and remove the SAML config to see that you can't do that. With this config applied, the the SAML config *not* removed, you should see this:

image


jaym (approved):

components/automate-dex/habitat/templates/config.yml
```diff
@@ -223,6 +219,15 @@ connectors:
...
 {{- end}}
+oauth2:
+  responseTypes:
+    - code
+    - token
+    - id_token
+  skipApprovalScreen: true
+{{- if cfg.connectors.saml}}
+  alwaysShowLoginScreen: true
+{{- end}}
```
msorens, line 230: I am not clear on what is the approval screen vs the login screen, or the whole process flow here for that matter. 🤷 You state:
> To not have users click "sign in with X" if they only have LDAP, or only local
> users, enabled, but have them go straight to the username/password field, we
> set alwaysShowLoginScreen to true only if SAML is in the mix.

Am I correct, for instance, that SAML is different than local or LDAP, in that we do not ask for the credentials, rather the SAML IdP does. So if we remove the "Sign in with SAML" (that, I guess, is on the login screen?) that is why one could never log out, yes?

msorens (approved):

components/automate-dex/habitat/templates/config.yml
srenatus: > Am I correct, for instance, that SAML is different than local or LDAP, in that we do not ask for the credentials, rather the SAML IdP does. 

Yes, the "SAML sign in page" is a redirect to the IdP, no username/password prompt.

> So if we remove the "Sign in with SAML" (that, I guess, is on the login screen?) that is why one could never log out, yes?

Signing out of A2 will get you back to dex. If dex decided that, since there's only one sign in method, that must be the one you want to use, it sends you to that method's sign in page. For SAML, that's a redirect to your IdP. If you still have an active session with the IdP (that's something we don't control), you'll be signed in automatically, and redirect to A2. So, from a user's perspective, there's no way to sign out. Clicking on "Sign out" gets you signed in again. (I'm just rephrasing, I think your understanding is correct.)

srenatus: The approval screen is this:
image

We have never enabled that, since it doesn't make sense in our setup.

msorens: Thanks!

PR #4391, opened 2020-10-05
tests: don't check cert validity period for internal CA

stevendanna:
We already put in a filter to ignore this check in
3a2e03b6e9e57bae4cd6fcf12c38fb153453cc11 but it appears that
drwetter/testssl.sh#1741 changed the name of the check.


susanev (approved):

srenatus (commented):
Thank you!

Issue #3759, opened 2020-05-20
Filtering Client Runs Data causes undesirable jump to top of screen


SEAjamieD:
## User Story
As a user, when i'm on the client runs page I want to be able to smoothly filter my client runs while maintaining my scroll depth on the page.

## Definition of Done
On Client runs, when switching from All to Successful Nodes to Failed Nodes etc, the new nodes are filtered correctly, and the scroll position is maintained on the page.

## Demo Script / Repro Steps
Navigate to https://a2-dev.test/infrastructure/client-runs
Click on any of the filters available
Notice the page jumps to top if scrolled down any distance.
![jump-to-top.gif](https://images.zenhubusercontent.com/5d7aa5d5d4249c0001cab607/03f51f26-d2e9-4514-b404-43e6698ee614)

PR #4392, opened 2020-10-06
added changes to prevent scrolling on client runs page tabs change

vinay033:

As a user, when i'm on the client runs page I want to be able to smoothly filter my client runs while maintaining my scroll depth on the page.
I have added some changes to preventing jump to top of screen

https://github.com/chef/automate/issues/3759

### :+1: Definition of Done
I have added one function which can find existing height of the component and again set the existing height after changing the tab.
```
STEP 1
inside the hab studio

STEP 2
open new window
go to automate UI path

$ cd components/automate-ui
and run the command 

npm run serve:hab

navigate to `https://a2-dev.test/infrastructure/client-runs`
add some sample data and then test the changes using tab change
```

- [ ] Tests added/updated?
- [ ] Docs added/updated?

client_runs

### DIFF
--- a/components/automate-ui/src/app/pages/client-runs/client-runs.component.ts
+++ b/components/automate-ui/src/app/pages/client-runs/client-runs.component.ts
@@ -242,8 +242,6 @@ export class ClientRunsComponent implements OnInit, OnDestroy {
   downloadFailed = false;
   downloadStatusVisible = false;
 
-  scrolling = false;
-
   constructor(
     private route: ActivatedRoute,
     private router: Router,
@@ -384,9 +382,9 @@ export class ClientRunsComponent implements OnInit, OnDestroy {
   }
 
   // prevent scrolling on tabs change
-  get height() {
+  get height(): { [cssClass: string]: string; }  {
     return {
-      'min-height': this.scrolling ? 'calc(100vh - 50px)' : 'calc(100vh - 49px)'
+      'min-height': 'calc(100vh - 49px)'
     };
   }
 
--- a/components/automate-ui/src/app/pages/client-runs/client-runs.component.ts
+++ b/components/automate-ui/src/app/pages/client-runs/client-runs.component.ts
@@ -242,6 +242,8 @@ export class ClientRunsComponent implements OnInit, OnDestroy {
   downloadFailed = false;
   downloadStatusVisible = false;
 
+  scrolling = false;
+
   constructor(
     private route: ActivatedRoute,
     private router: Router,
@@ -379,6 +381,13 @@ export class ClientRunsComponent implements OnInit, OnDestroy {
         verb: 'post'
       }
     ], []);
+  }
+
+  // prevent scrolling on tabs change
+  get height() {
+    return {
+      'min-height': this.scrolling ? 'calc(100vh - 50px)' : 'calc(100vh - 49px)'
+    };
   }
 
   hideNotification() {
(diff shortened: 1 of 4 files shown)


msorens (changes requested):
The browser experience is improved, ending up with the scroll position adjusted, but the image jumps around making it visually jarring. Would like to have the image not jump back and forth if possible.

components/automate-ui/src/app/pages/client-runs/client-runs.component.ts
```diff
@@ -381,6 +383,13 @@ export class ClientRunsComponent implements OnInit, OnDestroy {
     ], []);
   }
 
+  // prevent scrolling on tabs change
+  get height() {
+    return {
+      'min-height': this.scrolling ? 'calc(100vh - 50px)' : 'calc(100vh - 49px)'
```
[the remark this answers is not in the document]
vinay033: I have updated the PR, please check it once.

components/automate-ui/src/app/pages/client-runs/client-runs.component.ts
```diff
@@ -384,9 +382,9 @@ export class ClientRunsComponent implements OnInit, OnDestroy {
   }
 
   // prevent scrolling on tabs change
-  get height() {
+  get height(): { [cssClass: string]: string; }  {
     return {
-      'min-height': this.scrolling ? 'calc(100vh - 50px)' : 'calc(100vh - 49px)'
+      'min-height': 'calc(100vh - 49px)'
```
msorens, line 387: Since this is now a constant expression there is no reason for it to be in the code-behind; we could just use a normal style in the scss file, right?

SEAjamieD:
I agree with @msorens that the UX is certainly a bit jarring, however it does accomplish the task.  Would be ideal to keep the page from jumping

components/automate-ui/src/app/pages/client-runs/client-runs.component.ts
vinay033: sure, I am working on this to fix the UX jarring and I will update this PR soon.

vinay033:
@msorens , @SEAjamieD 
I have updated the PR and fix the page jumping when the user scrolls the page and switching the tabs.

scolling_tabs

SEAjamieD (approved):
Beautiful!  and so simple 🎉

msorens:
The fix is not working on my system; in fact it has regressed. When I click on a different tab it jumps to the top of the page and stays there. Note that with dev tools open I can confirm that the `content-container` div briefly gets `lock-scrolling` added to its CSS class list as it is supposed to, so it is running the right code, just not doing the right thing.

image

vinay033:
> The fix is not working on my system; in fact it has regressed. When I click on a different tab it jumps to the top of the page and stays there. Note that with dev tools open I can confirm that the `content-container` div briefly gets `lock-scrolling` added to its CSS class list as it is supposed to, so it is running the right code, just not doing the right thing.
> 
> image

@msorens 
I have debugged the `loading` function which is using in `content-container` div to apply `lock-scrolling` class, I found that the existing CSS property is not solving the page jumping issue when the tabs are changing, I have added changes to fix this. please check it once.

msorens:
Latest commits have made no improvement when I tested (checked both firefox and chrome). That is, I am seeing the originally reported issue: you click on a different "tab" and the position jumps to the top of the page.

SEAjamieD:
I am seeing the fix applied as described with no more jump.

client-runs jump fix

vinay033:
> Latest commits have made no improvement when I tested (checked both firefox and chrome). That is, I am seeing the originally reported issue: you click on a different "tab" and the position jumps to the top of the page.

have you checked my last commit because changes are working at my end and I have tested on ubuntu, can you please check it once again

Shadae (approved):
Pulled down the branch and loaded data with `chef_load_nodes` I see the same thing Jamie sees. LGTM

PR #4395, opened 2020-10-08
Introduce semgrep security/language validation

msorens:

Semgrep provides security validation via published rulesets for a variety of languages, including Go and soon TypeScript (now in alpha); see https://semgrep.dev/ for current offerings and ruleset browsing. It also allows writing arbitrary language validation rules, i.e. non-specific to security concerns, but rather to enforce conventions or patterns that the team views as "best practice".

## Focus of this PR

This PR introduces custom semgrep rules focussing on **rxjs lifecycle management**. The patterns required to properly manage rxjs subscriptions are both "best practice" and necessary, yet cannot be enforced by the compiler. Semgrep, though, provides just the mechanism to enforce these patterns.

Here are the rules I have introduced, separated into two files a la the single responsibility principle. All files in the semgrep directory are pulled in automatically.
```
rxjs-subscription.yaml
 - naming-for-isDestroyed
 - isDestroyed-needs-argument
 - isDestroyed-used-but-not-triggered
 - isDestroyed-triggered-but-not-used
 - ngOnDestroy-missing-next
 - ngOnDestroy-missing-complete

rxjs-syntax.yaml
 - combineLatest-not-combining
```

You can see the errors exposed (before fixes in this PR were applied) from this buildkite run:
https://buildkite.com/chef-oss/chef-automate-master-verify/builds/14170#d286ce19-5fb3-4276-a85f-6edadb38bbd6

Here is the error output there:
image

Note: some time ago a couple of us (@scottopherson and myself, maybe others) were suggesting that both Observables and things derived from them (Subjects, ReplaySubjects) should use [Finnish notation](https://medium.com/@benlesh/observables-and-finnish-notation-df8356ed1c9b) -- end with a dollar sign ($) -- but that is not the way the code base is at present. Before this commit, about 97% of `isDestroyed` instances existed vs. 3% of `isDestroyed$`. So to minimize changes I just went with the flow for now and implemented the rule to enforce it without the dollar sign. (If someone wants to follow-up globally switching all of them the other way, that is fine, too.)

## More on Semgrep

Semgrep comes in two flavors, **semgrep** and **semgrep agent**. The latter is currently in private beta (which we are participating in) and will eventually be more useful as an enterprise management tool. This is the dashboard for semgrep agent--looks sort of familiar, eh, auth-team? In theory, any Chef logging in to https://semgrep.dev/manage/projects has access to this dashboard.

image

In a nutshell, here are my learnings for both of them:
image

I am starting us out with the `r2c-ci` ruleset, provided by r2c. There is also an eponymous `r2c` ruleset that is the default if no rulesets are specified. There is a lot of overlap between the two, but neither is a subset of the other. I inquired about the difference: `r2c-ci` is the "kinder, gentler" version. Paraphrasing our friends at r2c: it includes rules that would not be annoying; those that largely everyone would say, "oh, yeah, we should fix that".

Note that `r2c-ci` includes (at the time of writing) 103 rules but not too many of those are for Go (and none are for TypeScript). I am assured that there is no performance penalty because rules are only applied if a given file has the appropriate extension matching the rule.

Like lint, semgrep can automatically fix rules that have been instrumented to support it; just add `--autofix` to the semgrep command. This applies equally to published rules as well as our own custom rules.

I have added `semgrep` and `semgrep-and-fix` as make targets in Makefile.common_go, so they are available to all Go projects, and automate-ui/Makefile for the UI.  Semgrep applies rules based on file extensions, so if you have a Go project it will only be running the Go rules (and perhaps the JSON rules if there are also JSON files in the project). It is thus safe--and prudent--to run semgrep at any level of the source tree, including the root of the repository, as is done in CI.

Semgrep Docs: https://github.com/returntocorp/semgrep/tree/develop/docs
Semgrep Action Docs: https://github.com/returntocorp/semgrep-action

### :+1: Definition of Done

Buildkite has a new badge. (Note that once my PR to introduce a Semgrep icon gets approved, that `:semgrep:` tag will turn into an icon.)
image

To run locally, install semgrep via homebrew; run as shown in the chart above. If you wanted to expose a failure, you could selectively revert any of the fix commits in this PR.

If you want to play with semgrep agent, I would be happy to provide the ID and token.

### Blocker with Semgrep Agent

It would be preferable to run Semgrep Agent instead of just Semgrep in CI, but I have not done that in this PR. Semgrep Agent needs an ID, which we can hard-code, and a token, which should be treated as a secret. I attempted to hook it up in buildkite following the secrets used in Automate, with this alternate recipe:
```
  - label: "Semgrep"
    command:
      - python -m semgrep_agent --publish-token $SEMGREP_TOKEN --publish-deployment CHEF_ID
    expeditor:
      secrets:
        SEMGREP_TOKEN:
          path: secret/chef-cloud/semgrep
          field: publish-token
      executor:
        docker:
          image: returntocorp/semgrep-action:v1
          workdir: /chef-cloud
```
But buildkite failed on the first step:
```
/app/lib/expeditor/cli/buildkite/executor/docker.rb:65:in `initialize':
accounts and secrets not supported
with 'returntocorp/semgrep action:v1'
```
According to @tduffield, we have no means to inject secrets into container images we don't build/own. To move forward we would need to install semgrep into the chefes/buildkite image, as @tduffield has suggested. But see next section...

### But Do We Need Semgrep Agent?
Though it would be nice to have the on-the-fly configurability provided in the GUI, plus soon to be tracking and stats, we can get by without it for the time being. One driving factor would be if we had either (a) more than one published ruleset we wanted to use regularly, or (b) wanting to publish our custom rules to multiple rulesets. The reason for both is that with the base semgrep, you can only specify a single ruleset per Buildkite task due to the way the semgrep docker image was constructed (with the entrypoint mode).
For our custom rules, at present I think it makes sense to commit those in the working repositories. Currently, I am only instrumenting chef-cloud and automate (a little bit of duplication, if it manifests, never hurt anyone, right...?). My plan is to have one Buildkite task against a published, public ruleset (`r2c-ci`) and a separate Buildkite task for Chef custom rules. (This PR, by the way, contains just the Chef custom rules; another PR will handle the `r2c-ci` and its necessary fixes.)

By the way, one very cool thing about semgrep agent--it is designed to be painless to turn on: it will only flag **new** things rather than flagging everything-you-have-previously-done-wrong in your code base, giving you breathing room to go back and fix the technical debt whenever you wish.

https://github.com/chef/chef-cloud/pull/385

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -12,7 +12,11 @@ steps:
         docker:
           image: returntocorp/semgrep:latest
           entrypoint: semgrep
-          command: ["--error", "--config", "/go/src/github.com/chef/automate/semgrep" ]
+          command: [
+            "--error",
+            "--config", "/go/src/github.com/chef/automate/semgrep",
+            "/go/src/github.com/chef/automate"
+          ]
 
   #
   # A2 protobufs - Detect if there is any component that has protobuf files that hasn't been compiled
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -12,7 +12,7 @@ steps:
         docker:
           image: returntocorp/semgrep:latest
           entrypoint: semgrep
-          command: ["--error", "--config", "/workdir/semgrep" ]
+          command: ["--error", "--config", "/go/src/github.com/chef/automate/semgrep" ]
 
   #
   # A2 protobufs - Detect if there is any component that has protobuf files that hasn't been compiled
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -5,6 +5,15 @@ expeditor:
         workdir: /go/src/github.com/chef/automate
 
 steps:
+
+  - label: ":semgrep: Semgrep"
+    expeditor:
+      executor:
+        docker:
+          image: returntocorp/semgrep:latest
+          entrypoint: semgrep
+          command: ["--error", "--config", "/workdir/semgrep" ]
+
   #
   # A2 protobufs - Detect if there is any component that has protobuf files that hasn't been compiled
   #
--- a/.expeditor/verify.pipeline.yml
+++ b/.expeditor/verify.pipeline.yml
@@ -5,19 +5,6 @@ expeditor:
         workdir: /go/src/github.com/chef/automate
 
 steps:
-
-  - label: ":semgrep: Semgrep"
-    expeditor:
-      executor:
-        docker:
-          image: returntocorp/semgrep:latest
-          entrypoint: semgrep
-          command: [
-            "--error",
-            "--config", "/go/src/github.com/chef/automate/semgrep",
-            "/go/src/github.com/chef/automate"
-          ]
-
   #
   # A2 protobufs - Detect if there is any component that has protobuf files that hasn't been compiled
   #
@@ -73,6 +60,18 @@ steps:
           environment:
             - HAB_STUDIO_SUP=false
             - HAB_NONINTERACTIVE=true
+
+  - label: ":semgrep: Semgrep"
+    expeditor:
+      executor:
+        docker:
+          image: returntocorp/semgrep:latest
+          entrypoint: semgrep
+          command: [
+            "--error",
+            "--config", "/go/src/github.com/chef/automate/semgrep",
+            "/go/src/github.com/chef/automate"
+          ]
 
   #
   # Static & Unit tests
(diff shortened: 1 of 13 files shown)


SEAjamieD (approved):
Excellent preamble.  Thank you for the extra education.  Worked well for me locally - great to have an automated extra check!

srenatus (approved):
Good job!

scottopherson (approved):
This looks great! Thanks for the excellent PR description.

This will definitely help cut down on the ":nit" and other code-formatting related review comments 😄

PR #4400, opened 2020-10-13
automate-dex: don't hardcode 24hrs expiry for saml, make it configurable

srenatus:
If unset, it's still going to be 24hrs when SAML is used. However, it can now be
overridden if desired.

If it's set to something too small, like one minute, you'll end up in a sign-in
loop.

## 🔧 

`rebuild components/automate-dex` and play with the existing id_token expiry setting, e.g.
```toml
[dex.v1.sys.expiry]
id_tokens = "10m"
[dex.v1.sys.connectors]
disable_local_users = true
```

sign in using saml, wait for 10 minutes, see what happens ✨

### DIFF
--- a/api/config/dex/config_request.go
+++ b/api/config/dex/config_request.go
@@ -41,8 +41,6 @@ func DefaultConfigRequest() *ConfigRequest {
 	c.V1.Sys.Service.Host = w.String("127.0.0.1")
 	c.V1.Sys.Service.Port = w.Int32(10117)
 
-	c.V1.Sys.Expiry.IdTokens = w.String("3m")
-
 	c.V1.Sys.Bootstrap.InsecureAdmin = w.Bool(false)
 
 	c.V1.Sys.Log.Level = w.String("info")
@@ -168,6 +166,15 @@ func (c *ConfigRequest) PrepareSystemConfig(creds *shared.TLSCredentials) (share
 
 	// default name_id_policy_format (SAML)
 	c.V1.Sys.GetConnectors().GetSaml().setNameIDPolicyDefault()
+
+	if c.V1.Sys.GetExpiry().GetIdTokens() == nil {
+		// Different defaults when SAML is or isn't used
+		if c.V1.Sys.GetConnectors().GetSaml() != nil {
+			c.V1.Sys.Expiry.IdTokens = w.String("24h")
+		} else {
+			c.V1.Sys.Expiry.IdTokens = w.String("3m")
+		}
+	}
 
 	return c.V1.Sys, nil
 }
--- a/api/config/dex/config_request.go
+++ b/api/config/dex/config_request.go
@@ -4,6 +4,7 @@ import (
 	"encoding/pem"
 	"fmt"
 	"strings"
+	"time"
 
 	google_protobuf "github.com/golang/protobuf/ptypes/wrappers"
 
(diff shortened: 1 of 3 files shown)


api/config/dex/config_request.go
```diff
@@ -41,7 +41,11 @@ func DefaultConfigRequest() *ConfigRequest {
 	c.V1.Sys.Service.Host = w.String("127.0.0.1")
 	c.V1.Sys.Service.Port = w.Int32(10117)
 
-	c.V1.Sys.Expiry.IdTokens = w.String("3m")
+	if c.V1.Sys.Connectors.Saml != nil {
```
srenatus, line 44: :x: this doesn't seem to work as intended. Ideas?

jaym: this will always be nil. this function is a constructor

srenatus: Yup. Any ideas how we could make this work, then? If SAML is configured but the expiry hasn't been set, it should default to 24h; if it wasn't configured and the expiry hasn't been set, it's 3m. If the expiry has been set, it should be whatever it's been set to.

jaym: I think `PrepareSystemConfig` is the last thing that gets called to render the config in a way the service expects. You could put it in there

msorens:
Did not seem to work for me. I rebuilt automate-dex on your branch, did a `chef-automate config patch test.toml` with the sample you provided, logged in with SAML, then waited... still logged in after 20 minutes.

srenatus:
> still logged in after 20 minutes.

Have you had your browser flicker through a non-interactive IdP page, though? That's the point. You wouldn't notice. Also, it's important that you get a new token after having applied the config: existing ID tokens are immutable.

msorens:
Ah, OK, I assumed that since this was expiring the token, it meant logout.
So, I tried again this time setting the expiry to 5 minutes. What is curious, though, watching both the network tab and the browser flicker, they occurred at about 3 minutes 45 seconds, not at the 5 minute mark.

api/config/dex/config_request.go
```diff
@@ -169,6 +167,15 @@ func (c *ConfigRequest) PrepareSystemConfig(creds *shared.TLSCredentials) (share
 	// default name_id_policy_format (SAML)
 	c.V1.Sys.GetConnectors().GetSaml().setNameIDPolicyDefault()
 
+	if c.V1.Sys.GetExpiry().GetIdTokens() == nil {
+		// Different defaults when SAML is or isn't used
+		if c.V1.Sys.GetConnectors().GetSaml() != nil {
+			c.V1.Sys.Expiry.IdTokens = w.String("24h")
+		} else {
+			c.V1.Sys.Expiry.IdTokens = w.String("3m")
+		}
```
msorens, line 176: Is this selecting 24 hours when the active connector is SAML or just when SAML is one of the possible login choices? With the name `GetConnectors().GetSaml()` it feels like that is just checking the latter, but that is not the behavior we want, right? (I tried to find the source of that code in VSCode but F12 was not being my friend.)

srenatus:
> So, I tried again this time setting the expiry to 5 minutes. What is curious, though, watching both the network tab and the browser flicker, they occurred at about 3 minutes 45 seconds, not at the 5 minute mark.

Yeah, that's got to do with the refresh happening at certain intervals, and session-service deciding "now is a good time to actually refresh" before the expiry. There's constants somewhere in session-service...

api/config/dex/config_request.go
srenatus: Re-posting what I had shared in slack for visibility:

👉If Saml is among the connectors, we need the large default. Token expiry is not per connector, unfortunately.

Also

> I tried to find the source of that code in VSCode but F12 was not being my friend.

I don't know how the indexing works, but the `GetSaml()` etc helpers are code generated from the protobuf definitions.

msorens (approved):

stevendanna (approved):
Overall this looks good.  A couple potential improvements.  

- You mention that setting this too low would be problematic.  Should we set a lower-bound in the `Validate()` function?

- In the existing code, we were setting it to "3m".  Before I thought about it much that struck me as pretty low.  Perhaps a comment explaining why that default makes sense?

srenatus:
@stevendanna good points. "3m" was selected for having fast feedback on changed LDAP groups or LDAP user status. I'll 
 add a note.

srenatus:
ℹ️ I'll make 3m the minimum.

stevendanna (approved):
Thanks!

PR #4406, opened 2020-10-16
automate-gateway: update swagger-ui to 3.35.2

stevendanna:


jaym:
Does the swagger ui still work? I was under the impression it was broken

stevendanna:
@jaym I was able to navigate to it after this change still...any pointer to why we think it doesn't work?

yzl (approved):

PR #4408, opened 2020-10-20
automate-gateway: use correct swagger configuration

stevendanna:
This got reset by re-vendoring the dependency.


stevendanna:
At the moment these swagger docs do seem broken unfortunately.

jaym (approved):

PR #4420, opened 2020-10-31
Additional semgrep validation

msorens:

_Semgrep adds checking beyond what the compiler and linter can do;
custom rules can validate Chef conventions, best practices, and the like!_

(A) Adds a few new rules shown below and applies a couple fixes stemming from them.
(B) Removes workaround from rule "combineLatest-not-combining" as bug was fixed in semgrep 0.29.0.
(C) A bit of general semgrep refinement.

**Category:** general syntax
**Rule ids:** verbose-booleans
**Detail:** returning exhaustive booleans should just return the expression instead

**Category:** rxjs subscriptions
**Rule ids:** onDestroy-not-implemented-in-class
**Detail:** subscriptions never complete because class does not implement OnDestroy Lifecycle hook

**Category:** rxjs syntax
**Rule ids:** observables-from-select-or-pipe-not-ending-with-dollar-sign,
observables-from-values-not-ending-with-dollar-sign
**Detail:** Observable variable should end with a dollar sign.

NA

### :+1: Definition of Done
Semgrep is a bit greener in buildkite.

You can run semgrep manually to confirm what buildkite does.
(1) Install from e.g., homebrew (`brew install semgrep`).
(2) `cd automate`
(3) Run `semgrep -f semgrep`
(The `-f semgrep` is specifying an eponymous directory name.)

- [x] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/Makefile
+++ b/components/automate-ui/Makefile
@@ -41,12 +41,12 @@ lint-typescript:
 
 #: Security validation via semgrep
 semgrep:
-	semgrep --config $(REPOROOT)/semgrep
+	semgrep --config $(REPOROOT)/semgrep --exclude *.spec.ts
 	semgrep --config https://semgrep.dev/p/r2c-ci
 
 #: Security validation via semgrep; autofix where possible
 semgrep-and-fix:
-	semgrep --config $(REPOROOT)/semgrep --autofix
+	semgrep --config $(REPOROOT)/semgrep --exclude *.spec.ts --autofix
 	semgrep --config https://semgrep.dev/p/r2c-ci --autofix
 
 pr-ready: unit-all-browsers lint e2e
--- a/components/automate-ui/Makefile
+++ b/components/automate-ui/Makefile
@@ -42,7 +42,7 @@ lint-typescript:
 #: Security validation via semgrep
 semgrep:
 	semgrep --config $(REPOROOT)/semgrep
-	semgrep --config https://semgrep.dev/p/r2c-ci --autofix
+	semgrep --config https://semgrep.dev/p/r2c-ci
 
 #: Security validation via semgrep; autofix where possible
 semgrep-and-fix:
--- a/semgrep/general.yaml
+++ b/semgrep/general.yaml
@@ -1,0 +1,15 @@
+rules:
+- id: verbose-booleans
+  patterns:
+    - pattern-either: 
+      - pattern: if ($EXPR) { return true; } else { return false; }
+      - pattern: if (!$EXPR) { return false; } else { return true; }
+      - pattern: if ($EXPR) { return true; } return false;
+      - pattern: if (!$EXPR) { return false; } return true;
+  message: returning exhaustive booleans should just return the expression instead
+  metadata: {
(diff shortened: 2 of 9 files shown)


components/automate-ui/Makefile
```diff
@@ -41,12 +41,12 @@ lint-typescript:
 
 #: Security validation via semgrep
 semgrep:
-	semgrep --config $(REPOROOT)/semgrep
-	semgrep --config https://semgrep.dev/p/r2c-ci --autofix
```
msorens, line 45: Autofix here was a mistake; that is just supposed to be in the `semgrep-and-fix` target.

semgrep/general.yaml
```diff
@@ -0,0 +1,15 @@
...
+      - pattern: if (!$EXPR) { return false; } else { return true; }
+      - pattern: if ($EXPR) { return true; } return false;
+      - pattern: if (!$EXPR) { return false; } return true;
+  message: returning exhaustive booleans should just return the expression instead
+  metadata: {
+      todo: "as of 0.29.0, the autofix sometimes works, and sometimes goes haywire"
+      }
+  languages: [ts]
+  severity: ERROR
+  fix: return $EXPR;
```
msorens, line 15: This was a fascinating exercise--all four bits of code in the matchers can be replaced with this simple fix. But wait, it is actually even better! To semgrep, these are the same (as they should be):
```
if ($EXPR) { return true; } else { return false; }
if ($EXPR) return true; else return false;
```
And, if you are wondering, line breaks are completely transparent, too.

semgrep/rxjs-subscription.yaml
```diff
@@ -116,3 +116,11 @@ rules:
     message: ngOnDestroy missing this.isDestroyed.complete()
     languages: [ts]
     severity: ERROR
+
+  - id: onDestroy-not-implemented-in-class
+    patterns:
+    - pattern: this.isDestroyed.complete()
+    - pattern-not-inside: class $CLASS implements OnDestroy { ... }
+    message: subscriptions never complete because class does not implement OnDestroy Lifecycle hook
```
msorens, line 124: That is, even if there is an `ngOnDestroy()` defined, it is never used without the `implements`.

semgrep/rxjs-syntax.yaml
```diff
@@ -2,11 +2,30 @@ rules:
...
     severity: WARNING
- 
\ No newline at end of file
+
+  - id: observables-from-select-or-pipe-not-ending-with-dollar-sign
+    patterns:
+    - pattern-regex: '\w+\s*(:.+?)?=.*?(select|pipe)'
+    - pattern-either:
+      - pattern-inside: $VAR = $EXPR.select(...);
+      - pattern-inside: $VAR = $EXPR.pipe(...);
```
msorens, line 15: Semgrep only resolves down to the level of language constructs. The characters that make up a variable name, though, are below that. The regex, therefore, is used to find identifiers consisting exclusively of word characters (possibly followed by a type designation)... and thus _without_ a trailing dollar sign.

msorens (commented):
Author notes

SEAjamieD (approved):
Ran through the testing section - worked great 🎉

Issue #4418, opened 2020-10-30
[infra-proxy-service] Add nodes list API


vsingh-msys:
Sub Task of: https://github.com/chef/automate/issues/4371

### User Story

- API to fetch the list of the nodes from the chef-server into the infra-proxy-service.
- As of now cookbooks, roles, environments other chef objects affected nodes API have been implemented.
- Node list attribute:
<img width="984" alt="Screen Shot 2020-09-23 at 17 27 03" src="https://user-images.githubusercontent.com/10341541/94084152-1241b400-fdc2-11ea-8ef4-ada0e53d4532.png">

PR #4422, opened 2020-11-03
[infra-proxy-service] node list HTTP & gPRC API

vsingh-msys:

- Add the node list API with attributes name, fqdn, IP address, environment, uptime and check-in time etc.
- Minor updates in affected node API to pass URL instead of name and version.

Fixes: #4418 

### :+1: Definition of Done
- Infra service responding with node list API.

Steps to build:
- `rebuild components/infra-proxy-service`

Steps to test:
- Add the chef-server test env direction given at step 3 https://github.com/chef/automate/blob/master/dev-docs/adding-data/adding_test_data.md#adding-data-to-infra-views 

- Hit the node list HTTP API

```
curl -sSkH "api-token: $(get_admin_token)" https://a2-dev.test/api/v0/infra/servers/chef-server-dev-test/orgs/chef-org-dev/nodes?pretty

{
  "nodes": [
    {
      "id": "c745433f-9517-4c9a-8525-65360bde854e",
      "name": "node-learn_chef_apache2",
      "fqdn": "vagrant.vm",
      "ip_address": "10.0.2.15",
      "check_in": "1.5901204929347618E+09",
      "uptime": "50 minutes 09 seconds",
      "platform": "ubuntu",
      "environment": "_default",
      "policy_group": ""
    }
  ]
}
``` 

- [ ] Tests added/updated?
- [ ] Docs added/updated?


api/external/infra_proxy/response/nodes.proto
```diff
@@ -13,16 +18,20 @@ message NodeAttribute {
...
+    string uptime = 6;
     // Node platform.
-    string platform = 5;
+    string platform = 7;
     // Node environment name.
-    string environment = 6;
+    string environment = 8;
     // Node policy group.
-    string policy_group = 7;
+    string policy_group = 9;
```
vsingh-msys, line 34: Re-arranged it for better readability, as these APIs have not consumed yet.

components/infra-proxy-service/server/nodes.go
```diff
@@ -12,14 +12,38 @@ import (
...
 )
 
+// GetNodes fetches the nodes from chef infra server
+func (s *Server) GetNodes(ctx context.Context, req *request.Nodes) (*response.Nodes, error) {
+	c, err := s.createClient(ctx, req.OrgId, req.ServerId)
+	if err != nil {
+		return nil, err
+	}
+
+	res, err := c.fetchAffectedNodes(ctx, "*:*")
```
vsingh-msys, line 22: Fetch node list from chef-server using search `node` index may we can refer: `knife search node "*:*"`

components/infra-proxy-service/server/nodes.go
```diff
@@ -12,14 +12,38 @@ import (
...
 
-	res, err := c.fetchAffectedNodes(ctx, req.ChefType, req.Name, req.Version)
+	var url string
+	if req.Version != "" {
+		url = fmt.Sprintf("%s_%s_version:%s", req.ChefType, req.Name, req.Version)
+	} else {
+		url = fmt.Sprintf("%s:%s", req.ChefType, req.Name)
+	}
+
+	res, err := c.fetchAffectedNodes(ctx, url)
```
vsingh-msys, line 46: Minor update to pass the constructed search URL.

api/external/infra_proxy/response/nodes.proto
vsingh-msys: Reverted it back, proto gen validation won't approve these changes

msorens: FYI: It is OK to make breaking changes when necessary, and certainly when the APIs are not yet in use. Just read the fine print on the error as to how to manage it:
```
🚨🚨🚨 If you NEED to get a breaking change through Ci: 🚨🚨🚨
add your file to the ignored files in buf.yaml and remove it in a subsequent pull request
```

vsingh-msys: Gotch, we are good for now, this would be helpful thanks!

lancewf (approved):

components/infra-proxy-service/server/nodes.go
```diff
@@ -12,14 +12,38 @@ import (
 	"github.com/chef/automate/components/infra-proxy-service/validation"
 )
 
+// GetNodes fetches the nodes from chef infra server
+func (s *Server) GetNodes(ctx context.Context, req *request.Nodes) (*response.Nodes, error) {
+	c, err := s.createClient(ctx, req.OrgId, req.ServerId)
+	if err != nil {
+		return nil, err
```
stevendanna, line 19: I wonder if we want to wrap these errors a bit to make sure we return grpc Status errors: https://pkg.go.dev/google.golang.org/grpc/status

components/infra-proxy-service/server/nodes.go
```diff
@@ -115,9 +139,11 @@ func (s *Server) UpdateNode(ctx context.Context, req *request.UpdateNode) (*resp
 // roles: it would be 'roles:ROLE_NAME'
 // environments: it would be 'chef_environment:ENVIRONMENT_NAME'
 // policyfiles: it would be 'chef_environment:POLICY_FILES_GROUP_NAME'
-func (c *ChefClient) fetchAffectedNodes(ctx context.Context, chefType, name, version string) (*chef.SearchResult, error) {
+func (c *ChefClient) fetchAffectedNodes(ctx context.Context, url string) (*chef.SearchResult, error) {
```
stevendanna, line 142: Small naming issue.  `url` here isn't really URL, it is the search query. And the variable we've called query is probably more accurately something like the "attrMap" or "attrFilter".

stevendanna (commented):
The code here mostly looks good.  My biggest question, however, is around pagination.  Do we need pagination? Do we think we will in the future?  If we think we'll need it in the future then it would be good to include it in the beginning so that we don't need a breaking API chagne.  The underlying library _appears_ to support pagination which would allow us to support it here as well in some form.

components/infra-proxy-service/server/nodes.go
vsingh-msys: Renamed `url` with `statement` and `query` with `params` seems matched with go-chef lib func signature https://github.com/go-chef/chef/blob/c278260aa167af1515bb3e2c282c2f02a2597543/search.go#L129-L138

components/infra-proxy-service/server/nodes.go
vsingh-msys: Yeah, I have moved wrapping in `proxy.go#createClient` with status.
here a bit of generalization

```
func ParseAPIError(err error) error {
	chefError, _ := chef.ChefError(err)
	if chefError != nil {
		switch chefError.StatusCode() {
		case http.StatusBadRequest:
			return status.Errorf(codes.InvalidArgument, chefError.StatusMsg())
		case http.StatusUnauthorized:
			return status.Errorf(codes.Unauthenticated, chefError.StatusMsg())
		case http.StatusForbidden:
			return status.Errorf(codes.PermissionDenied, chefError.StatusMsg())
		case http.StatusConflict:
			return status.Errorf(codes.AlreadyExists, chefError.StatusMsg())
		case http.StatusNotFound:
			return status.Errorf(codes.NotFound, chefError.StatusMsg())
		default:
			return status.Error(codes.InvalidArgument, chefError.StatusMsg())
		}
	}
```

vsingh-msys:
@stevendanna pagination stuff makes sense to me but it seems go-chef lib itself does not support the pagination first, we required to support pagination in go-chef lib then in further PR we can add it to here.

Ref: https://github.com/go-chef/chef/blob/c278260aa167af1515bb3e2c282c2f02a2597543/search.go#L129-L138

msorens (approved):
Not as familiar with all the details here, but looks OK.

PR #4427, opened 2020-11-05
Add test inspec profiles used for testing

alexpop:
Signed-off-by: Alex Pop <apop@chef.io>

Code will only change after this PR merges.The plan is to remove the dependency of our tests to some sample inspec profiles from Victoria's personal github account. For example:

https://github.com/chef/automate/blob/master/components/nodemanager-service/tests/scanjob_aws_api_integration_test.go#L88

After this merges, I'll open a PR to change URL like the one above to:

```
https://github.com/chef/automate/raw/master/components/compliance-service/test_data/inspec_profiles/test-aws-profile-2.0.0.tar.gz
```


srenatus (approved):
Thank you!

Issue #4423, opened 2020-11-04
[infra-proxy-service] Improve orgs validation using helpers. 


vsingh-msys:
# User Story

- Improve validation for orgs gRPC APIs using validation helper.

- As of now hardcoded helpers have been added:
https://github.com/chef/automate/blob/5cc1c7c365855dbac3194e102fe3711e643fe9e2/components/infra-proxy-service/server/orgs.go#L22-L45

replace it with a simple block of code

```
// Validate all request fields are required
	err := validation.New(validation.Options{
		Target:          "org",
		Request:         *req,
		RequiredDefault: true,
	}).Validate()

	if err != nil {
		return nil, err
	}
```

PR #4428, opened 2020-11-05
[infra-proxy-service] updates orgs test cases & validations

vsingh-msys:

- Used validation helpers for orgs params validations.
- Minor refactoring orgs test cases, might we can match exact string but skipping here and added similar assert check like `servers_test.go`.

Closes: https://github.com/chef/automate/issues/4423

### :+1: Definition of Done
- better & reused orgs validations.

- `make test`

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/infra-proxy-service/server/orgs.go
+++ b/components/infra-proxy-service/server/orgs.go
@@ -3,15 +3,13 @@ package server
 import (
 	"context"
 
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
-
 	"github.com/chef/automate/api/external/common/query"
 	secrets "github.com/chef/automate/api/external/secrets"
 	"github.com/chef/automate/api/interservice/infra_proxy/request"
 	"github.com/chef/automate/api/interservice/infra_proxy/response"
 	"github.com/chef/automate/components/infra-proxy-service/service"
 	"github.com/chef/automate/components/infra-proxy-service/storage"
+	"github.com/chef/automate/components/infra-proxy-service/validation"
 )
 
 // CreateOrg creates a new org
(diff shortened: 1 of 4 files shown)


components/infra-proxy-service/server/orgs_test.go
```diff
@@ -532,11 +532,22 @@ func TestOrgs(t *testing.T) {
...
+			grpctest.AssertCode(t, codes.InvalidArgument, err)
+		})
+
 		t.Run("when the org does not exists, return org not found", func(t *testing.T) {
 			ctx := context.Background()
 			orgID := "97e01ea1-976e-4626-88c8-43345c5d934f"
 			resp, err := cl.GetOrg(ctx, &request.GetOrg{
-				Id: orgID,
+				Id:       orgID,
+				ServerId: serverRes.Server.Id,
```
vsingh-msys, line 550: Added missing validation for server ID while fetching the org.

components/infra-proxy-service/server/orgs_test.go
```diff
@@ -1031,7 +1042,7 @@ func TestOrgs(t *testing.T) {
 				Projects:  []string{},
 			})
 			require.Nil(t, resp)
-			assert.Contains(t, err.Error(), "must supply org ID")
+			assert.Error(t, err, "must supply org ID")
```
vsingh-msys, line 1045: Loosely coupled assert to check the Err object only, avoiding hardcoded error message check here, any thoughts would be helpful!

components/infra-proxy-service/server/server.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/chef/automate/lib/tracing"
 )
 
-// NewGRPCServer creates a grpc server that serves all Teams GRPC APIs
+// NewGRPCServer creates a grpc server that serves all infra-proxy-service GRPC APIs
```
vsingh-msys, line 19: minor typo

stevendanna (approved):
I missed the PR for the new validation library, nice work there. Glad to see us spreading its use.

components/infra-proxy-service/server/orgs.go
```diff
@@ -3,45 +3,29 @@ package server
...
-	if req.AdminKey == "" {
-		s.service.Logger.Debug("incomplete create org request: missing org admin key")
-		return nil, status.Error(codes.InvalidArgument, "must supply org admin key")
-	}
+	// Validate all request fields are required
+	err := validation.New(validation.Options{
+		Target:          "org",
+		Request:         *req,
+		RequiredDefault: true,
+	}).Validate()
```
msorens, line 25: Looking over the validation library internals a bit, which I had not seen before. Looks like `RequiredDefault` says that every field of the `Request` must be non-empty, is that correct?
Then I am not clear on the second loop inside `Validate()` which seems to be doing the same check on `Rules` (though I do not see you using `Rules` anywhere in the codebase). 
Ideally would be good if you could add a couple comment lines inside validation.go::Validate that addresses my questions.

vsingh-msys: Yes, this validation library added to provide the validation out of the box. If `RequiredDefault` is set the all request field must be non-empty.


The second loop is the use case if the user wants to validate only specific fields to non-empty or other validations(TODO).

Name would be non-empty
```

err = validation.New(validation.Options{
		Target:          "org",
		Request:         *req,
		Rules: validation.Rules{
			"Name": []string{"required"}, // ip_address, email or multiple rules we can be add here but is in future scope.
		},
	}).Validate()
```


Will created further PRs and provides the doc & a couple of optimization for its uses. Thanks!

msorens (approved):
Looks good.

PR #4434, opened 2020-11-11
Introduce rxjs-spy diagnostic tool to UI

msorens:

rxjs-spy is a debugging library for RxJS, which is pervasive in our UI. This provides some infrastructure so you might be able to get away from a plethora of `console.log` statements when trying to diagnose an errant behavior.
With rxjs-spy you can add a simple tag to any observable in your code, then in the browser you can watch any or all tagged observables receiving values, being subscribed, and being unsubscribed.

Project: https://github.com/cartant/rxjs-spy
More Details: https://ncjamieson.com/debugging-rxjs-part-1-tooling/

### :+1: Definition of Done
rxjs-spy is available for use in development (not production).

1. Load the new package dependency on your machine: `npm install`.
2. Launch automate-ui using your method of choice (e.g. `make serve`).
3. Open automate-ui in your browser (https://a2-dev.test/).
4. Open the console in the browser's dev tools (<kbd>option</kbd> + <kbd>command</kbd> + <kbd>J</kbd>).
5. Enable rxjs-spy with this new command in the console: `enableRxjsSpy()`.

Now you can use methods on the `spy` object to activate/deactivate logging, show observables, and more (see https://github.com/cartant/rxjs-spy#module-api).

#### Sample session

We have not yet done anything so there is nothing to see:
```
> spy.show()
0 snapshot(s) matching /.+/
```

Navigate to Settings >> Teams and select any team to open its details page.
Now try the same call and you will see all the tagged observables. Here I have expanded one of them:
image
In there you will see that there is 1 subscriber; that the subscription is incomplete (i.e., the subscription is still active), and the value from the routeURL was `/settings/teams/t3`.

Go back to the parent teams page, run `spy.show()` again and you will see that the **State** has changed to `complete` and a new property, **Unsubscribed**, is `true`, and similarly for the other observables, so the observable lifecycle was properly cleaned up.

The `spy.show()` method also can take an  argument, either a string or a regular expression, to filter the output. Examples: `spy.show('team-routeUrl')` to see that single observable or `spy.show(/team-/)` to see all observables tagged with that prefix (which only coincidentally happens to be all of them at the moment).

Do the same exercise of viewing team details again, but this time witih logging. As setup just to be a bit more interesting, add a couple users to the team from the team details page, then return to the starting point at Settings >> Teams. The `spy.log()` method takes the same kind of argument. I will use that in this next example. Here I start on the Teams page after having added two users to the team, go into the details page, then back to the Teams page. Observe that I opened up the `team-users` tag so you actually see those two users.

image

If you want to deactivate any particular log that you previously activated, run `spy.undo()` to list things that you can undo. Here you will see the two activating loggers; I pick one to deactivate then list again and you can see that it is gone.

image

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/automate-ui/package-lock.json
+++ b/components/automate-ui/package-lock.json
@@ -2153,6 +2153,12 @@
         }
       }
     },
+    "@types/circular-json": {
+      "version": "0.4.0",
+      "resolved": "https://registry.npmjs.org/@types/circular-json/-/circular-json-0.4.0.tgz",
+      "integrity": "sha512-7+kYB7x5a7nFWW1YPBh3KxhwKfiaI4PbZ1RvzBU91LZy7lWJO822CI+pqzSre/DZ7KsCuMKdHnLHHFu8AyXbQg==",
+      "dev": true
+    },
     "@types/color-name": {
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/@types/color-name/-/color-name-1.1.1.tgz",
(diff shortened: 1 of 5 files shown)


lancewf (approved):
Thank you for the overview of this new debugging feature. 
So, there is no problem with having these tags in production?

msorens:
That is correct, @lancewf . I mentioned that in one of my commit messages but for convenience here it is:

These tags do not impact performance, per https://ncjamieson.com/debugging-rxjs-part-1-tooling/
"The tag operator can be used alone — import "rxjs-spy/add/operator/tag" — and the other rxjs-spy methods can be omitted from production builds, so the only overhead is the string annotations."

PR #4436, opened 2020-11-13
Apply default TLS cipher suite to nats-gateway

stevendanna:
This got picked up in a recent security scan. Someone familiar with
the application should review to ensure there wasn't a reason to allow
the less restrictive cipher suite.

Note that this still leaves the NATS ports 4222 and 10140 outside the
view of the scanner as the scanner says those don't appear to be TLS
ports.

Further this removes cipher suites with RSA key exchange from Erlang and Go defaults.

This keeps us in line with the "Intermediate" compatibility from
Mozilla. Further, because these are mostly used by internal
communications, it is likely we were never actually using these
ciphers, despite them coming up occasionally on scans.


stevendanna:
Looks like we have a few other violations that the scanner caught in CI that I didn't get locally.

stevendanna:
https://go-review.googlesource.com/c/go/+/175517 has made it a bit easier to make these internal cipher suites configurable, but I'm still -1 on them being configurable since in reality almost all of this communication should be happening over TLS1.3 where they aren't configurable anyway.

jaym (approved):

Issue #4366, opened 2020-09-22
Node client run history pagination selector disappears when accessing beyond end of multi-day history


jraidan:
## Describe the bug
From customer:
"When accessing multi-day node client run history, it seems when the selection goes past what is in the database, the pagination selector at the bottom disappears. The only fix is to refresh the page and then go back to the end.  This is annoying when we want to see the “beginning of history” on our recently Provisioned servers from a week or two ago.. We select 1 month then go through 20 pages then find the end and can’t change the selection. Then refresh and go one page at a time not to go past the last page."

## To Reproduce
- Navigate to the client run for a node, uptime 15 days for our repro case
- Click "Run History" blue button on right top
- Use nav buttons at bottom to get to last available page (page 8 for us), the dropdown selection is at 24 hrs (Default?)
- Select an older selection from drop-down, eg "Last month", Click "Apply"
- Use nav buttons at bottom to get to the last available page (page 12 now)
- The selection window closes
- Re-open the select window with "Run History" button
- The bottom pagination selector is now gone until you reload/refresh the page.

## Expected behavior
The "Run History" bottom page nav shouldn't poof.

## Versions (please complete the following information):
- Browser: Chrome
- Automate Build Number: 20200728181447

stevenoneill:
I experienced this one just earlier today.

susanev:
thanks for reporting this @jraidan!! weve pulled it into the workload for our ui team and they should be able to address it soon.

SEAjamieD:
I thought I had a fix for this today, but ultimately realized the issue is deeper in with the math involved for the optional `maxPageItems` property.  We haven't noticed it before because its the only place in the codebase were we use this property.  I'll also double check that this NOT being used isn't causing other pagination components to pop out of their containers.

PR #4441, opened 2020-11-16
Jamie/4366 disappearing pagination

SEAjamieD:
A customer reported that their pagination was disappearing/closing their sidebar on client runs > run history.  This turned out to be 2 different bugs.

1.  When there was only 1 selectable page left to view, the pagination component would disappear.  For example, if a user has 11 pages of content, on a default pagination component we show 10 selectable pages at a time.  When they get to page 11 - there is only 1 page available, so the page hides.  In this case, it would also cause the sidebar to close.

2.  The selectable pages were being calculated off the amount of selectable content, in this particular case - when the content is filtered, the page picker didn't know how to figure out what the current page was, so it was calculating against pages that didn't exist, which would lead the user to be able to select pages that had no content, which would then cause the pagination component to disappear - and in this case, the sidebar to close.

Fixes: https://github.com/chef/automate/issues/4366

### :+1: Definition of Done
Pagination component works as expected.  
1. Shows when there is more than one page of total selectable content.
2. Resets when filters are added.
3. Shows the current page after reset has been done.

cd Automate
`build components/automate-ui-devproxy && start_all_services`

cd Automate/components/automate-ui
`make serve`

in Hab Studio, load up a node with lots of client run history
`for i in {1..105}; do chef_load_nodes 1; done`

navigate to https://a2-dev.test/infrastructure/client-runs
if a fresh install, click on the only node available
if not a fresh install, likely search for the node named `chef-load-1`

on the single node page, click on the run history button in the top right corner.
navigate through the pagination on the bottom of the run history slideout.
use the filters on the top to change the pagination.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

Client Run history pagination

### DIFF
--- a/components/automate-ui/src/app/components/page-picker/page-picker.component.html
+++ b/components/automate-ui/src/app/components/page-picker/page-picker.component.html
@@ -1,4 +1,4 @@
-<ng-container *ngIf="selectablePages.length > 1">
+<ng-container *ngIf="selectablePages.length > 1 || last > 1">
   <ng-container *ngIf="!forDesktop">
     <chef-button secondary
       class="page-picker-item page-picker-first"
--- a/components/automate-ui/src/app/components/page-picker/page-picker.component.html
+++ b/components/automate-ui/src/app/components/page-picker/page-picker.component.html
@@ -17,7 +17,7 @@
       class="page-picker-item"
       [attr.primary]="item === page"
       [attr.secondary]="item !== page"
-      [attr.disabled]="selectablePages.length <= 1"
+      [attr.disabled]="selectablePages.length <= 1 && last <= 1"
       (click)="onItemTap(item)">
       {{ item }}
     </chef-button>
--- a/components/automate-ui/src/app/components/page-picker/page-picker.component.ts
+++ b/components/automate-ui/src/app/components/page-picker/page-picker.component.ts
@@ -13,7 +13,7 @@ export class PagePickerComponent implements OnChanges {
   @Input() total: number;
   @Input() perPage: number;
   @Input() page: number;
-  @Input() maxPageItems = 10;
+  @Input() maxSelectablePages = 10;
   @Input() @HostBinding('class.forDesktop') forDesktop = false;
   @Input() @HostBinding('class.fullScreened') fullScreened = false;
 
@@ -59,7 +59,7 @@ export class PagePickerComponent implements OnChanges {
     this.next = (this.page === this.last) ? null : this.page + 1;
 
     const pages = [];
-    const selectedIndex = ((this.page - 1) % this.maxPageItems);
+    const selectedIndex = ((this.page - 1) % this.maxSelectablePages);
 
     for (let pageIndex = selectedIndex; pageIndex >= 0; --pageIndex) {
       const prev = this.page + (pageIndex - selectedIndex);
@@ -68,7 +68,7 @@ export class PagePickerComponent implements OnChanges {
 
     pages[selectedIndex] = this.page;
 
-    for (let pageIndex = selectedIndex; pageIndex < this.maxPageItems; ++pageIndex) {
+    for (let pageIndex = selectedIndex; pageIndex < this.maxSelectablePages; ++pageIndex) {
       const next = this.page + (pageIndex - selectedIndex);
       if (next <= this.last) {
         pages[pageIndex] = next;
--- a/components/automate-ui/src/app/components/page-picker/page-picker.component.ts
+++ b/components/automate-ui/src/app/components/page-picker/page-picker.component.ts
@@ -54,6 +54,7 @@ export class PagePickerComponent implements OnChanges {
 
   ngOnChanges() {
     this.last = Math.ceil(this.total / this.perPage) || 1;
+    this.page = this.page > this.last ? this.last : this.page;
     this.prev = (this.page === this.first) ? null : this.page - 1;
     this.next = (this.page === this.last) ? null : this.page + 1;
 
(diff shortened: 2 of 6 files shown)


components/automate-ui/src/app/components/page-picker/page-picker.component.html
```diff
@@ -1,4 +1,4 @@
-<ng-container *ngIf="selectablePages.length > 1">
+<ng-container *ngIf="selectablePages.length > 1 || last > 1">
```
SEAjamieD, line 1: It's possible that sometimes we have a scenario with only 1 selectable page, however we're looking at say...page 11.  This makes sure that if we have only one selectable page, but more than 1 page in total, we still see the pagination component.

@msorens - thinking now that I probably need to pipe this, huh?

components/automate-ui/src/app/page-components/run-history/run-history.component.scss
```diff
@@ -180,6 +180,5 @@ app-date-selector {
 }
 
 .history-pagination {
-  margin-bottom: 1em;
-  justify-content: center;
+  padding: 1em;
```
SEAjamieD, line 183: Small styling update - gives the pagination component a little breathing room.

components/automate-ui/src/app/components/page-picker/page-picker.component.ts
```diff
@@ -54,6 +54,7 @@ export class PagePickerComponent implements OnChanges {
 
   ngOnChanges() {
     this.last = Math.ceil(this.total / this.perPage) || 1;
+    this.page = this.page > this.last ? this.last : this.page;
```
SEAjamieD, line 57: This was the main culprit, the current page data wasn't being updated, which was throwing off the math below creating selectable pages that didn't actually exist.  This makes sure that if we're dropping to a fewer amount of pages, the current page is adjusted to reflect that change.

components/automate-ui/src/app/components/page-picker/page-picker.component.html
msorens: No, using properties in expressions local to the template like this is fine; it is only when you introduce a function call that a pipe would improve performance.

components/automate-ui/src/app/components/page-picker/page-picker.component.ts
msorens: Good, but can simplify a bit:
```suggestion
    this.page = Math.min(this.page, this.last);
```

msorens (commented):
I am not quite following your "case 1". You indicate there is a default of 10 pages but I am only seeing 3 at a time (and your video also shows only 3). And I could not get the picker to show only page 11 by itself in order to test the fix for case 1.

SEAjamieD:
> I am not quite following your "case 1". You indicate there is a default of 10 pages but I am only seeing 3 at a time (and your video also shows only 3). And I could not get the picker to show only page 11 by itself in order to test the fix for case 1.

@msorens - I was speaking in general about the page-picker component.  By default it sets the amount of selectable pages as 10, however in this specific case, we have it set to max out at 3 pages.  You can see the issue if you remove the `maxSelectablePages` property in the `run-history.html` page.  

after looking at your comment above, I updated the property to be named `maxSelectablePages` so that it was more easily understandable.
bug

msorens:
Thanks for the additional details, @SEAjamieD. Looks one a minor issue: when I revert to the default of 10 on your branch I am seeing this--note that the page 11 icon is disabled and it should not be:
image

SEAjamieD:
> Thanks for the additional details, @SEAjamieD. Looks one a minor issue: when I revert to the default of 10 on your branch I am seeing this--note that the page 11 icon is disabled and it should not be:
> image

@msorens - Really fantastic catch there - thank you!

Shadae (approved):
:+1:

components/automate-ui/src/app/components/page-picker/page-picker.component.html
```diff
@@ -17,7 +17,7 @@
       class="page-picker-item"
       [attr.primary]="item === page"
       [attr.secondary]="item !== page"
-      [attr.disabled]="selectablePages.length <= 1"
+      [attr.disabled]="selectablePages.length <= 1 && last <= 1"
```
SEAjamieD, line 20: Similar to the addition above, this makes sure that the page selector button is _not_ disabled when we hit a scenario where there is only one selectable page showing, but may still have multiple pages.

Shadae: Nice!

components/automate-ui/src/app/components/page-picker/page-picker.component.ts
```diff
@@ -54,24 +54,17 @@ export class PagePickerComponent implements OnChanges {
...
-
-    for (let pageIndex = selectedIndex; pageIndex < this.maxPageItems; ++pageIndex) {
-      const next = this.page + (pageIndex - selectedIndex);
-      if (next <= this.last) {
-        pages[pageIndex] = next;
-      }
+    for (let pageIndex = 0, pg = this.page - selectedIndex;
+      pageIndex < this.maxSelectablePages && pg <= this.last;
+      ++pageIndex, pg = this.page + pageIndex - selectedIndex) {
+      pages[pageIndex] = pg;
```
msorens, line 67: Trying to understand all the nuances and edges of the original code was frazzling my brain; just had to tidy it up here (and add some test cases proving it is equivalent!).

msorens (approved):
Working as advertised!
While I was making sure I understood the code I took the liberty of tidying up some nearby calculations; hope you do not mind.

components/automate-ui/src/app/components/page-picker/page-picker.component.ts
SEAjamieD: amazing cleanup, thanks! 🎉   And tests always apreciated!

PR #4448, opened 2020-11-20
Add UTC label in the date picker

alexpop:
# :nut_and_bolt: Description: What code changed, and why?

After merging #4358 some users might get confused between local time selections (`Last 24 hours`) and the places where we select/show a day which is UTC.

As can be seen in the below git that Scott uploaded in the UI pull request:

In the same dropdown the user can choose between `Last 24 hours` in the local timezone to a DAY, without any hint that this day is not in their timezone, but in UTC.

In an effort to make this UTC aspect more clear, I'm submitting this PR with two changes as can be seen in the screenshots below:

---

## During date selection:

pick-day-utc

---

## After day selection from date picker: 

day-picked

### DIFF
--- a/components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.html
+++ b/components/automate-ui/src/app/pages/+compliance/+reporting/reporting-searchbar/reporting-searchbar.component.html
@@ -70,7 +70,7 @@
       <span *ngIf="last24h">Last 24 hours</span>
       <ng-container *ngIf="!last24h">
         <chef-icon>date_range</chef-icon>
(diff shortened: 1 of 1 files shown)


jonong1972:
@alexpop  yeah, Thats pretty much what I was thinking. just adding that  (UTC) to the date after selected.
Looks good to me for MVP.

jonong1972:
Actually one thing. Camel Case the "PICK DATE"    Maybe  "Choose Date"

jonong1972 (approved):
Looks good to me

alexpop:
Thank you Jon for the review, I pushed another commit to replace `PICK DATE (UTC)` with `Choose Date (UTC)`

PR #4452, opened 2020-11-25
Jamie/4343 connected services filter

SEAjamieD:
with the addition of https://github.com/chef/automate/pull/4339, we have the backend ability to filter service groups by only connected services.  This branch adds this ability to the UI

small UX bonus - hide the checkbox for deleting selected services when there all no services to select from

Backend Work: https://github.com/chef/automate/pull/4339

### :+1: Definition of Done
User is able to filter on connected services in services sidebar.

Terminal 1 - in Automate root folder
`hab studio enter`
`build components/automate-ui-devproxy`
`applications_populate_database`

Terminal 2 - cd into components/automate-ui
`make serve`

Navigate to https://a2-dev.test/applications/service-groups
in the services sidebar, see the "connected" option in the group of filter options
Click on the connected services option to see it filter

- [ ] Tests added/updated?
- [ ] Docs added/updated?

After:
Screen Shot 2020-11-25 at 11 32 48 AM

### DIFF
--- a/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
+++ b/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
@@ -58,7 +58,7 @@
         </chef-option>
       </chef-status-filter-group>
     </div>
-    <div class="bulk-action-bar">
+    <div class="bulk-action-bar" *ngIf="(services$ | async)?.length !== 0">
       <chef-checkbox
         [checked]="hasAllSelected"
         [indeterminate]="isIndeterminate"
--- a/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
+++ b/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
@@ -58,7 +58,7 @@
         </chef-option>
       </chef-status-filter-group>
     </div>
-    <div class="bulk-action-bar">
+    <div class="bulk-action-bar" [class.hidden]="(services$ | async)?.length === 0">
       <chef-checkbox
         [checked]="hasAllSelected"
         [indeterminate]="isIndeterminate"
--- a/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
+++ b/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
@@ -58,7 +58,7 @@
         </chef-option>
       </chef-status-filter-group>
     </div>
-    <div class="bulk-action-bar" [class.hidden]="(services$ | async)?.length === 0">
+    <div class="bulk-action-bar" [class.hidden]="(services$ | async)?.length === 0" [attr.aria-hidden]="(services$ | async)?.length === 0">
       <chef-checkbox
         [checked]="hasAllSelected"
         [indeterminate]="isIndeterminate"
--- a/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
+++ b/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
@@ -46,7 +46,6 @@
           <div class="filter-label">Unknown</div>
           <div class="filter-total">{{ serviceGroupsHealthSummary.unknown }}</div>
         </chef-option>
-        <br />
         <chef-option class="filter connected" value='connected' (click)="updateHealthFilter('connected')" (keyup.enter)="updateHealthFilter('connected')">
           <chef-icon class="filter-icon">link</chef-icon>
           <div class="filter-label">Connected</div>
--- a/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
+++ b/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
@@ -58,7 +58,7 @@
         </chef-option>
       </chef-status-filter-group>
     </div>
-    <div class="bulk-action-bar" *ngIf="(services$ | async)?.length !== 0">
+    <div class="bulk-action-bar">
       <chef-checkbox
         [checked]="hasAllSelected"
         [indeterminate]="isIndeterminate"
--- a/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
+++ b/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
@@ -46,6 +46,7 @@
           <div class="filter-label">Unknown</div>
           <div class="filter-total">{{ serviceGroupsHealthSummary.unknown }}</div>
         </chef-option>
+        <div class="spacer"></div>
         <chef-option class="filter connected" value='connected' (click)="updateHealthFilter('connected')" (keyup.enter)="updateHealthFilter('connected')">
           <chef-icon class="filter-icon">link</chef-icon>
           <div class="filter-label">Connected</div>
--- a/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
+++ b/components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
@@ -21,38 +21,38 @@
   <div *ngIf="serviceGroupsId && (serviceGroupsStatus$ | async) === 'loadingSuccess'">
     <div class="status-filter-bar">
       <chef-status-filter-group [value]="selectedHealth" lean>
-        <chef-option class="filter general" value="total" (click)="updateHealthFilter('total')" (keyup.enter)="updateHealthFilter('total')" selected>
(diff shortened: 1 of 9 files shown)


components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
```diff
@@ -51,9 +46,19 @@
...
+        <chef-option class="filter connected" value='connected' (click)="updateHealthFilter('connected')" (keyup.enter)="updateHealthFilter('connected')">
+          <chef-icon class="filter-icon">link</chef-icon>
+          <div class="filter-label">Connected</div>
+          <div class="filter-total">{{ serviceGroupsHealthSummary.total - serviceGroupsHealthSummary.disconnected }}</div>
+        </chef-option>
+        <chef-option class="filter disconnected" value='disconnected' (click)="updateHealthFilter('disconnected')" (keyup.enter)="updateHealthFilter('disconnected')">
+          <chef-icon class="filter-icon"></chef-icon>
+          <div class="filter-label">Disconnected</div>
+          <div class="filter-total">{{ serviceGroupsHealthSummary.disconnected }}</div>
+        </chef-option>
```
SEAjamieD, line 58: Moved these to the end of the group, per design spec

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
```diff
@@ -51,9 +46,19 @@
...
       </chef-status-filter-group>
     </div>
-    <div class="bulk-action-bar">
+    <div class="bulk-action-bar" [class.hidden]="(services$ | async)?.length === 0" [attr.aria-hidden]="(services$ | async)?.length === 0">
```
SEAjamieD, line 61: I noticed this checkbox/button was still visible even when there are no services to be selected.  This just hides the checkbox/button for UX purposes when theres none to be selected anyways.

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.scss
```diff
@@ -38,6 +38,10 @@ chef-alert {
   justify-content: left;
 }
 
+chef-option.filter.unknown {
+  margin-right: 50px; // forces remaining filters into new row
+}
+
```
SEAjamieD, line 44: per the new design spec for the connected filter option, this is a simple way to force the new row to happen, placing connected and disconnected on their on row.

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
```diff
@@ -51,9 +46,19 @@
           <div class="filter-label">Unknown</div>
           <div class="filter-total">{{ serviceGroupsHealthSummary.unknown }}</div>
         </chef-option>
+        <chef-option class="filter connected" value='connected' (click)="updateHealthFilter('connected')" (keyup.enter)="updateHealthFilter('connected')">
+          <chef-icon class="filter-icon">link</chef-icon>
+          <div class="filter-label">Connected</div>
+          <div class="filter-total">{{ serviceGroupsHealthSummary.total - serviceGroupsHealthSummary.disconnected }}</div>
```
SEAjamieD, line 52: the connected count isn't in the API yet, however its just the total minus the disconnected, so we can calculate this on the frontend for the time being.

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.scss
danielsdeleo: The other option is to put in an extra div with 0 height and 100% width which effectively works like ye olde `<br>`, is that true?

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
danielsdeleo: 💯

danielsdeleo (approved):
LGTM

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.scss
msorens: My concern here is that you are manipulating the line assuming a constant panel width. If that panel is changed in the future the desired behavior here might be lost. So I would also lean toward something that forces an actual line break.

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
```diff
@@ -51,9 +46,19 @@
           <div class="filter-label">Unknown</div>
           <div class="filter-total">{{ serviceGroupsHealthSummary.unknown }}</div>
         </chef-option>
+        <chef-option class="filter connected" value='connected' (click)="updateHealthFilter('connected')" (keyup.enter)="updateHealthFilter('connected')">
```
msorens, line 49: Would improve the code to tighten the signature of `updateHealthFilter` to accept only `ServiceGroupsFacadeService.allowedStatus` (plus `"total"` with a union type) instead of a general string... but that would require changing `allowedStatus` to be an enum, I suspect, so may be out of scope for this PR. 🤷

msorens (approved):
Just a couple notes.

msorens:
BTW, I would also suggest setting up some persistent data as you have outlined here but on an acceptance box (e.g. a2-local-inplace-upgrade-acceptance.cd.chef.co) so this could be exercised there as well.

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.scss
SEAjamieD: @danielsdeleo - you absolutely can 👍 
@msorens - I see your point, my goal was to add less code, and given that there are no current plans to change anything, simple CSS could give us what we needed.

I'm going to change this to Dan's method above and we'll be all set 👍

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
SEAjamieD: might as well update it while i'm in here.  I've updated now and would appreciate your feedback prior to merging

msorens: Getting there, but we really only need a single source of truth.
Should be able to eliminate `AllowedHealthStatus` completely and just use the `allowedStatus` enum directly (should capitalize to `AllowedStatus` though). Also, should then delete the entire line `health = includes(health, this.allowedHealthStatus) ? health : allowedStatus.TOTAL;` in `updateHealthFilter`; no longer need to check that constraint.

SEAjamieD: Thanks for the thoughts - I don't think I realized enums could be used as a type signature 👍

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
```diff
@@ -21,39 +21,45 @@
   <div *ngIf="serviceGroupsId && (serviceGroupsStatus$ | async) === 'loadingSuccess'">
     <div class="status-filter-bar">
       <chef-status-filter-group [value]="selectedHealth" lean>
-        <chef-option class="filter general" value="total" (click)="updateHealthFilter('total')" (keyup.enter)="updateHealthFilter('total')" selected>
```
msorens, line 24: Excellent; was hoping you would notice this improvement! 😉

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
```diff
@@ -21,39 +21,45 @@
...
+        <chef-option class="filter general" value="total" (click)="updateHealthFilter(allowedStatus.TOTAL)" (keyup.enter)="updateHealthFilter(allowedStatus.TOTAL)" selected>
           <chef-icon class="filter-icon">grain</chef-icon>
           <div class="filter-label">Total</div>
```
msorens, line 26: Contrast this with the change above, though just food for thought--you could also do this:
```suggestion
          <div class="filter-label">{{allowedStatus.TOTAL | capitalize}}</div>
```
But I am not really suggesting this; here "Total" is pure presentation so to me is fine as is, whereas on L24 that is code so should be using `allowedStatus.TOTAL` as you have done.

msorens (approved):
Nice!

components/automate-ui/src/app/page-components/services-sidebar/services-sidebar.component.html
SEAjamieD: food for thought, thanks for pointing it out

PR #4454, opened 2020-11-30
Resolve config parser

kalroy:
### :nut_and_bolt: Description: Added parser to parse resolv.conf file or similar file to get nameservers
This is required for the customer issue 268

### :+1: Definition of Done 
The nameservers should be fetched from /etc/resolv.conf file

- [x] Tests added/updated?
- [ ] Docs added/updated?


lib/config/configtest/resolve_config_parser_test.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+func TestParseNameServers(t *testing.T) {
+	filecontent := []byte(`
+	nameserver 10.0.0.3
+	`)
+
+	ns := config.ParseNameServers(filecontent)
+	expectedResult := []string{"10.0.0.3"}
+	if !reflect.DeepEqual(expectedResult, ns) {
```
stevendanna, line 17: We tend to use two small assertion libraries in our test that help with some of this:

https://godoc.org/github.com/stretchr/testify/assert
https://godoc.org/github.com/stretchr/testify/require

kalroy: Added "https://godoc.org/github.com/stretchr/testify/require" as the assertion library as suggested.

lib/config/configtest/resolve_config_parser_test.go
```diff
@@ -0,0 +1,85 @@
+package configtest
```
jaym, line 1: this file should be with the resolve_config_parser.go file. You can name the package `config_test` and go will allow that

lib/config/configtest/sample_resolv_conf
```diff
@@ -0,0 +1,19 @@
+# This file is managed by man:systemd-resolved(8). Do not edit.
```
jaym, line 1: the go pattern is to put such files in a `testdata` folder

lib/config/configtest/resolve_config_parser_test.go
```diff
@@ -0,0 +1,85 @@
...
+	ns := config.ParseNameServers(filecontent)
+	expectedResult := []string{"10.0.0.3"}
+
+	require.Equal(t, len(expectedResult), len(ns))
+
+	for _, m := range expectedResult {
+		require.Contains(t, ns, m)
+	}
+}
+
```
jaym, line 69: maybe add an ipv6 test?

lib/config/resolve_config_parser.go
```diff
@@ -0,0 +1,49 @@
...
+	for _, currentLine := range fileLines {
+		var contentToParse = currentLine
+		var commentIndicatorIndex = bytes.Index(currentLine, []byte("#"))
+		if commentIndicatorIndex != -1 {
+			// Only check the content before the comment section
+			contentToParse = currentLine[:commentIndicatorIndex]
+		}
+		server := nameServerRegex.FindSubmatch([]byte(contentToParse))
+
+		if len(server) > 0 {
```
jaym, line 43: should this be `== 2` or `> 1`?

lib/config/configtest/resolve_config_parser_test.go
kalroy: You mean to keep the file at the same level as resolve_config_parser.go? I wanted to keep the tests and related data separate.

lib/config/configtest/resolve_config_parser_test.go
kalroy: Although we are not validating the nameserver tests, but we can add to check the regex. Will do so.

lib/config/configtest/resolve_config_parser_test.go
jaym: thats correct. the test file and actual code go at the same level. go will let you append `_test` to the package name for the test file

lib/config/resolve_config_parser.go
kalroy: The regex `nameServerRegex` written has subexpression should always return either 0 ( in case of failure) or 2 ( in case of success ). We can set as 2 to make sure that it should not fail in case the subexpression is removed from the regular expression.

lib/config/configtest/resolve_config_parser_test.go
kalroy: The latest commit has the changes as suggested.

stevendanna (approved):
Thanks for working on this and adding test coverage.

We'll integrate this into the actual configuration in a follow-up PR.

PR #4457, opened 2020-12-01
esgateway: better support for DNS based external ES endpoints

stevendanna:
This change allows nginx to periodically re-resolve the DNS entries
for ES upstream servers. Specifically, we hope this will address ES
outages caused by cluster changes when AWS Elasticsearch.

This involves:

- Building nginx from source to include a new nginx module

- Adding a new `resolvers` configuration item to the esgateway
  configuration

- Modify template to use the jdomain module when a resolver is
  configured.

- Automatic resolver discovery that we plan to use in the future.

Users who want to try this new features, can opt-into it by specifying
the following configuration:

```
[esgateway.v1.sys.ngx.main]
# Multiple resolvers can be specified by
# a space-separated list.
resolvers = "127.0.0.53"
```

Follow-up work includes:

- Adding support to max_fails to the jdomain nginx module

- Adding support for deferred resolution to the jdomain nginx
  module (the module still uses getaddrinfo synchronously for the
  first lookup)

- Enabling the automatic resolver discovery once we are confident in
  the module and its features.


api/config/shared/global.proto
```diff
@@ -68,11 +68,18 @@ message External {
...
+      google.protobuf.StringValue address = 2;
+      google.protobuf.StringValue port = 3;
+    }
+    
     google.protobuf.BoolValue enable = 1;
     repeated google.protobuf.StringValue nodes = 2;
     Backup backup = 3;
     Authentication auth = 4;
     SSL ssl = 5;
+    repeated Endpoint parsed_endpoints = 6;
```
jaym, line 82: is this needed here? I don't see it being used anywhere

components/automate-es-gateway/habitat/config/nginx.conf
```diff
@@ -62,7 +78,7 @@ http {
     ssl_certificate_key {{pkg.svc_config_path}}/service.key;
     ssl_client_certificate {{pkg.svc_config_path}}/root_ca.crt;
     ssl_verify_client on;
-    ssl_verify_depth {{ cfg.ngx.http.ssl_verify_depth }};
+    ssl_verify_depth 7{{ cfg.ngx.http.ssl_verify_depth }};
```
jaym, line 81: what's the 7 for

jaym:
maybe leave a sample config in the pr description

api/config/esgateway/config_request.proto
```diff
@@ -79,11 +80,19 @@ message ConfigRequest {
 			message External {
 				google.protobuf.BoolValue enable = 1;
 				google.protobuf.BoolValue ssl_upstream = 2;
+				// TODO(ssd) 2020-12-01: Mark as deprecrated?
```
msorens, line 83:
```suggestion
				// TODO(ssd) 2020-12-01: Mark as deprecated?
```

api/config/esgateway/config_request_test.go
```diff
@@ -88,6 +95,176 @@ func TestExternalElasticsearch(t *testing.T) {
...
+			},
+		})
+
+		assert.True(t,
+			c.GetV1().GetSys().GetExternal().GetEnable().GetValue(),
+			"expected external ES to be enabled")
+
+		assert.True(t,
+			c.GetV1().GetSys().GetExternal().GetSslUpstream().GetValue(),
+			"expected http")
```
msorens, line 152:
```suggestion
			"expected https")
```

api/config/esgateway/config_request_test.go
```diff
@@ -88,6 +95,176 @@ func TestExternalElasticsearch(t *testing.T) {
...
```
msorens, line 236:
```suggestion
			"expected https")
```

api/config/esgateway/config_request_test.go
```diff
@@ -88,6 +95,176 @@ func TestExternalElasticsearch(t *testing.T) {
...
```
msorens, line 264:
```suggestion
			"expected https")
```

msorens (commented):
Just a couple general fixes noted.

yzl:
I think a sample config sounds like a good idea.

I hate to ask, but who is going to need to know about this?

stevendanna:
> I hate to ask, but who is going to need to know about this?

Until we are more comfortable with the module we are using here, we'd like this to remain a little-used option that can be tested by users experiencing issues.

PR #4474, opened 2020-12-11
deployment-service/ci: retry file-access errors when waiting for upgrade

stevendanna:
During upgrades from old versions, we may require a deployment-service
certificate regeneration. If the chef-automate command line tool
attempts to read the deployment-service certificate during this
regeneration, it may receive a FileAccessError such as:

    FileAccessError: Unable to access the file or directory: Failed to
    read deployment-service TLS certificates: Could not read the
    service cert: open
    /hab/svc/deployment-service/data/deployment-service.crt: no such
    file or directory

Since we expect this error to be transient, we now retry this error in
the wait_for_upgrade loop in CI. We also log a bit more information
about the deployment-service certificate regen.

### DIFF
--- a/integration/helpers/deployment.sh
+++ b/integration/helpers/deployment.sh
@@ -76,6 +76,9 @@ wait_for_upgrade() {
             0)
                 :
                 ;;
+            90)
+                log_warning "File access error for chef-automate upgrade status: likely cause deploymet-service certificate regen"
+                ;;
             98|99)
                 log_warning "Error calling deployment service"
                 ;;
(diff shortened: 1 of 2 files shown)


integration/helpers/deployment.sh
```diff
@@ -76,6 +76,9 @@ wait_for_upgrade() {
             0)
                 :
                 ;;
+            90)
+                log_warning "File access error for chef-automate upgrade status: likely cause deploymet-service certificate regen"
```
blakestier, line 80:
```suggestion
                log_warning "File access error for chef-automate upgrade status: likely cause is deployment-service certificate regen"
```

blakestier (approved):
I feel like i've been waiting my whole life for this change 😂

PR #4480, opened 2020-12-14
Create protobuf annotations for secrets and generate code

jaym:


components/automate-grpc/protoc-gen-a2-config/gen/service_config.go
```diff
@@ -385,3 +437,90 @@ func (m *A2ServiceConfigModule) generateGetPortMethod(message pgs.Message, acc *
...
+
+func (m *A2ServiceConfigModule) generateGetSecretMethod(message pgs.Message, acc *Acc) *jen.Statement {
+	// func (m *ConfigRequest) GetPort(name string) (uint16, error)
+	f := jen.Func().Params(
+		jen.Id("m").Id("*" + m.ctx.Name(message).String()),
+	).Id("GetSecret").Params(
+		jen.Id("name").String(),
+	).Params(jen.Op("*").Qual("github.com/golang/protobuf/ptypes/wrappers", "StringValue")).BlockFunc(func(g *jen.Group) {
+		// switch name {
+		g.Switch(jen.Id("name")).BlockFunc(func(sg *jen.Group) {
```
stevendanna, line 449: Perhaps we can special case an empty `acc.secretInfo` map so that we can generate more idiomatic code? Right now this will generate:

```
// GetSecret gets a secret by name. Returns nil if it is not set
func (m *ConfigRequest) GetSecret(name string) *wrappers.StringValue {
	switch name {
	default:
		return nil
	}
}
```

Might be nicer to generate

```
// GetSecret gets a secret by name. Returns nil if it is not set
func (m *ConfigRequest) GetSecret(_ string) *wrappers.StringValue {
	return nil
}
```

tylercloke (approved):
Good progress :)

PR #4486, opened 2020-12-16
esgateway pggateway: Changes to parsed nodes and resolvers to PG gateway HAProxy And Make resolvers configurable

kalroy:
Signed-off-by: Kallol Roy <karoy@progress.com>

Automate when configured with multiple postgres nodes (master + followers) fails to resolve the added/removed follower from the cluster beyond the ones mentioned in the config. 
The aim is to make sure that HAProxy should be able to resolve the nodes in a periodic manner based on the resolv.conf.

This involves changes in:
1. Modifying HaProxy template to include the `resolvers` and `server-template`
2. Parsing the resolver.conf to find the default `resolvers` IP addresses
3. Update the config request to populate the HaProxy Config.

The resolvers can be added and passed by adding the following configurations:

```
[pg_gateway.v1.sys.resolvers]
# Multiple resolvers can be specified by
# adding the resolvers in the list.
nameservers = ["127.0.0.53:52"]
enable_system_nameservers = false
```

In case of ESgateway
```
[esgateway.v1.sys.ngx.main.resolvers]
nameservers = ["127.0.0.53:52"]
enable_system_nameservers = false
```

### :+1: Definition of Done

Setting the configurations should make following changes in the deployed environment:
1. Setting the `esgateway` section should make a change in the `/hab/svc/automate-es-gateway/config/nginx.conf` and the resolvers should be added. Example:
```
resolver 192.168.65.1
```

2. Setting the `pg_gateway` section should make a change in the `/hab/svc/automate-pg-gateway/config/haproxy.conf` and the resolvers should be added. Example:
```
resolvers pgdns
        nameserver dns0 192.168.65.1:52
```

1. Pull the latest code
2. Enter hab studio
3. Rebuild the components
 ```
rebuild components/automate-pg-gateway
rebuild components/automate-es-gateway
```
4. Make the changes in the config.toml as above.
5. `start_all_services`
6. Have a look at `/hab/svc/automate-es-gateway/config/nginx.conf` and `/hab/svc/automate-pg-gateway/config/haproxy.conf`

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### CODE REVIEW

File: api/config/esgateway/config_request.go [hunk 1]
```diff
@@ -48,8 +48,9 @@ func DefaultConfigRequest() *ConfigRequest {
 	// resolvers once we are a bit more confident in the nginx
 	// module we are using to do dynamic resolution.
 	//
-	// c.V1.Sys.Ngx.Main.Resolvers = getSystemResolvers()
 
+	c.V1.Sys.Ngx.Main.Resolvers.EnableSystemNameservers = w.Bool(false)
+	c.V1.Sys.Ngx.Main.Resolvers.Nameservers = nil
```
stevendanna, line 53: We can probably avoid setting this to nil since it will be nil by default when the struct is initialized

api/config/esgateway/config_request.go
```diff
@@ -48,8 +48,9 @@ func DefaultConfigRequest() *ConfigRequest {
 	// resolvers once we are a bit more confident in the nginx
 	// module we are using to do dynamic resolution.
 	//
-	// c.V1.Sys.Ngx.Main.Resolvers = getSystemResolvers()
 
+	c.V1.Sys.Ngx.Main.Resolvers.EnableSystemNameservers = w.Bool(false)
```
stevendanna, line 52: `v.V1.Sys.Ngx.Main.Resolvers` is going to be nil. You'll want to initialize it to an empty struct in `NewConfigRequest`

api/config/esgateway/config_request.go
```diff
@@ -89,6 +90,20 @@ func (c *ConfigRequest) Validate() error {
 // to start the service.
 func (c *ConfigRequest) PrepareSystemConfig(creds *ac.TLSCredentials) (ac.PreparedSystemConfig, error) {
 	c.V1.Sys.Tls = creds
+
+	enableSystemNameServer := c.V1.Sys.Ngx.Main.Resolvers.EnableSystemNameservers.GetValue()
```
stevendanna, line 94: For safety, we try to use the accessor method `GetV1().GetSys()...` for all reads.

api/config/esgateway/config_request.go
kalroy: Removed the code to set Nameservers as nil in the latest changes

api/config/esgateway/config_request.go
```diff
@@ -89,6 +91,20 @@ func (c *ConfigRequest) Validate() error {
 // to start the service.
 func (c *ConfigRequest) PrepareSystemConfig(creds *ac.TLSCredentials) (ac.PreparedSystemConfig, error) {
 	c.V1.Sys.Tls = creds
+
+	enableSystemNameServer := c.GetV1().GetSys().GetNgx().GetMain().GetResolvers().GetEnableSystemNameservers().GetValue()
+	nameServers := c.GetV1().GetSys().GetNgx().GetMain().GetResolvers().GetNameservers()
+	if ((nameServers == nil) || (len(nameServers) == 0)) && enableSystemNameServer {
```
stevendanna, line 97: In Go, the length of a nil slice is 0, so I believe you can simplify this conditional to:

```suggestion
	if len(nameServers) == 0 && enableSystemNameServer {
```

See https://play.golang.org/p/BqOVEFB8YSp

components/automate-es-gateway/habitat/config/nginx.conf
```diff
@@ -24,15 +24,15 @@ http {
   gzip_proxied {{cfg.ngx.http.gzip_proxied}};
   gzip_types {{cfg.ngx.http.gzip_types}};
 
-  {{~#if cfg.ngx.main.resolvers}}
-  resolver {{cfg.ngx.main.resolvers}};
-  {{/if~}}
+  {{#if cfg.ngx.main.resolvers.nameservers}}
```
stevendanna, line 27: Might it be better to only reference the value that is actually required, `nameservers_string`?

kalroy: I was sceptical if `nameservers_string` being a string will also pass for empty string. But as discussed and as per the handlebarjs document, empty string would be inferred as false

api/config/pg_gateway/config_request.proto
```diff
@@ -22,12 +22,25 @@ message ConfigRequest {
 			chef.automate.infra.config.TLSCredentials tls = 3;
 			chef.automate.infra.config.Log log = 4;
 			Timeouts timeouts = 5;
+			Resolvers resolvers = 6;
+
+			message Resolvers {
+				repeated google.protobuf.StringValue name_servers = 1;
```
stevendanna, line 28: I think we should make this the same as it is in esgateway: `nameservers`

api/config/esgateway/config_request.go
```diff
@@ -89,6 +91,20 @@ func (c *ConfigRequest) Validate() error {
 // to start the service.
 func (c *ConfigRequest) PrepareSystemConfig(creds *ac.TLSCredentials) (ac.PreparedSystemConfig, error) {
 	c.V1.Sys.Tls = creds
+
+	enableSystemNameServer := c.GetV1().GetSys().GetNgx().GetMain().GetResolvers().GetEnableSystemNameservers().GetValue()
+	nameServers := c.GetV1().GetSys().GetNgx().GetMain().GetResolvers().GetNameservers()
+	if len(nameServers) == 0 && enableSystemNameServer {
+		c.V1.Sys.Ngx.Main.Resolvers.NameserversString = getSystemResolvers()
+	} else {
+		if (nameServers != nil) && (len(nameServers) > 0) {
```
stevendanna, line 100: Similar to the previous site, this could be 

```suggestion
		if len(nameServers) > 0 {
```

api/config/esgateway/config_request.proto
```diff
@@ -43,7 +43,14 @@ message ConfigRequest {
 				message Main {
 					google.protobuf.Int32Value worker_processes = 1;
 					google.protobuf.Int32Value max_fails = 2;
-					google.protobuf.StringValue resolvers = 3;
+					reserved 3;
```
jaym, line 46: i think we're saying this breaking change is ok because we haven't released this "resolvers" field to the world?

api/config/esgateway/config_request.go
```diff
@@ -89,6 +91,20 @@ func (c *ConfigRequest) Validate() error {
...
+
+	enableSystemNameServer := c.GetV1().GetSys().GetNgx().GetMain().GetResolvers().GetEnableSystemNameservers().GetValue()
+	nameServers := c.GetV1().GetSys().GetNgx().GetMain().GetResolvers().GetNameservers()
+	if len(nameServers) == 0 && enableSystemNameServer {
+		c.V1.Sys.Ngx.Main.Resolvers.NameserversString = getSystemResolvers()
+	} else {
+		if len(nameServers) > 0 {
+			ns := make([]string, 0, len(nameServers))
+			for _, n := range nameServers {
+				ns = append(ns, n.GetValue())
```
jaym, line 103: [nit] maybe check `n != nil`. I think the worst thing that will happen in this case is there will be extra spaces in the namservers string

jaym (approved):
You might be able to modify the config of some of the HA tests to get some integration coverage here

api/config/esgateway/config_request.proto
kalroy: Yes indeed. I added as the build was failing since there is a change in the filed format.

msorens:
Thanks for adding the "how to test..." and "definition of done" sections. That gets the reader much closer to understanding what you are trying to achieve. However, for someone unfamiliar with the context (like me 😁 ), that seems to only show that the entered config lines end up in the correct location. But how does that connect to what your "description" says?

kalroy:
I agree. I should have added more information there. The resolvers would make sure that the PG/ES node IPs can be resolved dynamically and any change in the cluster would not fail the application. In the current scenario when the node IPs are provided and if cluster node goes down, the gateway would keep on trying with the IP addresses. This encourages users to provide hostnames rather than the IPs.

kalroy:
Found an issue while doing acceptance testing for this. The issue can be summarised as below:

While setting up external postgres. In case if the config does not have
```
[global.v1.external.postgresql.ssl]
   enable =false
```

Automate-dex expects the connection to be SSL enabled and expects the root_cert path.
This can be fixed from the code or we can update the doc to mention the behaviour.
But we can take it up in the upcoming release as generally for any external ES/PG setup, the connections would be SSL secured for obvious reasons.

Issue #1880, opened 2019-10-16
investigate the possibility of adding an "xml" format option to compliance report download


vjeffrey:
## User Story
We've heard from the field that customers have been asking about getting an xml format option for compliance report downloads. This is b/c many security analysis tools use xml/xccdf.

Let's investigate if this is possible/how much work would be involved.

## Definition of Done
Summary of work needed to accomplish this or a pr providing the api functionality and a ui issue for adding the xml option if product/ux approves.

vjeffrey:
wip branch: vj/xml

vjeffrey:
https://stackoverflow.com/questions/30928770/marshall-map-to-xml-in-go

vjeffrey:
i think we're gonna have to map everything out similarly to how we've done this for csv conversion

rickmarry:
I believe that this requires us to export the report in the xccdf/1.1 format so that it may be imported and viewed in StigViewer

here's an example:
https://www.stigviewer.com/stig/apache_server_2.4_unix_site/2020-12-10/MAC-3_Sensitive/xml

PR #4493, opened 2020-12-17
[WIP] Extend export report API to support xml format

vsingh-msys:

- As of now we have two supported formats `JSON` & `CSV` to download the compliance report.
- This PR is for supporting the xml format to download the report.

Fixes: #1880 

### :+1: Definition of Done
- Export compliance report in XML format.

- `rebuild components/compliance-service/`
- `chef_load_compliance_nodes`
- hit export API with `type: xml`:

```
curl -sSkH "api-token: $(get_admin_token)" -X POST https://a2-dev.test/api/v0/compliance/reporting/export -d '{"type": "xml"}'
```

Response:

```
<Node>
      <ID>42838336-cec8-4ba5-9da2-e87a52bc082b</ID>
      <NodeID>8470e194-8f68-49ff-a61f-6a47670ff00a</NodeID>
      <NodeName>node1</NodeName>
      <EndTime>
          <Seconds>1608197927</Seconds>
      </EndTime>
      <Platform>
          <Platform>
              <Name>ubuntu</Name>
              <Release>16.04</Release>
              <Full>ubuntu 16.04</Full>
          </Platform>
      </Platform>
      <Environment>_default</Environment>
      <IPAddress>10.0.2.15</IPAddress>
      <FQDN>node1.test</FQDN>
      <Profiles>
          <Profile>
              <Name>ssl-baseline</Name>
              <Title>DevSec SSL/TLS Baseline</Title>
              <Version>1.6.3</Version>
              <Summary>Ensures a secure configuration for TCP ports</Summary>
              <Controles>
                  <Control>
                      <ID>au-anon</ID>
                      <Title>Disable ANON as AU from all exposed SSL/TLS ports and versions.</Title>
                      <Impact>0.5</Impact>
                  </Control>
                  <Waived>no</Waived>
                  <Waiver></Waiver>
                  <Results>
                      <Result>
                          <Status>skipped</Status>
                          <RunTime>3.032e-06</RunTime>
                          <CodeDescription>No-op</CodeDescription>
                          <SkipMessage>Skipped control due to only_if condition.</SkipMessage>
                      </Result>
                  </Results>
              </Controles>
              <Controles>
                  <Control>
                      <ID>au-export</ID>
                      <Title>Disable EXPORT as AU from all exposed SSL/TLS ports and versions.</Title>
                      <Impact>0.5</Impact>
                  </Control>
                  <Waived>no</Waived>
                  <Waiver></Waiver>
                  <Results>
                      <Result>
                          <Status>skipped</Status>
                          <RunTime>2.847e-06</RunTime>
                          <CodeDescription>No-op</CodeDescription>
                          <SkipMessage>Skipped control due to only_if condition.</SkipMessage>
                      </Result>
                  </Results>
              </Controles>
          </Profile>
      </Profiles>
      <Version>4.24.8</Version>
      <ChefServer>chef-server.test</ChefServer>
      <ChefOrganization>4thcoffee</ChefOrganization>


```

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/compliance-service/api/reporting/server/server.go
+++ b/components/compliance-service/api/reporting/server/server.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	sorter "sort"
 	"strings"
+	"time"
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -20,7 +21,6 @@ import (
 	"github.com/chef/automate/lib/grpc/auth_context"
 	"github.com/chef/automate/lib/io/chunks"
 	"github.com/sirupsen/logrus"
-	"time"
 )
 
 // Chosen somewhat arbitrarily to be a "good enough" value.
@@ -304,6 +304,8 @@ func getExportHandler(format string, stream reporting.ReportingService_ExportSer
 		return jsonExport(stream), nil
 	case "csv":
 		return csvExport(stream), nil
+	case "xml":
+		return xmlExport(stream), nil
 	default:
 		return nil, status.Error(codes.Unauthenticated, fmt.Sprintf(format+" export is not supported"))
 	}
@@ -315,6 +317,34 @@ func jsonExport(stream reporting.ReportingService_ExportServer) exportHandler {
 		raw, err := json.Marshal(data)
 		if err != nil {
 			return fmt.Errorf("Failed to marshal JSON export data: %+v", err)
+		}
+
+		if initialRun {
+			initialRun = false
+		} else {
+			raw = append([]byte(","), raw...)
+		}
+		reader := bytes.NewReader(raw)
+		buf := make([]byte, streamBufferSize)
+
+		writer := chunks.NewWriter(streamBufferSize, func(p []byte) error {
+			return stream.Send(&reporting.ExportData{Content: p})
+		})
+		_, err = io.CopyBuffer(writer, reader, buf)
+		if err != nil {
+			return fmt.Errorf("Failed to export JSON: %+v", err)
+		}
+
+		return nil
+	}
+}
+
+func xmlExport(stream reporting.ReportingService_ExportServer) exportHandler {
+	initialRun := true
+	return func(data *reporting.Report) error {
+		raw, err := util.ReportToXML(data)
+		if err != nil {
+			return fmt.Errorf("Failed to marshal XML export data: %+v", err)
 		}
 
 		if initialRun {
@@ -410,8 +440,8 @@ func (srv *Server) ListNodes(ctx context.Context, in *reporting.Query) (*reporti
 func (srv *Server) ReadNode(ctx context.Context, in *reporting.Id) (*reporting.Node, error) {
 	formattedFilters := formatFilters([]*reporting.ListFilter{})
 	// This method takes no filters, setting the widest start->end time range to find this node
-	formattedFilters["start_time"] = []string{ "2017-01-01T00:00:00Z" }
-	formattedFilters["end_time"] = []string{	time.Now().UTC().Format(time.RFC3339) }
+	formattedFilters["start_time"] = []string{"2017-01-01T00:00:00Z"}
+	formattedFilters["end_time"] = []string{time.Now().UTC().Format(time.RFC3339)}
 	formattedFilters, err := filterByProjects(ctx, formattedFilters)
 	if err != nil {
 		return nil, errorutils.FormatErrorMsg(err, in.Id)
--- a/components/compliance-service/reporting/util/xml.go
+++ b/components/compliance-service/reporting/util/xml.go
@@ -1,0 +1,133 @@
+package util
+
+import (
+	"encoding/xml"
+	"fmt"
+
+	reportingapi "github.com/chef/automate/api/interservice/compliance/reporting"
+)
+
+type xmlFields struct {
+	XMLName          xml.Name  `xml:"Node"`
+	ID               string    `xml:"ID,omitempty"`
+	NodeID           string    `xml:"NodeID,omitempty"`
+	NodeName         string    `xml:"NodeName,omitempty"`
+	EndTime          int64     `xml:"EndTime>Seconds,omitempty"`
+	Platform         Platform  `xml:"Platform,omitempty"`
+	Environment      string    `xml:"Environment,omitempty"`
+	IPAddress        string    `xml:"IPAddress,omitempty"`
+	FQDN             string    `xml:"FQDN,omitempty"`
+	Profiles         []Profile `xml:"Profiles,omitempty"`
+	Version          string    `xml:"Version,omitempty"`
+	ChefServer       string    `xml:"ChefServer,omitempty"`
+	ChefOrganization string    `xml:"ChefOrganization,omitempty"`
+}
+
+// Platform detail
+type Platform struct {
+	Name    string `xml:"Platform>Name,omitempty"`
+	Release string `xml:"Platform>Release,omitempty"`
+	Full    string `xml:"Platform>Full,omitempty"`
+}
+
+// Profile detail
+type Profile struct {
+	Name      string    `xml:"Profile>Name,omitempty"`
+	Title     string    `xml:"Profile>Title,omitempty"`
+	Version   string    `xml:"Profile>Version,omitempty"`
+	Summary   string    `xml:"Profile>Summary,omitempty"`
+	Controles []Control `xml:"Profile>Controles,omitempty"`
+}
+
+// Control detail
+type Control struct {
+	ID                  string   `xml:"Control>ID,omitempty"`
+	Title               string   `xml:"Control>Title,omitempty"`
+	Impact              float32  `xml:"Control>Impact,omitempty"`
+	Waived              string   `xml:"Waived,omitempty"`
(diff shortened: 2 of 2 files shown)


components/automate-ui/src/app/pages/+compliance/+reporting/reporting.component.html
```diff
@@ -28,6 +28,7 @@
             <chef-click-outside (clickOutside)="hideDownloadDropdown()" omit="dropdown-toggle">
               <chef-button tertiary (click)="onDownloadOptPressed('json')">Download JSON</chef-button>
               <chef-button tertiary (click)="onDownloadOptPressed('csv')">Download CSV</chef-button>
+              <chef-button tertiary (click)="onDownloadOptPressed('xml')">Download XML</chef-button>
```
vsingh-msys, line 31: Added `Download XML` button in a similar fashion.

components/compliance-service/api/reporting/server/server.go
```diff
@@ -337,6 +339,34 @@ func jsonExport(stream reporting.ReportingService_ExportServer) exportHandler {
...
+		}
+		reader := bytes.NewReader(raw)
+		buf := make([]byte, streamBufferSize)
+
+		writer := chunks.NewWriter(streamBufferSize, func(p []byte) error {
+			return stream.Send(&reporting.ExportData{Content: p})
+		})
+		_, err = io.CopyBuffer(writer, reader, buf)
+		if err != nil {
+			return fmt.Errorf("Failed to export JSON: %+v", err)
```
vsingh-msys, line 363: will change msg in next commit `Failed to export XML:`

components/compliance-service/reporting/util/xml.go
```diff
@@ -0,0 +1,133 @@
...
+	Full    string `xml:"Platform>Full,omitempty"`
+}
+
+// Profile detail
+type Profile struct {
+	Name      string    `xml:"Name,omitempty"`
+	Title     string    `xml:"Title,omitempty"`
+	Version   string    `xml:"Version,omitempty"`
+	Summary   string    `xml:"Summary,omitempty"`
+	Controles []Control `xml:"Controles>Control,omitempty"`
```
msorens, line 39:
```suggestion
	Controls []Control `xml:"Controls>Control,omitempty"`
```

components/compliance-service/reporting/util/xml.go
```diff
@@ -0,0 +1,133 @@
...
+type Result struct {
+	Status          string  `xml:"Status,omitempty"`
+	RunTime         float32 `xml:"RunTime,omitempty"`
+	CodeDescription string  `xml:"CodeDescription,omitempty"`
+	Message         string  `xml:"Message,omitempty"`
+	SkipMessage     string  `xml:"SkipMessage,omitempty"`
+}
+
+// ReportToXML converts a report to its XML representation as a string
+func ReportToXML(report *reportingapi.Report) ([]byte, error) {
```
msorens, line 63: Would like to see some unit tests for this.

msorens (changes requested):
Looking good, but please add unit tests.

components/compliance-service/api/reporting/server/server.go
```diff
@@ -337,6 +339,34 @@ func jsonExport(stream reporting.ReportingService_ExportServer) exportHandler {
 	}
 }
```
rickmarry, line 341: Minor nit but we can actually refactor these export functions into one and pass in the type (json, XML or csv), since they are nearly identical.  The only difference between the three functions is what we do when it's not the initial run, and the conversion to XML or csv if not json.

components/compliance-service/reporting/util/xml.go
rickmarry: nice job on this file!

components/compliance-service/api/reporting/server/server.go
```diff
@@ -337,6 +339,34 @@ func jsonExport(stream reporting.ReportingService_ExportServer) exportHandler {
...
+	return func(data *reporting.Report) error {
+		raw, err := util.ReportToXML(data)
+		if err != nil {
+			return fmt.Errorf("Failed to marshal XML export data: %+v", err)
+		}
+
+		if initialRun {
+			initialRun = false
+		} else {
+			raw = append([]byte(","), raw...)
```
rickmarry, line 353: I don't think this is what we want.. when we are exporting more than one node, this will create a comma separated set of xml, one for each node.  Instead, I think we should wrap the set with <nodes></nodes> and lose the ','

rickmarry (changes requested):

components/compliance-service/api/reporting/server/server.go
vsingh-msys: Yeah, thanks for pointing it out, I am keeping this to `raw = append([]byte("\n"), raw...)`

so the output will be:

```
  <Node>
    <ID>1</ID>
  </Node>
  <Node>
    <ID>2</ID>
  </Node>
```

 Wrapping with `<Nodes></Nodes>` make sense to me

```
       if initialRun {
            raw = append([]byte("<Nodes>\n"), raw...)
			initialRun = false
		} else {
			raw = append([]byte("\n"), raw...)
		}
		reader := bytes.NewReader(raw)


```


But not found a suitable way to add `\n</Nodes>` at the end of the byte string any pointers? Thanks!

rickmarry: >>But not found a suitable way to add \n</Nodes> at the end of the byte string any pointers? Thanks!
Good question.. let me think about that one a bit

msorens: Treating XML as if it was just strings like this is quite brittle. Would much prefer that you gather the collection of individual `<Node>` items together, then output `<Nodes>` + collection-of-node-items + `</Nodes>` at one time.

rickmarry: > Treating XML as if it was just strings like this is quite brittle. Would much prefer that you gather the collection of individual `<Node>` items together, then output `<Nodes>` + collection-of-node-items + `</Nodes>` at one time. That is, always emit valid XML elements. (Assuming it is possible in this case.)

not really possible here @msorens... we're streaming it

kalroy: This code should return a definitive number of IDs which we are looping through:
```
reportIDs, err := srv.es.GetReportIds(esIndex, formattedFilters)
```

Probably here:

```
total := len(reportIDs.Ids)
// Step 2: get all reports one by one in reverse order to be sorted asc by end_time
for idx := total - 1; idx >= 0; idx-- {
```

We may utilize this idx to identify the first and the last nodes and indicate the exporter.

rickmarry: I see how to fix this.. working on it.. will be up very soon

components/automate-gateway/gateway/services.go
```diff
@@ -672,18 +672,25 @@ func (s *Server) ReportExportHandler(w http.ResponseWriter, r *http.Request) {
...
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
+	} else if queryType == "xml" {
+		_, err := w.Write([]byte(fmt.Sprintf("<%s>\n", wrapper)))
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
```
msorens, line 693: Nice, but could (a) be a bit more concise and (b) show the intent more clearly:
```suggestion
	openBracket := nil
	closeBracket := nil
	switch queryType {
	case "json":
		openBracket = "["
		closeBracket = "]"
	case "xml":
		openBracket = fmt.Sprintf("<%s>\n", wrapper)
		closeBracket = fmt.Sprintf("\n</%s>", wrapper)
	}
	if openBracket != nil {
		_, err := w.Write([]byte(openBracket))
		if err != nil {
			http.Error(w, err.Error(), http.StatusInternalServerError)
			return
		}
	}
	for {
		data, err := stream.Recv()
		if err == io.EOF {
			if closeBracket != nil {
				_, err = w.Write([]byte(closeBracket))
				if err != nil {
					http.Error(w, err.Error(), http.StatusInternalServerError)
					return
				}
			}
```

rickmarry:
we have to make this xml/xccdf compliant

components/automate-gateway/gateway/services.go
rickmarry: I l

components/automate-gateway/gateway/services.go
```diff
@@ -672,14 +672,25 @@ func (s *Server) ReportExportHandler(w http.ResponseWriter, r *http.Request) {
...
+	var closeBracket string
+	switch queryType {
+	case "json":
+		openBracket = "["
+		closeBracket = "]"
+	case "xml":
+		openBracket = fmt.Sprintf("<%s>\n", wrapper)
+		closeBracket = fmt.Sprintf("\n</%s>", wrapper)
+	}
+	if len(openBracket) == 0 {
```
msorens, line 692: You seem to have the polarity of these guards backwards--should be `!=` here and on L702.
(Curious though, any reason `if openBracket != nil` would not work...? It is one less function call.)

PR #4495, opened 2020-12-17
Encrypt secrets that are put in the secrets store

vivekshankar1:
This modifies the secrets store/secrets helper utilities to encrypt secrets that are stored on disk. Any existing secret data will not be encrypted unless it is modified. This is fine, as all existing data is generated and uses the secret helpers to share that data between services.

The reason we are doing this is be able to store customer provided secrets in non plaintext. We will take the secrets that are provided in the config and place them in the secret store encrypted.

This is not a user facing change.
This does not provide any sort of security, as the key is stored on next to the secrets on disk.

### DIFF
--- a/lib/secrets/secrets.go
+++ b/lib/secrets/secrets.go
@@ -7,6 +7,8 @@ package secrets
 
 import (
 	"bytes"
+	"crypto/aes"
+	"crypto/cipher"
 	"crypto/rand"
 	"encoding/hex"
 	"io/ioutil"
@@ -19,6 +21,7 @@ import (
 
 	"github.com/chef/automate/lib/io/fileutils"
 	"github.com/chef/automate/lib/user"
+	"github.com/chef/toml"
 )
 
 // Commonly used secrets names. When accessing this library from
@@ -34,6 +37,14 @@ var BifrostSuperuserIDName = SecretName{
 type SecretName struct {
 	Group string
 	Name  string
+}
+
+// SecretKeyToml secret toml structure thats required to write and read
+// to the encrypted secret file
+type SecretKeyToml struct {
+	Algorithm  string `toml:"algorithm"`
+	IV         string `toml:"iv"`
+	Ciphertext string `toml:"ciphertext"`
 }
 
 // SecretNameFromString returns parses the string according to the
@@ -97,6 +108,9 @@ const (
 	// secrets store, read the secrets as root, and then drop privs when
 	// we exec the command.
 	DefaultDiskStoreDataOwner = "hab"
+
+	// SecretFileExtension is the extension name for the encrypted secrets file
+	SecretFileExtension = ".enc.toml"
 )
 
 func NewDiskStoreReader(basePath string) SecretsReader {
@@ -147,27 +161,68 @@ func (d *diskStore) Initialize() error {
 		return errors.Wrap(err, "could now chown shared data directory for secrets store")
 	}
 
+	// creates encryption key if it does not exist
+	keyExists, err := keyExists(d.basePath)
+	if err != nil {
+		return err
+	}
+	if !keyExists {
+		_, err := d.createEncryptionKey()
+		if err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 
 func (d *diskStore) Exists(secret SecretName) (bool, error) {
+	// added logic to check if secret is encrypted or not
+	encryptedPath := filepath.Join(d.basePath, secret.Group, addSecretFileExtension(secret.Name))
+	encryptedExists, err := fileutils.PathExists(encryptedPath)
+	if err != nil {
+		return false, err
+	}
+	if encryptedExists {
+		return encryptedExists, nil
+	}
 	path := filepath.Join(d.basePath, secret.Group, secret.Name)
 	return fileutils.PathExists(path)
 }
 
 func (d *diskStore) GetSecret(secret SecretName) ([]byte, error) {
-	path := filepath.Join(d.basePath, secret.Group, secret.Name)
-	ret, err := ioutil.ReadFile(path)
+
+	// checks if the encrypted secret exists
+	pathToEncryptedSecret := filepath.Join(d.basePath, secret.Group, addSecretFileExtension(secret.Name))
+	encryptedSecretExists, err := fileutils.PathExists(pathToEncryptedSecret)
 	if err != nil {
-		return nil, errors.Wrap(err, "could not read secrets data (secret may not have been generated yet)")
+		return nil, errors.Wrap(err, "unexpected error while checking if encrypted file exists (secret may not have been generated yet)")
+	}
+	if encryptedSecretExists {
+		value, err := ioutil.ReadFile(pathToEncryptedSecret)
+		if err != nil {
+			return nil, errors.Wrap(err, "unexpected error while reading encrypted secret data")
+		}
+		// decrypts and returns the secret value
+		ret, err := getDecryptedData(d.basePath, value)
+		if err != nil {
+			return nil, errors.Wrap(err, "could not decrypt secret with existing key")
+		}
+		return ret, err
 	}
 
+	// else return unencrypted secret
+	pathToUnencryptedSecret := filepath.Join(d.basePath, secret.Group, secret.Name)
+	ret, err := ioutil.ReadFile(pathToUnencryptedSecret)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not read unencrypted secrets data (secret may not have been generated yet)")
+	}
 	return ret, err
 }
 
 func (d *diskStore) SetSecret(secret SecretName, data []byte) error {
 	parentDir := filepath.Join(d.basePath, secret.Group)
-	secretPath := filepath.Join(parentDir, secret.Name)
+	secretPath := filepath.Join(parentDir, addSecretFileExtension(secret.Name))
 	err := os.MkdirAll(parentDir, 0700)
 	if err != nil {
 		return errors.Wrap(err, "could not create directory for secrets data")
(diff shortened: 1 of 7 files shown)


lib/secrets/secrets.go
```diff
@@ -229,3 +279,155 @@ func PrepareSSHPrivateKey(keyContent string) (string, error) {
...
+// return encrypted cipher, iv and error
+func getEncryptedData(data []byte, basePath string) ([]byte, []byte, error) {
+	// we create the encryption key if it doesnot exist
+	var secretKey []byte
+	keyExists, err := keyExists(basePath)
+	if err != nil {
+		return nil, nil, err
+	}
+	if !keyExists {
+		secretKey, err = createEncryptionKey(basePath)
```
jaym, line 364: I think if the encryption key doesn't exist at this point, you shouldn't create it. Deployment service should have already created it. If the key doesn't exist, we risk having a race condition here, as multiple services could end up executing this logic concurrently

lib/secrets/secrets.go
```diff
@@ -229,3 +279,155 @@ func PrepareSSHPrivateKey(keyContent string) (string, error) {
...
 	return keyPath.Name(), nil
 }
+
+// checks if encryption key exists or not
+func keyExists(basePath string) (bool, error) {
+	path := filepath.Join(basePath, "key")
+	return fileutils.PathExists(path)
+}
+
+func encodeBase64(b []byte) string {
```
jaym, line 289: given that we already use hex encoding in another part of this file, it might make more sense to continue using that encoding

lib/secrets/secrets.go
```diff
@@ -229,3 +279,155 @@ func PrepareSSHPrivateKey(keyContent string) (string, error) {
...
+		IV:         iv,
+		Ciphertext: ciphertext,
+	}
+
+	buf := new(bytes.Buffer)
+	if err := toml.NewEncoder(buf).Encode(secretData); err != nil {
+		log.Fatal(err)
+	}
+
+	err := fileutils.AtomicWrite(path, bytes.NewReader(buf.Bytes()), fileutils.WithAtomicWriteFileMode(0700))
```
jaym, line 347: This file doesn't need execute permissions. Also, i think you need to set the owner of the file

lib/secrets/secrets.go
```diff
@@ -229,3 +279,155 @@ func PrepareSSHPrivateKey(keyContent string) (string, error) {
...
+	secretKey, err := GenerateRandomBytes(32)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not generate random secret for encryption")
+	}
+
+	// creating path to store the secret key
+	secretKeyPath := filepath.Join(basePath, "key")
+	s := bytes.NewReader(secretKey)
+	// writing the key to a file
+	err = fileutils.AtomicWrite(secretKeyPath, s, fileutils.WithAtomicWriteFileMode(0700))
```
jaym, line 428: I think you need to set the owner of the file

lib/secrets/secrets.go
vivekshankar1: Removed this logic and returning error now

lib/secrets/secrets.go
vivekshankar1: Changed to hex now

lib/secrets/secrets.go
vivekshankar1: setting owner now

lib/secrets/secrets.go
vivekshankar1: setting owner now

lib/secrets/secrets.go
```diff
@@ -156,18 +179,37 @@ func (d *diskStore) Exists(secret SecretName) (bool, error) {
 }
 
 func (d *diskStore) GetSecret(secret SecretName) ([]byte, error) {
-	path := filepath.Join(d.basePath, secret.Group, secret.Name)
-	ret, err := ioutil.ReadFile(path)
+	var ret []byte
```
jaym, line 182: I don't think ret needs to be declared here. It gets unconditionally set outside the following if block

lib/secrets/secrets.go
```diff
@@ -156,18 +179,37 @@ func (d *diskStore) Exists(secret SecretName) (bool, error) {
...
+
+	// checks if the encrypted secret exists
+	pathToEncryptedSecret := filepath.Join(d.basePath, secret.Group, secret.Name+".enc.toml")
+	encryptedSecretExists, err := fileutils.PathExists(pathToEncryptedSecret)
```
jaym, line 186: Do we need to do anything with err?

lib/secrets/secrets.go
```diff
@@ -156,18 +179,37 @@ func (d *diskStore) Exists(secret SecretName) (bool, error) {
...
+
+	if encryptedSecretExists {
+		ret, err = ioutil.ReadFile(pathToEncryptedSecret)
+		if err != nil {
+			return nil, errors.Wrap(err, "could not read secrets data (secret may not have been generated yet)")
+		}
+		// decrypts and returns the secret value
+		retE, err := getDecryptedData(d.basePath, ret)
+		if err != nil {
+			return nil, errors.Wrap(err, "could not decrypt secret with existing key)")
```
jaym, line 196: There seems to be an unmatched parenthesis in the string

lib/secrets/secrets.go
```diff
@@ -156,18 +179,37 @@ func (d *diskStore) Exists(secret SecretName) (bool, error) {
...
+	var ret []byte
+
+	// checks if the encrypted secret exists
+	pathToEncryptedSecret := filepath.Join(d.basePath, secret.Group, secret.Name+".enc.toml")
+	encryptedSecretExists, err := fileutils.PathExists(pathToEncryptedSecret)
+
+	if encryptedSecretExists {
+		ret, err = ioutil.ReadFile(pathToEncryptedSecret)
+		if err != nil {
+			return nil, errors.Wrap(err, "could not read secrets data (secret may not have been generated yet)")
```
jaym, line 191: This error could be misleading. The file is being checked for existence before coming into the block

lib/secrets/secrets.go
```diff
@@ -229,3 +299,127 @@ func PrepareSSHPrivateKey(keyContent string) (string, error) {
...
+	encryptedData, iv, err := encrypt(secretKey, data)
+	if err != nil {
+		return nil, nil, err
+	}
+	return encryptedData, iv, nil
+}
+
+// gets the secret from the toml file and decrypts it
+// returns the decrypted data or error
+func getDecryptedData(basePath string, ret []byte) ([]byte, error) {
```
jaym, line 395: I think the name ret doesn't make sense here as its not the returned value

lib/secrets/secrets_test.go
```diff
@@ -95,3 +98,53 @@ func TestDiskSecretStore(t *testing.T) {
...
+	uid := os.Getuid()
+	gid := os.Getgid()
+	dataDir := filepath.Join(tmpDir, "data-dir")
+
+	dataStore := secrets.NewDiskStore(dataDir, uid, gid)
+	err := dataStore.Initialize()
+	require.NoError(t, err, "failed to initialize")
+
+	properties.Property("generate multiple random test cases to check if get and set can roundtrip a secret", prop.ForAll(
+		func(a string, b string, c string) bool {
```
jaym, line 119: maybe more descriptive parameter names

lib/secrets/secrets_test.go
```diff
@@ -95,3 +98,53 @@ func TestDiskSecretStore(t *testing.T) {
...
+			}
+			secretContent, err := dataStore.GetSecret(testSecretName)
+			if err != nil {
+				return reportErrorAndYieldFalse(t, err)
+			}
+			return string(testSecretContent) == string(secretContent)
+		},
+		stringGen(10).WithLabel("a"),
+		stringGen(10).WithLabel("b"),
+		stringGen(10).WithLabel("c"),
```
jaym, line 134: I think we want to generate a variety of secrets. We don't want to limit it just 10 alphanumerical characters

lib/secrets/secrets_test.go
```diff
@@ -95,3 +98,53 @@ func TestDiskSecretStore(t *testing.T) {
...
+			testSecretContent := []byte(c)
+			err := dataStore.SetSecret(testSecretName, testSecretContent)
+			if err != nil {
+				return reportErrorAndYieldFalse(t, err)
+			}
+			secretContent, err := dataStore.GetSecret(testSecretName)
+			if err != nil {
+				return reportErrorAndYieldFalse(t, err)
+			}
+			return string(testSecretContent) == string(secretContent)
```
jaym, line 130: the secret content need not be a string. I think any sequence of bytes is allowed

lib/secrets/secrets.go
vivekshankar1: removed now

lib/secrets/secrets.go
vivekshankar1: missed it. I've added now

lib/secrets/secrets.go
vivekshankar1: fixed

lib/secrets/secrets.go
vivekshankar1: changed it to be more relatable

lib/secrets/secrets.go
vivekshankar1: changed it to data

lib/secrets/secrets_test.go
vivekshankar1: changed to group, name and maxContentLength. MaxContentLength is for generating content upto 5kb

lib/secrets/secrets_test.go
vivekshankar1: content will be randomly generated from sizes 1 byte to 5 kb

lib/secrets/secrets_test.go
vivekshankar1: Comparing bytes now. bytes.Compare(testSecretContent, secretContent)

jaym (approved):

PR #4571, opened 2021-01-13
Move secrets from config to secret store

jaym:
Move secrets stored in the config to the secret store.

Added a watch mode the secrets-helper exec. When run with this flag, the
we run the requested executable as a child process. If any of the
watched secrets change, the child process is killed. This will make sure changes
to secrets still get picked up automatically. I've kept it simple for now. It polls the secrets
instead of using anything like inotify.

To run this:
- rebuild components/automate-platform-tools
- export local_platform_tools_origin=YOURORIGIN
- rebuild components/automate-deployment
- rebuild components/automate-dex

You can use the following config to patch automate:
```
[dex.v1.sys.connectors.ldap]
host = "ldap.localhost"
bind_dn = "uid=service_account,dc=example,dc=local"
bind_password = "testpassword3"
insecure_no_ssl = true
username_attr = "cn"
user_id_attr = "uid"
base_user_search_dn = "cn=users,dc=example,dc=local"
```

That's not a working ldap configuration, but it will be accepted (it may take 15 seconds). You can check that dex starts with `LDAP_PASSWORD=testpassword3` by looking at `/proc/PID/environ`

Better test would be to setup ldap for real

### DIFF
--- a/components/automate-platform-tools/cmd/secrets-helper/secrets-helper.go
+++ b/components/automate-platform-tools/cmd/secrets-helper/secrets-helper.go
@@ -21,6 +21,8 @@
 package main
 
 import (
+	"bytes"
+	"context"
 	"encoding/json"
 	"flag"
 	"fmt"
@@ -31,6 +33,7 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -39,6 +42,8 @@ import (
 	"github.com/chef/automate/lib/secrets"
 	"github.com/chef/automate/lib/user"
 )
+
+var defaultWatchInterval = 15 * time.Second
 
 //
 // Option structs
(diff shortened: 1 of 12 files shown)


components/automate-platform-tools/cmd/secrets-helper/secrets-helper.go
```diff
@@ -396,22 +428,73 @@ func execCommand(_ *cobra.Command, args []string) {
...
+		defer cancel()
+
+		cmd := exec.CommandContext(cmdCtx, fullExe, args[1:]...)
+		cmd.ExtraFiles = []*os.File{readPipe}
+		cmd.Env = append(os.Environ(), fmt.Sprintf("CHEF_SECRETS_FD=%d", 3)) // 3 + i from ExtraFiles
+		cmd.Stderr = os.Stderr
+		cmd.Stdout = os.Stdout
+		cmd.Stdin = os.Stdin
+		cmd.SysProcAttr = &syscall.SysProcAttr{
+			Pdeathsig: syscall.SIGTERM,
```
msorens, line 446: @jaym Just noticed that VSCode flagged this as a non-existent field name.

PR #4594, opened 2021-01-15
Minor cleanup (back-end)

msorens:

(1) Semgrep findings suggest as a best practice that when one knows one is returning nil, just return nil rather than obfuscate it inside a variable.
(2) Updating semgrep-ignored lines from `// nosem` to the less obfuscatory `// nosemgrep`; this will aid folks not intimately familiar with semgrep who attempt to do a web search to figure out what it is.

NA

### :+1: Definition of Done
Clean build.

NA

- [ ] Tests added/updated?
- [ ] Docs added/updated?

NA

### DIFF
--- a/components/automate-deployment/pkg/airgap/bundle_creator.go
+++ b/components/automate-deployment/pkg/airgap/bundle_creator.go
@@ -620,7 +620,7 @@ func (creator *InstallBundleCreator) createTar(InstallBundleCreatorProgress) err
 		return errors.Wrap(err, "Could not write archive")
 	}
 
-	return err
+	return nil
 }
 
 func uniq(xs []habpkg.VersionedPackage) []habpkg.VersionedPackage {
--- a/components/automate-deployment/pkg/airgap/bundle_creator.go
+++ b/components/automate-deployment/pkg/airgap/bundle_creator.go
@@ -101,13 +101,13 @@ func (dl *netHabDownloader) DownloadHabBinary(version string, release string, w 
 	}
 
 	// NOTE: We're downloading this via HTTPS from a location that's trusted.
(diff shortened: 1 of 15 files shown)


components/automate-deployment/pkg/airgap/bundle_creator.go
```diff
@@ -620,7 +620,7 @@ func (creator *InstallBundleCreator) createTar(InstallBundleCreatorProgress) err
 		return errors.Wrap(err, "Could not write archive")
 	}
 
-	return err
+	return nil
```
msorens, line 623: If you expand the context right above this, L619-621 has just caused a return if err was not nil, so here we know err is nil, and best practice suggested is to be more up front about it.

components/config-mgmt-service/grpcserver/node_export.go
```diff
@@ -241,7 +241,7 @@ func csvExport(stream service.CfgMgmtService_NodeExportServer) exportHandler {
 			return fmt.Errorf("Failed to export CSV: %+v", err)
 		}
 
-		return err
+		return nil
```
msorens, line 244: Same comment as in bundle_creator.go; we know it is nil.

components/data-feed-service/dao/destination.go
```diff
@@ -84,10 +84,6 @@ func (db *DB) DeleteDestination(delete *datafeed.DeleteDestinationRequest) error
 		}
 		return nil
 	})
-
-	if err != nil {
-		return err
-	}
 	return err
```
msorens, line 87: Deleted lines completely redundant.

msorens (commented):
Author notes

PR #4599, opened 2021-01-19
Fix docs symlink

jaym:
docs symlink was pointing at a folder that didn't exist. Looks like things have been moved to components/docs-chef-io


Shadae (approved):
👍

PR #4603, opened 2021-01-20
Enable HA Chef Server test

jaym:
the chef server tests are now passing the port as part of the host header. When erchef is configured to use the host header, it doesn't strip out the port if its passed via the host header even if its a default port. This change tells pedant to expect the port in the response URIs.


msorens (approved):
🎉 Thank you!

PR #4636, opened 2021-01-22
Revise the release notes template

kagarmoe:

Updates the Release Notes template for a better experience for developers and customers. I've added:

- **Compliance Profiles**: This is the space to note new or changed compliance profiles. Auditors are rarely sysadmins, so it makes sense to provide a section for compliance content.
- **Maintenance**: This is the space for upgrades that admins/maintainers should know about. Did you change the Go version? Add the note here.
- **Package manifest**: I added a link to the package manifest for `current/automate/latest`. It does NOT link to a versioned manifest. I'll be delighted to take your PR that adds a script to make links to versioned package manifests in the release notes happen!

- [ ] Tests added/updated?
- [ ] Docs added/updated?


.expeditor/publish-release-notes.sh
```diff
@@ -24,9 +24,15 @@ pushd ./automate.wiki
 ## Improvements
 -
 
+## Compliance Profiles
```
msorens, line 27: Should this perhaps be "Compliance Profile Updates" or is that obvious to those who would consume this info? (I am not amongst that group so I truly do not know. 😁 )

api/external/infra_proxy/infra_proxy.proto
```diff
@@ -215,6 +218,9 @@ service InfraProxy {
     option (google.api.http).get = "/api/v0/infra/servers/{server_id}/orgs/{org_id}/environments/{name}/recipes";
     option (chef.automate.api.iam.policy).resource = "infra:infraServers:{server_id}:orgs:{org_id}:environments";
     option (chef.automate.api.iam.policy).action = "infra:infraServers:get";
+    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
+			tags: "hidden";
```
msorens, line 222: Curious what we are suppressing here...?

msorens (approved):
On the versioned manifest question, I like the idea but I cannot recall ever seeing a reference except to latest. I agree others should be accessible but I have not seen the format. (Tried the obvious _sha_.json but that did not work.)

.expeditor/publish-release-notes.sh
```diff
@@ -38,6 +44,8 @@ This release uses:
 - Chef Infra Server version
 - Chef InSpec version
```
kalroy, line 46: Wondering if we can add a section to list down the version of the following components Automate uses:
- Postgres
- ElasticSearch
- Nginx
- Haproxy

This can give the users a view of the underlying softwares we are using and also if these are the ones customers need to know as they directly/indirectly touches while configuring external data stores.

Although, having said that, are we exposing these versions anywhere in any other docs?

kalroy (changes requested):

api/external/infra_proxy/infra_proxy.proto
kagarmoe: Picked up files from un-stashing.

.expeditor/publish-release-notes.sh
kagarmoe: Are those in the manifest?

Issue #4623, opened 2021-01-22
In Add Chef Organization Update the Scroll bar CSS as per the automate UI


himanshi-chhabra:
## Describe the bug
We need to show the Scroll bar as per the existing scroll bar we are using in the automate UI.

## To Reproduce
To update the organization in Infrastructure -> Chef Servers -> select server -> click on `Add Chef Organization`

## Expected behavior
Change the CSS of the scroll bar as per automate UI.

## Screenshots
![org-modal-scrollbar](https://user-images.githubusercontent.com/61003053/105479359-5ee2c080-5cca-11eb-98cd-dab5d4a58989.png)

PR #4637, opened 2021-01-23
[infra-proxy-read-only] fixed the scroll bar for the infra create org modal

vinay033:

Create org modal scroll bar is the same as other modal scroll bars.

https://github.com/chef/automate/issues/4623
### :+1: Definition of Done
I have added CSS changes to fix the create org modal scroll bar.
To load sample data:
- Load data into the infra views using `infra_service_load_sample_data` in hab studio (ref: [adding_test_data](https://github.com/chef/automate/blob/master/dev-docs/adding-data/adding_test_data.md))
- Enable the Chef Infra Server Views in the magic menu by typing **feat** anywhere in the automate ui

if you have sample data for the infra servers than 
go to >>  infrastructure tab >> click on Chef server on left side navigation bar >> click to any server where you can see the button `add chef organization` 
**before fix** 
org-modal-scrollbar
**after fix**
org-modal-scrollbar-fixed

### DIFF
--- a/components/automate-ui/src/app/modules/infra-proxy/create-org-modal/create-org-modal.component.html
+++ b/components/automate-ui/src/app/modules/infra-proxy/create-org-modal/create-org-modal.component.html
@@ -2,81 +2,79 @@
   <h2 slot="title">Add Chef Organization</h2>
   <div class="flex-container">
     <form [formGroup]="createForm">
-      <div class="scrollable-form-div">
-        <div class="input-margin">
-          <chef-form-field>
-            <label>
-              <span class="label">Name <span aria-hidden="true">*</span></span>
-              <input chefInput name="name" formControlName="name" type="text" (keyup)="handleNameInput($event)" data-cy="org-name" autocomplete="off">
-            </label>
-            <chef-error
-              *ngIf="(createForm.get('name').hasError('required') || createForm.get('name').hasError('pattern')) && createForm.get('name').dirty">
-              Display Name is required.
-            </chef-error>
-          </chef-form-field>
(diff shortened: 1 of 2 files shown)


vivekshankar1 (approved):
Thank you. Its working as expected

rickmarry:
I like this change.  In fact, I believe that is works better than some of the other model boxes that we have, specifically when zooming the browser window, in this one that you've changed, the zoom maintains the height of the model window and places scroll bar **(if and only if needed)**.
Now, if you want an example of a modal that misbehaves as I've described, take a look at this one: go to >> infrastructure tab >> click on Chef server on left side navigation bar >> click on add chef-server  Notice how when we zoom in, the model does not maintain it's height and no scroll bar is present when needed.. in fact, it's impossible to scroll the page at all.  I believe that the other models should behave the way you've made this one behave. 

nice work!  thank you!  
The one critique I would point out is that you could have made the modal screen a little bit taller so that it reduces the need for the scroll bar as much as possible. there is some spare space above and below that can be used for this.

rickmarry (changes requested):
could make the modal a bit taller to avoid scroll in more resolutions

vinay033:
> could make the modal a bit taller to avoid scroll in more resolutions

hello @rickmarry 
I have incorporated the review comment changes and increased the height of the modal for the add org. please check in the screenshot.
Before the change
org-previous

After the change
org-after

rickmarry:
> I have incorporated the review comment changes and increased the height of the modal for the add org. please check in the screenshot.

perfect.. thank you

rickmarry (approved):
thank you @vinay033

PR #4650, opened 2021-01-25
Fix segfault during migration of secrets

jaym:
Users who have secrets configured prior to last release (20210116...)
are stuck in a crash loop for deployment-service. The issue is that the
secrets store is being initialized too late in the process. This change
initializes it sooner. If a nil secrets store is provided, the code now
ignores it and moves on.

The release plan:
Deployment service will not be able to fix itself in this case. Customers who have already upgraded will need to follow the following steps after we promote the fix:
```
# Install the latest deployment service
hab pkg install chef/deployment-service/0.1.0/$NEW_VERSION
# Tell habitat that it must load the new version
sed -i -e 's|chef/deployment-service/0.1.0/20210116174651|chef/deployment-service/0.1.0/$NEW_VERSION|g' /hab/sup/default/specs/deployment-service.spec
# Tell automate its time to upgrade
chef-automate upgrade run
```

### DIFF
--- a/components/automate-deployment/pkg/deployment/deployment.go
+++ b/components/automate-deployment/pkg/deployment/deployment.go
@@ -2,6 +2,7 @@ package deployment
 
 import (
 	"fmt"
+	"reflect"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


msorens:
Not familiar with the secrets store handling, but your description sounds reasonable. Trying to understand the nuances a bit. You state "If a nil secrets store is provided, the code now ignores it and moves on."  So that would be the case for someone who does not have secrets configured at all, am I correct? Whereas for someone who does have secrets configured, they will now hit this earlier to fix the crash. Ultimately, I am wondering about your "too late in the process" comment. So after "the code now ignores it and moves on" it will never need to circle back and try again, right?

jaym:
Deployment service is responsible for initializing the secret store during its startup, so everyone will have a secret store. The nil check is now purely defensive so deployment-service cant crash because of it. It periodically converges its state to the right thing, so i believe it will eventually circle back. If not, worst case is the user has to reconfigure their config that has secrets.

msorens (approved):
Not in a position to test, but the code read looks good to me.

PR #4674, opened 2021-02-01
Add Integration Testing for the Notifications Service

danielsdeleo:

Create a full integration test suite for the notifications service in go. Everything is tested via external interface. This enables reimplementing the notifications-service while maintaining API stability.

NOTE: click the "Load Large Diff" button for `components/notifications-service/integration_test/notification_test.go`, that's most of the action in this PR.

No work is done to run this test suite in Ci yet. It will make sense to do so once we've created the reimplementation of the notifications-service.

This suite will probably need some adaption to work against the new implementation of the notifications-service. That will be done when there is a new implementation to adapt to.

In a fresh studio (if not fresh, source the `.studio` files):
* start all services
* run `notifications_integration` or `A2_SVC_NAME="notifications-service" A2_SVC_PATH="/hab/svc/notifications-service" go test -v "github.com/chef/automate/components/notifications-service/integration_test" -count 1`

### DIFF
--- a/components/notifications-service/integration_test/notification_test.go
+++ b/components/notifications-service/integration_test/notification_test.go
@@ -414,24 +414,9 @@ EXAMPLE EXCEPTION MESSAGE
 		assert.Equal(t, "EXAMPLE EXCEPTION MESSAGE\n(cookbook_example::recipe_example line 13)", actual.ExceptionMessage)
 		expectedBT := []string{"/path/to/recipe_example.rb: 42", "/path/to/recipe_example.rb: 23"}
 		assert.Equal(t, expectedBT, actual.ExceptionBacktrace)
-		// NOTE: the string values of the time will be different because of how the
-		// floating point value of the fractional seconds is converted to a string
-		// in different programming languages. E.g., trying to compare the strings
-		// can result in something like this:
-		// 		expected: "2021-01-29T19:11:09Z"
-		// 		actual  : "2021-01-29T19:11:09.000000Z"
-		// We can verify the time objects are the same, though.
-		actualStartTime, err := time.Parse(time.RFC3339Nano, actual.StartTimeUTC)
-		assert.NoError(t, err)
-		assert.Equal(t, startTime, actualStartTime)
-
-		actualEndTime, err := time.Parse(time.RFC3339Nano, actual.EndTimeUTC)
-		assert.NoError(t, err)
-		assert.Equal(t, endTime, actualEndTime)
-
-		actualTimestamp, err := time.Parse(time.RFC3339Nano, actual.EndTimeUTC)
-		assert.NoError(t, err)
-		assert.Equal(t, endTime, actualTimestamp)
+		assertRFC3339Equal(t, actual.StartTimeUTC, startTime)
+		assertRFC3339Equal(t, actual.EndTimeUTC, endTime)
+		assertRFC3339Equal(t, actual.TimestampUTC, endTime)
 	})
 	t.Run("ComplianceFailure alerts are sent to the ServiceNow URL", func(t *testing.T) {
 		_, err := suite.Client.Notify(ctx, complianceFailureEvent())
@@ -462,20 +447,9 @@ EXAMPLE EXCEPTION MESSAGE
 		assert.Equal(t, "https://chefautomate.example/compliance_failure/0", actual.AutomateFailureURL)
 		expectedFailureSnippet := "InSpec found a control failure on [chefnode.example](https://chefautomate.example/compliance_failure/0)"
 		assert.Equal(t, expectedFailureSnippet, actual.FailureSnippet)
-		// NOTE: the string values of the time will be different because of how the
-		// floating point value of the fractional seconds is converted to a string
-		// in different programming languages. E.g., trying to compare the strings
-		// can result in something like this:
-		// 		expected: "2021-01-29T19:11:09Z"
-		// 		actual  : "2021-01-29T19:11:09.000000Z"
-		// We can verify the time objects are the same, though.
-		actualEndTime, err := time.Parse(time.RFC3339Nano, actual.EndTimeUTC)
-		assert.NoError(t, err)
-		assert.Equal(t, endTime, actualEndTime)
 
-		actualTimestamp, err := time.Parse(time.RFC3339Nano, actual.TimestampUTC)
-		assert.NoError(t, err)
-		assert.Equal(t, startTime, actualTimestamp)
+		assertRFC3339Equal(t, actual.EndTimeUTC, endTime)
+		assertRFC3339Equal(t, actual.TimestampUTC, startTime)
 
 		assert.Equal(t, 1, actual.TotalNumberOfTests)
 		assert.Equal(t, 0, actual.TotalNumberOfSkippedTests)
@@ -540,10 +514,7 @@ EXAMPLE EXCEPTION MESSAGE
 		assert.Equal(t, "not good", resultActual.CodeDesc)
 		assert.Equal(t, 3.5, resultActual.RunTime)
 
-		resultStartTime, err := time.Parse(time.RFC3339Nano, resultActual.StartTime)
-		assert.NoError(t, err)
-		assert.Equal(t, startTime, resultStartTime)
-
+		assertRFC3339Equal(t, resultActual.StartTime, startTime)
 	})
 }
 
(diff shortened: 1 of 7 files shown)


components/notifications-service/integration_test/rules_crud_test.go
```diff
@@ -0,0 +1,409 @@
...
+	err := suite.DeleteEverything()
+	require.NoError(t, err)
+
+	ruleToCreate := validRule("TestUpdateRule_validations")
+	created, err := suite.Client.AddRule(ctx, ruleToCreate)
+	require.NoError(t, err)
+	createdRule := created.Rule
+	createdRule.Id = created.Id
+
+	t.Run("the Id field must not be empty", func(t *testing.T) {
```
msorens, line 303: Add test case for non-existent ID.

components/notifications-service/integration_test/suite_test.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// TestMain allow us to run a setup before running our tests and also
+// teardown everything after we have finished testing.
+//
+// (Check out 'suite_test.go')
+//
+// => Docs: https://golang.org/pkg/testing/#hdr-Main
+func TestMain(m *testing.M) {
+	// Global Setup hook: Here is where you can initialize anythings you need
```
msorens, line 84:
```suggestion
	// Global Setup hook: Here is where you can initialize any things you need
```

components/notifications-service/integration_test/notification_test.go
```diff
@@ -0,0 +1,1207 @@
...
+		// We can verify the time objects are the same, though.
+		actualStartTime, err := time.Parse(time.RFC3339Nano, actual.StartTimeUTC)
+		assert.NoError(t, err)
+		assert.Equal(t, startTime, actualStartTime)
+
+		actualEndTime, err := time.Parse(time.RFC3339Nano, actual.EndTimeUTC)
+		assert.NoError(t, err)
+		assert.Equal(t, endTime, actualEndTime)
+
+		actualTimestamp, err := time.Parse(time.RFC3339Nano, actual.EndTimeUTC)
```
msorens, line 432: Curious that this test passed with this bug:
```suggestion
		actualTimestamp, err := time.Parse(time.RFC3339Nano, actual.TimestampUTC)
```

components/notifications-service/integration_test/notification_test.go
```diff
@@ -0,0 +1,1207 @@
...
+		assert.NotEmpty(t, actual.AutomateFQDN)
+		assert.Equal(t, "https://automate.example/nodes/0/runs/1", actual.AutomateFailureURL)
+
+		actualStartTime, err := time.Parse(time.RFC3339Nano, actual.StartTimeUTC)
+		assert.NoError(t, err)
+		assert.Equal(t, startTime, actualStartTime)
+
+		actualEndTime, err := time.Parse(time.RFC3339Nano, actual.EndTimeUTC)
+		assert.NoError(t, err)
+		assert.Equal(t, endTime, actualEndTime)
```
msorens, line 667: No time stamp check needed for this test?

components/notifications-service/integration_test/notification_test.go
```diff
@@ -0,0 +1,1207 @@
...
+		expectedBT := []string{"/path/to/recipe_example.rb: 42", "/path/to/recipe_example.rb: 23"}
+		assert.Equal(t, expectedBT, actual.ExceptionBacktrace)
+		// NOTE: the string values of the time will be different because of how the
+		// floating point value of the fractional seconds is converted to a string
+		// in different programming languages. E.g., trying to compare the strings
+		// can result in something like this:
+		// 		expected: "2021-01-29T19:11:09Z"
+		// 		actual  : "2021-01-29T19:11:09.000000Z"
+		// We can verify the time objects are the same, though.
+		actualStartTime, err := time.Parse(time.RFC3339Nano, actual.StartTimeUTC)
```
msorens, line 424: Consider putting these 3 time checks--with the explanatory comments--into a helper function to DRY out the code a bit.

components/notifications-service/integration_test/notification_test.go
```diff
@@ -0,0 +1,1207 @@
...
+		expectedFailureSnippet :=
+			`Chef client run failure on [https://automate.example/nodes/0] chefnode.example : https://automate.example/nodes/0/runs/1
+Example Exception Occurs
+EXAMPLE EXCEPTION MESSAGE
+(cookbook_example::recipe_example line 13) 
+`
+		assert.Equal(t, expectedFailureSnippet, actual.FailureSnippet)
+		assert.Equal(t, "Example Exception Occurs", actual.ExceptionTitle)
+		assert.Equal(t, "EXAMPLE EXCEPTION MESSAGE\n(cookbook_example::recipe_example line 13)", actual.ExceptionMessage)
+		expectedBT := []string{"/path/to/recipe_example.rb: 42", "/path/to/recipe_example.rb: 23"}
```
msorens, line 415: What is a "BT" ?

components/notifications-service/integration_test/notification_test.go
```diff
@@ -0,0 +1,1207 @@
...
+			assert.NotNil(t, postData)
+			actualURLPaths[postData.URL] = true
+		}
+		expectedURLPaths := map[string]bool{
+			"/slackalert/":      true,
+			"/servicenowalert/": true,
+			"/webhookalert/":    true,
+		}
+
+		assert.Equal(t, expectedURLPaths, actualURLPaths)
```
msorens, line 799: Is order of the alerts really significant? (But then again, is it really worth writing a test to check for order independence? 🤷 )

msorens (approved):
Wow! Very thorough. Some notes below, mostly suggestions for some further tests; only found one bug.

kmacgugan (approved):

components/notifications-service/integration_test/rules_crud_test.go
```diff
@@ -0,0 +1,409 @@
...
+func TestAddRuleValidations(t *testing.T) {
+	err := suite.DeleteEverything()
+	require.NoError(t, err)
+
+	t.Run("the Id field must be empty", func(t *testing.T) {
+		rule := validRule("TestAddRule_validations_Id")
+		rule.Id = "43"
+		response, err := suite.Client.AddRule(ctx, rule)
+		require.NoError(t, err)
+		assert.Equal(t, api.RuleAddResponse_VALIDATION_ERROR, response.Code)
```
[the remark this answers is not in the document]
danielsdeleo: My experience is that we are uh, consistently inconsistent about this. I think some services return an error for internal server errors but handle validations separately, etc.

components/notifications-service/integration_test/rules_crud_test.go
```diff
@@ -0,0 +1,409 @@
...
+		rule := validRule("TestAddRule_validations_Name")
+		rule.Name = ""
+		response, err := suite.Client.AddRule(ctx, rule)
+		require.NoError(t, err)
+		assert.Equal(t, api.RuleAddResponse_VALIDATION_ERROR, response.Code)
+		assert.Empty(t, response.Id)
+		assert.Contains(t, response.Messages, "Rule name must be supplied.")
+	})
+
+	t.Run("the Name must be unique", func(t *testing.T) {
```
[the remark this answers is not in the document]
danielsdeleo: It just cannot be blank or a duplicate. https://github.com/chef/automate/blob/22545b9d39d4a575e21e7c738a63076fb178cff6/components/notifications-service/server/lib/validator.ex#L28-L50

components/notifications-service/integration_test/notification_test.go
danielsdeleo: There are many subtle differences between the different formats associated with the different rule types. This one doesn't have the timestamp field: https://github.com/chef/automate/blob/22545b9d39d4a575e21e7c738a63076fb178cff6/components/notifications-service/server/lib/formatters/webhook.compliance.ex

components/notifications-service/integration_test/notification_test.go
danielsdeleo: This test should not depend on the order of the alerts, that's why I accumulated the URLs into a map. The existing system may accidentally produce a stable order (I haven't checked at all) but perhaps the stable order is brittle (e.g., order of rows returned from postgres when no sort is specified is often "stable" until you make an unrelated schema change)

components/notifications-service/integration_test/notification_test.go
danielsdeleo: "backtrace." Seemed fair to abbreviate heavily since we use it just once on the next line

components/notifications-service/integration_test/rules_crud_test.go
```diff
@@ -0,0 +1,409 @@
...
+		// update would violate the uniqueness constraint on the name column, so it
+		// returns an internal error when the db request fails on the contstraint
+		// violation. This is a bug, but we don't plan to change the behavior until
+		// after the elixir implementation is replaced.
+		assert.NotEqual(t, api.RuleUpdateResponse_OK, response.Code)
+		//assert.Equal(t, api.RuleUpdateResponse_DUPLICATE_NAME, response.Code)
+		//assert.Contains(t, response.Messages, "A rule with this name already exists")
+	})
+
+	t.Run("the Action field must not be empty", func(t *testing.T) {
```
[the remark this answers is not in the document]
danielsdeleo: That's covered by the `NOTE` below, I decided against covering every validation code path for the update cases because of diminishing returns.

components/notifications-service/integration_test/notification_test.go
danielsdeleo: fixed!

PR #4695, opened 2021-02-05
Dan/rm eol workflow

danielsdeleo:

Workflow is EOL. This PR removes it from the repo.

NOTE that this is a minimum effort version of workflow removal. Depending on whether anyone is actually using it still, we may need additional effort to provide a good experience

 https://blog.chef.io/new-chef-automate-functionality-and-end-of-life-impacts


.bldr.toml
```diff
@@ -888,18 +888,6 @@ paths = [
   "lib/user/*"
 ]
 
-[automate-workflow-nginx]
```
msorens, line 891: Just naively searching for "workflow" in the codebase I see these occurrences in this file; are they relevant?
```
"api/config/workflow_nginx/*"
"api/config/workflow_server/*"
"api/config/workflow_nginx/*"
"api/config/workflow_server/*"
"api/config/workflow_nginx/*"
"api/config/workflow_server/*"
```

And these directories directly tied to the above:
api/config/workflow_nginx/
api/config/workflow_server/

dev-docs/a2-workflow.md

Seems to be a `--enable-workflow` flag in a variety of places, but not sure if that means the same "workflow" you are dealing with here.
https://github.com/chef/automate/blob/93cba9db02fc243a6dfbd9243a040382fc93b5f8/terraform/test-environments/modules/chef_automate_install/main.tf#L37

Also, here are a few others that might be relevant in other files:
https://github.com/chef/automate/blob/d096d223caf66106aa0f46ab8dfae3f56172b79b/.envrc#L11
https://github.com/chef/automate/blob/b6fd8b84196dd8e5dada741333000e5f5f0ac429/.expeditor/update_chef_server.sh#L45
https://github.com/chef/automate/blob/3d88cc006f1dde347873be85c6f8ce5e001d2c95/api/config/deployment/automate_config.proto#L41-L42
https://github.com/chef/automate/blob/eb6b8a6ce85b3b161b51a67f470d132f98a5ef2d/components/automate-deployment/a1-migration/setup.rb#L248-L249
https://github.com/chef/automate/blob/443f798abe38d90ef4d3c4d959ee631090774126/components/automate-deployment/pkg/manifest/client/http_test.go#L111-L112
https://github.com/chef/automate/blob/baa64657033a7aa5f9972e201b60ed4912dad30c/components/automate-load-balancer/habitat/plan.sh#L52
https://github.com/chef/automate/blob/1ca3deb611fa1ab5bbf28cfb55179011e16d4a3e/components/automate-load-balancer/habitat/config/automate-workflow-nginx-location.conf#L1
https://github.com/chef/automate/blob/1ca3deb611fa1ab5bbf28cfb55179011e16d4a3e/components/automate-load-balancer/habitat/config/automate-workflow-nginx-upstream.conf#L1
https://github.com/chef/automate/blob/37ecc1f4b0c3deba4ec0e2c00cee8aa99185f67d/inspec/a2-migrate-from-v1-integration/controls/file_data.rb#L40

msorens (commented):
That is a lot of nice chopping.
I am all for minimizing but have noted a few other references below that may be relevant; not familiar enough to know for sure.

danielsdeleo:
I'm gonna put this one on hold for now. There could possibly be some unfortunate outcomes if we merge this prior to the 2.1 release. Here are some concerns we should be thinking about when we finish this:

* If there are customers still using workflow and they plan to "get by" on EOL'd Automate for a while because they haven't built the replacement thing yet, and we merge this removal prior to the 2.1 milestone release, then those customers would be at risk of losing workflow functionality in a patch upgrade. This wouldn't violate any commitments we've made, but still not the best experience.
* What will we do with the workflow configuration? In particular, there may be saved configuration state within the deployment service's database, and customers could potentially have some config baked into automation/scripts/etc.
* Will we clean up databases and files automatically if the user upgrades a system with workflow enabled? Contrarily we might refuse to upgrade until the user affirms they are ok with that data being deleted(?)

PR #4710, opened 2021-02-08
Remove advice to call CS

kagarmoe:

We fixed a few issues that used to be tricky. This PR removes advice to contact customer support from the Automate docs, except for in the doc about migrating from A1 to A2.

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/components/docs-chef-io/content/automate/data_lifecycle.md
+++ b/components/docs-chef-io/content/automate/data_lifecycle.md
@@ -16,13 +16,6 @@ Data Lifecycle manages the retention of events, service groups, Chef Infra Clien
 Chef Automate stores data from the ingest-service,event-feed-service, compliance-service and applications-service in Elasticsearch or PostgreSQL.
 Over time, you may wish to remove that data from Chef Automate by using the data lifecycle settings.
 
-{{% warning %}}
-Note: Chef Automate data retention processes changed in 20191129172405.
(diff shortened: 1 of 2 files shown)


nrgetik:
> The directions provided below are intended for use only during initial deployment of Chef Automate.

@jaym Would this still be true given earlier internal clarifications around this? Should we mention instead that implementing such configuration in an environment that isn't new would not automatically kick off a migration of data to the externalized ES, but would technically work otherwise?

nrgetik (approved):

nrgetik:
Apparently I'm still terrible at GH reviews, sorry about that. I'll put this here for now:

> The directions provided below are intended for use only during initial deployment of Chef Automate.

@jaym Would this still be true given earlier internal clarifications around this? Should we mention instead that implementing such configuration in an environment that isn't new would not automatically kick off a migration of data to the externalized ES, but would technically work otherwise?

jaym:
that comment is still true. its not 100% clear to me switching while automate is running is safe and we'd need to test it out. At one point, if you switched while automate was running, not all the services would restart, and if they were writing data, they would write to a new ES without first creating the scheamas

nrgetik (approved):

danielsdeleo (approved):

IanMadd (approved):

PR #4712, opened 2021-02-08
Notifications Rewrite: Service Stub

danielsdeleo:

We are rewriting the notifications service in go to align better with our engineering capabilities.

This change creates a `notifications-service2` component in the source tree. It can be compiled and packaged manually, but is hidden from pipeline builds using the `dev_only` option in the habitat `plan.sh` (the `automate-ui-devproxy` uses the same mechanism).

The `notifications-service2` is missing most of the functionality that it will eventually have/need. Currently it only implements configuration file loading. When run as a service it prints its config to the logs in a loop. This is a first step towards implementing the infrastructural components of the new notifications service.

In a fresh habitat studio, `start_all_services`. Then run `build components/notifications-service2/ && sl`. The service should build without error and then get picked up by the deployment service. In the log output you will see the running configuration for the service.

The unit tests for the config package can be run by `cd /src/components/notifications-service2` and `go test -v pkg/config/*go`

### DIFF
--- a/components/notifications-service2/pkg/storage/postgres/schema/sql/01_one_dot_zero_migrations.up.sql
+++ b/components/notifications-service2/pkg/storage/postgres/schema/sql/01_one_dot_zero_migrations.up.sql
@@ -1,0 +1,65 @@
+-- ONE DOT ZERO MIGRATIONS:
+-- The migrations in this file are imported from the previous elixir
+-- implementation of the service.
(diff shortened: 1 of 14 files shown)


components/notifications-service2/pkg/storage/postgres/schema/sql/01_one_dot_zero_migrations.up.sql
```diff
@@ -0,0 +1,65 @@
+-- ONE DOT ZERO MIGRATIONS:
```
danielsdeleo, line 1: I'm going to handle migrating the database to a consistent state in a future PR. For now having this file here makes the schema migration handling logic in the plan.sh happy.

PR #4720, opened 2021-02-11
Notifications Service Rewrite: Implement Postgres storage and Schema Management

danielsdeleo:

We are rewriting the notifications service in go to better align with our engineering capabilities. In this PR we continue to implement the infrastructural components of the new implementation. 

This PR implements the lower-level database component for the new service. The only behavior implemented is to connect to Postgres and migrate the schema. Integration testing is added to verify that the migration tooling can correctly migrate an existing database/schema (from the elixir implementation) and the tooling can also correctly migrate an empty database instance (i.e., the fresh install case).

**UPGRADE CASE:**
* In a fresh hab studio, `start_all_services`
* build it and watch it get deployed: `build components/notifications-service2/ && sl`
* The build should succeed and you should see the service come up without error
* In the logs you should see the migrations from the new service happen. The relevant part looks like this:
```
msg="running db migrations from \"/hab/pkgs/YOUR_ORIGIN/notifications-service/2.0.0/20210211202635/schema\""
msg="Current schema version: 0 dirty: false"
msg="1/u one_dot_zero_migrations (15.2796ms)\n"
msg="2/u drop_old_migrations_table (29.0875ms)\n"
msg="Completed db migrations"
```
**FRESH INSTALL:**
This implementation isn't yet complete enough to start all of Automate with it, as it doesn't accept network connections on the notifications service port. Here we want to verify only that the service can start (e.g., no crash loop) and correctly manage the schema in a fresh database.
* In a fresh studio, build the new package before starting any services: `build components/notifications-service2/`
* `start_deployment_service`
* `chef-automate dev deploy-some chef/notifications-service --with-deps` -- when you get to the `Checking service health` spinner, you can ctrl-C.
* Use `chef-automate dev psql notifications_service` to connect to the database and verify the tables are there. `\d+` to show the tables and (e.g.) `\d+ rules` to see a table. You should see this:
```
notifications_service=# \d+
                               List of relations
 Schema |       Name        | Type  |     Owner     |    Size    | Description 
--------+-------------------+-------+---------------+------------+-------------
 public | processed_events  | table | notifications | 0 bytes    | 
 public | rules             | table | notifications | 8192 bytes | 
 public | schema_migrations | table | notifications | 8192 bytes | 
(3 rows)

notifications_service=# \d+ rules
                                                Table "public.rules"
         Column         |       Type        | Collation | Nullable | Default | Storage  | Stats target | Description 
------------------------+-------------------+-----------+----------+---------+----------+--------------+-------------
 id                     | uuid              |           | not null |         | plain    |              | 
 name                   | text              |           | not null |         | extended |              | 
 event                  | rule_event        |           | not null |         | plain    |              | 
 action                 | rule_action       |           | not null |         | plain    |              | 
 url                    | text              |           | not null |         | extended |              | 
 secret_id              | character varying |           |          |         | extended |              | 
 critical_controls_only | boolean           |           |          | false   | plain    |              | 
Indexes:
    "rules_pkey" PRIMARY KEY, btree (id)
    "rules_name_key" UNIQUE CONSTRAINT, btree (name)
```

**INTEGRATION TESTS:**
* You need to have automate deployed with either version of the notifications-service running.
* `cd /src/components/notifications-service2/`
* `go test -v integration_test/*go -count 1` -- this has scenarios that verify upgrade, fresh install, and schema update retries.


components/notifications-service2/pkg/storage/postgres/schema/sql/01_one_dot_zero_migrations.up.sql
```diff
@@ -1,65 +1,81 @@
...
-----  ALTER TABLE rules ADD COLUMN critical_controls_only BOOLEAN;
-----  ALTER TABLE rules ALTER COLUMN critical_controls_only SET DEFAULT FALSE;
-----  
-----  UPDATE rules SET critical_controls_only=FALSE;
-----  DELETE from rules where event = 'Assets'
+CREATE TABLE IF NOT EXISTS migrations (num INTEGER NOT NULL, descr TEXT, at TIMESTAMPTZ NOT NULL);
+
+DO $$
+BEGIN
+    IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'rule_event') THEN
```
danielsdeleo, line 15: Postgresql does not have `CREATE TYPE IF NOT EXISTS` so we have to implement idempotent behavior manually here.

PR #4738, opened 2021-02-16
Notifications Rewrite: gRPC Implementation Stub

danielsdeleo:

We are rewriting the notifications service in go to improve alignment with our organization's capabilities.

In this PR, we implement all the dependencies necessary to implement the notifications gRPC interface with stub implementations of the gRPC methods.

* Load service certificates, listen on the configured socket with the standard mutual TLS configuration used by other services.
* Instantiate a server struct that implements the required methods to satisfy the interface for `NotificationsServer` (defined in the generated code
* Implement standard health checking for the server
* Copy the proto files from the notifications service to `api/interservice`. Eventually we wish to remove the code that currently is located in `components/notifications-service`, which includes the proto files for the notifications service. Therefore it is a good opportunity to standardize the file locations. As part of this process, the _protobuf_ package name is changed from `notifications` to `chef.automate.domain.notifications.service`, which also improves conformance with our standards. NOTE, however that this makes the implementations incompatible at the wire level. For example, when accessing a new server with the old client, an error would be returned with the message `unknown service notifications.Notifications` (because the service is now `chef.automate.domain.notifications.service.Notifications`). That said, we do not provide any compatibility guarantees for interservice APIs and the go packages are interchangeable (with a named import). If necessary we can update the existing protos to change the package names, but in either case we should only need to make a mechanical change to import paths when we replace the elixir notifications service.
* Move the integration tests from the notifications-service component to the notifications-service2 component. We will now focus on using these tests to produce a compatible reimplementation. The tests are also updated to better handle some failure cases that occur with the stub implementations of the gRPC methods.

* In a fresh studio, `build components/notifications-service2/` then `start_all_services`
* `cd /src/components/notifications-service2/` then `go test -v integration_test/*go -count 1` -- the vast majority of the tests are expected to FAIL, however the _way_ things fail shows that the service is responding to gRPC requests and also that the test codebase is able to complete the test suite without hanging or panic.

I don't know of a clean way to show that the health checking is working. I resorted to modifying the `Health()` method in `server.go` like this:

```go
func (s *Server) Health() *health.Service {
	panic("health was checked")
	return s.health
}
```

Then I ran `build components/notifications-service2/ && sl` and observed that the health checks run by tab-sup triggered the expected panic/crash in the server.


api/interservice/notifications/service/events.proto
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    string id = 1;
+    float impact = 2;
+    string title = 3;
+    string code = 4;
+    string desc = 5;
+    SourceLocation source_location = 6;
+    repeated Refs refs = 7;
+    reserved 8;// google.protobuf.Struct tags = 8;
```
msorens, line 50: Is this for future use, or leftover, or what? Not sure what the commented item is supposed to be conveying here.

api/interservice/notifications/service/events.proto
```diff
@@ -0,0 +1,173 @@
+syntax = "proto3";
+
+package chef.automate.domain.notifications.service;
```
msorens, line 3: FWIW there are actually only three other services that end with the literal `.service` in the package name -- so those are the aberrations rather than the norm.

api/interservice/notifications/service/rules.proto
```diff
@@ -0,0 +1,132 @@
...
+    Empty none = 4;
+  }
+}
+
+message URLValidationResponse {
+  enum Code {
+    OK = 0;  // we made the request and it succeeded with a 20x response
+    ERROR = 1; // We made the request and it failed for any reason.
+    INVALID_URL = 4; // the url was not valid.
+    NOTIFICATIONS_UNAVAIALBLE = 98; // we could not make the request due to capacity
```
msorens, line 64:
```suggestion
    NOTIFICATIONS_UNAVAILABLE = 98; // we could not make the request due to capacity
```

components/notifications-service2/integration_test/rules_crud_test.go
```diff
@@ -409,7 +412,7 @@ func TestUpdateRuleValidation(t *testing.T) {
 	// as AddRule does. We don't explicitly test those because it's repetitive
 }
 
-func TestDeleteRule(t *testing.T) {
```
msorens, line 412: Not this line, but just above (L394) there is a typo: "contstraint".

msorens (approved):
Code read-through looks good; just a couple very minor items noted.

api/interservice/notifications/service/events.proto
danielsdeleo: This is a straight copy of what was previously in the notifications-service's source tree. I have opted to make fewer changes here for now in order to ensure that the generated go code for the proto package will be a drop-in replacement, so that the cutover from the old notifications service to the new one will be a simple as possible. I would like to do a cleanup pass on these protos after that.

api/interservice/notifications/service/events.proto
danielsdeleo: Unfortunately, `chef.automate.domain.notifications` is used for the `api/config` protos for notifications. So this is the most conforming thing I could think of :(

api/interservice/notifications/service/rules.proto
danielsdeleo: I'll add this to the list for cleanup after the cutover.

components/notifications-service2/integration_test/rules_crud_test.go
danielsdeleo: I will add this to the next PR I'm working on.

PR #4747, opened 2021-02-18
Notifications Rewrite: Run Lint/Unit in Ci

danielsdeleo:

Add linting and a basic Makefile to notifications-service2. Run the `ci` make target in Ci, which will run the linter and unit tests both. These are quite fast so it would be slower to parallelize them than to run in serial.

In the process it was necessary to upgrade golangci-lint. A bug in the tool prevented it from analyzing the notifications-service2. I attempted to be as conservative as possible. The (latest version of golangci-lint is 1.37.0)[https://github.com/golangci/golangci-lint/tags]; in this pull request I only upgraded us from 1.21.0 to 1.26.0. The reason for this is each version adds new checks, many of which fail in many components of the codebase. I am trying to keep this patch scoped to only adding testing and static analysis to the new notifications service implementation. In this spirit, I have disabled the new checks that I have found to fail on existing code. Some of these new checks seem kinda dubious, but `govet`'s `copylocks` is probably one we would like to enable. Nevertheless, this should be done in a separate work stream.

### :+1: Definition of Done

Same as the build and test steps below.

* view the "verify" build for this PR. It should include a `[unit] notifications-service2` job in the matrix
* In the studio, `cd /src/components/notifications-service2` and run `make ci` -- you should see a `go build`, `golangci-lint run` and `go test` run (with a lot of other args, etc.)
* `make ci` should also work on your host system if you have golang installed.


components/notifications-service2/Makefile
```diff
@@ -0,0 +1,23 @@
...
+	rm -f notifications-service
+
+test:
+	GOMAXPROCS=4 go test -v -cover $(GOFILES)
+
+.PHONY: ci review setup clean
+.PHONY: build clean fmt lint proto test
+.PHONY: edit view
+.PHONY: run
+.PHONY: generate
```
msorens, line 23: Only need to pay for what you use. 😁 
```suggestion
.PHONY: ci build clean test
```

components/notifications-service2/cmd/notifications-service/commands/serve.go
```diff
@@ -46,5 +46,4 @@ func runServeCommand(cmd *cobra.Command, args []string) error {
 		fmt.Println(string(s))
 		time.Sleep(5 * time.Second)
 	}
-	return nil
```
msorens, line 49: Don't we need to always return something?

msorens (approved):

components/notifications-service2/cmd/notifications-service/commands/serve.go
danielsdeleo: I guess go know's it's unreachable since the loop is never broken. `govet` failed on it 🤷‍♂️

PR #4755, opened 2021-02-22
Fix typo in proto package path

msorens:

Fix typo in proto package path.
I am guessing there are no repercussions to this change since all Buildkite verifications are passing.

Because it is effectively deleting one package and adding another, it is a breaking change to protos so also have to adjust the buf.yaml configuration temporarily.

NA

### :+1: Definition of Done
Product name spelled correctly.

NA

- [ ] Tests added/updated?
- [ ] Docs added/updated?

### DIFF
--- a/buf.yaml
+++ b/buf.yaml
@@ -415,7 +415,8 @@ lint:
     - external/habitat/event.proto # this file is owned by habitat and copied here
 breaking:
   use:
-    - PACKAGE
+    # TODO: TEMPORARY CHANGE JUST FOR THIS PR!
+    - FILE
   ignore:
     # Changes to interservice APIs are okay until (if ever) we support clusters
     # of interconnected internal services that would not upgrade in unison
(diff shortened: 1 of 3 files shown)


buf.yaml
```diff
@@ -415,7 +415,8 @@ lint:
     - external/habitat/event.proto # this file is owned by habitat and copied here
 breaking:
   use:
-    - PACKAGE
+    # TODO: TEMPORARY CHANGE JUST FOR THIS PR!
+    - FILE
```
msorens, line 419: Need to allow the breaking change for this commit; will revert after this PR merges.

kagarmoe: Thanks! I appreciate that you're fixing these errors!

danielsdeleo (approved):
👍 Thanks for fixing! This will work just fine because the client (automate-gateway) and server (CDS service) will change simultaneously (this will be the case as long as we don't support splitting the gateway and domain svcs on to different hosts).

Issue #4828, opened 2021-03-16
Make first focused element of Chef-Modal more flexible


SEAjamieD:
## User Story
As a developer, I want the ability to have the first focused element be more flexible depending on the type of modal i'm opening.

## Definition of Done
Developer can choose which element is focused first without the need for specialized code or overwriting

PR #4829, opened 2021-03-16
UI improvements for the edit mode

vinay033:
### :nut_and_bolt: Description: What code changed, and why?
UX identified some design issues in edit mode, this PR covered some of UX improvement changes.

### :chains: Related Resources
https://github.com/chef/automate/issues/4808
### :+1: Definition of Done
I have covered UX improvements which are mentioned below
1. Delete Data Bag - Button text: “Delete” ->  change to delete
2. Create/Remove Chef Server - The button on modal dialogue says “Delete” where on the drop down it says “Remove” we should make these the same. -  changed to delete
3. When you hover policy files, details page- the hover effect has some glitch. it not coverup the beta tag.
4. Client list page - Create client button misaligned
5. Delete popup modal width issue
6. Roles list page is misaligned when there is no data.
7. FIx pagination issue for the clients
### :athletic_shoe: How to Build and Test the Change
```
Go to the `infrastructure tab` from the top navigation bar then you can see the `Chef Server`
If you have data then you can see the list of servers --> then you can check all the mentioned points.
```
### :white_check_mark: Checklist

- [ ] I have read the [CONTRIBUTING document](https://github.com/chef/automate/blob/master/CONTRIBUTING.md).
- [ ] Tests added/updated?
- [ ] Docs added/updated?
- [ ] All commits have been signed-off for the [Developer Certification of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

PR #4830, opened 2021-03-17
Jamie/4828 modal first focus flexible

SEAjamieD:
Looking for a more flexible way to decide where the first focus on a modal should be so that we can provide a better UX experience to our customers.  This update will allow developers to more easily decide what gets focused first without hacky javascript overwriting.

addresses https://github.com/chef/automate/issues/4828

### :+1: Definition of Done
A developer is able to place `firstFocus` on an element that they would like to have focused first when opening a modal.

Terminal 1
in Automate Folder 
`hab studio enter`
`build components/automate-ui-devproxy && start_all_services`
If you do not currently have servers in your sample data you may need: `infra_service_load_sample_data`

Terminal 2
in Automate/components/automate-ui
`make update-ui-lib`
`make serve`

navigate to https://a2-dev.test/infrastructure/chef-servers
click on a server in your list
click on 'Add Chef Organization' button
when the modal opens, you should notice the first text input is automatically focused

This has now been activated in each of the "Add" modals within infrastructure

- [x] Tests added/updated?
- [x] Docs added/updated?

Example After:
first focus example

### DIFF
--- a/components/automate-ui/src/app/modules/infra-proxy/create-data-bag-modal/create-data-bag-modal.component.html
+++ b/components/automate-ui/src/app/modules/infra-proxy/create-data-bag-modal/create-data-bag-modal.component.html
@@ -5,8 +5,8 @@
       <div class="margin">
         <chef-form-field>
           <label>
-            <span class="label">Name <span aria-hidden="true">*</span></span>
-            <input chefInput formControlName="name" type="text" id="name-input" (keyup)="handleInput($event)" autocomplete="off"/>
+            <span class="label">Name</span>
+            <input chefInput firstFocus formControlName="name" type="text" id="name-input" (keyup)="handleInput($event)" autocomplete="off"/>
           </label>
           <chef-error
           *ngIf="(createForm.get('name').hasError('required') || createForm.get('name').hasError('pattern')) && createForm.get('name').dirty">
--- a/components/automate-ui/src/app/modules/infra-proxy/create-org-modal/create-org-modal.component.ts
+++ b/components/automate-ui/src/app/modules/infra-proxy/create-org-modal/create-org-modal.component.ts
@@ -75,6 +75,6 @@ export class CreateOrgModalComponent implements OnInit, OnChanges {
   }
 
   private isNavigationKey(event: KeyboardEvent): boolean {
-    return event.key === 'Shift' || event.key === 'Tab';
+    return event.key === 'Shift' || event.key === 'Tab' || event.key === 'Enter';
   }
 }
--- a/components/automate-ui/src/app/modules/infra-proxy/create-org-modal/create-org-modal.component.ts
+++ b/components/automate-ui/src/app/modules/infra-proxy/create-org-modal/create-org-modal.component.ts
@@ -9,6 +9,7 @@ import { Component,
 import { FormGroup } from '@angular/forms';
 import { IdMapper } from 'app/helpers/auth/id-mapper';
 import { ProjectConstants } from 'app/entities/projects/project.model';
+import { Utilities } from 'app/helpers/utilities/utilities';
 
 @Component({
   selector: 'app-create-org-modal',
@@ -49,7 +50,7 @@ export class CreateOrgModalComponent implements OnInit, OnChanges {
   }
 
   handleNameInput(event: KeyboardEvent): void {
-    if (!this.modifyID && !this.isNavigationKey(event)) {
+    if (!this.modifyID && !Utilities.isNavigationKey(event)) {
       this.conflictError = false;
       this.createForm.controls.id.setValue(
         IdMapper.transform(this.createForm.controls.name.value.trim()));
@@ -57,7 +58,7 @@ export class CreateOrgModalComponent implements OnInit, OnChanges {
   }
 
   public handleInput(event: KeyboardEvent): void {
-    if (this.isNavigationKey(event)) {
+    if (Utilities.isNavigationKey(event)) {
       return;
     }
     this.conflictError = false;
@@ -74,7 +75,4 @@ export class CreateOrgModalComponent implements OnInit, OnChanges {
     this.checkedProjectIDs = [];
   }
 
-  private isNavigationKey(event: KeyboardEvent): boolean {
-    return event.key === 'Shift' || event.key === 'Tab' || event.key === 'Enter';
-  }
 }
--- a/components/chef-ui-library/package-lock.json
+++ b/components/chef-ui-library/package-lock.json
@@ -2063,7 +2063,7 @@
     },
     "es6-promisify": {
       "version": "5.0.0",
-      "resolved": "https://registry.npmjs.org/es6-promisify/-/es6-promisify-5.0.0.tgz",
+      "resolved": "http://registry.npmjs.org/es6-promisify/-/es6-promisify-5.0.0.tgz",
       "integrity": "sha1-UQnWLz5W6pZ8S2NQWu8IKRyKUgM=",
       "dev": true,
       "requires": {
--- a/components/chef-ui-library/package-lock.json
+++ b/components/chef-ui-library/package-lock.json
@@ -2063,7 +2063,7 @@
     },
     "es6-promisify": {
       "version": "5.0.0",
-      "resolved": "http://registry.npmjs.org/es6-promisify/-/es6-promisify-5.0.0.tgz",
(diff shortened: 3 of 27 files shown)


components/automate-ui/src/app/modules/infra-proxy/create-data-bag-modal/create-data-bag-modal.component.html
```diff
@@ -6,7 +6,7 @@ <h2 slot="title">Create Data Bag</h2>
         <chef-form-field>
           <label>
             <span class="label">Name</span>
-            <input chefInput formControlName="name" type="text" id="name-input" (keyup)="handleInput($event)" autocomplete="off"/>
+            <input chefInput firstFocus formControlName="name" type="text" id="name-input" (keyup)="handleInput($event)" autocomplete="off"/>
```
SEAjamieD, line 9: simple proof of concept

components/automate-ui/src/app/modules/infra-proxy/create-org-modal/create-org-modal.component.ts
```diff
@@ -68,6 +68,6 @@ export class CreateOrgModalComponent implements OnInit, OnChanges {
   }
 
   private isNavigationKey(event: KeyboardEvent): boolean {
-    return event.key === 'Shift' || event.key === 'Tab';
+    return event.key === 'Shift' || event.key === 'Tab' || event.key === 'Enter';
```
SEAjamieD, line 71: In some of the modals, this may need to be adjusted so that when using keyboard navigation and the "Enter" key to open the modal, enter isn't recorded.  Nothing happens in the view, but an error is thrown in the console.

components/chef-ui-library/src/molecules/chef-modal/chef-modal.tsx
```diff
@@ -168,6 +177,26 @@ export class ChefModal {
...
+    if (lockStatus) {
+      return modal
+    } else if (firstFocus) {
+      return firstFocus.tagName === 'CHEF-BUTTON'
+        ? firstFocus.firstElementChild as HTMLElement
+        : firstFocus
+    }
+
+    return closeFocus;
+  }
```
SEAjamieD, line 198: This function has been updated so that when unlocked, it's searching for the `firstFocus` attribute in the modal.  If it finds that attribute, it will also check if it is a `chef-button` and then apply focus to the correct element.

I needed to add an exception for `chef-button` because the actual button element is nested below the `chef-button` tag.

`firstFocus` will work with any of the native input elements but may need similar alteration for other `chef-`elements.  I don't believe we have a need for many though.  Often times we are looking simply for a text input or a 'cancel' button.

SEAjamieD:
@jong-progress @vinay033 - This is sort of "proof of concept" around what we talked about today, giving developers an easier way to select the correct autofocused element without extra legwork.  The developer will add the attribute `firstFocus` to the element they'd like focused, otherwise it will default to the current implementation of the 'X' close button or the full modal.

@vinay033 and @msorens - would appreciate if you could poke holes in advancing this implementation.  The idea is to make it easier to implement some of the new UX paths

components/chef-ui-library/package-lock.json
```diff
@@ -2063,7 +2063,7 @@
     },
     "es6-promisify": {
       "version": "5.0.0",
-      "resolved": "https://registry.npmjs.org/es6-promisify/-/es6-promisify-5.0.0.tgz",
+      "resolved": "http://registry.npmjs.org/es6-promisify/-/es6-promisify-5.0.0.tgz",
```
msorens, line 2066: This should be reverted--always want https in there. Recall there was an issue which we solved for automate-ui; looks like the same fix is needed in chef-ui-library. (IIRC it was a caching issue of sorts.)

components/chef-ui-library/src/molecules/chef-modal/chef-modal.tsx
```diff
@@ -168,6 +177,26 @@ export class ChefModal {
     return '';
   }
 
+  // when Angular detects the modal is open
+  // it sets the focus by default on the close button for unlocked modals,
+  // or the modal div for locked modals
+  // Developer can specify element to focus first using firstFocus attribute
```
msorens, line 183: Need to say this only applies to unlocked modals.

components/chef-ui-library/src/molecules/chef-modal/chef-modal.tsx
msorens: It sounds like we could genericize this (without having to check for `chef-button` specifically) -- I am thinking that a native HTML input element would not have children (or only text node children), so that is how you could distinguish.

msorens (commented):
I like the notion. A few initial comments here; still need to play with it a bit.

components/automate-ui/src/app/modules/infra-proxy/create-data-bag-modal/create-data-bag-modal.component.html
vinay033: great.

msorens:
Missing a step in your repro steps: you said "click on a server in your list". There are no servers in the list. 😉

msorens:
Here is an interesting artifact (outside the scope here, of course):
image
The chef-button bounding box is not the same as the contained button bounding box!

components/automate-ui/src/app/modules/infra-proxy/create-org-modal/create-org-modal.component.ts
msorens: Good catch. I would like to suggest that this `isNavigationKey` should go into a utilities folder (perhaps under automate-ui/src/app/helpers/) rather than redefine it 10 places. Want an easy way to see the variations? Use semgrep!
```
$ semgrep -e 'isNavigationKey(...) {...}' --lang ts
src/app/components/create-object-modal/create-object-modal.component.ts
139:  private isNavigationKey(event: KeyboardEvent): boolean {
140:    return event.key === 'Shift' || event.key === 'Tab';
141:  }

src/app/modules/infra-proxy/create-chef-server-modal/create-chef-server-modal.component.ts
53:  private isNavigationKey(event: KeyboardEvent): boolean {
54:    return event.key === 'Shift' || event.key === 'Tab';
55:  }

src/app/modules/infra-proxy/create-org-modal/create-org-modal.component.ts
70:  private isNavigationKey(event: KeyboardEvent): boolean {
71:    return event.key === 'Shift' || event.key === 'Tab' || event.key === 'Enter';
72:  }

src/app/page-components/create-user-modal/create-user-modal.component.ts
154:  private isNavigationKey(event: KeyboardEvent): boolean {
155:    return event.key === 'Shift' || event.key === 'Tab';
156:  }

src/app/pages/+compliance/+node-credentials/create-node-credential-modal/create-node-credential-modal.component.ts
150:  private isNavigationKey(event: KeyboardEvent): boolean {
151:    return event.key === 'Shift' || event.key === 'Tab';
152:  }

. . .
```

vinay033:
> Here is an interesting artifact (outside the scope here, of course):
> image
> The chef-button bounding box is not the same as the contained button bounding box!

I have fixed this in UI improvements PR :)

PR: https://github.com/chef/automate/pull/4829

components/automate-ui/src/app/modules/infra-proxy/create-org-modal/create-org-modal.component.ts
SEAjamieD: this is beautiful ✨ 

And great idea, I'll incorporate that in the final product

components/chef-ui-library/src/molecules/chef-modal/chef-modal.tsx
SEAjamieD: Agreed - I would like to _not_ need to adjust in the future.  Checking for children is a good idea - i'll toy with that.

SEAjamieD: I played with this but came to discover that other possible elements that might want to be focused first, like a `chef-select` will have child nodes, but should't focus the child.  

So as it sits, I think this is the best way to do this for our needs for the current expected implementation.

vinay033 (commented):
Great work @SEAjamieD 🎉

vinay033 (approved):

himanshi-chhabra (approved):

Issue #4719, opened 2021-02-11
Add SELinux configuration to preflight checks


gscho:
## Describe the bug
Chef Automate fails to install if SELinux is enabled an not in `Permissive` mode

## To Reproduce
- Create a centos8 server with SELinux enabled
- Install Chef Automate
- See that the hab supervisor fails to start successfully

## Expected behavior
The preflight checks fail due to SELinux being enabled

## Versions (please complete the following information):
- OS: centos 8
- Latest stable version of Automate as of writing

## Additional context
Additionally, this leads to an issue where the credentials generated by a successful automate install fail to be generated. Any time you run a subsequent `chef-automate deploy`, the CLI reports that automate is already installed in its preflight check so even if you correct the SELinux issue, the credentials are never generated. You can work around this by using the `chef-automate iam` commands to generate a new set of credentials.

PR #4881, opened 2021-03-26
Check if SELinux is enabled during preflight

gscho:
Signed-off-by: gscho <greg.c.schofield@gmail.com>

SELinux is enabled by default on Centos and Redhat systems. When installing automate, there is no preflight check performed to see if SELinux is set to permissive or disabled which is a requirement for the install to success. What this leads to is a install which seems to succeed but actually the habitat supervisor fails to start properly. The leads to a weird state because you cannot run the deploy command again due to the `is_a2_deployed` check failing. The user must figure out the issue and restart the hab supervisor themselves.

#4719 

go test pkg/preflight/preflight_test.go

- [x] Tests added/updated?
- [ ] Docs added/updated?


shivangi-singh1:
Hi @gscho , 
In the above PR when the SELinux  is disabled we still  get the following error  in the preflight check as shown in the below image.
Screenshot 2022-06-13 at 1 00 28 PM

components/automate-deployment/pkg/preflight/checks.go
```diff
@@ -34,6 +34,39 @@ func RootUserRequiredCheck() Check {
...
+				t.ReportSuccess("SELinux is not enabled")
+				return nil
+			}
+			out, err := t.SELinuxStatus()
+			if err != nil {
+				logrus.WithError(err).Debug("Unable to determine SELinux status using 'getenforce'")
+				t.ReportFailure("Failed to determine if SELinux is enabled")
+				return nil
+			}
+			if strings.EqualFold(string(out), "disabled") || strings.EqualFold(string(out), "permissive") {
```
shivangi-singh1, line 59:
```suggestion
			var outValue = strings.TrimSuffix(string(out), "\n")
			if strings.EqualFold(outValue,"disabled") || strings.EqualFold(outValue , "permissive") {
```

shivangi-singh1:
> @shivangi-singh1 Looks like this functionality was already added in another PR? You can close this one if that is the case.

Sure @gscho. Thanks for letting us know!

PR #4914, opened 2021-04-06
Casing fix

SEAjamieD:
Simple Fix - Casing was wrong in e2e test


kagarmoe (approved):
Indeed. Use sentence case for this.

PR #4967, opened 2021-04-16
Jamie/clone e2e issue

SEAjamieD:
Ignore - for pipeline testing only


kalroy (approved):
Tried a couple of times and based on the tests this indeed looks like the fix for the pipeline issue. Needs more work but worth a try.

PR #5213, opened 2021-06-08
Praj+vinay/update chef server to 14.4.4

PrajaktaPurohit:
Update Infra Server from 14.1.0 to 14.4.4

This change includes the following:

- Add support for external Elasticsearch 7.0 for new versions of
  Automate
- Includes some security related updates
- Adds an omnibus configurable for HSTS max age

More details can be found at: https://docs.chef.io/release_notes_server/

Signed-off-by: Prajakta Purohit <prajakta@chef.io>


tas50:
We will want to make sure to call out the benefits of this in our release notes. Some, but not all of the changes can be found here: https://docs.chef.io/release_notes_server

Note: Most of these changes are omnibus only so don't impact Automate.

PR #6575, opened 2022-01-17
Sahiba/test

Sahiba31:
Added Bucket Lifecycle configurations for deleting records automatically after one day.


Sahiba31:
Not Required

PR #6577, opened 2022-01-17
Auto Delete after one Day

Sahiba31:
Signed-off-by: Sahiba Goyal <sgoyal@progress.com>

### DIFF
--- a/components/report-manager-service/worker/worker.go
+++ b/components/report-manager-service/worker/worker.go
@@ -273,6 +273,7 @@ func (t *GenerateReportTask) Run(ctx context.Context, task cereal.Task) (interfa
 		return nil, err
 	}
 
+	//If lifecycles are set
 	if lifecycles != nil {
 		flag := false
 		for i := range lifecycles.Rules {
--- a/components/report-manager-service/worker/worker.go
+++ b/components/report-manager-service/worker/worker.go
@@ -275,14 +275,14 @@ func (t *GenerateReportTask) Run(ctx context.Context, task cereal.Task) (interfa
 
 	//If lifecycles are set
 	if lifecycles != nil {
-		flag := false
-		for i := range lifecycles.Rules {
-			if !lifecycles.Rules[i].Expiration.IsNull() {
-				flag = true
+		isExpirationEnabled := false
+		for _, rule := range lifecycles.Rules {
+			if !rule.Expiration.IsNull() {
+				isExpirationEnabled = true
 				break
 			}
 		}
-		if !flag {
+		if !isExpirationEnabled {
 			newRule := lifecycle.Rule{
 				ID:     "expire-bucket",
 				Status: "Enabled",
--- a/components/report-manager-service/worker/worker.go
+++ b/components/report-manager-service/worker/worker.go
@@ -25,6 +25,7 @@ import (
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/minio/minio-go/v7"
+	"github.com/minio/minio-go/v7/pkg/lifecycle"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/protobuf/types/known/structpb"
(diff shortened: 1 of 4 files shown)


components/report-manager-service/worker/worker.go
```diff
@@ -249,6 +250,21 @@ func (t *GenerateReportTask) Run(ctx context.Context, task cereal.Task) (interfa
...
+		{
+			ID:     "expire-bucket",
+			Status: "Enabled",
+			Expiration: lifecycle.Expiration{
+				Days: 1,
+			},
+		},
+	}
+
+	err = t.ObjStoreClient.SetBucketLifecycle(context.Background(), job.RequestToProcess.RequestorId, config)
```
Dmaddu, line 264: Here we are setting life cycles rules always, we should set only if the bucket does not have the policy. So please check the existing policy and apply only if the bucket doesn't have any expiration policy.

Dmaddu (changes requested):

components/report-manager-service/worker/worker.go
```diff
@@ -248,6 +249,28 @@ func (t *GenerateReportTask) Run(ctx context.Context, task cereal.Task) (interfa
 			return nil, err
 		}
 	}
+	lifecycles, err := t.ObjStoreClient.GetBucketLifecycle(ctx, job.RequestToProcess.RequestorId)
+	// fmt.Println(lifecycles)
+
+	if lifecycles == nil && err.Error() == "The lifecycle configuration does not exist" {
```
Dmaddu, line 255: This code executes only if the bucket doesn't have any life cycle rules. What if the bucket has life cycle rules, and doesn't have any expiration. It is possible that bucket can have object transition rules and in these cases, we have to create an expiration rule as well.

Dmaddu, line 255: Does the GetBuckeLifeCycle return error if the bucket doesn't have any life cycle configuration?

components/report-manager-service/worker/worker_test.go
```diff
@@ -608,6 +627,8 @@ func TestRun(t *testing.T) {
 		isPresignedGetObjectFail        bool
 		isCSVExporter                   bool
 		isObjectNotAvailableInDataStore bool
+		isSetBucketLifeCycleFail        bool
+		isGetBucketLifeCycleFail        bool
 	}{
 		{
 			name:          "testRun_Success",
```
Dmaddu, line 634: Please check the unit tests, those should not fail.

Dmaddu (changes requested):

components/report-manager-service/worker/worker.go
```diff
@@ -249,6 +250,53 @@ func (t *GenerateReportTask) Run(ctx context.Context, task cereal.Task) (interfa
...
+			return nil, err
+		}
+	} else if err != nil {
+		return nil, err
+	}
+
+	if lifecycles != nil {
+		flag := false
+		for i := 0; i < len(lifecycles.Rules); i++ {
+			if !lifecycles.Rules[i].Expiration.IsNull() {
```
Dmaddu, line 279: The condition should check for Date and Days as well.
 some thing like Expiration.IsDateNull() || Expiration.IsDaysNull()

components/report-manager-service/worker/worker.go
```diff
@@ -249,6 +250,53 @@ func (t *GenerateReportTask) Run(ctx context.Context, task cereal.Task) (interfa
...
+				flag = true
+				break
+			}
+		}
+		if !flag {
+			newRule := lifecycle.Rule{
+				ID:     "expire-bucket2",
```
Dmaddu, line 286: what is the reason behind giving different ids ?

components/report-manager-service/worker/worker_test.go
```diff
@@ -560,6 +564,45 @@ func (m mockObjStore) MakeBucket(ctx context.Context, bucketName string, opts mi
...
+			Expiration: lifecycle.Expiration{
+				Days: 1,
+			},
+		},
+	}
+	if m.ForBucketLifeCycleWithoutExpiry {
+		newRule := lifecycle.Rule{
+			ID:     "expire-bucket2",
+			Status: "Enabled",
+			Expiration: lifecycle.Expiration{
```
Dmaddu, line 596: ForBucketLifeCycleWithoutExpiry case, there should not be any expiration.

Dmaddu (approved):

PR #6621, opened 2022-01-25
upgrading to minor/patch version

Dmaddu:
create hart files of deployment-service and automate-cli, and patch to any running automate instance.

**All PRs** must tick these:


components/automate-deployment/pkg/manifest/provider.go
```diff
@@ -174,3 +188,258 @@ func (provider *LocalHartManifestProvider) overlayLocalHarts(baseManifest *A2) (
...
+			if nextMajorVersion != "" && index == len(list)-1 {
+				return index, list[index]
+			}
+		}
+	}
+	return 0, ""
+}
+
+//GetAllVersions gives the list of all released versions of given channel in ascending order of version numbers
+func GetAllVersions(ctx context.Context, channel string, optionalURL ...string) ([]string, error) {
```
rickmarry, line 314: should this live in http.go along with the other manifest urls,  instead of here?

components/automate-deployment/pkg/manifest/provider.go
```diff
@@ -2,21 +2,31 @@ package manifest
...
 
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
 	"github.com/chef/automate/components/automate-deployment/pkg/habpkg"
 	"github.com/chef/automate/components/automate-deployment/pkg/services"
 )
 
+const (
+	automateVersionsURLFmt = "https://packages.chef.io/manifests/%s/automate/versions.json"
```
rickmarry, line 22: should this live in http.go along with the other manifest urls,  instead of here?

rickmarry (commented):
just one question about the placement of the versions url..

components/automate-deployment/pkg/manifest/provider.go
Dmaddu: No @rickmarry, that should go to http.go only when we are implementing this at the client level. But here we are implementing at the manifest provider level, and not at the client level, so this should be here only.

components/automate-deployment/pkg/manifest/provider.go
Dmaddu: No @rickmarry, that should go to http.go only when we are implementing this at the client level. But here we are implementing at the manifest provider level, and not at the client level, so this should be here only.

PR #6756, opened 2022-03-03
fix infinite render loop in checkin history component

rishichawda:
Signed-off-by: rishichawda <rishichawda@users.noreply.github.com>

While investigating an unrelated issue, I found a piece of code that raised alarms since it was modifying objects from the  rather than returning a new one. As a best practice, we avoid mutating objects in javascript. Along with that, the getter was being used inside a directive which means it is called indefinitely. I've fixed these two issues by 

- using [trackByFunction](https://angular.io/api/core/TrackByFunction) to avoid multiple (basically infinite) render cycles. 
- return a new history object instead of modifying the original in bucket array. (I'm not sure if this was being intentionally done and there's no comment either so I'm assuming it wasn't)

**All PRs** must tick these:

- [ ] Is the code clear? *(complicated code or lots of comments--subdivide and use well-named methods, meaningful variable names, etc.)*
- [ ] Consistency checked? *(user notifications, user prompts, visual patterns, code patterns, variable names)*
- [ ] Repeated code blocks eliminated? *(adapt and reuse existing components, blocks, functions, etc.)*
- [ ] Spelling, grammar, typos checked? *(at a minimum use `make spell` in any component directory)*
- [ ] Code well-formatted? *(indents, line breaks, etc. improve rather than hinder readability)*

- [ ] Tests added/updated? (all new code needs new tests)

### DIFF
--- a/components/automate-ui/src/app/modules/desktop/desktop-detail/desktop-detail.component.html
+++ b/components/automate-ui/src/app/modules/desktop/desktop-detail/desktop-detail.component.html
@@ -43,7 +43,7 @@
         </chef-tr>
       </chef-thead>
       <chef-tbody>
-        <chef-tr *ngFor="let history of labeledCheckInHistory; index as i">
+        <chef-tr *ngFor="let history of labeledCheckInHistory; index as i; trackBy: trackByFunction">
           <chef-td class="status-cell">
             <chef-icon [id]="'checkin-item-'+i+'text'" [ngClass]="history.status">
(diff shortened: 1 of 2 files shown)


rishichawda:
I'm not sure if similar issues exist in other components. I just happened to stumble upon this while looking into something else. But it is worth checking and fixing it as a performance improvement effort. 👍🏼

shaik80 (approved):

PR #7478, opened 2022-10-13
Kratos 180

rishabhjhs:

**All PRs** must tick these:

- [ ] All commits signed-off for the [Developer Certification of Origin](https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco).

- [ ] Is the code clear? *(complicated code or lots of comments--subdivide and use well-named methods, meaningful variable names, etc.)*
- [ ] Consistency checked? *(user notifications, user prompts, visual patterns, code patterns, variable names)*
- [ ] Repeated code blocks eliminated? *(adapt and reuse existing components, blocks, functions, etc.)*
- [ ] Spelling, grammar, typos checked? *(at a minimum use `make spell` in any component directory)*
- [ ] Code well-formatted? *(indents, line breaks, etc. improve rather than hinder readability)*

- [ ] Tests added/updated? (all new code needs new tests)


vviveksharma (approved):

components/automate-gateway/handler/ssoconfig.go
```diff
@@ -72,6 +69,54 @@ func (a *SsoConfig) GetSsoConfig(ctx context.Context, in *empty.Empty) (*sso.Get
...
+			return nil, err
+		}
+		go makeRequest("DELETE", *url, nil, fileName)
+		return &sso.DeleteSsoConfigResponse{
+            Message: "Started Disabling SSO Configuration",
+        }, nil
+    }
+
+	return &sso.DeleteSsoConfigResponse{
+		Message: "SSO Configuration not disabled successfully",
```
sandhi18, line 103: Please change the message to Dex config not present or SSO config not present

sandhi18 (changes requested):

components/automate-gateway/handler/ssoconfig.go
rishabhjhs: Fixed

vivek-yadav:
Questions:

1. Why only SaaS customers should have this feature?
1. How New GO server will be deployed?
  So far bastion was used only for administrative purpose. Now, we are changing its purpose and need active all times?
1. How current configs are applied from Automate like data retention etc..
1. What happen if SSH is taking longer than expected?
  For all machines
  Applied successfully to some but failed to others and in this case how Get SSO works?
  Is it synchronous/asynchronous call?
  Is it sone in parallel or sequence?
1. Why do we need this logic in Automate Gateway, as normally this work is done by a Service Component.
1. We can create a wrapper service to talk to Bastion Service if needed. So the Automate gateway handler doesn't do it.
1. This script is knowing too much about SSO config values, even though that should not be its concern. Better keep the logic of SSO config at one location in Automate Service
1. How long will the api to patch SSO config will take? and do we expect UI to wait for that period?
1. What will happen the config is wrong and SSO login will fail, can the user test it in UI before applying this config ?

components/automate-gateway/handler/ssoconfig.go
```diff
@@ -72,6 +69,52 @@ func (a *SsoConfig) GetSsoConfig(ctx context.Context, in *empty.Empty) (*sso.Get
...
+		return nil, err
+	}
+	
+	if res.Config.Dex != nil {
+		url, err := getBastionUrl()
+		if err != nil {
+			return nil, err
+		}
+		fileName := "revert-status.txt"
+		err = ioutil.WriteFile(ssoFilesPath+fileName, []byte("Pending"), 0777)
```
vivek-yadav, line 91: As I see this is a file which will save the state of the Revert Task.
Which means this file will be checked for status later, to know if the task is finished or not.

But here is my question, if check status is going to be called from UI via API, then if we had more than 1 Automate Node, then the status only going to be available in only one Automate Node, not Both (assuming this file is only the source of the revert task state).

This makes me wonder if this design had considered the Multiple Node of Automate and how state will be maintained across them. If this is considered in the design please highlight it.

@sandhi18  @karunesh-tech  ^

PR #7979, opened 2023-06-21
Added chages for nil pointer error due to uninitialized Config fields

daveaugustus:
Config struct was being called with some uninitialised fields which was causing nil pointer error. 
MicrosoftTeams-image8a8a976ac2328997e8ae761641c2e0bf8717337b9ce934fdfe3a49ed25703038

### :+1: Definition of Done
- All pointer fields should be initialised 

**All PRs** must tick these:

- [x] Docs added/updated? (all customer-facing changes)

### DIFF
--- a/components/automate-cli/pkg/verifyserver/models/batchcheck.go
+++ b/components/automate-cli/pkg/verifyserver/models/batchcheck.go
@@ -137,6 +137,19 @@ func appendCertsByIpToNodeCerts(certsByIP *[]config.CertByIP, ipList []string, p
 
 }
 
+func NewConfig() *Config {
+	return &Config{
+		SSHUser: &SSHUser{},
+		Backup: &Backup{
+			FileSystem:    &FileSystem{},
+			ObjectStorage: &ObjectStorage{},
+		},
+		Hardware:   &Hardware{},
+		ExternalOS: &ExternalOS{},
+		ExternalPG: &ExternalPG{},
+	}
+}
+
 func createMapforCertByIp(certsByIP *[]config.CertByIP) map[string]*config.CertByIP {
 	certByIPMap := make(map[string]*config.CertByIP)
 	if certsByIP != nil {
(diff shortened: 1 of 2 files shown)


components/automate-cli/pkg/verifyserver/models/batchcheck.go
```diff
@@ -123,6 +123,19 @@ func appendCertsByIpToNodeCerts(certsByIP *[]config.CertByIP, ipList []string, p
...
 }
 
+func NewConfig() *Config {
+	return &Config{
+		SSHUser: &SSHUser{},
+		Backup: &Backup{
+			FileSystem:    &FileSystem{},
+			ObjectStorage: &ObjectStorage{},
+		},
+		Hardware:   &Hardware{},
```
YashviJain01, line 133: We should add all the objects initialisation, like certificate?

daveaugustus: They will take their respective empty values, we just have to have types of objects initialized...

components/automate-cli/pkg/verifyserver/models/batchcheck.go
```diff
@@ -123,6 +123,24 @@ func appendCertsByIpToNodeCerts(certsByIP *[]config.CertByIP, ipList []string, p
 
 }
 
+func NewConfig() *Config {
+	return &Config{
+		SSHUser: &SSHUser{},
```
Dmaddu, line 128: Instead of initializing here, how about creating the instance at the config parser level? With this approach how the trigger knows the value is nil or empty? We have to handle each cases differently.
CC: @YashviJain01  @kalroy

Dmaddu (changes requested):
Lets have a discussion on this.

components/automate-cli/pkg/verifyserver/models/batchcheck.go
kalroy: Yes. If we can enforce that the struct can be created only by this function then it is fine. But if not, we should move it to individual functions in the parser so that we do not get null pointer exception

daveaugustus: @Dmaddu what do you suggest?

